PR #4, opened 2023-02-15
hotfix serp api for jest

lalalune:
This is a PR to fix the bug https://github.com/hwchase17/langchainjs/issues/3

This is a hotfix to get tests to pass.

After install, a postinstall script is run that edits the jsdomexceptions npm module. SerpApi depends on jsdom, which depends on jsdomexceptions, which is throwing an error when it trying to get the prototype of the prototype of a function. The script edits line 34 of the utils.js script in that module to only unwrap once, ignoring the error. Unless we have deep exceptions occurring, which I doubt, I think it's fine.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,8 @@
     "test": "turbo run test",
     "publish": "turbo run build lint test && yarn workspace langchain npm publish",
     "example": "turbo run start --filter langchain-examples --",
-    "prepare": "husky install"
+    "prepare": "husky install",
+    "postinstall": "node ./scripts/postinstall.js"
   },
   "author": "Langchain",
   "license": "MIT",
--- a/scripts/postinstall.js
+++ b/scripts/postinstall.js
@@ -1,19 +1,20 @@
 // postinstall functions to handle any tasks, fixes or hacks that may arise as a result of dependency issues
 
-function postinstall() {
+(function postinstall() {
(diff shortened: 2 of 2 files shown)


sullivan-sean (approved):

sebastianquek:
@lalalune @sullivan-sean How about setting the `@babel/preset-env` target to node instead since likely you want langchain to run on node?

Ref: https://github.com/serpapi/serpapi-javascript/issues/10#issuecomment-1430878401

sebastianquek:
See https://github.com/serpapi/serpapi-javascript/issues/10,
> I've updated serpapi to 1.1.1 which removes the production dependency on domexception. Let me know if you're still facing issues after the update. Will close this for now.

sullivan-sean:
> See [serpapi/serpapi-javascript#10](https://github.com/serpapi/serpapi-javascript/issues/10),
> 
> > I've updated serpapi to 1.1.1 which removes the production dependency on domexception. Let me know if you're still facing issues after the update. Will close this for now.
> 
> Nonetheless, I still think setting `preset-env` to target the current node version makes sense since this is meant to be ran in a node environment 🙂

Thanks I've bumped serpapi for now - we plan on supporting a subset of functionality browser side, but I'll also update the `preset-env` to node as well!

PR #8, opened 2023-02-15
Add VectorStore base class and HNSWLib vector store

nfcampos:
This implements a vector store using HNSWLib https://github.com/nmslib/hnswlib, using the node bindings at https://github.com/yoshoku/hnswlib-node

HNSWLib is the ANN library used by Chroma

### DIFF
--- a/langchain/embeddings/base.ts
+++ b/langchain/embeddings/base.ts
@@ -1,0 +1,5 @@
+export abstract class Embeddings {
+  abstract embedDocuments(documents: string[]): Promise<number[][]>;
+
+  abstract embedQuery(document: string): Promise<number[]>;
+}
--- a/langchain/embeddings/base.ts
+++ b/langchain/embeddings/base.ts
@@ -1,6 +1,4 @@
 export abstract class Embeddings {
-  abstract numDimensions: number;
-
   abstract embedDocuments(documents: string[]): Promise<number[][]>;
 
   abstract embedQuery(document: string): Promise<number[]>;
--- a/langchain/embeddings/base.ts
+++ b/langchain/embeddings/base.ts
@@ -1,4 +1,6 @@
 export abstract class Embeddings {
+  abstract numDimensions: number;
+
   abstract embedDocuments(documents: string[]): Promise<number[][]>;
 
   abstract embedQuery(document: string): Promise<number[]>;
--- a/langchain/embeddings/openai.ts
+++ b/langchain/embeddings/openai.ts
@@ -1,0 +1,91 @@
+import type {
+  Configuration as ConfigurationT,
+  OpenAIApi as OpenAIApiT,
+  CreateEmbeddingRequest,
+} from "openai";
+import { backOff } from "exponential-backoff";
+import { chunkArray } from "../util";
+import { Embeddings } from "./base";
+
+let Configuration: typeof ConfigurationT | null = null;
+let OpenAIApi: typeof OpenAIApiT | null = null;
+
+try {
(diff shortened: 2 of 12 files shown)


langchain/embeddings/base.ts
```diff
@@ -1,4 +1,6 @@
 export abstract class Embeddings {
+  abstract numDimensions: number;
```
nfcampos, line 2: This doesn't exist in the python library, but it felt right? Happy to change however

langchain/vectorstores/base.ts
```diff
@@ -0,0 +1,56 @@
...
+      await this.embeddings.embedQuery(query),
+      k
+    );
+  }
+}
+
+export abstract class SaveableVectorStore extends VectorStore {
+  abstract save(directory: string): Promise<void>;
+
+  static load(
```
nfcampos, line 50: Static methods cannot be abstract in TS

langchain/embeddings/base.ts
hwchase17: depends on how we do it imo. see comment below

langchain/embeddings/openai.ts
```diff
@@ -24,6 +24,8 @@ interface ModelParams {
 export class OpenAIEmbeddings extends Embeddings implements ModelParams {
   modelName = "text-embedding-ada-002";
 
+  numDimensions = 1536;
```
hwchase17, line 27: is it possible to avoid having to pass this in? Most people won't want to think about this when initializing an embedding class

its also tied to the model that is passed in. by specifying both (esp if there are defaults) it makes it real easy to get out of date

langchain/vectorstores/tests/hnswlib.test.ts
```diff
@@ -0,0 +1,59 @@
+import { test, expect } from "@jest/globals";
```
hwchase17, line 1: probably good to standardize on convention for whether test files are in a test directory or just their own files?

langchain/embeddings/openai.ts
nfcampos: Well we need this to be set somewhere, as it is necessary to know this number in order to create any kind of ANN index. We can magically read the length of the vector of the first text that you try to add to the vector store and use that, but feels really strange to do that?

langchain/vectorstores/tests/hnswlib.test.ts
nfcampos: I was just following what Sean had done in the other ones, I'm good with whatever

hwchase17: ah got it, i had thought hed done something else. never mind then

langchain/embeddings/openai.ts
hwchase17: thats what we do in python - it feels a bit strange but it makes it really easy for users not to have to think about embedding lengths

hwchase17: other option would be to have a map of embedding model to embedding length. but then becomes a pain to update that map...

nfcampos: Okay, a map would go out of date, I can do the magical one, but will throw an error in addTexts if user tries to change embeddings sizes halfway

hwchase17: thats fine!

nfcampos: Updated now

hwchase17 (approved):
looks good to me!

PR #12, opened 2023-02-15
add stuff chain and qa chain

hwchase17:


langchain/chains/combine_docs_chain.ts
```diff
@@ -0,0 +1,76 @@
...
+  async _call(values: ChainValues): Promise<ChainValues> {
+    if (!(this.inputKey in values)) {
+        throw new Error(`Document key ${  this.inputKey  } not found.`);
+      }
+    const docs: Document[] = values[this.inputKey];
+    const texts = docs.map(({ pageContent }) => pageContent);
+    const text = texts.join("\n\n");
+    delete values[this.inputKey];
+    values[this.documentVariableName] = text;
+    const result = this.llmChain.call(values);
```
nfcampos, line 51: Might be missing an await here?

hwchase17: good catch

PR #16, opened 2023-02-16
fix: add initial documentation with typedoc

sullivan-sean:


langchain/package.json
```diff
@@ -67,6 +36,7 @@
     "eslint-config-prettier": "^8.6.0",
     "eslint-plugin-import": "^2.27.5",
     "eslint-plugin-prettier": "^4.2.1",
+    "hnswlib-node": "^1.2.0",
```
nfcampos, line 39: Making this a required dependency on purpose?

PR #20, opened 2023-02-16
Harrison/add getting started

hwchase17:


docs/docs/getting-started.md
```diff
@@ -1,3 +1,264 @@
...
+
+```typescript
+const model = new OpenAI({temperature: 0.9});
+```
+
+We can now call it on some input!
+
+```typescript
+const res = await model.call("What would be a good company name a company that makes colorful socks?");
+console.log({ res });
```
nfcampos, line 64: You can remove the curly braces here maybe to make it simpler?

langchain/chains/conversation.ts
```diff
@@ -1,10 +1,10 @@
 import { LLMChain } from "./index";
 import { BaseLLM } from "../llms";
-import { BasePromptTemplate , PromptTemplate } from "../prompt";
+import { BasePromptTemplate, PromptTemplate } from "../prompt";
 
 import { BaseMemory, BufferMemory } from "../memory";
 
-
+// eslint-disable-next-line max-len
```
nfcampos, line 7: We might want to remove this eslint rule? Max length is what prettier is here for. Thoughts? @sullivan-sean

langchain/tsconfig.json
```diff
@@ -32,7 +32,8 @@
       "./prompt/index.ts",
       "./vectorstores/index.ts",
       "./text_splitter.ts",
-      "./index.ts"
+      "./index.ts",
+      "./memory/index.ts"
```
nfcampos, line 36: There's a gotcha with this config here for typedoc, the root index.ts needs to always be the last file in the list, otherwise it misbehaves

langchain/chains/conversation.ts
sullivan-sean: Sure, I'm not picky about linting let's leave this to prettier!

langchain/tsconfig.json
sullivan-sean: If only we could comment in `.json` 🙄 it's also annoying to duplicate this in the `package.json` and `tsconfig.json`, ideally the packages endpoint strategy would just pick up on our entrypoints automatically: https://typedoc.org/api/enums/EntryPointStrategy.html#Packages. Might be worth trying to upstream something

hwchase17: roger

PR #35, opened 2023-02-17
hnswlib: Throw error in index getter to dedupe code and have safer usage

micahriggan:
Feel free to close, but seems like we can provide safer access to .index with a getter that throws if it hasn't been init'd 

Also added a static method to init the HNSW Lib so we don't have to dupe the error handling each time

Oh yeah, thanks for making this, super cool

### DIFF
--- a/langchain/src/vectorstores/hnswlib.ts
+++ b/langchain/src/vectorstores/hnswlib.ts
@@ -24,7 +24,7 @@ export interface HNSWLibArgs {
 }
 
 export class HNSWLib extends SaveableVectorStore {
-  index?: HierarchicalNSWT;
+  _index?: HierarchicalNSWT;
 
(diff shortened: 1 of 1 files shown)


sullivan-sean:
Great call! This cleans things up nicely.

nfcampos:
Another option would be to make the index property private, if you need access to it you can always create the instance with an index created outside?

micahriggan:
Yeah you could make it private, and you could use 

```HNSWLib.getHierarchicalNSW(args)``` in the other files instead of accessing index. You might have to re-init the index that way though. I only noticed 2 places that were external to this file accessing the index property, but I didn't dig into whether they relied on it being initialized already

nfcampos:
Thanks so much for this @micahriggan , I've rebased this and pulling this in now

PR #36, opened 2023-02-17
Implement OpenAI LLM completion streaming

dqbd:
This PR implements streaming of inner completion results via `callbackManager`.

Similar PR to the Python counterpart: https://github.com/hwchase17/langchain/pull/986

*Usage*:
```typescript
const model = new OpenAI({
  streaming: true,
  callbackManager: {
    handleNewToken(token) {
      ...
    },
  },
});
```


sullivan-sean (approved):
perfect, looks good to go! thanks for including an example

PR #42, opened 2023-02-18
Add CI job (and make it green)

nfcampos:
- formatted all files that weren't formatted
- using the same prettier config in all packages
- separate unit and integration tests
- fix prettier on windows, do some setup for dev on windows machines
- fix `loadFromHub` to work on Windows


langchain/util/hub.ts
```diff
@@ -11,6 +11,8 @@ const URL_BASE =
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export type LoadValues = Record<string, any>;
 
+const URL_PATH_SEPARATOR = "/";
```
nfcampos, line 14: These changes make it work on Windows. filesystem path separator is different on windows, but remote URL path separator isn't

.editorconfig
```diff
@@ -0,0 +1,6 @@
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
```
nfcampos, line 6: For folks on windows to not clash with the linux/max defaults

PR #46, opened 2023-02-18
Added OpenAI CustomParameters to OpenAI constructor

chitalian:
Add configuration parameter to the constructor of OpenAI so that users can have more control of their

### DIFF
--- a/examples/src/customParameters/differentBaseUrl.ts
+++ b/examples/src/customParameters/differentBaseUrl.ts
@@ -1,0 +1,14 @@
+import { OpenAI } from "langchain";
+
+export const run = async () => {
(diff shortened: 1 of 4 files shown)


sullivan-sean:
We should add configuration (or some subset) to `identifyingParams` which is used for serializing/deserializing the model as well as caching generation results.

Caching in particular might result in stale results if somebody has two similar models that differ only in configuration.

chitalian:
That is a really good idea, should we do this in a follow-up PR? @sullivan-sean ?

sullivan-sean:
I think it makes sense to do atomically in this PR to avoid any caching issues

PR #47, opened 2023-02-18
Custom params on generate

chitalian:
Allows the ability to add custom parameters within your OpenAI generate calls
```typescript

  const res = await model.call(
    "What would be a good company name a company that makes colorful socks?",
    undefined,
    {
      headers: {
        "User-Id": "123",
      },
    }
  );
  console.log(res);
};
```


langchain/llms/openai.ts
```diff
@@ -87,6 +89,8 @@ type TokenUsage = {
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 type Kwargs = Record<string, any>;
 
+export type OpenAIRequestConfig = AxiosRequestConfig;
```
sullivan-sean, line 92: Wondering if we should nest the axios request config in an object type here to future-proof the third arg of generate for any other things we might want to add in the future, e.g.

```suggestion
export type OpenAIExtraArgs = {
    requestConfig: AxiosRequestConfig
};
```

sullivan-sean (commented):
LGTM pending type change here once https://github.com/hwchase17/langchainjs/pull/46 is merged

langchain/llms/openai.ts
chitalian: ahhh totally agree. We should also make it optional as well

nfcampos:
Note to me: combine #263

agola11 (approved):

PR #48, opened 2023-02-19
Add support for Pinecone vectorstore

dqbd:
Uses https://github.com/rileytomasek/pinecone-client as the client for Pinecone. 

The [official client](https://github.com/pinecone-io/pinecone-ts-client) is currently marked as `beta` and uses `axios`, which may further prevent us to support Edge runtime etc.

### DIFF
--- a/langchain/package.json
+++ b/langchain/package.json
@@ -81,6 +81,11 @@
   "peerDependencies": {
     "pinecone-client": "^1.0.1"
   },
+  "peerDependenciesMeta": {
+    "pinecone-client": {
+      "optional": true
+    }
+  },
   "dependencies": {
     "deepcopy": "^2.1.0",
     "eventsource-parser": "^0.1.0",
--- a/langchain/package.json
+++ b/langchain/package.json
@@ -54,6 +54,7 @@
     "@jest/globals": "^29.4.2",
     "@tsconfig/recommended": "^1.0.2",
     "@types/node-fetch": "2",
+    "@types/uuid": "^9",
     "@typescript-eslint/eslint-plugin": "^5.51.0",
     "@typescript-eslint/parser": "^5.51.0",
     "babel-jest": "^29.4.2",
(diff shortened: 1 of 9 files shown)


langchain/vectorstores/pinecone.ts
```diff
@@ -0,0 +1,116 @@
+import type { PineconeClient } from "pinecone-client";
```
sullivan-sean, line 1: We should make sure these imports don't fail if `pinecone-client` is not installed. We do this in a few other places as well for optional deps (e.g. see https://github.com/hwchase17/langchainjs/blob/80356338749618620cefdd1630120a570f5f1f62/langchain/llms/openai.ts#L18-L23)

langchain/package.json
```diff
@@ -77,12 +78,16 @@
     "typedoc-plugin-missing-exports": "^1.0.0",
     "typescript": "^4.9.5"
   },
+  "peerDependencies": {
+    "pinecone-client": "^1.0.1"
+  },
```
sullivan-sean, line 83: Let's add `peerDependenciesMeta` here too (https://docs.npmjs.com/cli/v9/configuring-npm/package-json#peerdependenciesmeta) so that langchain doesn't cause npm to emit warnings for users who aren't using pinecone.

sullivan-sean (changes requested):
Looks great, thanks for adding!

langchain/vectorstores/pinecone.ts
dqbd: I think in this case it should be fine to keep it as-is, as we're passing the instance of the client as an argument, not creating the instance ourselves. 

The type import is removed by `tsc` during build phase and the resulting JS file does not import "pinecone-client" at all.

CleanShot 2023-02-19 at 02 57 26@2x

dqbd: It might be the case that we need to actually create the instance ourselves, possibly due to serialisation? @sullivan-sean

sullivan-sean: oh yeah I see, this is fine!

sullivan-sean (approved):

PR #49, opened 2023-02-19
update examples to use template literals

samheutmaker:


examples/src/agents/load_from_hub.ts
```diff
@@ -18,9 +18,10 @@ export const run = async () => {
     returnIntermediateSteps: true,
   });
 
-  const input =
-    "Who is Olivia Wilde's boyfriend?" +
-    " What is his current age raised to the 0.23 power?";
+  const input = `
+    Who is Olivia Wilde's boyfriend?
```
sullivan-sean, line 22: This change will add tab/newline artifacts to the prompt. We can honestly just remove the `+` here and combine it into a single string. 

The only reason this is split across two lines is for the `max-len` lint rule which we've since removed from the primary `langchain` package, can be removed from `examples` as well if it isn't already.

sullivan-sean (changes requested):

PR #50, opened 2023-02-19
Add Docstore, InMemoryDocStore, update tests and other modules

samheutmaker:
Here is an attempt to add a base docstore implementation and an in-memory docstore using a TS Map. I moved a few files around to better match the file structure of python langchain. All feedback is appreciated :)


langchain/vectorstores/hnswlib.ts
```diff
@@ -153,12 +152,13 @@ export class HNSWLib extends SaveableVectorStore {
...
     }
-    return HNSWLib.fromDocuments(docs, embeddings);
+    return HNSWLib.fromDocuments(docs, embeddings, docstore);
   }
 
   static async fromDocuments(
     docs: Document[],
-    embeddings: Embeddings
+    embeddings: Embeddings,
+    docstore: Docstore,
```
hwchase17, line 161: is there anyway to have this be an optional param, and then if its not passed in just initialize with a `new InMemoryDocstore` here?

would make it a bit easier to not have to pass it in everytime

samheutmaker: This is a great idea, changes incoming.

samheutmaker: > is there anyway to have this be an optional param, and then if its not passed in just initialize with a `new InMemoryDocstore` here?
> 
> would make it a bit easier to not have to pass it in everytime

Done.

PR #53, opened 2023-02-19
feat: add callbacks for chains and agents

sullivan-sean:


langchain/callbacks/index.ts
```diff
@@ -0,0 +1,267 @@
...
+  setHandlers: (handlers: CallbackHandler[]) => void;
+}
+
+const isEventType = <T extends string>(
+  prefix: T,
+  event: CallbackEvent
+): event is CallbackEvent & { event: `${T}.${string}` } =>
+  event.event.startsWith(prefix);
+
+export abstract class EventTypeCallbackHandler extends BaseCallbackHandler {
```
nfcampos, line 143: I think we should make all handlers async (and by extension the manager too)

sullivan-sean: I think in most cases they are async already? (returning `Promise<void> | void`)

Looks like i missed it on the `PerEventCallbackHandler` object but should be easy fix

langchain/llms/base.ts
```diff
@@ -33,15 +26,26 @@ export abstract class BaseLLM {
...
    */
   verbose?: boolean = false;
 
-  constructor(callbackManager?: LLMCallbackManager, verbose?: boolean) {
-    this.callbackManager = callbackManager ?? getCallbackManager();
+  constructor(
+    callbackManager?: LLMCallbackManager | BaseCallbackManager,
+    verbose?: boolean
+  ) {
+    if (!callbackManager) {
```
hwchase17, line 40: some comment here explaining this would probably be nice

hwchase17 (commented):
high level:

- we recently updated the langchain code to have a concept of a tool class (which this package does already) and then we moved the callbacks from the agent executor to the tool class (added callbacks before/after tool.call, similar to how we do for chains)
-

langchain/agents/executor.ts
```diff
@@ -74,9 +84,33 @@ export class AgentExecutor extends BaseChain {
...
+          : `${action.tool} is not a valid tool, try another one.`;
+      } catch (err) {
+        this.callbackManager(
+          { event: "agent.tool_error", err },
+          { verbose: this.verbose }
+        );
+        throw err;
+      }
+
+      this.callbackManager(
```
nfcampos, line 109: With async I meant even this method should be async, otherwise we end up dangling promises

agola11 (commented):
@sullivan-sean -- since this PR is a bit stale with lots of merge conflicts, I'm going to take pieces of this and apply them in a new PR, might make some changes here and there. Hope that's okay with you. Will add you as a collaborator on the commit

PR #58, opened 2023-02-19
Create BaseDocumentLoader, BaseCheerioLoader, and loaders

samheutmaker:
This PR is an attempt to add the foundational code needed for loaders. It also adds a BaseCheerioLoader that uses [cheerio](https://www.npmjs.com/package/cheerio) in place of BeautifulSoup for dom-traversal. It builds on the cheerio loader to add loaders for:

- Hacker News
- College Confidential 
- AZLyrics
- IMSDB

It also adds  loaders for:
- Text Files
- SRT Files
- YouTube Videos

These loaders function identically to the corresponding loaders in Python LangChain (PyLangChain?). This PR is dependent on previous work in PR #50. 

Docs and examples to follow.


langchain/document_loaders/text.ts
```diff
@@ -0,0 +1,40 @@
+import type { readFile as ReadFileT } from 'fs/promises';
+import { Document } from "../docstore/document";
+import { BaseDocumentLoader } from "./base";
+
+let readFile: typeof ReadFileT | null = null;
+
+try {
+  // eslint-disable-next-line global-require,import/no-extraneous-dependencies
+  ({ readFile } = require("fs/promises"));
```
hwchase17, line 9: is `fs` package not a dependency/built in? if not, should we just make it one? how heavy is it - seems like should be light?

hwchase17 (commented):
this is AWESOME

my gut is we can probably just make the `fs` a normal dependency? dont know much about js packages but seems like it should be pretty light?

langchain/document_loaders/text.ts
samheutmaker: `fs` is only available in Node.js and not [V8 environments](https://edge-runtime.vercel.app/features/available-apis). Importing `fs` normally prevent users from deploying apps to Cloudflare workers and Vercel Edge functions,  which is a pretty common way to serve these apps. There's no downside to importing `fs` this way as far as I know, though it may be worth looking at this pattern more holistically and deciding if V8 support is a priority.

nfcampos: I think supporting deployment to Cloudflare workers is a worthwhile goal, but I'm sure there's other things in the library we'd need to change to make that possible, eg. the "load from hub" functionality

samheutmaker: There is a way to package TS libs in a manner that allows for different environments. Some features, like reading/writing files on a disk, would be disabled in V8 environments, but these features would work normally in Node.

I can do some digging to see what it would take to implement this at a build/package level.

PR #63, opened 2023-02-19
docs: changed notebook -> tutorial to keep consistent language in getting started guide

zaarheed:
The getting started guide references itself as a tutorial so the word notebook suggests copy-pasta from the Python lib


hwchase17 (approved):
oops thanks!

PR #66, opened 2023-02-20
Harrison/loaders

hwchase17:


langchain/package.json
```diff
@@ -58,6 +58,7 @@
     "@typescript-eslint/eslint-plugin": "^5.51.0",
     "@typescript-eslint/parser": "^5.51.0",
     "babel-jest": "^29.4.2",
+    "cheerio": "^1.0.0-rc.12",
```
sullivan-sean, line 61: Would be good to put these optional dependencies in `peerDependencies`/`peerDependenciesMeta` so that any consuming packages have ~the same version as we used in dev (and get a warning if not)

langchain/vectorstores/tests/hnswlib.int.test.ts
```diff
@@ -46,6 +46,28 @@ test("Test HNSWLib.fromTexts + addDocuments", async () => {
   expect(resultTwoMetadatas).toEqual([{ id: 2 }, { id: 1 }, { id: 3 }]);
 });
 
+test("Test HNSWLib.load and HNSWLib.save", async () => {
```
nfcampos, line 49: Wrong test name?

langchain/document_loaders/youtube.ts
```diff
@@ -0,0 +1,109 @@
...
+      throw new Error(
+        "Please install youtube-transcript as a dependency with, e.g. `yarn add youtube-transcript`"
+      );
+    }
+
+    this.videoId = videoId;
+    this.addVideoInfo = addVideoInfo;
+  }
+
+  public static fromYoutubeUrl(
```
nfcampos, line 50: fyi `public` is the default so usually in TS/JS folks don't write it

langchain/document_loaders/youtube.ts
```diff
@@ -0,0 +1,109 @@
...
+  // eslint-disable-next-line global-require,import/no-extraneous-dependencies
+  ytdl = require("ytdl-core");
+  // eslint-disable-next-line global-require,import/no-extraneous-dependencies
+  YouTubeTranscript = require("youtube-transcript");
+} catch {
+  // ignore error, will be throw in constructor
+}
+
+export class YouTubeLoader extends BaseDocumentLoader {
+  private videoId: string;
```
nfcampos, line 19: fyi these aren't truly private, they can still be accessed if you want to. See https://www.typescriptlang.org/docs/handbook/2/classes.html#caveats. If you want them to be truly impossible to access outside the class then this is what you want instead https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields

langchain/document_loaders/srt.ts
```diff
@@ -0,0 +1,59 @@
...
+export class SRTLoader extends BaseDocumentLoader {
+  constructor(public filePath: string) {
+    super();
+
+    /**
+     * Throw error at construction time
+     * if fs/promises is not installed.
+     */
+    if (readFile === null) {
+      throw new Error("Failed to load fs/promises.`");
```
nfcampos, line 27: Would probably be good to throw a more informative error, like

```js
if (readFile === null) {
  const { isBrowser, isNode, isWebWorker, isJsDom, isDeno } = require('browser-or-node');
  let env: string;
  if (isBrowser) {
    env = 'browser';
  } else if (isNode) {
    env = 'node';
  } else if (isWebWorker) {
    env = 'webworker';
  } else if (isJsDom) {
    env = 'jsdom';
  } else if (isDeno) {
    env = 'deno';
  }
  throw new Error(`Failed to load fs/promises. SRTLoader available only on environment 'node'. It appears you are running environment '${env}'. See https://<link to docs> for alternatives.`);
}
```

(obviously we'd want to encapsulate this in a util function we can use in multiple places

PR #69, opened 2023-02-20
add docstore entrypoints

hwchase17:


sullivan-sean:
Make sure to just add in the `langchain/create-entrypoints.js` script (see https://github.com/hwchase17/langchainjs/blob/main/CONTRIBUTING.md#adding-an-entrypoint)

PR #73, opened 2023-02-20
Add Chroma as vector store

jeffchuber:
Works with Chroma's dockerized backend. 

Feedback welcome! I tried to stay very similar to the `hnsw-node` implementation. 

The PR is open on chroma's repo to add this JS SDK - https://github.com/chroma-core/chroma/pull/150/

A fully functional version is on `npm` : https://www.npmjs.com/package/chromadb


langchain/vectorstores/chroma.ts
```diff
@@ -0,0 +1,187 @@
...
+        this.args.numDimensions = vectors[0].length;
+      }
+      if (ChromaClient === null) {
+        throw new Error(
+          "Please install chromadb as a dependency with, e.g. `npm install -S chromadb`"
+        );
+      }
+      this.index = new ChromaClient("http://localhost:8000");
+      try {
+        await this.index.createCollection(COLLECTION_NAME);
```
nfcampos, line 70: Wouldn't we want a different collection id for each instance?

jeffchuber: yeah - good point. there was no affordance for this obviously with the narrow `hnsw` version, but that should definitely be included here.

jeffchuber: This has been addressed. thanks for noting this!

langchain/package.json
```diff
@@ -58,6 +58,7 @@
     "@typescript-eslint/eslint-plugin": "^5.51.0",
     "@typescript-eslint/parser": "^5.51.0",
     "babel-jest": "^29.4.2",
+    "chromadb": "^1.2.1",
```
nfcampos, line 61: in peerDependencies like pinecone maybe?

langchain/vectorstores/chroma.ts
```diff
@@ -0,0 +1,194 @@
+import fs from "fs/promises";
+import path from "path";
+import type {
+  ChromaClient as ChromaClientT,
+} from "chromadb";
+import { ChromaClient as ChromaClientO } from "chromadb";
```
nfcampos, line 6: I think this will raise an exception if this file is imported and you don't have chromadb installed, we can probably just remove this line and rely on line 17?

langchain/vectorstores/chroma.ts
```diff
@@ -0,0 +1,194 @@
...
+  ({ ChromaClient } = require("chromadb"));
+} catch {
+  // ignore error
+}
+ChromaClient = ChromaClientO;
+
+export interface ChromaLibArgs {
+  space: string;
+  numDimensions?: number;
+  collection_name: string,
```
nfcampos, line 26: Not too particular about this, but do we want to make all the args the same case?

langchain/vectorstores/chroma.ts
```diff
@@ -0,0 +1,194 @@
...
+    await this.addVectors(await this.embeddings.embedDocuments(texts), documents);
+  }
+
+  async addVectors(vectors: number[][], documents: Document[]) {
+    if (vectors.length === 0) {
+      return;
+    }
+    if (!this.index) {
+      if (this.args.numDimensions === undefined) {
+        this.args.numDimensions = vectors[0].length;
```
nfcampos, line 63: If chromadb doesn't use this arg then Im not against just removing it, unless you think the error messages that compare it to the embedding length are worth it

langchain/vectorstores/chroma.ts
```diff
@@ -0,0 +1,194 @@
...
+      }
+      if (ChromaClient === null) {
+        throw new Error(
+          "Please install chromadb as a dependency with, e.g. `npm install -S chromadb`"
+        );
+      }
+      this.index = new ChromaClient("http://localhost:8000");
```
nfcampos, line 70: would anyone want to customise this url?

langchain/vectorstores/chroma.ts
```diff
@@ -0,0 +1,194 @@
...
+        "Please install chromadb as a dependency with, e.g. `npm install -S chromadb`"
+      );
+    }
+    const args = JSON.parse(
+      await fs.readFile(path.join(directory, "args.json"), "utf8")
+    );
+
+    args.collection_name = collection_name
+
+    const index = new ChromaClient("http://localhost:8000");
```
nfcampos, line 149: I think we might want to save the url in args

langchain/vectorstores/chroma.ts
```diff
@@ -0,0 +1,194 @@
...
+try {
+  // eslint-disable-next-line global-require,import/no-extraneous-dependencies
+  ({ ChromaClient } = require("chromadb"));
+} catch {
+  // ignore error
+}
+ChromaClient = ChromaClientO;
+
+export interface ChromaLibArgs {
+  space: string;
```
nfcampos, line 24: Looks like you don't use this arg, so im good with removing?

package.json
```diff
@@ -1,13 +1,12 @@
 {
-  "version": "0.0.4",
+  "version": "0.0.5",
```
nfcampos, line 2: I think you can undo these changes, we'll bump the version when releasing

nfcampos (commented):
Overall looks good, just a few comments

langchain/vectorstores/chroma.ts
jeffchuber: ah! should not be snakecase. thank you

langchain/vectorstores/chroma.ts
jeffchuber: Feels like a little local validation that keeps it instep with the `hnsw` implementation is nice. We do more validation on the JS client side as well.

langchain/vectorstores/chroma.ts
```diff
@@ -0,0 +1,194 @@
...
+      });
+      docs.push(newDoc);
+    }
+    return Chroma.fromDocuments(docs, embeddings, collection_name);
+  }
+
+  static async fromDocuments(
+    docs: Document[],
+    embeddings: Embeddings,
+    collection_name: string
```
nfcampos, line 179: one more, do we want to give this a default value (eg. a uuid, if you support that as a collection name, maybe prepended with `langchain` or something)? Would make it even easier to get started (at the cost of ending up with less predictable collection names in your chromadb instance)

jeffchuber: great idea, ill add that now

jeffchuber:
@nfcampos I cleaned up the diffs in the `package.json` files. Can you help me generate a new `yarn.lock` file where the merge got funky? What is the expected behavior here? Should I merge `main` into this branch and then generate the `yarn.lock`?

nfcampos:
@jeffchuber done, should be good now. when you get conflicts in yarn.lock you can just run `yarn` at the root during the merge, with the conflicts still in it, and yarn will solve it

nfcampos:
@jeffchuber I've added a few more commits which should make the CI checks green, one last thing to think about, can you think of a reasonable way of adding a test for this? It could be just an integration test that requires some setup etc (I guess it needs to run docker?), but would be great if we can have something

jeffchuber:
thank you @nfcampos ! 

re: tests -- I would also definitely like to have tests. 

One simple idea would be to stub out our API responses? That way at least changes in `langchainjs` that break the integration will show up.

nfcampos:
That sounds perfect

jeffchuber:
Great, I'll add that on tonight.

jeffchuber:
@nfcampos -- thinking more about this - adding the harness and all the supporting tools to `langchainjs` to do all the mocking I would want to do feels presumptive (and premature) at this point. My proposal is to merge this as-is, and then target that in a future PR that chroma team can take on. 

cc @hwchase17

nfcampos:
@jeffchuber that sounds very reasonable to me, do you want to open an issue for that?

nfcampos:
And one other question, I understand chroma has a builtin way to store metadata for documents (ie. a replacement for the in memory docstore we're using here). Would we want to use that here? (this is something that is very interesting to us as we work to make langchain runnable in non-node environment where the in-memory metadata store is not ideal). Also happy for this to be a followup if that makes more sense

jeffchuber:
@nfcampos https://github.com/hwchase17/langchainjs/issues/96

You read my mind :) Here is the API - https://github.com/chroma-core/chroma/blob/14c26f5c222e3bb72e811a4403cbc644846629d6/clients/js/src/index.ts

Let me know what you're thinking and let's make it happen!

nfcampos:
@jeffchuber check out the latest commit, does it look right?

jeffchuber:
@nfcampos looking now!

PR #74, opened 2023-02-20
Add Redis Cache, Make Cache interface async, cache clean up

samheutmaker:
Here's a PR for a Redis cache. It includes a few small change to the BaseCache and InMemory cache as well. 

One thing to note: In Python LangChain, the cache is set on the global langchain object like:
```.py
import langchain
from langchain.cache import InMemoryCache
langchain.llm_cache = InMemoryCache()
```

This doesn't translate well to TypeScript. It's worth discussing how this should be handled. IMHO, it would be best to optionally pass the cache to the LLM class upon instantiation with an in-memory cache as a fallback.


langchain/cache.ts
```diff
@@ -1,27 +1,79 @@
+import type { RedisClientType } from 'redis';
 import { Generation } from "./llms";
 
-const getKey = (prompt: string, llmKey: string): string =>
-  `${prompt}_${llmKey}`;
+/**
+ * This cache key should be consistent across all versions of langchain.
+ * It is currently NOT consistent across versions of langchain.
```
hwchase17, line 6: what does this mean?

hwchase17 (commented):
high level, looks good to me

re: your comments, have you been able to test this yet? was wondering how the global ness would scale

langchain/cache.ts
samheutmaker: I did some really simple testing just to make sure the Redis cache connects and can get/set values haven't done a test at scale.

This comment assumes you have a system with both Python and TS applications that are operating over the same data and can benefit from a shared cache. This may be something like a Python system that runs every 24 hours and primes the cache with FAQs and a TS system that endusers interact with.

The global-ness should scale fine as long as Redis is properly managed.

langchain/cache.ts
```diff
@@ -1,27 +1,79 @@
...
+
+    return generations.length > 0 ? generations : null;
   }
 
-  update(prompt: string, llmKey: string, value: T) {
-    this.cache[getKey(prompt, llmKey)] = value;
+  public update(prompt: string, llmKey: string, value: Generation[]): void {
+    for (let i = 0; i < value.length; i += 1) {
+      const key = getCacheKey(prompt, llmKey, String(i));
+      this.#redisClient.set(key, value[i].text);
```
nfcampos, line 76: Should this be awaited?

nfcampos:
Thanks a lot for this @samheutmaker

PR #75, opened 2023-02-21
UNTESTED attempt to replace default docusaurus social card image

aavetis:
on a mobile device so i can't test it, and i'm not familiar with docusaurus so i don't know for sure, but this image matches the one that twitter is unfurling the links with. saw @hwchase17 added a placeholder langchain image in the directory, so just slapping that one on as a temporary solution? 🤷 

image


hwchase17 (approved):
genius, i love it

PR #88, opened 2023-02-22
IFTTT + LangChain

SidU:
Tool to allow LangChain apps to connect to the services in the IFTTT ecosystem.

Tool can be added multiple times into an agent with a different `name`, `description`, and webhook `url` for each instance to create a more powerful LangChain app that can play Spotify music, interact with smart-lights, etc.


hwchase17 (commented):
do you have an example of using this?

SidU:
> do you have an example of using this?

Yup, added a wiki here https://github.com/SidU/teams-langchain-js/wiki/Connecting-IFTTT-Services

PR #90, opened 2023-02-22
Proposal 1 of 2: Add concurrency control to BaseChain

nfcampos:
See also #92. We only want 1 of these.

Using https://github.com/sindresorhus/p-queue

This would allow all chains to be "rate-limited", especially useful when dealing with APIs that are heavily rate-limited, like OpenAI.


nfcampos:
Closing in favour of #92

PR #103, opened 2023-02-22
update doc to suggest fewer more way of setting up the Api Key 

MaximeThoonsen:
… getting started section


hwchase17 (approved):
awesome - thank you!

PR #105, opened 2023-02-23
Add GitBook Document loader

aletna:


hwchase17 (approved):
you're a hero

i may change the usage slightly to take the `true` param in at init rather than load, similar to python interface

PR #118, opened 2023-02-23
[Non-Node Envs] Make openai classes use fetch, remove use of node-fetch 

nfcampos:
Adds installation instructions for node 16


sullivan-sean:
Looks like https://github.com/vespaiach/axios-fetch-adapter doesn't support CJS.

It's small enough and doesn't really have any deps besides axios that we could just pull the one adapter function into our codebase.

Also looks like axios has an open PR to add fetch support natively: https://github.com/axios/axios/pull/5146 so we could eventually remove whatever we go with here.

nfcampos:
@sullivan-sean yea or we wait until going for a full ESM build to merge this one, which was my plan, hopefully I can fix the remaining issues in #124

nfcampos:
@sullivan-sean I did what you suggested, in #124 , that package has a misconfigured package.json

PR #133, opened 2023-02-24
add json tools and agent

agola11:


langchain/agents/tools/json.ts
```diff
@@ -0,0 +1,101 @@
...
+
+function parseInput(input: string): (string | number)[] {
+    const regex = /\[.*?]/g;
+    const _res = input.match(regex) ?? [];
+    const res = _res.map((i) => i.slice(1, -1).replace(/['"]+/g, ''));
+    return res.map((i) => (Number.isNaN(+i) ? i : +i));
+}
+
+export class JsonSpec {
+    obj: { [key: string]: any };
```
agola11, line 11: any ideas on a better type here? Linter complaining about using any

langchain/agents/tests/json.test.ts
```diff
@@ -0,0 +1,68 @@
+import { test, expect } from "@jest/globals";
+import { JsonListKeysTool, JsonSpec, JsonGetValueTool } from "../tools/json";
+
+test("JsonListKeysTool", async () => {
+  const jsonSpec = new JsonSpec({
+    foo: "bar",
+    baz: { test: { foo: [1, 2, 3] } },
```
nfcampos, line 7: What about lists of dictionaries, do we support that?

langchain/agents/tools/json.ts
```diff
@@ -0,0 +1,103 @@
...
+    this.max_value_length = max_value_length;
+  }
+
+  public getKeys(input: string): string {
+    const keys = parseInput(input);
+    let res = this.obj;
+    for (const key of keys) {
+      res = res[key];
+    }
+    if (res.constructor === Object) {
```
nfcampos, line 26: We usually do `typeof res === 'object' && !Array.isArray(res)`

langchain/agents/tools/json.ts
```diff
@@ -0,0 +1,103 @@
+import { Tool, Toolkit } from "./base";
+
+function parseInput(input: string): (string | number)[] {
```
nfcampos, line 3: How about we use the JSONPointer spec, as implemented eg here https://www.npmjs.com/package/jsonpointer might even improve LLM accuracy given it's a more standard thing

langchain/agents/tools/json.ts
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  public getValue(input: string): string {
+    const keys = parseInput(input);
+    let res = this.obj;
+    for (const key of keys) {
+      res = res[key];
+    }
+    const str =
+      res.constructor === Object ? JSON.stringify(res) : res.toString();
```
nfcampos, line 41: Same as above

nfcampos (commented):
A few comments

langchain/agents/tools/json.ts
nfcampos: courtesy of supabase 

```
export type Json =
  | string
  | number
  | boolean
  | null
  | { [key: string]: Json }
  | Json[]

export type JsonObject = { [key: string]: Json };

export class JsonSpec {
    obj: JsonObject
    ...
}
```

agola11: Awesome, thanks for this!

langchain/agents/tests/json.test.ts
agola11: Yeah, we should. Will add test case for this

langchain/agents/tools/json.ts
agola11: Yeah this is great. Will implement. There seems to be something similar in python too, which I can add

examples/src/agents/json.ts
```diff
@@ -0,0 +1,45 @@
...
+import * as yaml from "js-yaml";
+import { JsonSpec, JsonToolkit, JsonObject } from "langchain/tools";
+
+export const run = async () => {
+  let data: JsonObject;
+  try {
+    const yamlFile = fs.readFileSync("openai_openapi.yaml", "utf8");
+    data = yaml.load(yamlFile) as JsonObject;
+    if (!data) {
+      console.error("Could not load YAML file");
```
nfcampos, line 13: Should we throw an error here?

langchain/agents/tests/json.test.ts
```diff
@@ -0,0 +1,71 @@
...
+  expect(await jsonListKeysTool.call("/foo")).toContain("not a dictionary");
+  expect(await jsonListKeysTool.call("/baz")).toBe("test");
+  expect(await jsonListKeysTool.call("/baz/test")).toBe("foo, qux");
+  expect(await jsonListKeysTool.call("/baz/test/foo")).toContain(
+    "not a dictionary"
+  );
+  expect(await jsonListKeysTool.call("/baz/test/foo/0")).toContain(
+    "not a dictionary"
+  );
+  expect(await jsonListKeysTool.call("/baz/test/qux")).toContain(
```
nfcampos, line 20: Arguable here it could return `0, 1` but not sure how useful that is in this case

langchain/agents/tools/json.ts
```diff
@@ -0,0 +1,110 @@
...
+  | null
+  | { [key: string]: Json }
+  | Json[];
+
+export type JsonObject = { [key: string]: Json };
+
+export class JsonSpec {
+  obj: JsonObject;
+
+  max_value_length = 4000;
```
nfcampos, line 17: usually in JS we use camel case, so `maxValueLength`, unless we're trying to keep the API exactly the same as python

langchain/agents/tools/json.ts
```diff
@@ -0,0 +1,110 @@
...
+    try {
+      return this.jsonSpec.getKeys(input);
+    } catch (error) {
+      return `${error}`;
+    }
+  }
+
+  description = `Can be used to list all keys at a given path. 
+    Before calling this you should be SURE that the path to this exists.
+    The input is a text representation of the path to the json in as json pointer syntax (e.g. /key1/0/key2).`;
```
nfcampos, line 77: nit: `in as` we probably need only one of the two

langchain/agents/tools/requests.ts
```diff
@@ -0,0 +1,69 @@
...
+interface RequestTool {
+  headers: headers;
+}
+
+export class RequestsGetTool extends Tool implements RequestTool {
+  name = "requests_get";
+
+  headers: headers;
+
+  constructor(headers?: headers) {
```
nfcampos, line 14: you can use TS "parameter properties" with a default value, like this `constructor(public headers?: Headers = {})`

langchain/agents/tools/requests.ts
```diff
@@ -0,0 +1,69 @@
+import { Tool, Toolkit } from "./base";
+
+type headers = { [key: string]: string };
```
nfcampos, line 3: Two things here
- types are usually title case so `Headers`
- while this is perfectly valid, types for objects are usually written with the `interface` keyword, like

```
interface Headers {
  [key: string]: string;
}
```

.gitignore
```diff
@@ -19,5 +19,6 @@ langchain/docs/
 
 .DS_Store
 
+openai_openapi.yaml
```
hwchase17, line 22: may actually be nice to check this in for easy example, its not that big

langchain/agents/agent_toolkits/prompt.ts
```diff
@@ -0,0 +1,75 @@
+export const SQL_PREFIX = `You are an agent designed to interact with a SQL database.
```
hwchase17, line 1: probably nice to split these up? not super scalable as we start to add more

maybe structure can be

json
  - agent
  - toolkit
  - prompt
sql
   - agent
   - toolkit
   - prompt

etc

hwchase17 (approved):

PR #134, opened 2023-02-24
Add support for (dis)allowed special and tokens in token text splitter

### LINKED ISSUE #122: TokenTextSplitter doesn't support special tokens
It seems that this is a [known issue](https://github.com/hwchase17/langchainjs/blame/4ecf055e760a66eeec4e15f755921e38d8372f51/langchain/text_splitter.ts#L214) but wanted to call it out explicitly since we're (Diagram) interested in helping bring the WASM tokenizer up to parity with python tiktoken for our usage. Not currently using langchain, but would be interested in adopting bits and pieces.

dqbd:
Upgrade the `@dqbd/tiktoken` dependency to version 0.4.0, which includes feature parity with its Python counterpart.

Closes #122

### DIFF
--- a/examples/package.json
+++ b/examples/package.json
@@ -19,7 +19,7 @@
   "author": "Langchain",
   "license": "MIT",
   "dependencies": {
-    "@dqbd/tiktoken": "^0.2.1",
+    "@dqbd/tiktoken": "^0.4.0",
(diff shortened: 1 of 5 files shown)


hwchase17 (approved):
you're a hero :)

PR #137, opened 2023-02-25
add requests toolkit and example for making requests based on openapi spec

agola11:


examples/src/agents/requests.ts
```diff
@@ -0,0 +1,107 @@
...
+  } catch (e) {
+    console.error(e);
+    return;
+  }
+
+  const toolkit = new JsonToolkit(new JsonSpec(data));
+  const model = new OpenAI({ temperature: 0 });
+
+  const jsonAgent = ZeroShotAgent.asJsonAgent(model, toolkit);
+  const jsonExecutor = AgentExecutor.fromAgentAndTools({
```
nfcampos, line 30: I wonder if we should have a method on agentexecutor to turn it into a (dynamic)tool. Would make it slightly easier to do so, and remind people that it is possible to do

langchain/agents/tools/requests.ts
```diff
@@ -0,0 +1,69 @@
...
+      return res.text();
+    } catch (error) {
+      return `${error}`;
+    }
+  }
+
+  description = `Use this when you want to POST to a website.
+  Input should be a json string with two keys: "url" and "data".
+  The value of "url" should be a string, and the value of "data" should be a dictionary of 
+  key-value pairs you want to POST to the url.
```
nfcampos, line 57: "...you want to POST to the url as a JSON body"

langchain/agents/tools/requests.ts
```diff
@@ -0,0 +1,69 @@
...
+  Be careful to always use double quotes for strings in the json string
+  The output will be the text response of the POST request.`;
+}
+
+export class RequestsToolkit extends Toolkit {
+  tools: Tool[];
+
+  constructor(headers?: headers) {
+    super();
+    this.tools = [new RequestsGetTool(headers), new RequestsPostTool(headers)];
```
nfcampos, line 67: Why not make a more complete RequestsTool that includes the ability to consult the api spec (openapi)?

nfcampos (commented):
Great

examples/src/agents/requests.ts
agola11: yeah good point, can add in a followup PR

langchain/agents/tools/requests.ts
agola11: Yeah can try and consolidate if people find useful. The goal is to create an interesting example for now

PR #158, opened 2023-02-27
partial format of prompt

hwchase17:


langchain/src/prompts/base.ts
```diff
@@ -35,14 +38,35 @@ export abstract class BasePromptTemplate implements BasePromptTemplateInput {
...
+  mergePartialAndUserVariables(userVariables: InputValues): InputValues {
+    const partialVariables = this.partialVariables ?? {};
+    const partialKwargs: InputValues = {};
+    for (let i = 0; i < Object.keys(partialVariables).length; i += 1) {
+      const key = Object.keys(partialVariables)[i];
+      const value = partialVariables[key];
+      if (typeof value === "string") {
+        partialKwargs[key] = value;
+      } else {
+        partialKwargs[key] = value();
```
agola11, line 65: 1) `partialKwargs` is pretty python-centric terminology. Probs better to rename to `partialValues`
2) We probably want to think about the typing for `inputValues` here and should support `value` being either an async function or string. Right now, if value is async, `partialKwargs[key]` would be assigned a promise which isn't what we want.

agola11 (commented):
a couple comments, plus should add tests

docs/docs/modules/prompts/partial_prompts.md
```diff
@@ -0,0 +1,70 @@
+# Partial Prompt Templates
+
+A prompt template is a class with a `.format` method which takes in a key-value map and returns a string (a prompt) to pass to the language model. Like other methods, it can make sense to "partial" a prompt template - eg pass in a subset of the required values, as to create a new prompt template which expects only the remaining subset of values.
+
+LangChain supports this in two ways: we allow for partially prompts (1) with string values, (2) with functions that return string values. These two different ways support different use cases. In the documentation below we go over the motivations for both use cases as well as how to do it in LangChain.
```
nfcampos, line 5: Missing the word "formatted" after "partially"?

docs/docs/modules/prompts/partial_prompts.md
```diff
@@ -0,0 +1,70 @@
...
+    inputVariables: ["foo"],
+    partialVariables: { bar: "baz" },
+  });
+  expect(await prompt.format({ foo: "foo" })).toBe("foobaz");
+});
+```
+
+## Partial With Functions
+
+The other common use is to partial with a function. The use case for this is when you have a variable you know that you always want to fetch in a common way. A prime example of this is with date or time. Imagine you have a prompt which you always want to have the current date. You can't hard code it the prompt, and passing it along with the other input variables is a bit annoying. In this case, it's very handy to be able to partial the prompt with a function that always returns the current date.
```
nfcampos, line 41: Missing the word "in" in "You can't hardcode it in the prompt"

docs/docs/modules/prompts/partial_prompts.md
```diff
@@ -0,0 +1,70 @@
...
+```typescript
+import { expect, test } from "@jest/globals";
+import { PromptTemplate } from "../prompt.js";
+test("Test partial with function", async () => {
+  const prompt = new PromptTemplate({
+    template: "Tell me a {adjective} joke about the day {date}",
+    inputVariables: ["adjective", "date"],
+  });
+  const partialPrompt = await prompt.partial({
+    date: () => dateFunction,
```
nfcampos, line 52: Maybe should use a real date function here? Something like `new Date().toLocaleDateString()`

docs/docs/modules/prompts/partial_prompts.md
```diff
@@ -0,0 +1,70 @@
...
+```typescript
+import { expect, test } from "@jest/globals";
+import { PromptTemplate } from "../prompt.js";
+test("Test partial with function", async () => {
+  const prompt = new PromptTemplate({
+    template: "Tell me a {adjective} joke about the day {date}",
+    inputVariables: ["adjective"],
+    partialVariables: { date: () => dateFunction },
+  });
+  expect(await prompt.format({ foo: "foo" })).toBe("foobaz");
```
nfcampos, line 68: This expect needs to be updated with the correct value

langchain/src/prompts/base.ts
```diff
@@ -9,6 +9,7 @@ export type SerializedBasePromptTemplate = ReturnType<
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export type InputValues = Record<string, any>;
+export type PartialValues = Record<string, string | (() => Promise<string>)>;
```
nfcampos, line 12: I think we can also allow functions that return a string (not wrapped in a promise)

langchain/src/prompts/base.ts
```diff
@@ -35,14 +39,37 @@ export abstract class BasePromptTemplate implements BasePromptTemplateInput {
...
   constructor(input: BasePromptTemplateInput) {
-    const { inputVariables } = input;
+    const { inputVariables, partialVariables } = input;
     if (inputVariables.includes("stop")) {
       throw new Error(
         "Cannot have an input variable named 'stop', as it is used internally, please rename."
       );
     }
     Object.assign(this, input);
+    Object.assign(this, partialVariables);
```
nfcampos, line 52: This doesn't do what you intent I think. This will copy each key of `partialVariables` onto `this` but then on the next method it looks like you access this.partialVariables. A simple assignment will do here

langchain/src/prompts/tests/few_shot.test.ts
```diff
@@ -0,0 +1,67 @@
...
+    prefix: "{foo}{bar}",
+    examples: [],
+    suffix: "",
+    templateFormat: "f-string",
+    exampleSeparator: "\n",
+    examplePrompt,
+    inputVariables: ["foo", "bar"],
+  });
+
+  const partialPrompt = await prompt.partial({ foo: "foo" });
```
nfcampos, line 47: Here let's add an expect against the original prompt to make sure that one is not modified

nfcampos (commented):
Looks good

Issue #130, opened 2023-02-24
Cache class should have an async lookup/update interface


nfcampos:
This is to be able to easily implement remote caches, eg. redis

nfcampos:
Partly addressed by #74

PR #165, opened 2023-02-28
Make cache local to each LLM (keeping the option of a global cache for ease of use)

nfcampos:
And add Redis cache from #74


nfcampos:
Closes #130

PR #168, opened 2023-02-28
More text loaders: CSV, JSON, JSONLines

nfcampos:


langchain/src/document_loaders/json.ts
```diff
@@ -0,0 +1,16 @@
...
+export class JSONLoader extends TextLoader {
+  constructor(filePath: string, public pointer: string = "") {
+    super(filePath);
+  }
+
+  protected async parse(raw: string): Promise<string[]> {
+    const json = JSON.parse(raw.trim());
+    const pointer = jsonpointer.compile(this.pointer);
+    const value = pointer.get(json);
+    return Array.isArray(value) ? value : [value];
```
hwchase17, line 14: for my edification, this seems like it returns an array? why an array and not a string? or rather, im guessing this actually does return a string and im just missing something... what am missing?

nfcampos: This does indeed return an array. The idea being that this way we support also reading array fields from json files (which will create one Document per array entry). Check the test for this one, and the file that it uses

PR #169, opened 2023-02-28
PDF loader

nfcampos:


langchain/package.json
```diff
@@ -142,6 +143,7 @@
     "gpt-3-encoder": "^1.1.4",
     "jsonpointer": "^5.0.1",
     "p-queue": "^7.3.4",
+    "pdf-parse": "^1.1.1",
```
hwchase17, line 146: assuming this is not a crazy dependency with known issues?

langchain/src/document_loaders/tests/pdf.test.ts
```diff
@@ -0,0 +1,16 @@
...
+test("Test PDF loader", async () => {
+  const filePath = path.resolve(
+    path.dirname(url.fileURLToPath(import.meta.url)),
+    "./example_data/1706.03762.pdf"
+  );
+  const loader = new PDFLoader(filePath);
+  const docs = await loader.load();
+
+  expect(docs.length).toBe(1);
+  expect(docs[0].pageContent).toContain("Attention Is All You Need");
```
hwchase17, line 15: hah nice

langchain/src/document_loaders/tests/pdf.test.ts
```diff
@@ -0,0 +1,16 @@
+import { test, expect } from "@jest/globals";
+import url from "url";
+import path from "path";
+import { PDFLoader } from "../pdf.js";
+
+test("Test PDF loader", async () => {
+  const filePath = path.resolve(
```
hwchase17, line 7: nice helpful util

hwchase17 (approved):
lgtm

langchain/package.json
nfcampos: Seems good to me (although last updated 4 years ago, but I guess pdfs don't change)

nfcampos: And btw if we we ever want to produce one document per page in a pdf all we have to do is start using the underlying dependency (pdf.js) directly

fraserxu: pdf.js is huge. I tried to use it, the code works but it has a dependency on a package called `canva` which is huge and made it impossible to use in any serverless(Vercel, AWS lambda) environment.

langchain/src/document_loaders/pdf.ts
```diff
@@ -0,0 +1,34 @@
...
+    super();
+  }
+
+  public async load(): Promise<Document[]> {
+    const { readFile } = await PDFLoader.imports();
+    const buffer = await readFile(this.filePath);
+    const parsed = await pdf(buffer);
+    const metadata = { source: this.filePath };
+    return [new Document({ pageContent: parsed.text, metadata })];
+  }
```
fraserxu, line 19: would it be good to provide a way to load from a buffer directly? I think there might be use cases such as upload a pdf file from a form from frontend and it will not be from file system. https://github.com/fraserxu/book-gpt/blob/main/pages/api/ingest.ts#L43

nfcampos: Very much agree, and for every file-based loader, in a future PR touching all of them

langchain/package.json
fraserxu: If we make it a peerDependency, it means whoever is going to use this loader would suffer from the size issue still.

nfcampos: Is there an alternative pdf parser that doesn't use pdf.js? Ideally I want a pure-JS one, not one with native bindings, which would bring other problems

fraserxu: I don't believe this package `pdf-parse` uses `pdf.js`. In the readme, it does mention a few alternatives tho https://www.npmjs.com/package/pdf-parse

fraserxu: Sorry, it uses it... I didn't see it in `package.json`, it seems that it include it in the repo instead https://gitlab.com/autokent/pdf-parse/-/tree/master/lib/pdf.js

nfcampos: I think it does, but the author bundled in the dist version, so probably a lot smaller than a direct install, see https://gitlab.com/autokent/pdf-parse/-/tree/master/lib/pdf.js

fraserxu: The one I had issue is https://www.npmjs.com/package/pdfjs-dist

nfcampos: Ha yes exactly! In your book-gpt you didn't have an issue getting this pdf-parse one deployed to a serverless function, right?

langchain/src/document_loaders/pdf.ts
onurays: @nfcampos any updates / plans on this?

PR #170, opened 2023-02-28
Directory loader, and docs for the other new loaders

nfcampos:
This is a slightly different interface from Python, imo a lot more powerful, but happy to change if not welcome


nfcampos:
Improved layout for document loaders docs sidebar, now that we have more

Screenshot 2023-02-28 at 23 08 51

hwchase17 (approved):

PR #173, opened 2023-03-01
Minor contribution improvements to streamline on-boarding

HunterGerlach:


.github/workflows/integration.yml
```diff
@@ -4,6 +4,9 @@
 name: Node.js Integration Tests
 
 on:
+  push:
```
hwchase17, line 7: we dont quite have this set up with env vars yet

hwchase17 (commented):
largely lgtm modulo the integration test stuff

HunterGerlach:
Removed integration update+ref

package.json
```diff
@@ -21,9 +21,12 @@
     "lint": "turbo run lint",
     "lint:fix": "yarn lint -- --fix",
     "test": "turbo run test",
+    "test:unit": "turbo run test --filter langchain --",
```
nfcampos, line 24: This is not quite true, we do have commands to run unit tests and integration tests, but they are inside the `langchain` folder

nfcampos: Specifically it's the next line that isn't correct, it can instead be `yarn workspace langchain run test:integration`

HunterGerlach: Yes, I actually went down the path of breaking out linting, unit tests, and integration tests into separate steps, but ultimately didn't think it provided much additional value when I noticed that you had the main parts covered in the CI action.

HunterGerlach: Oh, sorry I re-read that and I think I misunderstood what you were saying. I'll try to clean that up tomorrow. My goal was to have everything run from the root dir by default.

CONTRIBUTING.md
```diff
@@ -76,25 +76,71 @@ Now, you should be able to run the common tasks in the following section.
...
+
 ### Testing
 
 Tests should be added within a `tests/` folder alongside the modules they
 are testing.
 
+To run all tests, run:
+
+```bash
+yarn test
```
nfcampos, line 113: This command would run only unit tests

PR #175, opened 2023-03-01
Add vectorstore agent

agola11:
VectorStoreQAWithSourcesChain not supported in ts yet, so only added VectorStoreQA


langchain/src/agents/tools/vectorstore.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  static getDescription(name: string, description: string): string {
+    return `Useful for when you need to answer questions about ${name}. "
+            "Whenever you need information about ${description} "
+            "you should ALWAYS use this. "
+            "Input should be a fully formed question.`;
+  }
+
+  async call(input: string) {
+    const chain = VectorDBQAChain.fromLLM(this.llm, this.vectorStore);
```
nfcampos, line 36: Is it on purpose that we create a new chain for each call?

langchain/src/agents/tools/vectorstore.ts
```diff
@@ -0,0 +1,39 @@
...
+  constructor(name: string, description: string, fields: VectorStoreTool) {
+    super();
+    this.name = name;
+    this.description = description;
+    this.vectorStore = fields.vectorStore;
+    this.llm = fields.llm;
+  }
+
+  static getDescription(name: string, description: string): string {
+    return `Useful for when you need to answer questions about ${name}. "
```
nfcampos, line 29: I think the `"` shouldn't be here? In a ` string double quotes will just show up as any other character (so will the whitespace at the beginning of each line).

nfcampos (commented):
This looks great!

langchain/src/agents/tools/vectorstore.ts
agola11: yeah good point, can just initialize this in constructor, was just blindly doing what we do in python

langchain/src/agents/tools/vectorstore.ts
agola11: yup good catch

PR #180, opened 2023-03-01
Upgrade openai version to 3.2.0

Marviel:
Adds support for the long awaited [ChatGPT API backend](https://openai.com/blog/introducing-chatgpt-and-whisper-apis) to be added(!!!)

Per Greg Brockman, it's 10x cheaper, and higher quality. Woo!

https://platform.openai.com/docs/api-reference/chat/create


nfcampos:
Thanks a lot @Marviel we've just added support for this in #181

PR #181, opened 2023-03-01
Nc/chatgpt

nfcampos:


langchain/src/embeddings/openai.ts
```diff
@@ -1,10 +1,6 @@
-import type {
```
nfcampos, line 1: This is getting rid of the peer dependency dance

langchain/src/llms/openai.ts
```diff
@@ -1,6 +1,6 @@
-import type {
```
nfcampos, line 1: This is getting rid of the peer dependency dance

langchain/src/llms/openai.ts
```diff
@@ -123,7 +123,11 @@ export class OpenAI extends BaseLLM implements OpenAIInput {
 
   streaming = false;
 
-  private client: OpenAIApiT;
+  // Used for non-streaming requests
+  private batchClient: OpenAIApi;
```
nfcampos, line 127: This is fixing a bug I noticed, oops

PR #188, opened 2023-03-01
Feat/dynamic max tokens

evad1n:
References https://github.com/hwchase17/langchain/blob/master/langchain/llms/openai.py#L321 to implement the missing logic described in the comment. Passing -1 will calculate the max tokens dynamically.

Only done for 

There may be a problem with strongly typing the model name as a `TiktokenModel` as it would not let other models to be passed in if the API supports new ones but the type isn't updated. Therefore I cast it.


nfcampos (commented):
One comment, looks good otherwise

nfcampos:
@evad1n merged this in now, thank you!

PR #194, opened 2023-03-02
Add ability to serialize / deserialize steps in AgentExecutor

carlosdp:
Small change that allows for asynchronous continuation of an `AgentExecutor` when some of the steps are longer running tasks. Example:

```typescript
// CustomTool = a tool with returnDirect = true

const result = executor.call({ input }); // CustomTool returns direct with "wait", so we need to wait for a long task to finish

const serializedSteps = result.serializeSteps();

//... later

const toolResult; // contains result from CustomTool long running task

executor.deserializeSteps(serializedSteps, toolResult); // replaces the last Observation with the actual result. at the moment, CustomTool needs to be smart enough to respond with returnDirect = false now

const result = executor.call({ input }); // continue as usual, repeat
```

I just pulled this out of the custom executor fork I made for a project of mine, so not fully thought through, but I think this pattern makes sense.

### DIFF
--- a/langchain/src/agents/executor.ts
+++ b/langchain/src/agents/executor.ts
@@ -29,7 +29,9 @@ export class AgentExecutor extends BaseChain {
 
   maxIterations?: number = 15;
 
-  earlyStoppingMethod: StoppingMethod = "force";
+  earlyStoppingMethod: StoppingMethod = 'force';
+
+  private steps: AgentStep[] = [];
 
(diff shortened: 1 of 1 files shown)


bracesproul:
@carlosdp going to close due to merge conflicts. If this is still applicable/you'd still like to get this merged in, please open off of an up-to-date `main` branch, and I'll review. Thank you!

PR #195, opened 2023-03-02
Allow prompt customization for initializeAgentExecutor

Marviel:
My agent has been issues with using tools that it shouldn't... so the ability to customize the prompt that is passed to the agent executor would be very helpful, since that seems to be the preferred wrapper for getting an agent executor.

- Added ability for `promptArgs` to be passed in
- Moved `initializeAgentExecutor` to object parameters for better self-documentation

### DIFF
--- a/langchain/src/agents/initialize.ts
+++ b/langchain/src/agents/initialize.ts
@@ -1,17 +1,25 @@
-import { Tool } from "./tools/index.js";
 import { BaseLLM } from "../llms/index.js";
 import { AgentExecutor } from "./executor.js";
-import { ZeroShotAgent } from "./mrkl/index.js";
+import {
+  CreatePromptArgs,
+  ZeroShotAgent,
+} from "./mrkl/index.js";
+import { Tool } from "./tools/index.js";
 
-export const initializeAgentExecutor = async (
-  tools: Tool[],
-  llm: BaseLLM,
-  agentType = "zero-shot-react-description"
-): Promise<AgentExecutor> => {
+export interface InitializeAgentExecutorOptions {
+  tools: Tool[];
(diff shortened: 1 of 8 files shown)


jacobrosenthal:
I havent looked at this pr, but I have been digging into the codebase a ton. The current state of passing prompts to agents is a bit confusing but it does seem like theres a ton of ways to get prompts into agents, maybe too many. 

* `initializeAgentExecutor` uses `fromLLMAndTools` but does not pass `CreatePromptArgs` so it gets default prompt
* You can call instead new up a `ZeroShotAgent.fromLLMAndTools(llm, tools, args)` with the optional `CreatePromptArgs`, if not supplied uses its default hardcoded `createPrompt`
* `AgentExecutor.fromAgentAndTools(agent,tools)` is also available, it gets passed a constructed `ZeroShotAgent(llm)`, which calls `Agent` super constructor using the passed in llm's prompt

langchain/src/agents/initialize.ts
```diff
@@ -1,17 +1,27 @@
...
 import { AgentExecutor } from "./executor.js";
-import { ZeroShotAgent } from "./mrkl/index.js";
+import { CreatePromptArgs, ZeroShotAgent } from "./mrkl/index.js";
+import { Tool } from "./tools/index.js";
+
+export interface InitializeAgentExecutorOptions {
+  tools: Tool[];
+  llm: BaseLLM;
+  agentType?: "zero-shot-react-description";
+  promptArgs?: CreatePromptArgs;
```
hwchase17, line 10: will probably have to be made a bit more general to account for different types of agents

nfcampos:
Closing this in favour of #374

PR #197, opened 2023-03-02
add `ChatPrompt`

agola11:


langchain/src/prompts/chat.ts
```diff
@@ -0,0 +1,140 @@
...
+    Object.assign(this, input);
+
+    if (this.validateTemplate) {
+      if (!(this.templateFormat in DEFAULT_FORMATTER_MAPPING)) {
+        const validFormats = Object.keys(DEFAULT_FORMATTER_MAPPING);
+        throw new Error(`Invalid template format. Got \`${this.templateFormat}\`;
+                         should be one of ${validFormats}`);
+      }
+      // create a set of all input variables from all prompt messages
+      const inputVariables = new Set<string>();
```
nfcampos, line 58: I'm still not sold on the value of specifying this in addition to the prompts, what's the use case that is unlocked by this extra chance for the user making an error?

langchain/src/prompts/chat.ts
```diff
@@ -0,0 +1,140 @@
...
+    }
+  }
+
+  _getPromptType(): "chat" {
+    return "chat";
+  }
+
+  async format(values: InputValues): Promise<string> {
+    const messages = await this.formatChat(values);
+    return JSON.stringify(messages);
```
nfcampos, line 94: Is this sort of a placeholder? I imagine we could want a nicer stringified format for this

langchain/src/prompts/chat.ts
```diff
@@ -0,0 +1,140 @@
...
+    }
+    return messages;
+  }
+
+  serialize(): SerializedPromptTemplate {
+    throw new Error("ChatPromptTemplate.serialize() not yet implemented");
+  }
+
+  async partial(_: PartialValues): Promise<BasePromptTemplate> {
+    throw new Error("ChatPromptTemplate.partial() not yet implemented");
```
nfcampos, line 123: Btw this isn't that hard to implement basically just calling partial on each prompt (and returning a new instance), worth doing?

langchain/src/prompts/tests/chat.test.ts
```diff
@@ -0,0 +1,157 @@
+import { expect, test } from "@jest/globals";
+import { ChatPromptTemplate } from "../chat.js";
+import { PromptTemplate } from "../prompt.js";
+
+test("Test format", async () => {
+  const systemPrompt = new PromptTemplate({
```
nfcampos, line 6: Can we make this test slightly more comprehensive by having 1 variable shared by the 2 prompts

nfcampos (commented):
LGTM, juts a few minor comments

langchain/src/prompts/chat.ts
agola11: yeah fair enough, will change

langchain/src/prompts/chat.ts
agola11: yeah just placeholder for now

langchain/src/prompts/chat.ts
agola11: yeah shouldn't be too bad, can give it a go after other prompt changes

langchain/src/prompts/tests/chat.test.ts
agola11: will do

langchain/src/prompts/chat.ts
agola11: This is an artifact of `BasePromptTemplate`, and we'll need to change it there. This is outside the scope of this PR but we should definitely address. I don't think input variables should be specified anywhere, they should be just be inferred from the template.

PR #199, opened 2023-03-03
Add missing await in Summarization Chain example (docs)

rodgetech:
I was scratching my head for a while not knowing why the Summarization Chain example was not working for me, turns out I was not awaiting for the promise returned from this line of code which was missing in docs:

```typescript
const docs = textSplitter.createDocuments([text]);
```


nfcampos:
Thank you!

PR #201, opened 2023-03-03
feat: pgvector vector store and a Supabase client

jacobrosenthal:
Closes https://github.com/hwchase17/langchainjs/issues/128

Basic implementation of supabase pgvector querying. ~No adding of vectors atm.~

Seems complicated to figure out what the schema and query is for each user.. not sure how to generalize.

Also a problem is passing through match_count in addition to k, as well as limiting tokens used.


langchain/src/vectorstores/supabase.ts
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  export type SearchEmbeddingsParams = {
+    query_embedding: number[]
+    match_threshold: number // float
+    match_count: number // int
+  }
+  
+  
+export class PGVectorStore extends VectorStore {
```
nfcampos, line 23: I guess the name should refer to Supabase, given it requires a supabase client

langchain/src/vectorstores/supabase.ts
```diff
@@ -0,0 +1,107 @@
...
+      });
+  }
+
+  async similaritySearchVectorWithScore(
+    query: number[],
+    k: number
+  ): Promise<[Document, number][]> {
+    const matchDocumentsParams: SearchEmbeddingsParams = {
+      query_embedding: query,
+      match_threshold: k,
```
nfcampos, line 62: Is this right? K is meant to be the number of similar documents we want returned

langchain/src/vectorstores/supabase.ts
```diff
@@ -0,0 +1,107 @@
...
+  static async fromDocuments(
+    supabaseClient: SupabaseClient,
+    _docs: Document[],
+    embeddings: Embeddings,
+    textKey = "text"
+  ): Promise<PGVectorStore> {
+    return new PGVectorStore(supabaseClient, embeddings, textKey);
+  }
+
+  static async fromExistingIndex(
```
nfcampos, line 99: I don't think we need this method for this one, if it doesn't add anything on top of the regular constructor

nfcampos (commented):
Thanks for this!

(For reference for other folks reviewing https://supabase.com/blog/openai-embeddings-postgres-vector)

The biggest question for me is how to make easy all the setup that is described in the blogpost. If a user doesn't go through that first, and just tries to use this, it won't work. One option would be to add some `DDL` SQL statements here (copied from the blogpost) that we can optionally run using the supabase client for initial setup. Thoughts?

langchain/src/vectorstores/supabase.ts
jacobrosenthal: Yeah. theres also some other serverless pg providers, my original thought was it could be generalized

langchain/src/vectorstores/supabase.ts
jacobrosenthal: oh, is there a similarity score cut off?

langchain/src/vectorstores/supabase.ts
jacobrosenthal: The pinecone example does the same constructor pass through so this makes pgvector examples look the same. No strong preference

jacobrosenthal:
I updated the code to match the blog post. I think thats the best way forward, assume people used the blog post exactly as is. we can extend it to handle schema changes outside the post if people come forward with them

Currently doesnt set id, or handle metadata though, because the blog post didnt.

Also I dont think we can execute raw sql
https://github.com/supabase/supabase/discussions/3419#discussioncomment-1428126

So Im not sure I can just have a `prepare()` fn that creates the table and query sadly

examples/src/agents/vectorstore_supabase.ts
```diff
@@ -0,0 +1,48 @@
...
+  VectorStoreToolkit,
+  createVectorStoreAgent,
+  VectorStoreInfo,
+} from "langchain/agents";
+import { PGVectorStore } from "langchain/vectorstores";
+import { createClient } from '@supabase/supabase-js';
+
+
+export const run = async () => {
+  const model = new OpenAI({ temperature: 0 });
```
Johnrobmiller, line 13: **Question:** Might some users want a temp that is not 0? If so, should we allow them to pass in a custom temp as a part of the interface?

Johnrobmiller:
Not much to say on my end, looks good to me.

Also, first time looking at langchain, this lib looks fire 🔥

langchain/src/vectorstores/supabase.ts
```diff
@@ -0,0 +1,184 @@
+/// Youre not allowed to run raw sql on supabase apis I dont beleive
```
Johnrobmiller, line 1: **Comment:** Briefly looked into this. Indeed, Supabase doesn't like SQL getting executed outside of the safety of their own environment.

This is very hacky, but maybe we can bypass supabase altogether and use the postgres CLI in the terminal?

nfcampos:
(for me) to do here

- [ ] Move the sql setup code to the docs
- [ ] Add an integration test
- [ ] Rename to SupabaseVectorStore
- [ ] Handle error

sbusso:
> I updated the code to match the blog post. I think thats the best way forward, assume people used the blog post exactly as is. we can extend it to handle schema changes outside the post if people come forward with them
> 
> Currently doesnt use metadata though, because the blog post didnt.... Also not sure how to handle errors, throw?
> 
> Also I dont think we can execute raw sql [supabase/supabase#3419 (comment)](https://github.com/supabase/supabase/discussions/3419#discussioncomment-1428126)
> 
> So Im not sure I can just have a `prepare()` fn that creates the table and query sadly

Supabase client doesn't allow SQL but it is possible to connect directly to the database with a database client. Maybe the idea of a generic PG client would make sense.

jacobrosenthal:
@sbusso  yeah I found their urls and have something vaguely working with the postgres package. but then it feels like we need an abstraction so people can use whatever postgres library they want.. not sure how far to chase this.

jacobrosenthal:
Could go this route
```
export interface PGClient {
  similaritySearchVectorWithScore(
    query: number[],
    k: number
  ): Promise<[Document, number][]>;
  addVectors(
    vectors: number[][],
    documents: Document[],
    _ids?: string[]
  ): Promise<void>;
  prepare: () => void;
}
```
You dont even bring in any postgres libs into upstream this way just an example lib implements one, and they copy and paste it around with the example to build their schema and queries

nfcampos:
Thanks a lot for this @jacobrosenthal

nfcampos:
Per @jacobrosenthal's request on discord, re-posting here my message from Discord. We've merged this without the additional abstraction added in the last commits `PGVectorClient` because we already have an abstraction which is virtually identical, ie. the `VectorStore` base class, and the new abstraction didn't add anything additional specific to Postgres that warranted the additional complexity. Hope that helps, any questions let me know.

PR #204, opened 2023-03-03
add load from chat model

agola11:


langchain/package.json
```diff
@@ -218,4 +218,4 @@
       "import": "./document_loaders.js"
     }
   }
-}
```
nfcampos, line 221: this is so annoying haha, yarn formats the `package.json` files with newline, something else formats them without

langchain/src/chains/base.ts
```diff
@@ -9,6 +9,7 @@ import type {
   ChatQAChain,
 } from "./index.js";
 import { BaseMemory } from "../memory/index.js";
+import { ChatModelChain } from "./index.js";
```
nfcampos, line 12: This should be moved to the `import type` statement above (to avoid circular deps, which are allowed for type definitions but less allowed for actual code

langchain/src/chains/chat_vector_db_chain.ts
```diff
@@ -4,7 +4,7 @@ import {
   SerializedBaseChain,
   SerializedLLMChain,
   loadQAChain,
-  LLMChain,
+  LLMChain, loadQAChainFromChatModel,
```
nfcampos, line 7: Run `yarn format`

langchain/src/chains/question_answering/stuff_prompts.ts
```diff
@@ -1,8 +1,24 @@
...
+
+const chatTemplateSystem = `Use the following pieces of context to answer any user questions. If you don't know the answer, just say that you don't know, don't try to make up an answer
+{context}`;
+const chatPromptSystem = PromptTemplate.fromTemplate(chatTemplateSystem);
+const chatPromptUser = PromptTemplate.fromTemplate("{question}");
+
+export const DEFAULT_CHAT_QA_PROMPT = ChatPromptTemplate.fromPromptMessages([
+  {
+    role: "system",
+    message: chatPromptSystem,
```
nfcampos, line 18: How about defining them inline here , ie. 
```
{
     role: "system",
     message: PromptTemplate.fromTemplate(chatTemplateSystem)
}
```

Obviously not important for the code here, I'm thinking more about the example we want to set for people creating their own prompts

nfcampos (commented):
Some comments, this is where we'd want to consolidate with what I added in #190

PR #217, opened 2023-03-05
feat: Add Audio Loader

mosnicholas:
Add audio loader & OpenAI Whisper wrapper.

next up:
- Add chain to use voice as input
- Add agent that uses voice as input


hwchase17 (commented):
this is great. we'll probably treat as separate from an LLM. theres also an audio pr open in the python package - will merge them in at the same time

mosnicholas:
@hwchase17 was thinking the same -- are you suggesting creating a new package for it? Or just a new `BaseLLM` (ie. `BaseALM`) for audio work? Happy to pair on it :)

mosnicholas:
@hwchase17 was thinking the same -- are you suggesting creating a new package for it? Or just a new `BaseLLM` (ie. `BaseALM`) for audio work? Happy to pair on it :)

PR #223, opened 2023-03-06
agents: remove unsed allowedTools

jacobrosenthal:
Maybe I'm crazy, I dont think allowedTools is used, and it duplicates the tool.name anyway. Probably was refactored out at some point? 
Or If Im off base, feel free to close this


nfcampos:
Thanks @jacobrosenthal this is actually meant to be used, we have an open issue to add validation using it. Will be closing this one

PR #226, opened 2023-03-06
fix: Generated follow-up questions not being used

noughts:
Follow-up questions are generated when `chat_history` is specified in `ChatVectorDBQAChain` , but it appears that it is not being used.

I have fixed that issue.


hwchase17 (commented):
this is actually intentional - the new question should only be used in fetching documents. since we are passing in the chatHistory to the vectorDBQAchain, we can just pass in the original question

noughts:
@hwchase17 
I understand that is intentional. Thanks for the reply!

PR #231, opened 2023-03-06
memory refactor

hwchase17:


langchain/src/memory/base.ts
```diff
@@ -13,3 +20,37 @@ export abstract class BaseMemory {
...
+): string {
+  const string_messages: string[] = [];
+  for (const m of messages) {
+    let role: string;
+    if (m instanceof HumanChatMessage) {
+      role = human_prefix;
+    } else if (m instanceof AIChatMessage) {
+      role = ai_prefix;
+    } else if (m instanceof SystemChatMessage) {
+      role = "System";
```
nfcampos, line 47: Why isn't this one configurable?

langchain/src/memory/chat_memory.ts
```diff
@@ -0,0 +1,45 @@
...
+  chatHistory: ChatMessageHistory;
+
+  constructor(chatHistory?: ChatMessageHistory) {
+    super();
+    this.chatHistory = chatHistory ?? new ChatMessageHistory();
+  }
+
+  async saveContext(
+    inputValues: InputValues,
+    OutputValues: Promise<OutputValues>
```
nfcampos, line 39: This is a fairly strange API, receiving a promise of the argument. Should we take this opportunity to fix it?

nfcampos (approved):
LGTM

langchain/src/memory/chat_memory.ts
agola11: yeah this is weird for sure, but changing might be involved. probs best suited for another PR. Could create issue?

nfcampos: yea fair enough

PR #234, opened 2023-03-07
fix: await createDocuments() promise

iamjaredwalters:
## Description
Small update to Chat Vector DB & QA chain docs. `textSplitter.createDocuments([text])` returns a promise and should be awaited.


hwchase17 (approved):
thank you!

PR #236, opened 2023-03-07
feat: add logic to override default prompts and method to get default…

CarlosZiegler:
# Adding Options to the `fromLLM` Method in `ChatVectorDBQAChain`
The `ChatVectorDBQAChain` class provides a powerful mechanism for doing question answering over a large corpus of text documents. Given a question and a chat history, it first combines them into a standalone question, then looks up relevant documents from a vector database, and finally passes those documents and the question to a question answering chain to return a response. This process can be customized in a variety of ways to suit the specific needs of the user.

In a recent update to the library, the fromLLM method in the `ChatVectorDBQAChain` class has been enhanced to allow for additional options to be passed in during instantiation. These options provide a way for the user to customize the behavior of the `ChatVectorDBQAChain` object in various ways.

## Available Options
The following options are now available when creating a new ChatVectorDBQAChain instance using the fromLLM method:

`questionGeneratorTemplate`: A string that specifies a question generation template. If provided, the `ChatVectorDBQAChain` will use this template to generate a question from the conversation context, instead of using the question provided in the question parameter. This can be useful if the original question does not contain enough information to retrieve a suitable answer.
`qaTemplate`: A string that specifies a response template. If provided, the `ChatVectorDBQAChain` will use this template to format a response before returning the result. This can be useful if you want to customize the way the response is presented to the end user.
`returnSourceDocuments`: A boolean value that indicates whether the `ChatVectorDBQAChain` should return the source documents that were used to retrieve the answer. If set to true, the documents will be included in the result returned by the `call()` method. This can be useful if you want to allow the user to see the sources used to generate the answer. If not set, the default value will be false.

# Benefits of the Options
The addition of these options to the `fromLLM` method in the `ChatVectorDBQAChain` class provides a way for users to further customize the question answering process to their specific needs. By allowing for the customization of the question generation and response formatting steps, users can ensure that the `ChatVectorDBQAChain` object returns answers that are tailored to their requirements. Additionally, the `returnSourceDocuments` option provides users with greater transparency into the question answering process, allowing them to see the sources used to generate the answer and to perform additional analysis if necessary. Overall, the addition of these options enhances the flexibility and utility of the `ChatVectorDBQAChain` class, making it an even more powerful tool for doing question answering over large collections of text documents.

### DIFF
--- a/langchain/src/chains/chat_vector_db_chain.ts
+++ b/langchain/src/chains/chat_vector_db_chain.ts
@@ -21,9 +21,6 @@ Chat History:
 {chat_history}
 Follow Up Input: {question}
 Standalone question:`;
-const question_generator_prompt = PromptTemplate.fromTemplate(
-  question_generator_template
-);
 
 const qa_template = `Use the following pieces of context to answer the question at the end. If you don't know the answer, just say that you don't know, don't try to make up an answer.
 
@@ -31,7 +28,6 @@ const qa_template = `Use the following pieces of context to answer the question 
 
 Question: {question}
 Helpful Answer:`;
-const qa_prompt = PromptTemplate.fromTemplate(qa_template);
 
 export interface ChatVectorDBQAChainInput {
   vectorstore: VectorStore;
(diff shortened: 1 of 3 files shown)


hwchase17 (approved):
looks great to me! may need some refactoring after PromptCollections, but will tackle that then

langchain/src/chains/chat_vector_db_chain.ts
```diff
@@ -179,7 +175,30 @@ export class ChatVectorDBQAChain
     };
   }
 
-  static fromLLM(llm: BaseLLM, vectorstore: VectorStore): ChatVectorDBQAChain {
+  static getDefaultTemplates() {
```
nfcampos, line 178: FYI @CarlosZiegler removing this part to remain consistent with the other chains

alberduris:
Is there an example of using this anywhere?

PR #237, opened 2023-03-07
fix: supabase vectorstore addVectors chunking

jacobrosenthal:
upsert returns 500/502/504 (yes really any of them) if given too many rows/characters
~2000 trips it, but my data is probably smaller than average pageContent and metadata

Also was eating the error, now throws it. Finally use insert not upsert as the primary key isnt being inserted, not possible to update

Linking #201

### DIFF
--- a/langchain/src/vectorstores/supabase.ts
+++ b/langchain/src/vectorstores/supabase.ts
@@ -44,13 +44,25 @@ export class SupabaseVectorStore extends VectorStore {
   }
 
   async addVectors(vectors: number[][], documents: Document[]): Promise<void> {
-    await this.client.from(this.tableName).upsert(
-      vectors.map((embedding, idx) => ({
-        content: documents[idx].pageContent,
-        embedding,
-        metadata: documents[idx].metadata,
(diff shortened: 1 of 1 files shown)


nfcampos:
Great, thank you @jacobrosenthal

dk-crazydiv:
For others facing error "error inserting invalid input syntax for type json" while uploading to supabase with langchain, the work of this PR seems to have been overwritten by some other commits and insert code has been replaced by upsert again
https://github.com/langchain-ai/langchainjs/blame/main/langchain/src/vectorstores/supabase.ts#L123

Even tried manual external chunking and called `vectorStore.fromDocuments` and it doesn't work. Supabase seems to have some undocumented limit here. For my purposes, I'll probably move on to using some other vectorstore. Does anyone have any ideas on why this might be happening? 

Also this comment is still in the langchain code. 
```
upsert returns 500/502/504 (yes really any of them) if given too many rows/characters
~2000 trips it, but my data is probably smaller than average pageContent and metadata
```

PR #238, opened 2023-03-07
feat: try to delete timeout of browser to not exceed jest timeouts

CarlosZiegler:
According from our actions, we exceeded some timeout in puppeter :
https://github.com/hwchase17/langchainjs/actions/runs/4358676803/jobs/7619554448

This PR try to solve it, if the problem is the timeout conflict. If not resolves, need to refactor test.


nfcampos:
Closing in favour of #239 as discussed

PR #245, opened 2023-03-08
add chat agent

hwchase17:


examples/src/agents/chat_mrkl.ts
```diff
@@ -0,0 +1,31 @@
...
+import { SerpAPI, Calculator } from "langchain/tools";
+
+export const run = async () => {
+  const model = new ChatOpenAI({ temperature: 0 });
+  const tools = [new SerpAPI(), new Calculator()];
+
+  const executor = await initializeAgentExecutor(
+    tools,
+    model,
+    "chat-zero-shot-react-description"
```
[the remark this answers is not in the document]
hwchase17: no i dont think so - at least not currently (more is different than just the prompt)

langchain/src/agents/chat/index.ts
```diff
@@ -0,0 +1,124 @@
...
+  static createPrompt(tools: Tool[], args?: CreatePromptArgs) {
+    const { prefix = PREFIX, suffix = SUFFIX } = args ?? {};
+    const toolStrings = tools
+      .map((tool) => `${tool.name}: ${tool.description}`)
+      .join("\n");
+    const template = [prefix, toolStrings, FORMAT_INSTRUCTIONS, suffix].join(
+      "\n\n"
+    );
+    const messages = [
+      HumanMessagePromptTemplate.fromTemplate(template),
```
arnaudbzn, line 92: Is `SystemMessagePromptTemplate` more consistent for agent instructions?

langchain/src/agents/chat/index.ts
```diff
@@ -0,0 +1,124 @@
...
+      HumanMessagePromptTemplate.fromTemplate("{input}\n\n{agent_scratchpad}"),
```
arnaudbzn, line 93: About the `agent_scratchpad`:  A set of `AIMessage` can be used instead if it makes sense. These AI messages will be added after each step.

langchain/src/agents/chat/index.ts
hwchase17: yes! good call

langchain/src/agents/chat/index.ts
```diff
@@ -0,0 +1,125 @@
...
+  extractToolAndInput(text: string): { tool: string; input: string } | null {
+    if (text.includes(FINAL_ANSWER_ACTION)) {
+      const parts = text.split(FINAL_ANSWER_ACTION);
+      const input = parts[parts.length - 1].trim();
+      return { tool: "Final Answer", input };
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const [_, action, __] = text.split("```");
+    const response = JSON.parse(action.trim());
```
nfcampos, line 122: Should we do a try/catch here?

nfcampos (commented):
One comment, Lgtm otherwise

langchain/src/agents/chat/index.ts
agola11: I like this idea rather than stuffing everything in the user message. We'll experiment

PR #249, opened 2023-03-08
fix: Use new question when calling QA chain

nomadoda:
Hi!

I had some struggles in achieving a chat-like experience using the `ChatVectorDBQAChain`. After some digging I found that the generated question based on chat history is not used in the following `QAChain` call. Is this intentional?

In the python repo it seems like it's using new question

https://github.com/hwchase17/langchain/blob/master/langchain/chains/chat_vector_db/base.py#L96


hwchase17 (approved):
oops, thanks!

PR #250, opened 2023-03-08
Faster concurrent embeddings for OpenAI

vladgolubev:
I noticed that it takes too long to embed many documents, and realized that it's performed sequentially. 

On my machine, 20 documents take 10s to embed.
While with `concurrency: 20` it takes 1s.


vladgolubev:
@nfcampos just tried with 100 texts and batch size 20, all good.

2023-03-09T11 44 59-otYfGeq2

https://platform.openai.com/docs/guides/rate-limits/overview

3500 requests per minute gives ~60 req/s. So concurrency should not be a problem unless you set concurrency above `60`.

And even if it fails `embeddingWithRetry` function has retries, so likely it will succeed eventually.

nfcampos:
@vladgolubev we've increased the default value of batchSize, do you want to give it a go on the latest version? OpenAI recommend that batches are sent in the same API request (in Langchain this is done with the batchSize parameter) rather than as multiple concurrent requests, so I'd prefer to stick with OpenAI's recommendation if we can

PR #252, opened 2023-03-08
Nc/docs 8mar

nfcampos:


docs/docs/new.md
```diff
@@ -0,0 +1,78 @@
+# Composable, declarative prompts
```
agola11, line 1: Do we want to include this code in an actual example?

nfcampos: Oops committed this by mistake, will remove

PR #260, opened 2023-03-09
Consolidate sql db wrapper and toolkit

agola11:


examples/src/agents/sql.ts
```diff
@@ -28,6 +34,4 @@ export const run = async () => {
       2
     )}`
   );
-
-  db.close();
```
nfcampos, line 32: Is there no equivalent to this in the typeorm data source thing or something?

nfcampos (commented):
One comment

examples/src/agents/sql.ts
agola11: ah you're right, should be using ..destroy here. Will update

PR #261, opened 2023-03-09
fix: Remove unneeded `await` in constructor

zurawiki:
Javascript constructor are synchronous. The `await` can be dropped.

Also update imports for consistency


hwchase17 (approved):
looks pretty good to me!! thanks :)

PR #273, opened 2023-03-09
supabase: refactor to expose constructor options

### LINKED ISSUE #282: Make the remaining `fromX` methods in vector stores consistent across all classes

jacobrosenthal:
Closes #282

I hate to churn this, but in the refactor that got merged, the constructor options weren't exposed from `fromDocuments` or `fromTexts` or `fromExisting`
And a recent pr added an options struct for `fromDocuments` and moved the client into there

So why not always take client in the options at the same time?


jacobrosenthal:
Added https://github.com/hwchase17/langchainjs/issues/282 to try to speed this along
I havent tested chroma or pinecone.
Still needs docs if this is agreeable

nfcampos:
This is mostly good, and definitely leaves it in a better place. My only hesitation is the breaking change, any way we can ease that?

nfcampos:
Thanks a lot for this @jacobrosenthal I'm getting this in now

PR #277, opened 2023-03-10
Implement BaseLanguageModelParams interface that subclasses should extend

nfcampos:
cc @agola11 this is laying the groundwork for moving common functionality into BaseLanguageModel (such as callback manager) I moved verbose flag in this PR as an example

### DIFF
--- a/langchain/src/chat_models/base.ts
+++ b/langchain/src/chat_models/base.ts
@@ -3,6 +3,7 @@ import {
   AIChatMessage,
   BaseChatMessage,
   BaseLanguageModel,
+  BaseLanguageModelParams,
   BasePromptValue,
   ChatGeneration,
   ChatResult,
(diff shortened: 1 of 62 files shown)


langchain/src/chat_models/base.ts
```diff
@@ -22,26 +23,22 @@ const getCallbackManager = (): LLMCallbackManager => ({
...
-const getVerbosity = () => true;
-
 export type SerializedChatModel = {
   _model: string;
   _type: string;
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
 } & Record<string, any>;
 
+export interface BaseChatModelParams extends BaseLanguageModelParams {
+  callbackManager?: LLMCallbackManager;
```
agola11, line 33: `callbackManager` is present in both `BaseChatModelParams` and `BaseLLMParams`, let's move to `BaseChatModelParams`

nfcampos: For sure, and a bunch of other things too. I didn't move it on purpose, as it sounded easier to do at the same time as you actually implement it.

PR #279, opened 2023-03-10
Create reusable AsyncCaller class for calling "expensive" resources like APIs or LLMs

nfcampos:
This class contains generic logic useful for calling any "expensive" resource, eg. an API or LLM. This logic up to now has lived in various forms in subclassess of BaseLLM. As a follow on this will be applied across all those places.

As a follow on to this and to #277 will open a future PR using this in BaseLanguageModel

### DIFF
--- a/examples/tsconfig.json
+++ b/examples/tsconfig.json
@@ -6,7 +6,7 @@
       "ESNext",
       "DOM"
     ],
-    "target": "ES2020",
+    "target": "ES2021",
     "module": "nodenext",
     "sourceMap": true,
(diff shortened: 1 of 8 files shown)


agola11 (approved):
looks reasonable to me. Is the queue actually needed? We don't have a similar concept for that in python

nfcampos:
@agola11 it makes it super easy to work around api rate limits (ie. queue up requests instead of having them fail), it's disabled by default

agola11:
kk makes sense, thanks!

PR #280, opened 2023-03-10
Update hnswlib, new version adds cosine metric

nfcampos:


langchain/src/vectorstores/hnswlib.ts
```diff
@@ -189,7 +189,7 @@ export class HNSWLib extends SaveableVectorStore {
     docstore: InMemoryDocstore = new InMemoryDocstore()
   ): Promise<HNSWLib> {
     const args: HNSWLibArgs = {
-      space: "ip", // dot product
+      space: "cosine",
```
nfcampos, line 192: This will throw an error for people who don't update the hnswlib dependency, but it is a better default going forward

PR #283, opened 2023-03-10
Allow options to be passed to fromLLM constructor in ChatVectorDBQAChain

bsantos09:
Allow options to be passed to fromLLM constructor. 
Allow the options: inputKey, outputKey, k, returnSourceDocuments to be passed when creating a chain fromLLM.


hwchase17 (approved):
looks pretty good to me! thanks!

PR #284, opened 2023-03-10
Add `CallbackHandler`, `CallbackManager` and `Tracer` with tests

agola11:
Add callback handlers, manager and tracer implementation with tests.

Has not been wired up to base classes yet! Will do that in a followup PR.


langchain/src/callbacks/index.ts
```diff
@@ -0,0 +1,259 @@
...
+    tool: { name: string },
+    input: string,
+    verbose?: boolean
+  ): Promise<void>;
+
+  handleToolError?(err: Error, verbose?: boolean): Promise<void>;
+
+  handleToolEnd?(output: string, verbose?: boolean): Promise<void>;
+
+  handleText?(text: string, verbose?: boolean): Promise<void>;
```
nfcampos, line 65: Remind me when this one is fired?

langchain/src/callbacks/index.ts
```diff
@@ -0,0 +1,259 @@
...
+  }
+
+  async handleLLMStart(
+    llm: { name: string },
+    prompts: string[],
+    verbose?: boolean
+  ): Promise<void> {
+    for (const handler of this.handlers) {
+      if (!handler.ignoreLLM && (verbose || handler.alwaysVerbose)) {
+        await handler.handleLLMStart?.(llm, prompts);
```
nfcampos, line 99: How about adding a try catch around each call to a handler. That way of one handler errors your chain run (or whatever it was) doesn't fail, and a failure in one handler doesn't prevent calling the next handlers in the list

langchain/src/callbacks/index.ts
```diff
@@ -0,0 +1,259 @@
...
+
+  addHandler(handler: BaseCallbackHandler): void {
+    this.handlers.push(handler);
+  }
+
+  removeHandler(handler: BaseCallbackHandler): void {
+    this.handlers = this.handlers.filter((_handler) => _handler !== handler);
+  }
+
+  setHandlers(handlers: BaseCallbackHandler[]): void {
```
nfcampos, line 219: Maybe this could be named replaceHandlers to be clearer?

langchain/src/callbacks/index.ts
```diff
@@ -0,0 +1,259 @@
...
+    this.handlers = handlers;
+  }
+}
+
+export class ConsoleCallbackHandler extends BaseCallbackHandler {
+  constructor() {
```
nfcampos, line 225: This isn't needed (when it's empty)

nfcampos (commented):
Some comments, looks good

langchain/src/callbacks/index.ts
agola11: This is used whenever we need to log arbitrary messages from chains or agents that are not necessarily on start/end events. TBH I'm not in love with this, and want to look into another soln but it's not a priority rn, so I just copied what's in python

langchain/src/callbacks/index.ts
agola11: Great idea

agola11: resolved in followup PR

PR #285, opened 2023-03-10
add mysql support for sql_db chain

MaximeThoonsen:


nfcampos:
@MaximeThoonsen were you able to test this?

MaximeThoonsen:
hey @nfcampos . Yes I did. 
I was wondering if we want to create a directory with tests involving all databases types that we can run manually (with docker). It would be better than only sqlite ones. And allow you in this can of situation to tests it quickly. 
I can also setup mysql tests in the CI if you think it is mandatory. There is a tradeoff with the speed of the CI. 
What do you think?

nfcampos:
I think I'll merge this one without that, as it sounds like a fairly sizeable piece of work. I'm not the biggest fan of tests that need lots of other things to be running, but am not against it

PR #296, opened 2023-03-12
plug in callbacks

agola11:


langchain/src/llms/openai-chat.ts
```diff
@@ -247,10 +247,10 @@ export class OpenAIChat extends LLM implements OpenAIInput {
...
                 innerCompletion += part.delta?.content ?? "";
 
-                this.callbackManager.handleNewToken?.(
-                  part.delta?.content ?? "",
-                  this.verbose
-                );
+                // await this.callbackManager.handleLLMNewToken(
+                //     part.delta?.content ?? "",
+                //     this.verbose
+                // );
```
agola11, line 253: @nfcampos -- any ideas how to to get streaming to work with async callbacks? `createParser` expects sync function. Can we use something else?

langchain/src/base_language/index.ts
```diff
@@ -0,0 +1,40 @@
+import {BasePromptValue, LLMResult} from "../schema/index.js";
```
agola11, line 1: I had to move this out to avoid circular dependency: `BaseLanguageModel` in `langchain/schema` used `CallbackManager`, `CallbackManager` used `LLMResult` and other things in `langchain/schema`

langchain/src/chains/base.ts
```diff
@@ -29,16 +29,30 @@ export type SerializedBaseChain = ReturnType<
...
 
-  constructor(memory?: BaseMemory) {
+  verbose: boolean;
+
+  callbackManager: CallbackManager;
+
+  constructor(
+    memory?: BaseMemory,
+    verbose?: boolean,
+    callbackManager?: CallbackManager
```
agola11, line 51: This seemed like the most bw-compat option, although ideally, we should do what @nfcampos did with `BaseLanguageModelParams`

langchain/src/chat_models/openai.ts
```diff
@@ -288,10 +288,10 @@ export class ChatOpenAI extends BaseChatModel implements OpenAIInput {
...
                 innerCompletion += part.delta?.content ?? "";
                 role = part.delta?.role ?? role;
-                this.callbackManager.handleNewToken?.(
-                  part.delta?.content ?? "",
-                  this.verbose
-                );
+                // this.callbackManager.handleNewToken?.(
+                //   part.delta?.content ?? "",
+                //   this.verbose
+                // );
```
agola11, line 294: @nfcampos -- there's some issues with using async callbacks for streaming i.e. async handleLLMNewToken in the LLM classes bc you're using createParser which is sync. What's the best workaround? I've commented out all the streaming callbacks and int tests that use streaming

langchain/src/agents/tools/base.ts
```diff
@@ -1,5 +1,41 @@
...
+  verbose: boolean;
+
+  callbackManager: CallbackManager;
+
+  constructor(verbose?: boolean, callbackManager?: CallbackManager) {
+    this.verbose = verbose ?? getVerbosity();
+    this.callbackManager = callbackManager ?? getCallbackManager();
+  }
+
+  protected abstract _call(arg: string): Promise<string>;
```
agola11, line 20: All tools now implement abstract `_call` instead of `call` s.t. we are able to plug in callbacks

langchain/src/agents/tests/calculator.test.ts
```diff
@@ -3,30 +3,30 @@ import { Calculator } from "../tools/calculator.js";
 
 test("Calculator tool, sum", async () => {
   const calculator = new Calculator();
-  const result = await calculator.call("1 + 1");
+  const result = await calculator._call("1 + 1");
```
agola11, line 6: will revert all of these tests to use `call`

langchain/src/base_language/index.ts
```diff
@@ -0,0 +1,40 @@
+import { BasePromptValue, LLMResult } from "../schema/index.js";
```
agola11, line 1: I had to move this out to avoid circular dependency: BaseLanguageModel in langchain/schema used CallbackManager, CallbackManager used LLMResult and other things in langchain/schema

langchain/src/callbacks/utils.ts
```diff
@@ -0,0 +1,49 @@
+import * as process from "process";
+import { LangChainTracer } from "./tracers.js";
+import { CallbackManager, ConsoleCallbackHandler } from "./base.js";
+
+export class SingletonCallbackManager extends CallbackManager {
```
agola11, line 5: I made a `SingleTonCallbackManager` since there should be ideally one instance of `LangChainTracer` (since it is stateful, it has a stack). @nfcampos -- I'm not knowledgable about typescript to know if this will breakdown with concurrency. I know that, if you have multiple async coroutines in python, sharing a tracer is bad news since there could be collisions when the trace is being collected (everything runs on the same thread).

langchain/src/chains/base.ts
nfcampos: This is an abstract class, so the only people we'd be breaking by changing these args would be someone who has created a custom subclass, not sure how likely that is

PR #306, opened 2023-03-13
Support concurrent runs in tracer

nfcampos:
Note this currently does not try to guess which concurrent run a child run should get associated to, that would be an improvement to make on top of this


agola11:
Can you elaborate on how you were planning to "guess which concurrent run a child run should get associated to?"

This won't solve the concurrency problem as it stands if using a shared tracer, as concurrent runs all get pushed to the same stack

agola11:
have a couple of ideas I'll try out

PR #307, opened 2023-03-13
json document loader can handle more complex json

MaximeThoonsen:


nfcampos:
Hi @MaximeThoonsen thanks for this, would it be possible to do this all inside the existing JSONLoader, by changing that one to accept either one JSONPointer or an array of multiple JSONPointer?

MaximeThoonsen:
Hey @nfcampos. 
Sure I can add the handling of arrays of JSONpointer. 

On top of that, in this version of the code, the JSONpointer loader is for use case where you want exact match of the path of the data you want to retrieve. And the JSON loader can handle more unstructured JSON file to retrieve strings from matching keys in any level of the JSON object. I did it to be able to handle [Corrado's use case on discord](https://discord.com/channels/1038097195422978059/1076182741374214285/1083153092658331658).

I see 3 options if I move everything in the JSON loader:

1. I remove the logic I implemented around what I call "keys".
2. I can make the JSON pointer follows quite the same behaviour (ie not looking for exact match from the root node)
3. I add an argument to the constructor of the JSONLoader class to handle both cases (strict and not strict)

I see value of both usage but I understand there is a tradeoff with simplicity of documentation. 
Tell me what you prefer.

MaximeThoonsen:
Hey @nfcampos,
I chose to do 2). It should be ok for most use of json use case including the one from [Corrado](https://discord.com/channels/1038097195422978059/1076182741374214285/1083153092658331658).
(I am AFK until Sunday evening)

PR #309, opened 2023-03-13
feat: executor returns additional object from agent prepareForOutput

jacobrosenthal:
Im looking to say return suggestions from final output of an agent, but since theres no `_call` to override in agents I cant use the chain style of doing that.

from agent, Im overriding like
```
  prepareForOutput(
    _: Record<string, any>,
    __: AgentStep[]
  ): Promise<Record<string, any>> {
    return new Promise((resolve) => {
      resolve({ suggestions: [] });
    });
  }
```

this seems to be one way of doing that. I wonder if that callback wouldnt want the input as well....

Not wedded to naming or anything


langchain/src/agents/executor.ts
```diff
@@ -72,11 +72,13 @@ export class AgentExecutor extends BaseChain {
 
     const getOutput = async (finishStep: AgentFinish) => {
       const { returnValues } = finishStep;
+      const additional = await this.agent.prepareForOutput(returnValues, steps);
+
       if (this.returnIntermediateSteps) {
-        return { ...returnValues, intermediateSteps: steps };
+        return { ...returnValues, intermediateSteps: steps, ...additional };
       }
       await this.callbackManager.handleAgentEnd(finishStep, this.verbose);
```
jacobrosenthal, line 80: Seems likely youd want to track these in the trace as well? not sure though havent looked at this new stuff yet

hwchase17 (approved):
seems good to me!

agola11 (approved):

PR #312, opened 2023-03-14
convo agent chat

hwchase17:


langchain/src/chains/base.ts
```diff
@@ -98,8 +97,7 @@ export abstract class BaseChain implements ChainInputs {
    * Wraps {@link _call} and handles memory.
    */
   async call(values: ChainValues): Promise<ChainValues> {
-    const fullValues = deepcopy(values);
-
+    const fullValues = values;
```
hwchase17, line 100: @agola11 @nfcampos this was giving me issues. basically deepcopy was not respecting the types of the objects, and was convertiing HumanChatMessage just to generic `{"text": "foo"}` object, which was leading to issues down the road

i **think** this deepcopy is not super important... but if theres a better way to do this, just to be safe, it might be good...

langchain/src/memory/base.ts
```diff
@@ -21,13 +21,16 @@ export abstract class BaseMemory {
   ): Promise<void>;
 }
 
-export const getInputValue = (inputValues: InputValues) => {
+export const getInputValue = (inputValues: InputValues, inputKey?: string) => {
```
agola11, line 24: can you explain why these changes are needed? If I'm understanding correctly, `inputValues` can only have one key rn, why not just always grab the only value?

langchain/src/agents/chat_convo/index.ts
```diff
@@ -0,0 +1,195 @@
...
+    const invalidTool = tools.find((tool) => !tool.description);
+    if (invalidTool) {
+      const msg =
+        `Got a tool ${invalidTool.name} without a description.` +
+        ` This agent requires descriptions for all tools.`;
+      throw new Error(msg);
+    }
+  }
+
+  constructScratchPad(steps: AgentStep[]): BaseChatMessage[] {
```
agola11, line 114: do we want to construct the scratchpad as messages for the regular chat agent too?

langchain/src/memory/base.ts
hwchase17: for conversational agent, there ends up being multiple keys (the memory key and the input key)

langchain/src/agents/chat_convo/index.ts
hwchase17: probably not

langchain/src/agents/chat_convo/index.ts
```diff
@@ -0,0 +1,195 @@
...
+  BaseChatMessage,
+  AIChatMessage,
+  HumanChatMessage,
+} from "../../schema/index.js";
+import {
+  BaseOutputParser,
+  SerializedOutputParser,
+} from "../../output_parsers/base.js";
+
+export class AgentOutputParser extends BaseOutputParser {
```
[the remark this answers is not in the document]
hwchase17: nah just used here

langchain/src/chains/base.ts
nfcampos: I don't think the deep copy is needed either, I think the fix would be `const fullValues = deepcopy(values) as typeof values;` but I couldn't reproduce whatever issue you were facing @hwchase17

PR #313, opened 2023-03-14
Zapier integration

agola11:
* Add Zapier wrapper (based on https://github.com/hwchase17/langchain/pull/1539/)
* Add Zapier tool, toolkit
* Add example `zapier_mrkl`


examples/src/agents/zapier_mrkl.ts
```diff
@@ -0,0 +1,25 @@
+import { OpenAI } from "langchain";
+import { initializeAgentExecutor, ZapierToolKit } from "langchain/agents";
+import { ZapierNLAWrapper } from "langchain/zapier";
```
nfcampos, line 3: Do we really want to have an entrypoint dedicated to zapier? Why not stick this in tools or something like that

nfcampos (approved):
lgtm with 1 comment

examples/src/agents/zapier_mrkl.ts
agola11: did a small refactor

PR #315, opened 2023-03-14
handleLLMNewToken should not use the overall verbose flag of the model, as it is only fired when asked for by the user

nfcampos:


agola11:
Can we just remove the verbose checks in the CallbackManager for this callback?

nfcampos:
I thought about that, but there may be cases of future LLMs that always call handleLLMNewToken even when the user didn't request streaming, so felt better this way

PR #323, opened 2023-03-14
Add eslint rule to detect (some) side effects (that prevent tree shaking)

nfcampos:


nfcampos:
https://github.com/mrdoob/three.js/issues/24199
https://github.com/evanw/esbuild/issues/1420
https://github.com/mapbox/node-pre-gyp/issues/661
https://esbuild.github.io/api/

langchain/src/cache.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable tree-shaking/no-side-effects-in-initialization */
```
nfcampos, line 1: remove this

agola11 (approved):
lgtm

PR #327, opened 2023-03-14
BaseLLM -> Base language

jacobrosenthal:
This mostly does the chains.  I havent looked into agents yet. I know recent Prs are making a whole new class for agent for the new model anyway..

I looked around the python I dont think ser/de is done there either so not sure of the format or if it needs to be supported in order to merge this.

But tests pass and the touched examples run with either model. I dont know if its time to default examples into the ChatOpenAI class yet?

I have not tested the langchain/base_language export as my typescript tooling is poor these days, still working that out but this would seem to do it.

### DIFF
--- a/langchain/src/chains/llm_chain.ts
+++ b/langchain/src/chains/llm_chain.ts
@@ -1,5 +1,6 @@
 import { BaseChain, ChainInputs } from "./base.js";
-import { BaseLLM, SerializedLLM } from "../llms/index.js";
+import { SerializedLLM } from "../llms/index.js";
+
 import { BaseMemory, BufferMemory } from "../memory/index.js";
 import {
   BasePromptTemplate,
@@ -97,7 +98,7 @@ export class LLMChain extends BaseChain implements LLMChainInput {
     >("prompt", data);
 
     return new LLMChain({
-      llm: await BaseLLM.deserialize(serializedLLM),
+      llm: await BaseLanguageModel.deserialize(serializedLLM),
       prompt: await BasePromptTemplate.deserialize(serializedPrompt),
     });
   }
@@ -105,7 +106,7 @@ export class LLMChain extends BaseChain implements LLMChainInput {
   serialize(): SerializedLLMChain {
     return {
       _type: this._chainType(),
-      // llm: this.llm.serialize(), TODO fix this now that llm is BaseLanguageModel
+      llm: this.llm.serialize(),
       prompt: this.prompt.serialize(),
     };
   }
--- a/langchain/src/chains/prompt_selector.ts
+++ b/langchain/src/chains/prompt_selector.ts
@@ -1,5 +1,4 @@
 import { BaseChatModel } from "../chat_models/base.js";
-import { BaseLLM } from "../llms/base.js";
 import { BasePromptTemplate } from "../prompts/base.js";
 import { BaseLanguageModel } from "../base_language/index.js";
 
@@ -38,8 +37,8 @@ export class ConditionalPromptSelector extends BasePromptSelector {
   }
 }
 
-export function isLLM(llm: BaseLanguageModel): llm is BaseLLM {
-  return llm instanceof BaseLLM;
+export function isLLM(llm: BaseLanguageModel): llm is BaseLanguageModel {
+  return llm instanceof BaseLanguageModel;
 }
 
 export function isChatModel(llm: BaseLanguageModel): llm is BaseChatModel {
--- a/docs/docs/modules/chains/chat_vector_db_qa.md
+++ b/docs/docs/modules/chains/chat_vector_db_qa.md
@@ -42,7 +42,7 @@ In this code snippet, the fromLLM method of the ChatVectorDBQAChain class has th
 
 ```typescript
 static fromLLM(
-  llm: BaseLLM,
+  llm: BaseLanguageModel,
   vectorstore: VectorStore,
   options?: {
     questionGeneratorTemplate?: string;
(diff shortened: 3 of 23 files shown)


langchain/src/chains/prompt_selector.ts
```diff
@@ -38,8 +37,8 @@ export class ConditionalPromptSelector extends BasePromptSelector {
   }
 }
 
-export function isLLM(llm: BaseLanguageModel): llm is BaseLLM {
-  return llm instanceof BaseLLM;
+export function isLLM(llm: BaseLanguageModel): llm is BaseLanguageModel {
```
hwchase17, line 40: this should actually stay the same - its checking the subtype

langchain/src/chains/tests/combine_docs_chain.test.ts
```diff
@@ -1,37 +1,55 @@
...
-import { LLMResult } from "../../schema/index.js";
+import { BasePromptValue, LLMResult } from "../../schema/index.js";
 
 test("Test MapReduceDocumentsChain", async () => {
   let nrMapCalls = 0;
   let nrReduceCalls = 0;
 
-  class FakeLLM extends BaseLLM {
-    _llmType(): string {
+  class FakeLLM extends BaseLanguageModel {
```
hwchase17, line 14: i think its fine for this to remain the same and not change? eg extend BaseLLM

jacobrosenthal:
@hwchase17 Addressed those nits. 

Im also sneaking some exports in here as the reason I need this is Im trying to build/extend agents on the new BaseLanguageModel out of tree

related, since I have your attention :) 

custom agents need to be able to hook agents output, much like you can for custom chains  (instead of extending the entire executor to get access to `_call`) https://github.com/hwchase17/langchainjs/pull/309

langchain/src/chains/llm_chain.ts
```diff
@@ -97,15 +98,15 @@ export class LLMChain extends BaseChain implements LLMChainInput {
     >("prompt", data);
 
     return new LLMChain({
-      llm: await BaseLLM.deserialize(serializedLLM),
+      llm: await BaseLanguageModel.deserialize(serializedLLM),
```
nfcampos, line 101: This method throws a "not implemented" error, therefore this will fail

langchain/src/prompts/index.ts
```diff
@@ -28,3 +28,4 @@ export {
   SerializedMessagePromptTemplate,
   SerializedChatPromptTemplate,
 } from "./serde.js";
+export { interpolateFString } from "./template.js";
```
nfcampos, line 31: This is an implementation detail, which we do not want exported

nfcampos (commented):
Looks good overall, needs 2 changes

langchain/src/prompts/index.ts
jacobrosenthal: I often need to create my agents out of tree, which means implementing createPrompt which means people will want to be able to export this or something like it

langchain/src/chains/llm_chain.ts
jacobrosenthal: That seems like an ok solution, since its not implemented right? Other options?

langchain/src/prompts/index.ts
nfcampos: Yea, for that you should use `parseTemplate` or `renderTemplate` from the same file. I'd be happy to see those exported to everyone.

jacobrosenthal:
Why doesn't agent use those then. People are going to copy the agent they
want to extend

nfcampos:
That's something we need to fix, I'll create an issue to make agent use parseTemplate, thanks for noticing

langchain/src/chains/llm_chain.ts
jacobrosenthal: Probably core team has more view into all the serde stuff. I dont use it atm, and i have no idea what desired standard formats are. Ill have to leave that you all.

jacobrosenthal:
Rebased to fix the conflicts and squash this mess down.

jacobrosenthal:
Moved serde functions to base like Basellm. Not sure thats the best solution, but it matches upstream

Then add a test something like this
```
test.only("Test deserialize", async () => {
  const model = new OpenAI({ modelName: "text-ada-001" });
  const prompt = new PromptTemplate({
    template: "Print {foo}",
    inputVariables: ["foo"],
  });
  const chain = new LLMChain({ prompt, llm: model });

  const serialized = chain.serialize();
  console.log(serialized)
  const chain2 = await LLMChain.deserialize({ ...serialized });

  const res = await chain2.run("my favorite color");
  console.log({ res });

  // chain === chain2?
});
```
upstream prints this (no llm because wasnt supported)
```
{
  _type: 'llm_chain',
  prompt: {
    _type: 'prompt',
    input_variables: [ 'foo' ],
    output_parser: undefined,
    template: 'Print {foo}',
    template_format: 'f-string'
  }
}
```
this branch now prints
```
    {
      _type: 'llm_chain',
      llm: {
        model_name: 'text-ada-001',
        model: 'text-ada-001',
        temperature: 0.7,
        max_tokens: 256,
        top_p: 1,
        frequency_penalty: 0,
        presence_penalty: 0,
        n: 1,
        best_of: 1,
        logit_bias: undefined,
        stop: undefined,
        stream: false,
        apiKey: 'sk-xxxxxxxxxxxxxxxxxxxx',
        _type: 'openai',
        _model: 'base_llm'
      },
      prompt: {
        _type: 'prompt',
        input_variables: [ 'foo' ],
        output_parser: undefined,
        template: 'Print {foo}',
        template_format: 'f-string'
      }
    }
```

Is that enough implementation? Correct fields?
Should llm serialize your apiKey?
other tests I have api keys for still seem to pass, but theres basically no tests for serde

PR #341, opened 2023-03-15
Remove import of process module, this doesnt play well with browsers/edge

nfcampos:


langchain/src/callbacks/tracers.ts
```diff
@@ -1,4 +1,3 @@
-import * as process from "process";
```
nfcampos, line 1: @agola11 `process` is a global in node, and injected by bundlers in edge/browser/etc envs

agola11:
makes sense, thanks!

agola11 (approved):

ericlewis (approved):

PR #354, opened 2023-03-17
[WIP] Several changes to support concurrency with tracing and other callback handlers

agola11:
`yarn start agents/concurrent_mrkl` now just works:
Screenshot 2023-03-16 at 11 55 41 PM

Changes to callback handling/managing:
* Enable callback handlers to return a record
* Enable callback managers to return a record (which they aggregate from their child callback handlers)
* Have callbackhandlers take in optional `callerId` for `start` methods and `runId` for `error` and `end` methods.
* In tracer, ditch the idea of having a shared stack and shared execution order -- instead, create a map from runID to run. Add each child run to parent runs children by referencing `callerId`. On `end` and `error` methods, use the passed in `runId`. When you end a run with no caller_id, i.e. a top-level run, persist it

Changes to langchain modules:
* In each base class (i.e. any class with callback manager wired up), we now retrieve the `runId` from the callbackManager when tracing is enabled, pass that `runId` to child modules, and also pass it to the callback manager methods.
* The reason why pass the caller runId through rather than retrieving the childRun id from a return value is because changing the return value is a less bw-compat change than passing through an optional parameter. Additionally, it would mess up the simple return values we have for some convenience methods (like `chain.run`)
* You have access to the runId when using `generate` for llms and `call` for chains, it's in the LLMResult and ChainValues respectively.

The downside is that we now have to pass through the `runId` for in a chain or agent if we want it traced. I already did this for `agent` and `llmchain` but some others will need to be changed in order for tracing to work properly.


langchain/src/agents/tools/base.ts
```diff
@@ -19,21 +24,27 @@ export abstract class Tool {
...
     const _verbose = verbose ?? this.verbose;
-    await this.callbackManager.handleToolStart(
+    const values = await this.callbackManager.handleToolStart(
       { name: this.name },
       arg,
+      callerId,
       _verbose
     );
     let result;
+    const runId = values[TRACER_RUN_ID];
```
nfcampos, line 40: I don't think this is a good idea, we're leaking an implementation detail of the tracer (one of many callback handlers) into the whole codebase. This is why I had the callback manager be the one that assigns runIds. What's the advantage of this?

PR #355, opened 2023-03-17
json based convo agent - for discussion

jacobrosenthal:
This is a json based convo fork ive been playing with. It takes json only from chatgpt, defining its api via typescript examples. This is less token verbose and ive found it to be more reliable than the upstream one as constraining to a json array for actions really seems to give it less ways to screw up

I think the COT prompting with thought: is really necessary, and with the current agent interface its pretty ackwardly written. Not particularly extendable. 

Its a bit slower, because I dont use the stop, so it has to receive all the tokens. Maybe theres a way to reliably detect like the comma between two objects

For discussion, maybe someone has a way to clean it up or just abandon the idea that its extendable

~All the other commits are just Prs I needed to build this on top of out of tree, just the last commit is the agent when everything else clears.~
related to my request in https://github.com/hwchase17/langchainjs/pull/349

### DIFF
--- a/examples/src/agents/chat_convo.ts
+++ b/examples/src/agents/chat_convo.ts
@@ -1,0 +1,49 @@
+import {
+  Tool,
+  ChatConversationalAgent2,
+  AgentExecutor,
+} from "langchain/agents";
+import { CallbackManager } from "langchain/callbacks";
+import { ChatOpenAI } from "langchain/chat_models";
+import { LLMResult } from "langchain/schema";
+import { SerpAPI, Calculator } from "langchain/tools";
+
+export const run = async () => {
+  const callbackManager = CallbackManager.fromHandlers({
+    async handleLLMStart(_llm: { name: string }, prompts: string[]) {
+      console.log(JSON.stringify(prompts, null, 2));
+    },
+    async handleLLMEnd(output: LLMResult) {
+      for (const generation of output.generations) {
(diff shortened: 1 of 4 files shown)


jacobrosenthal:
Much of the time chatgpt responds with
```
langchain-examples:start:   "action_input": "pow(Harry Styles age, 0.23)",
```
but occasionally it uses
```
"action_input": `${Math.pow((new Date().getFullYear() - 1994), 0.23)}`
```
which kills the json parsing.. whats a good escape library for json parsing? Or any ideas on prompt engineering? If you tell it to never use ` in the prompt itll fail with
```
langchain-examples:start:   "action_input": "round(" + Math.pow(age, 0.23) + ")",
```

folke:
This is really funny. I also started using TypeScript types and initially used JSON. But as you also discovered, it's hard to parse invalid json.
I eventually just used the same types, but said it had to respond with YAML instead. It's much better in writing yaml in my experience.

jacobrosenthal:
Oh interesting. Or your agent? ;)

jacobrosenthal:
Pr. Pull request

jacobrosenthal:
I tried some yaml but didnt have a lot of luck, it doesnt feel constrained enough and it likes to not follow the schema. Not using certain fields randomly. This json is still working best for me, assuming im not using the math tool and no backticks show up.

PR #362, opened 2023-03-17
Update GPT-4 model context size

nauxliu:


nfcampos:
@nauxliu thanks for this. This isn't actually used for gpt-4 models, as like gpt-3.5-turbo it natively supports not passing max_tokens, which is equivalent to "as many tokens as possible", see https://platform.openai.com/docs/api-reference/chat/create#chat/create-max_tokens

PR #363, opened 2023-03-17
In LLMChain include format_instructions when formatting the prompt if the prompt accepts it

nfcampos:


langchain/src/chains/tests/llm_chain.test.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  expect(res).toEqual({ [chain.outputKey]: "a completion" });
+  expect(model._call).toHaveBeenCalledWith(
+    "Print my favorite color",
+    undefined
+  );
+});
+
+test("Test with output_parser without format_instructions", async () => {
+  const model = new FakeLLM();
```
agola11, line 65: Should we consider throwing an error if prompt has an output parser but no format_instructions variable?

agola11 (approved):
This is to get around needing to use `partialValues` right? Makes sense to me

langchain/src/chains/tests/llm_chain.test.ts
nfcampos: I guess we could, but would be less backwards compatible

nfcampos:
note to me: do same in python

Issue #365, opened 2023-03-17
Replace all direct usage of interpolateFString with renderTemplate/parseTemplate


nfcampos:
interpolateFString is an implementation that we don't want people depending on / seeing as an example of something they should use

PR #367, opened 2023-03-17
feat: export parseTemplate and renderTemplate

### LINKED ISSUE #366: chat convo agent should use `parseTemplate` or `renderTemplate` from the same file. I'd be happy to see those exported to everyone.
Yea, for that you should use `parseTemplate` or `renderTemplate` from the same file. I'd be happy to see those exported to everyone.

_Originally posted by @nfcampos in https://github.com/hwchase17/langchainjs/pull/327#discussion_r1140520534_

jacobrosenthal:
Closes #366 #365

### DIFF
--- a/langchain/src/agents/agent_toolkits/sql/sql.ts
+++ b/langchain/src/agents/agent_toolkits/sql/sql.ts
@@ -8,7 +8,7 @@ import {
 import { Toolkit } from "../base.js";
 import { BaseLLM } from "../../../llms/index.js";
 import { SQL_PREFIX, SQL_SUFFIX } from "./prompt.js";
-import { interpolateFString } from "../../../prompts/template.js";
+import { renderTemplate } from "../../../prompts/template.js";
 import { LLMChain } from "../../../chains/index.js";
 import { ZeroShotAgent, CreatePromptArgs } from "../../mrkl/index.js";
 import { AgentExecutor } from "../../executor.js";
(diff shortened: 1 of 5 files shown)


jacobrosenthal:
Also sneaking another few chat convo agent things in here. Im not like enamored with the Conversational Agent but I want to try to get the patterns right so agents are extendable in how Ive tended to use them. 

Some of the input can be copy pasted, but output parser comes with templates and other stuff. I know these will be standardized more in the parsers folder, but this one is defined here and should be exported.

I think a common use case im using is extending an agent I already mostly like, and maybe doing something in the new `prepareForOutput` to add or massage output. You want to wrap up all your prefix/suffix changes So the most minimal out of tree agent that for instance add follow up question suggestions would look something like
```
export class ChatConversationalAgentSuggestions extends ChatConversationalAgent {
  questionGeneratorChain: LLMChain

  constructor(llm: BaseLanguageModel,
    tools: Tool[],
    args: ChatConversationalAgentInput = {
      systemMessage: MY_BETTER_PREFIX,
      humanMessage: MY_NEW_SUFFIX,
    }) {

    const prompt = ChatConversationalAgentSuggestions.createPrompt(tools, args)
    const chain = new LLMChain({ prompt, llm })
    const questionGeneratorChain = new LLMChain({
      prompt: question_generator_prompt,
      llm,
    })
    const { outputParser = new ChatConversationalAgentOutputParser() } = args ?? {}

    super({
      llmChain: chain,
      allowedTools: tools.map((t) => t.name),
    }, outputParser)
    this.questionGeneratorChain = questionGeneratorChain
  }

  async prepareForOutput(
    _returnValues: Record<string, any>,
    _intermediateSteps: AgentStep[]
  ): Promise<Record<string, any>> {
    // generatesuggestions
    return { suggestions: [] }
  }
}
```

hwchase17 (approved):
seems reasonable to me, thanks for renaming stuff

PR #368, opened 2023-03-17
feat: Add Github Loader 

mayooear:
@nfcampos it's mayo here. As discussed, here is the first draft of the github loader that parses the files of a given github url into the `Document` format. This is done recurssively for directories (if it's not set to false).

You can test it with this url: "https://github.com/hwchase17/langchainjs/tree/main/langchain". It works on my end 

I have also filtered out binary files in the logic.

Note that the test file has not yet been included, as I anticipate you may want to make some changes via discussion first. Cheers.

### DIFF
--- a/langchain/src/document_loaders/github.ts
+++ b/langchain/src/document_loaders/github.ts
@@ -1,0 +1,135 @@
+import { Document } from "../document.js";
+import { BaseDocumentLoader } from "./base.js";
+
+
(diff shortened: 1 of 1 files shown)


agola11:
Do you mind adding an integration test and example like we do for other loaders? (see examples/src/document_loaders). Can take over doing that if you don't have time

PR #370, opened 2023-03-17
Use OpenAIChat class to initialize gpt-4 models

jacoblee93:
GPT-4 is classified the same way as GPT-3.5, and uses the `v1/chat/completions` endpoint rather than the `v1/completions` endpoint.


agola11 (approved):
thanks! just FYI we might deprecate this class soon in favor of ChatOpenAI (which uses chat based primitives)

PR #371, opened 2023-03-18
build: Improve exported types with declarationMap

zurawiki:
Modifies the global tsconfig.json to enable source maps from .d.ts files back to the original source code. This makes is easier for downstream developers to understand the internals of langchain.


nfcampos:
Thanks @zurawiki this is great. One question, for this to work we need to also update package.json to publish the src files to npm no? Otherwise the files it links to won't be there to be linked to

zurawiki:
Yes, that’s a good point! If one is using this package via NPM install, we will need to expose the sources somehow.

One downside I see with publishing the sources is an increase in package size. That said I’d prefer having the ability into LangChain sources when I’m developing my apps. That helps me debug what’s going wrong and propose any changes to upstream.

In case it helps, I’ve been referencing this resource for publishing npm libraries.

https://snyk.io/blog/best-practices-create-modern-npm-package/

This article recommends putting the source files in a `lib` folder. I know this is just one way to package libraries. I’m curious to learn more about the benefits of doing packing the apps other ways.

Roger Zurawicki
zura.wiki ( https://zura.wiki ) | linkedin.com/in/zurawiki ( http://linkedin.com/in/zurawiki )

PR #374, opened 2023-03-18
Adds support for memory and prompt passing to a ChatConversationalAgent

jacoblee93:
This PR adds a new method for initializing agent executors, `initializeAgentExecutorWithOptions`, which takes a`Tool[]` parameter and a `BaseLanguageModel` like the existing `initializeAgentExecutor`, but also takes an `options` object to support additional parameters without breaking existing code. 

This new method also adds `BufferMemory` to the executor when the agent type is set as `chat-conversational-react-description`.

We'll now be able to pass a prompt into a `ChatConversationalAgent` and have built in memory like this:

```js
import { initializeAgentExecutorWithOptions } from "langchain/agents";
const model = new OpenAIChat({
  openAIApiKey: process.env.OPENAI_API_KEY,
  modelName: process.env.OPENAI_MODEL_NAME
});
const tools = [new Calculator()];
const executor = await initializeAgentExecutorWithOptions(
  tools,
  model,
  {
    agentType: "chat-conversational-react-description",
    verbose: true,
    prompt: `Answer the following questions as best you can, but speaking as a pirate might speak.`
  }
);

const result0 = await executor.call({input: "How's your day going?"});

// result0.output
// Arr, me day be goin' well, matey! What be ye needin' help with?

const result1 = await executor.call({input: "What is 9 raised to the second power?"});

// result1.output
// Arr, 9 raised to the second power be 81, matey!

const result2 = await executor.call({ input: "What is that result times 2?" });

// result2.output
// Arr, the result of 9 raised to the second power times 2 be 162, matey!
```


hwchase17:
the refactor to use `Options` is done to allow for more optional params in a more elegant way?

If thats the case, this should be the interface we promote, correct? if so, maybe lets update the examples to reflect that

langchain/src/agents/chat_convo/index.ts
```diff
@@ -127,11 +132,10 @@ export class ChatConversationalAgent extends Agent {
    * @param args.prefix - String to put before the list of tools.
    */
   static createPrompt(tools: Tool[], args?: CreatePromptArgs) {
-    const {
-      systemMessage = PREFIX,
-      humanMessage = SUFFIX,
-      outputParser = new AgentOutputParser(),
-    } = args ?? {};
+    const systemMessage =
```
hwchase17, line 135: why is the prefix/suffix helpful here? its just same as system/human right? seems confusing to have two with same name

jacoblee93:
> the refactor to use `Options` is done to allow for more optional params in a more elegant way?
> 
> If thats the case, this should be the interface we promote, correct? if so, maybe lets update the examples to reflect that

Yes, the existing method had the following signature:

```js
export const initializeAgentExecutor = async (
  tools: Tool[],
  llm: BaseLanguageModel,
  agentType = "zero-shot-react-description",
  verbose = false,
  callbackManager: CallbackManager = getCallbackManager()
): Promise<AgentExecutor>
```

And tacking on more parameters at the end seemed like the wrong approach. I can update this PR with some docs updates tomorrow!

langchain/src/agents/chat_convo/index.ts
jacoblee93: The comment block above that line may be wrong (maybe copy-pasted from elsewhere?) but they implied that `args` should contain:
```js
   * @param args.suffix - String to put after the list of tools.
   * @param args.prefix - String to put before the list of tools.
```
I'm a bit new to Langchain's/prompt terminology in general so `prefix/suffix` seemed clearer but I can update the comment block and change the parameter names to `systemMessage` and `humanMessage` instead if that's preferred.

jacoblee93:
@hwchase17 I've updated the docs and changed the parameter names as requested, and found out how to get rid of the ugly `.memory` hack:
https://github.com/hwchase17/langchainjs/pull/374/files#diff-59a2c3ddfc3c7e38be163dd2b5f2a48f76da2b51370b98755cf6a24a2d290c70R42

@GautierT I also updated the `initializeAgentExecutorWithOptions` method to allow you to pass in a custom `memory` arg to override the default settings - does that fit your use case?

jacoblee93:
Ok, merged in `main` as well. @hwchase17 PTAL when you get some time!

jacoblee93:
The formatting check is passing on my end - is it possible to rerun it? Might be a flake or have run on a previous commit.

jacoblee93:
> If you look close it says you need to run prettier on
> 
> ```
> docs:format:check: cache miss, executing 01977ba0ac428b4d
> langchain:format:check: src/agents/chat_convo/index.ts
> ```

I was trying it with `yarn format:check` as specified in `CONTRIBUTING.md` but didn't see any warnings or alerts - could definitely be something wrong with my setup though:

Screenshot 2023-03-19 at 6 09 41 PM

jacobrosenthal:
If I do yarn `format:check` in your branch which is what that test runs, it shows the error and do a `yarn format` it does change it

examples/src/agents/chat_convo_with_tracing.ts
```diff
@@ -8,12 +8,10 @@ export const run = async () => {
...
-    tools,
-    model,
-    "chat-conversational-react-description",
-    true
-  );
+  const executor = await initializeAgentExecutorWithOptions(tools, model, {
+    agentType: "chat-conversational-react-description",
+    verbose: true,
+  });
   executor.memory = new BufferMemory({
```
agola11, line 15: This is no longer needed with your new changes right? `initializeAgentExecutorWithOptions` handles initializing the memory for the executor

agola11 (commented):
Thanks @jacoblee93! Overall lgtm, just a small comment and need to fix formatting

jacoblee93:
> Thanks @jacoblee93! Overall lgtm, just a small comment and need to fix formatting

Ah, yes! Fixed. Also fixed a typo in the Zapier docs page I noticed.

> If I do yarn `format:check` in your branch which is what that test runs, it shows the error and do a `yarn format` it does change it

Strangely, pushing this unrelated docs fix triggered the CI again and it passed this time without issue 🤷.

jacoblee93:
Ok, sounds good!

langchain/src/agents/chat_convo/index.ts
```diff
@@ -49,10 +50,14 @@ export class ChatConversationalAgentOutputParser extends BaseOutputParser {
 }
 
 export type CreatePromptArgs = {
-  /** String to put after the list of tools. */
+  /** DEPRECATED: String to put after the list of tools. */
   systemMessage?: string;
-  /** String to put before the list of tools. */
+  /** DEPRECATED: String to put before the list of tools. */
   humanMessage?: string;
```
jacoblee93, line 56: I opted to deprecate these properties to be in line with the other agents - all prompt args should probably all inherit from the same type anyway, no?

langchain/src/agents/chat_convo/index.ts
jacoblee93: I feel like we should deprecate these `systemMessage` and `humanMessage` properties to be in line with the other agents, or deprecate `prefix` on `suffix` on the other agents to avoid confusion here.

langchain/src/agents/initialize.ts
```diff
@@ -46,3 +47,68 @@ export const initializeAgentExecutor = async (
...
+      const executor = AgentExecutor.fromAgentAndTools({
+        agent: ChatConversationalAgent.fromLLMAndTools(llm, tools, {
+          prefix: options.prefix,
+          suffix: options.suffix,
+        }),
+        tools,
+        returnIntermediateSteps,
+        verbose,
+        callbackManager,
+        memory: options.memory,
```
jacoblee93, line 106: @nfcampos I removed the memory default here, but don't love the fact that we need to initialize memory with specific options like `memoryKey: "chat_history"`. Would it make sense to make a wrapper class with those defaults? 

Or even just change the memory key in `BufferMemory` to `chat_history`?

nfcampos: This agent is "special" in that it actually is designed to work with memory, the other ones aren't. I agree the `chat_history` hidden requirement is bad, but we do need to default here to creating a memory

langchain/src/agents/chat_convo/index.ts
nfcampos: So these have different arg names because they are designed to work with chat models, instead of string LLMs as the other agents. Hence the different arg names. Thoughts?

jacoblee93: Ah ok understood, thanks for the context. What would you think of subclassing it into a `ChatAgent`?

Happy to revert and leave it the way it was if you prefer that.

langchain/src/agents/initialize.ts
jacoblee93: I also changed the options input syntax for `initializeAgentExecutorWithOptions` to take a `promptArgs` option that will just get passed through directly to `fromLLMAndTools`, so we don't need to worry about conflating `systemMessage` and `prefix` - the user can look at the specific agent they're using and figure out which one is appropriate.

Longer term, subclassing at least the prompt args could make sense to me so we could do something like:
```js
promptArgs?: AgentCreatePromptArgs | ChatAgentCreatePromptArgs;
```
`ChatAgentCreatePromptArgs` could also take a memory key too and decouple the hidden reliance on a `memoryKey` set to `chat_history`.

langchain/src/agents/chat_convo/index.ts
jacoblee93: Added a bit more discussion here: https://github.com/hwchase17/langchainjs/pull/374#discussion_r1151175672

PR #376, opened 2023-03-18
Change url used in getActionRequest

GautierT:
The url use should be "https://nla.zapier.com/api/v1/dynamic/" when calling an action.

If not we have a 400 Bad Request Error.


nfcampos:
@agola11 check this one

agola11:
@GautierT I'm not getting any bad request errors, everything is still working for me with the existing endpoint. In fact, even the Zapier team wrote our python integration and even they used the non-dynamic endpoint in their wrapper: https://github.com/hwchase17/langchain/blob/master/langchain/utilities/zapier.py

Where did you see that it should be v1/dynamic?

PR #381, opened 2023-03-19
Support concurrency with callbacks by passing through callback manager

agola11:
**Context**
This PR attempts to solve the following problems without any type of context var/async local storage which are either python/node only (so can't use and both python and js should have similar ways of doing things):
1) Support tracing and callbacks in concurrent environments
2) Scope a callback manager with a tracer or other handler to a single agent/chain run while avoiding the following code:
```typescript
// This will work, because each executor has its own Tracer, avoiding concurrency issues.
// VERY cumbersome
const executors = [];
for (let i = 0; i < 3; i += 1) {
  const callbackManager = new CallbackManager();
  callbackManager.addHandler(new ConsoleCallbackHandler());
  callbackManager.addHandler(new LangChainTracer());

  const model = new OpenAI({ temperature: 0, callbackManager });
  const tools = [new SerpAPI(), new Calculator()];
  for (const tool of tools) {
    tool.callbackManager = callbackManager;
  }
  const executor = await initializeAgentExecutor(
    tools,
    model,
    "zero-shot-react-description",
    true,
    callbackManager
  );
  executor.agent.llmChain.callbackManager = callbackManager;
  executors.push(executor);
}

const results = await Promise.all(
  executors.map((executor) => executor.call({ input }))
);
for (const result of results) {
  console.log(`Got output ${result.output}`);
}
```
3. Developers want to access the id of the trace being recorded in order to pull it up later via API.

To get around this, we can pass through the CallbackManager and have it also handle the creation of a runId for each run. We need to get rid of the tracer's stack and, instead, explicitly assign the parentId for a specific run. This is because, in the middle of a run, we could spawn multiple concurrent child runs (this would be the case with multi-output agents — in each step, we could spawn multiple tool runs concurrently). To avoid having to pass the parent run id to each call of the callback manager, we can wrap this with a child CallbackManager. The child callback manager would get a `parentRunId` from the parent callback manager and pass it through to the callbacks.

**TL;DR/Key Changes**
1) Pass through `CallbackManager` instance in `Agent`, `Chain`, and `LLM` by taking an extra optional argument in key methods (run, generate, call, etc)
2) Bake the `currentRunId` and `parentRunId` in the `CallbackManager` to avoid having to pass them around. Make `currentRunId` and `parentRunId` is immutable. Make each `CallbackManager` `handle*Start` call return a bound `CallbackManager` bound to that specific run instance.
3) Implement a `getChild` method on the `CallbackManager` that assigns the `parentRunId` of the child `CallbackManager` to the `currentRunId` of parent bound `CallbackManager`
4) We still need to have the option of assigning a `CallbackManager` to the object, we use this for streaming and other applications. In order to do this, we need to merge callbackmanagers. I added a `configureCallbackManager` that merges the pass-through manager with the one on the object. In order to avoid passing through these object-specific callback handlers through to child runs, there is a concept of `inheritedHandlers` that get inherited.
5) CallbackManagers are now optional. The verbose flag now has the following behavior: ALWAYS trigger callbacks on the callbackmanager if one has been passed through, regardless of verbose flag. If a callback manager has not been passed through AND verbose flag is set, create callback manager with a stdout/console callbacks and use that. If a callbackmanager has been passed through AND verbose flag is set, add a console/stdout callback handler to the passed through callbackmanager.
6) When you pass a callbackmanager to children runs, we need to be careful to pass the unaltered passed-in callback manager and not one that merges the local callbackmanager with one attached to the class. However, the `currentRunId` needs to be updated on the one being passed through.

**Todos**

- [x] Update callback manager documentation, any references to verbose
- [x] Add a couple of test cases that ensure the proper callbackmanager is being passed through (local one and not merged one)


langchain/src/agents/executor.ts
```diff
@@ -79,20 +83,28 @@ export class AgentExecutor extends BaseChain {
       if (this.returnIntermediateSteps) {
         return { ...returnValues, intermediateSteps: steps, ...additional };
       }
-      await this.callbackManager.handleAgentEnd(finishStep, this.verbose);
+      await this.callbackManager.handleAgentEnd(
```
nfcampos, line 86: Shouldn't this now be just callbackManager. ?

langchain/src/agents/executor.ts
```diff
@@ -79,20 +83,28 @@ export class AgentExecutor extends BaseChain {
...
+        finishStep,
+        runId ?? "",
```
nfcampos, line 88: Assigning a constant runId as a fallback value sounds like it will produce some hard to track down bugd

langchain/src/agents/tools/base.ts
```diff
@@ -17,23 +17,32 @@ export abstract class Tool {
...
-    const _verbose = verbose ?? this.verbose;
-    await this.callbackManager.handleToolStart(
+  protected abstract _call(
+    arg: string,
+    callbackManager?: CallbackManager
+  ): Promise<string>;
+
+  async call(
+    arg: string,
+    verbose?: boolean,
```
nfcampos, line 27: This verbose argument sounds useless now?

langchain/src/agents/tools/base.ts
```diff
@@ -17,23 +17,32 @@ export abstract class Tool {
...
+    callbackManager?: CallbackManager
+  ): Promise<string> {
+    const verbose_ = verbose ?? this.verbose;
+    const callbackManager_ = callbackManager ?? this.callbackManager;
```
nfcampos, line 31: To ensure this is all working correctly I'd actually completely remove this.callbackManager. Thoughts?

langchain/src/agents/executor.ts
agola11: yup good catch

langchain/src/agents/tools/base.ts
agola11: yeah probably a good idea

langchain/src/agents/executor.ts
nfcampos: But that assumes people will always use it correctly, we know they won't, and should design for it imo

nfcampos: And eg imagine you're running a multi output agent, or a mapreducechain, isn't that correct use of this that produces multiple simultaneous async calls to the same callbackManager?

agola11: [resolved offline]

* same callback not meant to be used in async, can clone to guard against that
* mapreduce chain is non issue, issues one callback before all the calls

langchain/src/base_language/index.ts
```diff
@@ -12,6 +12,45 @@ export type SerializedLLM = {
...
+ */
+export abstract class BaseLangChain {
+  /**
+   * Whether to print out response text.
+   */
+  verbose: boolean;
+
+  callbackManager?: CallbackManager;
+
+  protected configureCallbackManager(
```
nfcampos, line 26: I think this belongs in the CallbackManager class, namely as a `static` method

langchain/src/callbacks/base.ts
```diff
@@ -93,21 +131,43 @@ export abstract class BaseCallbackManager extends BaseCallbackHandler {
...
+    this._currentRunId = currentRunId;
+    this._parentRunId = parentRunId;
+  }
+
+  get currentRunId(): string | undefined {
+    return this._currentRunId;
+  }
+
+  // needed to avoid ESLint no-param-reassign error
+  setCurrentRunId(runId: string | undefined) {
```
nfcampos, line 152: IMO this method should not exist, and currentRunId should be immutable after set in the constructor, as we discussed. If currentRunId needs to change for any reason a new instance should be created

langchain/src/callbacks/base.ts
```diff
@@ -333,9 +441,21 @@ export class CallbackManager extends BaseCallbackManager {
...
 
+  getChild(): CallbackManager {
+    const manager = new CallbackManager(undefined, this._currentRunId);
+    manager.setHandlers(this.handlers);
+    return manager;
+  }
+
+  copy(additionalHandlers: BaseCallbackHandler[] = []): CallbackManager {
+    const manager = new CallbackManager(this._currentRunId, this._parentRunId);
+    manager.setHandlers([...this.handlers, ...additionalHandlers]);
```
nfcampos, line 452: This line assumes handlers are immutable, which is true of the ones we write, but is it true of the ones other folks will write? ie. if handlers do have state, this becomes then a shallow copy of the callback manager, still liable to cause issues. An alternative pattern would be

```js
manager.setHandlers([...this.handlers, ...additionalHandlers].map(handler => handler.copy()))
```

Where we provide a default implementation of `copy` in handler base class

langchain/src/callbacks/base.ts
```diff
@@ -333,9 +441,21 @@ export class CallbackManager extends BaseCallbackManager {
...
+    return manager;
+  }
+
   static fromHandlers(handlers: BaseCallbackHandlerMethods) {
     class Handler extends BaseCallbackHandler {
-      alwaysVerbose = true;
+      name = "handler";
```
nfcampos, line 458: Hmm should this one have a unique name? eg. `uuid()`

langchain/src/llms/openai.ts
```diff
@@ -302,10 +308,9 @@ export class OpenAI extends BaseLLM implements OpenAIInput {
                       choice.finish_reason = part.finish_reason;
                       choice.logprobs = part.logprobs;
                       // eslint-disable-next-line no-void
-                      void this.callbackManager.handleLLMNewToken(
-                        part.text ?? "",
-                        true
-                      );
+                      void this.configureCallbackManager(
```
nfcampos, line 311: So we create a new callback manager for every token logged? That's a lot of callback managers

langchain/src/tools/base.ts
```diff
@@ -1,39 +1,39 @@
...
+    callbackManager?: CallbackManager
+  ): Promise<string>;
+
+  async call(
+    arg: string,
+    _verbose?: boolean,
+    callbackManager?: CallbackManager
+  ): Promise<string> {
+    const localCallbackManager = this.configureCallbackManager(callbackManager);
+    await localCallbackManager?.handleToolStart({ name: this.name }, arg);
```
nfcampos, line 25: Wouldn't it make more sense to return a bound callback manager copy from `handleToolStart` rather than have 5 lines of code here?

langchain/src/tools/base.ts
```diff
@@ -1,39 +1,39 @@
...
+    if (callbackManager) {
+      callbackManager.setCurrentRunId(localCallbackManager?.currentRunId);
```
nfcampos, line 27: Why?

langchain/src/llms/openai.ts
```diff
@@ -235,7 +237,11 @@ export class OpenAI extends BaseLLM implements OpenAIInput {
    * const response = await openai.generate(["Tell me a joke."]);
    * ```
    */
-  async _generate(prompts: string[], stop?: string[]): Promise<LLMResult> {
+  async _generate(
+    prompts: string[],
+    stop?: string[],
+    callbackManager?: CallbackManager
```
nfcampos, line 243: this creates an issue with this PR https://github.com/hwchase17/langchainjs/pull/47

nfcampos (commented):
Some comments, I think easier to handle/reply to these before I review the rest

nfcampos:
@agola11 Thoughts on removing `verbose` arg to tools? unclear to me what purpose it serves

langchain/src/callbacks/base.ts
agola11: Yeah that's fair. I'm not too worried about the currentRunId changing around a lot as each time, before the callbackManager is used, it's cloned. But what you suggested sounds like a better abstraction

langchain/src/callbacks/base.ts
agola11: make sense. To be pedantic here, our tracer maintains a map of runId -> run now so not technically immutable but it still seems to work for our logic

langchain/src/llms/openai.ts
agola11: yep, can fix

langchain/src/tools/base.ts
agola11: I agree it's a bit cumbersome but this is something that needs to just be done in the base classes I'm pretty sure. I don't envision users needing to do this in custom chains. The reason for doing this is to update the callbackmanager that has been passed through with the currentRunId of the callbackManager copy created with `configureCallbackManager`. We need to preserve the passed-through one to avoid leaking callback handlers in a parent class not present in a child class (like if llmchain has consolecallback but not llm)

agola11:
@nfcampos yes we should remove verbose on tool. We can just set on the tool object itself

langchain/src/chat_models/base.ts
```diff
@@ -38,28 +42,31 @@ export abstract class BaseChatModel extends BaseLanguageModel {
...
+    stop?: string[],
+    callbackManager?: CallbackManager
   ): Promise<LLMResult> {
     const generations: ChatGeneration[][] = [];
     const llmOutputs: LLMResult["llmOutput"][] = [];
     const messageStrings: string[] = messages.map((messageList) =>
       getBufferString(messageList)
     );
-    await this.callbackManager.handleLLMStart(
+    const callbackManager_ = await this.configureCallbackManager(
```
nfcampos, line 53: I still think this should move to CallbackManager? (I get this is accessing properties on instance, but those can be passed in?

PR #385, opened 2023-03-19
feat: Implement Chroma fromExistingCollection

Alw3ys:
Currently there's no possibility to load an existing imported vectorStore unless loading from document again using Chroma.fromDocuments, making using an external database not useful.

With this pull request a method fromExistingCollection is implemented that provides the library the possibility to load collections that are already loaded.

Happy answer any questions about the Pull Request :)

Available here on discord Alw3ys#5643


langchain/src/vectorstores/chroma.ts
```diff
@@ -38,6 +38,18 @@ export class Chroma extends VectorStore {
...
   }
 
+  async loadExistingCollection() {
+    if (!this.index) {
+      const { ChromaClient } = await Chroma.imports();
+      this.index = new ChromaClient(this.url);
+      try {
+        await this.index.getCollection(this.collectionName);
+      } catch {
+        // ignore error
```
nfcampos, line 48: I don't think we'd want to ignore this error. Any thoughts on this @jeffchuber ?

Alw3ys: I must add this.index.getCollection is not actually being used the value it returns. is more to force the connection to chroma otherwise it doesn't happen. and if we should handle it here as you suggest we should change the whole class too since is not handled anywhere else.

jeffchuber: I think the right way to do this is we should implement `getOrCreateCollection` like we have in Python https://github.com/chroma-core/chroma/blob/64940789be360527e43f3ea819d91acdb46d8b72/chromadb/api/fastapi.py#L79

thoughts @Alw3ys ?

Alw3ys: @jeffchuber you are right. this will also solve what you mention @nfcampos which made sense somehow had to be visible if it failed. Now it will create the collection if it doesn't exist.

thanks you both!

langchain/src/vectorstores/chroma.ts
```diff
@@ -38,6 +38,18 @@ export class Chroma extends VectorStore {
     );
   }
 
+  async fromExistingCollection() {
```
nfcampos, line 41: This method name is confusing, given you're creating a static method of the same name. Can you rename to eg. `_ensureCollection` ?

Alw3ys: Renamed to `ensureCollection` as agreed over discord.

nfcampos (approved):
Thanks!

PR #389, opened 2023-03-20
Add ability to create child callback managers

agola11:
We need to get rid of the tracer's stack and, instead, explicitly assign the parent uuid for a specific run. This is because, in the middle of a run, we could spawn multiple concurrent child runs (this would be the case with multi-output agents, in each step, we could spawn multiple tool runs concurrently).

To avoid having to pass the parent run id to each call of the callback manager, we can wrap this with a child callback manager. The child callback manager would get a parentRunId from the parent callback manager and pass it through to the callbacks.

### DIFF
--- a/langchain/src/callbacks/base.ts
+++ b/langchain/src/callbacks/base.ts
@@ -170,7 +170,13 @@ export class CallbackManager extends BaseCallbackManager {
       this.handlers.map(async (handler) => {
         if (!handler.ignoreLLM && (verbose || handler.alwaysVerbose)) {
           try {
-            await handler.handleLLMStart?.(llm, prompts, runId, undefined, this._parentRunId);
+            await handler.handleLLMStart?.(
+              llm,
+              prompts,
+              runId,
+              undefined,
+              this._parentRunId
+            );
           } catch (err) {
             console.error(
               `Error in handler ${handler.constructor.name}, handleLLMStart: ${err}`
@@ -192,7 +198,12 @@ export class CallbackManager extends BaseCallbackManager {
       this.handlers.map(async (handler) => {
         if (!handler.ignoreLLM && (verbose || handler.alwaysVerbose)) {
           try {
-            await handler.handleLLMNewToken?.(token, runId, undefined, this._parentRunId);
+            await handler.handleLLMNewToken?.(
+              token,
+              runId,
+              undefined,
+              this._parentRunId
+            );
           } catch (err) {
             console.error(
               `Error in handler ${handler.constructor.name}, handleLLMNewToken: ${err}`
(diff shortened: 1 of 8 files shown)


langchain/src/callbacks/base.ts
```diff
@@ -408,6 +504,12 @@ export class CallbackManager extends BaseCallbackManager {
     this.handlers = handlers;
   }
 
+  getChild(): CallbackManager {
+    const manager = new CallbackManager(this._currentRunId);
```
nfcampos, line 508: We can actually remove need for any state tracking currentRunId here if we take it as an arg to getChild

nfcampos: The caller of getChild will always have its own runId around somewhere

langchain/src/callbacks/tracers.ts
```diff
@@ -98,43 +97,45 @@ export abstract class BaseTracer extends BaseCallbackHandler {
...
-          this.stack.at(-1)?.type === "tool" ||
-          this.stack.at(-1)?.type === "chain"
-        )
-      ) {
-        throw new Error("Nested run can only be logged for tool or chain");
+    if (run.parent_uuid) {
+      const parentRun = this.runMap.get(run.parent_uuid);
+      if (parentRun) {
+        if (!(parentRun.type === "tool" || parentRun.type === "chain")) {
+          throw new Error("Caller run can only be a tool or chain");
```
nfcampos, line 107: I know this was already here, but what is the intended result of throwing this error? Not tracking this run?

langchain/src/callbacks/tracers.ts
```diff
@@ -98,43 +97,45 @@ export abstract class BaseTracer extends BaseCallbackHandler {
...
 
-  protected async _endTrace(runId: string) {
-    const run = this.runMap.get(runId);
-    this.stack.pop();
-    if (this.stack.length === 0 && run) {
-      this.executionOrder = 1;
+  protected async _endTrace(run: LLMRun | ChainRun | ToolRun): Promise<void> {
+    if (!run.parent_uuid) {
       await this.persistRun(run);
+      this.executionOrder = 1;
```
nfcampos, line 121: Do we still need this? Sounds like it can instead be a calculated field?

langchain/src/callbacks/base.ts
agola11: I think it's better for getChild to not have any arguments, no?

langchain/src/callbacks/tracers.ts
agola11: Potentially not, but would rather not change this for now as it's something our backend uses atm

Issue #386, opened 2023-03-19
Zapier error are not handled.


GautierT:
Hi !

When Zapier output an error like this response : 

```
result :  {
  id: '01GVXS8S08MZ8RS8Y4F47Y31DM',
  action_used: 'Google Calendar: Create Detailed Event',
  input_params: {
    Calendar: 'yousphere@gmail.com',
    Start_Date___Time: 'tomorrow 11am',
    End_Date___Time: 'tomorrow 11:30am\nHelp: Date and time of when this event starts.'
  },
  rating_url: 'https://nla.zapier.com/execution/01GVXS8S08MZ8RS8Y4F47Y31DM/',
  result: null,
  status: 'error',
  error: 'Error during execution: The app returned "The specified time range is empty.".'
}
```

It's not handled by langchain and it's considered a success.

Used code for repro : 

```
export const run = async () => {
	const model = new OpenAI({ temperature: 0 })
	const zapier = new ZapierNLAWrapper()
	const toolkit = await ZapierToolKit.fromZapierNLAWrapper(zapier)

	const executor = await initializeAgentExecutor(toolkit.tools, model, 'zero-shot-react-description', true)
	console.log('Loaded agent.')

	const input = `create an event tomorrow from 11am to 11:30am to remind me to do my workout`

	console.log(`Executing with input "${input}"...`)

	const result = await executor.call({ input })

	console.log(`Got output ${result.output}`)
}

run()
```

Full log : 
```
❯ npx ts-node -r dotenv/config ./src/langchain.ts
Loaded agent.
Executing with input "create an event tomorrow from 11am to 11:30am to remind me to do my workout"...
Entering new agent_executor chain...
 I need to create an event in my calendar
Action: Google Calendar: Create Detailed Event
Action Input: create an event tomorrow from 11am to 11:30am
result :  {
  id: '01GVXV3PT15RNZ6BGA1MP1BFDW',
  action_used: 'Google Calendar: Create Detailed Event',
  input_params: {
    Calendar: 'yousphere@gmail.com',
    Start_Date___Time: 'tomorrow 11am',
    End_Date___Time: 'tomorrow 11:30am\nHelp: Date and time of when this event starts.'
  },
  rating_url: 'https://nla.zapier.com/execution/01GVXV3PT15RNZ6BGA1MP1BFDW/',
  result: null,
  status: 'error',
  error: 'Error during execution: The app returned "The specified time range is empty.".'
}
null
 I need to provide more information
[...]

PR #395, opened 2023-03-20
Update zapier.ts to handle errors in return JSON object

stunningpixels:
Hey @GautierT this should fix the error you described in #386 

Zapier errors are not always defined by a non 2XX status code, but instead Zapier can return an object with an error key.

Co-authored by [bloop](https://bloop.ai/).

### DIFF
--- a/langchain/src/agents/tools/zapier.ts
+++ b/langchain/src/agents/tools/zapier.ts
@@ -63,7 +63,16 @@ export class ZapierNLAWrapper {
         `Failed to execute action ${actionId} with instructions ${instructions}`
       );
     }
-    return resp.json();
+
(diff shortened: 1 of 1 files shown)


agola11:
Thanks! Can you provide an example of what this looks like? I would add an example that triggers an error in the zapier integration test (see `zapier_toolkit.int.test`)

agola11:
can take this over actually

conroywhitney:
@agola11 -- I'd refactored the Zapier int specs a little yesterday -- I can add some checks for the error if you want, or you can take over/cherry-pick that PR

https://github.com/hwchase17/langchainjs/pull/419

PR #396, opened 2023-03-20
Review treatment of max_tokens for chat models

nfcampos:


langchain/src/chat_models/openai.ts
```diff
@@ -209,6 +210,7 @@ export class ChatOpenAI extends BaseChatModel implements OpenAIInput {
       top_p: this.topP,
       frequency_penalty: this.frequencyPenalty,
       presence_penalty: this.presencePenalty,
+      max_tokens: this.maxTokens,
```
nfcampos, line 213: this was actually being ignored until now

PR #400, opened 2023-03-21
github loader (#368)

agola11:


langchain/src/document_loaders/github.ts
```diff
@@ -39,7 +65,7 @@ export class GithubRepoLoader extends BaseDocumentLoader {
     path: string;
   } {
     const match = url.match(
-      /https:\/\/github.com\/([^\/]+)\/([^\/]+)(\/tree\/[^\/]+\/(.+))?/i
+      /https:\/\/github.com\/([^/]+)\/([^/]+)(\/tree\/[^/]+\/(.+))?/i
```
nfcampos, line 68: Do we really need regex for this? Can't we just do a string split

langchain/src/document_loaders/github.ts
```diff
@@ -75,48 +101,52 @@ export class GithubRepoLoader extends BaseDocumentLoader {
...
   }
 
-  private async fetchFileContent(file: any): Promise<string> {
-    // Check if the file has a MIME type that starts with "image/" or "application/pdf"
+  private async fetchFileContent(file: GithubFile): Promise<string> {
+    const fileExtension = file.name.split(".").pop()?.toLowerCase();
     if (
-      file.mime &&
-      (file.mime.startsWith("image/") || file.mime === "application/pdf")
+      fileExtension === "jpg" ||
```
nfcampos, line 137: Hmm I don't like this, if someone has some other kind of file we didn't think of (eg. docx) then we'll just let it go through?

langchain/src/document_loaders/tests/github.int.test.ts
```diff
@@ -0,0 +1,11 @@
...
+import { GithubRepoLoader } from "../github.js";
+
+test("Test GithubRepoLoader", async () => {
+  const loader = new GithubRepoLoader(
+    "https://github.com/hwchase17/langchainjs",
+    { branch: "main", recursive: false, unknown: "warn" }
+  );
+  const documents = await loader.load();
+  console.log(documents[0].pageContent);
+}, 10000);
```
nfcampos, line 11: Btw we don't need this timeout on each test, I've set it here for all integration tests https://github.com/hwchase17/langchainjs/blob/810d8be60903ba6e2a7941811b179a8fbd1a3112/langchain/package.json#L64

langchain/src/document_loaders/github.ts
```diff
@@ -75,48 +101,52 @@ export class GithubRepoLoader extends BaseDocumentLoader {
...
     } catch (error) {
-      this.handleError(`Failed to process directory: ${path}`);
+      this.handleError(`Failed to process directory: ${path}, ${error}`);
     }
   }
 
-  private async fetchRepoFiles(path: string): Promise<any[]> {
+  private async fetchRepoFiles(path: string): Promise<GithubFile[]> {
     const url = `https://api.github.com/repos/${this.owner}/${this.repo}/contents/${path}?ref=${this.branch}`;
     const response = await fetch(url);
```
nfcampos, line 117: If we wanted to be really nice here we could do an exponential backoff retry, we have this to make it painless https://github.com/hwchase17/langchainjs/blob/main/langchain/src/util/async_caller.ts#L22

langchain/src/document_loaders/github.ts
nfcampos: We can eg use this instead https://github.com/sindresorhus/is-binary-path or the opposite https://github.com/sindresorhus/is-text-path

nfcampos (commented):
Some small changes to make, overall looks good

langchain/src/document_loaders/github.ts
mayooear: Yeah you're right, we also need to exclude lock files

mayooear:
> _ No description provided. _

@agola11 does this new PR override my original `github-loader` PR alongside commits and attribution?

agola11:
@mayooear -- No, once we squash and merge you'll get attributed too.

langchain/src/document_loaders/github.ts
agola11: I'll create a separate issue for us to use this in all document loaders / tool, everywhere use fetch basically

cktang88:
So this loads a repo by downloading each file one at a time???
Would it not have been more efficient to download the entire git repo at once to a temp dir?

PR #410, opened 2023-03-21
add getNumTokensFromMessages (Chat)

Marviel:
Adds convenient wrapper to help quickly calculate number of tokens for a list of messages.

This is useful when trying to determine the number of messages that should be fed to the agent.

Also returns each message's count, to help make decisions about what may need to be cut.


hwchase17 (commented):
this is actually a bit innacurate (based on https://github.com/openai/openai-cookbook/blob/main/examples/How_to_format_inputs_to_ChatGPT_models.ipynb, 4. Counting tokens section)

could we update to have that logic?

langchain/src/chat_models/openai.ts
```diff
@@ -355,6 +357,35 @@ export class ChatOpenAI extends BaseChatModel implements OpenAIInput {
...
+    if (["gpt-3.5-turbo", "gpt-3.5-turbo-0301"].includes(this.modelName)) {
+      tokensPerMessage = 4;
+      tokensPerName = -1;
+    } else if (["gpt-4", "gpt-4-0314"].includes(this.modelName)) {
+      tokensPerMessage = 3;
+      tokensPerName = 1;
+    }
+
+    const countPerMessage = messages.map((message) => {
+      const textCount = this.getNumTokens(message.text);
```
nfcampos, line 378: Note this method (from the base class) needs to be fixed for this to be accurate. It is using the wrong tokeniser. The solution is probably to convert it to use tiktoken. This doesn't have to be done in this PR, just raising it as an FYI

Marviel: gotcha, good catch.

I can try to flip it over today if it's already a dependency

nfcampos (approved):
Thanks! Have made a few updates to fix conflicts

PR #413, opened 2023-03-21
Fix: Pinecone upsert errors due to large size

mayooear:
@nfcampos as discussed, users are frequently complaining about upsert errors whilst using LangChain alongside Pinecone. The main culprit is that Pinecone has a max size for an upsert request at 2MB and a recommended upsert limit of 100 vectors per request as per the [docs](https://docs.pinecone.io/docs/limits#:~:text=Max%20size%20for%20an%20upsert,to%20queries%20immediately%20after%20upserting.)

The solution is simply to add logic that splits the vectors into chunks of approx. 50, before calling the upsert function for each chunk.

I've tested this locally and it works fine.

### DIFF
--- a/langchain/src/vectorstores/pinecone.ts
+++ b/langchain/src/vectorstores/pinecone.ts
@@ -47,20 +47,26 @@ export class PineconeStore extends VectorStore {
     ids?: string[]
   ): Promise<void> {
     const documentIds = ids == null ? documents.map(() => uuidv4()) : ids;
-
-    await this.pineconeIndex.upsert({
-      upsertRequest: {
-        vectors: vectors.map((values, idx) => ({
-          id: documentIds[idx],
-          metadata: {
-            ...documents[idx].metadata,
-            [this.textKey]: documents[idx].pageContent,
-          },
-          values,
-        })),
-        namespace: this.namespace,
+    const pineconeVectors = vectors.map((values, idx) => ({
+      id: documentIds[idx],
+      metadata: {
+        ...documents[idx].metadata,
+        [this.textKey]: documents[idx].pageContent,
       },
-    });
+      values,
+    }));
+
+    // Pinecone recommends a limit of 100 vectors per upsert request
+    const chunkSize = 50;
+    for (let i = 0; i < pineconeVectors.length; i += chunkSize) {
+      const chunk = pineconeVectors.slice(i, i + chunkSize);
+      await this.pineconeIndex.upsert({
+        upsertRequest: {
+          vectors: chunk,
+          namespace: this.namespace,
+        },
+      });
+    }
   }
 
   async similaritySearchVectorWithScore(


hwchase17 (approved):
nice, i like this - lets get this in!

nfcampos:
@mayooear we cannot get this in as is because it also includes the commits from #415 can you remove those from here? Also seems to include the commits from another PR that was already merged

mayooear:
> @mayooear we cannot get this in as is because it also includes the commits from #415 can you remove those from here? Also seems to include the commits from another PR that was already merged

strange. I wonder why, let me resend again

rawsh:
I'm still getting upsert errors on 0.0.39 with every doc I tried:

```
$ tsx -r dotenv/config testIndex.ts
PineconeClient: Error calling upsert: PineconeClient: Error calling upsertRaw: FetchError: The request failed and the interceptors did not return an alternative response
```

Here is my loader code, I can share an example doc that I'm getting upsert errors with if that is helpful.

```
import { PineconeClient } from "@pinecone-database/pinecone";
import * as dotenv from "dotenv";
import { PDFLoader } from "langchain/document_loaders";
import { OpenAIEmbeddings } from "langchain/embeddings";
import { PineconeStore } from "langchain/vectorstores";
import { RecursiveCharacterTextSplitter } from "langchain/text_splitter";

dotenv.config();

async function main() {
    const client = new PineconeClient();
    await client.init({
      apiKey: process.env.PINECONE_API_KEY || "",
      environment: process.env.PINECONE_ENVIRONMENT || "",
    });
    const pineconeIndex = client.Index(process.env.PINECONE_INDEX || "");
    const loader = new PDFLoader("example_data/lecture20.pdf");
    const rawDocs = await loader.load();
        chunkSize: 1000,
        chunkOverlap: 200,
    });
    const docs = await textSplitter.splitDocuments(rawDocs);

    try {
        await PineconeStore.fromDocuments(docs, new OpenAIEmbeddings(), {
            pineconeIndex,
        });
    } catch (err) {
        console.error(err);
    }
}

main()
```

rawsh:
After trying a few more things it looks like the metadata langchain is including with the documents is over the Pinecone limits.

rmbrntt:
> After trying a few more things it looks like the metadata langchain is including with the documents is over the Pinecone limits.

If your pinecode db dependency is on `0.0.11`, try downgrading to `"@pinecone-database/pinecone": "^0.0.10"`

Solved it for me using the doc examples 
image

The only way I figured that out was due to pinecone's own chatbot example using that version - https://github.com/pinecone-io/chatbot-demo/blob/main/package.json#L15

Also, the pinecone dimensions had to be 1536 on my index.

boboweiqi:
> > 在尝试了更多操作之后，文档中包含的元数据 langchain 似乎超出了 Pinecone 限制。
> 
> 如果您的 pinecode db 依赖项是 on `0.0.11`，请尝试降级到`"@pinecone-database/pinecone": "^0.0.10"`
> 
> 使用文档示例为我解决了它 图像
> 
> 我想出来的唯一方法是使用该版本的 pinecone 自己的聊天机器人示例 - https://github.com/pinecone-io/chatbot-demo/blob/main/package.json#L15
> 
> 此外，我的索引中的松果尺寸必须为 1536。

According to the modification, the same error still occurs

omikader:
Hi @mayooear! Why do we process the upserts sequentially vs in parallel? In the Pinecone Python docs, there is an [example](https://docs.pinecone.io/docs/insert-data#sending-upserts-in-parallel) of using a threadpool to send 30 concurrent upsert requests (each 100 vectors in size).

I believe we could do something similar here by sending up to 30 chunks at a time using `Promise.all` which could significantly improve the client throughput without adding extra significant extra load to the target index.

PR #414, opened 2023-03-21
Fix: Missing 'await' keyword is most examples using the format function

mayooear:
Most examples and documentation using the format function i.e. `prompt.format` fail to return the promise because the `await` keyword is missing.

These changes fix these bugs across the repo. Example: `await prompt.format`


hwchase17 (approved):
thanks!

PR #416, opened 2023-03-21
docs: update Chroma usage docs example

Alw3ys:
- Updated Chroma example to use native langchainjs query func
- Added example usage for query docs from existing collection on Chroma

### DIFF
--- a/docs/docs/modules/indexes/vector_stores/chroma.md
+++ b/docs/docs/modules/indexes/vector_stores/chroma.md
@@ -19,6 +19,28 @@ npm install -S chromadb
 import { Chroma } from "langchain/vectorstores";
 import { OpenAIEmbeddings } from "langchain/embeddings";
 
+// text sample from Godel, Escher, Bach
+const vectorStore = await Chroma.fromTexts(
+  [
+    "Tortoise: Labyrinth? Labyrinth? Could it Are we in the notorious Little\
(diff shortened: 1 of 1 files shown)


hwchase17 (commented):
i actually think we want to keep this section focused on chroma, irregardless of how its used in downstream chains. so im all for adding mroe examples of chroma.from_documents, etc, but i dont think we should use in chains (theres other sections of the docs for that)

Alw3ys:
> i actually think we want to keep this section focused on chroma, irregardless of how its used in downstream chains. so im all for adding moe examples of chroma.from_documents, etc, but i dont think we should use in chains (theres other sections of the docs for that)

For sure, updated it! @hwchase17

docs/docs/modules/indexes/vector_stores/chroma.md
```diff
@@ -13,32 +13,31 @@ Use [chroma](https://github.com/chroma-core/chroma) with langchainjs.
...
 ```
 
-## Usage
+## Index and query docs
 
 ```typescript
 import { Chroma } from "langchain/vectorstores";
 import { OpenAIEmbeddings } from "langchain/embeddings";
 
-// text sample from Godel, Escher, Bach
```
nfcampos, line 22: Can we just keep both? These are two valid ways of using Chroma VectorStore

hwchase17 (approved):
lgtm! thanks!

PR #424, opened 2023-03-22
Pinecone metadata filtering

conroywhitney:
I noticed this open issue around filtering Pinecone by metadata: https://github.com/hwchase17/langchainjs/issues/333

Since the Pinecone NodeJS lib supports a `filter` param: https://docs.pinecone.io/docs/metadata-filtering

I added an optional `filter` param to the `VectorStore` methods `similaritySearch`, `similaritySearchVectorWithScore`, and `similaritySearchWithScore`. Then implemented it in `PineconeStore`.

I updated the specs, docs, and examples to reflect this new functionality.

One thing I couldn't figure out, though, is how to allow metadata filtering within a chain. It'd be cool to be able to add it as a param in `VectorDBQAChain.fromLLM`. What do y'all think? Doable?


hwchase17 (commented):
thank you for adding this! lots of people have been asking for filtering pinecone, so would love to get this in

re: using it in a chain - we'll be introducing a new abstraction shortly that should make this easier, so i wouldnt worry about it for now

ElektrikSpark:
Funny how we were working on this at the exact same time. #426 

My approach was to add `filter` the same way `namespace` was implemented. I want to get metadata filtering in ASAP, so I would love to talk about the best approach to do so.

This is the way I plan on implementing filter:
```
// metadata filtering
      const filter = {
        metadataId: { $in: metadataIds },
      };

      const index = pinecone.Index(PINECONE_INDEX_NAME);
      /* create vectorstore*/
      const vectorStore = await PineconeStore.fromExistingIndex(
        new OpenAIEmbeddings({ modelName: 'text-embedding-ada-002' }),
        {
          namespace: PINECONE_NAME_SPACE,
          pineconeIndex: index,
          textKey: 'text',
          filter: filter,
        }
      );

      const model = openai;
      // create the chain
      const chain = VectorDBQAChain.fromLLM(model, vectorStore);

      //Ask a question
      const response = await chain.call({
        query: sanitizedQuestion,
      });
```

conroywhitney:
@hwchase17 @ElektrikSpark -- I'm open to either approach. I can move the `filter` param into `PineconeLibArgs` if you'd prefer. Let me know what y'all think?

|Approach|Pros|Cons|
|-|-|-|
|`PineconeLibArgs`| Pinecone-specific | Have to re-init VS to change filters for query |
|`similaritySearch`| Filters associated with query | Not all VS have meta filtering |

hwchase17:
i think it should definitley be an option to pass it in during init. the reason i think that is important is that its annoying to plumb the search kwargs through all the time when youre using it in a deep nested way in a chain

but im also happy to allow it in kwargs, and just do some validation that only one exists, etc

hwchase17:
@conroywhitney i merged @ElektrikSpark s pr in. im gonna take a stab at fixing merge conflicts and get this one in as well

PR #427, opened 2023-03-22
Adds AnthropicChat LLM class for interfacing with Claude

jacoblee93:
This PR adds support for Anthropic's recently released chatbot, Claude, as a LangChain LLM by creating a wrapper `AnthropicChat` class around Anthropic's TypeScript SDK.

I based the new class of the existing `OpenAIChat` class, including adding support for users to pass in OpenAI style `prefixMessages` (`ChatCompletionRequestMessage` from their SDK). The new class will translate those messages into Anthropic's very particular prompting syntax, allowing for a relatively similar user-facing API.

This PR adds a dependency on Anthropic's official `@anthropic-ai/sdk`, and requires an Anthropic API key set as an environment variable named `ANTHROPIC_API_KEY` environment variable for the new test cases to pass, so it may require an update to the CI/CD runner.


conroywhitney:
I don't have Anthropic yet to test this out 😢 but I'm looking forward to speaking with Claude!

Since you added a new `.env` var, would you mind updating `langchain/.env.example` and `examples/.env.example`?

I noticed the two actually differed slightly, so here's a combined list (including the new `ANTHROPIC_API_KEY`):

```
ANTHROPIC_API_KEY=ADD_YOURS_HERE
COHERE_API_KEY=ADD_YOURS_HERE
HUGGINGFACEHUB_API_KEY=ADD_YOURS_HERE
OPENAI_API_KEY=ADD_YOURS_HERE
PINECONE_API_KEY=ADD_YOURS_HERE
PINECONE_ENVIRONMENT=ADD_YOURS_HERE
PINECONE_INDEX=ADD_YOURS_HERE
SERPAPI_API_KEY=ADD_YOURS_HERE
SERPER_API_KEY=ADD_YOURS_HERE
SUPABASE_PRIVATE_KEY=ADD_YOURS_HERE
SUPABASE_URL=ADD_YOURS_HERE
```

nfcampos:
Thanks for this. I think ideally this would be a subclass of ChatModel, not LLM

nfcampos:
@jacoblee93 thanks! You can ignore that combine method, it is optional, and as you say doesn't apply to this one

langchain/src/chat_models/anthropic.ts
```diff
@@ -0,0 +1,302 @@
...
+          },
+        });
+      };
+    } else {
+      if (!this.batchClient) {
+        this.batchClient = new AnthropicApi(this.apiKey);
+      }
+      makeCompletionRequest = async () => this.batchClient.complete(request);
+    }
+    return backOff(makeCompletionRequest, {
```
nfcampos, line 288: Replace this with `AsyncCaller`

jacoblee93: Swapped it out!

jacoblee93:
PTAL @nfcampos @agola11

jacoblee93:
Ah whoops, I merged my personal `main` branch and not the upstream one - will fix now.

jacoblee93:
Alright, merged the correct `main` branch this time!

langchain/src/chat_models/tests/chatanthropic.int.test.ts
```diff
@@ -0,0 +1,106 @@
...
+  const responseA = await chat.generatePrompt([
+    await chatPrompt.formatPromptValue({
+      input_language: "English",
+      output_language: "French",
+      text: "I love programming.",
+    }),
+  ]);
+
+  console.log(responseA.generations);
+}, 50000);
```
agola11, line 106: This is not needed, as I believe @nfcampos added a limit for all int tests

agola11 (approved):
This lgtm, just fix CI

langchain/src/chat_models/tests/chatanthropic.int.test.ts
jacoblee93: Ok, just removed the timeouts.

jacoblee93:
I think the last failing test is some approvals check? This shouldn't touch any docs pages.

PR #430, opened 2023-03-22
Fix process exit on `new OpenAI({concurrency: 1}).call('')`, Add AsyncCaller to BaseLanguageModel, Make all existing LLMs and Chat models use async caller to control concurrency and retries

### LINKED ISSUE #418: Bug: Setting `concurrency=1` in OpenAI model params immediately crashes process with no error message on first call
This one was quite the bugger to find. Because of GPT4's rate limit, I set the `concurrency` param to `1` when instantiating the `OpenAI` model as you can see below:

```
const model = new OpenAI({
  openAIApiKey: <redacted>,
  modelName: 'gpt-4',
  temperature: 0.7,
  verbose: true,
  concurrency: 1,
  callbackManager: callbackManager,
  maxTokens: 1000,
  maxRetries: 5
})
```

Even with explicit callbacks, this causes the program to exit *immediately* upon making its first call. Values higher than one (including `2`) work fine.

nfcampos:
Closes #418

### DIFF
--- a/examples/src/emb_openai_concurrency.ts
+++ b/examples/src/emb_openai_concurrency.ts
@@ -1,0 +1,12 @@
+import { OpenAIEmbeddings } from "langchain/embeddings";
+
+export const run = async () => {
+  const model = new OpenAIEmbeddings({
+    maxConcurrency: 1,
+  });
+  console.log("yooooo");
+  const res = await model.embedQuery(
+    "What would be a good company name a company that makes colorful socks?"
+  );
+  console.log("hello", { res });
+};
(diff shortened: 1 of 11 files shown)


hwchase17 (approved):
i think seems reasonable to me

examples/src/emb_openai_concurrency.ts
```diff
@@ -0,0 +1,12 @@
+import { OpenAIEmbeddings } from "langchain/embeddings";
+
+export const run = async () => {
+  const model = new OpenAIEmbeddings({
+    maxConcurrency: 1,
+  });
+  console.log("yooooo");
```
agola11, line 7: did you mean to leave this in lol?

nfcampos: Oops, at least it's an example haha

PR #438, opened 2023-03-22
Add dynamoDB store

zammitjames:


hwchase17 (commented):
awesome!

an example (and corresponding documenation page) would be really helpful

langchain/src/vectorstores/dynamodb.ts
```diff
@@ -0,0 +1,157 @@
...
+    query: number[],
+    k: number,
+    filter?: object
+  ): Promise<[Document, number][]> {
+    // Fetch all items from the DynamoDB table
+    const scanOutput = await this.dynamoDb
+      .scan({
+        TableName: this.tableName,
+      })
+      .promise();
```
danamajid, line 128: Explored doing a `query` here instead of a `scan`? You'll probably want to store different docs in different PKs and query all the docs for that partition separately instead of the entire table.

langchain/src/vectorstores/dynamodb.ts
```diff
@@ -0,0 +1,157 @@
...
+// eslint-disable-next-line @typescript-eslint/ban-types, @typescript-eslint/no-explicit-any
+type DynamoMetadata = Record<string, any>;
+
+export interface DynamoLibArgs {
+  tableName: string;
+  dynamoDb: DynamoDB.DocumentClient;
+  textKey?: string;
+}
+
+export class DynamoStore extends VectorStore {
```
danamajid, line 18: Would suggest renaming this to `DynamoDBStore` to match the service's name. Same thing for every other exported interface.

danamajid (commented):
Nice addition! Curious if you ended up using this. Or did you end up going down another memory store?

+1 on @hwchase17's comment for adding usage instructions (docs/example/tests).

PR #440, opened 2023-03-23
Add implementation of Prisma vectorstore

dqbd:
Add an implementation of Prisma vectorstore, with type safety in mind. 

__1. Prisma vectorstore implements a mock interface, as it depends on the types generated via `prisma generate`__
This does mean that the code might be unnecessary brittle compared to hardcoding `@prisma/client` directly in user codebases. 

__2. Added optional generic type argument for Document metadata__
Developers can now specify a concrete type for `metadata` prop in `Document`. AFAIK this does not propagate in chains etc, but I believe this to be an useful starting block. 
 
__3. `fromDocuments` and `fromTexts` is currently unsupported__
As there is a significant difficulty upserting entire documents, I've marked these methods as `@deprecated` for now.


hwchase17 (commented):
oh nice i like this a lot

an example would probably hel[

examples/src/indexes/vector_stores/prisma_vectorstore/.env.example
```diff
@@ -0,0 +1,2 @@
+# Add DATABASE_URL to .env file in this directory
+DATABASE_URL=postgresql://[USERNAME]:[PASSWORD]@[ADDR]/[DBNAME]
```
nfcampos, line 2: Can we put this in the main `examples/` .env file ?

nfcampos: Ah I see this example needs some more things, makes sense to leave it here

langchain/src/vectorstores/prisma.ts
```diff
@@ -0,0 +1,249 @@
...
+    return { create };
+  }
+
+  async addModels(models: TModel[]) {
+    return this.addDocuments(
+      models.map((metadata) => {
+        const pageContent = typeof metadata[this.contentColumn];
+        if (pageContent !== "string")
+          throw new Error("Content column must be a string");
+        return new Document({ pageContent, metadata });
```
nfcampos, line 151: I think we'd want to remove the pageContent from `metadata`?

langchain/src/vectorstores/prisma.ts
```diff
@@ -0,0 +1,249 @@
...
+
+  async similaritySearchVectorWithScore(
+    query: number[],
+    k: number
+  ): Promise<[Document<SimilarityModel<TModel, TSelectModel>>, number][]> {
+    const vectorQuery = `[${query.join(",")}]`;
+    const articles = await this.db.$queryRaw<
+      Array<SimilarityModel<TModel, TSelectModel>>
+    >`
+      SELECT ${this.selectSql}, ${this.vectorColumnSql} <=> ${vectorQuery}::vector as "_distance"
```
nfcampos, line 198: will prisma escape these interpolated values? otherwise could be potentially dangerous

langchain/src/vectorstores/prisma.ts
```diff
@@ -0,0 +1,249 @@
...
+      }
+    }
+
+    return results;
+  }
+
+  /**
+   * @deprecated Not implemented in Prisma vectorstore, please use PrismaVectorStore.withModel instead
+   */
+  static fromTexts(
```
nfcampos, line 224: These are the main methods we see people using, worth implementing I think. Given you implemented `addDocuments` doing these should be trivial, I'm happy to do them too

nfcampos (commented):
A few comments, looks good otherwise

MaximeThoonsen:
hello @dqbd, I was wondering if we could use typeorm to avoid adding another orm in the dependencies?

dqbd:
Hi @MaximeThoonsen!

> I was wondering if we could use typeorm to avoid adding another orm in the dependencies?

Both `prisma` and `@prisma/client` were added as a peer-dependency, it is not being actively installed when consuming `langchain` project. That being said, agreed, it can be removed entirely from the `langchain` project, as I'm stubbing the definitions.

Regardless, I think a TypeORM vectorstore implementation is certainly doable. Will look into it and possibly create a separate PR.

langchain/src/vectorstores/prisma.ts
dqbd: I believe that should be the case for `vectorQuery`, as the query has been built using `$queryRaw` instead of `$queryRawUnsafe` and will use parametrised queries: https://github.com/prisma/prisma-client-js/issues/727#issuecomment-650096790

However, `this.selectSql` (coming from `dbConfig.columns`), `this.vectorColumnSql` (coming from `dbConfig.vectorColumnName`) and `this.tableSql` (coming from `dbConfig.tableName`). 

Assuming these values come strictly from the developer and _not_ from user input, I believe it would be acceptable to leave it as-is and add a disclaimer to shift the responsibility of sanitising to the consumer. What do you think @nfcampos ?

langchain/src/vectorstores/prisma.ts
dqbd: At the moment, it returns the subset a Prisma model (`TModel` and `TSelectModel`), which do include the `this.contentColumn`). 

We could remove the content column from the metadata, but I would argue that the metadata should contain the same columns, as user specifies in the `dbConfig`.

langchain/src/vectorstores/prisma.ts
nfcampos: Okay great!

nfcampos (approved):
This looks great, thanks!

Issue #441, opened 2023-03-23
Pinecone metadata filtering - "cannot provide both `filter` and `this.filter`"


ElektrikSpark:
I am going to work on a pull request to fix this issue. An empty object is being passed in as filter by default on `similaritySearchWithScore`, so I am not able to pass filter into `fromExistingIndex()`.

Here is where the problem is in the langchain source:

`pinecone.ts`
```
async similaritySearchVectorWithScore(
    query: number[],
    k: number,
    filter?: object
  ): Promise<[Document, number][]> {
    if (filter && this.filter) {
      throw new Error("cannot provide both `filter` and `this.filter`");
    }
    const _filter = filter ?? this.filter;
    const results = await this.pineconeIndex.query({
      queryRequest: {
        includeMetadata: true,
        namespace: this.namespace,
        topK: k,
        vector: query,
        filter: _filter,
      },
    });
```

`vectorstores/base.ts`
```
async similaritySearch(
    query: string,
    k = 4,
    filter: object = {}
  ): Promise<Document[]> {
    const results = await this.similaritySearchVectorWithScore(
      await this.embeddings.embedQuery(query),
      k,
      filter
    );

    return results.map((result) => result[0]);
  }

  async similaritySearchWithScore(
    query: string,
    k = 4,
    filter: object = {}
  ): Promise<[object, number][]> {
    return this.similaritySearchVectorWithScore(
      await this.embeddings.embedQuery(query),
      k,
      filter
    );
  }
```

nfcampos:
Thanks for fixing this @ElektrikSpark !

PR #442, opened 2023-03-23
fixed filter to check for empty object and undefined

ElektrikSpark:
Issue #441 

Right now on similaritySearch an empty object is being passed in by default `filter: object = {}`. I opted to expand the validation to check for an empty object as well as undefined and keep the functionality the same.

We could change similaritySearchVectorWithScore from `filter?: object` to `filter: object = {}` to mimic `similaritySearch` and `similaritySearchWithScore` and remove the undefined checks.


nfcampos:
@ElektrikSpark thanks. Yea my preference would be just remove the default value in the base class method, as that would eg allow people to set a default filter when instantiating (this.filter) and then disable it by passing an empty object on a call to similaritySearchVectorWithScore

ElektrikSpark:
@nfcampos Sounds good. I removed the default value for filter, and to get around the `default-param-last` linting rule, and not break all the similaritySearch calls, I got rid of the default value of 4 for k on `similaritySearch`. We were setting a default value of k = 4 at the class level.

nfcampos:
Or we can give the filter a default value of undefined? @ElektrikSpark

ElektrikSpark:
Changed it to:
```
k = 4,
filter: object | undefined = undefined
```

PR #444, opened 2023-03-23
Adding Pinecone batch upsert capability

danmo:
Pinecone upsert has a hard cap of 1000 vectors per upsert request. Even more, their documentation advices on capping the vector size per upsert request to 100. 

To prevent pinecone upsert exceptions when upserting large embeddings, I've added the 1000 vector size default batch cap and the ability to configure batch size to lower values via constructor params.


langchain/src/vectorstores/pinecone.ts
```diff
@@ -49,22 +53,24 @@ export class PineconeStore extends VectorStore {
...
-      id: documentIds[idx],
-      metadata: {
-        ...documents[idx].metadata,
-        [this.textKey]: documents[idx].pageContent,
-      },
-      values,
-    }));
 
-    // Pinecone recommends a limit of 100 vectors per upsert request
-    const chunkSize = 50;
```
nfcampos, line 62: Thanks for this. We already have this functionality (see the diff around this line). Can you clarify what this improves @danmo

PR #450, opened 2023-03-23
feat(vectorstores): allow single metadata object to span across all text

skwak22:
This allows a single metadata object to be applied to an entire collection of text, for when we want to insert text from a single corpus into a vectorstore, or similar.


nfcampos:
Note to me: update the docs

PR #462, opened 2023-03-25
Fix ChatConversationalAgent Memory Bug

GRVYDEV:
The ChatConversationalAgent was not properly saving AI output into the message history since it was being passed the raw output rather than the parsed output from the executor. This PR fixes that issue but adding parsed output value in the `prepareForOutput` call. It also adds a fix that prevents undefined messages from being added into the history on ChatHistory.

### DIFF
--- a/langchain/src/agents/chat_convo/index.ts
+++ b/langchain/src/agents/chat_convo/index.ts
@@ -22,6 +22,7 @@ import {
   BaseOutputParser,
 } from "../../schema/index.js";
 import { AgentInput } from "../types.js";
+import { AgentFinish } from "../../schema/index.js";
 import { Tool } from "../tools/base.js";
 
 export class ChatConversationalAgentOutputParser extends BaseOutputParser {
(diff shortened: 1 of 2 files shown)


nfcampos:
@hwchase17 one for you I think

hwchase17:
not sure what the bug is, as far as I can tell this is working as I expect

do you have any examples for how to replicate and what the expected input/output are?

GRVYDEV:
> not sure what the bug is, as far as I can tell this is working as I expect
> 
> do you have any examples for how to replicate and what the expected input/output are?

Yes. When using the ChatConversationalAgent with the AgentExecutor the 'Final Answer' output from the LLM was not properly being saved into the chat history. This is because when `memory.saveContext` was being called the output passed in did not have the proper key to save the memory because it was happening before the AgentExecutor parsed and formatted the output. This led to the AiChatMessages in the memory containing text that was `undefined`.

GRVYDEV:
> Hi @GRVYDEV looking at your diff here, I think maybe the issue is that you're passing the memory to the LLMChain inside the agent, rather than to the executor as we expect, see this example for how we expect to do it https://js.langchain.com/docs/modules/agents/agents/examples/conversational_agent Can you give that a try?

I think you may be right. My apologies. I'll close this PR

PR #464, opened 2023-03-25
Fixed: use the name_for_model and description_for_model for AIPluginTool

VoidIsVoid:
Using name_for_model, description_for_model in AIPluginTool  makes more sense.

### DIFF
--- a/langchain/src/agents/tools/aiplugin.ts
+++ b/langchain/src/agents/tools/aiplugin.ts
@@ -51,7 +51,7 @@ export class AIPluginTool extends Tool implements AIPluginToolParams {
 
     return new AIPluginTool({
       name: aiPluginJson.name_for_model,
-      description: `Call this tool to get the OpenAPI spec (and usage guide) for interacting with the ${aiPluginJson.name_for_human} API. You should only call this ONCE! What is the ${aiPluginJson.name_for_human} API useful for? ${aiPluginJson.description_for_human}`,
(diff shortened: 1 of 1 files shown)


langchain/src/agents/tools/aiplugin.ts
```diff
@@ -51,7 +51,7 @@ export class AIPluginTool extends Tool implements AIPluginToolParams {
 
     return new AIPluginTool({
       name: aiPluginJson.name_for_model,
-      description: `Call this tool to get the OpenAPI spec (and usage guide) for interacting with the ${aiPluginJson.name_for_human} API. You should only call this ONCE! What is the ${aiPluginJson.name_for_human} API useful for? ${aiPluginJson.description_for_human}`,
```
nfcampos, line 54: From our testing it actually produces better results as it is currently

nfcampos: Did you have better results with your changes?

VoidIsVoid: I think it's more reasonable to use xx_for_model in prompt rather than xx_for_human.
If we use xx_for_human in prompt, so what does xx_for_model mean for ?

nfcampos: Notice they're used in the other properties, `name` and `apiSpec`. This description is meant to be a human-readable description of what this tool is for, not a description of how to use it (that's `apiSpec`)

PR #490, opened 2023-03-27
feat: print top level object describe

jacobrosenthal:
Zod allowed me to make the .describe on a top level object, but we didnt print it. 

Seems like we should? But I honestly dont incredibly need this right now, did what I wanted by templating around the schema anyway


nfcampos:
Great, thanks @jacobrosenthal

PR #492, opened 2023-03-28
fix:update retrieval docs with correct chain name

toonverbeek:
The retrieval docs were still mentioning the `chatVectorDBQAChain` in some places. I've updated and replaced those with `ConversationalRetrievalQAChain`.


nfcampos:
Thanks @toonverbeek

PR #496, opened 2023-03-28
Make docker compose work in m1 mac

nfcampos:
@agola11 thoughts on this? Only one of my Macs needed this, no clue why


agola11:
I remember running into a similar issue, where running "docker compose" instead of "docker-compose" fixed it. This looks fine though, I will investigate the docker issues in more detail when we revive the langchainplus work

agola11 (approved):

PR #506, opened 2023-03-28
feat(MapReduce): run at least one the map chain

louis-sanna-eki:
Solves https://github.com/hwchase17/langchainjs/issues/347

# Changes

In the MapReduce chain, the map chain runs at least one even if the maxToken condition is already met.

# Remark

This align with the python class: https://github.com/hwchase17/langchain/blob/master/langchain/chains/combine_documents/map_reduce.py.


langchain/src/chains/combine_docs_chain.ts
```diff
@@ -154,31 +156,40 @@ export class MapReduceDocumentsChain
 
     let currentDocs = docs as Document[];
 
-    for (let i = 0; i < this.maxIterations; i += 1) {
```
nfcampos, line 157: Would probably be easier to just do `this.maxIterations = Math.min(fields.maxIterations ?? this.maxIterations, SANITY_CHECK_MAX_ITERATION_COUNT)` in the constructor?

langchain/src/chains/combine_docs_chain.ts
```diff
@@ -154,31 +156,40 @@ export class MapReduceDocumentsChain
...
+      if (iterationCount >= this.maxIterations) {
+        break;
+      }
+
       const inputs = currentDocs.map((d) => ({
         [this.documentVariableName]: d.pageContent,
         ...rest,
       }));
-      const promises = inputs.map(async (i) => {
+      const results = await this.llmChain.apply(inputs);
```
nfcampos, line 169: I think we'd want to have the new behaviour behind a flag? Some people might prefer to save some money and not run the map chain if not needed

nfcampos (commented):
Some comments

langchain/src/chains/combine_docs_chain.ts
louis-sanna-eki: @nfcampos ok I'll do another implem with a flag.

PR #508, opened 2023-03-28
Allow structured parser to work with array inputs, numbers

iamnafets:
This change allows us to work with schemas where the base object is an array and which contain number fields.


iamnafets:
@jacobrosenthal you've got experience with this code if you want to take a look.

PR #515, opened 2023-03-29
Nc/replicate llm fixes

nfcampos:


hwchase17 (approved):

langchain/src/llms/tests/replicate.int.test.ts
```diff
@@ -0,0 +1,16 @@
+import { test, expect } from "@jest/globals";
+import { Replicate } from "../replicate.js";
+
+test("Test Cohere", async () => {
```
cbh123, line 4:
```suggestion
test("Test Replicate", async () => {
```

langchain/src/llms/replicate.ts
```diff
@@ -0,0 +1,75 @@
...
+      userAgent: "langchain",
+      auth: this.apiKey,
+    });
+
+    const output = await this.caller.call(() =>
+      replicate.run(this.model, {
+        wait: true,
+        input: {
+          ...this.input,
+          prompt,
```
cbh123, line 53: Technically not all models will have a `prompt` input. One way to fix this is a block above that looks like this:

```
// get model version
const  model_owner = model.split("/")[0]
const model_name, const model_version = model.split("/")[0].split(":")

// get the model openapi schema
const { schema } = await replicate.models.versions.get(
  model_owner,
  model_name,
  version_id
);

// 
const inputProperties = Object.entries(schema.properties).sort(
    ([, a], [, b]) => a["x-order"] - b["x-order"]
  );

const firstInputName = firstInputName[0][0]

this.input = {...this.input, firstInputName: prompt}
```

I haven't actually tested this, but that should match the logic we're [running in the python version](https://github.com/cbh123/langchain/blob/30862c1fe0cfc0e6ca5818049706c5d83052ec77/langchain/llms/replicate.py#LL90:LL107)

PR #516, opened 2023-03-29
Make verbose flag default to true when a callbackManager is passed in

### LINKED ISSUE #507: Default verbose to true when passing in callbackManager
This isn't an ideal api but that's a different problem

nfcampos:
Closes #507

### DIFF
--- a/langchain/src/callbacks/base.ts
+++ b/langchain/src/callbacks/base.ts
@@ -350,27 +350,35 @@ export class CallbackManager extends BaseCallbackManager {
 }
 
 export class ConsoleCallbackHandler extends BaseCallbackHandler {
+  async handleLLMStart(llm: { name: string }, prompts: string[]) {
+    console.log(`Starting LLM ${llm.name} with prompts: ${prompts.join(", ")}`);
+  }
+
+  async handleLLMEnd(output: LLMResult) {
+    console.log("Finished LLM with output:", output);
+  }
+
   async handleChainStart(chain: { name: string }) {
     console.log(`Entering new ${chain.name} chain...`);
   }
 
-  async handleChainEnd(_output: ChainValues) {
-    console.log("Finished chain.");
+  async handleChainEnd(output: ChainValues) {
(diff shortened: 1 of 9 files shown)


langchain/src/callbacks/base.ts
```diff
@@ -350,27 +350,35 @@ export class CallbackManager extends BaseCallbackManager {
 }
 
 export class ConsoleCallbackHandler extends BaseCallbackHandler {
+  async handleLLMStart(llm: { name: string }, prompts: string[]) {
```
hwchase17, line 353: not sure if intentional or was just for testing, but flagging that this changed

hwchase17 (approved):

langchain/src/callbacks/base.ts
nfcampos: Yea that was on purpose, seemed a bit strange that our console callback handler only implements some methods? But can revert if that's on purpose for some reason

agola11: I think we'll want to keep as is otherwise there will be too much output during agent runs

agola11 (commented):
Looks good, I would just revert the console callback handler to keep the agent output as is

langchain/src/callbacks/base.ts
nfcampos: Alright, undid these changes

langchain/src/chains/llm_chain.ts
```diff
@@ -26,7 +26,7 @@ export interface LLMChainInput extends ChainInputs {
   outputParser?: BaseOutputParser;
 
   /** @ignore */
-  outputKey: string;
+  outputKey?: string;
```
agola11, line 29: why did this change?

agola11 (approved):

langchain/src/chains/llm_chain.ts
nfcampos: Because I'm now using this type in the constructor, and it's optional there

PR #525, opened 2023-03-29
feat(MapReduceDocumentsChain): add ensureMapStep flag

louis-sanna-eki:
resolves https://github.com/hwchase17/langchainjs/issues/347

# Changes

In the MapReduce chain, the map chain runs at least once if the ensureMapStep flag is set to true.

# Remark

This align with the python class: https://github.com/hwchase17/langchain/blob/master/langchain/chains/combine_documents/map_reduce.py.


louis-sanna-eki:
@nfcampos here is an implementation with flag.

nfcampos (approved):

PR #526, opened 2023-03-29
Enable streaming OpenAI requests to work outside of Node

nfcampos:
Streaming requests now use our axios fetch adapter

### DIFF
--- a/langchain/src/chat_models/openai.ts
+++ b/langchain/src/chat_models/openai.ts
@@ -271,7 +271,7 @@ export class ChatOpenAI extends BaseChatModel implements OpenAIInput {
             {
               responseType: "stream",
               onmessage: (event) => {
-                if (event.data === "[DONE]") {
+                if (event.data?.trim?.() === "[DONE]") {
                   resolve(response);
                 } else {
                   const message = JSON.parse(event.data) as {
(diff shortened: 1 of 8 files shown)


langchain/src/util/axios-fetch-adapter.d.ts
```diff
@@ -2,3 +2,14 @@
...
+
+export interface StreamingAxiosConfiguration {
+  responseType: "stream";
+
+  /**
+   * Called when a message is received. NOTE: Unlike the default browser
+   * EventSource.onmessage, this callback is called for _all_ events,
+   * even ones with a custom `event` field.
+   */
+  onmessage?: (ev: EventSourceMessage) => void;
```
agola11, line 14: nit: `onMessage` ?

langchain/src/chat_models/openai.ts
```diff
@@ -258,68 +255,94 @@ export class ChatOpenAI extends BaseChatModel implements OpenAIInput {
...
+              ...params,
+              messages: messages.map((message) => ({
+                role: messageTypeToOpenAIRole(message._getType()),
+                content: message.text,
+              })),
+            },
+            {
+              responseType: "stream",
+              onmessage: (event) => {
+                if (event.data === "[DONE]") {
```
agola11, line 274: could there ever be extra newlines or carriage returns here?

agola11 (approved):
lgtm, thanks!

langchain/src/chat_models/openai.ts
nfcampos: yea I guess it doesn't hurt to call .trim, will add

PR #533, opened 2023-03-29
Enable passing timeout to OpenAI models

### LINKED ISSUE #511: Add ability to add timeOut to when initialising OpenAI Clients
Description:

Currently, when initializing a new ChatOpenAI instance, there is no option to pass a custom timeout value to configure the underlying Axios instances. This makes it difficult to handle cases where the server never responds (which we've noticed at times recently with OpenAI's API).

To improve the flexibility of the library, it would be helpful to allow users to set a custom timeout value when initialising new  instance of classes that use the OpenAI API.

nfcampos:
Closes #511


jasondotparse:
Maybe the docs should mention that the timeout is in ms? Some people might think it's measured in seconds :) 

(Some people === me)

loui7 (approved):

nfcampos:
Yep good point, have added that

Issue #388, opened 2023-03-20
How to set proxy for openai


sunlujing:
1. i need to config a proxy to call the openai.
2. so where should i to config the proxy ?

Cyansce:
same question + 1

duanwei913:
node's native fetch api is still in experiment, which doesn't offer a way to setup proxy AFAICT

nfcampos:
Thanks for providing the solution @yych42

allanbritos:
> ```ts
> { basePath: 'https://proxy.your-server.com/v1' }
> ```

@yych42 , https://proxy.your-server.com/v1 refers to an internet proxy or to a reverse_proxy?

yych42:
> ```ts
> ```ts
> { basePath: 'https://proxy.your-server.com/v1' }
> ```
> 
> 
>     
>       
>     
> 
>       
>     
> 
>     
>   
> 
> @yych42 , https://proxy.your-server.com/v1 refers to an internet proxy or to a reverse_proxy?
> ```

sorry about the delays. its a reverse proxy.

qingtian1771:
it seems do not work for new version now

fancn21th:
> it seems do not work for new version now

me 2

PR #550, opened 2023-03-31
change the order of fetchAdaptor and axios baseOptions

zombieyang:
Then we have a different way to solve #388.
```
const model = new OpenAI({
    openAIApiKey: '',
    temperature: 0.9
}, {
    baseOptions: {
        httpsAgent: new SocksProxyAgent({
            hostname: 'xxx',
            port: xxx
        }),
        adapter: null
    }
});
```

### DIFF
--- a/langchain/src/llms/openai-chat.ts
+++ b/langchain/src/llms/openai-chat.ts
@@ -320,8 +320,8 @@ export class OpenAIChat extends LLM implements OpenAIInput {
         ...this.clientConfig,
         baseOptions: {
           timeout: this.timeout,
-          ...this.clientConfig.baseOptions,
           adapter: fetchAdapter,
+          ...this.clientConfig.baseOptions,
         },
       });
       this.client = new OpenAIApi(clientConfig);
--- a/langchain/src/llms/openai.ts
+++ b/langchain/src/llms/openai.ts
@@ -366,8 +366,8 @@ export class OpenAI extends BaseLLM implements OpenAIInput {
         ...this.clientConfig,
         baseOptions: {
           timeout: this.timeout,
-          ...this.clientConfig.baseOptions,
           adapter: fetchAdapter,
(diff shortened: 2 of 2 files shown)


nfcampos:
Thanks

zitudu:
This should solve the problem of invalid proxy settings. However, its configuration seems not intuitive. The axios' adapter itself supports the http_proxy and https_proxy environment variables, and it does not require any configuration. But now, to use axios's default behavior, we need to add `adapter: null`, which users don't need to care about.

nfcampos:
Yes, but if we don't patch the adapter no one can use this outside of Node.

zitudu:
Good consideration. Axios supports the [browser environment](https://github.com/axios/axios#Browser-support) and it is implemented based on XHR. Langchain uses fetch API for both NodeJS and Browser, which is good. By using fetchAdapter, we can set a proxy for OpenAI interfaces, but it could be challenging to set a proxy for the original fetch API used directly by LangChain.

zombieyang:
> Yes, but if we don't patch the adapter no one can use this outside of Node.

Why do we patch the adapter for all environment but not just for non-node? Does the adapter has potential to not compatiable with node's fetch?

nfcampos:
This adapter is fully compatible with all environments which implement the fetch spec, including node. We prefer to patch it everywhere to avoid differences in behaviour between environments

lazyrong:
`const model = new OpenAI({
    openAIApiKey: '',
    temperature: 0.9,
   streaming: true,
   callbacks: [
            {
                handleLLMNewToken(token) {
                    process.stdout.write(token);
                },
            },
        ],
}, {
    baseOptions: {
        httpsAgent: new SocksProxyAgent({
            hostname: 'xxx',
            port: xxx
        }),
        adapter: null
    }
});`

But it seems it doesn't work if I set streaming to true.
throw network error like this：
image

PR #565, opened 2023-04-01
Add support for ChatAnthropic as the LLM in ZeroShotAgent

jacoblee93:
Builds on config changes in #534 

Adds an `llmType` arg to `CreatePromptArgs` for the ZeroShotAgent's `createPrompt` method. This allows us to conveniently set a default prompt for `ZeroShotAgent` based on the LLM type when the user creates the agent with the common `fromLLMAndTools` entry method. This will allow us to tweak default prompts for future supported models as well.


jacoblee93:
CC @nfcampos the issue appears that essentially Claude was treating the `Thought:` line in the ZeroShotAgent suffix as part of the user's message, rather than something it had written in the past. Adding the `AI_PROMPT` prefix seemed to help:
```
Begin!
Question: {input}
${AI_PROMPT} Thought: {agent_scratchpad}
```
I added a test case where it reliably used a calculator as well.

langchain/src/chat_models/anthropic.ts
```diff
@@ -27,6 +27,8 @@ function getAnthropicPromptFromMessage(type: MessageType): string {
   }
 }
 
+const DEFAULT_STOP_SEQUENCES = [HUMAN_PROMPT];
```
nfcampos, line 30: I've merged your other PR, this one might have to be rebased?

jacoblee93:
Any more thoughts on this @nfcampos?

nfcampos:
@jacoblee93 sorry, im working through the backlog of PRs, will get to this one in next couple days

jacoblee93:
No worries at all, didn't mean to press!

jacoblee93:
Closing for now, may revisit someday.

PR #567, opened 2023-04-01
`Fix passing chat history into buffer memory, add tests, update docs

jacoblee93:
Followup to #538 

The `chatHistory` field was not getting passed through to the `BaseChatMemory` constructor. This fixes that issue, adds tests, and updates the docs with an example.


jacoblee93:
Would it also make sense to just pass through the entire `fields` object into the call to `super()` e.g. here?:
https://github.com/hwchase17/langchainjs/pull/567/files#diff-18454f03f31ddfe1bf30aaeed806985f3089ecb3be4ea0424e31db55a80481a4R26

nfcampos (approved):
Thanks!

PR #576, opened 2023-04-02
Update the implements of MapReduceDocumentsChain and fix typo

hui0808:
I found that `MapReduceDocumentsChain` implements `StuffDocumentsChainInput`, but `MapReduceDocumentsChainInput` was not being used. So I updated it and fixed a typo.


langchain/src/chains/combine_docs_chain.ts
```diff
@@ -96,7 +96,7 @@ export class StuffDocumentsChain
 export interface MapReduceDocumentsChainInput extends StuffDocumentsChainInput {
   maxTokens: number;
   maxIterations: number;
-  combineDocumentsChain: BaseChain;
+  combineDocumentChain: BaseChain;
```
nfcampos, line 99: This is technically a breaking change so we'd want to keep around both old and new names

PR #583, opened 2023-04-03
Optimize the sample code for Streaming documentation

rxliuli:
Fix typos in the sample code and simply optimize it to match the output in the docs, link to the docs: https://js.langchain.com/docs/getting-started/guide-chat#streaming

previous output
image

current output
image


agola11 (approved):
Looks great, thanks!

rxliuli:
Ah, a better way is `process.stdout.write(token)`, which doesn't clear the previous command line, but just doesn't wrap the output.

image

> This answer is from chatgpt...

PR #599, opened 2023-04-04
Add s3 file loader

jasondotparse:
## Summary 
Add a new loader type exactly like this one which already exists in the [python repo](https://python.langchain.com/en/latest/modules/indexes/document_loaders/examples/s3_file.html#).

## Notes
* The logic is largely the same as that which is in the python package, except for one difference: the python version benefits from its use of the `unstructured` python package which we do not have at our disposal in node. Therefore, the best I could do was allow a user to pass in the url of an existing unstructured API endpoint and make use of that within the S3 loader. I'd be open to suggestions if someone thinks a better solution is possible, but for now this seemed like the optimal path forward. 
* I could not find a way to use jest to mock the 'fs' and 'Unstructured' class dependencies of `S3Loader` for unit testing purposes. After multiple days of trying to mock these imports for a unit test, I finally fell back on just using dependency injection and passing them into the S3Loader class constructor so they could easily be stubbed in the test file. 
* After this work is done I will probably come back and update the class interface for the class to allow S3 access keys or perhaps a single S3 pre-signed URL to be passed into the S3 loader. And also add a [S3 directory loader](https://python.langchain.com/en/latest/modules/indexes/document_loaders/examples/s3_directory.html) to match the on in the python repo.


langchain/package.json
```diff
@@ -74,6 +74,7 @@
   "author": "Langchain",
   "license": "MIT",
   "devDependencies": {
+    "@aws-sdk/client-s3": "^3.306.0",
```
nfcampos, line 77: This should be added as a peer dependency as well

langchain/src/document_loaders/s3.ts
```diff
@@ -0,0 +1,95 @@
...
+        `Failed to load file ${filePath} using unstructured loader.`
+      );
+    }
+  }
+}
+
+async function S3LoaderImports() {
+  try {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    return await import("@aws-sdk/client-s3" as any);
```
nfcampos, line 88: Why is this `any` required?

langchain/src/document_loaders/s3.ts
```diff
@@ -0,0 +1,95 @@
...
+    this.key = key;
+    this.unstructuredAPIURL = unstructuredAPIURL;
+    this._fs = _fs;
+    this._UnstructuredLoader = _UnstructuredLoader;
+  }
+
+  public async load() {
+    const { S3Client, GetObjectCommand } = await S3LoaderImports();
+
+    const tempDir = this._fs.mkdtempSync(
```
nfcampos, line 36: I think we'd want to not use `fs` and work with a Buffer in memory. This would enable using this outside of Node, eg in Edge environments

I'm going to open a PR very soon to make the Unstructured loader work with Buffers directly

nfcampos (commented):
Looks good, just a few comments

langchain/src/document_loaders/s3.ts
jasondotparse: Good catch, it's not required. I just forgot to go back and add the strict type.

langchain/package.json
jasondotparse: Done!

examples/src/document_loaders/unstructured.ts
```diff
@@ -2,7 +2,7 @@ import { UnstructuredLoader } from "langchain/document_loaders";
 
 export const run = async () => {
   const loader = new UnstructuredLoader(
-    "http://localhost:8000/general/v0.0.4/general",
+    "http://localhost:8000/general/v0/general",
```
nfcampos, line 5: Why this change @jasondotparse ?

langchain/package.json
nfcampos: @jasondotparse can you add the optional flag under `peerDependenciesMeta` like the others?

examples/src/document_loaders/unstructured.ts
jasondotparse: I think the url was wrong. I tried it myself and I could only get Unstructured to work with my local docker instance after removing the ".0.4"

jasondotparse: I probably should have made that a separate PR though. I'll remove this change since it's not relevant to the rest of the S3 feature.

hwchase17 (approved):
lgtm!

PR #607, opened 2023-04-04
feat: Add RefineChain 

RohitMidha23:
Added Refine Chain with prompts as present in the python library for QA.

For issue: https://github.com/hwchase17/langchainjs/issues/483


langchain/src/chains/question_answering/refine_prompts.ts
```diff
@@ -0,0 +1,80 @@
...
+  PromptTemplate,
+  ChatPromptTemplate,
+  SystemMessagePromptTemplate,
+  HumanMessagePromptTemplate,
+  AIMessagePromptTemplate,
+} from "../../prompts/index.js";
+import { ConditionalPromptSelector, isChatModel } from "../prompt_selector.js";
+
+export const DEFAULT_REFINE_PROMPT_TMPL = `The original question is as follows: {question}
+    We have provided an existing answer: {existing_answer}
```
hwchase17, line 13: it looks like theres some rouge tabs here?

langchain/src/chains/combine_docs_chain.ts
```diff
@@ -215,3 +218,152 @@ export class MapReduceDocumentsChain
...
+    const documentInfo: Record<string, unknown> = {};
+    this.documentPrompt.inputVariables.forEach((value) => {
+      documentInfo[value] = baseInfo[value];
+    });
+    const baseInputs: Record<string, unknown> = {
+      [this.documentVariableName]: this.documentPrompt.format({
+        ...documentInfo,
+      }),
+    };
+    const inputs = { [this.initialResponseName]: res, ...baseInputs };
```
hwchase17, line 314: res right now is ChainValues, but it should be a just a strings

since this is an llm chain, there should only be one value in there. i think we actually expose `predict` in the JS LLMChain, which returns a string (so you could try using that)

hwchase17 (commented):
this is great! would be even better to have an example that we can plop in the documentation :)

one major comment around the `existing_answer` vairable

langchain/src/chains/question_answering/refine_prompts.ts
RohitMidha23: Fixed this.

langchain/src/chains/combine_docs_chain.ts
RohitMidha23: **Resolution:** 
Used `predict` instead of call. 
The `_call` function now returns the string response `res` using the `outputKey`.

RohitMidha23:
> this is great! would be even better to have an example that we can plop in the documentation :)
> 
> one major comment around the `existing_answer` vairable

Will try to add an example but facing 429 issue when querying using OPENAI_API_KEY.

hwchase17 (approved):
this lgtm! thanks

langchain/src/chains/combine_docs_chain.ts
```diff
@@ -215,3 +218,152 @@ export class MapReduceDocumentsChain
...
+    this.refineLLMChain = fields.refineLLMChain;
+    this.documentVariableName =
+      fields.documentVariableName ?? this.documentVariableName;
+    this.inputKey = fields.inputKey ?? this.inputKey;
+    this.documentPrompt = fields.documentPrompt ?? this.documentPrompt;
+    this.initialResponseName =
+      fields.initialResponseName ?? this.initialResponseName;
+  }
+
+  _constructInitialInputs(docs: Document[], rest: Record<string, unknown>) {
```
hwchase17, line 281: uber nit: maybe just pass only `docs[0]` into this function since thats all thats needed

PR #613, opened 2023-04-04
Add MetalRetriever

nfcampos:


langchain/src/retrievers/metal.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    this.client = fields.client;
+  }
+
+  async getRelevantDocuments(query: string): Promise<Document[]> {
+    const res = await this.client.search({ text: query });
+
+    const items = ("data" in res ? res.data : res) as ResponseItem[];
+    return items.map(
+      ({ text, metadata }) =>
```
softboyjimbo, line 29: we also return a `dist` which shows the similarity distance when running a search with a `text` query. Perhaps we should include this in the `metadata` of `Document` below?

softboyjimbo (approved):

softboyjimbo:
One minor comment but this is great! Thanks @nfcampos

PR #618, opened 2023-04-04
Support combining multiple output parsers

jmandel:
Not sure if this is a recommended approach, but I'm finding in my prompts that I want to cause outputs like:

```
{
  "system": "http://www.nlm.nih.gov/research/umls/rxnorm",
  "display": "hydroxychloroquine",
  "code": "5521",
}

New Query: ?system=loinc&display=medication
  
Final Grade: A

Justification: The code "5521" for "hydroxychloroquine" is the most appropriate code for the medication
mentioned in the clinical note. It is a valid RXNORM code and accurately represents the medication mentioned.
```

It's easier for me to create a JSONOutputParser and combined it with a couple of RegexOutpuParsers,and the logic for this combination can be abstracted into something like this "Combining" output parser.


nfcampos:
@jmandel thanks for this, could you provide an example of this in action?

jmandel:
@nfcampos sure thing -- here's how I'm using it currently, to pare output like in the example above:

```ts
    const chain = new LLMChain({
      outputParser: new MultiOutputParser(
        new CodingsParser("codings"), // parses all JSON objects in output --> `Record<string, any>[]`
        new RegexParser(/Final Grade: (A|B|C)/i, ["grade"], "noGrade"),
        new RegexParser(/Justification: (.*)/i, ["rationale"], "noRationale"),
        new RegexParser(
          /New Query: \?system=(\S+)&display=(\S+)/i,
          ["newQuerySystem", "newQuery"],
          "noQuery"
        )
      ),
      prompt,
      llm
    });

```

hwchase17 (commented):
oh this is pretty cool. i like this a lot - i can do something similar in the python version, and then we can release at the same time and make a cool announcement

i think all thats missing from this PR is an example to put in the docs

jmandel:
Added an example and updated the class name.

examples/src/prompts/combining_parser.ts
```diff
@@ -0,0 +1,66 @@
+import { OpenAI, PromptTemplate } from "langchain";
```
nfcampos, line 1:
```suggestion
import { OpenAI } from "langchain/llms/openai";
import { PromptTemplate } from "langchain/prompts";
```

PR #621, opened 2023-04-05
[Work in progress] Create composition.mdx

jasondotparse:
The docs could use a clear example of agent composition to make it obvious that agents can be used to coordinate actions with other agents doing intermediary work. 

To make sure this capability of Langchain is made crystal clear, this CR adds a new page to the docs under the "Agent examples" section which showcases this feature with a simple example. 

This is a work in progress with https://github.com/atbe, as the source code example has not yet been completed.


jacoblee93:
@jasondotparse This is a great idea if you're willing to revive this PR!

PR #626, opened 2023-04-05
Add CJS build, add package testing CJS exports

nfcampos:
FYI I'm aware of the "dual package hazard", our library is entirely free of global state (and we make very little use of `instanceof` checks), we have tests in place that check this on CI. This is following this approach recommended by Node https://nodejs.org/api/packages.html#approach-2-isolate-state


langchain/src/document_loaders/tests/csv-blob.test.ts
```diff
@@ -1,6 +1,6 @@
-import url from "node:url";
-import path from "node:path";
-import fs from "node:fs/promises";
+import * as url from "node:url";
```
transitive-bullshit, line 1: this is great

prettier is super powerful w/ imports btw; I recommend using one of the prettier import sorting plugins ala https://github.com/transitive-bullshit/chatgpt-api/blob/main/.prettierrc.cjs

just completely removes having to think about imports and ordering best practices and is a huge time saver

langchain/src/llms/openai-chat.ts
```diff
@@ -326,11 +326,13 @@ export class OpenAIChat extends LLM implements OpenAIInput {
...
-      request,
-      options
-    );
+    return this.caller
+      .call(
+        this.client.createChatCompletion.bind(this.client),
+        request,
+        options
+      )
+      .then((res) => res.data);
```
transitive-bullshit, line 335: why are you removing `.then((res) => res.data)` in some places but still using it here?

not saying it's wrong; just making sure cause it looks inconsistent

langchain/src/util/async_caller.ts
```diff
@@ -1,5 +1,5 @@
 import { backOff } from "exponential-backoff";
-import PQueue from "p-queue";
+import PQueueMod from "p-queue";
```
transitive-bullshit, line 2: LOVE `p-queue` and `p-map` 🔥

langchain/src/vectorstores/hnswlib.ts
```diff
@@ -133,6 +131,8 @@ export class HNSWLib extends SaveableVectorStore {
   }
 
   async save(directory: string) {
+    const fs = await import("node:fs/promises");
```
transitive-bullshit, line 134: is it clear which methods and classes require node imports?

what happens if I try to call `save` in browser/edge/deno/etc?

test-exports-cjs/import.js
```diff
@@ -37,6 +38,13 @@ async function test() {
   );
 
   assert((await vs.similaritySearchVectorWithScore([0, 0, 1], 1)).length === 1);
+
+  // Test CSVLoader
+  const loader = new CSVLoader(new Blob(["a,b,c\n1,2,3\n4,5,6"]));
+
+  const docs = await loader.load();
```
transitive-bullshit, line 45: `loader.load` feels a lil off; I understand why it's separate from the ctor cause it's async. maybe `loader.init`?

not sure..

test-exports-cjs/index.mjs
```diff
@@ -0,0 +1,46 @@
...
+import { InMemoryDocstore, Document } from "langchain/docstore";
+import { CSVLoader } from "langchain/document_loaders";
+
+// Test exports
+assert(typeof OpenAI === "function");
+assert(typeof loadPrompt === "function");
+assert(typeof HNSWLib === "function");
+
+// Test dynamic imports of peer dependencies
+const { HierarchicalNSW } = await HNSWLib.imports();
```
transitive-bullshit, line 15: this line surprised me; I'm not used to seeing the type of thing happening at runtime. I know you're just testing here, but the fact that `imports()` is part of the public api made me glance twice.

dealing w/ deps / peer deps / optional peer deps is really tricky as I'm sure you're well aware, especially across diff environments.

for now, if this abstraction works for you, I'd say run with it. just wanted to point out that it was surprising

test-exports-cjs/index.mjs
```diff
@@ -0,0 +1,46 @@
...
+
+const vs = new HNSWLib(new OpenAIEmbeddings({ openAIApiKey: "sk-XXXX" }), {
+  space: "ip",
+  numDimensions: 3,
+  docstore: new InMemoryDocstore(),
+  index: new HierarchicalNSW("ip", 3),
+});
+
+await vs.addVectors(
```
transitive-bullshit, line 24: all of the actual usage of these classes is so 🔥 btw

test-exports-cjs/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "description": "CJS Tests for the things exported by the langchain package",
+  "main": "./index.mjs",
+  "scripts": {
+    "test": "npm run test:esm && npm run test:cjs && npm run test:cjs:import && npm run test:ts",
+    "test:esm": "node ./index.mjs",
+    "test:cjs": "node ./require.js",
+    "test:cjs:import": "node ./import.js",
+    "test:ts": "tsc && node dist/index.js",
+    "format": "prettier --write \"**/*.ts\"",
+    "format:check": "prettier --list-different \"**/*.ts\""
```
transitive-bullshit, line 14: consider adding `test:format` that runs `prettier --check` like this

also same feedback w/ using `npm-run-all` (or a similar alternative to the `&&` approach)

transitive-bullshit (commented):
looked through the PR and tried to provide some hopefully useful feedback.

I don't have time to dive as deep as I'd like unfortunately.

a few more high-level pieces of feedback:

### 1. Do you really want to / need to support CommonJS?

The community is pretty split on this; it's like the old days of the python2 => python3 painful upgrade.

A lot of maintainers like [Sindre Sorhus](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c) are choosing to explicitly only support ESM. It will make your life so so sooooo much easier. That's what I decided to do for the [chatgpt](https://github.com/transitive-bullshit/chatgpt-api) package btw. 

If you support both, it puts more work on you guys, and you'll get a lot of random, esoteric build chain issues like "when I import into this webpack app, this fails". This is *not* the type of stuff you really want to waste your team's dev resources on.

On the flip side, if you only support ESM, you'll get a lot of issues with less experienced devs who don't understand how to use the package. Sindre has a [guide](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c) he uses for all of his packages that I like. Just make sure to mention it prominently in your docs like I do here:

```ts
async function example() {
  // To use ESM in CommonJS, you can use a dynamic import like this:
  const { ChatGPTAPI } = await import(https://github.com/transitive-bullshit/chatgpt-api)
  // You can also try dynamic importing like this:
  // const importDynamic = new Function('modulePath', 'return import(modulePath)')
  // const { ChatGPTAPI } = await importDynamic('chatgpt')

  const api = new ChatGPTAPI({ apiKey: process.env.OPENAI_API_KEY })

  const res = await api.sendMessage('Hello World!')
  console.log(res.text)
}
```

It's up to you whether you want to support both; just don't be surprised if the CJS side of things ends up being painful to maintain w/ random build chain bugs down the road.

### Optional / peer dependencies

I couldn't really tell, but just wanted to make sure you're aware of NPM's support for [optionalDependencies](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#optionaldependencies) and [peerDependenciesMeta](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#peerdependenciesmeta), which lets you specify optional peer dependencies.

### Misc

Also, two useful toys you want to add to your codebase would be using `esbuild` to replace `tsc` (still using `tsc` to output types), maybe a wrapper around `esbuild` like `tsup` to help with transpilation & bundling dependencies. Generally relying on `tsc` (which I think is what this is currently doing) isn't always enough to really get you what you want, but I know you're doing some custom build script stuff here, so maybe it's nice to have full control. Also check out `tsx` which is my go-to for running TS-based scripts with `esbuild` without needing a build step. Super fast for local testing && running scripts.

---

Hope that's useful feedback. Sorry it's not more detailed. Really really love what you guys are building w/ LangChain && excited to see the JS/TS side of things mature 🔥 💕

langchain/package.json
```diff
@@ -241,83 +263,104 @@
   "exports": {
     ".": {
       "types": "./index.d.ts",
-      "import": "./index.js"
+      "import": "./index.js",
```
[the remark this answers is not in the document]
nfcampos: Yep adding the default in the next PR

langchain/scripts/check-tree-shaking.js
```diff
@@ -8,6 +8,9 @@ export function listEntrypoints() {
   const entrypoints = [];
 
   for (const [key, value] of Object.entries(exports)) {
+    if (key === "./package.json") {
```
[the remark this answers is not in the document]
nfcampos: This is a test file, but it's run in the build command because it requires the build output

langchain/scripts/move-cjs-to-dist.js
```diff
@@ -0,0 +1,38 @@
+import { resolve, dirname, parse, format } from "node:path";
```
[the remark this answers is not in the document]
nfcampos: Yea I want to avoid bundling dependencies (for now at least) as that's a bit too opinionated for me

langchain/src/document_loaders/tests/csv-blob.test.ts
nfcampos: Oh need to set that up

langchain/src/llms/openai-chat.ts
nfcampos: It's being moved here (from where this function is called)

langchain/src/vectorstores/hnswlib.ts
nfcampos: In deno it would work. In other envs it wouldn't. One thing I haven't done yet is add some indication in the docs of which methods are node-only, that's planned

test-exports-cjs/import.js
nfcampos: Hmm yea I think we went with this because it's the name used in python

test-exports-cjs/index.mjs
nfcampos: Hmm yea not without its issues, but it has mostly worked for us. Surprisingly hard to do optional deps in JS nowadays

PR #632, opened 2023-04-05
Enable support for Browsers, Cloudflare Workers, Next.js Browser/Serverless/Edge

nfcampos:
~- [ ] Split File loaders from Blob loaders (File loaders remain with the existing name, Blob loaders are new)~ Will tackle later in #663
- [x] Fix `Cannot find module '@fortaine/fetch-event-source/parse' from '../langchain/dist/util/axios-fetch-adapter.js'` replace with eventsource-parser ?
- [x] Add package to test on CF Workers
- [x] Add package to test on browser with `create-react-app` template
- [x] Test when peer dependencies are not present (not sure how to do this in CI, but that's secondary)


agola11 (approved):

dankolesnikov:
this is very exciting

pegak:
I am hyped about this. Still have problems with streaming, this might be huge step forward. :)

PineappleExpress808:
+1, i will test w/ the lex-gpt app (https://lex-gpt.vercel.app/) as soon as this goes in :)

Sameedahmad:
@PineappleExpress808 Any updates? Is it working and if so let us know of any roadblocks ahead.

nfcampos:
This has been published as a prerelease, you can install it with `npm i langchain@next`.

See upgrade instructions here https://langchainjs-docs-git-nc-test-exports-cf-langchain.vercel.app/docs/getting-started/install

If you test it let me know any issues

test-exports-vercel/src/pages/api/hello-edge.ts
```diff
@@ -0,0 +1,69 @@
...
+  // Test a chain + prompt + model
+  const chain = new LLMChain({
+    llm,
+    prompt: ChatPromptTemplate.fromPromptMessages([
+      HumanMessagePromptTemplate.fromTemplate("{input}"),
+    ]),
+  });
+
+  // Run the chain but don't await it
+  chain.run("hello").catch(console.error);
```
agola11, line 61: why don't we await here?

langchain/src/util/event-source-parse.ts
agola11, line 1: why do we need this?

langchain/scripts/create-entrypoints.js
```diff
@@ -45,29 +187,51 @@ const generateFiles = () => {
...
+
+          // If there is a *.lite.js file add it as the root `import` export,
+          // which should/will then be used by non-Node environments.
+          const litePath = `./dist/${entrypoints[key]}.lite.js`;
+
+          if (deprecatedNodeOnly.includes(key)) {
+            entryPoint = {
+              node: entryPoint,
+            };
+          } else if (fs.existsSync(litePath)) {
```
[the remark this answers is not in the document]
agola11: yeah should probably remove

agola11 (approved):
From my perspective (mostly looking at import refactoring) looks good! Thanks!

test-exports-vercel/src/pages/api/hello-edge.ts
nfcampos: Because we need the response to start being sent to the user right away. If we await then we'll only start sending response after the chain run finishes, therefore no streaming

langchain/src/util/event-source-parse.ts
nfcampos: Because there was some packaging issue with the dependency we were using for this and it was easiest to include a copy here. Same licence as we have

PR #643, opened 2023-04-06
docs: Add Databerry.ai to the ecosystem doc

gmpetrov:
Add Databerry.ai https://github.com/gmpetrov/databerry to the ecosystem documentation


hwchase17 (commented):
this is really exciting!

two quick comments:

1. we should definitley include an example of using the retriever in the modules/indexes/retrievers section (will help more people find it)

2. the RemoteLangChainRetriever is still pretty in flux. I want to make sure we dont make any big changes that mess you up. One idea would be to write a databerry specific wrapper (would also be helpful in terms of letting people more easily specify params like `top k`) - thoughts?

gmpetrov:
Thx for the feedback!

LGTM

- Added a DataberryRetriever 
- Added doc usage in modules/indexes/retrievers  + associated mdx
- Updated docs/ecosystem/databerry.md

hwchase17 (commented):
awesome! this lgtm

in terms of releasing and publicizing, i know you did something similar on the python side (will leave some comments there) but i think it could be most effective to do a retriever wrapper there as well, and then publicize both toghether?

gmpetrov:
Done! https://github.com/hwchase17/langchain/pull/2486

gmpetrov:
> awesome! this lgtm
> 
> in terms of releasing and publicizing, i know you did something similar on the python side (will leave some comments there) but i think it could be most effective to do a retriever wrapper there as well, and then publicize both toghether?

@hwchase17 let me know when you want to communicate about this. 
I m close to my Openai's billing usage limit, I requested an increase but no news from them yet.
So it might be better to wait until this issue is resolved. 

By any chance, do you know someone at OpenAI that could quickly validate the request?

hwchase17:
woo - will try to get both in release tmrw

gmpetrov:
Cool! I'll be back to work next Tuesday, ok if we wait to communicate about this until then?

hwchase17:
> Cool! I'll be back to work next Tuesday, ok if we wait to communicate about this until then?

ah yes absolutely

PR #650, opened 2023-04-06
Update url in UnstructuredLoader example

awalker4:
Unstructured now uses the major version in its urls to avoid having to track every change.


hwchase17 (approved):
thanks!

PR #654, opened 2023-04-06
docs: Add apiKey param for model instantiation

mvikrammenon:
Added key to huggingface, cohere and replicate


hwchase17 (commented):
they can also be set via environment variables right? i dont want to make it seem like this is the only way to do this

nfcampos:
Thanks!

PR #655, opened 2023-04-06
WIP MongoVectorStore implementation

floomby:
I have implemented a somewhat functional mongo vector store however there some inconveniences/issues.

- `knnVector` indexes are only available in atlas and not on the community server version of mongo which means testing needs to be done against a cloud instance.
- The index needs to be setup in atlas using the atlas cli, atlas ui or through the atlas api. (Pursuing a way to automatically setup the index through the atlas api may be doable, but I would be inclined to think that that is probably outside the scope of this library)
- There appears to be a rather small bound on the dimensionality of a `knnVector` index which precludes using the OpenAI embeddings (`dimension: 1536`). I have not tested to see what exactly this bound is.
- Testing is a little awkward as well because afaik there is no straightforward way to know when the indexing of added documents is complete with just the mongo driver apart from trying again after a timeout a couple of times until something is returned or it is decided that something failed.

Knn vector search is in beta for atlas and it shows somewhat as the UI, namely the visual index editor as well as the query tester don't support it yet. I didn't find much information about the bound limitation on the index, I am hoping that it is either configurable somewhere or just a temporary bound which will increased in the near future.


floomby:
I would like to get this merged at some point. Currently the integration test on in the workflow is going to fail.

I will be happy setup atlas to test against, or work with a maintainer to make this happen. Alternatively the integration test could be no-oped in a manner like the current chroma integration test.

nfcampos:
Hi @floomby I've reviewed this and fixed the merge conflicts, this is great work thank you! I've skipped the integration test, and will merge this in now

nikhilkotiya:
is there any docs for this, How can i use this? i'm bit confused

prakul:
@nikhilkotiya The docs are listed https://python.langchain.com/docs/integrations/vectorstores/mongodb_atlas

PR #659, opened 2023-04-07
Pinecone Hybrid Search

trevorpfiz:
I tried to emulate the langchain python implementation of Pinecone Hybrid Search. I also used the Supabase Hybrid Search implementation as a reference. Thanks @hwchase17 @ericciarla 

I went with `BertWordPieceTokenizer` from `tokenizers` as the tokenizer. I also had to update the `@pinecone-database/pinecone` to `^0.0.12` as I could not add `sparseVectors` otherwise.

Not 100% sure if everything is in working order yet, but I will work on testing it in my local project when I get the chance. A good starting point nonetheless!


ElektrikSpark:
I had an oversight with using the `tokenizers` library. It only supports `"node": ">=10 < 11 || >=12 <14"`. Will evaluate alternative methods to create the sparse vectors.

ElektrikSpark:
@nfcampos @hwchase17 I think this is ready to be reviewed whenever the chance. 

I am using `bert-tokenizer` as the tokenizer for now, but down the road we will want to support a greater variety of tokenizer options. I initially tried to implement `tokenizers` from Hugging Face, but the node bindings are too old.

nfcampos:
Note to me: use tiktoken instead of bert

examples/src/retrievers/pinecone_hybrid.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  await client.init({
+    environment: process.env.PINECONE_ENVIRONMENT!,
+    apiKey: process.env.PINECONE_API_KEY!,
+  });
+
+  const embeddings = new OpenAIEmbeddings();
+  const pineconeIndex = client.Index(process.env.PINECONE_INDEX!);
+
+  const tokenizer = new BertTokenizer(undefined, true, 512);
```
dqbd, line 18: Consider using `js-tiktoken` instead, which should match the behavior of `OpenAIEmbeddings`

langchain/src/retrievers/pinecone-hybrid-search.ts
```diff
@@ -0,0 +1,148 @@
...
+  namespace?: string;
+
+  filter?: PineconeMetadata;
+
+  constructor(embeddings: Embeddings, args: PineconeHybridSearchParams) {
+    super();
+    this.embeddings = embeddings;
+    this.pineconeIndex = args.pineconeIndex;
+    this.tokenizer = args.tokenizer;
+    this.topK = args.topK;
```
dqbd, line 40: _nit_: the default values are being overriden, should that be the case?


```suggestion
    this.topK = args.topK ?? 4;
```

langchain/src/retrievers/pinecone-hybrid-search.ts
```diff
@@ -0,0 +1,148 @@
...
+    this.alpha = args.alpha;
```
dqbd, line 41: _nit_: the default values are being overriden, should that be the case?

```suggestion
    this.topK = args.topK ?? 4;
    this.alpha = args.alpha ?? 0.5;
```

langchain/src/retrievers/index.ts
```diff
@@ -1,5 +1,9 @@
 export { RemoteRetriever } from "./remote/base.js";
 export { ChatGPTPluginRetriever } from "./remote/chatgpt-plugin.js";
+export {
+  PineconeHybridSearchRetriever,
```
dqbd, line 4: Instead of adding a file into a barrel file, create a new entrypoint instead.

langchain/src/retrievers/pinecone-hybrid-search.ts
```diff
@@ -0,0 +1,148 @@
...
+  topK: number;
+  alpha: number;
+}
+
+export class PineconeHybridSearchRetriever extends BaseRetriever {
+  embeddings: Embeddings;
+
+  pineconeIndex: VectorOperationsApi;
+
+  tokenizer: BertTokenizer;
```
dqbd, line 23: I think it would make more sense for the sparse encoder to be more configurable, not assuming BERT tokenization.

langchain/src/retrievers/pinecone-hybrid-search.ts
```diff
@@ -0,0 +1,148 @@
...
+    }
+  }
+
+  hybridScale(
+    dense: number[],
+    sparse: SparseValues,
+    alpha: number
+  ): [number[], SparseValues] {
+    if (alpha < 0 || alpha > 1) {
+      throw new Error("Alpha must be between 0 and 1");
```
dqbd, line 102: _nit_: consider validating `this.alpha` in the constructor. 

_nit_: do we need to pass `this.alpha` as a parameter? Why not access it directly?

dqbd: _nit_: consider validating `this.alpha` in the constructor. 
_nit_: do we need to pass `this.alpha` as a parameter? Why not access it directly?

dqbd (changes requested):

bracesproul:
@carlosdp going to close due to merge conflicts. If you'd still like to get this merged in, please open off of an up-to-date `main` branch, and I'll review. Thank you!

PR #662, opened 2023-04-07
add better error message when the tableinfo method return too much info

MaximeThoonsen:


langchain/src/chains/sql_db/sql_db_chain.ts
```diff
@@ -127,4 +130,27 @@ export class SqlDatabaseChain extends BaseChain {
...
+  private async verifyNumberOfTokens(
+    inputText: string,
+    tableinfo: string
+  ): Promise<void> {
+    // We verify it only for OpenAI for the moment
+    if (!(this.llm instanceof OpenAI)) {
+      return;
+    }
+    const promptTemplate = this.prompt.template;
+    const stringWeSend = `${inputText}${promptTemplate}${tableinfo}`;
```
MaximeThoonsen, line 143: it's an easy to understand approximation of the final string

MaximeThoonsen:
I did this PR because the current error message (a 400 from the API) is not really helping to debug and I was asked for help to debug this kind of error on discord. 
Screenshot from 2023-04-07 14-58-07


The python error message is a bit better but we could also back port this in the python version. ping @hwchase17 

Screenshot from 2023-04-07 14-53-05

MaximeThoonsen:
The new error message looks like this.
Screenshot from 2023-04-07 15-48-23

hwchase17 (approved):
this looks pretty good to me! thanks for the improvement

nfcampos:
Thanks!

PR #666, opened 2023-04-07
docs: Mendable v79 new features + bug fixes

nickscamara:


nfcampos:
It seems the source links open a new tab, a little odd?
Screenshot 2023-04-07 at 22 05 29

nickscamara:
Good point. The reason it opens a new tab is that sometimes the user might want to continue the conversation after looking at a source. Because we currently don't save the conversation context when it is dismissed, the conversation gets wiped. 

Will look into adding that in newer versions. Tks for the feedback.

PR #669, opened 2023-04-08
Improve Logging Output

more-by-more:
Added some simple logging colours to the ConsoleCallbackHandler along with a utility for Chalk. Have also improved the formatting of the log outputs to create consistent newlines and trim leading / trailing whitespace.


nfcampos:
@agola11 to get this in we'd want to move the ConsoleCallbackHandler to a separate entry point to avoid including this dependency for everyone, thoughts?

langchain/package.json
```diff
@@ -234,6 +234,7 @@
     "@anthropic-ai/sdk": "^0.4.2",
     "@fortaine/fetch-event-source": "^3.0.6",
     "browser-or-node": "^2.1.1",
+    "chalk": "^5.2.0",
```
nfcampos, line 237: We need to stick to `chalk@4` as 5.x is ESM only

langchain/src/callbacks/base.ts
```diff
@@ -351,26 +353,30 @@ export class CallbackManager extends BaseCallbackManager {
...
   async handleAgentAction(action: AgentAction) {
-    console.log(action.log);
+    if (action.log) {
+      colorLog(LogColor.Yellow, `\n${action.log.trim()}`);
+    }
   }
 
   async handleToolEnd(output: string) {
-    console.log(output);
+    if (output) {
```
nfcampos, line 370: @agola11 when would it be valid for a tool to return no output?

langchain/src/callbacks/handlers/tracers.ts
```diff
@@ -39,6 +39,10 @@ export interface ChainRun extends BaseRun {
   child_tool_runs: ToolRun[];
 }
 
+export interface AgentRun extends ChainRun {
+  actions: AgentAction[];
```
nfcampos, line 43: @agola11 I checked and looks like the backend will just ignore this

langchain/src/callbacks/handlers/tracers.ts
```diff
@@ -264,8 +273,48 @@ export abstract class BaseTracer extends BaseCallbackHandler {
...
+    }
+    const agentRun = run as AgentRun;
+    agentRun.actions = agentRun.actions || [];
+    agentRun.actions.push(action);
+    await this.onAgentAction?.(run as AgentRun);
+  }
+
+  // custom event handlers
+
+  onLLMStart?(run: LLMRun): void | Promise<void>;
```
nfcampos, line 293: @agola11 I've added these separate methods because
1. It's much nicer to work with the Run objects
2. It's tricky to know exactly when to call them, eg. in start is in the last line, in end methods it's right before we delete the run etc

langchain/src/callbacks/manager.ts
```diff
@@ -395,8 +395,15 @@ export class CallbackManager
     inherit = true
   ): CallbackManager {
     const manager = new CallbackManager(this._parentRunId);
-    manager.setHandlers([...this.handlers].map((handler) => handler.copy()));
+    for (const handler of this.handlers) {
```
nfcampos, line 398: @agola11 this wasn't right, it was losing the "inheritable" flag on each handler

langchain/src/callbacks/manager.ts
```diff
@@ -395,8 +395,15 @@ export class CallbackManager
...
+      const inheritable = this.inheritableHandlers.includes(handler);
+      const copied = handler.copy();
+      manager.addHandler(copied, inheritable);
+    }
     for (const handler of additionalHandlers) {
+      if (manager.handlers.some((h) => h.name === handler.name)) {
```
nfcampos, line 404: This is to ensure we don't end up with multiple copies of the same handler in here, if eg someone sets it both in constructor and `call()`

langchain/package.json
```diff
@@ -450,6 +450,7 @@
   "dependencies": {
     "@anthropic-ai/sdk": "^0.4.3",
     "@dqbd/tiktoken": "^1.0.4",
+    "ansi-styles": "^5.0.0",
```
nfcampos, line 453: This is a much lighter dependency which should work in every env. It is used by chalk under the hood

nfcampos:
Thanks a lot for this @more-by-more I've made a few updates for latest changes in main, and replaced chalk with a lighter dependency.

langchain/src/callbacks/handlers/tracers.ts
```diff
@@ -114,14 +111,43 @@ export abstract class BaseTracer extends BaseCallbackHandler {
...
-  protected async _endTrace(run: LLMRun | ChainRun | ToolRun): Promise<void> {
+  protected async _endTrace(run: Run): Promise<void> {
     if (!run.parent_uuid) {
       await this.persistRun(run);
-      this.executionOrder = 1;
     }
     this.runMap.delete(run.uuid);
   }
 
+  protected _getExecutionOrder(parentRunId: string | undefined): number {
```
nfcampos, line 121: @agola11 check this is what execution order is meant to be

langchain/src/callbacks/handlers/tracers.ts
```diff
@@ -378,15 +452,6 @@ export class LangChainTracer extends BaseTracer {
   }
 
   copy(): LangChainTracer {
```
nfcampos, line 454: Move this to BaseTracer

langchain/src/callbacks/manager.ts
agola11: I see the reason for this but what if someone has an actual usecase where they use the same callback handler (say a websockethandler) two different ways? Let's say there is one they pass through that writes to one websocket but one on the object itself that is writes to a different one. In this case, the one that was passed through will be used while the other one on the object won't be used, potentially leading to unexpected behavior.

langchain/src/callbacks/manager.ts
agola11: ah yeah, thanks for fixing.

langchain/src/callbacks/handlers/tracers.ts
agola11: looks reasonable. Do we really need the onAgentRun/handleText here though?

langchain/src/callbacks/handlers/tracers.ts
```diff
@@ -378,15 +452,6 @@ export class LangChainTracer extends BaseTracer {
...
-    // TODO: this is a hack to get tracing to work with the current backend
-    // we need to not use execution order, then remove this check
-    if (this.executionOrder === 1) {
-      const copy = new LangChainTracer();
-      copy.session = this.session;
-      copy.runMap = new Map(this.runMap);
-      copy.executionOrder = this.executionOrder;
-      return copy;
-    }
     return this;
```
agola11, line 455: This is safe now because `execution_order` gets calculated every time it's accessed?

agola11 (approved):
overall looks great, thanks!

langchain/src/callbacks/handlers/tracers.ts
nfcampos: Yes exactly

langchain/src/callbacks/handlers/tracers.ts
```diff
@@ -114,14 +111,43 @@ export abstract class BaseTracer extends BaseCallbackHandler {
...
+    if (parentRun === undefined) {
+      throw new Error(`Parent run ${parentRunId} not found`);
+    }
+
+    // If a run has a parent then execution order is
+    // parent's execution order + (count of previous child runs) + 1
+
+    let executionOrder = parentRun.execution_order;
+
+    if (parentRun.type === "chain" || parentRun.type === "tool") {
```
agola11, line 138: think we should raise an error if the parent run type is "llm", that shouldn't be possible.

langchain/src/callbacks/handlers/tracers.ts
```diff
@@ -114,14 +111,43 @@ export abstract class BaseTracer extends BaseCallbackHandler {
...
+  protected async _endTrace(run: Run): Promise<void> {
     if (!run.parent_uuid) {
       await this.persistRun(run);
-      this.executionOrder = 1;
     }
     this.runMap.delete(run.uuid);
   }
 
+  protected _getExecutionOrder(parentRunId: string | undefined): number {
+    // If a run has no parent then execution order is 1
```
agola11, line 122: actually, this is not going to work because it looks at direct children, not at all children recursively.

let's say top level chain calls tool, then tool calls llm. Then the top level chain calls llm. The llm's execution order will incorrectly be computed as 3 (`parent_run.execution_order` = 1, number of direct children = 1)when it really should be 4

PR #673, opened 2023-04-08
Update queue type to avoid build error

ashburnham:
Existing code gives this error

node_modules/langchain/dist/util/async_caller.d.ts:30:22 - error TS2702: 'PQueueMod' only refers to a type, but is being used as a namespace here.
30     protected queue: PQueueMod.default;

This PR fixes it..

### DIFF
--- a/langchain/src/util/async_caller.ts
+++ b/langchain/src/util/async_caller.ts
@@ -32,7 +32,7 @@ export class AsyncCaller {
 
   protected maxRetries: AsyncCallerParams["maxRetries"];
 
-  protected queue: PQueueMod.default;
+  protected queue: PQueueMod;
 
   constructor(params: AsyncCallerParams) {
     this.maxConcurrency = params.maxConcurrency ?? Infinity;
--- a/langchain/src/util/async_caller.ts
+++ b/langchain/src/util/async_caller.ts
@@ -32,7 +32,7 @@ export class AsyncCaller {
 
   protected maxRetries: AsyncCallerParams["maxRetries"];
 
-  protected queue: PQueueMod;
+  private queue: typeof import("p-queue")["default"]["prototype"];
 
   constructor(params: AsyncCallerParams) {
     this.maxConcurrency = params.maxConcurrency ?? Infinity;


nfcampos:
@ashburnham see the "Build and check types" CI job result, this appears to make the build fail on CI (and in my laptop). Can you share more about what's different about your environment

ashburnham:
I'm using it in a `commonjs` module (which I had due to other dependencies. My `tsconfig` is as follows:

```
{
  "include": [
    "src/**/*",
  ],
  "compilerOptions": {
    "target": "ES2020",           /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019' or 'ESNEXT'. */
    "module": "commonjs",         /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */
    "outDir": "./dist",           /* Redirect output structure to the directory. */
    "strict": true,              /* Enable all strict type-checking options. */
    "baseUrl": "./src",
    "typeRoots": [
      "node_modules/@types"
    ],                            /* List of folders to include type definitions from. */
    "types": [
      "node",
      "jest"
    ],
    "lib":  [
      "ES2019",
      "dom",
    ], /* Type declaration files to be included in compilation. */
    "esModuleInterop": true,      /* Enables emit interoperability between CommonJS and ES Modules via creation of namespace objects for all imports. Implies 'allowSyntheticDefaultImports'. */
    "inlineSourceMap": true,
    "resolveJsonModule": true,
    "moduleResolution": "node10",
    "allowJs": true
  }
}
```

It took me a while to get it does so that this is the only error left, and it seems it was as my PR suggests only 3 days ago... so I was hoping it could be changed back. Alternatively, any other workaround to make it work in a commonjs project would be very helpful! 🙏🙏🙏🙏 

p.s. I've been followinf [this issue](https://github.com/hwchase17/langchainjs/issues/185), which is related, but none of it works for me.

ashburnham:
p.s. I'm going `skipLibCheck ` on `langchain` for now, but that's not a long-term solution...

langchain/src/util/async_caller.ts
```diff
@@ -32,7 +32,7 @@ export class AsyncCaller {
 
   protected maxRetries: AsyncCallerParams["maxRetries"];
 
-  protected queue: PQueueMod.default;
+  protected queue: PQueueMod;
```
nfcampos, line 35:
```suggestion
  private queue: typeof import("p-queue")["default"]["prototype"];
```

nfcampos:
I've committed an alternative here which I think fixes it, but in testing your tsconfig it seems to be invalid, eg. `"moduleResolution": "node10",` doesn't exist according to the docs. I'd usually recommend adding `skipLibCheck` on your projects, you don't gain anything in type checking your dependencies (ie. it doesn't make your own application any more type safe), which were most likely designed for a different tsconfig than you are using, hence the type "errors" you then see

ashburnham (commented):
This looks like a good solution! Thanks!

nfcampos:
This was released on 0.0.51, please try again

PR #675, opened 2023-04-08
feat(vectorstore): implement opensearch vector store

igorshapiro:
# Summary

Implement OpenSearch vector Store.
The implementation only supports the kNN approximate search


nfcampos:
Thanks!

PR #677, opened 2023-04-08
Use signal option in axios adapter if passed in

nfcampos:


langchain/src/util/axios-fetch-adapter.js
```diff
@@ -315,6 +315,10 @@ function createRequest(config) {
   if (config.timeout && config.timeout > 0) {
     options.signal = AbortSignal.timeout(config.timeout);
   }
+  if (config.signal) {
+    // this overrides the timeout signal if both are set
```
iosifboanca, line 319: thanks!!!

PR #682, opened 2023-04-08
Nc/granular entrypoints

nfcampos:


langchain/scripts/create-entrypoints.js
```diff
@@ -1,13 +1,21 @@
-import fs from "fs";
+import * as fs from "fs";
+import * as path from "path";
 
 const entrypoints = {
```
nfcampos, line 4: @hwchase17 this is where you can see all the entrypoints

PR #688, opened 2023-04-09
Provide instructions for how to run examples

jameshfisher:
I hit a few stumbling blocks when trying to run these. I've tried to make it clearer.


agola11 (approved):
Thanks!

PR #693, opened 2023-04-09
Merge CallbackManagers when passed through

agola11:
I think it will be really hard to get rid of self.callback_manager / this.callbackManager the way things are set up now, because sometimes we only want certain callbacks on certain objects (like we in the case of the ConversationalRetrievalChain , we want the llm associated with the QA chain to hit the on_llm_new_token callback and not the question gen one).

### DIFF
--- a/langchain/src/callbacks/base.ts
+++ b/langchain/src/callbacks/base.ts
@@ -510,6 +510,12 @@ export class CallbackManager extends BaseCallbackManager {
     return manager;
   }
 
+  copy(): CallbackManager {
+    const manager = new CallbackManager(this._parentRunId);
+    manager.setHandlers(this.handlers);
+    return manager;
+  }
+
   static fromHandlers(handlers: BaseCallbackHandlerMethods) {
     class Handler extends BaseCallbackHandler {
       alwaysVerbose = true;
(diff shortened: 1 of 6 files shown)


langchain/src/callbacks/base.ts
```diff
@@ -510,6 +510,12 @@ export class CallbackManager extends BaseCallbackManager {
     return manager;
   }
 
+  copy(): CallbackManager {
+    const manager = new CallbackManager(this._parentRunId);
+    manager.setHandlers(this.handlers);
```
nfcampos, line 515: This assumes handlers are stateless which isn't true eg of the tracer. Is the tracer "stateless enough" now that this can never be a problem?

langchain/src/callbacks/base.ts
```diff
@@ -510,6 +510,12 @@ export class CallbackManager extends BaseCallbackManager {
     return manager;
   }
 
+  copy(): CallbackManager {
```
nfcampos, line 513: Might as well accept here as an argument another list of handlers to add?

nfcampos (approved):
lgtm otherwise

langchain/src/callbacks/base.ts
agola11: yup, this should not be a problem, no more stack. it uses a map in instead which will prevent the "collisions" problem

langchain/src/callbacks/base.ts
agola11: hmm yeah good point

PR #695, opened 2023-04-09
Add docs for cancelling an LLM call

nfcampos:


nfcampos:
Merge after #47 (and update example)

PR #698, opened 2023-04-10
Change to PDFLoader to us Legacy pdf.js  Accepts Polyfills

hillis:
pdfjs-dist is causing an error with message The browser/environment lacks native support for critical functionality used by the PDF.js library (e.g. `Path2D` and/or `ReadableStream`); please use a `legacy`-build instead.

According to pdf.js legacy is the only way to run inside node.
https://github.com/mozilla/pdf.js/wiki/Frequently-Asked-Questions#faq-support


LarchLiu:
We can use PDFLoader like this
```
new PDFLoader(path, {
    pdfjs: () =>
      import('pdfjs-dist/legacy/build/pdf.js').then((mod) => mod.default)
  })
```

hillis:
> We can use PDFLoader like this
> 
> ```
> new PDFLoader(path, {
>     pdfjs: () =>
>       import('pdfjs-dist/legacy/build/pdf.js').then((mod) => mod.default)
>   })
> ```

Thank you that is much better.

PR #703, opened 2023-04-10
Add .nvmrc with minimum supported Node version

jameshfisher:
I had some mysterious failures with examples due to using Node 16. With an nvmrc to auto-switch to Node 18, I would have avoided this mistake.


nfcampos:
Thanks! Should this instead be `node` to default to the latest (19)?

jameshfisher:
@nfcampos yes, if you prefer! I wish `.nvmrc` would let you specify e.g. `v18+`, but I don't think it's possible. But I've always set `.nvmrc` to the minimum supported Node version (`18` in this project), so that developers don't accidentally assume newer features.

nfcampos:
Cool lets leave as is

PR #707, opened 2023-04-10
Adds JavascriptTextSplitter and GenericCodeSplitter

kryptoklob:
Adds an abstract splitter `JavascriptBaseTextSplitter` which splits on regex, implemented by:

`JavascriptClassTextSplitter` - which splits right before class and interface declarations (and any comments for them)
`JavascriptFunctionTextSplitter` - which splits right before function and constructor declarations (and any comments for them)
`JavascriptTextSplitter` - combines the above two

Adds a `GenericCodeTextSplitter` which splits on provided keywords, and any comments that immediately precede them.

These classes split text *right before* the keywords they're looking for, where *right before* means "immediately before the first comment that is immediately line-adjacent to the keyword in question"

### DIFF
--- a/langchain/src/text_splitter.ts
+++ b/langchain/src/text_splitter.ts
@@ -449,7 +449,7 @@ export abstract class BaseCodeSplitter extends TextSplitter {
   }
 }
 
-export class GenericCodeSplitter extends BaseCodeSplitter {
+export class GenericCodeTextSplitter extends BaseCodeSplitter {
   constructor(keywords: string[], fields?: Partial<TextSplitterParams>) {
     super(fields);
 
--- a/langchain/src/text_splitter.ts
+++ b/langchain/src/text_splitter.ts
@@ -407,8 +407,6 @@ export abstract class BaseCodeSplitter extends TextSplitter {
 
     // find our first match
     while (match && match[0]) {
-      console.log("match", match);
-
       // find and remove our first match - this is the start of our string
       // it should be index 0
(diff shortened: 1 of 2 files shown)


langchain/src/text_splitter.ts
```diff
@@ -330,3 +330,134 @@ export class MarkdownTextSplitter
...
+export class JavascriptClassTextSplitter extends JavascriptBaseTextSplitter {
+  regex = /(((.*\/\/.*\n)*)|(.*\/\*.*\n))(.*((class)|(interface)).*)/;
+}
+export class JavascriptFunctionTextSplitter extends JavascriptBaseTextSplitter {
+  regex = /(((.*\/\/.*\n)*)|(.*\/\*.*\n))(.*((function)|(constructor)).*)/;
+}
+
+export class JavascriptTextSplitter extends JavascriptBaseTextSplitter {
+  regex =
+    /(((.*\/\/.*\n)*)|(.*\/\*.*\n))(.*((function)|(constructor)|(class)|(interface)).*)/;
```
nfcampos, line 394: I'd consider adding `type` here, and maybe `export` too?

nfcampos (commented):
looks good, two comments

langchain/src/text_splitter.ts
TimPietrusky: I would also add these:

* `default`
* `async function`

And I'm not sure about `constructor`, as that is already part of a `class` and there is usually not so much code between the class and the `constructor`.

PR #713, opened 2023-04-10
Clarify that examples require installing example dependencies

jameshfisher:
Also I don't think README.md should be under src/

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -6,22 +6,32 @@ This folder contains examples of how to use LangChain.
 
 What you'll usually want to do.
 
-First, build langchain. From the repository root, run:
+First, build LangChain. From the repository root, run:
 
 ```sh
 yarn
 yarn build
 ```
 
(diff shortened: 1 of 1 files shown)


examples/README.md
```diff
@@ -6,22 +6,32 @@ This folder contains examples of how to use LangChain.
...
 yarn
 yarn build
 ```
 
-Most examples require API keys. Run `cp .env.example .env`, then edit `.env` with your API keys.
+Then switch to this `examples/` directory, and run:
 
-Then from the `examples/` directory, run:
+```sh
+yarn  # Install dependencies for examples
```
nfcampos, line 19: This is a yarn monorepo, therefore it shouldn't be necessary to install dependencies in a subdirectory, running yarn at the root installs all

jameshfisher: Hmm, you're right; I wonder why that didn't work for me 🤔

PR #719, opened 2023-04-10
Instructions for running Pinecone example

jameshfisher:
Using default Pinecone indexes results in error due to dimension mismatch.


nfcampos:
@jameshfisher Thanks for all these improvements!

PR #728, opened 2023-04-11
feat(document_loader): Add optional selector

magick93:
This adds an optional selector parameter to the `load()` function. The type is then specified in `langchain/src/document_loaders/cheerio_web_base.ts`


langchain/src/document_loaders/base.ts
```diff
@@ -10,7 +10,7 @@ export interface DocumentLoader {
 }
 
 export abstract class BaseDocumentLoader implements DocumentLoader {
-  abstract load(): Promise<Document[]>;
+  abstract load(selector?: unknown): Promise<Document[]>;
```
nfcampos, line 13: Thanks for this, I think the selector should be moved to a constructor argument on CheerioWebBaseLoader

langchain/src/document_loaders/cheerio_web_base.ts
```diff
@@ -47,9 +47,9 @@ export class CheerioWebBaseLoader
     );
   }
 
-  async load(): Promise<Document[]> {
+  async load(selector?: SelectorType): Promise<Document[]> {
     const $ = await this.scrape();
-    const text = $("body").text();
+    const text = $(selector).text() ?? $("body").text();
```
nfcampos, line 52: This should instead be `$(selector ?? "body").text();`

nfcampos (commented):
two comments

langchain/src/document_loaders/base.ts
arronKler: I do think so, its weird to add a parameter 'selector' in the abstract BaseDocumentLoader, especially you do implementation in other tools. Its not an universal parameter

langchain/src/document_loaders/cheerio_web_base.ts
magick93: Nice!

nfcampos:
Hi @magick93 can you give me permission to push to this branch (there should be a checkbox you can tick here in the PR page) so I can fix the conflicts, and move the selector the fields object

magick93:
@nfcampos

I couldnt find the checkbox - but merge conflicts should be resolved now.

PR #748, opened 2023-04-11
Adds `returnDirect` to DynamicTool

francofantini:
DynamicTool is missing the option to return direct the output of the tool.


nfcampos:
Thanks!

PR #755, opened 2023-04-12
Improve pgvector docs

ankane:
Hi, thanks for adding pgvector to the LangChain ecosystem!

The docs currently instruct users to run `CREATE INDEX` right after creating the table. However, this will lead to poor recall. I think it'd be better for users to use exact nearest neighbor search (for perfect recall) and only add an index when querying becomes slow (and the table has data).

More info: https://github.com/pgvector/pgvector#indexing


nfcampos:
Thank you! I was about to do this myself

ankane:
Thanks!

PR #766, opened 2023-04-12
serpapi results coming back in another region/languge

jacobrosenthal:
closes https://github.com/hwchase17/langchainjs/issues/765

Screenshot from 2023-04-12 11-13-26

I contacted support because just going back to a search from last weeks It was all in english (Im in Phoenix USA)
So I dont think we changed anything in the api. 
TLDR is unless you set location (and possibly hl and gl) they dont seem to guarantee anything

Just setting location fixed it for me for now.. It seems like alot to add the hg and gl as well, maybe we can get away without it

Updated all examples with a default location. their docs mention Austin so I just used that

>The first thing I notice is that you have not used any location parameters. For consistent language/location results, you need to specify the location, hl and gl parameters. Without these, your search can take the proxy's location, which may deliver unexpected results. We recommend always using the most specific location parameters possible in order to simulate a real user's search.
I understand that you have not experienced this before, and your parameters may not have changed. However, by not using these parameters you are not exercising control over the location for which the results are delivered.
Let me know if you continue to experience this after including location parameters.

> For best results I would still recommend using the hl and gl as well, however you may not notice the difference, I can't say for sure.

### DIFF
--- a/docs/docs/getting-started/guide-chat.mdx
+++ b/docs/docs/getting-started/guide-chat.mdx
@@ -281,6 +281,8 @@ And finally, we can use the AgentExecutor to run an agent:
 const tools = [
   new SerpAPI(process.env.SERPAPI_API_KEY, {
     location: "Austin,Texas,United States",
+    hl: "en",
+    gl: "us",
   }),
 ];
 // Create the agent from the chat model and the tools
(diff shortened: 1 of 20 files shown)


jacobrosenthal:
Nope still happening. Let me try to set those other 2

>langchain-examples:start: Перевод "emolument" на русский. Сущ. вознаграждение n.
langchain-examples:start:       "observation": "«Новый агент Макгайвер» — американский приключенческий телесериал, ремейк «Секретного агента Макгайвера», транслировавшегося на ABC с 1985 по 1992 год. Премьера сериала состоялась 23 сентября 2016 года на телеканале CBS."

nfcampos:
Thanks!

PR #782, opened 2023-04-13
feat: wikipedia loader - broken

jacobrosenthal:
@nfcampos 

This is a half ass paste, but `extractSections` was proven out in some code I have elsewhere so I know that function is good

### DIFF
--- a/examples/src/document_loaders/wikipedia.ts
+++ b/examples/src/document_loaders/wikipedia.ts
@@ -1,0 +1,9 @@
+import { WikipediaLoader } from "langchain/document_loaders/web/wikipedia";
+
+export const run = async () => {
+  const loader = new WikipediaLoader(
+    "https://en.wikipedia.org/wiki/2020_Summer_Olympics"
+  );
+  const docs = await loader.load();
+  console.log({ docs });
+};
--- a/langchain/src/document_loaders/web/wikipedia.ts
+++ b/langchain/src/document_loaders/web/wikipedia.ts
@@ -1,0 +1,127 @@
+import wikipedia from "wikipedia";
+import { Document } from "../../document.js";
+import { BaseDocumentLoader } from "document_loaders/base.js";
+
+const discardCategories = [
+  "See also",
+  "References",
+  "External links",
+  "Further reading",
+  "Footnotes",
+  "Bibliography",
+  "Sources",
+  "Citations",
+  "Literature",
+  "Footnotes",
+  "Notes and references",
+  "Photo gallery",
+  "Works cited",
+  "Photos",
+  "Gallery",
+  "Notes",
+  "References and sources",
+  "References and notes",
+];
+
+export class WikipediaLoader extends BaseDocumentLoader {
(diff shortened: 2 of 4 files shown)


examples/src/document_loaders/wikipedia.ts
```diff
@@ -0,0 +1,9 @@
+import { WikipediaLoader } from "langchain/document_loaders/web/wikipedia";
+
+export const run = async () => {
```
dqbd, line 3: A documentation entry would be great!

langchain/src/document_loaders/web/wikipedia.ts
```diff
@@ -0,0 +1,131 @@
+import wiki, { Page } from "wikipedia";
+import { Document } from "../../document.js";
+import { BaseDocumentLoader } from "../base.js";
+
+const DISCARDCATEGORIES = [
```
dqbd, line 5: I think these should be moved to examples instead, as this assumes only English locale will be loaded.

langchain/src/document_loaders/web/wikipedia.ts
```diff
@@ -0,0 +1,131 @@
...
+  "Gallery",
+  "Notes",
+  "References and sources",
+  "References and notes",
+];
+
+export class WikipediaLoader extends BaseDocumentLoader {
+  discardCategories: string[];
+
+  constructor(public webPath: string, discardCategories?: string[]) {
```
dqbd, line 29: To match the feature parity of `wikipedia` loader in Langchain, it would make sense to add an option for locale via `wiki.setLang(lang)`

langchain/src/document_loaders/web/wikipedia.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  public async load(): Promise<Document[]> {
+    const matchResult = this.webPath.match(/\/wiki\/(.+)/);
+    if (!matchResult) {
+      throw new Error(
+        "Invalid webPath format. Use full url like https://en.wikipedia.org/wiki/2020_Summer_Olympics"
+      );
+    }
+
+    const pageTitle = decodeURIComponent(matchResult[1]);
```
dqbd, line 42: I think perform an actual search instead of decoding the last URL portion is more resilient and matches the behavior of Wikipedia loader.

langchain/src/document_loaders/web/wikipedia.ts
```diff
@@ -0,0 +1,131 @@
...
+          pageContent: section,
+          metadata: { source: this.webPath },
+        })
+    );
+    return docs;
+  }
+
+  // converted from the openai cookbook example
+  // https://github.com/openai/openai-cookbook/blob/main/examples/fine-tuned_qa/olympics-1-collect-data.ipynb
+  private extractSections(
```
dqbd, line 63: Would rather see section splitting being done by a splitter instead of by the loader.

dqbd (changes requested):

Issue #751, opened 2023-04-11
Trying to get `loadQARefineChain` to use the `input_documents` as context


spark-ming:
Just noticed the addition of `loadQARefineChain` in `0.0.53`. Thank you!

However, I currently have the code:

```ts
const refineChain = loadQARefineChain(textDavinci);
const result = await refineChain.call({
  input_documents: docs,
  question,
});
```

The is the result I get from calling

```js
{
  output_text: 'Sorry, I cannot answer the question as the provided context information of "[object Promise]" is not related to pre-pay and does not provide enough information to understand the question. Can you please provide more context or clarify the question?'
}
```

Is the context not being serialised correctly?

As a side, I noticed the result is keyed `output_text` instead of `answer` in this chain.

RohitMidha23:
@spark-ming apologies, I seem to have missed this case. Have raised a PR to account for async functionality. You should be able to use RefineChain in your code post that.

nfcampos:
This has been fixed already in 0.0.54

PR #786, opened 2023-04-13
Fix issue in Refine Chain

RohitMidha23:
Noted in issue #751 that the `format` method is returning a promise. 

Updated the code to take into consideration `async` functionality.


RohitMidha23:
@nfcampos @hwchase17 can you please review?

nfcampos:
I fixed this already here https://github.com/hwchase17/langchainjs/blob/main/langchain/src/chains/combine_docs_chain.ts#L281 Thanks

PR #787, opened 2023-04-14
fix: printSchema in StructuredOutputParser failing when using Zod

justindra:
When using `Zod` with the `StructuredOutputParser`, it always fails as the `instanceof` checks never passes as `instanceof` only works when the class used is the exact same one as the one being compared to. This issue occurs when dealing with monorepos where the same package may have multiple instances.

```ts
    z.object({
      name: z.string().describe('The name'),
      artists: z
        .array(z.string())
        .describe('An array of artists'),
    })
  );
```

The error received is:
```
Error: Cannot read properties of undefined (reading 'typeName')
```

So using the `schema._def.typeName` is a lot more reliable in performing these checks.

### DIFF
--- a/langchain/src/output_parsers/structured.ts
+++ b/langchain/src/output_parsers/structured.ts
@@ -64,7 +64,7 @@ ${Object.entries(shape)
 ${indent}}`;
   }
 
-  throw new Error(`Unsupported type: ${schema._def.innerType.typeName}`);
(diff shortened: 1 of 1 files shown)


nfcampos:
Thanks @justindra agreed, this is a good fix, I think we should just remove the instanceof checks in favour of this?

justindra:
@nfcampos yea, I think we can remove the `instanceof` checks. I'll send through another PR for it.

PR #788, opened 2023-04-14
Remove mongo vector store from docs

nfcampos:


harisvsulaiman:
@nfcampos Why was this removed?

PR #797, opened 2023-04-14
fix: remove instanceof checks in favour of typeName checks

justindra:
Continuing the fix started in #787 by removing `instanceof` checks now that `typeName` checks is in.


nfcampos:
Thanks!

PR #802, opened 2023-04-14
Use Output Parsers for agents

joshsny:
refactors agents to use output parsers


langchain/src/agents/chat/outputParser.ts
```diff
@@ -0,0 +1,33 @@
...
+export class ChatAgentOutputParser extends AgentActionOutputParser {
+  async parse(text: string) {
+    if (text.includes(FINAL_ANSWER_ACTION)) {
+      const parts = text.split(FINAL_ANSWER_ACTION);
+      const output = parts[parts.length - 1].trim();
+      return { returnValues: { output }, log: text } satisfies AgentFinish;
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const [_, action, __] = text.split("```");
```
nfcampos, line 15: I know you're just moving this from elsewhere but this can probably be improved by optionally allowing for code blocks with `json` keyword after the three backticks

langchain/src/agents/initialize.ts
```diff
@@ -1,10 +1,10 @@
+import { BaseLanguageModel } from "../base_language/index.js";
```
nfcampos, line 1: Is this import reordering something we can add to CI / pre-commit hook?

nfcampos (approved):

langchain/src/agents/initialize.ts
joshsny: You can find an implementation of this that we could add as a pre-commit hook [here](https://github.com/microsoft/TypeScript/issues/32818). Do you want me to follow up with that in another PR?

PR #805, opened 2023-04-15
Add missing function execution in streaming example in getting started guide

choonkeat:
followup for https://github.com/hwchase17/langchainjs/pull/691

1. import statement has warning
2. example exits immediately with no output

### DIFF
--- a/examples/src/models/chat/chat_streaming_stdout.ts
+++ b/examples/src/models/chat/chat_streaming_stdout.ts
@@ -14,23 +14,23 @@ const chat = new ChatOpenAI({
 await chat.call([
   new HumanChatMessage("Write me a song about sparkling water."),
 ]);
-  /*
-  Verse 1:
-  Bubbles rise, crisp and clear
-  Refreshing taste that brings us cheer
-  Sparkling water, so light and pure
-  Quenches our thirst, it's always secure
+/*
+Verse 1:
+Bubbles rise, crisp and clear
+Refreshing taste that brings us cheer
+Sparkling water, so light and pure
+Quenches our thirst, it's always secure
 
-  Chorus:
-  Sparkling water, oh how we love
-  Its fizzy bubbles and grace above
-  It's the perfect drink, anytime, anyplace
-  Refreshing as it gives us a taste
+Chorus:
+Sparkling water, oh how we love
+Its fizzy bubbles and grace above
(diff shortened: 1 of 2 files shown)


examples/src/models/chat/chat_streaming_stdout.ts
```diff
@@ -36,3 +36,5 @@ export const run = async () => {
     ...
     */
 };
+
+run();
```
nfcampos, line 40: The examples are meant to be run with the `yarn example` script, therefore this is not needed

nfcampos (commented):
Thanks, one comment

examples/src/models/chat/chat_streaming_stdout.ts
choonkeat: I see. Is my understanding correct that
- the files are meant to be executed through `yarn example` and thus should not contain `run()`
- the content of these files, however, are embedded into docs, e.g. https://js.langchain.com/docs/getting-started/guide-chat#streaming which would not work without `run()`

Do we prefer
1. to keep `run()` absent from the file
2. add a `// run()` along with a comment to uncomment it (I can't come up with a good enough phrasing that feels right for the file & docs)
3. insert the `run()` at docs/docs/getting-started/guide-chat.mdx and docs/docs/getting-started/guide-llm.mdx instead, e.g.

    ```diff
    - <CodeBlock language="typescript">{Example}</CodeBlock>
    + <CodeBlock language="typescript">{Example}
    + run()</CodeBlock>
    ```

(3) perhaps?

nfcampos: That's a good point. I think my preference would actually be to remove these run() functions from all the examples, as top-level await is allowed in ESM modules we can just run them without that. But that's a larger change

nfcampos: FYI I've opened this PR https://github.com/hwchase17/langchainjs/pull/841 which makes it possible to declare examples without `function run()`

choonkeat: I've added a commit to make use of your work in #841

choonkeat: 🤦‍♂️ fixed formatting

PR #812, opened 2023-04-15
Export weaviate store

KadoBOT:


nfcampos:
Hi, as written at the top of that file, those `index` entrypoints are deprecated and will be removed in a future version, therefore we won't be adding any more exports to them. You can import Weaviate from `import { WeaviateStore } from "langchain/vectorstores/weaviate";`

PR #813, opened 2023-04-15
Zod enum support

ashburnham:
- Added enum support for Zod schemas.


langchain/src/output_parsers/structured.ts
```diff
@@ -24,6 +23,9 @@ function printSchema(schema: z.ZodTypeAny, depth = 0): string {
   if (schema instanceof z.ZodDate || schema._def.typeName === "ZodDate") {
     return "date";
   }
+  if (schema instanceof z.ZodEnum) {
```
nfcampos, line 26:
```suggestion
  if (schema instanceof z.ZodEnum || schema._def.typeName === "ZodEnum") {
```

PR #821, opened 2023-04-15
Lambda Tool improvements

jasondotparse:
### Summary
1. Added the tool to `langchain/src/tools/index.ts` so that it can be imported the same was as shown in the [docs](https://js.langchain.com/docs/modules/agents/tools/lambda_agent)
2. Made `region`, `accessKeyId`, and `secretAccessKey` optional, because I received feedback that some people like to just manage those by keeping them in environment variables, since it is easy to do that via the AWS CLI tool.

### Todo:
After this is merged, some people might be confused as to why the S3 File Loader class doesn't say anything about providing a region/accessKey/secretAccessKey. So, I'll have to make a small update to that class so that it matches this one, along with its docs.


nfcampos:
Hi, as written at the top of that file, those index entrypoints are deprecated and will be removed in a future version, therefore we won't be adding any more exports to them.

langchain/src/tools/index.ts
```diff
@@ -23,3 +23,4 @@ export { VectorStoreQATool } from "./vectorstore.js";
 export { ZapierNLARunAction, ZapierNLAWrapper } from "./zapier.js";
 export { Serper } from "./serper.js";
 export { AIPluginTool } from "./aiplugin.js";
+export { AWSLambda } from "./aws_lambda.js";
```
nfcampos, line 26: Remove this

nfcampos:
Thanks!

PR #826, opened 2023-04-16
Suggestion on how to make callback managers more type safe and harder to misuse

nfcampos:
- A callback manager without a currentRunId does not offer non-start methods 
- A callback manager cannot be mutated to add a currentRunId, to do so create a new specialised instance for that run, by calling the respective handleStart method
- An unbound callback manager does not offer getChild, as you shouldn't be able to create a child manager without before calling start

### DIFF
--- a/langchain/src/agents/executor.ts
+++ b/langchain/src/agents/executor.ts
@@ -9,7 +9,7 @@ import {
   AgentStep,
   ChainValues,
 } from "../schema/index.js";
-import { CallbackManager } from "../callbacks/index.js";
+import { CallbackManagerForChainRun } from "../callbacks/manager.js";
 
 interface AgentExecutorInput extends ChainInputs {
   agent: BaseSingleActionAgent | BaseMultiActionAgent;
@@ -69,7 +69,7 @@ export class AgentExecutor extends BaseChain {
 
   async _call(
     inputs: ChainValues,
-    callbackManager?: CallbackManager
+    runManager?: CallbackManagerForChainRun
   ): Promise<ChainValues> {
     const toolsByName = Object.fromEntries(
       this.tools.map((t) => [t.name.toLowerCase(), t])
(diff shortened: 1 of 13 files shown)


langchain/src/agents/executor.ts
```diff
@@ -84,17 +84,15 @@ export class AgentExecutor extends BaseChain {
       if (this.returnIntermediateSteps) {
         return { ...returnValues, intermediateSteps: steps, ...additional };
       }
-      await this.configureCallbackManager(callbackManager)?.handleAgentEnd(
```
nfcampos, line 87: this 2nd configure wasn't needed as it was done in the Chain.call method too

langchain/src/callbacks/manager.ts
```diff
@@ -7,7 +7,13 @@ import {
 } from "../schema/index.js";
```
nfcampos, line 7: Git makes this diff quite unreadable, I'd recommend just reading the new file, instead of the diff

langchain/src/chains/base.ts
```diff
@@ -77,29 +77,24 @@ export abstract class BaseChain extends BaseLangChain implements ChainInputs {
...
-    if (callbackManager) {
-      callbackManager.setCurrentRunId(localCallbackManager?.currentRunId);
-    }
+    const runManager = await this.configureCallbackManager(
+      callbackManager
+    )?.handleChainStart({ name: this._chainType() }, fullValues);
     let outputValues;
     try {
-      outputValues = await this._call(fullValues, callbackManager);
+      outputValues = await this._call(fullValues, runManager);
```
agola11, line 92: This is going to lead to issues. `configureCallbackManager` merges the passed-in `CallbackManager` with `this.callbackManager`. This merged manager should not be passed into the `_call/_generate` methods... there is a potential these methods call `getChild` when calling `tools`, other `chains`, `llms` etc. 

I think one solution for this is to keep a separate list of `inheritedHandlers` that get copied when calling `getChild`. This list can be set by the `configureCallbackManager` method

agola11 (commented):
Really like this idea, but one main comment

PR #827, opened 2023-04-16
feat: Add minSimilarityScore to VectorStore.similaritySearch

mayt:
Allow users to add a cutoff score to include for similarity search. 
Update `vector_db_qa` chain to support this new parameter

### DIFF
--- a/docs/docs/modules/indexes/vector_stores/index.mdx
+++ b/docs/docs/modules/indexes/vector_stores/index.mdx
@@ -24,7 +24,8 @@ interface VectorStore {
   similaritySearch(
     query: string,
     k?: number,
-    filter?: object | undefined
+    filter?: object,
+    minSimilarityScore?: number
   ): Promise<Document[]>;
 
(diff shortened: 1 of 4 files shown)


martinseanhunt:
Nice, you beat me to it! I'd implemented this functionality a week or so ago on a project I'm working on and was just going to write a test and open a PR today, glad I checked here first ;)

jacoblee93:
Sorry for the delayed review!

Similarity scores can differ drastically between different vector store implementations, so I worry this would be misleading if it were built into the base vector store implementation with unit tests like the ones you've added - I think I'd rather keep the abstraction one level higher in the `VectorDBQAChain` and have it call `similaritySearchWithScore` instead.

Does that work for you?

jacoblee93:
These chains have been superseded by the `RetrievalQAChains` - we could look at adding support there in the future. Going to close this for now.

mayt:
Thanks for the update.

PR #830, opened 2023-04-16
feat: serpapi returns list of links

jacobrosenthal:
Not sure if back compat matters but this pairs with webbrowser really well

### DIFF
--- a/langchain/src/tools/serpapi.ts
+++ b/langchain/src/tools/serpapi.ts
@@ -347,28 +347,35 @@ export class SerpAPI extends Tool {
       throw new Error(`Got error from serpAPI: ${res.error}`);
     }
 
+    let links = "";
+    if (res.organic_results) {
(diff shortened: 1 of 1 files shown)


nfcampos:
Can we put this behind a constructor arg? That way it will be backwards compatible

jacobrosenthal:
Yeah whatever you decide. I realized Ive also never seen any of those other fields in my searches, so the if case that always hits is the last, which give 1 link.

So this way if it falls through you get all links as well

jacobrosenthal:
Actually I think the returning n links always is a proper fix, the input args allow you to specify how many links you want back, so the code wasnt respecting the inputs before.

bracesproul:
@jacobrosenthal going to close due to merge conflicts. If you'd still like to get this merged in, please open off of an up-to-date main branch, and I'll review. Thank you!

Issue #475, opened 2023-03-26
SimpleSequentialChain


danny-waite:
How to combine chains, ie. take the output of prompt A and feed it as an input into prompt B?

The python version has the following:

https://python.langchain.com/en/latest/modules/chains/generic/sequential_chains.html

Am I missing something?

danny-waite:
hey @nfcampos I can familiarise myself with the python code and give this a shot towards the end of the week if that helps?

nfcampos:
Yes thanks @danny-waite Any questions let me know!

Potrock:
Happy to take this on if you don't have time @danny-waite :)

kwkr:
Is anyone working on this @danny-waite, @nfcampos ?

danny-waite:
Hey @kwkr I was trying to get around to it, but didn't have chance as yet, so feel free to jump on it.  I'm happy to help if necessary

kwkr:
> Hey @kwkr I was trying to get around to it, but didn't have chance as yet, so feel free to jump on it. I'm happy to help if necessary

Thanks! I will start working on it. Could you assign it to me @nfcampos ?

nfcampos:
@kwkr assigned, thanks!

PR #834, opened 2023-04-16
475/simple sequential chain

kwkr:
addresses #475 

Added SimpleSequentialChain implementation based on the one from the [python version](https://python.langchain.com/en/latest/_modules/langchain/chains/sequential.html#SimpleSequentialChain)

This is my first PR here so please excuse me if I messed some conventions up. The functionality is there but due to some differences between Python and TS classes, I wasn't sure how should I proceed with the validation that is included in the Python version.

I also added the some example docs site for this specific class as it's done for the `LLM`.

Few things to consider:
1. I added both integration and unit tests for the implementation. I'm not sure if both of them are required because some of them cover the same scenario (i.e. joining two chains together).
2. The original implementation validates the chains before creating the SimpleSequentialChain object i.e. it checks whether there is a single input and output for each chain. In the case of the TS implementation, the property `outputKey` is not available on the `BaseChain` but as a part of some chains (for example `LLMChain`). Because I used `BaseChain` as a type of the `chains` that are passed to the `SimpleSequentialChain` the way to check for this would be to do some casting and then check if the property `outputKey` is available on each chain. I would appreciate some input on how I should proceed with this.
3. The original implementation assumes that both input and output are strings. Should this condition be also preserved when validating the chains?

I would appreciate some hints on this @nfcampos. Thanks in advance for any input!

### DIFF
--- a/langchain/src/chains/simple_sequential_chain.ts
+++ b/langchain/src/chains/simple_sequential_chain.ts
@@ -1,0 +1,119 @@
+import { BaseChain, ChainInputs } from "./base.js";
+import { ChainValues } from "../schema/index.js";
+import {
+  SerializedBaseChain,
+  SerializedSimpleSequentialChain,
+} from "./serde.js";
+
+export interface SimpleSequentialChainInput extends ChainInputs {
+  /** Array of chains to run as a sequence. The chains are run in order they appear in the array. */
+  chains: Array<BaseChain>;
+  /** Whether or not to trim the intermediate outputs. */
+  trimOutputs?: boolean;
+}
+
+/**
+ * Simple chain where a single string output of one chain is fed directly into the next.
+ * @augments BaseChain
+ * @augments SimpleSequentialChainInput
+ *
+ * @example
+ * ```ts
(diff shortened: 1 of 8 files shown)


langchain/src/chains/simple_sequential_chain.ts
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  async _call(values: ChainValues): Promise<ChainValues> {
+    let input: string = values[this.inputKey];
+    for (const chain of this.chains) {
+      input = await chain.run(input);
+      if (this.trimOutputs) {
+        input = input.trim();
+      }
+      await this.callbackManager.handleText(input, this.verbose);
```
nfcampos, line 93: @agola11 (not to do this in PR) but thoughts on adding a new callback method called `handleChainIntermediateResult` could be used in lots of places and might be a good answer for some things we've been discussing?

nfcampos:
This is great @kwkr I'm adding `outputKeys` to BaseChain here #843 once that's in you can rebase and finish the validation logic

nfcampos (approved):
This looks great, thanks!

nfcampos (approved):

PR #836, opened 2023-04-17
docs: Improved getting-started guide

erkhem42:
- Added package installation step for `serpapi`


erkhem42:
> @erkhem42 Please undo the change to Calculator import, it does work if you update to the latest version

Ok, done

PR #838, opened 2023-04-17
Update s3 loader so that you can provide credentials in class constructor

jasondotparse:
## Summary 
This change allows the access key / secret access key / region of your S3 bucket to be provided in the constructor of `S3Loader`. This means that you do not need to run `aws configure` before using the class, which was an item that had caused some confusion. 

Also updated the documentation to reflect this change.


nfcampos:
Thanks!

PR #846, opened 2023-04-17
Proposal: return parse error to agent

francofantini:
This PR tries to solve an issue with Agents that causes the agent to return an output without an Action or with a misformatted one.

The changes in this PR introduce a new type of action for `AgentStep`: `AgentError`.

Please let me know if this make sense and if you have suggestions or comments.


jacoblee93:
Thanks for this - we merged something similar to this in a later PR!

PR #847, opened 2023-04-17
Use lite version of @dqbd/tiktoken

dqbd:
Use the lite version of @dqbd/tiktoken, fetching the JSON mappings in runtime

- [ ] Switch to cached ranks instead to avoid pummeling OpenAI Azure CDNs etc.
- [ ] Consider importing full blown WASM when not used in edge runtimes for improved performance


nfcampos:
Closing in favour of #1239

PR #853, opened 2023-04-17
updating import links.

sachio222:
links were incorrect:

import { ChatOpenAI } from 'langchain/chat_models/openai'; raised errors in execution, until replaced with import { ChatOpenAI } from 'langchain/chat_models';


sachio222:
making sure i'm on latest langchain, found breaking-changes notice on installation page. closing.

PR #857, opened 2023-04-17
feat(Make textSplitter  of WebBrowser Tool  configurable):

lxfater:
Make the TextSplitter in WebBrowser Tool configurable

### DIFF
--- a/langchain/src/tools/webbrowser.ts
+++ b/langchain/src/tools/webbrowser.ts
@@ -2,7 +2,7 @@ import axiosMod, { AxiosRequestConfig } from "axios";
 import { isNode } from "browser-or-node";
 import * as cheerio from "cheerio";
 import { BaseLanguageModel } from "../base_language/index.js";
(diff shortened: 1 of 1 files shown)


jacobrosenthal:
Greetings, just another contrib here, but i wrote webbrowser so i thought id chime in.

Maybe just take the constructed RecursiveCharacterTextSplitter? or better yet, lets take the base class TextSplitter. Then people can pass whatever TextSplitter type they want?

ideally thow a test in showing usage of passing in some html (theres some html in a fixtures folder) you can split on that with the default and something else and see it change

lxfater:
Great idea, this makes it more flexible. However, I feel that most of the time, agents cannot choose a very good delimiter, so the current text separator is still the best choice. But I think the best solution is to allow users to customize it and make the original setting the default value. @jacobrosenthal

nfcampos:
Hi, closing this in favour of #1748 which accepts an instance of textsplitter instead

PR #859, opened 2023-04-17
Fix RegexParser including double slashes in prompt

jameshfisher:
RegexParser was including the /regex/ slashes in its output, but those are already included when converting a RegExp to string. Therefore prompts looked like //0-9//, which is not a common regex notation. Models get confused and include literal // in their responses, which can be seen in the examples.

Fix: remove slashes from RegexParser output, and rely on those from converting RegExp to string.

Also included an example of how to use RegexParser.

Also re-ran examples and pasted output in comments.

### DIFF
--- a/examples/src/prompts/combining_parser.ts
+++ b/examples/src/prompts/combining_parser.ts
@@ -42,12 +42,14 @@ export const run = async () => {
 
   ```json
   {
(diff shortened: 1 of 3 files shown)


nfcampos:
@jameshfisher Thanks!

PR #864, opened 2023-04-18
fix: fix document loader examples

jtyjty99999:
hi
document loader refer has been changed to "langchain/document_loaders" 
but examples are not, so fix it.


nfcampos:
Hi, that's actually the old deprecated import, any use going forward should use the import paths specific to each loader (to avoid pulling in dependencies you're not using)

jtyjty99999:
@nfcampos but docs have not been updated https://js.langchain.com/docs/modules/indexes/document_loaders/examples/file_loaders/directory

Issue #720, opened 2023-04-10
Where is the SequentialChain exported from ?


vpaul18:
Hi !

I'm trying to convert the python implementation of Langchain to JS, yet I cannot seem to find where the `SequentialChain` is exported from in this repo. Or is it not supported yet?

Thanks!

hyuraku:
https://github.com/hwchase17/langchainjs/pull/834 will add `SequentialChain`

nfcampos:
This was just merged today #834 will release in a few hours

nfcampos:
Ah yes quite right @kwkr my mistake. If you want to contribute that it would be great, but no pressure

nfcampos:
Yea just start, we can use this issue, I've reopened it

vpaul18:
@nfcampos @kwkr Thanks guys!

PR #868, opened 2023-04-18
720/sequential chain

kwkr:
Added initial implementation of `SequentialChain` to address #720 

There are a few problems because of the current implementation that should be addressed before finalizing it:

1. The straightforward serialization/deserialization works but there are some problems with more complex scenarios (for example serialization/deserialization of the chain from example code). The example code uses custom input/output keys that are required for the proper passing of the data between inner chains. The problem is that for example for LLMChain the `outputKey` is not serialized which causes an error after trying to deserialize it and run it afterward. You can see some sample code in the provided integration tests (I skipped the one because it wasn't working).

2. Chains' validation logic requires knowing the memory keys that will be used in the chain beforehand. You can see the Python logic [here](https://python.langchain.com/en/latest/_modules/langchain/chains/sequential.html#SequentialChain) in the method `validate_chains`. In the current implementation, the validation logic is done at the end of the constructor call. The problem is `loadMemoryVariables` is function that returns a Promise which doesn't allow use within a constructor and there is no way to get available keys. The solution would be to either make the keys from the memory available via some getter or move the validation logic to some async static factory method and make the constructor private. Please let me know if you have some other thoughts on this matter.

3. How to proceed when there are 0 chains passed to the `SequentialChain`? Is there some use case when an empty array could be passed?

4. Should the callback manager be invoked for intermediate results within the chain? It's not done in Python implementation but I was wondering whether it might be useful.

Could you provide me some input here @nfcampos ? I don't think I would be able to figure it out on my own. Thanks a lot!


nfcampos:
Thanks a lot for this @kwkr I've finished the remaining questions, will merge this soon

PR #873, opened 2023-04-18
expose checkValidTemplate, TemplateFormat from template.ts

ppramesi:
checkValidTemplate and TemplateFormat should be exposed so other devs can create other types of templates.


nfcampos:
Agreed, thanks

PR #882, opened 2023-04-18
Support passing in and attaching handlers to language models, chains, and tools

agola11:


langchain/src/chains/base.ts
```diff
@@ -89,8 +98,10 @@ export abstract class BaseChain extends BaseLangChain implements ChainInputs {
         fullValues[key] = value;
       }
     }
-    const callbackManager_ = await this.configureCallbackManager(
-      callbackManager
+    const callbackManager_ = await CallbackManager.configure(
+      callbacks,
+      Array.isArray(this.callbacks) ? this.callbacks : this.callbacks?.handlers,
+      { verbose: this.verbose }
```
nfcampos, line 104: Could also pass as an arg here just `this`, instead of picking the properties

langchain/src/chains/base.ts
```diff
@@ -109,7 +120,7 @@ export abstract class BaseChain extends BaseLangChain implements ChainInputs {
     }
     // add the runManager's currentRunId to the outputValues
     outputValues.__runMetadata = runManager
```
nfcampos, line 122: How about renaming this to just __run ?

langchain/src/chains/base.ts
```diff
@@ -119,10 +130,10 @@ export abstract class BaseChain extends BaseLangChain implements ChainInputs {
    */
   async apply(
     inputs: ChainValues[],
-    callbackManagers?: CallbackManager[]
+    callbacks?: CallbackManager[] | BaseCallbackHandler[][]
```
nfcampos, line 133: Oh I hadn't realised before that here we accept an array of callback managers. I can see the usefulness, but someone can accidentally pass an array of handlers here (if not using typescript)?

langchain/src/chat_models/base.ts
```diff
@@ -43,15 +44,17 @@ export abstract class BaseChatModel extends BaseLanguageModel {
...
     const llmOutputs: LLMResult["llmOutput"][] = [];
     const messageStrings: string[] = messages.map((messageList) =>
       getBufferString(messageList)
     );
-    const callbackManager_ = await this.configureCallbackManager(
-      callbackManager
+    const callbackManager_ = await CallbackManager.configure(
+      callbacks,
+      Array.isArray(this.callbacks) ? this.callbacks : this.callbacks?.handlers,
+      { verbose: this.verbose }
```
nfcampos, line 57: Wouldn't you want to pass here this.callbacks too?

langchain/src/llms/base.ts
```diff
@@ -69,10 +70,12 @@ export abstract class BaseLLM extends BaseLanguageModel {
...
     stop?: string[],
-    callbackManager?: CallbackManager
+    callbacks?: CallbackManager | BaseCallbackHandler[]
   ): Promise<LLMResult> {
-    const callbackManager_ = await this.configureCallbackManager(
-      callbackManager
+    const callbackManager_ = await CallbackManager.configure(
+      callbacks,
+      Array.isArray(this.callbacks) ? this.callbacks : this.callbacks?.handlers,
+      { verbose: this.verbose }
```
nfcampos, line 78: Same as above

langchain/src/tools/base.ts
```diff
@@ -3,25 +3,40 @@ import {
...
-    const callbackManager_ = await this.configureCallbackManager(
-      callbackManager
+  async call(
+    arg: string,
+    callbacks?: CallbackManager | BaseCallbackHandler[]
+  ): Promise<string> {
+    const callbackManager_ = await CallbackManager.configure(
+      callbacks,
+      Array.isArray(this.callbacks) ? this.callbacks : this.callbacks?.handlers,
+      { verbose: this.verbose }
```
nfcampos, line 39: Same as above

langchain/src/base_language/index.ts
```diff
@@ -25,51 +21,14 @@ export abstract class BaseLangChain {
...
-        callbackManager_.addHandler(await getTracingCallbackHandler(), true);
-      }
-    }
-    return callbackManager_;
-  }
+  callbacks?: CallbackManager | BaseCallbackHandler[];
 
-  constructor(verbose?: boolean, callbackManager?: CallbackManager) {
+  constructor(
+    verbose?: boolean,
```
nfcampos, line 27: I think this should take an args object with these keys rather than the positional args

nfcampos (commented):
Some small comments, overall lgtm

langchain/src/chains/base.ts
agola11: Yeah tbh, I'm not sure how much this method is used rn. We can restrict it to one callback manager or a single array of handlers?

langchain/src/chains/base.ts
agola11: sure

PR #884, opened 2023-04-19
added ignoreFiles on github document loader

ppramesi:
added ignoreFiles parameter on GithubRepoLoader constructor so it can ignore irrelevant large files (e.g. package-lock.json). It accepts array of strings or RegExp to test path name.


nfcampos:
Thanks, this is great!

PR #886, opened 2023-04-19
feat(tools): fix WebBrowser and AWSLambda export

pftom:
When I see the documentation at https://js.langchain.com/docs/modules/agents/tools/webbrowser, I can't use webbrowser at 0.0.59 version, and I fix it


nfcampos:
This is not a mistake, these tools need to be imported from `"langchain/tools/webbrowser"` and `"langchain/tools/aws_lambda"`, the docs have been updated to reflect that some days ago. The reason for this is they have optional dependencies that we don't want to force on all users.

PR #889, opened 2023-04-19
Add HumanTool class for agent-human interaction

NJordan72:
This pull request introduces a new HumanTool class that extends the base Tool class. The purpose of this class is to allow agents to seek guidance from a human when they are stuck or unsure of what to do next.

Key changes include:

- A new HumanTool class has been added to represent the tool for agent-human interaction.

This enhancement will enable the agent to better handle situations where its knowledge or capabilities are insufficient by allowing it to seek assistance from a human.

### DIFF
--- a/langchain/src/tools/human.ts
+++ b/langchain/src/tools/human.ts
@@ -5,7 +5,7 @@ import { Tool } from "./base.js";
 *    or unsure of what to do next.
 *    @extends Tool
 */
-export class HumanTools extends Tool {
+export class HumanTool extends Tool {
     /**
     * The name of the tool.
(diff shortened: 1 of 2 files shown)


jacoblee93:
Hey I don't quite understand this one - it just echoes back what the agent calls the tool with?

NJordan72:
@jacoblee93 It was ported from the python version

jacoblee93:
Seems like there's some other components in the Python version to make it functional: https://github.com/hwchase17/langchain/blob/master/langchain/tools/human/tool.py

Could you add them and add an integration test?

PR #897, opened 2023-04-19
feat: getDefaultOutputParser fields

jacobrosenthal:
I have an issue with the new outputparser refactors. you cant take in Final Tool name as a variable to getDefaultOutputParser (which is abstract so has to be there)

### DIFF
--- a/langchain/src/agents/agent.ts
+++ b/langchain/src/agents/agent.ts
@@ -17,6 +17,9 @@ import {
   StoppingMethod,
 } from "./types.js";
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type OutputParserArgs = Record<string, any>;
+
 class ParseError extends Error {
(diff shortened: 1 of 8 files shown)


hwchase17 (approved):
conceptually seems great to me!

jacobrosenthal:
The other use case for this is to make it a fixing parser, and give down the llm in the fields

jacobrosenthal:
As a bonus, this allows  putting Format instructions inside the output parser where they probably belong.

also refactoring out allowedTools which is clunky now

You can throw that top commit out, I dont care either way, just wanted to point at it. It needs the `ZeroShotAgent.deserializer` to be fixed if you want to keep it (presuming that doesnt break backcompat or whatever)

nfcampos (approved):

nfcampos:
Thanks!

PR #900, opened 2023-04-19
feat: add verbose and callbackManager to ChainTool

bambrose24:
This brings the ChainTool up to snuff with the base Tool, and allows users of the ChainTool to include a callbackManager.

I'm exploring making a standard callback manager for my project where we are able to log to a database every interaction consistently. So this will let us use this utility tool.


Crizzooo (approved):
LGTM

Seems there are a few other extensions of Tools that are missing the ability to forward these arguments to the core tool?

nfcampos:
Thanks! Yes I think there's still a few other tools that need the same fix done to them

PR #902, opened 2023-04-19
feat: agent make constructScratchPad async

jacobrosenthal:
Ive been carrying this patch for a bit, so it feels worth pushing

Ive done stuff where i reach back out to get something (dynamicContext) or i have an agent with a vectorstore built in where I add and get from the vector store


nfcampos:
Thanks, yea this is a good change

PR #909, opened 2023-04-20
Add `verbose` option to `ConversationalRetrievalQAChain` and question answering chains

yuku:
This change makes it easier to investigate what is going on inside ~~`ChatVectorDBQAChain`~~ `ConversationalRetrievalQAChain`.

### DIFF
--- a/langchain/src/chains/question_answering/load.ts
+++ b/langchain/src/chains/question_answering/load.ts
@@ -25,7 +25,9 @@ interface qaChainParams {
   questionPrompt?: BasePromptTemplate;
   refinePrompt?: BasePromptTemplate;
   type?: string;
+  verbose?: boolean;
 }
+
 export const loadQAChain = (
   llm: BaseLanguageModel,
   params: qaChainParams = {}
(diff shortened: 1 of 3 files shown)


agola11:
Thanks for adding this in @yuku! Great idea. Do you mind making the changes to `ConversationalRetrievalQAChain` as well? We're in the process of deprecating `ChatVectorDBQAChain` in favor of that one. Sorry that this wasn't more clear.

yuku:
@agola11 Thanks for the feedback. I pushed two commits:

* Add `@deprecated` comment to `ChatVectorDBQAChain`
* Add `verbose` option to `ConversationalRetrievalQAChain.fromLLM`

langchain/src/chains/question_answering/load.ts
```diff
@@ -25,7 +25,9 @@ interface qaChainParams {
   questionPrompt?: BasePromptTemplate;
   refinePrompt?: BasePromptTemplate;
   type?: string;
+  verbose?: boolean;
```
yuku, line 28: Should I add `callbackManager` too?

https://github.com/hwchase17/langchainjs/blob/f78f2e90e95878da7c6067fd52e6a75eb0efb5fd/langchain/src/chains/base.ts#L9-L13

If yes, which way is better?

* Just add `callbackManager?: CallbackManager`
* Make `qaChainParams` extend `Omit<ChainInputs, "memory">`

langchain/src/chains/question_answering/load.ts
```diff
@@ -73,21 +80,23 @@ export const loadQAChain = (
...
 }
 
 export const loadQAStuffChain = (
   llm: BaseLanguageModel,
   params: StuffQAChainParams = {}
 ) => {
-  const { prompt = QA_PROMPT_SELECTOR.getPrompt(llm) } = params;
-  const llmChain = new LLMChain({ prompt, llm });
+  const { prompt = QA_PROMPT_SELECTOR.getPrompt(llm), verbose } = params;
+  const llmChain = new LLMChain({ prompt, llm, verbose });
```
yuku, line 91: (continue from the previous comment) then pass the `outputKey` to the `LLMChain` constructor. Does this make sense?

langchain/src/chains/question_answering/load.ts
yuku: Based on https://github.com/hwchase17/langchainjs/pull/909#pullrequestreview-1394896326, it seems the `StuffQAChainParams` should extend `Omit<LLMChainInput, "prompt" | "llm" | "memory">` 🤔

yuku:
I have marked the my last three comments as off-topic as they are outside the scope of this PR. Please forget them.

langchain/src/chains/question_answering/load.ts
yuku: I realized that `callbackManager` has been deprecated.

https://github.com/hwchase17/langchainjs/blob/033666e1a89d791483ce6345a8dab9a907754bd7/langchain/src/chains/base.ts#L17-L20

yuku:
Rebased.

nfcampos:
Thanks!

PR #910, opened 2023-04-20
Azure open api support changes.

FirozKuraishi:


nfcampos:
Thanks for this, I'm going to close this in favour of #1038 which is more complete

PR #915, opened 2023-04-20
Add contextual compression base abstractions

dev2049:
Adding just the base classes from [langchain#2915](https://github.com/hwchase17/langchain/pull/2915)

First TS pr, @nfcampos pls edit freely!


langchain/src/retrievers/index.ts
agola11, line 1: Importing from this file is deprecated. We'll want to create entrypoints for `retrievers/contextual_compression` and `retrievers/document_compressors` (see create-entrypoint.js)

PR #925, opened 2023-04-21
Make __run special key non-enumerable, ie. you can only access it specifcally by name

nfcampos:
- it does not show up when calling Object.keys
- it isn't included when calling JSON.stringify on the output object (eg. in the tracer)

This ensures that we don't break eg the logic that powers memory updates, and sequential application of chains

### DIFF
--- a/langchain/src/chains/base.ts
+++ b/langchain/src/chains/base.ts
@@ -1,5 +1,5 @@
 import { BaseMemory } from "../memory/base.js";
-import { ChainValues } from "../schema/index.js";
+import { ChainValues, RUN_KEY } from "../schema/index.js";
 import { BaseCallbackHandler, CallbackManager } from "../callbacks/index.js";
 import { SerializedBaseChain } from "./serde.js";
 import { BaseLangChain, BaseLangChainParams } from "../base_language/index.js";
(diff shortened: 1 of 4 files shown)


netzhuffle:
This seems to break a ChatConversionalAgent using a ChainTool:

```
Error when generating agent message TypeError: Cannot redefine property: __run
    at Function.defineProperty (<anonymous>)
    at VectorDBQAChain.call (node_modules/langchain/dist/chains/base.cjs:64:16)
    at VectorDBQAChain.run (node_modules/langchain/dist/chains/base.cjs:29:30)
    at ChainTool.call (node_modules/langchain/dist/tools/base.cjs:23:22)
    at node_modules/langchain/dist/agents/executor.cjs:101:23
    at async Promise.all (index 0)
    at AgentExecutor._call (node_modules/langchain/dist/agents/executor.cjs:97:30)
    at AgentExecutor.call (node_modules/langchain/dist/chains/base.cjs:53:28)
```

nfcampos:
@netzhuffle fixed in #959 thanks for reporting

nfcampos:
@netzhuffle this is fixed in 0.0.63

PR #929, opened 2023-04-21
Added support for Zod nativeEnum

baloodevil:
Added support for Zod nativeEnum so you don't have to define your enum values in two places.  Added a test to `structured.test.ts` which includes an Enum definition `StateProvinceEnum`.  It would be better if there was an enum already in langchainjs that could be referred to, but this at least makes it clear what is going on.


baloodevil:
> Thanks!

Glad to contribute!  Any chance this will get included in next release?  Hey- can you open Discord DM's to @Baloo#8072?

PR #931, opened 2023-04-21
Add framework to do evaluation of agent test cases against all agents

nfcampos:


langchain/src/agents/tests/chat_agent.int.test.ts
```diff
@@ -0,0 +1,116 @@
...
+  // Test that the agent throws the same error
+  await expect(() => executor.call({ input })).rejects.toThrowError(
+    "Request failed with status code 401"
+  );
+});
+
+test.skip("Run tool web-browser", async () => {
+  const model = new ChatOpenAI({ temperature: 0 });
+  const tools = [
+    new SerpAPI(process.env.SERPAPI_API_KEY, {
```
vowelparrot, line 92: Should we remove the serpapi tool here?  Not sure what it adds specifically

langchain/src/agents/tests/evaluation.int.test.ts
```diff
@@ -0,0 +1,91 @@
...
+  test.concurrent.each(scenarios)(`With scenario %#`, async (scenario) => {
+    const agentIndex = agents.indexOf(initializeAgentExecutorWithTools);
+    const scenarioIndex = scenarios.indexOf(scenario);
+    const { tools, input } = await scenario();
+    const agent = await initializeAgentExecutorWithTools(tools);
+    const result = await agent.call({ input });
+    console.log(`Agent #${agentIndex}`, `Scenario #${scenarioIndex}`, {
+      result,
+    });
+    expect(typeof result.output).toBe("string");
```
vowelparrot, line 89: I need to refresh my memory, but are there situations where the agent would return an error string rather than raising an error?

vowelparrot (approved):
This looks good to me..

langchain/src/agents/tests/evaluation.int.test.ts
nfcampos: Not unless the LLM returns an error string as "final action"

PR #937, opened 2023-04-22
Improves pdf parsing

Nishkalkashyap:
The below changes improve upon `PDFLoader.parse` method. Instead of adding a newline character after every `TextItem`, we only add a newline if the `TextItem` has an EOL character. This significantly improves the output text. Attaching an example below - 

Before:
```
Contents

Foreword
 
iii
Preface
 
v

1.
 
Real Numbers
 
1

1.1
 
Introduction
 
1
1.2
 
Euclid's Division Lemma
 
2
1.3
 
The Fundamental Theorem of Arithmetic
 
7
1.4
 
Revisiting Irrational Numbers
 
11
1.5
 
Revisiting Rational Numbers and Their Decimal Expansions
 
15
1.6
 
Summary
``` 

After: 
```
Contents
Foreword iii
Preface v
1. Real Numbers 1
1.1 Introduction 1
1.2 Euclid's Division Lemma 2
1.3 The Fundamental Theorem of Arithmetic 7
1.4 Revisiting Irrational Numbers 11
1.5 Revisiting Rational Numbers and Their Decimal Expansions 15
1.6 Summary 18
```

### DIFF
--- a/langchain/src/document_loaders/fs/pdf.ts
+++ b/langchain/src/document_loaders/fs/pdf.ts
@@ -39,13 +39,23 @@ export class PDFLoader extends BufferLoader {
         continue;
       }
 
-      const text = content.items
-      .map((i) => {
-        const item = i as TextItem;
-        if (item.hasEOL) {return `${item.str}\n`;}
-        return item.str;
-      })
-      .join("");
+      let lastY: number | undefined;
+      let text = "";
+      // Adapted from https://gitlab.com/autokent/pdf-parse/-/blob/master/lib/pdf-parse.js#L21
+      // https://github.com/mozilla/pdf.js/issues/8963
+      // https://github.com/mozilla/pdf.js/issues/2140
+      // https://gist.github.com/hubgit/600ec0c224481e910d2a0f883a7b98e3
+      // https://gist.github.com/hubgit/600ec0c224481e910d2a0f883a7b98e3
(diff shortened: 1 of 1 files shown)


nfcampos:
That doesn't seem to generalise to all pdfs, see eg the output for this pdf https://arxiv.org/pdf/1706.03762.pdf 

Screenshot 2023-04-22 at 18 58 20

nfcampos:
I've committed here the solution used by https://gitlab.com/autokent/pdf-parse, do you want to try on your pdf and see if that's better?

Nishkalkashyap:
> That doesn't seem to generalise to all pdfs, see eg the output for this pdf https://arxiv.org/pdf/1706.03762.pdf
> 
Are you sure you ran the PDF against my commit? Because I executed it against the PDF you mentioned (https://arxiv.org/pdf/1706.03762.pdf) and I'm getting a different result, as shown below. 

Perhaps your console is not rendering the `\n` character and that's why the result seems off? 

image

Nishkalkashyap:
> I've committed here the solution used by https://gitlab.com/autokent/pdf-parse, do you want to try on your pdf and see if that's better?

Unfortunately, the solution that you have committed doesn't seem to work for my pdfs. Example - red side of the diff editor is the output from my commit and green side would be from your commit.

image

PR #941, opened 2023-04-22
feat: pass _runManager to func in the DynamicTool

toxic-johann:
It will be helpful to have the `_runManager` in the dynamic function.

We can use it as a context with the `runId`.


nfcampos:
Agreed, thanks!

PR #951, opened 2023-04-23
Adding Redis Memory Support

### LINKED ISSUE #731: RedisMemory support
Similar to [Python](https://python.langchain.com/en/latest/modules/memory/examples/redis_chat_message_history.html) (and the current Motorhead implementation), would it be possible to get a RedisMemory client? 

Also - if you ask in the interactive JS docs, the response hallucinates the client

Screenshot 2023-04-10 at 8 31 55 PM

cmtoomey:
Adds Redis Memory to close #731. Instantiation is similar to Motorhead implementation. Includes docs and tests, based on Motorhead test suite.

- Net new tests are passing: 
Screenshot 2023-04-22 at 8 15 00 PM

- Linting is clean for net new files.
Screenshot 2023-04-22 at 8 37 22 PM

- Yarn formatted before most recent commit


jacoblee93:
Having looked a little more, I think this would fit better as a `ChatMessageHistory` store like this: https://github.com/hwchase17/langchainjs/pull/1016

That way other memory types could use Redis persistence in the future.

cmtoomey:
@jacoblee93 - that makes sense.

cmtoomey:
@jacoblee93 - I've made updates per your suggestions

jacoblee93:
Thanks! Looks great overall - I think it'd be a bit more user-friendly to just initialize the client within the message store class rather than having the user import directly from a hidden dep, and there are a few standardization things we added more recently. I can clean it up from here!

PR #956, opened 2023-04-23
Update schema.prisma to inclue pgvector extension, simplify docs

praveen-palanisamy:
This PR updates the `schema.prisma` sample to create/add the [pgvector](https://github.com/pgvector/pgvector) postgresql extension and simplifies the steps listed in the [Prisma vector store integration doc](https://js.langchain.com/docs/modules/indexes/vector_stores/integrations/prisma).

With the updated `schema.prisma`, `prisma migration dev` should automatically generate and apply a `migration.sql` to the database with `CREATE EXTENSION IF NOT EXISTS "vector" WITH SCHEMA "extensions";` to add the pgvector extension. 
This avoids the manual migration SQL update step and reduces the commands the user needs to run.
Simplified the user/developer docs.


jacoblee93:
Let's wait until it's no longer a preview feature - otherwise we'll have to update the docs again once that's no longer necessary.

PR #958, opened 2023-04-23
Update Chroma class with new functionality for collection creation

arajooj:
The Chroma class has been updated with new functionality for creating and retrieving collections. The ensureCollection method now takes a new optional parameter 'newCollection', which allows the caller to specify whether or not to create a new collection. If 'newCollection' is true, a new collection will be created if one with the same name does not already exist. If 'newCollection' is false (or not provided), the method will return without doing anything if the collection already exists.

Additionally, a new method 'getVectorStore' has been added to the Chroma class. This method retrieves the vector store without attempting to create a new collection if one with the same name already exists. This can be useful in cases where the user wants to retrieve an existing vector store, but does not want to create a new one if it does not exist.

The constructor has also been updated to use the new 'ensureCollection' method with the 'newCollection' parameter set to true by default, ensuring that a new collection is created if one does not already exist.

Before:
image

After:
image

And it's a small change, but every millisecond counts.

### DIFF
--- a/docs/docs/modules/indexes/vector_stores/integrations/chroma.md
+++ b/docs/docs/modules/indexes/vector_stores/integrations/chroma.md
@@ -45,6 +45,13 @@ const vectorStore = await Chroma.fromDocuments(docs, new OpenAIEmbeddings(), {
   collectionName: "goldel-escher-bach",
 });
 
+// or alternatively from a existing collection
+const vectorStore = await Chroma.fromExistingCollection(new OpenAIEmbeddings(),
+  {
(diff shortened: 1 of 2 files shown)


l4b4r4b4b4:
that looks exactly like the PR I need :D

nfcampos:
@arajooj thanks for the PR, there is an earlier PR which does the same without changing the API so I'll be closing this one #954

arajooj:
I thank you, my dear. I will continue improving what I have achieved!

PR #964, opened 2023-04-23
Fix ChatConversationalAgentOutputParser cannot handle markdown in the JSON

micahriggan:
The output parser doesn't handle when the AI's action_input contains markdown, or ``` backticks really, as it's matching on the first set of backticks it finds rather than the last.

Tested with `yarn test:single langchain/src/agents/tests/chat_output_parser.test.ts`


micahriggan:
For the lint, how should I resolve the no-unused-expressions on the tests calling expect?

langchain/src/agents/tests/chat_output_parser.test.ts
```diff
@@ -0,0 +1,54 @@
+import { ChatConversationalAgentOutputParser } from "../chat_convo/outputParser.js";
```
nfcampos, line 1:
```suggestion
import { test, expect } from "@jest/globals";
import { ChatConversationalAgentOutputParser } from "../chat_convo/outputParser.js";
```

nfcampos:
@micahriggan I've fixed it, it was missing the `()` for doing the function call

NJordan72:
FWIW, I think [StructuredOutputParser](https://github.com/hwchase17/langchainjs/blob/main/langchain/src/output_parsers/structured.ts) has the same problem

micahriggan:
@NJordan72 agreed, just took a look at that file and it has the same issue.

micahriggan:
Anything else I need to do on this one? @nfcampos

nfcampos:
Thanks, merging this now. The same issue in StructuredOutputParser is resolved in another PR

Issue #887, opened 2023-04-19
Do we have any plan for Azure OpenAI?


baobo5625:

amoreaulemay:
So you can do that already. When you create a model, you can pass in a `ConfigurationParams` as a second argument on the constructor:

```typescript
const model = new OpenAI({temperature: 0}, {basePath: '<your-azure-endpoint>'})
```

bxav:
Hi @amoreaulemay, did you make it work? Looks like langchainjs is using the openai package and the issue there is still not resolved https://github.com/openai/openai-node/issues/53 . 
Also, the header for azure are not the same for auth the request to there models (api-key for azure, Authorization/Bearer token for openai).

amoreaulemay:
Sorry @bxav you're right. I based this off Helicone which acts as a proxy and assumed azure used the same request schema

Taylorsuk:
[This solution works well for me](https://github.com/openai/openai-node/issues/53#issuecomment-1517604780)

Issue #601, opened 2023-04-04
Azure Open API Support: available in Python version but not in this package


FirozKuraishi:
Support is available in python package as 
from langchain.llms import AzureOpenAI

and llm = AzureOpenAI(deployment_name="text-davinci-003", model_name="text-davinci-003")

but ther is no option to define the same here.

nfcampos:
Hi, yes sadly as far as I know the node sdk for OpenAI doesn't offer support for Azure out of the box, see https://github.com/openai/openai-node/issues/53

See also this discussion https://github.com/hwchase17/langchainjs/issues/244

FirozKuraishi:
@nfcampos I used https://github.com/hwchase17/langchainjs/issues/244 to create a sample application which worked fine but don't know how to integrate Langchain using this as when trying to use const chain = new LLMChain({ llm: model, prompt }); and passing model created using 
const { Configuration, OpenAIApi } = require("azure-openai");
model= new OpenAIApi(configuration);
it's giving error as model not supported.

Any suggestion/workaround to use to pass this model to llm chain

Elvincth:
+1 for this, any update? Maybe supporting the [azure openai package](https://www.npmjs.com/package/azure-openai) ?

nfcampos:
@Elvincth i didn't know about that package, thanks for sharing. I can open a PR with this, but I'm not able to test it, hopefully one of you can?

Elvincth:
> @Elvincth i didn't know about that package, thanks for sharing. I can open a PR with this, but I'm not able to test it, hopefully one of you can?

Yes, sure I can help you test it.

firoz-qureshi:
@Elvincth @nfcampos I was able to use Azure OpenAI after doing some modification with current langchain llms. I did the following changes :

**1. Added AzureLLM class at src/llms package.**

 import { CallbackManager } from 'callbacks/base.js';
import { chunkArray } from "../util/chunk.js";
import { BaseLLM } from "./base.js";
import { calculateMaxTokens } from "../base_language/count_tokens.js";
import { LLMResult } from "../schema/index.js";
import { TiktokenModel } from '@dqbd/tiktoken';

const promptToAzureArgs = ({
  prompt,
  temperature,
  stop,
  maxTokens,
}: {
  prompt: string[]
  temperature: number
  stop: string[] | string | undefined
  maxTokens: number
}): LLMPromptArgs => ({
    prompt,
    temperature,
    max_tokens: maxTokens,
[...]

PR #966, opened 2023-04-23
Added implementation for Azure OpenAI

dersia:
I added an implementation for using langchain with Azure OpenAI. In generel, it would be sufficient to make the `openAIApiKey` non mandatory by removing the check and throw for it, but I thought it makes sense to add some api to make it more obvious to setup the models to use Azure OpenAI.
I added it to the completion models, the embedding model and the chat-model. 
All Azure parameters can also be set using envvars.

I also adjusted the docs accordingly.
I haven't yet adjusted the unit-tests and I would like to adjust those, before this is merge.

I know that there is already a PR for Azure OpenAI, but I think it makes more sense to use the same models for Azure OpenAI and OpenAI's API, becuase maintanace would be easier this way and when new features are added, both profit from it. So this PR is not adding new models for Azure, but using the same models, by just extending the model with a few extra parameters and enforcing those, when the Azure OpenAI Api Key is set. 

This implementation also throws, if both Api Keys `openAIApiKey` and `azureOpenAIApiKey` are missing.


dersia:
I don't have access to Vercel, can please one of the maintainers look for the issues and let me know what to fix?

kunjan13:
when will these be available in langchain package.. we are waiting for this one.

dersia:
I am working on the tests. I think will have them done by tomorrow around this time.

dersia:
So I added a few test for the exports, but honestly the whole experience for running the tests is really a pain for me. 
* I have problems compiling the project on windows, because node-gyp isn't finding the vc tools and I really don't know why.
* on wsl I can get it compiling and running the tests, but it takes like forever and so some test timeout because it is running really that slow
* I looked into adding integration test for azure, but I this would mean, that for the test to run on the github ci action, this project would need an azure api key (and deployment and instance etc) and I have no clue, if this project has access to it. Maybe our friends from MS can provide an instance running on Azure for this project?
* I would also like to add some examples to the repo, but again the problem would be with the api keys, anyhow I will add the examples in another PR
* so far I am done, but I want to run some last manual tests and will afterwards edit the PR name so it can be merged

I would really love for someone to please run the tests

dersia:
ok, I got it all working with node-gyp and I was able to run all the tests and everything looks good. I also rebased from main.

@nfcampos could you please check and merge?

dersia:
fixes #887 and #601

liuxuan1-1:
Hi @nfcampos, could you please review this PR and merge it? It is really useful. Thank you very much.

iosifboanca:
@dersia Please don't merge this until you test streaming as well :) - I believe END event will never happen

baobo5625:
do we have an eta to enable this function in langchain?

isshesure:
Has this been tested on chat completion streaming? It seems Azure's response has a single blank line (compared to OpenAI's 2 blank lines) after `data: [DONE]` so `getMessages` in `event-source-parse` never invokes `onMessage` to send the `[DONE]` event?

nfcampos:
Hi folks, is one of you able to test this? I don't have access to an Azure model to test it with

dersia:
I did run the streaming tests against azure and they looked good to me. But I'd be happy to run others tests, if someone has any specific tests in mind.

acshulk:
I see the following when trying to build.  Were references moved on the latest build?

Error : 
node_modules/langchain/dist/llms/openai.d.ts:2:66 - error TS2307: Cannot find module 'types/open-ai-types.js' or its corresponding type declarations.

Original Call :

nfcampos:
@acshulk Thanks for reporting, this is fixed in 0.0.71

bryantpa:
@nfcampos I'm on 0.0.71 and getting the error mentioned by @acshulk still. on both import { OpenAI } from "langchain/llms/openai" and import { OpenAIEmbeddings } from "langchain/embeddings/openai";. 

`node_modules/langchain/dist/embeddings/openai.d.ts(2,34): error TS2307: Cannot find module '../types/open-ai-types.js' or its corresponding type declarations.
node_modules/langchain/dist/llms/openai-chat.d.ts(2,70): error TS2307: Cannot find module '../types/open-ai-types.js' or its corresponding type declarations.
node_modules/langchain/dist/llms/openai.d.ts(2,66): error TS2307: Cannot find module '../types/open-ai-types.js' or its corresponding type declarations.`

nfcampos:
@bryantpa That is fixed in 0.0.72

baobo5625:
Tested in 0.0.72, still same.

Using ChatOpenAI with gpt-4-32k

baobo5625:
I think it shouldn't be depend on model, 3.5 should also produce the same result.

baobo5625:
```
const chatStreaming = new ChatOpenAI({
    temperature: 0,
    streaming: true,
    azureOpenAIApiKey: azureOpenAIConfig.openApiKey,
    azureOpenAIApiInstanceName: azureOpenAIConfig.accountName,
    azureOpenAIApiDeploymentName: azureOpenAIConfig.deploymentName,
    azureOpenAIApiVersion: '2023-03-15-preview',
  });

  const response = await chatStreaming.call([new HumanChatMessage(message)], undefined, [
    {
      handleLLMNewToken(token: string) {
        logger.info(token);
      },
    },
  ]);
  return response;
```

baobo5625:
found the reason, the last message in azure streaming is:

{"id":"chatcmpl-7DxBV9AgR8GuvqLyrsmRtDav2M9zJ","object":"chat.completion.chunk","created":1683560125,"model":"gpt-4-32k","choices":[{"index":0,"finish_reason":"stop","delta":{}}],"usage":null}

baobo5625:
I tested with my local axios implementation, azure openai does return the "[DONE]" message

baobo5625:
Tried with Azure OpenAI vs OpenAI, the "onmessage" can capture [DONE] from openai, but not the azure one.

if not set streaming, then it works.

nfcampos:
@baobo5625 I believe https://github.com/hwchase17/langchainjs/pull/1167 will fix your issue, any chance you could pull this locally and test it?

baobo5625:
it could be one of the ways to fix it, however I tested out with my local implementation of Azure OpenAI in axios, it does return the [DONE] message.

I tested locally and it works.

I suggest to put a todo there to find out why

PR #968, opened 2023-04-24
unrolled error in supabase vectorstore and retrievers

ppramesi:
So now if there's an error its clear what the error is, instead of printing out [object object].


nfcampos:
Thanks!

j1philli:
Just ran into this exact error with SupabaseHybridSearch and getRelevantDocuments. Will be great to have this so I have a path for debugging. Thanks @ppramesi

PR #971, opened 2023-04-24
nc/babyagi

nfcampos:


agola11 (approved):
Skimmed and lgtm. I think there are things we can factor out of this and AutoGPT, namely things like `vectorstore, maxIterations, executionChain` but can wait until we move out of experimental

PR #973, opened 2023-04-24
Relax axios peer dep range

nfcampos:


vowelparrot:
Was this originally done to match `openai`'s?

nfcampos:
@vowelparrot correct, but as far as I can test more recent versions are compatible with openai sdk too.

agola11 (approved):

PR #998, opened 2023-04-26
feat: Add Apify integration

jirimoravcik:
JS version of https://github.com/hwchase17/langchain/pull/2201

If you have any suggestions, feel free to comment on this PR.
Also if there are some issues concerning Apify integrations in langchain, feel free to contact me.


langchain/src/tools/apify.ts
```diff
@@ -0,0 +1,93 @@
+import { Document } from "../document.js";
```
jacoblee93, line 1: Tools have a specific meaning in LangChain - this is in the wrong place. Can we bundle it into the document loader and run `callActor` when `load`'ing documents?

jacoblee93 (changes requested):

langchain/src/tools/apify.ts
jirimoravcik: Thanks for the review.
In the Python version, it's in the utilities. https://github.com/hwchase17/langchain/blob/master/langchain/utilities/apify.py
Not sure if there's an equivalent of it in the JS version?

jacoblee93: In that case I'd prefer to put this as a `callActor` method in the `document_loader` class. I can make the change today and polish everything up if that's ok with you?

jacoblee93: The other reason is so that we can give everything that requires Apify one entrypoint.

jancurn: Thank you @jacoblee93 If you prefer so, feel free to update it that way.

jacoblee93:
Extended and merged here: #1271 

Thanks for your patience!

PR #1011, opened 2023-04-27
Serp base param

### LINKED ISSUE #982: serpApi base url
I wish to be able to use the SerpApi tool in browser. 

As far as I can tell, 
https://github.com/hwchase17/langchainjs/blob/8bf7e9028d048c194f543fb652dcb35a93ab398b/langchain/src/tools/serpapi.ts#L294

Assumes that one will always call the serpApi.com directly. This means one cannot instantiate a proxy between this tool, and serpapi. In turns, this results in a CORS error. 

Proposal: Similar to the paradigm in the `OpenAiChat` modules, introduce a "base" parameter (by defult the existing serpapi.com) which one can override, to point the tool at a proxy.

A PR looks fairly simple, so I would consider attempting it if this proposal gets a thumbs up.

Quafadas:
Aims close #982 by adding a base url parameter to the serpapi class. Should allow one to push the serpapi tool through a proxy.

### DIFF
--- a/langchain/src/tools/serpapi.ts
+++ b/langchain/src/tools/serpapi.ts
@@ -278,21 +278,10 @@ export interface SerpAPIParameters extends BaseParameters {
   ijn?: string;
 }
 
-type UrlParameters = Record<
+export type UrlParameters = Record<
   string,
   string | number | boolean | undefined | null
 >;
-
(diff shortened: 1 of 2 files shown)


nfcampos:
Thanks!

PR #1013, opened 2023-04-27
Fix the issue where the track function does not support the LANGCHAIN_SESSION environment variable

rxliuli:
When using the langchain track, I found that the LANGCHAIN_SESSION environment variable does not take effect, and in python's documentation, it is supported.
ref: https://python.langchain.com/en/latest/tracing.html#changing-sessions

### DIFF
--- a/langchain/src/callbacks/manager.ts
+++ b/langchain/src/callbacks/manager.ts
@@ -464,7 +464,15 @@ export class CallbackManager
           (handler) => handler.name === "langchain_tracer"
         )
       ) {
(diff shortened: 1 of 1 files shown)


rxliuli:
Oddly, CI fails due to failure to load a stylesheet
ref: https://fonts.googleapis.com/css2?family=Inter:wght@100..900&display=swap

nfcampos:
@rxliuli ah yes ive fixed that in another PR and updated your branch now

agola11 (approved):
thanks!

PR #1014, opened 2023-04-27
Added additional check for empty object when upserting to Pinecone Index

gullerg:
Langchain checks for null properties and deletes these before upserting to the Index. I've added an additional check for empty objects in the metadata, as these are not handled when the array is flattened. 

I had issues when upserting to Pinecone, and I realised this was because of the fact that sometimes there were empty objects as properties in the PDF metadata. The proposed change fixed this issue.


nfcampos:
Thanks!

PR #1023, opened 2023-04-27
Add inputKey and outputKey fields to BufferWindowMemory

nSimonFR:
Useful when using `returnIntermediateSteps` - makes it ISO with `BufferMemory`.


nfcampos (approved):
Thanks!

PR #1024, opened 2023-04-27
chore(version): update package @zilliz/milvus2-sdk-node from 2.2.0 to 2.2.7

Todmy:
The update includes the change in SDK interface

logs when I tried to run langchainjs with the new version of milvus2-sdk-node@2.2.7:
```bash
dataManager are no longer necessary, you can call methods directly on the client object.
collectionManager are no longer necessary, you can call methods directly on the client object.
indexManager are no longer necessary, you can call methods directly on the client object.
```


jacoblee93:
Thank you!

yongyi520:
When I use the search parameter now, it does not return database field. any pointers on this? Have you tested the results from vector similarity search?

Screenshot 2023-06-21 at 19 30 14

PR #1025, opened 2023-04-27
Better structured output parsing using zod-to-json-schema

jasongill:
Now that [zod-to-json-schema](https://github.com/StefanTerdell/zod-to-json-schema) is included in the project since https://github.com/hwchase17/langchainjs/commit/4a0c69185b4a79f704a01927e2cb9998acbef606, we can stop trying to manually parse all of the different types supported by Zod and instead switch to generating (and parsing) valid JSON schema in our format instructions.

[This mirrors the way that the Python version of Langchain generates structured-data prompts](https://github.com/hwchase17/langchain/blob/master/langchain/output_parsers/format_instructions.py#L11-L19) using Pydantic, so the output should be identical between the two packages now.

Plus, this should improve reliability and reduce the amount of code we have to maintain to keep up with [all of the different object types that Zod supports](https://github.com/colinhacks/zod#table-of-contents).


jasongill:
This fixes tickets like #929 permanently by including support for ALL of the Zod types, without having to re-invent the wheel by creating our own handler for each type. This PR basically replaces the work done committed today in bb602ef742cbc41280826ab525f6289b8cc941de (sorry!)

jasongill:
I've updated this PR to include the changes from 11397ab30e73761815c49e6b23d4509d1ee10cef so that it will merge without conflicts.

nfcampos:
Thanks! This is great

PR #1030, opened 2023-04-27
Making sure that the SQL command is properly displayed in the logs

felipegtx:
This change allows for the information related to `sqlCommand` to be displayed correctly in the logs.

At first, I thought this was a bug in my prompt/code because I was seeing the following output:
```
SQLQuery:NaN
SQLResult: [{"COUNT(*)":0}]
Answer:
```

With the new code this should now display the actual Command the LLM is suggesting.


nfcampos:
Thanks!

PR #1046, opened 2023-04-29
Fixed import for TextLoader in example

philrwebb:
import { TextLoader } from "langchain/document_loaders" was sufficient


nfcampos:
Hi @philrwebb that import path is deprecated and will be removed in a future version

PR #1049, opened 2023-04-29
Adds input parsers for validating and documenting LLM inputs

jacoblee93:
While working with LLMs as a step in a pipeline with other APIs, I found it useful several times to use a prompt like this in conjunction with an output parsers:

```
The JSON array below represents X. Each item in the array is an object containing two fields:
  - foo, a string that represents Y
  - bar, a number that represents Z
```

I've framework-itized it as a `StructuredInputParser` in this PR. In addition to giving the LLM more context as to what each field does beyond just the field name and overall structure, there's the nice benefit of encouraging people to define input types, more easily support complex inputs, and generally think more about what they're passing into LLMs.

I've included a unit and integration test that illustrate what using such a parser looks like in practice - here's an example of what I'm envisioning:

```js
const chat = new ChatOpenAI({ modelName: "gpt-3.5-turbo", temperature: 0 });
const zodSchema = z
  .array(
    z.object({
      id: z.string().describe("the unique record id"),
      created_at: z.string().describe("creation date of the database record"),
      fields: z.object({
        title: z.string().describe("the book title"),
        seriesId: z.string().describe("id of the series the book belongs to"),
        createdAt: z.string().describe("publish date of the book"),
      }),
    })
  )
  .describe("an array of database records, each representing a book");
const inputParser = StructuredInputParser.fromZodSchema(zodSchema);
const outputParser = StructuredOutputParser.fromZodSchema(zodSchema);
const chain = new LLMChain({
  llm: chat,
  prompt: new PromptTemplate({
    template: `Answer the user's question factually and as best you can.

{input_format_instructions}

{input}

{output_format_instructions}

{query}`,
    inputVariables: ["query", "input"],
    partialVariables: {
      input_format_instructions: inputParser.getFormatInstructions(),
      output_format_instructions: outputParser.getFormatInstructions(),
    },
  }),
});

  query: `List all the books published before 2000.`,
  input: await inputParser.parse([
    {
      id: "rec0",
      created_at: "2021-11-06T00:00:00.000Z",
      fields: {
        title: "Book 1",
        seriesId: "1",
        createdAt: "1996-08-01T00:00:00.000Z",
      },
    },
    {
      id: "rec1",
      created_at: "2022-08-17T00:00:00.000Z",
      fields: {
        title: "Book 2",
        seriesId: "1",
        createdAt: "1999-08-01T00:00:00.000Z",
      },
    },
    {
      id: "rec3",
      created_at: "1994-01-08T00:00:00.000Z",
      fields: {
        title: "Novel 1",
        seriesId: "2",
        createdAt: "2010-08-31T00:00:00.000Z",
      },
    },
    {
      id: "rec4",
      created_at: "1995-01-08T00:00:00.000Z",
      fields: {
        title: "Novel 2",
        seriesId: "2",
        createdAt: "2011-09-01T00:00:00.000Z",
      },
    },
  ]),
});

console.log(await outputParser.parse(res.text));
```

I also made a small tweak to the output parser prompt to more strongly hint at the required output format as I noticed some flakiness while writing my tests (s/The output should/Your output must). 

I didn't want to spend too much time writing docs as I'm sure there'll be a lot of feedback and requested changes, but am happy to do so as well!


jacoblee93:
Maybe `StructuredInputParameterParser` is a clearer name?

langchain/src/util/zod.ts
```diff
@@ -0,0 +1,70 @@
+import { z, util } from "zod";
+
+export function printZodSchema(schema: z.ZodTypeAny, depth = 0): string {
```
nfcampos, line 3: We've actually replaced this now in #1025

jacoblee93:
Found that `gpt-3.5-turbo` was a little bit worse at picking up the "description" with this new JSON Schema format - kept getting silly responses like:

`Based on the input data provided, it is not possible to determine which books were published before 2000 as the "createdAt" field in the "fields" property only specifies the publication date of each book, not the year it was published.`

`gpt-4` worked better but decided to simplify the integration test to make it a bit less intentionally misleading for the LLM and it worked fine.

langchain/src/util/zod.ts
jacoblee93: Removed and am now using the `zodToJsonSchema` method

PR #1050, opened 2023-04-29
Update broken docs links

jacoblee93:
While looking into writing docs for #1049, I noticed a few broken links in the `Prompts` section of the docs. I've fixed them here.


nfcampos:
Thanks!

PR #1058, opened 2023-04-30
Update unstructured_directory.ts

CHAOWEICHIU:
## Abstract 
Format the example that follows the pattern of other example

## Reason
Follow the convention of other examples


nfcampos:
Hi @CHAOWEICHIU this actually is the new pattern for examples, we've done away with the requirement to have a run function surrounding the code.

PR #1060, opened 2023-04-30
Update BaseMemory.getInputValue() so it can access nested keys

ivanblagdan:
Update that allows key traversal for Memory that is useful when dealing with structured output. E.g., Motorhead can only store string values, which might be nested deep in the outputValues object.


bracesproul:
@ivanblagdan going to close due to merge conflicts. If you'd still like to get this merged in, please open off of an up-to-date main branch, and I'll review. Thank you!

PR #1076, opened 2023-05-01
Reverse motorhead init

Czechh:
Motorhead returns messages sorted by descending order (newest first).

So, `init` has to reverse to `add` messages in the right order.


softboyjimbo (approved):
lgtm

nfcampos:
Thanks!

PR #1080, opened 2023-05-02
feat: added `returnIntermediateSteps` flag to `MapReduceChain`

### LINKED ISSUE #171: [Question]: Ability to see intermediate steps in MapReduceChain?
Is there any way to see the intermediate steps in the mapreduce chain when using the `OpenAI` llm?

I'm only able to see the original prompts at this time, even when setting it to "verbose"

justindra:
Added the `returnIntermediateSteps` flag to `MapReduceChain` to add towards feature parity with the Python Library. 

Fixes #171 

Side note:
I noticed while doing this that in the `loadQAChain` it is calling slightly different prompts compared to when you directly use the `loadQAMapReduceChain`, `loadQAStuffChain` and `loadQARefineChain` instead of specifying a `type`. Is there a reason for this? If not, I can make another PR to fix that up and to make it reference the functions directly.

### DIFF
--- a/docs/docs/modules/chains/other_chains/summarization.mdx
+++ b/docs/docs/modules/chains/other_chains/summarization.mdx
@@ -1,8 +1,15 @@
 import CodeBlock from "@theme/CodeBlock";
 import SummarizeExample from "@examples/chains/summarization_map_reduce.ts";
+import SummarizeExampleIntermediateSteps from "@examples/chains/summarization_map_reduce_intermediate_steps.ts";
 
 # Summarization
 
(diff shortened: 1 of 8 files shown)


justindra:
@nfcampos btw, not sure why but that test against Ubuntu - Node 20 seems to have been going for a couple hours there. Might want to retry it?

rohanprasadofficial:
i think everyone is going love this ❤️ , Thanks for bring this up. 
Eagerly waiting to consume this in something i have been building

jacoblee93:
LGTM! Thank you!

rohanprasadofficial:
Thanks @justindra for adding this. @nfcampos When we can expect new package coming with this change ?

PR #1087, opened 2023-05-02
Add Vespa retriever

lesters:
Add Vespa as a LangChain retriever. A very thin wrapper around RemoteRetriever.

Not sure I got all the correct places to insert this, please let me know if there are something else that needs to be added.


dqbd (approved):
Made some small changes addressing lint and formatting issues, otherwise lgtm!

jacoblee93:
Hey apologies for losing track of this @lesters - happy to merge more or less as is but is it possible to add a type `query_body`? Would make it easier for folks to use.

Also, `query_body` is a bit confusing to me as they're both HTTP terms, and it's more JS convention to use camel case (`queryBody`) instead of snake. Since it's relatively isolated though (and you all probably have your own conventions) it's no big deal. Let me know if you'd just like to get it merged though as I know it's been a while!

docs/docs/modules/indexes/retrievers/vespa-retriever.mdx
```diff
@@ -0,0 +1,42 @@
+# Vespa retriever
+
+This shows how to use Vespa.ai as a LangChain retriever.
+Vespa.ai is a platform for highly efficient structured text and vector search.
+Please refer to [Vespa.ai](https://vespa.ai) for more information.
+
+The following sets up a retriever that fetches results from Vespa's documentation search:
+
+```
```
jacoblee93, line 9: Would ideally also prefer this to be in the `@examples` folder and imported

langchain/src/retrievers/index.ts
```diff
@@ -10,3 +10,4 @@ export {
 } from "./supabase.js";
 export { RemoteLangChainRetriever } from "./remote/remote-retriever.js";
 export { MetalRetriever } from "./metal.js";
+export { VespaRetriever } from "./vespa.js";
```
jacoblee93, line 13: We're deprecating all the `index.ts` files, wouldn't cause a problem in this case but best not to include it here.

jacoblee93:
Didn't work for locally running instances since the port wasn't getting passed through, just using the passed in URL directly. Will merge and it will be out with the next release!

PR #1094, opened 2023-05-02
add support of parsing multi-line text

wizd:
for example, an image generator may need the following input. the current regex can not match multi line text.

<pre>I need to generate an image of the Mona Lisa in the style of traditional Chinese painting.
Action: image-generator
Action Input: 
{
  "Subject": "Mona Lisa",
  "Style": "Chinese traditional painting",
  "Color": "Mainly wash tones of ink, with small color blocks in some parts",
  "Details": "Mona Lisa should have long hair, a silk dress, holding a fan. The background should have mountains and trees.",
  "Emotion": "Serene and elegant"
}<pre>


nfcampos:
Thanks!

PR #1095, opened 2023-05-02
[WIP] Structured tool input agent

vowelparrot:
At first I thought I could do some magic with TS default generics to avoid making a new agent class, but I realized the static methods wouldn't be able to bind to the type.

We relased a structured chat agent to be compatible with structured tools in python. This is something of an attempt at porting it over.

### DIFF
--- a/langchain/src/agents/agent.ts
+++ b/langchain/src/agents/agent.ts
@@ -9,7 +9,7 @@ import {
   BaseChatMessage,
   ChainValues,
 } from "../schema/index.js";
-import { Tool } from "../tools/base.js";
+import { StructuredTool, Tool } from "../tools/base.js";
 import {
   AgentActionOutputParser,
   AgentInput,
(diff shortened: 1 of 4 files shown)


langchain/src/agents/agent.ts
```diff
@@ -195,7 +195,10 @@ export interface AgentArgs {
  * include a variable called "agent_scratchpad" where the agent can put its
  * intermediary work.
  */
-export abstract class Agent extends BaseSingleActionAgent {
+export abstract class Agent<
```
vowelparrot, line 198: Clearly this doesn't work. Basically I wanted to avoid making another layer of inheritance just for `tool` typing

langchain/src/agents/structured_chat/index.ts
```diff
@@ -0,0 +1,124 @@
...
+   * @param tools - List of tools the agent will have access to, used to format the prompt.
+   * @param args - Arguments to create the prompt with.
+   * @param args.suffix - String to put after the list of tools.
+   * @param args.prefix - String to put before the list of tools.
+   */
+  static createPrompt(tools: StructuredTool[], args?: ChatCreatePromptArgs) {
+    const { prefix = PREFIX, suffix = SUFFIX } = args ?? {};
+    const toolStrings = tools
+      .map((tool) => `${tool.name}: ${tool.description}, args: ${JSON.stringify(
+        (zodToJsonSchema(tool.schema) as JsonSchema7ObjectType).properties
```
vowelparrot, line 90: technically slightly different right now since the args schema in python is sub-selecting on `properties` but I think we're going to change that soon

PR #1096, opened 2023-05-02
feat: add support for extensions as globs in directory loader

rpidanny:
This PR adds support for glob patterns, similar to the Python interface. With this change, more file types can be loaded with some common default loaders, and common loaders can be used for similar file types.

One useful application of this feature is when loading source code or similar text-based files. Using glob patterns makes it easier to use a TextLoader for multiple extensions at once.

Please note that this change is backward compatible with the previous approach of static extensions, but requires the installation of the [minimatch](https://www.npmjs.com/package/minimatch) package.


rpidanny:
Hi @nfcampos , not sure what is causing Vercel to fail since I don't have permission to view the build. Please let me know if I need to fix something on my side 🙏

langchain/src/document_loaders/fs/directory.ts
```diff
@@ -19,7 +19,7 @@ export class DirectoryLoader extends BaseDocumentLoader {
   constructor(
     public directoryPath: string,
     public loaders: {
-      [extension: string]: (filePath: string) => BaseDocumentLoader;
+      [extension: `.${string}`]: (filePath: string) => BaseDocumentLoader;
```
rpidanny, line 22: Since we're always expecting the extension to start with a `.`, this should help TS users to have better DX. wdyt?

rpidanny:
HI @nfcampos any thoughts on this PR?

langchain/src/document_loaders/fs/directory.ts
```diff
@@ -41,6 +41,7 @@ export class DirectoryLoader extends BaseDocumentLoader {
   }
 
   public async load(): Promise<Document[]> {
+    const { minimatch } = await minimatchImports();
```
dqbd, line 44: I would suggest importing `minimatch` statically instead dynamically. 

As this file is already a custom entrypoint, `minimatch` will not be imported if unused. And existing users of `DirectoryLoader` will get an error message if upgrading, which can be considered as a breaking change.

langchain/src/document_loaders/fs/directory.ts
```diff
@@ -59,11 +60,15 @@ export class DirectoryLoader extends BaseDocumentLoader {
...
       } else {
-        // I'm aware some things won't be files,
-        // but they will be caught by the "unknown" handling below.
-        const loaderFactory = this.loaders[extname(file.name)];
-        if (loaderFactory) {
-          const loader = loaderFactory(fullPath);
+        let matchedLoader = null;
+        for (const pattern in this.loaders) {
+          if (minimatch(extname(file.name), pattern)) {
+            matchedLoader = this.loaders[pattern as `.${string}`];
```
dqbd, line 66: _nit_: unnecessary as `.${string}` typing.


```suggestion
            matchedLoader = this.loaders[pattern];
```

dqbd (changes requested):

langchain/src/document_loaders/fs/directory.ts
dqbd: _note_: this does mean that the bundle will increase by ~20.6kB minified (https://bundlephobia.com/package/minimatch@9.0.1), but that should not be such a big concern, as I don't think people will use fs-based loaders in serverless?

docs/docs/modules/indexes/document_loaders/examples/file_loaders/directory.md
```diff
@@ -14,10 +14,17 @@ src/document_loaders/example_data/example/
...
 ├── example.txt
-└── example.csv
+├── example.csv
+└── example.md
 ```
 
-Example code:
+## Setup
+
+```bash npm2yarn
```
dqbd, line 23: Remove the following instruction if decided to move `minimatch` from an optional to a required dependency.

dqbd:
Thanks for the PR! Looks great, if possible, please rebase so that we can merge it.

docs/docs/modules/indexes/document_loaders/examples/file_loaders/directory.md
rpidanny: Done :)

bracesproul:
@rpidanny going to close due to merge conflicts. If you'd still like to get this merged in, please open off of an up-to-date main branch, and I'll review. Thank you!

PR #1098, opened 2023-05-02
update variable name to responseB

bmacer:
If you're just copy-pasting you'll run into a const redefine error from `const response` being defined above.  in the below code block `const responseC` is defined, so this continues with that convention.


nfcampos:
Thanks!

PR #1101, opened 2023-05-02
Fix cohere embed

mkozakov:
This is a fix for a bug which results in the wrong inputs being sent to Cohere Embed API.


nfcampos:
Thanks!

PR #1111, opened 2023-05-03
Allow filter value in ".call" invocation for VectorDBQAChain

brettgrigsby:
- extract "filter" key from ChainValues passed to ".call"
- pass filter value to "similaritySearch" call
- if user does not supply "filter" value it remains "undefined" as before
- if the shape of filter does not conform to value expected by VectorStore, error returns as expected


jacoblee93:
Thank you! Added a small integration test and ran formatting.

PR #1113, opened 2023-05-03
add different templates for different sql engines for SqlDatabaseChain

MaximeThoonsen:


MaximeThoonsen:
This PR is to get closer to the python behavior, I took the templates from python repository.

nfcampos:
Great, thank you!

PR #911, opened 2023-04-20
Add TimeWeightedVectorStoreRetriever to retrievers

winor30:
This pull request introduces the TimeWeightedVectorStoreRetriever class to the retrievers module, which retrieves documents based on their time-weighted relevance. The implementation is based on the reference Python implementation found at https://github.com/hwchase17/langchain/blob/master/langchain/retrievers/time_weighted_retriever.py. The new class includes methods for preparing documents, getting relevant documents, and computing final results based on the combined scores. Additionally, the PR updates the retrievers' index file with new exports for the added class.

PR #1116, opened 2023-05-03
Adds thrown error when using TimeWeightedVectorStoreRetriever on unsupported vector stores

jacoblee93:
Builds off of #911. Also adds docs and another test.


nfcampos:
Cherry-picked onto original PR

PR #1119, opened 2023-05-04
Allow configurable responses for the Search tools

Altaflux:
Currently the search tools response is very limited and tend to only respond back with a single answer. This is typically not enough for many cases where the main answer is either incomplete of irrelevant to the original question.

I have added to the Serper and the SerpAPI search tool's constructor an optional "response extractor" that users of the tool can set to fully customize the output of the search tool. This way each user can decide what type of results they want to include and in which presentation.

A good use-case for this feature is for Agents that use the Browser tool, developers that have setup agents with access to the browser may want the Search tools to return found links so the agent can navigate to those links thru the browser.

Open question:

- The type signature of the extractor function is `(response: any) => Promise<string>;` where response is the raw result of the search call. Should `response` be typed or should can it stay as `any`?

### DIFF
--- a/langchain/src/tools/index.ts
+++ b/langchain/src/tools/index.ts
@@ -1,4 +1,8 @@
-export { SerpAPI, SerpAPIParameters, SerpAPIResponseExtractor } from "./serpapi.js";
+export {
+  SerpAPI,
+  SerpAPIParameters,
+  SerpAPIResponseExtractor,
+} from "./serpapi.js";
 export { DadJokeAPI } from "./dadjokeapi.js";
 export { BingSerpAPI } from "./bingserpapi.js";
 export { Tool, ToolParams, StructuredTool } from "./base.js";
--- a/langchain/src/tools/index.ts
+++ b/langchain/src/tools/index.ts
@@ -1,4 +1,4 @@
-export { SerpAPI, SerpAPIParameters } from "./serpapi.js";
+export { SerpAPI, SerpAPIParameters, SerpAPIResponseExtractor } from "./serpapi.js";
 export { DadJokeAPI } from "./dadjokeapi.js";
 export { BingSerpAPI } from "./bingserpapi.js";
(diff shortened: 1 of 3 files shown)


jacoblee93:
I'm not sure this is worth building in vs. just wrapping it in a `DynamicTool` since the end user would have to write custom logic anyway?

Altaflux:
@jacoblee93 I am not sure I understand the question. The problem I see with Serper and the SerpAPI search tools is that the type of search results the tools return is hardcoded to be built based on a hierarchy of preferred results types, but depending on certain use cases this causes it to omit a lot of information that would have been desirable.

https://github.com/hwchase17/langchainjs/blob/main/langchain/src/tools/serper.ts#L63

An example use case could be an Agent that has access to the browser would want to get link results from the search tools.

My pull requests make this part optionally configurable for those who want different results:
https://github.com/hwchase17/langchainjs/pull/1119/files#diff-67cbaeebc253b29ce7c2cebcb84c6893b4c801c3d9ec0868bbfd411604d61bbeR70

jacoblee93:
Weighing whether it's worth adding complexity by building it into the tool vs.:
```
new DynamicTool({
  name: "web-search",
  description: "Searches links on the web. Takes a search query as input",
  func: async (input) => {
    const result = await fetch("https://serpapi.com?q=" + input);
    // Custom processing
    return "some result";
  }
})
```
You'd need to add custom processing for a response extractor anyway.

Altaflux:
I see what you mean. The problem I see with using DynamicTool is that you have to build the tool's functionality again and looses some of the features already offered by the Serper and the SerpAPI tools like error handling, configuration and whatever might be implemented next.

bracesproul:
@Altaflux going to close as stale/due to merge conflicts. If you'd still like to get this merged in, please open off of an up-to-date main branch, and I'll review. Thank you!

PR #1122, opened 2023-05-04
chore: DRY-up the `loadQAChain` code so it's just using the same functions

justindra:
The `loadQAChain` code was repeating the same code defined in other functions specified already for each type. Just DRY-ing it up to reduce any issues in the future when these functions are updated.


nfcampos:
This is great, thanks!

PR #1008, opened 2023-04-26
feat: add API key support to unstructured loaders

MthwRobinson:
### Summary

Adds support for passing in an Unstructured API key to the Unstructured loaders. Currently, the Unstructured API does not require an API key, but it will in the future. The Unstructured documentation will include instructions on how to obtain an API key once they are available.  Once they are available, the Unstructured team can provide an API key to `langchainjs` can call the API during integration tests.

cc: @cragwolfe and @ryannikolaidis who are working to set up API keys on the Unstructured side.

PR #1128, opened 2023-05-04
Add API key support to Unstructured loaders

jacoblee93:
Extends #1008, adds a temporary shim to avoid breaking people currently using Unstructured document loaders and fixes a bug.


langchain/src/document_loaders/fs/unstructured.ts
```diff
@@ -95,26 +143,34 @@ export class UnstructuredLoader extends BaseDocumentLoader {
 
 export class UnstructuredDirectoryLoader extends DirectoryLoader {
   constructor(
-    public webPath: string,
-    public directoryPath: string,
+    directoryPathOrLegacyApiUrl: string,
+    optionsOrLegacyDirectoryPath: UnstructuredOptions | string,
     public recursive: boolean = true,
```
nfcampos, line 148: I think ideally we move these into the options arg too? @jacoblee93

jacoblee93: Done, left them that way originally to mirror the underlying `DirectoryLoader` syntax but agree that this is nicer.

Issue #1125, opened 2023-05-04
VectorStoreToolkit repeatedly erroring due to "Could not parse LLM output" error


barneyfoxuk:
Hey y'all.

I've been trying to use the VectorStoreToolkit to do some PDF question answering, but unfortunately it almost always (>90% of the time) is giving me the `Error(`Could not parse LLM output: ${text}`);`

I am using GPT-4 and temperature is set to `0`. From previously exploration with agents those settings have been fine previously for getting reliable output, so I'm wondering if there's a code change that's causing the issue (or an update to the model)? 

An example failing prompt is:

```
You are an agent designed to answer questions about sets of documents.\nYou have access to tools for interacting with the documents, and the inputs to the tools are questions.\nSometimes, you will be asked to provide sources for your questions, in which case you should use the appropriate tool to do so.\nIf the question does not seem relevant to any of the tools provided, just return \"I don't know\" as the answer.\n\nNapo_Generic_Policy_Wording_Document: Useful for when you need to answer questions about Napo_Generic_Policy_Wording_Document. Whenever you need information about A document outlining what pet insurance is covered in that policy you should ALWAYS use this. Input should be a fully formed question.\n\nUse the following format:\n\nQuestion: the input question you must answer\nThought: you should always think about what to do\nAction: the action to take, should be one of [Napo_Generic_Policy_Wording_Document]\nAction Input: the input to the action\nObservation: the result of the action\n... (this Thought/Action/Action Input/Observation can repeat N times)\nThought: I now know the final answer\nFinal Answer: the final answer to the original input question\n\nBegin!\n\nQuestion: Age when compulsory excess rises: At what age do customers have to pay more for a claim?\nThought:
```
and the result of that is
```
Exiting LLM run with output: {
  "generations": [
    [
      {
        "text": "I should check the Napo_Generic_Policy_Wording_Document to find information about the age when compulsory excess rises."
      }
    ]
  ]
}
```

So it looks like the issue from that is it not following the ReAct format properly and just returning the `thought`.

Any assistance in solving would be appreciated.

The only thought I had is if the formatting instructions are sent as `system` message, or just as part of the user message? From work I've done separate to Langchain (using my own similar lib I've written) I've found using system messages to specify the  formatting info seemed to work better. You might already be doing that, I haven't gone that deep into the LC code to see, but it might be an avenue to explore if the fix for this bug isn't something obvious. 

Happy to help work on any fix etc if you'd like as this is a blocker for me right now

[...]

PR #1138, opened 2023-05-05
MRKL prompt fix, integration test

### LINKED ISSUE #1042: Unable to parse JSON response from chat agent
Hi im getting an error when trying to use the example from here https://js.langchain.com/docs/modules/agents/agents/examples/chat_mrkl

Im using this template https://github.com/domeccleston/langchain-ts-starter and my node version is 18.16.

The error is this 
image

jacoblee93:
A small prompt change helped cue GPT-4 better. Should fix #1042, and we'll be looking into more robust ways to test agents long-term.


jacoblee93:
This should also fix #1125 since that uses the MRKL agent behind the scenes.

PR #1146, opened 2023-05-06
Move tiktoken to devdeps

nSimonFR:
It seemed unnecessary as a core dependency instead of a devdeps as most of the imports were converted to dynamic.

This should allow users to use their own version of tiktoken or dismiss it completely (Which reduces bundle sizes - but may be redundant with #847 which also does).


nSimonFR:
Lint fails with `import/no-extraneous-dependencies`, not sure if I should allow it with an eslint-disable or not

nfcampos:
Thanks, closing in favour of https://github.com/hwchase17/langchainjs/pull/1239

PR #1149, opened 2023-05-07
fix bug in getNumTokensFromMessages

pczekaj:
getNumTokensFromMessages wasn't doing the same thing as described in: https://github.com/openai/openai-cookbook/blob/main/examples/How_to_format_inputs_to_ChatGPT_models.ipynb

because it didn't include in calculation message role, name content and didn't add 3 to total count


jacoblee93:
Thanks for the catch!

PR #1150, opened 2023-05-07
Router chain (MultiRetrievalQAChain & MultiPromptChain)

ppramesi:
Added MultiRetrievalQAChain and MultiPromptChain. Translated from the python's version (https://github.com/hwchase17/langchain/tree/master/langchain/chains/router)

MultiRetrievalQAChain allows an LLM to pick between a number of different retrievers to pass the input to. 
MultiPromptChain allows an LLM to pick between a number of different prompts to pass the input to.

The router chain works slightly differently than the python's version because we can directly define the structure of the next_inputs by defining a zod schema, use the schema to create StructuredOutputParser and incorporate StructuredOutputParser's getMarkdownJsonInstructionFromSchema into the template.


nfcampos:
Thanks!

PR #1153, opened 2023-05-07
Add Redis VectorStore

mvaker:
### Redis VectorStore
Integration with the Redis VectorStore (RediSearch).

## Side-note
This is my first contribution. I followed the contribution guidelines as closely as possible but it might be that I missed something. Feel free to tell me if so 🙂


langchain/src/vectorstores/index.ts
```diff
@@ -8,3 +8,4 @@ export { PineconeStore } from "./pinecone.js";
 export { VectorStore, SaveableVectorStore } from "./base.js";
 export { SupabaseVectorStore } from "./supabase.js";
 export { PrismaVectorStore } from "./prisma.js";
+export { RedisVectorStore } from "./redis.js";
```
ewfian, line 11: This index file is deprecated, see: https://github.com/hwchase17/langchainjs/pull/685#discussion_r1166712827

docs/docs/modules/indexes/vector_stores/integrations/redis.md
```diff
@@ -0,0 +1,122 @@
+# Redis
+
+[Redis](https://redis.io/) is a fast open source, in-memory data store.
+As part of the [Redis Stack](https://redis.io/docs/stack/get-started/), [RediSearch](https://redis.io/docs/stack/search/) is the module that enables vector similarity semantic search, as well as many other types of searching.
+
+Langchain.js accepts [node-redis](https://github.com/redis/node-redis) as the client for Redis vectorstore.
+
```
ewfian, line 7: Maybe need a compatibility tips here.
https://github.com/hwchase17/langchainjs/blob/5ab15052177b641fd56685739bac75ab95c970d1/docs/docs/modules/indexes/vector_stores/integrations/hnswlib.mdx?plain=1#L8-L12

langchain/src/vectorstores/index.ts
mvaker: Fixed

docs/docs/modules/indexes/vector_stores/integrations/redis.md
mvaker: I think it's not necessary since it's already available in the Python package. See:
- [the docs](https://python.langchain.com/en/latest/modules/indexes/vectorstores/examples/redis.html)
- [the repo](https://github.com/hwchase17/langchain/blob/master/langchain/vectorstores/redis.py)

ewfian: My understanding is: this tips indicates it can only be used in the Node.js environment, not in the browser and other environments. [Support additional JS environments? #152](https://github.com/hwchase17/langchainjs/discussions/152)

For example, milvus has implementations on both js and Python, but it has the tips.
https://js.langchain.com/docs/modules/indexes/vector_stores/integrations/milvus
https://python.langchain.com/en/latest/modules/indexes/vectorstores/examples/milvus.html

mvaker: Oh I see... then my bad. I totally misunderstood it. Then, it should indeed have the tip then since it can only be used in the Node.js env.
I've added it now, thanks for the heads up!

examples/src/indexes/vector_stores/redis/redis_query.ts
```diff
@@ -0,0 +1,65 @@
...
+*/
+
+/* Search in the vector DB using filters */
+const filterRes = await vectorStore.similaritySearch("redis", 3, ["qux"]);
+console.log(filterRes);
+/*
+[
+  Document {
+    pageContent: "consectetur adipiscing elit",
+    metadata: { baz: "qux" },
```
nfcampos, line 36: How will this metadata filtering work when there are multiple keys in metadata?

nfcampos (commented):
Just one question, lgtm otherwise, thanks!

examples/src/indexes/vector_stores/redis/redis_query.ts
mvaker: The implemented filtering is only text-based for now. The filter matches the metadata that contains any of the provided strings in the filters array.
For example, with two records having the following metadata:
1. `metadata: { foo: 'bar', baz: 'qux' }`
2. `metadata: { bar: 'bla' }`

A search with `filter = ['bar']` will return both results because they both have 'bar' in their stringified metadata. It checks for either a value or a key containing that text.

Redis has more complex ways of filtering but this one was the most straightforward to implement and the most useful one imo.
An alternative to allow more complex filtering without overcomplicating the library would be making `filter` a `string` and allow providing a [Redis query syntax string](https://redis.io/docs/stack/search/reference/query_syntax/#a-few-query-examples).

tylerhutcherson: Part of the challenge(?) here is that Redis expects you to define the schema declaratively in your index construction. This was by design, and ultimately gives devs more power. But expressing this through LangChain (or any abstraction layer) makes it a bit harder to expose without lot's of additional code. We're ideating this concept on the LC python lib now.

Allowing a user to pass in a Redis query string for more complex metadata filtering would be interesting.. but also still wouldn't handle anything beyond the stringified metadata field right? (redis also supports Tag fields, Numeric fields, in addition to Text).

@nfcampos @mvaker my recommendation would be we start here and then work on a full, unified implementation for metadata filtering with Redis + LangChain across all languages.

mvaker: Thanks for your comment @tylerhutcherson. It’s true that a query string wouldn’t handle anything beyond the stringified metadata field, but ultimately it would let users to use more complex syntax such as negative operator, AND queries, and so on.
Of course, having a full, unified implementation for metadata filtering would be great.
Let’s build up from here.

PR #1156, opened 2023-05-08
Fix the problem that OutputFixingParser parse directly throws an error when it encounters a ZodError

Vinlic:
When encountering a ZodError it should first try to fix the output instead of throwing an error.

```
    z.object({
      answer: z.string().describe("answer to the user's question"),
      sources: z
        .array(z.string())
        .describe("sources used to answer the question, should be websites."),
    })
  );
  const badOutput = `\`\`\`json
  {
    "answer": "foo",
    "sources": "foo.com"
  }
  \`\`\``;
const fixParser = OutputFixingParser.fromLLM(
    new ChatOpenAI({ temperature: 0 }),
    parser
  );
  const output = await fixParser.parse(badOutput);  //throw ZodError
  console.log("Fixed output: ", output);
```


jacoblee93:
Output parsers should all throw an `OutputParserException` upon failures like this - Zod is an implementation detail rather than something we should track at a higher level here.

Is there a specific parser that's not behaving like this?

khanguslee:
@jacoblee93 Although I agree that this PR isn't necessary, I was wondering why the `fixParser.parse` throws a zod error. I would assume that the system would be able to fix itself after running it a second time. Is this expected behaviour? I am running "gpt-4".

PR #1169, opened 2023-05-08
add mssql to the compatible database list for SqlDatabaseChain

MaximeThoonsen:


jacoblee93:
Looks good to me, just removed an unnecessary dev dependency. Thank you!

PR #1174, opened 2023-05-08
Add V2 Tracer

vowelparrot:
Add V2 Tracer.
- Add an 'integration test' (no assertions though)


langchain/src/callbacks/handlers/tracers.ts
```diff
@@ -55,8 +75,33 @@ export interface ToolRun extends BaseRun {
 
 export type Run = LLMRun | ChainRun | ToolRun;
 
+export interface RunV2Base {
+  id: string; // v4 uuid
```
nfcampos, line 79: Why would we care what kind of uuid it is here

langchain/src/callbacks/handlers/tracers.ts
```diff
@@ -55,8 +75,33 @@ export interface ToolRun extends BaseRun {
...
+  serialized: object;
+  inputs: RunInputs;
+  outputs: RunOutputs;
+  session_id: string; // v4 uuid
+  reference_example_id?: string; // v4 uuid
+  run_type: RunType;
+}
+
+export interface RunV2Create extends RunV2Base {
+  name?: string;
```
nfcampos, line 94: What's the name of a run?

langchain/src/callbacks/handlers/tracers.ts
```diff
@@ -458,3 +514,158 @@ export class LangChainTracer extends BaseTracer {
...
+    } else if (run.type === "chain") {
+      const chainRun = run as ChainRun;
+      inputs = chainRun.inputs;
+      outputs = chainRun.outputs ?? {};
+      child_runs = [
+        ...chainRun.child_llm_runs,
+        ...chainRun.child_chain_runs,
+        ...chainRun.child_tool_runs,
+      ];
+    } else {
```
nfcampos, line 625: I think this should be else if?

langchain/src/callbacks/handlers/tracers.ts
```diff
@@ -458,3 +514,158 @@ export class LangChainTracer extends BaseTracer {
...
+        ...toolRun.child_tool_runs,
+      ];
+    }
+
+    return {
+      id: run.uuid,
+      name: run.serialized.name,
+      start_time: run.start_time,
+      end_time: run.end_time,
+      extra: {}, // TODO: VWP
```
nfcampos, line 642: What will this be used for?

langchain/src/callbacks/handlers/tracers.ts
```diff
@@ -458,3 +514,158 @@ export class LangChainTracer extends BaseTracer {
...
+      error: run.error,
+      execution_order: run.execution_order,
+      serialized: run.serialized,
+      inputs,
+      outputs,
+      session_id: session.id,
+      run_type: run.type,
+      reference_example_id: this.exampleId,
```
nfcampos, line 650: This isn't right, it would leave the same example as reference example for the root run and all its children runs, it should apply to the root run only

nfcampos (commented):
Some comments. Overall is fine, but I don't agree with using deprecated v1 runs everywhere and converting to v2 only in the v2 tracer. It should be the other way around, use the new format everywhere and only convert to v1 in deprecated v1 tracer

langchain/src/callbacks/handlers/tracers.ts
vowelparrot: Not strongly. Meant to differentiate between the number `id` of the v1 run. I can remove the comments entirely if that's cleaner

langchain/src/callbacks/handlers/tracers.ts
vowelparrot: That's a great point

langchain/src/callbacks/handlers/tracers.ts
vowelparrot: On the python side, there are kwargs in the `on_{type}_start` that are passed in as `extra` values in the run

langchain/src/callbacks/handlers/tracers.ts
vowelparrot: Yeah `else if` then `else {throw ...}`?

vowelparrot:
> Some comments. Overall is fine, but I don't agree with using deprecated v1 runs everywhere and converting to v2 only in the v2 tracer. It should be the other way around, use the new format everywhere and only convert to v1 in deprecated v1 tracer


Yeah I was hesitant to impact the v1 tracer definitions while we're finishing up the v2 tracing definition, but we can chat about this today.

langchain/src/callbacks/handlers/tracers.ts
```diff
@@ -458,3 +595,123 @@ export class LangChainTracer extends BaseTracer {
...
+      end_time: run.end_time,
+      run_type: run.run_type,
+      reference_example_id: example_id,
+      extra: run.extra,
+      execution_order: run.execution_order,
+      serialized: run.serialized,
+      error: run.error,
+      inputs: run.inputs,
+      outputs: run.outputs,
+      session_id: run.session_id,
```
vowelparrot, line 692: There must be a better way of initializing objects like `...run` with everything except 1 or 2 fields. Will look into it

langchain/src/callbacks/handlers/tracers.ts
```diff
@@ -371,11 +413,102 @@ export class LangChainTracer extends BaseTracer {
...
+  protected async persistRun(
+    run: RunV2 | LLMRun | ChainRun | ToolRun
+  ): Promise<void> {
     let endpoint;
-    if (run.type === "llm") {
+    let v1Run: LLMRun | ChainRun | ToolRun;
+    if ((run as RunV2).run_type !== undefined) {
+      v1Run = await this.convertV2RunToRun(run as RunV2);
+    } else {
+      v1Run = run as LLMRun | ChainRun | ToolRun;
```
nfcampos, line 507: How would we get here?

langchain/src/callbacks/handlers/tracers.ts
nfcampos: Which fields are you trying to remove?

nfcampos: Note the backend actually needs to accept (and ignore) extra unknown fields for forwards compatibility (if it doesn't we need to fix that @agola11)

vowelparrot: Oh OK that makes sense. Really the `parent_run_id` and `child_execution_order` fields

langchain/src/callbacks/handlers/tracers.ts
vowelparrot: It should never get here. I could actually probably update the `persistRun` argument typing  - thanks for pointing that out

langchain/src/callbacks/handlers/tracers.ts
agola11: extra/unknown fields are already ignored. (This is the default behavior of pydantic btw)

vowelparrot: Hm If I do 
```ts
const persistedRun: RunV2Create = {
      id: run.id,
      name: run.name,
      start_time: run.start_time,
      end_time: run.end_time,
      run_type: run.run_type,
      extra: run.extra,
      execution_order: run.execution_order,
      serialized: run.serialized,
      error: run.error,
      inputs: run.inputs,
      outputs: run.outputs,
      session_id: run.session_id,
      reference_example_id: example_id,
      child_runs: await Promise.all(
        run.child_runs.map((child_run) => this._convertToCreate(child_run))
      ),
    };
    ```
    it works but if I do
    
    ```
    const persistedRun: RunV2Create = {
      ...run,
      reference_example_id: example_id,
      child_runs: await Promise.all(
        run.child_runs.map((child_run) => this._convertToCreate(child_run))
      ),
    };
```

It returns a 404

langchain/src/callbacks/handlers/tracers.ts
```diff
@@ -55,10 +74,39 @@ export interface ToolRun extends BaseRun {
 
 export type Run = LLMRun | ChainRun | ToolRun;
 
+export interface BaseRunV2 {
+  id: string;
+  name: string;
+  start_time: number;
+  end_time: number;
+  extra: object;
```
agola11, line 82: this should be optional and some version of `Record<string, Any>`

langchain/src/callbacks/handlers/tracers.ts
```diff
@@ -55,10 +74,39 @@ export interface ToolRun extends BaseRun {
...
+  error?: string;
+  execution_order: number;
+  serialized: object;
+  inputs: RunInputs;
+  outputs: RunOutputs;
```
agola11, line 87: should also be optional

langchain/src/callbacks/handlers/tracers.ts
```diff
@@ -159,17 +199,22 @@ export abstract class BaseTracer extends BaseCallbackHandler {
...
+      name: llm.name,
+      extra: {},
+      parent_run_id: parentRunId,
       start_time: Date.now(),
       end_time: 0,
       serialized: llm,
-      prompts,
-      session_id: this.session.id,
+      inputs: { prompts },
+      outputs: {},
```
agola11, line 212: this and `extra` can go away once optional

langchain/src/callbacks/handlers/tracers.ts
```diff
@@ -55,10 +74,39 @@ export interface ToolRun extends BaseRun {
...
+export interface RunV2Create extends BaseRunV2 {
+  child_runs: RunV2Create[];
+}
+export interface RunV2 extends RunV2Create {
+  child_execution_order: number;
+  child_runs: RunV2[];
+  parent_run_id?: string; // uuid
+}
+
+export interface AgentRunV2 extends RunV2 {
```
agola11, line 102: Why do we need this?

agola11 (approved):
overall looks great. It's a little weird to me that we are calling this `RunV2` though, maybe just `Run` or `LangChainRun`? Not a big deal though, know it's annoying to change rn

nfcampos:
I agree that it would be good if we don't name the new interface `RunV2` and instead call it just `Run`

langchain/src/callbacks/handlers/tracers.ts
vowelparrot: @agola11 FYI currently still has to be provided - I'll update the backend to let it be optional

agola11: sg, my bad for overlooking that

PR #1177, opened 2023-05-09
ignore like .gitignore in GithubLoader

tonisives:
The current GithubLoader implementation only ignores specific files with exact match.

This PR uses .gitignore style file ignoring, with which you can ignore folders and all files with the same(.gitignore = .gitignore in all subfolders, not only root path)

### Removes
https://github.com/hwchase17/langchainjs/pull/1177

### Adds
Uses the https://www.npmjs.com/package/ignore package to ignore Github files.
User can input array of .gitignore file lines

### Note
I am not an expert with this project layout, and didn't have time to run tests. However, I implemented this locally(outside of langchain project) and it worked fine. But maybe I missed something when adding these lines in to the PR.

This means this is a general idea how to implement it. It should be tested


tonisives:
> Hey, thanks for this PR! I was looking out for this feature but turns out it's not implemented yet.
> 
> My 2 cents:
> 
> * Relocate the ignore package: 
> * Adjust the dependency type: 

Thanks for the information! I made these changes

langchain/package.json
```diff
@@ -410,6 +410,7 @@
     "epub2": "^3.0.1",
     "hnswlib-node": "^1.4.2",
     "html-to-text": "^9.0.5",
+    "ignore": "^5.2.0",
```
jacoblee93, line 413: This also needs to go under `devDependencies`

langchain/src/document_loaders/web/github.ts
```diff
@@ -32,7 +33,7 @@ export interface GithubRepoLoaderParams {
   recursive?: boolean;
   unknown?: UnknownHandling;
   accessToken?: string;
-  ignoreFiles?: (string | RegExp)[];
+  ignorePaths?: string[]
```
jacoblee93, line 36: This will break anyone currently using this functionality. Let's make a new argument name instead and deprecate this one.

jacoblee93:
Ah unfortunately given that this loader is exported from the main `langchain/document_loaders` file, adding `ignore` as a dependency here would require us to add it as a core dependency. If we added at as a peer dependency, everyone currently using this would break.

We have a few options:
1. Treat the old GitHub loader as deprecated and create a new one that requires `ignore` as a peer dependency
2. Add `ignore` as a core dependency
3. Find a way to implement this without a dependency

All are not ideal but I think option 2 is actually our best bet since it's 0 dep and quite popular - let me think a bit more though.

jacoblee93:
After discussion with @nfcampos, we'll merge this as is and expect folks to add the dependency when upgrading versions.

Thank you @tonisives! We'll merge tomorrow.

Issue #1159, opened 2023-05-08
Stop streaming call on an LLM chain mid-generation


Emsu:
We have a realtime conversation use case where we need to cancel generations and modify the bot response in the chain history (eg. because the bot didn't finish speaking so history should have a partial sentence).

If a user wants to interrupt an LLM in the middle of the response stream and start a new call, there's no way currently. I have a real time use case where we may want to interrupt a call. Ideally there would be something returned when we run `chain.call` that returns a cancellation callback. Right now we unhook from handleNewToken, but the generation continues and saveContext is still called.

For this type of real time use case, I think it would be nice to have:
- A cancel function return on a `chain.call` with `stream=true`. Could be cancelling the whole function or maybe stopping the axios request.
- The ability to stop saveContext with the cancellation. Would pair nicely with #1158 

Are these on the roadmap or potentially something we could help implement?

nfcampos:
Hi @Emsu we already have this functionality when calling the model directly, let me know if that's what you're after https://js.langchain.com/docs/modules/models/llms/additional_functionality#cancelling-requests 
We're working on adding it to `chain.call` as well

Emsu:
@nfcampos oh awesome that's what I'm after. Is there a way to get that today from the chain as a workaround? Is there a way we can help with this?

Emsu:
@nfcampos I'm not sure if this is the way you wanted to implement but I added a PR. I saw you've self-assigned as well now so feel free to put in a different implementation.

jacoblee93:
Thanks again for raising the issue - support was merged here: https://github.com/hwchase17/langchainjs/pull/1182

PR #1180, opened 2023-05-09
Add ability to pass in signal to the call via ChainValues

Emsu:
Should be able to call with `chain.call({ input: "Hi", options: { signal: abortController.signal }})`

@nfcampos not sure if this is how you want it implemented but this should pass through to the openai llm #1159


nfcampos:
@Emsu yes, thank you, its something like that that I'm thinking about, just need to make it more generic for other LLMs too, I'll do that on top of this branch

nfcampos:
I've opened #1182 for this

Issue #262, opened 2023-03-09
Add vector store adapter for Voy


nfcampos:
https://github.com/tantaraio/voy

A WASM vector store written in Rust, so a good choice for non-Node envs

kwkr:
Could you assign it to me @nfcampos ? I would like to start working on it.

nfcampos:
Done, thanks!

kwkr:
Merged in #1181

PR #1181, opened 2023-05-09
add initial voy implementation

kwkr:
This addresses #262 
Hello, @nfcampos ! I added some initial implementation but still have a few questions:

1. Should the implementation be based on some specific abstract class? I saw `VectorStore` and `SaveableVectorStore`. I went with the first for now but I would be glad to hear some opinion here. Since `voy-search` is meant for browser env, I would say it doesn't make much sense to save the index data to drive.
2. What is the preferred approach for using external clients? I saw two different approaches: the first is to inject the client within the constructor (for example `langchain/src/vectorstores/supabase.ts`), and the second is to instantiate the class from the external library within the initializing code (for example `langchain/src/vectorstores/hnswlib.ts`). Which one should be used here?

Thanks for your help!

TODO: for now it's missing integration tests and docs for the newly added class


nfcampos:
Hi @kwkr sorry for the delay in getting back to you here, in answer to your questions
1. Yes I agree SaveableVectorStore makes most sense for Node.js vector store
2. I think we prefer to have the client as a argument, and people instantiate outside, how you've done it here
This is looking good, with some docs and integration tests we can get this in, thanks!

kwkr:
I needed to patch the `voy-search` package in order to be able to import it within integration tests and provide a working example script:
https://github.com/hwchase17/langchainjs/pull/1181/files#diff-300cdd47b9511e809b968c658e0ea9e1267ec27afb73f50310bcead507ae07adR8-R9

jacoblee93:
This is awesome - thanks for following through!

PR #1182, opened 2023-05-09
Implement cancellation and timeout for all LLMs, allow passing LLM call options via LLMChain (like was already possible for stop)

nfcampos:


langchain/src/chains/llm_chain.ts
```diff
@@ -86,19 +86,35 @@ export class LLMChain<T extends string | object = string>
     return finalCompletion;
   }
 
+  /**
+   * Run the core logic of this chain and add to output if desired.
+   *
+   * Wraps _call and handles memory.
+   */
+  call(
+    values: ChainValues & this["llm"]["CallOptions"],
```
jacoblee93, line 95: My ideal signature here (when v1 comes around?) would be:
```
call(
  values: ChainValues
  options: this["llm"]["CallOptions"] // Includes "callbacks?"
)
```
So as not to have any restricted/meta properties as I'm sure we'll keep adding more and more. Would that make sense long-term?

Otherwise, if this is the long-term solution, could it make sense to make just `options` or `callOptions` (and I guess `stop`) a special value within `values` here?
```
call(
  values: ChainValues & {
    options: this["llm"]["CallOptions"]
  }
)
```
I see you have something like this below for `openai-chat` but building it in lower level might make sense to me if this is the long-term fix.

langchain/src/llms/base.ts
```diff
@@ -54,28 +54,28 @@ export abstract class BaseLLM extends BaseLanguageModel {
...
-    stop?: string[] | this["CallOptions"],
+    options?: Omit<this["CallOptions"], "timeout">,
     runManager?: CallbackManagerForLLMRun
   ): Promise<LLMResult>;
 
   /** @ignore */
   async _generateUncached(
     prompts: string[],
-    stop?: string[] | this["CallOptions"],
+    options?: this["CallOptions"],
```
jacoblee93, line 78: I think this should technically be `Omit<this["CallOptions"], "timeout">`?

langchain/src/llms/openai.ts
```diff
@@ -251,15 +255,10 @@ export class OpenAI extends BaseLLM implements OpenAIInput, AzureOpenAIInput {
    */
   async _generate(
     prompts: string[],
-    stopOrOptions?: string[] | this["CallOptions"],
+    options?: this["CallOptions"],
```
jacoblee93, line 258: Should be `Omit<this["CallOptions"], "timeout">`

langchain/src/llms/base.ts
```diff
@@ -37,7 +37,7 @@ export interface BaseLLMCallOptions extends BaseLanguageModelCallOptions {}
  * LLM Wrapper. Provides an {@link call} (an {@link generate}) function that takes in a prompt (or prompts) and returns a string.
  */
 export abstract class BaseLLM extends BaseLanguageModel {
-  declare CallOptions: BaseLanguageModelCallOptions;
+  declare CallOptions: BaseLLMCallOptions;
```
jacoblee93, line 41: How do you feel about this so we don't need `Omit` everywhere?
```
  declare ParsedCallOptions: Omit<this["CallOptions"], "timeout">
```

langchain/src/llms/base.ts
```diff
@@ -229,20 +237,20 @@ export abstract class LLM extends BaseLLM {
...
     prompt: string,
-    stop?: string[] | this["CallOptions"],
+    options?: Omit<this["CallOptions"], "timeout">,
     runManager?: CallbackManagerForLLMRun
   ): Promise<string>;
 
   async _generate(
     prompts: string[],
-    stop?: string[] | this["CallOptions"],
+    options?: Omit<this["CallOptions"], "timeout">,
```
jacoblee93, line 246: I think we can just make this required since it's always called via `generate` if we always pass in an options object - one less case to worry about

langchain/src/llms/base.ts
```diff
@@ -229,20 +237,20 @@ export abstract class LLM extends BaseLLM {
    */
   abstract _call(
     prompt: string,
-    stop?: string[] | this["CallOptions"],
+    options?: Omit<this["CallOptions"], "timeout">,
```
jacoblee93, line 240: Wait `call` doesn't actually call `_call`?

langchain/src/util/async_caller.ts
```diff
@@ -100,6 +104,27 @@ export class AsyncCaller {
     );
   }
 
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  callWithOptions<A extends any[], T extends (...args: A) => Promise<any>>(
+    options: AsyncCallerCallOptions,
```
jacoblee93, line 109: Options second?

jacoblee93:
Hey @nfcampos, implemented several of the discussed changes here: https://github.com/jacoblee93/langchainjs/commit/953f848dc3d30c95f61871a6e6bdfd06eab80630

I didn't want to just directly merge into your branch before you looked at them.

langchain/src/chains/llm_chain.ts
agola11: I think it could make sense to add callbacks in call options

agola11 (commented):
overall looks good

langchain/src/llms/base.ts
nfcampos: Agreed

langchain/src/util/async_caller.ts
nfcampos: No, because with of the rest args signature it can't come last, so must come first

langchain/src/chains/llm_chain.ts
nfcampos: Yea, the thing is only language models have call options, whereas lots of other objects in the library have callbacks, so not sure about that

PR #1184, opened 2023-05-09
feat: add basic duckduckgo search tool

BEN00262:


langchain/package.json
```diff
@@ -31,6 +31,9 @@
     "tools/webbrowser.cjs",
     "tools/webbrowser.js",
     "tools/webbrowser.d.ts",
+    "tools/duckduckgo.cjs",
```
dqbd, line 34: Consider adding the tool into `tsconfig.json`

langchain/src/tools/duckduckgo.ts
```diff
@@ -0,0 +1,26 @@
...
+import { convert as htmlToText } from 'html-to-text';
+import { Tool } from "./base.js";
+
+export class DuckDuckGo extends Tool {
+    name = "search"
+
+    /** @ignore */
+    async _call(input: string) {
+        const searchResults = await DDGsearch(input, {
+            safeSearch: SafeSearchType.STRICT
```
dqbd, line 11: _nit_: could this be configurable?

langchain/src/tools/duckduckgo.ts
```diff
@@ -0,0 +1,26 @@
+import { search as DDGsearch, SafeSearchType } from 'duck-duck-scrape';
```
dqbd, line 1: Use dynamic import whenever possible, as this is being imported into the barrel file

dqbd, line 1: Consider adding the tool into `langchain/src/tools/index.ts` as well

dqbd (changes requested):
Nice! If possible, could you add examples and docs entry?

Clad3815:
What you think about using LLM to get only the data needed from the search results ? I made the DDG tool some days ago with this approch, if it's can help:

(I use some keyword like "google" because i'm using GPT3.5 and I found it better understanding like that)

```js
export class DDG extends Tool {
  model: ChatOpenAI;
  embeddings: Embeddings;
  name = "google-search";
  description = `A search engine (Google) connected to the internet. Input: a single, specific, SEO-optimized query. Output: an answer with relevant links. Avoid complex or multi-part queries for optimal results.`;

  constructor(params: { model: ChatOpenAI; embeddings: Embeddings }) {
    super();
    this.model = params.model;
    this.embeddings = params.embeddings;
  }

  async _call(input: string) {
    try {
      let googleResult: any = await search(input, {
        safeSearch: SafeSearchType.STRICT,
      });
      if (googleResult.noResults) {
        return "No good search result found";
      }

      let googleResults = googleResult.results;
      googleResults = googleResults.map(
        ({ title, description, url }: SearchResult) => ({
          title,
          body: description.replace(/<.*?>/g, ""),
          href: url,
        })
      );

      // Keep only the first 10 results
      googleResults = googleResults.slice(0, 10);

      let date = new Date();
      let fullDate = `Date: ${date.getDate()}/${
        date.getMonth() + 1
      }/${date.getFullYear()}, Time: ${date.getHours()}:${date.getMinutes()}:${date.getSeconds()}`;
      const prompt = `
${fullDate}
Question: ${input} 
Task: Find the best match result based on the user question and try to answer the question, if not possible provide up to 5 markdown links from within that would be of interest (always including URL and text). Links should be provided, if present, in markdown syntax as a list under the heading "Relevant Links:". 
You have to return a sentence which answer to the question, or the markdown links.

The markdown links must be only included if they are the same topic as the question.
You have to ensure the data got are current data, and not outdated.
Results: 
${JSON.stringify(googleResults)}`;

      const gptResponse = await this.model.call([new HumanChatMessage(prompt)]);

      return gptResponse.text;
    } catch (error) {
      console.log(error);
      return "No good search result found";
    }
  }
}

```

3a1b2c3:
Any progress?

jacoblee93:
Merged in #4943

PR #1191, opened 2023-05-10
Scrape all pages, not only level 1 navigation

emilsedgh:
Gitbook's server-side rendering means that when we fetch the contents of the index page
the sub-menu items are not rendered and thus are invisible to $('nav a').
This patch makes the loader use sitemap.xml which has an index of all pages.

Inspired by the Python version.

### DIFF
--- a/langchain/src/document_loaders/tests/gitbook.int.test.ts
+++ b/langchain/src/document_loaders/tests/gitbook.int.test.ts
@@ -7,13 +7,13 @@ test("Test GitbookLoader", async () => {
   );
 
   const docs = await loader.load();
-  console.log(docs);
+  console.log('Loaded', docs.length, 'Gitbook documents');
 });
 
 test.only("Test GitbookLoader with shouldLoadAllPaths", async () => {
-  const loader = new GitbookLoader("https://docs.gitbook.com", {
+  const loader = new GitbookLoader("https://docs.maildrop.cc", {
     shouldLoadAllPaths: true,
   });
   const docs = await loader.load();
-  console.log(docs);
+  console.log('Loaded', docs.length, 'Gitbook documents');
 });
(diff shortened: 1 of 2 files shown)


emilsedgh:
This is a fairly simple change that fixes the `shouldLoadAllPaths` parameter. Is it possible for it to be reviewed? The python version uses the same logic to make sure all pages are scraped.

dqbd (approved):

dqbd:
This should match the behaviour found in the original Python version, lgtm :)

emilsedgh:
Thank you David.

jacoblee93:
I can't push to your branch, but this change breaks the canonical example in the docs. Can we make sure both are supported? Maybe use both sources and dedupe?

I added this short integration test:

`langchain/src/document_loaders/tests/gitbook.int.test.ts`
```
import { test } from "@jest/globals";
import { GitbookLoader } from "../web/gitbook.js";

test("Test GitbookLoader", async () => {
  const loader = new GitbookLoader(
    "https://docs.gitbook.com/product-tour/navigation"
  );

  console.log(docs);
})

test.only("Test GitbookLoader with shouldLoadAllPaths", async () => {
  const loader = new GitbookLoader("https://docs.gitbook.com", {
    shouldLoadAllPaths: true,
  });
  console.log(docs);
});
```

Martinlopez-hash:
Playson

emilsedgh:
@jacoblee93 Thank you for the test. I pushed it to the branch.

How does this break the example?

jacoblee93:
> @jacoblee93 Thank you for the test. I pushed it to the branch.
> 
> How does this break the example?

It doesn't return any links, while the old code does.

jacoblee93:
Thank you!

PR #1195, opened 2023-05-10
feat: add ical tool

rpidanny:
Adds a new `ical` tool that allows agents to create calendar events.

I'm open to suggestions for other names for the tool if you have a better idea.


langchain/src/tools/ical.ts
```diff
@@ -0,0 +1,115 @@
...
+    this.llm = llm;
+  }
+
+  /** @ignore */
+  async _call(input: string): Promise<string> {
+    try {
+      const { name, events } = await this.getEventsObject(input);
+      const { ical } = await this.imports();
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
```
dqbd, line 53: AFAICT ical-generator does provide typescript definitions, no need to add @ts-ignore

dqbd (approved):
Looks great! Only a small nitpick

langchain/src/tools/ical.ts
rpidanny: I had to add that because of `ts-jest`. It strangely doesn't seem to pick up the types. 

Screenshot 2023-05-12 at 00 37 54

dqbd: Understandable, let's keep it that way.

langchain/src/tools/ical.ts
```diff
@@ -0,0 +1,104 @@
...
+          location: event.location,
+          url: event.url,
+        });
+      }
+
+      await this.store.writeFile(`${name}.ics`, calendar.toString());
+
+      return `${description} event created successfully and saved as ${name}.ics.`;
+    } catch (err) {
+      return "Failed to create calendar event.";
```
jacoblee93, line 72: What happens if this errors due to something non-retryable? I suspect the agent may just continually retry.

Might be better to just throw it so that execution stops.

examples/src/tools/ical.ts
```diff
@@ -0,0 +1,21 @@
+import { ICalTool } from "langchain/tools/ical";
```
jacoblee93, line 1: Would prefer to have an end-to-end example here with an agent.

langchain/src/tools/ical.ts
```diff
@@ -0,0 +1,104 @@
...
+      action: inputStr,
+    });
+
+    const response = await this.llm.call(input);
+
+    try {
+      const eventsObject = parser.parse(response);
+      return eventsObject;
+    } catch (err) {
+      console.error(err);
```
jacoblee93, line 100: Remove extraneous log

langchain/src/tools/tests/ical.test.ts
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  async _call(prompt: string): Promise<string> {
+    return prompt;
+  }
+}
+
+describe("ICalTool", () => {
+  let tool: ICalTool;
+  let store: BaseFileStore;
```
jacoblee93, line 25: Could we make this a tool for a given provider (e.g. Google Calendar, Outlook) instead?

A little bit uncertain about allowing tools to rely on the existence of a filesystem, and am not certain about the utility of storing these locally anyway.

langchain/src/tools/ical.ts
```diff
@@ -0,0 +1,104 @@
+import { z } from "zod";
+import { ICalCalendar } from "ical-generator";
```
jacoblee93, line 2: We can just do this directly now if we use a separate entrypoint!

PR #1196, opened 2023-05-10
Run the Client over a Dataset

vowelparrot:
Part 3 of the client PRs.
- Part 1 https://github.com/hwchase17/langchainjs/pull/1174 creates the V2 tracer and breaks the type signatures of the BaseTracer 
- Part 2 https://github.com/hwchase17/langchainjs/pull/1190 adds the client along with creation / read / deletion methods for datasets and examples. 

This PR adds a method to run a chain or LLM over a dataset.

It doesn't yet introduce concurrency or properly support chat models. I'll try to add that in a fourth PR (this is too big already).


agola11 (approved):
This looks good, but would be great to add an example in the experimental section of examples. Can honestly just copy a lot of your int test

Issue #1200, opened 2023-05-10
ZeroShotAgentOutputParser trims quotes regardless of whether they match


jollytoad:
ZeroShotAgentOutputParser strips all leading and trailing double quotes from the tool input string, regardless of whether they actually match, this can produce an unexpected effect such as an infinite loop.

For example, if the following is generated:

`Action Input: foo = "Foo"`

and you have that fed into a tool to validate the syntax, the regex in ZeroShotAgentOutputParser mistakenly strips the last quote, so the tool gets:

`foo = "Foo`

and the tool will return an error due to the invalid syntax, if that error states the closing quote is missing, then the feedback misleads the model, and then the next generation may include additional quotes, which also gets stripped, and you end up in a feedback loop of nonsense.

I think a less surprising behaviour would be to trim matching quotes only, so that the quotes are not stripped from the above, but it would strip something like `"""hello"""` for example.

This is the existing replacement: `.replace(/^"+|"+$/g, "")`, that strips all leading and trailing quotes regardless.

And this is an alternative: `.replace(/^(\"+)(.*?)(\1)$/, '$2')`, that will only strip matching leading and trailing quotes.

PR #1201, opened 2023-05-10
Strip matching quotes only in output parser #1200

jollytoad:
Addresses #1200 with an alternative regular expression that strips only leading and trailing quotes that balance.


nfcampos:
Thanks!

PR #1216, opened 2023-05-11
feat: allow passing in verbose flag in loadSummarizationChain

justindra:
Expose the `verbose` option to the `loadSummarizationChain` similar to other `loadX` functions.


nfcampos:
Thanks!

PR #1218, opened 2023-05-11
Add Chat Message Handling

vowelparrot:
- Add `handleChatModelStart` to callback handler and manager
- Fallback to `handleLLMStart` if not implemented

Still needs tests - will add tomorrow

### DIFF
--- a/langchain/src/callbacks/handlers/tracers.ts
+++ b/langchain/src/callbacks/handlers/tracers.ts
@@ -1,6 +1,7 @@
 import * as uuid from "uuid";
 import {
   AgentAction,
+  BaseChatMessage,
   ChainValues,
   LLMResult,
   RunInputs,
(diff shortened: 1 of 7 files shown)


langchain/src/callbacks/handlers/tracers.ts
```diff
@@ -220,6 +221,42 @@ export abstract class BaseTracer extends BaseCallbackHandler {
...
+    runId: string,
+    parentRunId?: string
+  ): Promise<void> {
+    if (this.session === undefined) {
+      this.session = await this.loadDefaultSession();
+    }
+    const execution_order = this._getExecutionOrder(parentRunId);
+    const session = this.session as TracerSession;
+    const convertedMessages = messages.map((batch) =>
+      batch.map((message) => ({
```
vowelparrot, line 236: The python message format looks like
```
{
                        "_type": "human|ai|system",
                        "content": "the text",
                        "example": false, // only used for PALM currently?
                        "additional_kwargs": {} // not sure when this is ever used either
                    }
```

langchain/src/callbacks/manager.ts
```diff
@@ -309,6 +311,47 @@ export class CallbackManager
...
+    messages: BaseChatMessage[][],
+    runId: string = uuidv4()
+  ): Promise<CallbackManagerForLLMRun> {
+    let messageStrings: string[];
+    await Promise.all(
+      this.handlers.map(async (handler) => {
+        if (!handler.ignoreLLM) {
+          try {
+            if (handler.handleChatModelStart)
+              await handler.handleChatModelStart?.(
```
vowelparrot, line 325: I'm sure we could make this more concise with some ternary ops or better use of conditional chaining...

nfcampos (approved):

PR #1221, opened 2023-05-11
Separate base tracer from langchain tracers, Remove concept of session from base tracer, add missing exports, update console tracer accordingly

nfcampos:


langchain/src/callbacks/handlers/tracer_langchain.ts
```diff
@@ -0,0 +1,207 @@
...
+  ): Promise<TracerSession> {
+    const tenantId = this.tenantId ?? (await this.updateTenantId());
+    const configured_endpoint = `${endpoint}&tenant_id=${this.tenantId}`;
+    const response = await fetch(configured_endpoint, {
+      method: "GET",
+      headers: this.headers,
+    });
+    let tracerSession: TracerSession;
+    if (!response.ok) {
+      throw new Error(
```
nfcampos, line 145: I removed creating a fake session if it fails, as the fake session won't be accepted by the backend anyway, so there's no point

langchain/src/callbacks/handlers/tracer_langchain.ts
```diff
@@ -0,0 +1,207 @@
...
+      start_time: run.start_time,
+      end_time: run.end_time,
+      run_type: run.run_type,
+      reference_example_id: example_id,
+      extra: run.extra ?? {},
+      execution_order: run.execution_order,
+      serialized: run.serialized,
+      error: run.error,
+      inputs: run.inputs,
+      outputs: run.outputs ?? {},
```
nfcampos, line 181: This is a little strange as runs with error shouldn't have outputs

vowelparrot: It's supposed to be optional - I can double check that it works, may have added unnecessarily

langchain/src/callbacks/handlers/tracer_langchain.ts
vowelparrot: Should we add a `try / catch ` in `manager.ts` so the program doesn't crash when the env var is set?

nfcampos: We already have that

Issue #336, opened 2023-03-15
Bad Request 400 status code: "You must provide a model parameter" when attempting any openAI .call method


snapdeus:
Greetings,

**When using Node 18.1.0, or lower, on both Ubuntu 20.0.4 and Windows 10:**

Attempting to use the method `.call() ` results in a 400 status code. 
![image](https://user-images.githubusercontent.com/44621390/225312964-21f434c9-1981-4e99-8df5-1f322bc2d5a6.png)

Here is my code:

```
import { ChatOpenAI } from "langchain/chat_models"
import { HumanChatMessage } from 'langchain/schema'
import dotenv from 'dotenv'
dotenv.config()

const chat = new ChatOpenAI({
    openAIApiKey: process.env.OPENAI_API_KEY,
    temperature: 0.9

})
async function getChat() {
    const response = await chat.call([
        new HumanChatMessage(
            "Translate this sentence from English to French. I love programming."
        ),
    ]);


}

getChat()
```

Things I have tried: 

- Instead of passing the openAIApiKey as a parameter with .call(), I instead ran package.json script: `npm run dev` which was defined as `node -r dotenv/config index.js`
- Downgrading to Node 16.8.0 and setting NODE_ENV to experimental
- Downgrading the version of langchain to 0.0.22

### TEMPORARY SOLUTION

After changing my Node version to 18.2.0, the error does NOT happen! 

### What I suspect may be the cause:
Something with AxiosFetchAdapter...that's the only thing I can think of!

nfcampos:
Hi

In general it's not recommended to stick with old minor versions of Node. We only support running Langchain on the most recent versions of Node.js 16.x, 18.x, 19.x

jacobcwright:
Still getting this error with node v18.16.0

nfcampos:
Duplicate of #1010

Developergovind:
same error

PR #1222, opened 2023-05-11
Convert to Uint8Array to avoid Content-Type override

### LINKED ISSUE #1010: Broken OpenAI LLM implementation for node version 18
### Issue
When running the vanilla OpenAI LLM function, it returns a 400 for node version 18.

### To reproduce
The following code was used to create this error.

```ts

  const model = new OpenAI({ modelName: "gpt-3.5-turbo"});
  try {

    const resA = await model.call(
      "What would be a good company name a company that makes colorful socks?"
    );
    console.log({ resA });
    } catch (e: any) {
      console.log(e!.response!.data)
    }
};

run()
```

Error from the response body

```
{
  error: {
    message: 'you must provide a model parameter',
    type: 'invalid_request_error',
    param: null,
    code: null
  }
}
```

As seen, even though a model parameter is provided it still returns an error asking to provide one.

### Additional notes

This issue is not present in node v19 and v20.

dqbd:
Fixes #1010, #336

### DIFF
--- a/langchain/src/util/axios-fetch-adapter.js
+++ b/langchain/src/util/axios-fetch-adapter.js
@@ -332,6 +332,12 @@ function createRequest(config) {
       headers.delete("Content-Type");
     }
   }
+  // Some `fetch` implementations will override the Content-Type to text/plain
+  // when body is a string. 
+  // See https://github.com/hwchase17/langchainjs/issues/1010
+  if (typeof options.body === "string") {
+    options.body = new TextEncoder().encode(options.body);
+  }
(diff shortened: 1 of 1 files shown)


nfcampos:
Great investigation, thanks!

nfcampos:
Looks like might need to run prettier @dqbd

nfcampos (approved):

dqbd:
> Looks like might need to run prettier @dqbd

My bad, fixed now 👍

PR #1226, opened 2023-05-11
Update Chat Model Serialization

vowelparrot:
Unify format with the python lib, which was [defined of yore as](https://github.com/hwchase17/langchain/blob/4ee47926cafba0eb00851972783c1d66236f6f00/langchain/schema.py#L121)

```
def _message_to_dict(message: BaseMessage) -> dict:
    return {"type": message.type, "data": message.dict()}


def messages_to_dict(messages: List[BaseMessage]) -> List[dict]:
    return [_message_to_dict(m) for m in messages]


def _message_from_dict(message: dict) -> BaseMessage:
    _type = message["type"]
    if _type == "human":
        return HumanMessage(**message["data"])
    elif _type == "ai":
        return AIMessage(**message["data"])
    elif _type == "system":
        return SystemMessage(**message["data"])
    elif _type == "chat":
        return ChatMessage(**message["data"])
    else:
        raise ValueError(f"Got unexpected type: {_type}")


def messages_from_dict(messages: List[dict]) -> List[BaseMessage]:
    return [_message_from_dict(m) for m in messages]
```

### DIFF
--- a/langchain/src/stores/message/utils.ts
+++ b/langchain/src/stores/message/utils.ts
@@ -6,10 +6,17 @@ import {
   SystemChatMessage,
 } from "../../schema/index.js";
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type AdditionalKwargs = Record<string, any>;
+export interface StoredMessageData {
+  content: string;
+  role: string | undefined;
+  additional_kwargs?: AdditionalKwargs;
+}
+
 export interface StoredMessage {
   type: string;
-  role: string | undefined;
-  text: string;
+  data: StoredMessageData;
 }
 
 export function mapStoredMessagesToChatMessages(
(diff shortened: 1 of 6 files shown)


langchain/src/stores/message/utils.ts
```diff
@@ -6,10 +6,17 @@ import {
...
+  content: string;
+  role: string | undefined;
+  additional_kwargs?: AdditionalKwargs;
+}
+
 export interface StoredMessage {
   type: string;
-  role: string | undefined;
-  text: string;
+  data: StoredMessageData;
```
jacoblee93, line 19: I do feel like I'd rather shim this, mark it, then remove the shim when 1.0.0 comes out as a breaking change - I know all this is recent but have seen a few people using it already.

vowelparrot: @jacoblee93 thoughts on the mapV1MessageToStoredMessage ?

langchain/src/stores/message/utils.ts
```diff
@@ -6,29 +6,64 @@ import {
...
+    };
+  }
+}
+
 export function mapStoredMessagesToChatMessages(
   messages: StoredMessage[]
 ): BaseChatMessage[] {
   return messages.map((message) => {
-    switch (message.type) {
+    const stored_msg = mapV1MessageToStoredMessage(message);
```
jacoblee93, line 49: Camel case to match JS style

langchain/src/stores/message/utils.ts
```diff
@@ -6,29 +6,64 @@ import {
...
+interface StoredMessageV1 {
   type: string;
   role: string | undefined;
   text: string;
 }
 
+export function mapV1MessageToStoredMessage(
+  message: StoredMessage | StoredMessageV1
+): StoredMessage {
+  if ("data" in message) {
```
jacoblee93, line 31: Might be some weird edge case where message has the `data` property set but is `undefined` - the more idiomatic way to do this is (weirdly):
```
if (message.data !== undefined) {
}
```

langchain/src/stores/message/utils.ts
jacoblee93: I'm for it! Just would add a comment saying to remove when ready to make a breaking change so we don't forget about it

langchain/src/stores/message/utils.ts
vowelparrot: Cool thank you!

PR #1228, opened 2023-05-11
[test-exports-esbuild] Add missing entrypoints

jeffreylin:
Syncing entrypoints with `esm` modules since esbuild/esm should have the same imports. This should give us the expected CI test coverage in `esbuild`.

It seems like these were missed because in the time `esbuild` was added, `esm` already had two more commits that added entrypoints.

Compare to `esm`:
https://github.com/hwchase17/langchainjs/blob/main/test-exports-esm/src/entrypoints.js

Test plan:
 - CI tests

### DIFF
--- a/langchain/scripts/create-entrypoints.js
+++ b/langchain/scripts/create-entrypoints.js
@@ -203,6 +203,10 @@ const testExports = [
     (p) => `import * as ${p.replace(/\//g, "_")} from "langchain/${p}";`,
   ],
   [
+    "test-exports-esbuild",
+    (p) => `import * as ${p.replace(/\//g, "_")} from "langchain/${p}";`,
+  ],
+  [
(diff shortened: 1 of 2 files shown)


jacoblee93:
These should be auto-generated by script - CC @nfcampos?

jeffreylin:
@jacoblee93 - ah, I see - I think I found the upstream codegen script - it's `langchain/scripts/create-entrypoints.js`

Added a commit that adds `test-exports-esbuild` to that script. Now it autogenerates the same output as the previous commit when `yarn build` is run.
https://github.com/hwchase17/langchainjs/pull/1228/commits/11796a0ff147716e2f184d02adfe588b8589be0d

nfcampos:
Great, thanks!

PR #1232, opened 2023-05-12
Adds chat model support in the LC+ dataset runner

jacoblee93:
Also adds integration tests with a simpler data set that should run easier


jacoblee93:
CC @vowelparrot

PR #1233, opened 2023-05-12
add qdrant vector store Integration

ppxu:


langchain/src/vectorstores/qdrant.ts
```diff
@@ -0,0 +1,116 @@
...
+}
+
+export class QdrantVectorStore extends VectorStore {
+  client: QdrantClient;
+
+  collectionName: string;
+
+  constructor(embeddings: Embeddings, args: QdrantLibArgs) {
+    super(embeddings, args);
+    this.client = args.client;
```
jacoblee93, line 29: Instead of taking a client as an arg here, can we initialize the client internally from a config object like this?

https://github.com/hwchase17/langchainjs/pull/1271/files#diff-87a2d06951a95c0abda758677c0617d601b183627559a1cb331cf184e91fe56bR37

jacoblee93:
This looks solid overall - can you add a docs page with some examples like this one too?

https://js.langchain.com/docs/modules/indexes/vector_stores/integrations/milvus

langchain/src/vectorstores/qdrant.ts
ppxu: well, I support both taking a exist client as an arg and initializing a client internal, and will verify whether the collection exists.

langchain/src/vectorstores/qdrant.ts
```diff
@@ -0,0 +1,116 @@
+import { v4 as uuid } from "uuid";
+import type { QdrantClient } from "@qdrant/js-client-rest";
+
+import { Embeddings } from "../embeddings/base.js";
+import { VectorStore } from "./base.js";
+import { Document } from "../document.js";
+
+interface QdrantSearchResponse {
```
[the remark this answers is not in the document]
ppxu: Yes, I updated the type declaration.

jacoblee93:
Closing as the extension #1396 has just been merged. Thank you!

PR #1234, opened 2023-05-12
feat: Add filter support for Prisma

erictt:
Closes #1426

### DIFF
--- a/examples/src/indexes/vector_stores/prisma_vectorstore/prisma/migrations/00_init/migration.sql
+++ b/examples/src/indexes/vector_stores/prisma_vectorstore/prisma/migrations/00_init/migration.sql
@@ -3,6 +3,7 @@ CREATE EXTENSION IF NOT EXISTS vector;
 CREATE TABLE "Document" (
     "id" TEXT NOT NULL,
     "content" TEXT NOT NULL,
+    "namespace" TEXT NOT NULL DEFAULT "default",
     "vector" vector,
 
     CONSTRAINT "Document_pkey" PRIMARY KEY ("id")
--- a/examples/src/indexes/vector_stores/prisma_vectorstore/prisma/migrations/00_init/migration.sql
+++ b/examples/src/indexes/vector_stores/prisma_vectorstore/prisma/migrations/00_init/migration.sql
@@ -3,7 +3,7 @@ CREATE EXTENSION IF NOT EXISTS vector;
 CREATE TABLE "Document" (
     "id" TEXT NOT NULL,
     "content" TEXT NOT NULL,
-    "namespace" TEXT NOT NULL DEFAULT "default",
+    "namespace" TEXT DEFAULT 'default',
     "vector" vector,
 
     CONSTRAINT "Document_pkey" PRIMARY KEY ("id")
--- a/langchain/src/vectorstores/prisma.ts
+++ b/langchain/src/vectorstores/prisma.ts
@@ -78,7 +78,7 @@ export class PrismaVectorStore<
 
   selectSql: Sql;
 
-  filterSql?: Sql;
+  filter?: PrismaSqlFilter;
 
   idColumn: keyof TModel & string;
 
@@ -132,7 +132,7 @@ export class PrismaVectorStore<
     );
 
     if (config.filter) {
-      this.setSqlFilter(config.filter);
+      this.filter = config.filter;
     }
   }
 
@@ -265,10 +265,11 @@ export class PrismaVectorStore<
 
     const vectorQuery = `[${query.join(",")}]`;
 
-    // Override with the local filter if exists
-    const filterSql = filter
-      ? this.Prisma.raw(`WHERE ${this.buildSqlFilterStr(filter)}`)
-      : this.filterSql;
+    let whereCond = '';
+    if (filter) whereCond = this.buildSqlFilterStr(filter);
+    else if (this.filter) whereCond = this.buildSqlFilterStr(this.filter);
+
+    const filterSql = this.Prisma.raw(whereCond ? `WHERE ${whereCond}` : '');
 
     const articles = await this.db.$queryRaw<Array<SimilarityModel<TModel, TSelectModel>>>`
       SELECT ${this.selectSql}, ${this.vectorColumnSql} <=> ${vectorQuery}::vector as "_distance"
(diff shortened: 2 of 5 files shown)


langchain/src/vectorstores/prisma.ts
```diff
@@ -25,6 +25,10 @@ type PrismaNamespace = {
   raw: (sql: string) => Sql;
 };
 
+type PrismaSqlFilter = {
+  [key: string]: string|number
```
dqbd, line 29: _nit_: Don't forget to run Prettier / tests!

langchain/src/vectorstores/prisma.ts
```diff
@@ -276,6 +294,18 @@ export class PrismaVectorStore<
     return results;
   }
 
+  buildSqlFilterStr(filter: PrismaSqlFilter) {
+    return Object.keys(filter)
```
dqbd, line 298: _note_: Not sure whether that would be possible, but as Prisma reexports `sql-template-tag`, it might be more sane to defer the joining logic to that instead?

langchain/src/vectorstores/prisma.ts
```diff
@@ -276,6 +294,18 @@ export class PrismaVectorStore<
...
+      .map((key) => {
+        const value = filter[key];
+        if (typeof value === 'number') {
+          return `${key} = ${value}`;
```
dqbd, line 302: This does seem to be prone to SQL injection, especially when dealing with `value`. Will look into it more as well to see if we can handle it more safely.

dqbd (changes requested):

dqbd:
Hi! I've done some small adjustments and polish, addressing some of the pain points mentioned in the review above.

Notably:
- Using Prisma.join and Prisma.sql, we can send values separately without string interpolation, reducing the risk of SQL injection.
- Filters have been extended to support multiple operators (=, <>, <, <=, >=, >), modelling the same filter structure similar to how `where` works in Prisma
- Migration and schema has been fixed

dqbd (approved):

examples/src/indexes/vector_stores/prisma_vectorstore/prisma/migrations/00_init/migration.sql
```diff
@@ -3,6 +3,7 @@ CREATE EXTENSION IF NOT EXISTS vector;
 CREATE TABLE "Document" (
     "id" TEXT NOT NULL,
     "content" TEXT NOT NULL,
+    "namespace" TEXT DEFAULT 'default',
```
jacoblee93, line 6: Just making sure, this won't break legacy users right? It's completely optional?

langchain/src/vectorstores/prisma.ts
```diff
@@ -245,19 +286,38 @@ export class PrismaVectorStore<
...
-      SELECT ${this.selectSql}, ${this.vectorColumnSql} <=> ${vectorQuery}::vector as "_distance" 
-      FROM ${this.tableSql}
-      ORDER BY "_distance" ASC
-      LIMIT ${k};
-    `;
+    >(
+      this.Prisma.join(
+        [
+          this.Prisma.sql`
+            SELECT ${this.selectSql}, ${this.vectorColumnSql} <=> ${vectorQuery}::vector as "_distance"
```
jacoblee93, line 309: This is ok because everything here is generated in from a config file right or a vector array right? 

I still think it would be nice to parameterize things for best practices.

jacoblee93: Or does `this.Prisma.sql` sanitize?

jacoblee93 (commented):
A few questions but looks good

langchain/src/vectorstores/prisma.ts
erictt: Base on the [doc](https://www.prisma.io/docs/concepts/components/prisma-client/raw-database-access#sql-injection):
> $executeRaw can only run one query at a time. You cannot append a second query - for example, adding DROP bobby_tables to the end of an ALTER.

Although it didn't mention queryRaw, but i found a [related issue](https://github.com/prisma/docs/issues/524) says:

> queryRaw(sql) and executeRaw(sql) will only be executing a single statement, which means some injections won't work (the one with multiple queries trying to drop tables for example) but some will (unveiling additional data using UNION for example) 

So I don't think this will be a problem.

dqbd: Yep! Parameters are by default sanitised unless they are wrapped in a `Prisma.raw` call. 

There are some dynamic fields, which cannot be parametrised (such as table names and column names). Moved these `Prisma.raw` calls closer to the `executeRaw` and `queryRaw` invocations alongside a note in docs to warn users.

langchain/src/vectorstores/prisma.ts
```diff
@@ -220,14 +257,18 @@ export class PrismaVectorStore<
...
 
     await this.db.$transaction(
       vectors.map(
         (vector, idx) => this.db.$executeRaw`
-          UPDATE ${this.tableSql}
-          SET ${this.vectorColumnSql} = ${`[${vector.join(",")}]`}::vector
-          WHERE ${idSql} = ${documents[idx].metadata[this.idColumn]}
+          UPDATE ${tableNameRaw}
+          SET ${vectorColumnRaw} = ${`[${vector.join(",")}]`}::vector
+          WHERE ${idColumnRaw} = ${documents[idx].metadata[this.idColumn]}
```
jacoblee93, line 271: What if `documents[idx].metadata[this.idColumn]` can't be trusted? E.g. if you're loading documents in from a different source?

jacoblee93: Oh, right didn't see the other comment about executeRaw only being able to execute one query. I still don't love this and it should be escaped.

langchain/src/vectorstores/prisma.ts
jacoblee93: Ok. Using a single statement helps but I'm sure there's still sneaky issues and damage some bad metadata on a loaded document could do even in a single query, so please escape here: 

https://github.com/hwchase17/langchainjs/pull/1234/files#diff-9bc13719b32d4ab3d65a0d4f561ae4427f1c746ad7d140e30c0005b4e0d91018R271

Raw SQL from user provided fields is also really not ideal but at least it'll be the user's own fault if something happens there.

langchain/src/vectorstores/prisma.ts
dqbd: afaict `documents[idx].metadata[this.idColumn]` and `[${vector.join(",")}]` is still escaped and parametrised.

jacoblee93:
Thanks for the patience here, will merge with the next release!

Issue #809, opened 2023-04-15
Investigate bundle size


nfcampos:
Is it tiktoken, or something else?

dqbd:
Hello! 

Looking at the bundle size and metafile when bundled through esbuild + wrangler, the main reason the bundle size is large is due to the big WASM file included by @dqbd/tiktoken. 

The WASM file contains the rankings for all models (~3 MB uncompressed) + the tokenizer logic itself (~700 kB uncompressed). 

Even if ranks only for `gpt-3.5`/`gpt-4` are bundled (skipping older gpt-2 ranks etc.), the `cl100k_base` (1.1 MB uncompressed) + tokenizer fits just right into the 1 MB limit when compressed, not leaving enough space for rest of `langchain`.

I think the best way forward is to either:
- Split the `@dqbd/tiktoken` package into a separate endpoint for the edge. That way, we're not bundling anything. However, as this would be an additional network call, this would be at the expense of slower tokenization. Some preliminary work has been done in that regard, we could release an isomorphic package, which would call an API if imported in edge runtimes.
- Avoid tokenization altogether when just consuming `OpenAI` and other LLMs. This would be harder to do as far as I can see and would most likely be a breaking change. 

---

Adding the analyzed `meta.json` bundle when just importing `OpenAi` LLM. 

<img width="1091" alt="CleanShot 2023-04-15 at 15 17 04@2x" src="https://user-images.githubusercontent.com/1443449/232226197-b1a528da-4724-44d4-8d11-c8daaa655ac7.png">

Note: this is not only related to CF Workers but for all other edge runtimes as well.

nfcampos:
Thanks a lot for this investigation @dqbd 
 
Approach 1 sounds interesting to me, anything I can do to help here?

Approach 2 is harder as the tokeniser is eg used deep inside some chains, so not terribly clear how to make it more optional. 

That said, we could conceivably move the countTokens methods outside of the LLM classes so that they need to be imported separately, I'd be okay with that, I don't think many people are using them directly. 

What's harder is 
1. The chains that use it, those would need to be moved to separate endpoints from other chains
2. The maxTokens=-1 behaviour on OpenAI LLM which uses the tokeniser to calculate max tokens, that'd be harder to remove

but I would however be perfectly okay with making it require a remote download when used as you suggested 

thoughts?

protoEvangelion:
Thanks @dqbd I'm seeing something similar when running cloudflare worker cli:

```
[...]

Issue #62, opened 2023-02-19
Add support for Cloudflare Workers / Vercel Edge Functions


dqbd:
So far it seems like there are 3 culprits which may block us from running `Langchainjs` on Edge.

- [x] 1) OpenAI and it's usage of `axios` #118 

    - Switch to a different library / create a upstream PR for the official library (the pending MR seems to be doing way too much IMO)

- [x] 2) `**/index.ts` files might include other modules, which aren't needed 

     - better treeshaking support, which

- [x] 3) `langchain/util/hub` and the general usage of NodeJS APIs (`fs`, etc.) #97 

- [x] 4) Have at least 1 vector store which 1. doesn't need Node native modules and 2. stores metadata for each document #73 

- [ ] Add tag in the docs for node-only things (eg hnswlib)

- [x] Create and test cloudlfare workers and vercel edge templates

- [x] Remove usage of node-fetch (which uses node-only APIs) #118 

- [x] Remove all usage of require #124  

- [x] Get TS to output ESM #124 

- [x] Remove usage of node `crypto` #213 

- [x] Replace node-only tokeniser with one that works in browser-like envs #214 

This is the error output when attempting to deploy the `examples/src/chains/llm_chain.ts` via Vercel:

<img width="582" alt="CleanShot 2023-02-19 at 19 28 23@2x" src="https://user-images.githubusercontent.com/1443449/219971669-827be1f6-fd2c-4b3c-995c-b5d572a0847d.png">

jasongill:
I'm also investigating this, seeing if I can clean up or pare down the package a bit to make it work in Cloudflare Workers. In theory `node_compat = true` in wrangler.toml should fix many of those missing modules by adding polyfills, but will also need to replace node's crypto with WebCrypto I think.

nfcampos:
Another thing to fix here is to add support for at least one docstore that is external, as there is no filesystem support to load one from a file

nfcampos:
One option for openai adapter is mentioned here https://github.com/openai/openai-node/issues/30#issuecomment-1342350056

jasongill:
@nfcampos @dqbd I would be happy to pay to encourage support for Cloudflare Workers for Langchain, if there's a place/dev I can send $500 to contribute to getting this done please let me know! Langchain at the edge is very exciting.

hwchase17:
@jasongill trying to think/understand the use case here. this may sound like a really stupid question, but what exactly is it about langchain at the edge that excites you so much?

[...]

PR #1239, opened 2023-05-12
Use a JS based tokenizer for token counting

dqbd:
The idea behind this PR is to reduce the friction when adopting LangchainJS by rewriting the tokenizer library into JS and fetching ranks on demand from a CDN (`https://tiktoken.pages.dev`). As a side effect, this will reduce the bundle size (necessary for edge functions, see #809, #62), at the expense of slower tokenization (which should be a non-issue for pure token counting). 

Tokenization behavior should match the @dqbd/tiktoken, which has been entirely removed. 

Supersedes #847 and #1146

- [x] Move the tests to ensure correct tokenization 
- [x] Remove WASM tiktoken entirely in count_tokens (weak dependencies are not well supported afaict)
- [x] Serve other encoders from CDN (w/ cache)


langchain/src/util/tiktoken.ts
```diff
@@ -0,0 +1,29 @@
...
+} from "js-tiktoken/lite";
+
+const cache: Record<string, TiktokenBPE> = {};
+
+export async function getEncoding(
+  encoding: TiktokenEncoding,
+  extendedSpecialTokens?: Record<string, number>
+) {
+  if (!(encoding in cache)) {
+    const res = await fetch(`https://tiktoken.pages.dev/js/${encoding}.json`);
```
nfcampos, line 16: Let's use AsyncCaller class here for retry behaviour

langchain/src/util/tiktoken.ts
```diff
@@ -0,0 +1,29 @@
...
+  Tiktoken,
+  TiktokenBPE,
+  TiktokenEncoding,
+  TiktokenModel,
+  getEncodingNameForModel,
+} from "js-tiktoken/lite";
+
+const cache: Record<string, TiktokenBPE> = {};
+
+export async function getEncoding(
```
nfcampos, line 11: Is it possible this will ever be called multiple times concurrently? If so (and not yet cached) we'd end up with multiple inflight requests for the same thing. If we think that's likely can avoid by also caching the promise synchronously

nfcampos (approved):
Looks great!

nfcampos:
Let's also update the test-exports packages to remove any custom config we had put in there for wasm etc

test-exports-vercel/next.config.js
```diff
@@ -1,14 +1,6 @@
 /** @type {import('next').NextConfig} */
```
nfcampos, line 1: We can probably remove this file completely?

anzemur:
@nfcampos @dqbd I think this is not quite right - the unpackaged size of `js-tiktoken` is almost 10MB and this package is listed as the dependency. The usage of the library in `langchainjs` is implemented in a way that the encoders are always loaded from the CDN the first time they are used, so the actual encoder files are never used - this means that we have 10MB of overhead here. The "light" version of `js-tiktoken` should only include the code required for running the tokenizers, and the user should provide the actual encoder - from local files or from CDN.

 The size of `langchainjs` itself is almost 5MB and with all of the dependencies, this can cause troubles in environments where package size is limited (ex. AWS Lambda).

dqbd:
Hey @anzemur! As mentioned in https://github.com/dqbd/tiktoken/issues/68 but applicable for LangchainJS as well, the best course of action is to add an intermediate build step for tree shaking and bundling dependencies, such as `esbuild` or `rollup` (as documented [here](https://docs.aws.amazon.com/lambda/latest/dg/nodejs-package.html#nodejs-package-dependencies:~:text=If%20you%20bundle,I/O%20operations.)). That should address your concerns with the bundle size as well as increase your performance of Lambda functions

PR #1250, opened 2023-05-12
Update to Use a Factory Method

vowelparrot:
To enable independent trials over datasets

### DIFF
--- a/langchain/src/client/langchainplus.ts
+++ b/langchain/src/client/langchainplus.ts
@@ -108,14 +108,18 @@ const stringifyError = (err: Error | unknown): string => {
   return result;
 };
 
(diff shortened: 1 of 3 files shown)


langchain/src/client/langchainplus.ts
```diff
@@ -110,28 +110,57 @@ const stringifyError = (err: Error | unknown): string => {
...
+  const bchFactory = llm as () => Promise<BaseChain>;
+  const bch = await bchFactory();
   return (
     typeof bch?._chainType === "function" && bch?._chainType() !== undefined
   );
 }
 
+type _ModelType = "llm" | "chatModel" | "chainFactory";
+
+async function getModelOrFactoryType(
```
vowelparrot, line 147: Extracted this to avoid instantiating multiple times just to check type

PR #1252, opened 2023-05-13
feat: store information about the sql query

L-Qun:
We cannot get the information about the sql directly, so I have added the intermediateStep field to mark the sql statement and the results of the sql query


nfcampos:
Hi, thanks for this @L-Qun we've added support for this in #1313

PR #1253, opened 2023-05-13
Nc/fix tracer chat output

nfcampos:


langchain/src/callbacks/handlers/tracer.ts
```diff
@@ -132,17 +131,14 @@ export abstract class BaseTracer extends BaseCallbackHandler {
...
-    );
     const run: Run = {
       id: runId,
       name: llm.name,
       parent_run_id: parentRunId,
       start_time: Date.now(),
       end_time: 0,
       serialized: llm,
-      inputs: { messages: convertedMessages },
+      inputs: { messages },
```
vowelparrot, line 141: So we are storing directly as messages here?

vowelparrot (approved):

PR #1259, opened 2023-05-13
fix: maintain human and ai prefix in buffer window memory

aseem2625:
Same as in python repository

https://github.com/hwchase17/langchain/blob/master/langchain/memory/buffer_window.py#L33-L37


nfcampos:
Thanks!

PR #795, opened 2023-04-14
Add GoogleSearchAPIWrapper

sajadghawami:
Tried to stay very close to the [implementation](https://python.langchain.com/en/latest/modules/agents/tools/examples/google_search.html) of the python package.

The following env variables are needed:

`GOOGLE_CSE_ID`
`GOOGLE_API_KEY`

PR #1260, opened 2023-05-13
Added google custom search as a tool

YIZHUANG:
Didn't choose serpapi or serper cause it requires you to pay up front and serpapi is subscription based rather than usage base. Otherwise works quite ok, see screenshots:

Screenshot 2023-05-14 at 1 31 56
Screenshot 2023-05-14 at 1 30 48


jacoblee93:
Thanks for this!

jacoblee93:
This will close #795 as well.

PR #1262, opened 2023-05-14
feat: add datetime tool

rpidanny:
This tool empowers agents with the ability to determine the present date and time.
Its usefulness lies in liberating models from the constraints of being stuck in a snapshot of the past, where they were originally trained.


jacoblee93:
Thank you for the contribution - IMO this is a bit too trivial to do with a `DynamicTool` and specific with its return format.

PR #1264, opened 2023-05-14
feat: add token buffer based memory, port of ConversationTokenBufferMemory in langchain python

aseem2625:
JS port of python version [here](https://github.com/hwchase17/langchain/blob/master/langchain/memory/token_buffer.py)
`k window` based memory doesn't work in scenarios where the answer length cannot be predicted, so openai doesn't return any response in such cases or the answer doesn't have natural STOP. So, this `token` based buffer memory allows to insert chat history safely.

- [x] Local testing
- [ ] Test cases - TBD after [this warning](https://github.com/hwchase17/langchainjs/issues/985#issuecomment-1546898014) is fixed. Not a blocker as it fallbacks to approx tiktoken calculation but prior to this PR, langchainJS doesn't have any token based library, so this warning has never surfaced. Since it is polluting the global logs, this must be fixed in original repo.

image

### DIFF
--- a/langchain/src/memory/token_buffer.ts
+++ b/langchain/src/memory/token_buffer.ts
@@ -1,0 +1,136 @@
+import { BaseLanguageModel } from "../base_language/index.js";
+import {
+    getBufferString,
+    InputValues,
+    MemoryVariables,
+    OutputValues,
+} from "./base.js";
+import { BaseChatMemory, BaseChatMemoryInput } from "./chat_memory.js";
+
+interface BufferMemoryInput extends BaseChatMemoryInput {
+    humanPrefix?: string;
+    aiPrefix?: string;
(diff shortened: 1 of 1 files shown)


jacoblee93:
Hey @aseem2625 is this ready for a look?

dmanresa-saes:
This would be very good. If this branch is not merged. Is there any workaround to do the same thing that ConversationTokenBufferMemory does with what is implemented in the JS version of Langchain?

dmanresa-saes:
This would be very good. If this branch is not merged. Is there any workaround to do the same thing that ConversationTokenBufferMemory does with what is implemented in the JS version of Langchain?

jacoblee93:
This was later merged in #3211

PR #1267, opened 2023-05-15
feat: Add sagemaker endpoint support

cmanou:
- Kept it as close to the python implementation as possible
- I have not implemented the `credentials_profile_name` arg, given the changes with the new aws-sdk v3 I am wondering if we should just allow passing in the full credential arg so people can use whatever aws [credential provider](https://github.com/aws/aws-sdk-js-v3/tree/main/packages/credential-providers) they want.
- I don't know how the integration test should work given it would need a sagemaker endpoint, this looks like it was also an issue for the python implementation as it is also missing the tests.


jacoblee93:
This looks really cool!

Could you add an integration test to demonstrate how to set up and use this?

cmanou:
@jacoblee93 have added the docs and integration test albeit skipped due to the need for a sagemaker endpoint that's accessible for ci to test.

jacoblee93:
> @jacoblee93 have added the docs and integration test albeit skipped due to the need for a sagemaker endpoint that's accessible for ci to test.

Thanks! Made a few modifications:
1. You don't need to skip the integration tests as they don't run as part of CI right now.
2. We don't need to do the `await import` syntax anymore - just import it, make sure it's under a separate entrypoint, and add it under `requiresOptionalDependency` in `langchain/scripts/create-entrypoints.js`: https://github.com/hwchase17/langchainjs/pull/1267/files#diff-d541783a759c4660823be9d9adc2406939edbc517f876c28b18136abd4db6f86R160

Also moved a few args and types around - worked great overall with HuggingFace GPT-2 for me. Thank you!

cmanou:
@jacoblee93 thanks for making those changes. Is there anything else needed to get this over the line ?

PR #1269, opened 2023-05-15
Fixed typo with example

Njima1572:
I was checking out the `langchainjs` and found out that there was a typo in the example.


jacoblee93:
Thanks for the catch!

PR #1270, opened 2023-05-15
Add Environment Info to Run Logs

vowelparrot:
I'm very confident I'm doing this sub-optimally so could use some feedback


nfcampos:
@vowelparrot I've pushed a commit removing some of this info so that this us guaranteed to work across all envs, we can add more keys over time

vowelparrot:
I like it - thank you!

Issue #1240, opened 2023-05-12
Google Vertex AI PaLM Support


afirstenberg:
Reference: hwchase17/langchain#4532

Although there is an open issue for the PaLM API integration (#1129), these APIs are different and should be tracked separately.

I would be willing to work on this one as well and coordinate with @Jflick58 on the Python side.

danny-avila:
Would love to see this integrated. I’m curious how it will perform given the examples parameter which explicitly defines expected outputs to the API. I still need to test this bit myself.

I just integrated PaLM 2 into my project here: https://github.com/danny-avila/chatgpt-clone

My chief complaint is that the moderation filter is pretty high. It will false flag certain topics consistently. Also despite its ability to translate well, it will refuse to answer messages without any English.

Jflick58:
@afirstenberg  let's do it. Want to connect on the langchain discord?

sounishnath003:
I am also eager to work on this integration. I would love to hear more on this as soon as possible.

It certainly will be a nice feature to introduce to langchain ecosystem.

afirstenberg:
For those interested, the first integration (text) was just merged into main with #1273. Expect Chat and Embedding support soon.

danny-avila:
> For those interested, the first integration (text) was just merged into main with #1273. Expect Chat and Embedding support soon.

Awesome work!!!

afirstenberg:
These are now available as of versions 0.0.80 and 0.0.82.

sounishnath003:
Thank you so much @afirstenberg for the Google Palm support for langchainjs.

PR #1273, opened 2023-05-15
Feature: Google Vertex AI PaLM Support

afirstenberg:
This adds basic support for both text and chat based models through the Vertex AI API.

Addresses #1240


langchain/src/llms/googlevertexai.ts
```diff
@@ -0,0 +1,387 @@
...
+ * methods in place:
+ * - You are logged into an account permitted to the Google Cloud project
+ *   using Vertex AI.
+ * - You are running this on a machine using a service account permitted to
+ *   the Google Cloud project using Vertex AI.
+ * - The `GOOGLE_APPLICATION_CREDENTIALS` environment variable is set to the
+ *   path of a credentials file for a service account permitted to the
+ *   Google Cloud project using Vertex AI.
+ */
+export class GoogleVertexAiChat
```
jacoblee93, line 310: Can you move this under `chat_models`? E.g.:

langchain/src/chat_models/anthropic.ts

langchain/src/llms/googlevertexai.ts
```diff
@@ -0,0 +1,387 @@
...
+interface GoogleVertexAiLLMTextInstance {
+  content: string;
+}
+
+interface GoogleVertexAiChatExample {
+  input: GoogleVertexAiChatMessage;
+  output: GoogleVertexAiChatMessage;
+}
+
+interface GoogleVertexAiChatMessage {
```
jacoblee93, line 110: Can you try to make this match: https://github.com/hwchase17/langchainjs/blob/main/langchain/src/schema/index.ts#L65

langchain/src/llms/googlevertexai.ts
```diff
@@ -0,0 +1,387 @@
...
+  data: {
+    predictions: Prediction[];
+  };
+}
+
+interface GoogleVertexAiLLMTextInstance {
+  content: string;
+}
+
+interface GoogleVertexAiChatExample {
```
jacoblee93, line 105: Is this exported from some Google TypeScript package somewhere?

jacoblee93 (changes requested):

langchain/src/llms/googlevertexai.ts
afirstenberg: Sure, tho I had it here since the OpenAI chat is also in LLM

afirstenberg: I guess, perhaps, I'm confused about the openai-chat.ts code and what it's there for, since I do see it under chat_models/openai.ts. Will move it there.

langchain/src/llms/googlevertexai.ts
afirstenberg: No, the chat model allows for an array of examples, which I'm just modeling here.
(There is no Google TypeScript package for this at all. The dependency is just for auth.)

afirstenberg:
Code should reflect all of the above changes.
I still have to go in and review examples and documentation.

langchain/src/llms/googlevertexai.ts
jacoblee93: The idea was to have an entrypoint for newer people who weren't aware `gpt-3.5-turbo` and `gpt-4` are a separate class of model:

https://github.com/hwchase17/langchainjs/blame/main/langchain/src/llms/openai.ts#L112

Definitely a bit confusing and something we should clarify better going forwards.

afirstenberg:
Tests, examples, and documentation added.
Refactored types and a connection class to separate directories.
I think this is ready for further review.

jacoblee93:
Would you mind/would you mind if I split this up into two PRs? I think the LLM part and the general auth pieces looks pretty much good (can we call it just `GoogleVertexAI` rather than `GoogleVertexAiTextLLM` to be more standard though? And also capitalize `AI` in the general helper classes?) while the chat model part has a few non-standard things (like `context` where we've typically used `SystemChatMessage`) that I'd love to try to standardize.

I'm happy to do the work to split it if you'd like - just don't want to push such a big change directly to the open branch without your ok first!

afirstenberg:
I have no objection to the name changes for consistency.
I have no strong objections to splitting out the Chat changes if you feel they need further baking, however it *does* respect the standard of SystemChatMessage to set the context. I added a context field to both the constructor and call options for consistency with Google's model, however providing them as part of messages overrides either of those. See the unit tests (just committed), and I'll provide clarification and usage in the documentation about context and examples.

jacoblee93:
> I have no objection to the name changes for consistency. I have no strong objections to splitting out the Chat changes if you feel they need further baking, however it _does_ respect the standard of SystemChatMessage to set the context. I added a context field to both the constructor and call options for consistency with Google's model, however providing them as part of messages overrides either of those. See the unit tests (just committed), and I'll provide clarification and usage in the documentation about context and examples.

I still think I'd to split it into two just to keep individual PRs compact and focused. I'll start on it now!

jacoblee93 (approved):

afirstenberg:
Looking forward to release!
(And fine to publish my twitter handle: @afirstenberg, same as github.)
Thanks for your help with this.

PR #1274, opened 2023-05-15
Add PromptLayer chat model support, return promptLayerRequestID in generationInfo

SalehHindi:
We want to return promptLayerRequestID if PromptLayerOpenAI and PromptLayerOpenAIChat if and only if returnPromptLayerID = true when initializing the model. Added tests as well.

Here is some sample output:

```typescript
  const model = new PromptLayerOpenAIChat({
    prefixMessages: [
      {
        role: "system",
        content: "You are a helpful assistant that answers in pirate language",
      },
    ],
    returnPromptLayerID: true,
  });
  const res = await model.generate([
    "What would be a good company name a company that makes colorful socks?"
  ]);
  console.log(JSON.stringify(res));
  
 // result of res
 {
   "generations": [
      [
         {
            "text": "Arrr, ye could name yer company \"Rainbow Boot Socks\" or \"Jolly Jester Socks\" or \"Parrot Feathers Footwear\" or \"Sunny Seadog Socks\". Choose any name that represents the vibrant and flamboyant nature of yer socks, matey!",
            "generationInfo": {
               "promptLayerRequestID": 1998796
            }
         }
      ]
   ]
}

```


SalehHindi:
Hey @Jped just finished this PR. Do you mind taking a look? 

@hwchase17 @dev2049

langchain/src/util/prompt-layer.ts
```diff
@@ -0,0 +1,42 @@
+import fetch from 'cross-fetch';
+import { AsyncCaller } from "../util/async_caller.js";
+
+export const getPromptLayerRequestID = async (
```
Jped, line 4: @SalehHindi is there a reason why this function is called `getPromptLayerRequestID`? Why is the name focusing on getting the request ID when the main functionality of this function is to make a promptlayer track-request.

SalehHindi: I tried making the function narrow in scope but I can change it so that it's something like `makePromptLayerTrackRequest` or `promptLayerTrackRequest`. I think now that you say that, that makes more sense.

SalehHindi:
Sorry, i had asked for a review earlier, but let me push up more code to add `PromptLayerChatOpenAI`

natarius:
Thank you so much!!! Can't wait for this to ship!

langchain/src/util/prompt-layer.ts
jacoblee93: Agreed on this, can you rename?

It's also more conventional on our end to do `getPromptLayerRequestId` with the "D" in `id` lower-cased.

langchain/src/llms/tests/openai.int.test.ts
```diff
@@ -180,3 +180,27 @@ test("Test OpenAI prompt value", async () => {
   }
   console.log({ res });
 });
+
+test("Test PromptLayerOpenAI returns promptLayerID iff returnPromptLayerID=true", async () => {
```
jacoblee93, line 184: Fix typo. Also this should be in its own integration test, not part of the existing one.

langchain/src/llms/tests/openai-chat.int.test.ts
```diff
@@ -101,3 +101,34 @@ test("Test OpenAI with signal in call options and node adapter", async () => {
     return ret;
   }).rejects.toThrow();
 }, 5000);
+
+test("Test PromptLayerOpenAI returns promptLayerID iff returnPromptLayerID=true", async () => {
```
jacoblee93, line 105: Fix typo. Also this should be in its own integration test, not part of the existing one.

docs/docs/modules/models/llms/integrations.mdx
```diff
@@ -118,6 +118,73 @@ const res = await model.call(
...
+//         "completionTokens": 5,
+//         "promptTokens": 23,
+//         "totalTokens": 28
+//      }
+//   }
+//}
+```
+
+### PromptLayerOpenAIChat
+PromptLayer is compatible with the OpenAIChat model as well. Here is an example of getting the PromptLayerOpenAIChat requestID:
```
jacoblee93, line 160: Remove, `OpenAIChat` is mostly deprecated in favor of the `ChatOpenAI` model (which you've added support to!).

jacoblee93: Add docs to `chat_models/integrations.mdx` instead. Also, we have been shifting to use imported examples as show here:

https://github.com/hwchase17/langchainjs/blob/main/.github/contributing/INTEGRATIONS.md#documentation-and-integration-tests

langchain/src/llms/openai.ts
```diff
@@ -373,6 +376,24 @@ export class OpenAI extends BaseLLM implements OpenAIInput, AzureOpenAIInput {
             }
           );
 
+      const requestEndTime = Date.now();
```
jacoblee93, line 379: PromptLayer specific code should not be in the main OpenAI LLM code path - can you move it into the `PromptLayerOpenAI` class?

jacoblee93 (changes requested):
Thanks for this! A few questions. Also don't forget to format and lint.

langchain/src/util/prompt-layer.ts
SalehHindi: Agreed on the name change. Changed the function name to `promptLayerTrackRequest` which is typically called like `let promptLayerRespBody = await promptLayerTrackRequest(...)`. Lmk if there is something better you suggest. 

Also thx for the tip about naming id like `varNameId` instead of `varNameID`.

SalehHindi:
@jacoblee93 Thanks for the feedback, addressed all your comments. Do you mind taking a final look? Much appreciated!

jacoblee93 (approved):
Looks good! Removed an excess import and fixed up things for the linter. Give it a last look if you'd like but will merge later tonight or tomorrow morning and will be part of the next release!

langchain/src/util/prompt-layer.ts
```diff
@@ -0,0 +1,42 @@
+import fetch from "cross-fetch";
```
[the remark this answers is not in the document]
SalehHindi: Oh thank you for removing for me! VSCode was complaining for me but didn't realize it somehow was a global.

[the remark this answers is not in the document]
jacoblee93: One big reason we upgraded to Node 18 :)

PR #1275, opened 2023-05-16
Add metadata support in VectorStoreRetrieverMemory

baobo5625:
Add metadata support while creating more document, use same metadata for filter as well.


baobo5625:
@nfcampos could you take a look at this one?

PR #1282, opened 2023-05-16
Add metadata support in memory vector

baobo5625:
Add metadata support while creating more document, use same metadata for filter as well.

Add history prompt indicator if the variable name contains history

### DIFF
--- a/langchain/src/memory/vector_store.ts
+++ b/langchain/src/memory/vector_store.ts
@@ -5,7 +5,7 @@ import {
   getInputValue,
   InputValues,
   MemoryVariables,
-  OutputValues,
+  OutputValues
 } from "./base.js";
 
 export interface VectorStoreRetrieverMemoryParams {
@@ -14,6 +14,8 @@ export interface VectorStoreRetrieverMemoryParams {
   outputKey?: string;
   memoryKey?: string;
   returnDocs?: boolean;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  metadata?: Record<string, any>;
 }
 
 export class VectorStoreRetrieverMemory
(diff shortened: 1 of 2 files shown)


langchain/src/prompts/chat.ts
```diff
@@ -62,7 +62,14 @@ export class MessagesPlaceholder extends BaseMessagePromptTemplate {
   }
 
   formatMessages(values: InputValues): Promise<BaseChatMessage[]> {
-    return Promise.resolve(values[this.variableName] as BaseChatMessage[]);
+    return Promise.resolve([
+      values[this.variableName].length > 0 &&
+      this.variableName.includes("history")
+        ? `Relevant pieces of previous conversation:: ${
```
dqbd, line 68: Not sure about hardcoding the prompt at this location. We need to also make sure `BaseChatMessage` is returned instead of a `string`. 

If I understand correctly, this is desired for LLMs which do not support the ChatML syntax? Shouldn't that be the responsibility of the LLM?

dqbd, line 68: _note_: Consider switching from `any` to `unknown` for `InputValues` in `langchain/src/schema/index.ts` to catch the invalid signature issues.

langchain/src/prompts/chat.ts
```diff
@@ -27,7 +27,7 @@ export abstract class BaseMessagePromptTemplate {
   serialize(): SerializedMessagePromptTemplate {
     return {
       _type: this.constructor.name,
-      ...JSON.parse(JSON.stringify(this)),
+      ...JSON.parse(JSON.stringify(this))
```
dqbd, line 30: If possible, please invoke `yarn format` / Prettier to properly format the code.

langchain/src/memory/vector_store.ts
```diff
@@ -14,6 +14,8 @@ export interface VectorStoreRetrieverMemoryParams {
   outputKey?: string;
   memoryKey?: string;
   returnDocs?: boolean;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  metadata?: Record<string, any>;
```
dqbd, line 18: _nit_: consider using `unknown` instead to avoid unintended disabling of type-checking

dqbd (commented):
Thanks for the PR @baobo5625, some notes and questions:

dqbd (commented):
Thanks for the PR @baobo5625, some notes and questions

langchain/src/prompts/chat.ts
baobo5625: Well, the bug I encountered was that, while I was using the conversation Agent with memory, the chat_history will be converted to a pure string message, which doesn't not have the type nor text attributes.

It turns out it was due to the placeholder format result.

Due to the limit understanding of the whole project, I tested with this hard coded prompt to verify the bug. Ideally we should create a message type for chat history.

langchain/src/memory/vector_store.ts
baobo5625: good call, i'll update this part

PR #1287, opened 2023-05-16
Add Invocation Params to onLLMStart

vowelparrot:


langchain/src/chat_models/base.ts
```diff
@@ -72,9 +73,13 @@ export abstract class BaseChatModel extends BaseLanguageModel {
       this.callbacks,
       { verbose: this.verbose }
     );
+    const invocationParams = { invocation_params: this?.invocationParams() };
     const runManager = await callbackManager_?.handleChatModelStart(
       { name: this._llmType() },
-      messages
+      messages,
+      uuidv4(),
```
nfcampos, line 80: This should be undefined

langchain/src/llms/base.ts
```diff
@@ -88,10 +96,15 @@ export abstract class BaseLLM extends BaseLanguageModel {
       this.callbacks,
       { verbose: this.verbose }
     );
+    const invocationParams = { invocation_params: this?.invocationParams() };
     const runManager = await callbackManager_?.handleLLMStart(
       { name: this._llmType() },
-      prompts
+      prompts,
+      uuidv4(),
```
nfcampos, line 103: This should be undefined

nfcampos (commented):
Looks good, two comments

langchain/src/llms/base.ts
vowelparrot: Makes sense! TY! I'm assuming that means explicitly passing `undefined`

PR #1288, opened 2023-05-16
Support input schemas for prompts and LLMChains, allow multi-input keys if only one is required

jacoblee93:
Given a prompt created with an `inputSchema`, this PR should also allow TS autocompletion of LLMChain call input parameters.


jacoblee93:
Should be backwards compatible but I'm actually almost certain there's a way to make the autocomplete apply to the existing style of prompts/chains as well - will think more about it tomorrow.

marcusglowe:
@jacoblee93 any interest in picking this back up? 

a while ago (may be incompatible with more recent versions of langchain) I prototyped adding some type safety to inputs
```ts
import { LLMChain, OpenAI, PromptTemplate } from "langchain";
import { parseTemplate } from "langchain/prompts";

type Parts<Path> = Path extends `${infer _ignore}{${infer PartA}}${infer PartB}`
  ? PartA | Parts<PartB>
  : never;

type Input<T extends string> = { [key in Parts<T>]: string }; // Record<"a" | "b" | "c", string>

const testTemplate = `
hello world
{foo}
hello world
{bar}
`;

const getTemplate = async <T extends string>(template: T, input: Input<T>) => {
  const llm = new OpenAI({});
  const promptTemplate = new PromptTemplate({
    template,
    inputVariables: parseTemplate(template, "f-string")
      .map((node) => (node.type === "variable" ? node.name : null))
      .filter((s): s is string => s !== null),
  });

  const chain = new LLMChain({
    llm,
    prompt: promptTemplate,
  });

  const { text } = await chain.call(input);

  return text;
};

getTemplate(testTemplate, {
  foo: "foo",
  bar: "bar",
}).then((t) => console.log(t));
```

bracesproul:
> @jacoblee93 any interest in picking this back up?
> 
> a while ago (may be incompatible with more recent versions of langchain) I prototyped adding some type safety to inputs
> 
> ```ts
> import { LLMChain, OpenAI, PromptTemplate } from "langchain";
> import { parseTemplate } from "langchain/prompts";
> 
> type Parts<Path> = Path extends `${infer _ignore}{${infer PartA}}${infer PartB}`
>   ? PartA | Parts<PartB>
>   : never;
> 
> type Input<T extends string> = { [key in Parts<T>]: string }; // Record<"a" | "b" | "c", string>
> 
> const testTemplate = `
> hello world
> {foo}
> hello world
> {bar}
> `;
> 
> const getTemplate = async <T extends string>(template: T, input: Input<T>) => {
>   const llm = new OpenAI({});
>   const promptTemplate = new PromptTemplate({
>     template,
>     inputVariables: parseTemplate(template, "f-string")
>       .map((node) => (node.type === "variable" ? node.name : null))
>       .filter((s): s is string => s !== null),
>   });
> 
>   const chain = new LLMChain({
>     llm,
>     prompt: promptTemplate,
>   });
> 
>   const { text } = await chain.call(input);
> 
>   return text;
> };
> 
> getTemplate(testTemplate, {
>   foo: "foo",
>   bar: "bar",
> }).then((t) => console.log(t));
> ```
> 
> Would love to see something like this built directly in!

Will look into this/take it on

PR #1289, opened 2023-05-16
chore: export conditional prompt selector

justindra:
I'm finding this Conditional Prompt Selector to be super useful. But it is not exported anywhere. So I've moved it to be under prompts and exported out so it can be used externally.


jacoblee93:
Thank you! Hopefully others will find it useful as well. We should add some docs about it at some point.

PR #1292, opened 2023-05-17
Improve chat conversation agent reliability by using an output fixing parser

jacoblee93:
Should address and fix parsing issues for the chat conversation agent raised in #1203 around tools that return stringified JSON by adding a retry mechanism for poorly formatted outputs. Should be more flexible for a wider range of failures vs. excessive prompt/output parsing changes.


nfcampos:
Great!

PR #1298, opened 2023-05-17
Fix env detection for Deno in Supabase Edge Functions

### LINKED ISSUE #1268: LLM run errored with error: t is not a function
Was trying out the new MultiPromptChain when I encountered the error below.
Tried debugging but couldn't progress much.

Trying this example - https://js.langchain.com/docs/modules/chains/other_chains/multi_prompt_chain

Runtime: Deno - Supabase edge function
Supabase Version - 1.49.4 (as recommended [here](https://github.com/langchain-ai/langchain-template-supabase/)
Lib CDN: https://esm.sh/langchain@0.0.75

It works on Node (using the v0.0.75 package from npm) buts fails on Deno.

image

dqbd:
Closes #1268

It does seem like Supabase Deno Functions only provides a global Deno functions _without_ `versions` property, which will mark `isDeno` as `false` and because of NodeJS compat, will mark `isNode` as `true`.

This PR aims to correct both of the properties, which is required to force the `axios-fetch-adapter` in these environments.

### DIFF
--- a/langchain/src/util/env.ts
+++ b/langchain/src/util/env.ts
@@ -9,15 +9,15 @@ declare global {
     | undefined;
 }
 
-export const isBrowser =
+export const isBrowser = () =>
   typeof window !== "undefined" && typeof window.document !== "undefined";
 
-export const isWebWorker =
+export const isWebWorker = () =>
   typeof globalThis === "object" &&
   globalThis.constructor &&
   globalThis.constructor.name === "DedicatedWorkerGlobalScope";
 
-export const isJsDom =
+export const isJsDom = () =>
   (typeof window !== "undefined" && window.name === "nodejs") ||
   (typeof navigator !== "undefined" &&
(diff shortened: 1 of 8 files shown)


nfcampos (approved):
Looks great!

langchain/src/util/env.ts
```diff
@@ -1,22 +1,50 @@
...
+  (typeof navigator !== "undefined" &&
+    (navigator.userAgent.includes("Node.js") ||
+      navigator.userAgent.includes("jsdom")));
+
+// Supabase Edge Function provides a `Deno` global object
+// without `version` property
+export const isDeno = () => typeof Deno !== "undefined";
+
+// Mark not-as-node if in Supabase Edge Function
+export const isNode = () =>
```
dqbd, line 31: Have tried using `/* #__PURE__*/`, but without much luck

nfcampos (approved):
Great

PR #1305, opened 2023-05-17
Add couple of run and session crud

vowelparrot:


langchain/src/client/langchainplus.ts
```diff
@@ -234,6 +244,63 @@ export class LangChainPlusClient {
...
+  public async readRun(runId: string): Promise<Run> {
+    return await this._get<Run>(`/runs/${runId}`);
+  }
+
+  public async listRuns({
+    sessionId,
+    sessionName,
+    executionOrder = 1,
+    runType,
+  }: ListRunsParams): Promise<Run[]> {
```
vowelparrot, line 256: I'm being inconsistent here... What would you recommend as the preferred way to specify args?

PR #1306, opened 2023-05-17
Fix escaping of special characters in JsonListKeyTool

### LINKED ISSUE #1300: Fix prompts for JsonListKeysTool and JsonGetValueTool to work with paths containing `/`
There is specific syntax to escape `/` in jsonpointer strings, `/paths/~1IDSGenericFXCrossRate` means `['paths', '/IDSGenericFXCrossRate']`, see https://datatracker.ietf.org/doc/html/rfc6901#section-3

dqbd:
This PR addresses escaping special characters (`/` and `~`) when listing keys in JsonListKeyTool.

The following agent will now work properly:

```typescript
import { JsonSpec, JsonObject } from "langchain/tools";
import { JsonToolkit, createJsonAgent } from "langchain/agents";

  const data: JsonObject = {
    paths: {
      "/IDSGenericFXCrossRate": "32",
    },
  };

  const toolkit = new JsonToolkit(new JsonSpec(data));
  const executor = createJsonAgent(model, toolkit);

  const input = `What is the value found in IDSGenericFXCrossRate?`;

  console.log(`Executing with input "${input}"...`);

  const result = await executor.call({ input });

  console.log(`Got output ${result.output}`);

  console.log(
    `Got intermediate steps ${JSON.stringify(
      result.intermediateSteps,
      null,
      2
    )}`
  );
};
```

Closes #1300

### DIFF
--- a/langchain/src/agents/tests/json.test.ts
+++ b/langchain/src/agents/tests/json.test.ts
@@ -29,6 +29,22 @@ test("JsonListKeysTool", async () => {
   expect(await jsonListKeysTool.call("/bar")).toContain("not a dictionary");
 });
 
+test("JsonListKeysTool, paths containing escaped characters", async () => {
+  const jsonSpec = new JsonSpec({
+    paths: {
+      "a~b": 1,
+      "a/b": 2,
+      "a~/b": 3,
+      "a//~b": 4,
+    },
+  });
+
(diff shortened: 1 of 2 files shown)


nfcampos (approved):
Great!

PR #1307, opened 2023-05-18
Handle incompatibility of `ChatOpenAI` with Azure OpenAI

### LINKED ISSUE #1299: Incompatibility of `ChatOpenAI` with Azure OpenAI due to EventStream Handling
### Description

Currently, the `langchain.js` library is unable to successfully resolve `ChatOpenAI` when used with Azure OpenAI streaming because of the differences in the responses from OpenAI's and Azure's APIs. 

The OpenAI's chat completions API ends with a blank line:

response from openai

In contrast, Azure's OpenAI API doesn't send a blank line at the end:

response from azure openai

`Langchain.js` uses `src/event-source-parse.ts`, originally copied from [fetch-event-source](https://github.com/gfortaine/fetch-event-source/blob/main/src/parse.ts), to handle EventSource. This code dispatches `onMessage` when a blank line is encountered, based on the standard:

> * If the line is empty (a blank line)
>     * Dispatch the event, as defined below.
>
> *[9.2.6 Interpreting an event stream](https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation)*

https://github.com/hwchase17/langchainjs/blob/d60eae5995a0d414f7684fc9c113c91c712c7af0/langchain/src/util/event-source-parse.ts#L148-L149

The problem arises in `ChatOpenAI`, where due to the absence of a blank line in Azure's OpenAI response, the promise never resolves:

https://github.com/hwchase17/langchainjs/blob/d60eae5995a0d414f7684fc9c113c91c712c7af0/langchain/src/chat_models/openai.ts#L302-L308

### Possible Solutions

1. **Microsoft sends a blank line at the end of response**: This would make Azure's OpenAI compatible with OpenAI's API. However, convincing Microsoft to alter their API behavior may be difficult, especially since their current implementation is valid according to the standard.

2. **Patch `src/event-source-parse.ts`**: Modifying this module to allow handling of Azure's OpenAI response without requiring a blank line could solve the issue.

In order to maintain broad compatibility with different APIs, I recommend exploring the second solution. We need to ensure that our library can handle responses from different APIs that follow the standard, even if they interpret it slightly differently.

yuku:
This pull request introduces several enhancements to the handling of event stream parsing. This pull request closes #1299.

In summary, the changes made include:

1. Fixing existing test cases to handle an additional 'done' flag, which ensures that all message chunks have been successfully processed from the stream.
2. Introducing a new test case designed to simulate the behaviour of Azure OpenAI's event streams. This case was initially expected to fail as it illustrated the problem of the last message being ignored due to the absence of a trailing blank line.
3. Implementing a 'flush' mechanism in the stream processing functions within `event-source-parse.ts`. This mechanism ensures that any pending messages are correctly processed when the stream ends, addressing the issue identified in the newly introduced test case.

These modifications are essential to ensuring complete and accurate processing of all messages, particularly from streams such as Azure OpenAI, which may not always terminate with a blank line.

By implementing these changes, we enhance the library's robustness in handling varying event stream behaviours across different platforms, making 'langchainjs' a more flexible and reliable tool for developers.

### DIFF
--- a/langchain/src/util/tests/openai-stream.test.ts
+++ b/langchain/src/util/tests/openai-stream.test.ts
@@ -33,46 +33,56 @@ const mockFetchForOpenAIStream = async ({
   );
 
   let error: Error | null = null;
+  let done = false;
   const receivedChunks: Array<string> = [];
   const resp = await fetchAdapter({
     url: "https://example.com",
     method: "POST",
     responseType: "stream",
-    onmessage: (strChunk: { data: string }) => {
-      receivedChunks.push(
-        JSON.parse(strChunk.data).choices[0].delta.content ?? ""
-      );
+    onmessage: (event: { data: string }) => {
+      if (event.data?.trim?.() === "[DONE]") {
+        done = true;
+      } else {
+        receivedChunks.push(
+          JSON.parse(event.data).choices[0].delta.content ?? ""
+        );
+      }
     },
   } as unknown as never).catch((err) => {
     error = err;
     return null;
   });
 
-  return { resp, receivedChunks, error } as {
+  return { resp, receivedChunks, error, done } as {
     resp: AxiosResponse | null;
     receivedChunks: Array<string>;
     error: Error | null;
+    done: boolean;
   };
 };
 
 describe("OpenAI Stream Tests", () => {
   it("should return a 200 response chunk by chunk", async () => {
-    // When stream mode enabled, OpenAI responds with a stream of `data: {...}\n\n` chunks.
-    const { resp, receivedChunks, error } = await mockFetchForOpenAIStream({
-      status: 200,
-      contentType: "text/event-stream",
-      chunks: [
-        'data: {"choices":[{"delta":{"role":"assistant"},"index":0,"finish_reason":null}]}\n\n',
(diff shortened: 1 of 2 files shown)


langchain/src/util/tests/openai-stream.test.ts
```diff
@@ -131,3 +141,30 @@ describe("OpenAI Stream Tests", () => {
     expect(receivedChunks).toEqual([]);
   });
 });
+
+describe("Azure OpenAI Stream Tests", () => {
+  it("should return a 200 response chunk by chunk", async () => {
+    // When stream mode enabled, Azure OpenAI responds with chunks without tailing blank line.
+    // In addition, Azure sends chunks in batch.
```
yuku, line 148: OpenAI sends data line by line:

https://github.com/hwchase17/langchainjs/assets/96157/5bd0173f-6aff-4e38-a9b4-ce09e07d2cbc

Azure OpenAI sends in batch:

https://github.com/hwchase17/langchainjs/assets/96157/7d7b171e-7d26-4f15-9a13-deb91cd2f62b

nfcampos (approved):
This is great, many thanks

PR #1320, opened 2023-05-18
Add getPromptAsync method to prompt selectors, deprecate existing sync methods in favor of async methods

jacoblee93:
Will allow us to apply partial variables to prompts selected with a `PromptSelector`.

### DIFF
--- a/docs/docs/modules/chains/prompt_selectors/index.mdx
+++ b/docs/docs/modules/chains/prompt_selectors/index.mdx
@@ -34,11 +34,11 @@ This will return `DEFAULT_QA_PROMPT` if the model is not a chat model, and `CHAT
 The example below shows how to use a prompt selector when loading a chain:
 
 ```typescript
-const loadQAStuffChain = (
+const initializeQAStuffChain = async (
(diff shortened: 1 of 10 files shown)


jacoblee93:
Will allow us to better support use cases like: #1176

docs/docs/modules/chains/prompt_selectors/index.mdx
```diff
@@ -34,11 +34,11 @@ This will return `DEFAULT_QA_PROMPT` if the model is not a chat model, and `CHAT
 The example below shows how to use a prompt selector when loading a chain:
 
 ```typescript
-const loadQAStuffChain = (
+const initializeQAStuffChain = async (
   llm: BaseLanguageModel,
   params: StuffQAChainParams = {}
 ) => {
-  const { prompt = QA_PROMPT_SELECTOR.getPrompt(llm) } = params;
+  const { prompt = await QA_PROMPT_SELECTOR.getPromptAsync(llm) } = params;
```
nfcampos, line 41: I don't think we need to make this change now, we can do this in a separate PR

PR #1321, opened 2023-05-18
Allow creation of a SQL toolkit with a custom language model

### LINKED ISSUE #1202: SqlToolkit does not allow for setting OpenAI API Key outside of environment variables.
When instantiating the `SqlToolkit`, by default it adds the `QueryCheckerTool`. The tool takes an optional `LLMChain` as an argument but the `SqlToolkit` does not pass this option. If the `LLMChain` is not passed, a default `LLMChain` is created using the `OPENAI_API_KEY` environment variable but if your app passes the key through something other than environment variables this will break.

Suggestion:

- Modify the `QueryCheckerTool` to accept a custom `BaseLLM`, use this `LLM` to create the internal `LLMChain`.
- Modify `SqlToolkit` to accept a custom `BaseLLM`, pass this `LLM` to `QueryCheckerTool` on instantiation.

jacoblee93:
Addresses and fixes #1202 by allowing the user to pass a preconfigured model into a SQL Toolkit.

### DIFF
--- a/langchain/src/tools/sql.ts
+++ b/langchain/src/tools/sql.ts
@@ -4,6 +4,7 @@ import { LLMChain } from "../chains/llm_chain.js";
 import { PromptTemplate } from "../prompts/prompt.js";
 import type { SqlDatabase } from "../sql_db.js";
 import { SqlTable } from "../util/sql_utils.js";
+import { BaseLanguageModel } from "../base_language/index.js";
 
 interface SqlTool {
   db: SqlDatabase;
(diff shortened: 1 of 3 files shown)


langchain/src/tools/sql.ts
```diff
@@ -103,17 +110,22 @@ If there are any of the above mistakes, rewrite the query. If there are no mista
 
   llmChain: LLMChain;
 
-  constructor(llmChain?: LLMChain) {
+  constructor(llmChainOrOptions?: LLMChain | QueryCheckerToolArgs) {
     super();
-    if (llmChain) {
-      this.llmChain = llmChain;
+    if (typeof llmChainOrOptions?._chainType === "function") {
```
jacoblee93, line 115: CC @nfcampos saw this elsewhere to determine whether an object was a chain, but does this look ok with you?

PR #1322, opened 2023-05-19
Export calculateMaxTokens

kushp:
The calculateMaxTokens is a handy utility. I'll like to use it along with some math to provide max tokens. I essentially want to do minimum of some value and the output of calculateMaxTokens to save on API limits. 

For example, calculateMaxTokens might yield some value like 2000 for a given model and prompt, but I want to do something like minimum of 128 and 2000. In that case my max tokens will be 128. However, if it yields 100 I would use 100 for max tokens.


jacoblee93:
I'm not sure if this is stable enough yet - @nfcampos to weigh in?

kushp:
Thank you @jacoblee93 @nfcampos, apologies for the CI failure, I ran prettier and amended my commit. Should be good now!

jacoblee93:
Ok! Will merge and it'll go out with the next release. Thank you!

PR #1325, opened 2023-05-19
Allow passing options to LLM chains inside MultiRetrievalQAChain and MultiPromptChain.

ppramesi:
Changed MultiRetrievalQAChain.fromRetrievers parameter into one option. Added multiRetrievalChainOpts and retrievalQAChainOpts into the fromRetrievers's parameters, so we can pass options to the multi retrieval chain and retrieval QA chains (so e.g. we can pass returnSourceDocuments to the RetrievalQAChain).

Same with MultiPromptChain.fromPrompts.


jacoblee93:
So of course this will break anyone currently using this, but looks like this was a very recent addition anyway - will leave it up to @nfcampos

jacoblee93:
So I think the best option since the behavior isn't that different and it's now been out for a little while is to just deprecate the old methods and have them wrap new ones, `fromLLMAndPrompts` and `fromLLMAndRetrievers`.

Does that sound ok to you @ppramesi?

ppramesi:
> So I think the best option since the behavior isn't that different and it's now been out for a little while is to just deprecate the old methods and have them wrap new ones, `fromLLMAndPrompts` and `fromLLMAndRetrievers`.
> 
> Does that sound ok to you @ppramesi?

Sounds good to me

jacoblee93:
Alright cool, it'll be part of tomorrow's release :)

PR #1343, opened 2023-05-20
Update motorhead_memory.md

bfpill:
motorheadUrl -> motorheadURL 

Assuming this is what was wanted, motorheadURL is the correct property name on class MotorheadMemory.


jacoblee93:
Thank you!

PR #1345, opened 2023-05-20
change typeorm vectorstore to handle better the updates of embeddings

MaximeThoonsen:
This allows the use of typeorm to interact with postgresql database to store vectors.
Also it implements an algo to avoid recreate embeddings if not needed.
See my [article here](https://www.maximeai.com/blog/scale-your-embeddings-with-langchain).
Small thread on [discord here](https://discord.com/channels/1038097195422978059/1106938194085285928)
The [tweet about my article](https://twitter.com/maxthoon/status/1657336947629490176)


MaximeThoonsen:
Implement typeorm vectorstore + implement algo to avoid recreate embeddings if not needed. 
See my [article here](https://www.maximeai.com/blog/scale-your-embeddings-with-langchain)

langchain/src/vectorstores/typeorm.ts
```diff
@@ -0,0 +1,307 @@
...
+        const document = new Document(doc) as TypeORMVectorStoreDocument;
+        document.id = doc.id;
+        results.push([document, doc._distance]);
+      }
+    }
+
+    return results;
+  }
+
+  async initTable(): Promise<void> {
```
MaximeThoonsen, line 239: was wondering if we should call this function at the beginning of fromDocuments function to be sure the table and the extension exist as it doesn't cost much in time.

langchain/src/vectorstores/typeorm.ts
```diff
@@ -0,0 +1,307 @@
...
+      try {
+        await documentRepository.upsert(chunk, [`id`]);
+      } catch (e) {
+        console.error(e);
+        throw new Error(`Error inserting: ${chunk[0].pageContent}`);
+      }
+    }
+  }
+
+  async similaritySearchVectorWithScore(
```
MaximeThoonsen, line 211: nothing new

langchain/src/vectorstores/typeorm.ts
```diff
@@ -0,0 +1,307 @@
...
+        metadata jsonb,
+        "sourceType" text,
+        "sourceName" text,
+        hash text,
+        embedding vector
+      );
+    `);
+  }
+
+  static async fromTexts(
```
MaximeThoonsen, line 255: I add a split of the texts because either the user has already done it and it should be transparent or it hasn't and it could lead to errors. I had people asking for my help because of this.

langchain/src/vectorstores/typeorm.ts
```diff
@@ -0,0 +1,307 @@
...
+    }
+    const splitter = new CharacterTextSplitter();
+
+    const splittedDocs = await splitter.splitDocuments(docs);
+
+    return TypeORMVectorStore.fromDocuments(splittedDocs, embeddings, dbConfig);
+  }
+
+  static async fromDocuments(
+    docs: Document[],
```
MaximeThoonsen, line 278: nothing new

langchain/src/vectorstores/typeorm.ts
```diff
@@ -0,0 +1,307 @@
...
+      embeddings,
+      dbConfig
+    );
+    await instance.addDocuments(docs);
+
+    return instance;
+  }
+
+  static async fromExistingIndex(
+    embeddings: Embeddings,
```
MaximeThoonsen, line 292: nothing new

langchain/src/vectorstores/typeorm.ts
```diff
@@ -0,0 +1,307 @@
...
+    super(embeddings, fields);
+    this.tableName = fields.tableName || defaultDocumentTableName;
+    this.filter = fields.filter;
+  }
+
+  static async fromDataSource(
+    embeddings: Embeddings,
+    fields: TypeORMVectorStoreArgs
+  ): Promise<TypeORMVectorStore> {
+    const TypeORMDocumentEntity = new EntitySchema<TypeORMVectorStoreDocument>({
```
MaximeThoonsen, line 52: this is the way to construct dynamically a table in TypeORM

langchain/src/vectorstores/typeorm.ts
```diff
@@ -0,0 +1,307 @@
...
+      await this.embeddings.embedDocuments(texts),
+      documents
+    );
+  }
+
+  async addVectors(vectors: number[][], documents: Document[]): Promise<void> {
+    const { createHash } = await import("node:crypto");
+    const rows = vectors.map((embedding, idx) => {
+      const embeddingString = `[${embedding.join(",")}]`;
+      const hash = createHash("sha1");
```
MaximeThoonsen, line 114: sha1 is quick and there is no security issue here

langchain/src/vectorstores/typeorm.ts
```diff
@@ -0,0 +1,307 @@
...
+      );
+
+      await documentRepository.delete({
+        id: {
+          in: idsDocumentsThatWeShouldDeleted,
+        },
+      });
+    }
+
+    const documentsToUpsert = [];
```
MaximeThoonsen, line 170: we compare the hash of the document in the database with the one we want to insert to see if we need to recreate the embeddings. It's faster and cost less

jacoblee93:
Started looking at this and then realized @dqbd had already started looking - will leave it to you!

langchain/src/document.ts
```diff
@@ -17,12 +21,23 @@ export class Document<
 {
   pageContent: string;
 
+  id?: string;
```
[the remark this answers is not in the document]
MaximeThoonsen: @dqbd to keep it simple I would keep one class. The fields are optionnal and the next step would be to update the document loaders to feed more info into this class

langchain/src/vectorstores/typeorm.ts
```diff
@@ -0,0 +1,307 @@
...
+
+  async similaritySearchVectorWithScore(
+    query: number[],
+    k: number,
+    filter?: this["FilterType"]
+  ): Promise<[TypeORMVectorStoreDocument, number][]> {
+    const embeddingString = `'[${query.join(",")}]'`;
+    const _filter = filter ?? "'{}'";
+    const queryString = `
+      SELECT *, embedding <=> ${embeddingString}::vector as "_distance"
```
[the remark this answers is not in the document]
MaximeThoonsen: @dqbd I guess the type of query which is number[] prevents a sql injection here.

This.tableName is at risk but only people with already an access to the db can modify it so I think it's safe for now

langchain/src/vectorstores/typeorm.ts
```diff
@@ -0,0 +1,307 @@
...
+  async addDocuments(documents: Document[]): Promise<void> {
+    const texts = documents.map(({ pageContent }) => pageContent);
+    return this.addVectors(
+      await this.embeddings.embedDocuments(texts),
+      documents
+    );
+  }
+
+  async addVectors(vectors: number[][], documents: Document[]): Promise<void> {
+    const { createHash } = await import("node:crypto");
```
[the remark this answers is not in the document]
MaximeThoonsen: @dqbd I'll dig that

langchain/src/vectorstores/typeorm.ts
```diff
@@ -0,0 +1,307 @@
+import { Metadata } from "@opensearch-project/opensearch/api/types.js";
+import {
+  DataSource as DataSourceT,
```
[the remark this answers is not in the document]
MaximeThoonsen: done

langchain/src/vectorstores/typeorm.ts
```diff
@@ -0,0 +1,307 @@
...
+
+  tableName: string;
+
+  documentEntity: EntitySchema;
+
+  filter?: Metadata;
+
+  appDataSource: DataSourceT;
+
+  _verbose?: boolean;
```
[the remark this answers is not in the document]
MaximeThoonsen: done

langchain/src/vectorstores/typeorm.ts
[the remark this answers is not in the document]
MaximeThoonsen: done

langchain/src/vectorstores/typeorm.ts
MaximeThoonsen: done

MaximeThoonsen:
hey @dqbd and @jacoblee93 . I took your feedbacks into account. Look good to you?

langchain/src/vectorstores/index.ts
```diff
@@ -8,3 +8,4 @@ export { PineconeStore } from "./pinecone.js";
 export { VectorStore, SaveableVectorStore } from "./base.js";
 export { SupabaseVectorStore } from "./supabase.js";
 export { PrismaVectorStore } from "./prisma.js";
+export { TypeORMVectorStore } from "./typeorm.js";
```
jacoblee93, line 11: Can't export from here as it requires an optional dep - will break people currently (and mistakenly) importing from here.

langchain/src/document.ts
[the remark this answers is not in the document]
jacoblee93: I'm with David, let's keep this general and in line with Python's implementation:

https://github.com/hwchase17/langchain/blob/master/langchain/schema.py#L269

Is there a reason these extra fields can't just go into `metadata`?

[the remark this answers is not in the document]
jacoblee93: Either way, it's out of scope for this PR. I'm up for opening another one and having a more focused discussion later!

langchain/src/vectorstores/typeorm.ts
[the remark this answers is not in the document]
jacoblee93: Always, always, always good to parameterize though (even just to show best practices for someone else looking through the code) and it's not hard. Should just be:
```
    const queryString = `
      SELECT *, embedding <=> $1::vector as "_distance" 
      FROM $2
      WHERE metadata @> $3
      ORDER BY "_distance" ASC
      LIMIT $4;
    const documents = await this.appDataSource.query(queryString, [param1, param2, param3, param4]);
```
Have had bad things happen to things that seem safe before - what happens if someone trusts some sketchy embeddings provider?

langchain/src/vectorstores/typeorm.ts
jacoblee93: You can call this `ensureTable` which might be a bit more clearly named.

jacoblee93:
Hey @MaximeThoonsen, this looks fantastic and useful but a bit overscoped - we definitely should be thoughtful about any changes we make to the base `Document` class and discuss them in their own PR. 

The hashing stuff is a great idea and is probably broadly useful in other vector stores as its own concept/module as well (though I'm cool with you adding it as a one-off here - just thinking ahead to a potential refactor!).

I'm happy to take over if you'd like - let me know.

langchain/src/vectorstores/typeorm.ts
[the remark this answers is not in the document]
jacoblee93: How does this actually get set? It's not in the superclass is it?

langchain/src/vectorstores/typeorm.ts
```diff
@@ -0,0 +1,322 @@
...
+    });
+
+    const appDataSource = new DataSource({
+      entities: [TypeORMDocumentEntity],
+      ...fields.postgresConnectionOptions,
+    });
+
+    const postgresqlVectorStore = new TypeORMVectorStore(embeddings, fields);
+    postgresqlVectorStore.appDataSource = appDataSource;
+    postgresqlVectorStore.documentEntity = TypeORMDocumentEntity;
```
jacoblee93, line 86: This should all go in the constructor IMO

langchain/src/vectorstores/typeorm.ts
```diff
@@ -0,0 +1,322 @@
...
+    // do anything if the table already exists, and it is not expensive in terms of performance
+    await this.initTable();
+    return this.addVectors(
+      await this.embeddings.embedDocuments(texts),
+      documents
+    );
+  }
+
+  async addVectors(vectors: number[][], documents: Document[]): Promise<void> {
+    const { createHash } = await TypeORMVectorStore.imports();
```
jacoblee93, line 113: We don't need to do this, just put it at the top.

langchain/src/vectorstores/typeorm.ts
jacoblee93: This would be a broader problem right? Let's make this a separate PR and try to think of a more general solution. There are a lot of different splitting strategies.

docs/docs/modules/indexes/vector_stores/integrations/typeorm.mdx
```diff
@@ -0,0 +1,81 @@
+---
+sidebar_class_name: node-only
+---
+
+# TypeORM
+
+For augmenting existing models in a PostgreSQL database with vector search, Langchain supports using [TypeORM](https://typeorm.io/) together with PostgreSQL and the [`pgvector`](https://github.com/pgvector/pgvector) Postgres extension.
```
jacoblee93, line 7: We're not really augmenting an existing model though right? This would be creating a new model in a database?

langchain/src/vectorstores/typeorm.ts
MaximeThoonsen: OK I removed it

langchain/src/vectorstores/typeorm.ts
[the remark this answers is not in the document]
MaximeThoonsen: I agree it's way better for good practices. sorry @dqbd I got lazy on this one :p.

langchain/src/document.ts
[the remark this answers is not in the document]
MaximeThoonsen: I don't want to put everything into metadata for few reasons:
- It's a easier for dev to understand columns rather than property inside a JSON object
- it's easier to manipulate the data in columns than in a JSON
- it will be easier for dev to enforce indexes on columns 
- SQL visualization tools work out of the box for columns but some of them won't work as fine for Json object

I need it to add the concept of not recreating all embeddings.

langchain/src/vectorstores/index.ts
MaximeThoonsen: ok

langchain/src/vectorstores/typeorm.ts
MaximeThoonsen: ok

docs/docs/modules/indexes/vector_stores/integrations/typeorm.mdx
MaximeThoonsen: indeed, I change this

MaximeThoonsen:
Hey @jacoblee93 ,
I agree that this PR is "add TypeORM vectorstore" + "new behavior when creating documents" which is a lot. 
If you prefer to do a separate PR we can, there is a bit of rework. I did it like it because I though it would be a good idea to add it in a new vectorstore because it doesn't break anything and let us assess before refacto the other vectorstore.

I also understand you want to separate it into multiple PR. 
Feel free to take over from here, I'm here to contribute, not to be annoying 😅
Thanks for your help to make it forward!

jacoblee93:
> Hey @jacoblee93 , I agree that this PR is "add TypeORM vectorstore" + "new behavior when creating documents" which is a lot. If you prefer to do a separate PR we can, there is a bit of rework. I did it like it because I though it would be a good idea to add it in a new vectorstore because it doesn't break anything and let us assess before refacto the other vectorstore.
> 
> I also understand you want to separate it into multiple PR. Feel free to take over from here, I'm here to contribute, not to be annoying 😅 Thanks for your help to make it forward!

Not at all - some really great ideas here! We just don't want to break anyone or create abstractions without planning them out first.

I'll make some updates and pass it back to you for an OK in a few days.

MaximeThoonsen:
Hey @jacoblee93, 
I felt guilty of making you work to clean my stuff so I did a [PR](https://github.com/hwchase17/langchainjs/pull/1436) with only the addition of typeorm as a vectorstore and took into all comments you did.
I guess it will be easier to merge this one and then we can discuss for the other changes I suggested.

MaximeThoonsen:
Hey @jacoblee93 ,

To finish this one, I see two decisions:

1 - what do we do with the Document class?
- we extend the Document class (my recommendation but I agree we need to agree with the python implem)
- we create a new EnrichedDocument class
- other option?

I don't want to put everything into metadata for few reasons:
- It's a easier for dev to understand columns rather than property inside a JSON object
- it's easier to manipulate the data in columns than in a JSON
- it will be easier for dev to enforce indexes on columns
- SQL visualization tools work out of the box for columns but some of them won't work as fine for Json object

2 - how do we introduce the upsert mechnism with the hash?
- we update the addVectors method (this PR)
- I create an upsertVectors method so we don't do any break change (new idea)

What do you think?

Thanks again for your help on this,

Maxime

jacoblee93:
Hey @MaximeThoonsen, sorry for all the pings - I opened #1562 and #1563 to talk about the issues raised here. Going to close this one, thank you for your patience!

PR #1347, opened 2023-05-21
Add support for memoryPrompts in StructuredChatAgent

andrewlei:
- Add support for `memoryPrompts` in `StructuredChatAgent`, inspired by python library: https://github.com/hwchase17/langchain/blob/master/langchain/agents/structured_chat/base.py#L94
- Fix the `inputVariables` provided by `StructuredChatCreatePromptArgs`


jacoblee93:
We'd love to support memory in this agent, but the ideal implementation would probably look more like this:

https://github.com/hwchase17/langchainjs/blob/main/langchain/src/agents/chat_convo/index.ts

For this use-case, I'd suggest just passing a custom `LLMChain` with a custom prompt into the constructor directly - happy to give an example if helpful!

jacoblee93:
Seems like I was presumptive in closing this as it's what Python's implemented - reopening!

emilsedgh:
GH-1414 is another attempt at providing memory support for this agent that is closer in spirit to the rest of ts codebase.

PR #1348, opened 2023-05-21
Added filter to ChromaDB vector store + update chroma package.

ppramesi:
The old version of ChromaDB js package doesn't work with the current docker image, so the package is updated to 1.4.2. 

Also added filter to Chroma vector store + ChromaDB integration test.


jacoblee93:
This is excellent, thank you! 

Only changes were added a example to the docs and changed the peer dep from a wildcard to caret.

PR #1351, opened 2023-05-21
chore: remove useless code

L-Qun:
I found that after adding the sqlOutputKey, the intermediateStep had no effect, so I removed it


jacoblee93:
Nice catch, thank you!

PR #1354, opened 2023-05-21
fix input keys in refine and stuff chains

FluffyKebab:
In the refine and stuff chain the input keys are the input keys of the llm chains used and the input key of the chain. This causes issues because in these chains some of the inputs to the llm chain are given by the chain: 

```typescript
const chain = loadSummarizationChain(model, { type: "refine" });
const textSplitter = new RecursiveCharacterTextSplitter({ chunkSize: 1000 });
const docs = await textSplitter.createDocuments(["test"]);

console.log(chain.inputKeys)
// ['input_documents', 'existing_answer', 'text']. Should be ['input_documents']

console.log(await chain.call({"input_documents":docs}))
// output_text: Test is a process used to evaluate a syst…termine its performance and functionality.'

console.log(await chain.run(docs))
// Uncaught Error Error: Chain refine_documents_chain expects multiple inputs, cannot use 'run' 
```

This fixes the bug by filtering out the input keys of the llm chains added by the refine and stuff chains and adds a test for the refine chain.

### DIFF
--- a/langchain/src/chains/combine_docs_chain.ts
+++ b/langchain/src/chains/combine_docs_chain.ts
@@ -38,7 +38,7 @@ export class StuffDocumentsChain
 
   get inputKeys() {
     return [this.inputKey, ...this.llmChain.inputKeys].filter(
-      (key) => key != this.documentVariableName
+      (key) => key !== this.documentVariableName
     );
   }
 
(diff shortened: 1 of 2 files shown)


nfcampos (approved):
Thanks! This is great

PR #1356, opened 2023-05-22
fix: query_constructor prompt fix

ppramesi:
Fix a misspelling and added double quote on json formatting.


jacoblee93:
Still seemed to pass integration tests for me - thanks for the fixes!

PR #1358, opened 2023-05-22
chore: fixing docs - document_qa

orrgal1:


orrgal1:
Pleasure. Will make more of these as I find them. Hopefully will also get to code contribution at some point :)

PR #1359, opened 2023-05-22
Make API URL optional when specifed via env var

vowelparrot:
Add env var defaults and make it not required in the `create()` method

### DIFF
--- a/langchain/src/client/langchainplus.ts
+++ b/langchain/src/client/langchainplus.ts
@@ -16,6 +16,7 @@ import { BaseLLM } from "../llms/base.js";
 import { BaseChatModel } from "../chat_models/base.js";
 import { mapStoredMessagesToChatMessages } from "../stores/message/utils.js";
 import { AsyncCaller, AsyncCallerParams } from "../util/async_caller.js";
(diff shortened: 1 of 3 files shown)


langchain/src/client/langchainplus.ts
```diff
@@ -203,11 +211,24 @@ export class LangChainPlusClient {
   }
 
   public static async create(
-    apiUrl: string,
+    apiUrl: string | undefined = undefined,
```
jacoblee93, line 214: Would it be too late to make an interface change here and in the constructor?

I think a config object makes more sense given things are under flux and it's bit uncertain what's going to be optional and what isn't:

```
  public static async create(config: {apiUrl?: string, apiKey?: string}) {}
```

jacoblee93: I usually only like args if it's an important required parameter (`tenantId` might qualify in this case) - LMK and I'm happy to do some refactoring

vowelparrot: This is the perfect time to do that. I'll go over the other methods to make the requested changes as well

PR #1367, opened 2023-05-23
docs: Add file path resolves for nodejs loaders

mvikrammenon:
Would this be needed for all loaders examples?


jacoblee93:
Thanks for opening a PR. I think it's a bit too opaque - the examples in that case are just illustrative, and aren't intended to be taken literally. It could be replaced with `/path/to/your/file`.

PR #1369, opened 2023-05-23
Adds contribution guidelines for integrations

jacoblee93:
Some sections marked as TODO, but this should be a good start!


.github/contributing/integrations/MEMORY.md
```diff
@@ -0,0 +1,22 @@
+# Contributing third-party memory
+
+This page contains some specific guidelines and examples for contributing integrations with third-party memory providers.
```
nfcampos, line 3: Let's cross link here to MESSAGE_STORES, as many times when people think they want to add a memory class they actually want to add a message store

PR #1374, opened 2023-05-23
Fix redis chat message history message hydration

### LINKED ISSUE #1339: Chat History not beeing retrieved / saved
Hi,
I am wondering if anyone else experience the same issue ?
I had some issues on my project and found out examples were returning an anomaly.

image

Buffer Memory is working fine.
image

I have noticed that the error appeared when model used was ChatOpenAI, but not when it is OpenAI. But in both case I can't seems to be able to save history.

nfcampos:
Closes #1339

### DIFF
--- a/langchain/src/stores/message/redis.ts
+++ b/langchain/src/stores/message/redis.ts
@@ -7,7 +7,6 @@ import {
   RedisScripts,
 } from "redis";
 import {
-  StoredMessage,
   BaseChatMessage,
   BaseListChatMessageHistory,
 } from "../../schema/index.js";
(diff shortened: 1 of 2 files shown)


langchain/src/stores/message/redis.ts
```diff
@@ -57,19 +56,7 @@ export class RedisChatMessageHistory extends BaseListChatMessageHistory {
     const orderedMessages = rawStoredMessages
       .reverse()
       .map((message) => JSON.parse(message));
-    const previousMessages = orderedMessages
```
vowelparrot, line 60: Ah so it was being assigned empty values before going to the `mapStoredMessagesToChatMessages` :( thanks for catching

vowelparrot (commented):
I think dynamoDB also needs fixed!

langchain/src/stores/tests/redis.int.test.ts
```diff
@@ -15,7 +15,13 @@ afterAll(async () => {
   await client.disconnect();
 });
 
-test.skip("Test Redis history store", async () => {
+/**
```
vowelparrot, line 18: Helpful comment. Thanks!

nfcampos:
@vowelparrot no I tested dynamo and it seems fine, because it has its own custom format for saving (not ideal)

vowelparrot (approved):

PR #1382, opened 2023-05-23
PR for merging adding Zep Memory support to LangchainJS

rsharath:
Hi Harrison, 

In continuation of providing Zep Memory support to LangChain, I am creating a PR for adding Zep Memory support to LangchainJS. The changes include Zep Memory support for both Retrievers and Memory with relevant example and doc. Please let me know if you are good with this. Thanks! Sharath


rsharath:
@jacoblee93 I have made the update recommended and removed export from index.ts. Also made some minor edits, cleanups. Please take a look and let me know if you see any other concerns/issues. Good to go from my end.

docs/docs/modules/memory/examples/zep_memory.md
```diff
@@ -0,0 +1,156 @@
+# Zep Memory
```
jacoblee93, line 1: Can you convert this to an `.mdx` file? Helps with making sure examples are linted and formatted correctly:

https://github.com/hwchase17/langchainjs/blob/main/.github/contributing/INTEGRATIONS.md#documentation-and-integration-tests

package.json
```diff
@@ -50,7 +50,10 @@
     "typescript": "^5.0.0"
   },
   "dependencies": {
-    "turbo": "latest"
+    "@docusaurus/core": "latest",
```
jacoblee93, line 53: This shouldn't change - can you revert?

langchain/src/memory/zep_memory.ts
```diff
@@ -0,0 +1,252 @@
...
+ *
+ *  For server installation instructions and more, see: https://getzep.github.io/
+ *  This class is a thin wrapper around the zep-python package. Additional
+ *  Zep functionality is exposed via the zepSummary and zepMessages properties.
+ *
+ *  For more information on the zep-python package, see:
+ *  https://github.com/getzep/zep-js
+ *
+ */
+export class ZepChatMessageHistory extends BaseChatMessageHistory {
```
jacoblee93, line 42: I think this should live in `src/stores/zep.ts`? Could you theoretically use Zep with other message stores like Redis with the proper adapter?

Message stores are generally intended to be simple storage - can we move more of the processing into the memory class?

https://github.com/hwchase17/langchainjs/blob/main/.github/contributing/integrations/MESSAGE_STORES.md

jacoblee93 (changes requested):
A few questions and comments - I can make some of these changes when I have more time as well.

docs/docs/modules/memory/examples/zep_memory.md
rsharath: Done

docs/package.json
```diff
@@ -22,6 +22,7 @@
     "@docusaurus/core": "2.4.0",
     "@docusaurus/preset-classic": "2.4.0",
     "@docusaurus/remark-plugin-npm2yarn": "^2.4.0",
+    "@getzep/zep-js": "^0.2.9",
```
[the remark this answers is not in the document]
rsharath: Done

langchain/package.json
```diff
@@ -425,6 +429,7 @@
     "@aws-sdk/client-sagemaker-runtime": "^3.310.0",
     "@clickhouse/client": "^0.0.14",
     "@getmetal/metal-sdk": "*",
+    "@getzep/zep-js": "*",
```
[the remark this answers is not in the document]
rsharath: Done

docs/docs/modules/memory/examples/zep_memory.md
rsharath: It appears like API References that are auto-generated by docusaurus is not picking up class definition for zep_memory classes. Possibly because i am using hard path in the memory example imports. in my memory example under example/src/memory/zep.ts: I am unable to import {
  ZepChatMessageHistory,
  ZepMemory,
  ZepMemoryInput,
} from "**_langchain/memory/_**"

and instead have to use:
 from "**_../../../langchain/src/memory/zep_memory.js_**";

Side effect is that doc build is not able to autofind the class definition docs for zep_memory. 

Any ideas why this may be the case?

langchain/src/memory/zep_memory.ts
rsharath: zep_memory is implemented as a memory class - since we summarize and process the chat histories (not just store them).

package.json
rsharath: done

docs/docs/modules/memory/examples/zep_memory.md
rsharath: Added a create_endpoint for memory (thanks for the tip @jacoblee93 I appear to have missed this for memory). Done

langchain/src/memory/zep_memory.ts
rsharath: Done

jacoblee93:
I'm unable to run the unit test:

Screenshot 2023-05-25 at 2 40 34 PM

And the example in the docs fails with:

Screenshot 2023-05-25 at 2 41 34 PM

I tried to follow the instructions in the docs and got the docker container running ok, but am not sure where to find a session id.

rsharath:
@jacoblee93 I have refactored the unit/int tests, should work fine. Please see github.com/getzep for docker image. Let me know if you run into any issues.

examples/src/memory/zep.ts
```diff
@@ -0,0 +1,143 @@
+import { BufferMemory } from "langchain/memory";
+import { ChatOpenAI } from "langchain/chat_models/openai";
+import { ConversationChain } from "langchain/chains";
+
+import { ZepChatMessageHistory } from "langchain/memory/";
```
jacoblee93, line 5: I think you have the right abstractions but are presenting the wrong interface in the example here - people will be using these in chains, which themselves take a memory object directly, not a `ChatMessageHistory`:

https://js.langchain.com/docs/modules/memory/examples/motorhead_memory#usage

jacoblee93:
I think we're almost there!

examples/src/memory/zep.ts
rsharath: @jacoblee93 thanks. There are two ways you can use Zep Memory from Langchain:
// Option 1: Use ZepMemory directly:
import { ConversationChain } from "langchain/chains";
import { ZepMemory } from "langchain/memory/zep_memory";
const memory = new ZepMemory(sessionID, zepURL);
const chain = new ConversationChain({ llm: model, memory });

// Option 2: Use ZepChatMessageHistory with BufferMemory:
import { ConversationChain } from "langchain/chains";
import { BufferMemory } from "langchain/memory";
import { ZepChatMessageHistory } from "langchain
const memory = new BufferMemory({chatHistory: new ZepChatMessageHistory(sessionID, zepURL)});
const chain = new ConversationChain({ llm: model, memory });

I have added 2 examples, this will show up in docs under Memory as Zep Memory and Zep Chat Memory. Please advise - happy to merge them if there is some meaningful way to do so or re-organize them as appropriate.

jacoblee93: Ahhh ok. Was a bit confused about that. What's the difference between 1 and 2? Why not just have option 1?

The right place for`ZepChatMessageHistory` if you want it exported is with the other message stores in `src/stores/zep.ts` (same place as the Redis example).

jacoblee93: I think my preference unless there's a really compelling time you'd want to do option 2 over option 1 would be keeping things simple: one memory class that internally takes care of storing messages and presenting those messages to the LLM (your memory class), then one retriever class responsible for loading from a chat session.

rsharath: @jacoblee93 Yes, for now thats fine. I have removed the example reference and the related doc for Option 2.

rsharath: I have also remove the related code to make this cleaner and refactored the int tests. Thanks, hope this is everything. If you see any issues please let me know.

Joshua-Shepherd: Thank you both for all your hard work on this 🙏

jacoblee93:
Closing since the extension PR #1450 was merged!

PR #1384, opened 2023-05-23
Feature: Google Vertex AI Chat model

afirstenberg:
Implements the Google Vertex AI PaLM chat model, as referenced in #1240 and building off the work first done in pull request #1273.

Note that the LangChain BaseChatMessage pattern is completely followed here, although Google-specific extensions are included (and documented through unit and integration tests, examples, and the documentation).

### DIFF
--- a/langchain/src/chat_models/googlevertexai.ts
+++ b/langchain/src/chat_models/googlevertexai.ts
@@ -1,0 +1,300 @@
+import { BaseChatModel } from "./base.js";
+import {
+  AIChatMessage,
+  BaseChatMessage,
+  ChatGeneration,
+  ChatResult,
+  LLMResult,
+  MessageType,
+  SystemChatMessage,
+} from "../schema/index.js";
+import { BaseLanguageModelCallOptions } from "../base_language/index.js";
+import { GoogleVertexAIConnection } from "../util/googlevertexai-connection.js";
+import {
+  GoogleVertexAIBaseLLMInput,
+  GoogleVertexAIBasePrediction,
+  GoogleVertexAILLMResponse,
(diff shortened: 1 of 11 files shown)


langchain/src/chat_models/googlevertexai.ts
```diff
@@ -0,0 +1,300 @@
...
+      context,
+      messages,
+    };
+  }
+
+  convertMessage(baseMessage: BaseChatMessage): GoogleVertexAIChatMessage {
+    const ret: GoogleVertexAIChatMessage = {
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
+      author: this.roleAlias[baseMessage._getType()],
```
jacoblee93, line 238: I'd rather hard-code an alias for now to limit the number of specific abstractions we're introducing here - `BaseChatMessages` have a `name` field that could override this in the future.

langchain/src/chat_models/googlevertexai.ts
```diff
@@ -0,0 +1,300 @@
...
+
+  topP = 0.8;
+
+  topK = 40;
+
+  context: string;
+
+  examples: ChatExample[];
+
+  roleAlias: RoleAlias = {
```
jacoblee93, line 109: Having a look here: https://cloud.google.com/vertex-ai/docs/generative-ai/chat/chat-prompts

It looks like `SystemChatMessages` could be mapped to a vertex message that has an `author` set to `context`? We can avoid the special case below if so.

jacoblee93:
Having a look now! Will have a commit up by the end of the day.

Martinlopez-hash:
Į got you bro

PR #1385, opened 2023-05-23
Add docs on hidden faiss-node dependency, change unit test to an integration test

jacoblee93:
Mitigates https://github.com/hwchase17/langchainjs/issues/1383.

Faiss-node has a hidden dependency on libomp, which causes unit tests to fail on some machines. This PR adds docs and changes the Faiss unit test to an integration test to signify the need for more setup.


jacoblee93:
CC @ewfian @nfcampos

PR #1386, opened 2023-05-24
Move axios from optional peer to dependency

dsherret:
axios ^0.26 is already a dependency of openai (https://www.npmjs.com/package/openai?activeTab=dependencies) and axios is used by the openai specific code (https://github.com/hwchase17/langchainjs/blob/29b69455ba73194fa42999bfc8c3b405f852116f/langchain/src/chat_models/openai.ts#L16). It seems this might as well just be made a required dependency given that openai is required and axios is already in the dependency tree.

### DIFF
--- a/langchain/package.json
+++ b/langchain/package.json
@@ -378,7 +378,6 @@
     "@typescript-eslint/parser": "^5.58.0",
     "@zilliz/milvus2-sdk-node": ">=2.2.7",
     "apify-client": "^2.7.1",
-    "axios": "^0.26.0",
     "cheerio": "^1.0.0-rc.12",
     "chromadb": "^1.4.2",
     "cohere-ai": "^5.0.2",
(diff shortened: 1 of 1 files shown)


jacoblee93:
Hey @dsherret, is there a visible benefit to this change?

dsherret:
@jacoblee93 yes, it removes the need to specify the optional peer in Deno:

```ts
import "https://deno.land/std@0.188.0/dotenv/load.ts";
// import "npm:axios"; // no longer necessary
import { OpenAI } from "npm:langchain@0.0.81/llms/openai";
```

jacoblee93:
I see - I think @nfcampos may have a bit more context as to some of the reasoning in `util/axios-fetch-adapter.js`. Will pass it along to him!

dsherret:
I'm going to close this PR. I believe the reason that langchain requires a peer dependency on axios, is because it exposes the `axios.AxiosError` type here, then people are checking for it when catching errors:

https://github.com/hwchase17/langchainjs/blob/01673d8496d27f0e84bb4ef4c51bf67eb6187f68/langchain/src/util/axios-fetch-adapter.js#L398-L406

I think maybe axios should be removed as a dependency, peer dependency, and dev dependency of langchain and instead a langchain specific error is thrown.

PR #1387, opened 2023-05-24
fix: Changed expression parser from meriyah to peggy.

ppramesi:
The current expression parser hits a snag with javascript's reserved words. To address this, I've crafted a custom grammar very similar to the old meriyah parser, but it allows all words (excluding "null", "true" and "false") as identifiers. This adjustment is critical for introducing the "in" filter to the self-query retriever, given that "in" is a reserved word in JavaScript - though only Pinecone allows "in" filter for now.


jacoblee93:
This looks great! Will review as soon as I can.

jacoblee93:
@ppramesi could you add an example test case that fails with the old parser and succeeds with the new one?

langchain/src/output_parsers/expression_type_handlers/grammar/parser_grammar.ts
```diff
@@ -0,0 +1,281 @@
+export const GRAMMAR = `{{
```
jacoblee93, line 1: Is this something we'll have to maintain ourselves? And are there any docs on this you could link to in a comment here?

langchain/src/output_parsers/expression.ts
```diff
@@ -22,20 +21,31 @@ import { ASTParser } from "./expression_type_handlers/base.js";
...
+  parser: ParseFunction;
+
+  /**
+   * We should separate loading the parser into its own function
+   * because loading the grammar takes some time. If there are
+   * multiple concurrent parse calls, it's faster to just wait
+   * for building the parser once and then use it for all
+   * subsequent calls. See expression.test.ts for an example.
+   */
+  async loadParser() {
```
jacoblee93, line 33: We've called methods like this `ensureParser()` and wrapped it in an `if (!this.parser)` statement.

langchain/src/output_parsers/expression_type_handlers/base.ts
```diff
@@ -1,104 +1,78 @@
...
+        const parser = peggy.generate(GRAMMAR);
+        const { parse } = parser;
+        ASTParser.astParseInstance = parse as ParseFunction;
       }
       return ASTParser.astParseInstance;
     } catch (e) {
-      console.log(e);
       throw new Error(
-        "Failed to import meriyah. Please install meriyah (i.e. npm install meriyah)."
+        `Failed to import peggy. Run 'npm install peggy' to install the package.`
```
jacoblee93, line 26: Keep style like the old comment, as people may be using `yarn`

ppramesi: Ok I'll change it.

langchain/src/output_parsers/expression_type_handlers/grammar/parser_grammar.ts
ppramesi: The grammar is a pared down version of Javascript grammar from [Peggy JS's example](https://github.com/peggyjs/peggy/blob/main/examples/javascript.pegjs). In terms of maintaining the grammar, yes you'd need to maintain it yourself, but imho since the expression grammar is rather simple, it shouldn't require too much maintenance. I'll add links to the docs for PeggyJS.

ppramesi:
> @ppramesi could you add an example test case that fails with the old parser and succeeds with the new one?

@jacoblee93 the old parser can't handle Javascript reserved keywords, especially `in`. So it'll fail if you feed it something like `in("test", [1, 2, 3])`. The last 2 expressions in the valid expressions array in expression.test.ts would fail with the old parser. (i.e. `and(in("a", ["a", "b", "c"]), eq("a", "b"))` and `with(const("a", ["a", "b", "c"]), eq("a", "b", true))`)

langchain/src/output_parsers/expression.ts
ppramesi: Noted. I'll rename it.

jacoblee93:
Very sorry about the delay - will merge with the next release. Thank you!

PR #1389, opened 2023-05-24
Update tracer to use new simpler api

nfcampos:
- Remove GET /tenants as the API will infer the tenant_id from the api key
- Remove session creation as POST /runs now accepts session name
- Add timeout (default to 5s), limit retries to 2
- Update Client in the same vein

### DIFF
--- a/langchain/src/callbacks/handlers/tracer_langchain.ts
+++ b/langchain/src/callbacks/handlers/tracer_langchain.ts
@@ -97,8 +97,10 @@ export class LangChainTracer
       body: JSON.stringify(persistedRun),
       signal: AbortSignal.timeout(this.timeout),
     });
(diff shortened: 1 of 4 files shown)


langchain/src/callbacks/handlers/tracer_langchain.ts
```diff
@@ -48,24 +30,19 @@ export class LangChainTracer
...
-  protected session: TracerSession;
+  sessionName?: string;
 
   exampleId?: string;
 
-  tenantId?: string;
-
   caller: AsyncCaller;
 
+  timeout = 5000;
```
vowelparrot, line 39: Is this default too long? I suppose it's not a problem in TS since it's async but 5s * 3 seems long

PR #1392, opened 2023-05-24
Make callbacks not be awaited by default, with option to enable waiting globally or per handler

nfcampos:


langchain/src/callbacks/promises.ts
```diff
@@ -0,0 +1,27 @@
...
+  wait: boolean
+): Promise<void> {
+  if (wait === true) {
+    await promiseFn();
+  } else {
+    void queue.add(promiseFn);
+  }
+}
+
+export function awaitAllCallbacks(): Promise<void> {
```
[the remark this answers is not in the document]
nfcampos: Yea I don't want to cll it flush because it doesn't really do anything, ie. it won't suddenly send out queued callbacks, it will just wait for them to finish

langchain/scripts/jest-setup-after-env.js
```diff
@@ -0,0 +1,3 @@
+import { awaitAllCallbacks } from "../src/callbacks/promises.js";
+
+afterAll(awaitAllCallbacks);
```
jacoblee93, line 3: Nice!

langchain/src/callbacks/base.ts
```diff
@@ -176,6 +176,12 @@ export abstract class BaseCallbackHandler
 
   ignoreAgent = false;
 
+  awaitHandlers =
+    typeof process !== "undefined"
+      ? // eslint-disable-next-line no-process-env
+        process.env?.LANGCHAIN_CALLBACKS_BACKGROUND !== "true"
```
jacoblee93, line 182: How about `LANGCHAIN_CALLBACK_MODE="async"` or `LANGCHAIN_CALLBACK_MODE="nonblocking"`?

langchain/src/callbacks/manager.ts
```diff
@@ -324,34 +359,36 @@ export class CallbackManager
...
+            try {
+              if (handler.handleChatModelStart)
+                await handler.handleChatModelStart?.(
+                  llm,
+                  messages,
+                  runId,
+                  this._parentRunId,
+                  extraParams
+                );
+              else if (handler.handleLLMStart) {
```
jacoblee93, line 374: Does the linter not enforce putting brackets here? I think it's a bit hard to read

langchain/src/callbacks/promises.ts
```diff
@@ -0,0 +1,34 @@
+import PQueueMod from "p-queue";
+
+let queue: typeof import("p-queue")["default"]["prototype"];
+
+function createQueue() {
+  const PQueue = "default" in PQueueMod ? PQueueMod.default : PQueueMod;
+  return new PQueue({
+    autoStart: true,
+    concurrency: 1,
```
jacoblee93, line 9: Concurrency 1 seems a little low?

langchain/src/callbacks/promises.ts
```diff
@@ -0,0 +1,34 @@
...
+ */
+export async function consumeCallback<T>(
+  promiseFn: () => Promise<T> | T | void,
+  wait: boolean
+): Promise<void> {
+  if (wait === true) {
+    await promiseFn();
+  } else {
+    if (typeof queue === "undefined") {
+      queue = createQueue();
```
jacoblee93, line 26: No strong feelings but we could just make this a global singleton?

jacoblee93 (commented):
Small suggestions on env var name etc. - functionality seems fine overall

langchain/src/callbacks/base.ts
nfcampos: @agola11 do you prefer that?

langchain/src/callbacks/promises.ts
nfcampos: It's to prevent handlers executing out of order

langchain/src/callbacks/promises.ts
nfcampos: It's to avoid the side effect the bundlers would complain about in that case

langchain/src/callbacks/manager.ts
nfcampos: I guess not, we need to review the linter config anyway, lots of silly rules in there

PR #1409, opened 2023-05-25
Add SingleStore vectorstore integration

volodymyr-memsql:
# SingleStore vector store

* Added SingleStoreVectorStore class that allows using SingleStoreDB as a vector store. 
* Added integration test
* Added documentation

## Future work 
* Enable filtering by metadata in `similaritySearch`


volodymyr-memsql:
Hello, @jacoblee93! I cannot assign a reviewer to the PR. What is the correct procedure to get this reviewed and merged?  Thanks!

jacoblee93:
Hey @volodymyr-memsql, looking now!

package.json
```diff
@@ -43,13 +43,15 @@
...
     "@tsconfig/recommended": "^1.0.2",
+    "@types/mysql": "^2",
     "dotenv": "^16.0.3",
     "husky": "^8.0.3",
     "lint-staged": "^13.1.1",
     "prettier": "^2.8.3",
     "typescript": "^5.0.0"
   },
   "dependencies": {
+    "mysql": "^2.18.1",
```
jacoblee93, line 54: Shouldn't be a hard dependency here. Please see this guide for the right way to add third-party deps:
https://github.com/hwchase17/langchainjs/blob/main/.github/contributing/INTEGRATIONS.md

langchain/src/vectorstores/singlestore.ts
```diff
@@ -0,0 +1,118 @@
...
+
+  async similaritySearchVectorWithScore(query: number[], k: number, _filter?: undefined): Promise<[Document, number][]> {
+    const contentColumn = this.contentColumnName;
+    const metadataColumn = this.metadataColumnName;
+    // use vector DOT_PRODUCT as a distance function 
+    const searchQuery = `SELECT ${this.contentColumnName}, ${this.metadataColumnName}, DOT_PRODUCT(${this.vectorColumnName}, JSON_ARRAY_PACK('[${query}]')) as __score  FROM ${this.tableName} ORDER BY __score DESC LIMIT ${k}`;
+    return new Promise((resolve, reject) => {
+      this.connectionPool.getConnection((err: MysqlError, connection: PoolConnection) => {
+        if (err) reject(err);
+        connection.query(searchQuery, (err: MysqlError, rows: any[]) => {
```
jacoblee93, line 83: Try to avoid `any`

langchain/src/vectorstores/singlestore.ts
```diff
@@ -0,0 +1,118 @@
...
+    });
+  }
+
+  async similaritySearchVectorWithScore(query: number[], k: number, _filter?: undefined): Promise<[Document, number][]> {
+    const contentColumn = this.contentColumnName;
+    const metadataColumn = this.metadataColumnName;
+    // use vector DOT_PRODUCT as a distance function 
+    const searchQuery = `SELECT ${this.contentColumnName}, ${this.metadataColumnName}, DOT_PRODUCT(${this.vectorColumnName}, JSON_ARRAY_PACK('[${query}]')) as __score  FROM ${this.tableName} ORDER BY __score DESC LIMIT ${k}`;
+    return new Promise((resolve, reject) => {
+      this.connectionPool.getConnection((err: MysqlError, connection: PoolConnection) => {
```
jacoblee93, line 81: Is there no `async` method on `connectionPool`? Prefer not to mix callbacks and promises.

If there isn't, is this the right package to use? Last publish was 3 years ago, seems abandoned.

jacoblee93 (changes requested):
Thank you, and apologies for the delay. Left a few comments.

jacoblee93:
Also, please remember to `yarn lint` and `yarn format`!

volodymyr-memsql:
Thanks for the feedback, @jacoblee93! I switched to the connection pool from `mysql2/promise` and got rid of callbacks. Also, addressed other comments. Could you please check it again?

langchain/src/vectorstores/singlestore.ts
```diff
@@ -0,0 +1,135 @@
...
+  async addVectors(vectors: number[][], documents: Document[]): Promise<void> {
+    await this.createTableIfNotExists();
+    const { tableName } = this;
+
+    await Promise.all(
+      vectors.map(async (vector, idx) => {
+        try {
+          await this.connectionPool.execute(
+            `INSERT INTO ${tableName} VALUES ('${
+              documents[idx].pageContent
```
jacoblee93, line 63: I don't think we can assume documents are friendly/sanitized - can you use a prepared statement here?

jacoblee93: It's just good practice regardless

langchain/src/vectorstores/index.ts
```diff
@@ -8,3 +8,4 @@ export { PineconeStore } from "./pinecone.js";
 export { VectorStore, SaveableVectorStore } from "./base.js";
 export { SupabaseVectorStore } from "./supabase.js";
 export { PrismaVectorStore } from "./prisma.js";
+export { SingleStoreVectorStore } from "./singlestore.js";
```
jacoblee93, line 11: Exporting from here is deprecated, please remove.

langchain/src/vectorstores/singlestore.ts
```diff
@@ -0,0 +1,135 @@
...
+        | RowDataPacket[][]
+        | OkPacket
+        | OkPacket[]
+        | ResultSetHeader
+      ),
+      FieldPacket[]
+    ] = await this.connectionPool.query(
+      `SELECT ${this.contentColumnName},
+    ${this.metadataColumnName},
+    DOT_PRODUCT(${this.vectorColumnName}, JSON_ARRAY_PACK('[${query}]')) as __score
```
jacoblee93, line 93: I feel like this should be ok but best practices is still to parameterize

jacoblee93 (changes requested):
A few more changes/questions - potential SQL injection

examples/src/indexes/vector_stores/singlestore.ts
```diff
@@ -0,0 +1,26 @@
...
+import { createPool } from "mysql2/promise";
+
+export const run = async () => {
+  const pool = createPool({
+    host: process.env.SINGLESTORE_HOST,
+    port: Number(process.env.SINGLESTORE_PORT),
+    user: process.env.SINGLESTORE_USERNAME,
+    password: process.env.SINGLESTORE_PASSWORD,
+    database: process.env.SINGLESTORE_DATABASE,
+  });
```
jacoblee93, line 12: I think generally I'd have a preference for making `mysql2` an implementation detail within your store - basically passing in `connectionParams` rather than passing in the entire `connectionPool`. Fewer imports for the developer, and allows you to swap packages more seamlessly if something should happen that dependency. 

But up to you! Looks good otherwise and will merge by tomorrow with or without that change.

jacoblee93 (approved):
Looks good, one small suggestion Will merge by tomorrow with or without that change.

examples/src/indexes/vector_stores/singlestore.ts
volodymyr-memsql: The `createPool` allows a developer to pass many different parameters (not only those four that I used in the example) depending on the type of connection. I'd prefer to keep this flexibility and not copy-paste all parameters as parameters of the vector store.

jacoblee93: They may export a type like `ConnectionPoolParams` with all options that you can then initialize a pool with. No worries though, will merge shortly.

PR #1413, opened 2023-05-25
Factor out and wrap environment variable access in a try/catch

### LINKED ISSUE #1412: Fail gracefully if Deno.env is not accessible
At [Val Town](https://val.town) we block access to Deno.env, which means that LangChain will fail to run. It would be nice if it could wrap calls to `Deno.env.get` in a try/catch and treat a thrown exception there as you would if the environment variable simply wasn't set.

image

Alternatively, we could directly folks to using langchain via npm which sometimes succeeds and sometimes fails.

Success:

image

Fail:

image

We haven't yet figured out the cause of this intermittent error - let us know if you have any clue to it.

Here's my langChain example val: https://www.val.town/v/stevekrouse.langchainEx. You can view past versions to see what I mean.

jacoblee93:
Some Deno configurations will throw an error if any attempt is made to access an environment variable. This PR refactors `process.env` access into a util method. Should also remove the need to have TS compiler ignore comments in many places as well.

Fixes #1412

### DIFF
--- a/langchain/src/callbacks/handlers/tracer_langchain.ts
+++ b/langchain/src/callbacks/handlers/tracer_langchain.ts
@@ -1,5 +1,8 @@
 import { AsyncCaller, AsyncCallerParams } from "../../util/async_caller.js";
-import { getRuntimeEnvironment } from "../../util/env.js";
+import {
+  getEnvironmentVariable,
+  getRuntimeEnvironment,
+} from "../../util/env.js";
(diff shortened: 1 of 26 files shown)


stevekrouse:
Thank you!!

WadeWarren:
@nfcampos This would be great to merge so we can use langchainjs with Deno / edge functions. :-)

jacoblee93:
Will go out in today's release!

PR #1414, opened 2023-05-25
Structured agent memory support

emilsedgh:
This provides Memory support for `StructuredChatAgent` similar to `ChatConversationalAgent`.

There's only one gotcha so far.

The prompt of this agent tries really hard to force the chat model to respond in the `{action, action_input}` format.
Most of the prompt is dedicated to that. However, when we add memory to it, the memory seems to pollute the prompt and after a couple of back and forth messages, the language model (chatgpt 3.5 in this case) seems to forget that it must respond in the json format.

This is, however, automatically recovered by the error correction mechanism built in to the agent, so it ends up working fine.

I wonder if:

1. The same behavior exists on Python version (It doesn't appear they do anything differently so in theory they should be the same but I haven't tried it in practice)
2. How the prompts can be improved to avoid this.


emilsedgh:
[This commit](https://github.com/hwchase17/langchainjs/pull/1414/commits/b1af087c0ad6fa9917d1303c7673b08513d5c181) essentially removes the gotcha and makes it very very consistently use the provided format.

jacoblee93:
Will leave this open for now as maybe it'll fit in well as a completely separate agent :)

PR #1416, opened 2023-05-26
feat: add Momento as a standard cache and chat history provider

malandis:
# Add Momento as a standard cache and chat message history provider

This PR adds Momento as a standard caching provider. Implements the interface, adds integration tests, and documentation. We also add Momento as a chat history message provider along with integration tests, and documentation.

[Momento](https://www.gomomento.com/) is a fully serverless cache. Similar to S3 or DynamoDB, it requires zero configuration, infrastructure management, and is instantly available. Users sign up for free and get 50GB of data in/out for free every month.

# Before submitting
✅ We have added documentation, notebooks, and integration tests demonstrating usage. We verified the docs build and render correctly. We also ran the examples locally.


langchain/src/cache/momento.ts
```diff
@@ -0,0 +1,166 @@
...
+   * @param {ICacheClient} props.client The Momento cache client.
+   * @param {string} props.cacheName The name of the cache to use to store the data.
+   * @param {number} props.ttlSeconds The time to live for the cache items. If not specified,
+   * the cache client default is used.
+   * @param {boolean} props.ensureCacheExists If true, ensure that the cache exists before returning.
+   * If false, the cache is not checked for existence. Defaults to true.
+   * @throws {@link InvalidArgumentError} if {@link props.ttlSeconds} is not strictly positive.
+   * @returns The Momento-backed cache.
+   */
+  public static async CreateAsync(
```
jacoblee93, line 71: We've standardized `fromX` for factory methods rather than capitals (e.g. `fromProps()` or `fromOptions()` - would you mind renaming?

langchain/src/stores/message/momento.ts
```diff
@@ -0,0 +1,180 @@
...
+    } else {
+      throw new Error(`Unknown response type: ${deleteResponse.toString()}`);
+    }
+  }
+}
+
+/**
+ * Ensure that the cache exists.
+ */
+async function ensureCacheExists(
```
jacoblee93, line 165: Should this not be on the class?

langchain/src/cache/momento.ts
malandis: Refactored in https://github.com/hwchase17/langchainjs/pull/1416/commits/de103e65f50d8d4a563feba4564d56c90942bf68

langchain/src/stores/message/momento.ts
malandis: Good catch. Since this code is common to both the cache and chat history modules, I intended to extract to a common utility module. I've made this change in https://github.com/hwchase17/langchainjs/pull/1416/commits/b0750e9c1ce3171225d89b5eb8d5a0ccdae620cb.

jacoblee93:
Thank you! Will merge this with the next release tonight or tomorrow.

jacoblee93 (approved):

PR #1418, opened 2023-05-26
New feat: Added advanced filter builder function for Supabase vector store

ppramesi:
This pull request introduces an advanced filter query builder into Supabase vector store, which allows users to apply complex query operators as defined in [Postgrest documentation](https://postgrest.org/en/stable/references/api/tables_views.html?highlight=operators#operators). An example of building a filter can be found in the Supabase integration test file.

The current implementation only allows exact matching of metadata properties, which can be limiting for users needing more intricate queries. The added feature will enable a greater variety of filters, improving usability and versatility, and aligning Supabase more closely with other vector databases. You can even do a full text search with the new query builder.

I also added more tests in the Supabase integration test to make sure it works correctly and making sure the new implementation preserves backwards compatibility by maintaining the ability to use an object as the filter parameter, as was done prior to this pull request. This means existing users won't have to modify their current code and it won't introduce any breaking changes.

However, it should be noted that changes were made to the Postgres function definition on the Supabase vector store and retriever documentation. Specifically, a null default was included for the match_count parameter. This change was necessary to allow limit query to be used after the rpc is called and ensure proper functioning of the new query builder function.

This feature also serves as the groundwork for future development of Supabase self-query retriever.

Before finalizing this pull request, I would appreciate inputs and feedback from the community. This is particularly important because the changes to the Postgres function could potentially affect users who have been using this function previously but would like to use the filter builder.


langchain/src/vectorstores/supabase.ts
```diff
@@ -84,17 +88,32 @@ export class SupabaseVectorStore extends VectorStore {
...
 
-    const { data: searches, error } = await this.client.rpc(
-      this.queryName,
-      matchDocumentsParams
-    );
+    let filterFunction: SupabaseFilterRPCCall;
+
+    if (typeof _filter === "function") {
+      filterFunction = (rpcCall) => {
+        if (!_filter) {
```
jacoblee93, line 100: Is it possible to hit this case? Aren't we checking `typeof _filter` above

docs/docs/modules/indexes/retrievers/supabase-hybrid.mdx
```diff
@@ -29,7 +29,7 @@ create table documents (
 -- Create a function to similarity search for documents
 create function match_documents (
   query_embedding vector(1536),
-  match_count int,
+  match_count int DEFAULT null,
```
jacoblee93, line 32: Is it possible to avoid this? 

We need it for advanced filtering because it basically filters down retrieved documents after they're initially fetched from Supabase? Is this really different then from setting a really high `k` and just using a `.filter()` method in JS?

langchain/src/vectorstores/supabase.ts
ppramesi: No yeah it really shouldn't be there. I'll change it.

docs/docs/modules/indexes/retrievers/supabase-hybrid.mdx
ppramesi: The problem I've had was that without that default value, the similaritySearch function doesn't return the correct number of results since the procedure already limits the results before it even gets to the filter. But yeah it is possible to just set the match_count really high and then add limit in the filter itself. I suppose if we include that in the example, we don't have to add the default value.

jacoblee93:
Small update to tests to make them re-runnable, this is very cool though, thank you! Will be part of the next release.

PR #1420, opened 2023-05-26
fix: #1383 update faiss-node to avoid libomp missing

### LINKED ISSUE #1383: Faiss-node has a hidden dependency on `libomp`, causes unit tests to fail on some machines
Introduced in https://github.com/hwchase17/langchainjs/pull/685, will change unit tests to integration tests to signify the need for more setup. Ideally this would be bundled in or the dependency would somehow otherwise be represented in the future.

ewfian:
# Update faiss-node to avoid `libomp` missing

- [x] Update faiss-node.
- [x] Revert `faiss_simple.int.test.ts` to a unit test.
- [x] Remove  `libomp` installation info from docs.

Fixes #1383


jacoblee93:
Thank you!

PR #1425, opened 2023-05-26
Remove outdated remark in docs/modules/models/chat/index.mdx

DanielPopOut:
# Remove outdated remark in docs/modules/models/chat/index.mdx

According to the docs there is more than ChatOpenAi model only --> ChatAnthropic etc...

What was removed:

> Note: Currently, the only chat-based model we support is ChatOpenAI (with gpt-4 and gpt-3.5-turbo), but anticipate adding more in the future.


jacoblee93:
Thank you!

PR #1432, opened 2023-05-28
feature: added support for POST to ApiChain

vjsai:
# added support for POST to ApiChain

- Changed Prompt to support POST call as well
- Made code changes to support POST via fetch


langchain/src/chains/tests/api_chain.int.test.ts
```diff
@@ -65,6 +82,24 @@ test("Test APIChain fromLLMAndApiDocs", async () => {
...
+  const model = new OpenAI({ modelName: "text-davinci-003" });
+  const chain = APIChain.fromLLMAndAPIDocs(model, POST_API_DOCS);
+  const res = await chain.call({
+    question: "send a message hi langchain to channel3 with token5",
+  });
+  console.log({ res });
+});
+
 test("Load APIChain from hub", async () => {
   const chain = await loadChain("lc://chains/api/meteo/chain.json");
```
vjsai, line 104: Will fix this chain If this PR changes are approved before merging.

langchain/src/chains/api/prompts.ts
```diff
@@ -3,11 +3,23 @@ import { PromptTemplate } from "../../prompts/prompt.js";
 
 export const API_URL_RAW_PROMPT_TEMPLATE = `You are given the below API Documentation:
 {api_docs}
-Using this documentation, generate the full API url to call for answering the user question.
-You should build the API url in order to get a response that is as short as possible, while still getting the necessary information to answer the question. Pay attention to deliberately exclude any unnecessary pieces of data in the API call.
 
-Question:{question}
-API url:`;
+<< FORMATTING >>
+Return a JSON string with a JSON object formatted to look like:
```
jacoblee93, line 8: Can we use an output fixing parser here?

https://js.langchain.com/docs/modules/prompts/output_parsers/#output-fixing-parser

langchain/src/chains/api/api_chain.ts
```diff
@@ -64,12 +64,26 @@ export class APIChain extends BaseChain implements APIChainInput {
...
+    const api_json = await this.apiRequestChain.predict(
       { question, api_docs: this.apiDocs },
       runManager?.getChild()
     );
 
-    const res = await fetch(api_url, { headers: this.headers });
+    const { api_url, api_body, api_method } = JSON.parse(api_json);
+
+    const request_options =
+      api_method === "GET" || api_body === "HEAD"
```
jacoblee93, line 75: `api_body === "HEAD"` looks like a typo?

The `DELETE` method also does not support a body.

jacoblee93 (changes requested):
Love the idea, a few comments though!

langchain/src/chains/api/api_chain.ts
vjsai: Yes will fix this good catch!

langchain/src/chains/api/prompts.ts
vjsai: Yes will make use of it @jacoblee93

vjsai:
@jacoblee93 Can you review and let me know if there are any other inputs. Thanks!

langchain/src/chains/api/api_chain.ts
```diff
@@ -64,16 +70,46 @@ export class APIChain extends BaseChain implements APIChainInput {
...
-    const api_url = await this.apiRequestChain.predict(
+    const api_json = await this.apiRequestChain.predict(
       { question, api_docs: this.apiDocs },
       runManager?.getChild()
     );
 
-    const res = await fetch(api_url, { headers: this.headers });
+    let api_options;
+    try {
+      api_options = await APIChain.getApiParser().parse(api_json);
```
jacoblee93, line 80: `getAPIParser` can actually just return the output fixing parser directly:

```
      const fixParser = OutputFixingParser.fromLLM(
        this.llm,
        APIChain.getApiParser()
      );
```
The above will automatically use the internal output parser first, then retry with a more focused prompt.

langchain/src/chains/api/prompts.ts
jacoblee93: Can you try replacing this formatting instructions with the built-in ones for your output parser?

https://js.langchain.com/docs/modules/prompts/output_parsers/#structured-output-parser-with-zod-schema

```
  z.object({
    answer: z.string().describe("answer to the user's question"),
    sources: z
      .array(z.string())
      .describe("sources used to answer the question, should be websites."),
  })
);

const formatInstructions = parser.getFormatInstructions();

const prompt = new PromptTemplate({
  template:
    "Answer the users question as best as possible.\n{format_instructions}\n{question}",
  inputVariables: ["question"],
  partialVariables: { format_instructions: formatInstructions },
});
```

langchain/src/chains/api/api_chain.ts
```diff
@@ -64,16 +70,46 @@ export class APIChain extends BaseChain implements APIChainInput {
...
+    } catch (e) {
+      const fixParser = OutputFixingParser.fromLLM(
+        this.llm,
+        APIChain.getApiParser()
+      );
+      api_options = await fixParser.parse(api_json);
+    }
+
+    const request_options =
```
jacoblee93, line 89: I think it would be nice to actually have this configurable - basically pass in a list of `allowedMethods` that defaults to `GET` and `POST`. 

Just get a little bit nervous now that I see DELETE here in code.

jacoblee93:
A few more comments!

langchain/src/chains/api/api_chain.ts
vjsai: Oh should have seen the code. Documentation made me a bit confused. Will fix the documentation as well in another PR.

langchain/src/chains/api/api_chain.ts
vjsai: Sure will do that. Makes sense @jacoblee93 !

langchain/src/chains/api/api_chain.ts
jacoblee93: You’re right, definitely could be clearer. Thank you!

langchain/src/chains/tests/api_chain.int.test.ts
vjsai: Have raised following PR : https://github.com/hwchase17/langchain-hub/pull/47

vjsai:
@jacoblee93 addressed the comments and also raised PR for LC hub as well. Let me know if anything more needs to be added. Thanks!

vjsai:
@jacoblee93 Have addressed review comments let me know if I am missing something. Thanks !

jacoblee93:
Hey @vjsai, sorry for the delay. I actually think we should think a bit more about the effect on existing folks using this before making a big prompt change - I'd be more open to adding this as an extension (use a different prompt if the user wants to enable the chain to use different methods) if you're open to that?

jacoblee93:
Now that #1803 is merged I think this is superseded - sorry for the confusion.

PR #1435, opened 2023-05-28
Ability to load DSV files 

EhsanHadid:
# New Loader for DSV like pipe delimited files

Created A new loader for loading custom delimited files


jacoblee93:
Could this extend the CSVLoader instead of being a separate class? I've seen that pretty frequently.

jacoblee93:
Functionality was added in #1968.

PR #1441, opened 2023-05-29
Support string arrays in Pinecone metadata

yujinakayama:
# Support string arrays in Pinecone metadata

Since #415, PineconeStore automatically flattens metadata (all nested objects and arrays) when adding documents.
However Pinecone officially supports list of strings in metadata, so this PR will add support for string arrays in Pinecone metadata by not flattening them.

https://docs.pinecone.io/docs/metadata-filtering#supported-metadata-types


jacoblee93:
Hey @yujinakayama, thanks for this, and sorry for the delay! Going to go with #1518 which fixes the same issue without changing any tests but will give you a shoutout as well.

PR #1447, opened 2023-05-29
Initial implementation for Structured Conversational agent

emilsedgh:
Implements Structured Conversational Agent.

This is a continuation of the conversation from GH-1414. @jacoblee93 prefers this to be a new agent so the original StructuredAgent could be kept closer to the python version.

Note that it has a slight modification to the prompt to ensure better performance with the memory, as introducing the memory caused it to very frequently forget instructions.

It also uses a Human prompt rather than a System prompt because according to [OpenAI docs](https://platform.openai.com/docs/guides/chat/introduction) `gpt-3.5-turbo-0301` does not always pay strong attention to  system messages. And since that's a very popular chat model, it'd hinder the performance of the agent.


emilsedgh:
I'm closing this MR in favor of the new OpenAI Functions Agent.

emilsedgh:
Also Relevant: ~1638

PR #1458, opened 2023-05-30
New serialisation format

nfcampos:


langchain/src/load/tests/load.test.ts
```diff
@@ -0,0 +1,580 @@
...
+            "type": "constructor",
+            "identifier": [
+              "langchain",
+              "llms",
+              "openai",
+              "OpenAI"
+            ],
+            "arguments": [
+              {
+                "callbacks": [
```
hwchase17, line 119: do we want to serialize callbacks? my default thinking would be no...

langchain/src/chains/base.ts
```diff
@@ -26,6 +26,14 @@ export interface ChainInputs extends BaseLangChainParams {
 export abstract class BaseChain extends BaseLangChain implements ChainInputs {
   declare memory?: BaseMemory;
 
+  get lc_namespace(): ["langchain", "chains"] {
```
nfcampos, line 29: Fix this

langchain/scripts/identify-secrets.js
```diff
@@ -0,0 +1,77 @@
...
+      fs.readFileSync(p, "utf-8")
+    ),
+    ts.sys,
+    "./src/"
+  );
+
+  for (const fileName of tsConfig.fileNames.filter(
+    (fn) => !fn.endsWith("test.ts")
+  )) {
+    const sourceFile = ts.createSourceFile(
```
jacoblee93, line 18: Dang this is gnarly - purpose is to look for required env variables in source files?

langchain/src/schema/load.ts
```diff
@@ -0,0 +1,105 @@
...
+        type: "secret",
+        id: [secretId],
+      };
+    }
+  }
+  return result;
+}
+
+export abstract class Serializable {
+  abstract lc_namespace: string[];
```
jacoblee93, line 60: No way around this eh?

Could we make it a little bit friendlier by making it recursive? e.g. `LangChainSerializationName = "structured"` and computing the full namespace by going up through the superclasses?

Does get a bit hairy if we decide to change the what a class subclasses though...

jacoblee93: I think generally I'd name these so that it's a bit clearer they're for serialization - a little easier for contributors to grok.

nfcampos: We could make it recursive (you'll notice the lc_secrets one works like that) but the way we've organised things (eg. base classes in /schema it actually wouldn't do the right thing.

langchain/src/schema/load.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  lc_arguments: unknown[];
+
+  get lc_secrets(): { [key: string]: string } | undefined {
```
jacoblee93, line 64: What's the purpose behind the `0.SECRET_NAME`?

jacoblee93 (commented):
Left some feedback/questions!

langchain/src/schema/load.ts
nfcampos: It's to prevent those values from being included in the serialised object, eg api keys

langchain/scripts/identify-secrets.js
nfcampos: To look for the secret ids defined in Serializable classes yes

langchain/src/schema/load.ts
nfcampos: Because it's in the first argument to the constructor, thus 0. If you hunt around all the files you'll see some constructors (sadly) take more than one arg

langchain/scripts/identify-secrets.js
jacoblee93: Might be good to refactor into different methods for clarity

PR #1465, opened 2023-05-30
Zep::getMemory should return [] instead of throwing an error

rsharath:
# Fixes : zepMemory should return [] when no records were found (instead of raising error)


jacoblee93:
Thanks! We're doing another release tomorrow so it'll go out with that.

PR #1478, opened 2023-05-31
add managed motorhead

softboyjimbo:
# add managed motorhead

### DIFF
--- a/langchain/src/memory/motorhead_memory.ts
+++ b/langchain/src/memory/motorhead_memory.ts
@@ -13,12 +13,6 @@ export interface MotorheadMemoryMessage {
   content: string;
 }
 
-export interface MotorheadMemoryHeaders {
-  "Content-Type": string;
-  "x-metal-api-key"?: string;
-  "x-metal-client-id"?: string;
-}
-
 /**
  * @interface
  */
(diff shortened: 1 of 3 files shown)


langchain/src/memory/motorhead_memory.ts
```diff
@@ -19,13 +19,18 @@ export interface MotorheadMemoryMessage {
 export type MotorheadMemoryInput = BaseChatMemoryInput &
   AsyncCallerParams & {
     sessionId: string;
-    motorheadURL?: string;
+    url?: string;
```
jacoblee93, line 22: This will break backwards compatibility - you ok if I add a shim?

langchain/src/memory/motorhead_memory.ts
```diff
@@ -19,13 +19,18 @@ export interface MotorheadMemoryMessage {
...
+    apiKey?: string;
+    clientId?: string;
   };
 
+const MANAGED_URL = "https://api.getmetal.io/v1/motorhead";
+// const LOCAL_URL = "localhost:8080"
+
 export class MotorheadMemory extends BaseChatMemory {
-  motorheadURL = "localhost:8080";
+  url = MANAGED_URL;
```
jacoblee93, line 33: I think this is fair because anyone using this in production probably isn't running it on `localhost`?

jacoblee93:
Merged with extension #1561

PR #1484, opened 2023-05-31
fix: Tracing python doc link 404 not found redirection problem.

gkhngyk:
Document Page: https://js.langchain.com/docs/production/tracing

Replace the https://python.langchain.com/en/latest/tracing.html link in the document with the https://python.langchain.com/en/latest/additional_resources/tracing.html link.


jacoblee93:
Thanks for this! Can you please actually just change it to the `python.langchain.com` site directly? Might as well save people the hop.

jacoblee93:
I've squashed it, shouldn't be a problem. Thank you!

PR #1489, opened 2023-06-01
ported over the generative agent abstractions from python (and docs)

dylanintech:
# LangChain.js is closer to being at parity with the python implementation with the addition of the GenerativeAgent and GenerativeAgentMemory abstractions 🦜

I translated these abstractions over from the python implementation and also wrote a script for using these agents and added it in the `examples/src/experimental/generative_agents` folder. I used this example in a new section on agent simulations in the docs that I added.

This took me a couple of days to do haha. I tweeted about this on my twitter for context: [@dxlantxch](https://twitter.com/dxlantxch)

*Also, if there are any additional requests for these abstractions, please don't hesitate to ping me in the comments below :D*


dylanintech:
What should be edited so that the vercel deployment doesn't fail?

jacoblee93:
> What should be edited so that the vercel deployment doesn't fail?

Nothing, it just needs approval to run. Excited to look at this tomorrow!

jacoblee93:
Hey @dylanintech, awesome stuff! 

Did some refactoring around method signatures (there's no named args in JS like in Python, so it's nice to use a config object for readability) and there were some issues with memory not getting passed in properly, but really excited to get this merged.

PR #1490, opened 2023-06-01
Add unit tests to chromadb integration

### LINKED ISSUE #96: Additions for chromadb integration
There are few things that can be improved in the integration:
- [ ] Add tests - this will require mocking the chroma backend responses to have any value and to be a part of regular CI. 
- [x] Add ability for documents to be stored in Chroma. Chroma can operate purely as a metadata-index store, but some users store documents in Chroma as well and and provide some hybrid search affordances. For example search where document `$contains` a string. https://github.com/chroma-core/chroma/blob/main/examples/where_filtering.ipynb

cdiddy77:
# Add unit tests for chromadb integration

Per request, mocked out the chroma backend, and wrote unit tests that validate basic functionality of the Chroma VectorStore class

image

Fixes #96  (issue)

Twitter handle: https://twitter.com/cparker7

### DIFF
--- a/langchain/src/vectorstores/tests/chroma.test.ts
+++ b/langchain/src/vectorstores/tests/chroma.test.ts
@@ -3,6 +3,7 @@ import { jest, test, expect } from "@jest/globals";
 
 import { type Collection } from "chromadb";
 import { Chroma } from "../chroma.js";
+import { FakeEmbeddings } from "../../embeddings/fake.js";
 
 const mockCollection = {
   add: jest.fn<Collection["add"]>().mockResolvedValue(undefined as any),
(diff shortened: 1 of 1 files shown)


cdiddy77:
Happy to write a few more, just want to validate that I'm on the right track. First time contributing.

langchain/src/vectorstores/tests/chroma.test.ts
```diff
@@ -1,11 +1,112 @@
...
+  // peek: jest.fn<Collection["peek"]>().mockResolvedValue(undefined as any),
+  // delete: jest.fn<Collection["delete"]>().mockResolvedValue(undefined as any),
+} as any;
 
-  expect(ChromaClient).toBeDefined();
+const mockClient = {
+  getOrCreateCollection: jest.fn<any>().mockResolvedValue(mockCollection),
+} as any;
+
+const createMockEmbeddings = (): any => ({
```
jacoblee93, line 24: We have a `FakeEmbeddings` class you can use!

https://github.com/hwchase17/langchainjs/blob/main/langchain/src/embeddings/fake.ts

cdiddy77: I saw that. The problem is that, since it is not a jest thing, I can't do verification with it. I guess I could use jest.spyOn. Would that make more sense?

jacoblee93: Yeah if you can do it that would be best - otherwise let me know and we can merge as is.

PR #1500, opened 2023-06-01
Add new filetypes supported by Unstructured

qued:
# Expands list of file types supported by Unstructured to reflect updated to Unstructured library/API

Update the list of file extensions supported by Unstructured to add the following:
* `.htm` (`.html` was already supported, no reason this shouldn't be included)
* `.rtf` (Rich Text Format)
* `.xlsx`(MS Excel files since 2007)
* `.xls` (older MS Excel files)
* `.odt` (OpenDocument files)
* `.epub` (open e-book format)


jacoblee93:
Thank you!

PR #1503, opened 2023-06-01
improving MongoDBAtlasVectorSearch vectorstore and normalizing its name with the python package

### LINKED ISSUE #1502: MongoVector is not the correct name and lacks the possibility to add custom field names for content and embedding
Currently "content" and "embedding" are forced on the user.
This should not be the case.
Also "metadata" field contains the metadata. MongoDB users may have already existing collections with vector embeddings as a field and other fields at the root of the document. This should not be enforced by this library (contrary to what the python langchain library does).
Also the name of the vectorstore is misleading and not correct. We want to use the same name as the one in the python library.

P-E-B:
improving MongoDBAtlasVectorSearch vectorstore and normalizing its name with the python package

Fixes #1502

@nfcampos


P-E-B:
@nfcampos tagging you here since you validated the first version of this vectorstore :)

archie-swif (approved):
LGTM!

jacoblee93:
I don't mind the name change but this would be very breaking - can we try aliasing it instead?

P-E-B:
> I don't mind the name change but this would be very breaking - can we try aliasing it + deprecating instead?

We (MongoDB) want to change the name that does not reflect the product and is actually misleading. What do you suggest?

jacoblee93:
> > I don't mind the name change but this would be very breaking - can we try aliasing it + deprecating instead?
> 
> We (MongoDB) want to change the name that does not reflect the product and is actually misleading. What do you suggest?

I suggest what I wrote above. Basically making another entrypoint with the new name so developers can import it while keeping the old one with a deprecation notice.

LMK if you still don't understand what that means and I can just do it and see if you like it.

P-E-B:
> > > I don't mind the name change but this would be very breaking - can we try aliasing it + deprecating instead?
> > 
> > 
> > We (MongoDB) want to change the name that does not reflect the product and is actually misleading. What do you suggest?
> 
> I suggest what I wrote above. Basically making another entrypoint with the new name so developers can import it while keeping the old one with a deprecation notice.
> 
> LMK if you still don't understand what that means and I can just draft something for your review.

Done!

jacoblee93:
Looks great, thanks for making that update!

PR #1504, opened 2023-06-01
change tracer to use post/patch

agola11:
# Your PR Title Here (What it does)

Fixes # (issue)


langchain/src/callbacks/handlers/tracer.ts
```diff
@@ -101,7 +101,7 @@ export abstract class BaseTracer extends BaseCallbackHandler {
       name: llm.name,
       parent_run_id: parentRunId,
       start_time: Date.now(),
-      end_time: 0,
+      end_time: Date.now(),
```
agola11, line 104: This was done to maintain same behavior as python (otherwise, an in-progress run would show up with a negative latency)

langchain/src/callbacks/handlers/tracer_langchain.ts
```diff
@@ -106,4 +118,65 @@ export class LangChainTracer
...
+    const body = await response.text();
+    if (!response.ok) {
+      throw new Error(
+        `Failed to update run: ${response.status} ${response.statusText} ${body}`
+      );
+    }
+  }
+
+  async onLLMStart(run: Run): Promise<void> {
+    await this._persistRunSingle(run);
```
agola11, line 148: lmk if I shouldn't be awaiting these @nfcampos

langchain/src/callbacks/handlers/tracer.ts
nfcampos: We shouldn't do this, we should just change the schema to make end_time nullable. We shouldn't save wrong information to the db

nfcampos:
@agola11 I've pushed a commit here removing the fake end_time values

nfcampos (approved):

PR #1505, opened 2023-06-02
AI21 Labs Integration

paaatrrrick:
# A new file for integrating AI21 Labs API with unit tests.

I followed the python ai21 labs file as closely as possible to keep the similarity between the two codebases.

Please be thorough in your feedback

Twitter: @PatrickRFoster


jacoblee93:
Thank you! Will look as soon as I can.

jacoblee93:
Added some polish, tests, and docs - thank you!

langchain/src/llms/ai21.ts
```diff
@@ -0,0 +1,183 @@
...
+        : "https://api.ai21.com/studio/v1";
+
+    const url = `${baseUrl}/${this.model}/complete`;
+    const headers = {
+      Authorization: `Bearer ${this.ai21ApiKey}`,
+      "Content-Type": "application/json",
+    };
+    const data = { prompt, stopSequences: stop, ...this.defaultParams };
+    const responseData = await this.caller.callWithOptions(
+      { signal: options.signal },
```
nfcampos, line 153: Given you're passing signal to fetch you don't need to pass it here too

PR #1507, opened 2023-06-02
Add human as a tool

andrewlei:
# Add human as a tool

Used for asking the human for clarifications when the AI agent is confused
Inspired by python library human tool: https://github.com/langchain-ai/langchain/blob/master/libs/langchain/langchain/tools/human/tool.py


jacoblee93:
Can we implement this similarly to Python with a `promptFunc` and an `inputFunc`? It would ideally have some example of how to implement a good `promptFunc` for the command line:

https://github.com/hwchase17/langchain/blob/master/langchain/tools/human/tool.py

andrewlei:
> Can we implement this similarly to Python with a `promptFunc` and an `inputFunc`? It would ideally have some docs example of how to implement a good `promptFunc` for the command line:
> 
> https://github.com/hwchase17/langchain/blob/master/langchain/tools/human/tool.py

yup let me make those updates

PR #1508, opened 2023-06-02
fix: pass all prior inputs in sequential chains

miladh:
# fix(SequentialChain): pass all prior inputs

Follows the Python implementation.
Also see https://github.com/hwchase17/langchain/issues/4116#issuecomment-1546798238


jacoblee93:
Makes sense, and thank you for spotting this!

Can you add a test for this as well?

miladh:
> Makes sense, and thank you for spotting this!
> 
> Can you add a test for this as well?

I see you have already added a test for this. Thanks!

jacoblee93:
Yes, will merge tomorrow!

PR #1509, opened 2023-06-02
Enhance flexibility and type safety in NotionDBLoader

yuku:
This pull request introduces significant improvements to the `NotionDBLoader` class, enhancing its flexibility and type safety. It contains two key commits:

1. e148d23615f2da3b89ad0e83cd0308da6f4ba6e5 In this commit, we've refactored our `NotionDBLoader` to use the official types provided by [the Notion SDK](https://github.com/makenotion/notion-sdk-js) rather than implementing our own. This change brings about improved type safety and consistency within our codebase.
2. 896f625d72386627c30667d1f49594ec33590bcd The second commit introduces a new `createMetadata` method in `NotionDBLoader` which allows developers to customize metadata creation. This feature improves the flexibility of `NotionDBLoader`, enabling it to handle diverse metadata structures. The `createMetadata` also utilizes the type imported from the official SDK for better developer experience.

### Movitation

The motivation behind this pull request is two-fold:

* **Better Handling of Notion Page Properties**: Our current implementation doesn't fully exploit the information available in Notion page properties. These properties have a complex structure and can contain a wealth of data. However, the ideal structure for metadata depends on the vector store used - in our case, Pinecone. 
  Given this situation, we believe it is best to enable developers to implement their own parsers, tailored to their specific needs. Hence, the new `createMetadata` method has been introduced. 
* **Solution to Property Name Collision**: The current implementation forcibly creates `"id"` metadata and potentially overwrites properties named `"id"`. To mitigate this, a prefix, such as `"property_"`, could be added to metadata entries sourced from properties, effectively preventing name collisions but it is backward incompatible.

### Additional Information

* This PR adds the official notion sdk [`@notionhq/client`](https://github.com/makenotion/notion-sdk-js) to devDependencies and peerDependencies but not to dependencies, so it will not have any impact on the bundle size.


yuku:
I decided to write my own document loader from scratch. I'm going to close this PR.

jacoblee93:
This could definitely be improved if you're down to make another PR with your latest one!

PR #1518, opened 2023-06-02
Add pinecone metadata string array support

### LINKED ISSUE #1446: addVectors does not support array value in metadata
In the code below, I am adding ownerIds as array in the metadata and then embedding to Pinecone. Pinecone supports array value in the metadata but in Pinecone, it was saved as "ownerIds.0". 
```
const documents = await splitter.splitDocuments([
            new Document({
                pageContent: content,
                metadata: {
                    entityName,
                    entityType,
                    recordId,
                    ownerIds: [recordId],
                    organizationId,
                    solutionId,
                    userId
                }
            })
        ]);
```

I believe it was caused by the code in addVectors, it seems does not support array value.
Could you please improve it?

Thanks
Gary

jaclar:
# Add pinecone metadata string array support

Pinecone documents support string arrays, so they shouldn't be flattened.
This fix basically removes all arrays from the metadata object which contain only strings and appends them after the remaining metadata object was flattened.

Fixes #1446

### DIFF
--- a/langchain/src/vectorstores/pinecone.ts
+++ b/langchain/src/vectorstores/pinecone.ts
@@ -57,12 +57,24 @@ export class PineconeStore extends VectorStore {
     const documentIds = ids == null ? documents.map(() => uuid.v4()) : ids;
     const pineconeVectors = vectors.map((values, idx) => {
       // Pinecone doesn't support nested objects, so we flatten them
+      const documentMetadata = {...documents[idx].metadata};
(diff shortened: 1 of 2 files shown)


jacoblee93:
Fantastic PR, thank you!

Joshua-Shepherd:
Thanks a ton! I was actually working on this in my own personal project 🙏

jacoblee93:
Not announced but 0.0.90 is live with it :)

PR #1526, opened 2023-06-03
Adds pipeline prompt

jacoblee93:
Adds a new type of prompt that can use intermediate prompts in the final output value.

### DIFF
--- a/langchain/src/prompts/pipeline.ts
+++ b/langchain/src/prompts/pipeline.ts
@@ -3,30 +3,31 @@ import { BasePromptTemplate, BasePromptTemplateInput } from "./base.js";
 import { ChatPromptTemplate } from "./chat.js";
 import { SerializedBasePromptTemplate } from "./serde.js";
 
-export type PipelinePromptParams<T extends BasePromptTemplate> = {
+export type PipelinePromptParams<
+  PromptTemplateType extends BasePromptTemplate
+> = {
   name: string;
-  prompt: T;
+  prompt: PromptTemplateType;
 };
 
-export type PipelinePromptTemplateInput<T extends BasePromptTemplate> = Omit<
-  BasePromptTemplateInput,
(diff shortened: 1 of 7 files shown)


hwchase17 (commented):
conceptually lgtm, would be helpful to have example maybe?

jacoblee93:
> conceptually lgtm, would be helpful to have example maybe?

Yes, will add docs before merging.

langchain/src/prompts/pipeline.ts
```diff
@@ -0,0 +1,117 @@
+import { InputValues, PartialValues } from "../schema/index.js";
+import { BasePromptTemplate, BasePromptTemplateInput } from "./base.js";
+import { ChatPromptTemplate } from "./chat.js";
+import { SerializedBasePromptTemplate } from "./serde.js";
+
+export type PipelinePromptParams<T extends BasePromptTemplate> = {
```
jacoblee93, line 6: TODO: Change `T` to a more descriptive name

jacoblee93:
> > conceptually lgtm, would be helpful to have example maybe?
> 
> Yes, will add docs before merging.

Done!

langchain/src/prompts/pipeline.ts
```diff
@@ -0,0 +1,120 @@
...
+      .map((pipelinePrompt) =>
+        pipelinePrompt.prompt.inputVariables.filter(
+          (inputValue) => !intermediateValues.includes(inputValue)
+        )
+      )
+      .flat();
+    return [...new Set(inputValues)];
+  }
+
+  protected static extractRequiredInputValues(
```
nfcampos, line 48: Why protected on a static field?

jacoblee93: Just as documentation - that it's not intended to be part of the public interface

PR #1528, opened 2023-06-03
Adds inferred generic type parameters for prompts

jacoblee93:
This sets the groundwork for autocomplete in chain inputs, as well as optional chain values etc.


langchain/src/prompts/base.ts
```diff
@@ -28,59 +31,77 @@ export class StringPromptValue extends BasePromptValue {
...
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  InputVariableName extends string = any,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  PartialVariableName extends string = any
+> {
   /**
    * A list of variable names the prompt template expects
    */
-  inputVariables: string[];
+  inputVariables: InputVariableName[];
```
nfcampos, line 43: Should this an actually be a tuple type? But might be hard to make that generic

jacoblee93: It's particularly nice to do it this way here: https://github.com/hwchase17/langchainjs/pull/1528/files#diff-6b62f3e2fdc338c91c2930a51218d6cc23baa2103b6e34a40e03adf0018a2390L8

I suppose this does technically allow multiple of the same key in the array - is that what you're worried about?

jacoblee93: Or are you thinking ahead to optional variables?

nfcampos:
@jacoblee93 I'll review this tomorrow

nfcampos (approved):

PR #1530, opened 2023-06-04
New Feats: Weaviate, Supabase, HNSWLib and Memory self-query retrievers.

ppramesi:
Added Weaviate, Supabase, HNSWLib and memory self-query retrievers, along with their integration tests, docs and examples. Also added Chroma and Pinecone self query integration tests. Separated the basic translator for Pinecone and Chroma into two separate translator classes because they have two different allowed comparators, while keeping the BasicTranslator class backwards compatible.


jacoblee93:
Wow! This is a big one - would you be ok with doing this piecemeal over several PRs? Would make it easier to review + test.

ppramesi:
> Wow! This is awesome but also a pretty big one - would you be ok with doing this piecemeal over several PRs? Would make it easier to review + test.

Yeah I think I can do that. I'll split this PR 4 ways: Weaviate self-query, Supabase self-query, HWNSLib and memory self-query, and the Chroma and Pinecone self-query integration test. Does that sound good to you @jacoblee93?

jacoblee93:
Sounds great, that's exactly what I was thinking

ppramesi:
@jacoblee93 Okay in that case I'm going to close this pull request. The new pull requests are #1541, #1544, #1543 and #1542.

Issue #1539, opened 2023-06-05
Zep memory doesn't return memoryKey when getMemory() is not found.


chrisle:
When `memory = await this.zepClient.getMemory(this.sessionId, lastN);` fails with `NotFoundError` it should return the `memoryKey` if true.

chrisle:
See: https://github.com/hwchase17/langchainjs/pull/1540

PR #1540, opened 2023-06-05
Include memory key when getMemory returns NotFoundError.

chrisle:
Fixes # 1539


jacoblee93:
Fixes #1539

jacoblee93:
Can you run `yarn format`?

PR #1541, opened 2023-06-05
New Feat: Added Weaviate self-query retriever

ppramesi:
Added Weaviate self query retriever, along with the doc, example and integration test.


jacoblee93:
Hey unfortunately I think the end-to-end example requires some schema setup and I wasn't able to get it working - could you add something like that?

ppramesi:
@jacoblee93 hmmm that's odd. Mine works out of the box using Weaviate cloud sandbox, at least the weaviate_self_query.int.test.ts

ppramesi:
> Hey unfortunately I think the end-to-end example requires some schema setup and I wasn't able to get it working - could you add something like that?

@jacoblee93 @nfcampos I think I figured out what's wrong. Jacob mentioned that he's using local instance of Weaviate, which I presume using Docker image. The docker compose yaml file generated by Weaviate does not include API key or OIDC settings, and on Weaviate self query integration test, the client by default includes api key on the Weaviate client contstructor, which will throw error is used with Weaviate instance that does not have api key setup.

nfcampos:
Thanks, merged!

PR #1543, opened 2023-06-05
New Feat: Added HNSWLib and in-memory self-query retriever

ppramesi:
Added HNSWLib and in-memory self-query retriever. Also added docs, examples and integration tests. Also added filter for in-memory vector store.


docs/docs/modules/indexes/retrievers/hnswlib-self-query.mdx
```diff
@@ -0,0 +1,14 @@
+# Self Query HNSWLib Retriever
```
jacoblee93, line 1: I'll move these into their own section of the retriever docs

jacoblee93 (commented):
Looks awesome, one small question

langchain/src/retrievers/self_query/functional_translator.ts
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  getComparatorFunction<C extends Comparator>(
+    comparator: Comparator
+  ): (a: string | number, b: ValueType[C]) => boolean {
+    switch (comparator) {
+      case Comparators.eq: {
+        return (a: string | number, b: ValueType[C]) => a === b;
+      }
+      case Comparators.ne: {
```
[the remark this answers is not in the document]
ppramesi: I messed up here. `ne` and `neq` are shorthand for the same comparator "not equal", and there should be only one of them and it should be `ne`. I'm going to change all occurrences of `neq` and change it to `ne` here and the other 2 pull requests.

langchain/src/retrievers/self_query/tests/hnswlib_self_query.int.test.ts
```diff
@@ -0,0 +1,110 @@
+import { test } from "@jest/globals";
+import { Document } from "../../../document.js";
+import { AttributeInfo } from "../../../schema/query_constructor.js";
+import { OpenAIEmbeddings } from "../../../embeddings/openai.js";
+import { SelfQueryRetriever } from "../index.js";
+import { OpenAI } from "../../../llms/openai.js";
+import { FunctionalTranslator } from "../functional_translator.js";
+import { HNSWLib } from "../../../vectorstores/hnswlib.js";
+
+test.skip("HNSWLib Store Self Query Retriever Test", async () => {
```
jacoblee93, line 10: Also @ppramesi this caused OOM errors for me sporadically - any idea why?

jacoblee93:
`langchain/src/chains/query_constructor/tests/query_chain.int.test.ts` is failing - you can run with:
`yarn test:single langchain/src/chains/query_constructor/tests/query_chain.int.test.ts`

ppramesi:
@jacoblee93 Fixed failing query_chain test and HNSWLib test failing sometimes by removing `in` comparator. The LLM will use multiple `or` comparators instead when building queries.

PR #1546, opened 2023-06-05
fix #1475 update faiss-node to catch native exception

### LINKED ISSUE #1475: FAISS crashes NestJS app when index doesn't exist
I'm using langchain and the FAISS integration in a NestJS app, and I noticed that trying to do "FaissStore.load" with a path that doesn't exist results in a hard crash that cannot be catched. The app crashes with the following error:

```
Error: ENOENT: no such file or directory, open 'C:\...\docstore.json'
error Command failed with exit code 3221226505.
```

I'm not entirely sure why this happens, but the crash seems to occur when trying to run "FAISS.read" with a non-existent "faiss.index" file. Checking for the file existence before running that method seems to prevent the crash, although this is just a workaround, as the proper fix should be implemented on the faiss-node side.

In faiss.js, line 105
```js
    static async load(directory, embeddings) {
        const fs = await import("node:fs/promises");
        const path = await import("node:path");
        const readStore = (directory) => fs
            .readFile(path.join(directory, "docstore.json"), "utf8")
            .then(JSON.parse);
        const readIndex = async (directory) => {
            const { IndexFlatL2 } = await this.importFaiss();
            // FIX
            const filePath = path.join(directory, "faiss.index");
            await fs.access(filePath, fs.F_OK);
            return IndexFlatL2.read(filePath);
            // ---
        };
        const [[docstoreFiles, mapping], index] = await Promise.all([
            readStore(directory),
            readIndex(directory),
        ]);
        const docstore = new InMemoryDocstore(new Map(docstoreFiles));
        return new this(embeddings, { docstore, index, mapping });
    }
```

ewfian:
- [x]  Update faiss-node.
- [x]  Add Unit Test.

Fixes #1475


jacoblee93:
Thank you very much!

PR #1548, opened 2023-06-05
Store events on each run

nfcampos:


langchain/src/callbacks/tests/tracer.test.ts
agola11, line 1: might be good to add a check for `error` here?

agola11 (approved):
looks good

PR #1549, opened 2023-06-05
Use SDK Package for LCP Client

vowelparrot:
Remove the implementation of LCP client and instead import from the SDK. Split out the `runOnDataset` method into a separate function (rather than extending the client. I think that could be more confusing. Open to pushback)

### DIFF
--- a/examples/src/client/tracing_datasets.ts
+++ b/examples/src/client/tracing_datasets.ts
@@ -1,6 +1,7 @@
 /* eslint-disable no-process-env */
 import { ChatOpenAI } from "langchain/chat_models/openai";
-import { LangChainPlusClient, Dataset } from "langchain/client";
+import { runOnDataset } from "langchain/client";
+import { LangChainPlusClient, Dataset } from "langchainplus-sdk";
 import { SerpAPI } from "langchain/tools";
 import { Calculator } from "langchain/tools/calculator";
 import { initializeAgentExecutorWithOptions } from "langchain/agents";
@@ -55,8 +56,9 @@ export const run = async () => {
   // 4. Add the other examples to the dataset as well
 
   // So that you don't have to create the dataset manually, we will create it for you
-  const client: LangChainPlusClient = await LangChainPlusClient.create({
-    apiUrl: "http://localhost:1984",
+  const client: LangChainPlusClient = new LangChainPlusClient({
+    // apiUrl: "https://api.langchain.plus", // Default: LANGCHAIN_ENDPOINT environment variable of "http://localhost:1984"
+    // apiKey: "<Your API Key>", // Default: LANGCHAIN_API_KEY environment variable
   });
   const csvContent = `
 input,output
@@ -101,6 +103,6 @@ what is 1213 divided by 4345?,approximately 0.2791714614499425
 
   // If using the traced dataset, you can update the datasetName to be
   // "calculator-example-dataset" or the custom name you chose.
-  const results = await client.runOnDataset(datasetName, executorFactory);
+  const results = await runOnDataset(datasetName, executorFactory, { client });
   console.log(results);
 };
--- a/examples/src/client/tracing_datasets.ts
+++ b/examples/src/client/tracing_datasets.ts
@@ -1,7 +1,7 @@
 /* eslint-disable no-process-env */
+import { LangChainPlusClient, Dataset } from "langchainplus-sdk";
 import { ChatOpenAI } from "langchain/chat_models/openai";
 import { runOnDataset } from "langchain/client";
-import { LangChainPlusClient, Dataset } from "langchainplus-sdk";
 import { SerpAPI } from "langchain/tools";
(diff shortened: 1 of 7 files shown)


jacoblee93:
Will look at this by EOD!

examples/src/client/tracing_datasets.ts
```diff
@@ -101,6 +103,6 @@ what is 1213 divided by 4345?,approximately 0.2791714614499425
 
   // If using the traced dataset, you can update the datasetName to be
   // "calculator-example-dataset" or the custom name you chose.
-  const results = await client.runOnDataset(datasetName, executorFactory);
+  const results = await runOnDataset(datasetName, executorFactory, { client });
```
jacoblee93, line 106: I'm not such a fan of these freestanding imported functions as they have less context (also applies to all those `initialize/load` methods as well - definitely prefer the old `client.runOnDataset` personally.

langchain/src/client/index.ts
```diff
@@ -1,6 +1 @@
-export {
-  LangChainPlusClient,
-  Dataset,
-  Example,
-  DatasetRunResults,
-} from "./langchainplus.js";
+export { DatasetRunResults, runOnDataset } from "./langchainplus.js";
```
jacoblee93, line 1: Do we actually need this export at all anymore?

jacoblee93: Or just for backwards compat?

langchain/src/client/langchainplus.ts
```diff
@@ -1,94 +1,18 @@
-import { BaseRun, Run, RunType } from "../callbacks/handlers/tracer.js";
+import { Example, LangChainPlusClient } from "langchainplus-sdk";
```
jacoblee93, line 1: What's the rationale for keeping some methods in the main framework but not others? Why not move all these into the new SDK?

examples/src/client/tracing_datasets.ts
```diff
@@ -1,6 +1,7 @@
 /* eslint-disable no-process-env */
 import { ChatOpenAI } from "langchain/chat_models/openai";
-import { LangChainPlusClient, Dataset } from "langchain/client";
+import { runOnDataset } from "langchain/client";
+import { LangChainPlusClient, Dataset } from "langchainplus-sdk";
```
jacoblee93, line 4: Importing from `"langchain/client"` and then not importing `LangChainPlusClient` from that entrypoint looks a little odd to me

jacoblee93 (commented):
Left some comments/questions

langchain/src/client/langchainplus.ts
vowelparrot: The main rationale is that the SDK doesn't depend on any LangChain primitives, so things like "chains", "llms", etc. aren't accessible (unless we add duplicate typing stubs/schema for the BaseLanguageModel, Chain, etc. to the SDK).

So then when running in LangChain, the most straightforward things to do would either be:
- Standalone functions (current proposal)
- Extended client

Between the latter two, I don't have a strong preference but it's not a 1-way door

langchain/src/client/index.ts
vowelparrot: Specifically for importing from the index? Are we discouraging importing from the index for ALL directories or just the ones with optional dependencies?

examples/src/client/tracing_datasets.ts
vowelparrot: Ya it's less clean looking. Added more [context](https://github.com/hwchase17/langchainjs/pull/1549#discussion_r1219717005) in that message but the reason I initially chose this over extending the client is it also seemed slightly confusing to have an extended client

examples/src/client/tracing_datasets.ts
vowelparrot: Ya I'm not sure what to call the directory now. Could be langchainplus but then will likely rename anyway

jacoblee93: It feels like `runOnDataset` specifically seems like it should be in `langchainplus-sdk` - `Dataset` is imported from the SDK and it's odd that the concept of a dataset wouldn't be self-contained. There's no way to abstract out the LangChain specific references from this method and just put everything in the new SDK then?

jacoblee93: Could we frame this as a "LangChain LangChain Plus adapter" maybe?

vowelparrot: Hm there's a lot of types to include: llms, chat models, callbacks. I think that could be in another PR.

While it would be nice to have run utils in the SDK, they're only relevant to langchain primitives so I'm not sure we'd need to bundle them there

examples/src/client/tracing_datasets.ts
nfcampos: I think this is fine, we can always mimic the old behaviour by doing `import * as client from "langchain/blabla"`

PR #1552, opened 2023-06-06
remove interface dependency for milvus sdk

### LINKED ISSUE #1551: Remove interface dependency for milvus-sdk-node
We should remove the dependency from @zilliz/milvus2-sdk-node/dist/milvus/const/Milvus.js.  
In newer version, it would be removed. 

You can just import these from the package

From: 
```javascript
import {
  MilvusClient,
  DataType,
  DataTypeMap,
} from "@zilliz/milvus2-sdk-node/dist/milvus/const/Milvus.js";
```

To:
```javascript
import {
  MilvusClient,
  DataType,
  DataTypeMap,
} from "@zilliz/milvus2-sdk-node";
```

shanghaikid:
Fixes #1551

### DIFF
--- a/langchain/src/vectorstores/milvus.ts
+++ b/langchain/src/vectorstores/milvus.ts
@@ -1,13 +1,11 @@
 import * as uuid from "uuid";
-import { MilvusClient } from "@zilliz/milvus2-sdk-node";
 import {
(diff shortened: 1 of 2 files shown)


jacoblee93:
Thank you!

PR #1555, opened 2023-06-06
Fix: Allow Redis addDocuments without overriding existing keys

Haim763:
The current behavior of Redis VectorStore is overriding current keys (since it starts with idx 0 and not considering the previous state of the DB.

This PR takes the previous keys count and uses it as a reference for the next run of addDocument.


Haim763:
@jacoblee93 Why Vercel is failing?

jacoblee93:
Hey @Haim763 apologies for the delay - getting caught back up here.

`keys` is an `O(n)` operation that Redis doesn't recommend using in production - could you try using `ft.info` and seeing if the `num_docs` or other fields there help? https://redis.io/commands/ft.info/

jacoblee93:
Thank you!

PR #1557, opened 2023-06-06
[FIX] s3loader credential interface

QasimRRizvi:
Fixes #[1460](https://github.com/hwchase17/langchainjs/issues/1460)


jacoblee93:
This has been broken the whole time?

jacoblee93:
Ok, either syntax seems to work - updating with backwards compatibility in mind for those folks using the old syntax.

QasimRRizvi:
> patibility in mind for those folks using t

Yes it was, last time when I was using this I created patch for myself but I saw someother are facing the same issue so I decided to solve it.

jacoblee93:
Thanks again and sorry about the wait! Had an issue with the mocks.

Issue #1532, opened 2023-06-04
Support for Vertex AI Matching Engine as a Vector Store


afirstenberg:
See also (from the Python version):
- hwchase17/langchain#2892
- hwchase17/langchain#2971
- hwchase17/langchain#3104

Documentation:
- https://cloud.google.com/vertex-ai/docs/matching-engine/overview

I'll be working on this.

nfcampos:
Thank you

afirstenberg:
Only a little - was going to do more this weekend, but if you'd like to get started, feel free.
(Particularly if you understand the Matching Engine. I've been trying to wrap my head around it, and I feel like I have only the barest grasp of it.)
Would love if you shared your branch so I can follow along or work with you.

My main thoughts:
- Use the Docstore to store the document (so I needed to get that in the codebase first)
- Assume streaming updates
- Use the metadata to generate "restricts" values?

Your thoughts?

afirstenberg:
@jaythegeek 
Only a little - was going to do more this weekend, but if you'd like to get started, feel free.
(Particularly if you understand the Matching Engine. I've been trying to wrap my head around it, and I feel like I have only the barest grasp of it.)
Would love if you shared your branch so I can follow along or work with you.

My main thoughts:

- Use the Docstore to store the document (so I needed to get that in the codebase first, which it now is)
- Assume streaming updates
- Use the metadata to generate "restricts" values?

Your thoughts?

afirstenberg:
@jaythegeek - Have you done anything on this? It's rising closer to the top of my list and want to help out if you have things in progress.

davidoort:
What's the status of this @afirstenberg ?

afirstenberg:
I have the basics working - add a vector and query for a nearby vector.
But the documentation for the REST interface is... non-existent. And matching with a filter doesn't work at all.

I'm in email exchanges with Google DevRel, but haven't gotten clear answers yet.

So I can add what I have (which doesn't support restrictions or matching filters) which will change once I get clear documentation about the filtering.

Thoughts?

afirstenberg:
@davidoort 

I have the basics working - add a vector and query for a nearby vector.
But the documentation for the REST interface is... non-existent. And matching with a filter doesn't work at all.

I'm in email exchanges with Google DevRel, but haven't gotten clear answers yet.

So I can add what I have (which doesn't support restrictions or matching filters) which will change once I get clear documentation about the filtering.

Thoughts?

[...]

PR #1564, opened 2023-06-07
Docstore updates - change base class and add GoogleCloudStorage Docstore

afirstenberg:
Prerequisite for #1532 

This changes the Docstore base class to make the methods asynchronous, and adds a new Synchronous base class for the in-memory docstore to use.
Additionally, this new Docstore base class is the basis for a Google Cloud Storage (GCS) implementation.


nfcampos (approved):
Thanks, LGTM

afirstenberg:
Just to make sure I followed, @nfcampos - we're getting rid of the docstore directory and, essentially, moving everything from it to stores/doc?

nfcampos:
Yes exactly. It's something I've had in mind to do for a while, and now that you're introducing the first implementation of doc store (other than in memory) is the best time to do it. The other change removing the synchronous base class is just to make it less likely people open PRs implementing more synchronous doc stores, which we don't want

afirstenberg:
Got it, and looks good. Thanks!

PR #1565, opened 2023-06-07
Add Trace Grouping Helper

vowelparrot:
Add helper function for grouping traces similar to the context manager in the [python library](https://github.com/hwchase17/langchain/blob/a0d847f63628c77d34e9e6003f45af6d59d83288/langchain/callbacks/manager.py#L131)

### DIFF
--- a/langchain/src/callbacks/manager.ts
+++ b/langchain/src/callbacks/manager.ts
@@ -603,7 +603,7 @@ export class TraceGroup {
   }
 }
 
-export async function traceAsChainGroup<T, A extends any[]>(
+export async function traceAsGroup<T, A extends any[]>(
   groupOptions: {
     name: string;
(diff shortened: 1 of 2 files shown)


langchain/src/callbacks/manager.ts
```diff
@@ -556,3 +560,61 @@ function ensureHandler(
...
+
+  async end(): Promise<void> {
+    if (this.runManager) {
+      await this.runManager.handleChainEnd({});
+      this.runManager = undefined;
+    }
+  }
+}
+
+export async function traceAsChainGroup<T, A extends any[]>(
```
agola11, line 606: nit: maybe just `traceAsGroup`?

agola11 (commented):
lgtm

PR #1568, opened 2023-06-07
Use swc-loader to build docs instead of babel-loader

jacoblee93:
Should speed up build times and hopefully reduce memory/flaky CI

See discussion here: https://github.com/facebook/docusaurus/issues/4765
And here: https://github.com/rancher/rancher-docs/pull/667


jacoblee93:
CC @nfcampos

PR #1582, opened 2023-06-08
GoogleVertexAI "Codey" support for code-bison and code-gecko models

### LINKED ISSUE #1575: GoogleVertexAI class does not support "code-bison" model.
It looks like the required parameters for `code-bison` are different than those for `text-bison`, so just changing the value of `model` doesn't work.
I'm looking at the best way to handle this.

Do we have a best practice for handling code models vs text models for LLMs?

afirstenberg:
Fix #1575


nfcampos (approved):
Great, thanks!

PR #1586, opened 2023-06-08
docs: Remove `node-only` label from the SingleStoreVectorStore doc

volodymyr-memsql:
As the Python implementation is also available, we do not need to mark it as `Node-only` anymore.

@hwchase17


volodymyr-memsql:
@jacoblee93 could you please take a look? It should be a trivial change

nfcampos:
Note, Node-only here means it isn't available in other JS environments, eg browser, edge etc

PR #1595, opened 2023-06-08
Add method to identify BaseChatMessage type without using hard coded text like "human"

borel:
When trying to guess the type of ChatMessage there is a lot of equality calculation based on hard coded string like === "human", this MR is just about refactoring this to use isXX ( eg isHuman() ) instead .


christian-bromann:
@borel thanks for the contribution 🙏 we have overhauled the standard message format in the `v1` branch and don't think we can incorporate this change anymore. Would love to get your feedback though on the new format.

PR #1599, opened 2023-06-09
add delete function to hnswlib index

### LINKED ISSUE #1107: Delete function for hnswlib
I needed this in my app since I create different indexes on one server. Not sure If I should update the interface since this might be a specific feature for disk / in-memory stores only. 

https://github.com/hwchase17/langchainjs/compare/main...anandkumarpatel:feature-hnswlib-delete?expand=1

I can add tests and flesh it out more if other people find this useful.

anandkumarpatel:
Adding delete function for `hnswlib.ts`. Useful when you have limited disk space or use many indexes and need to clean them up.

Fixes #1107


tomi-mercado:
Cool improvement!

jacoblee93:
Thanks for your patience and not giving up on this!

Updated a few things but should be good to go in tonight's release.

PR #1603, opened 2023-06-10
Fix docs typo from `unstructuredClone` to `structuredClone`

MaxMusing:
The docs for running Langchain with Node.js 16 have a typo that says you need to polyfill `unstructuredClone`, when it should really say `structuredClone`.

Docs link: https://js.langchain.com/docs/getting-started/install#unsupported-nodejs-16


nfcampos:
Thanks!

Celthi:
Why does the structuredClone need to be polyfill ? I cannot search the code base and find the function is being used.

PR #1615, opened 2023-06-12
Improved build environment compatibility

skarard:
As a small enhancement aimed at making our Windows developers' lives a little easier, I've replaced instances of `rm -rf` in the codebase with `rimraf`.

`rimraf` effectively mirrors the functionality of the `rm -rf` command, but with the added benefit of cross-platform consistency. This means it helps us maintain uniform behaviour when deleting files and directories, regardless of whether we're working on Windows, MacOS, Linux.


nfcampos (approved):
Thanks!

PR #1617, opened 2023-06-12
Optimized Notion Document Loader with Extended Features

skarard:
I've been building upon the superb groundwork set by the teams behind the web/notiondb and fs/notion document loaders. While their work couldn't be directly incorporated, it was an invaluable foundation for what I've managed to create.

Today, I'm excited to present to you an optimized Notion document loader. Here's why it stands out:

- It's been optimized for fewer API calls and faster speed.
- Type safety has been ensured by incorporating the official [notion-sdk-js](https://github.com/makenotion/notion-sdk-js). 
- It is fully compatible with all Notion block types, which are seamlessly translated into markdown using [notion-md](https://github.com/souvikinator/notion-to-md).
- It can also load pages and databases, which includes loading child pages.
- Plus, the document metadata now includes all the page properties for improved vectordb referencing.

I believe this PR is ready for the final review stage. I've followed the contribution guidelines closely, but since this is my first PR into this repo, I'm eager to hear any feedback you may have. Your insights are always appreciated.


nfcampos:
This looks great, thanks! Would you be able to add an example and a docs page?

skarard:
@nfcampos documents and examples have been added.

gkhngyk:
Hi @hwchase17  
Do you have a pull request plan for this commits? I wanted to write it specifically because it looks great.

skarard:
> This looks great, thanks! Would you be able to add an example and a docs page?

@nfcampos is there anything stopping this from being merged?

skarard:
!remind @nfcampos

jacoblee93:
It looks like this supersedes the existing NotionDBLoader? Should we deprecate that one?

skarard:
> It looks like this supersedes the existing NotionDBLoader? Should we deprecate that one?

Yep, the new code does the same job as NotionDBLoader but in a snappier and more feature-packed way. Having both might confuse users and be a headache to maintain. So, let's make life simpler and stick to the new stuff. Thanks for flagging this.

gkhngyk:
> It looks like this supersedes the existing NotionDBLoader? Should we deprecate that one?

After seeing this commit, I tested the existing NotionDBLoader in langchain with 3 Notion Databases. None of them yielded results. I think the current NotionDBLoader is not working well. But I think NotionDBLoader in this commit works fine.

jacoblee93:
Thanks again for doing this! Ran through it and it's excellent.

I changed the constructor to take a `clientOptions` arg for the Notion client rather than a client instance itself to cut down on an import for devs (typed with the Notion client's options type), deprecated the `NotionDBLoader`, and updated the docs with a notice about the other required peer dep. Should be good to go with tomorrow's release!

PR #1620, opened 2023-06-13
Upgrade SDK Version

vowelparrot:


langchain/src/callbacks/handlers/tracer_langchain.ts
```diff
@@ -1,12 +1,18 @@
...
-import { Run, RunCreate, RunUpdate } from "langchainplus-sdk/schemas";
+import { BaseRun, RunCreate, RunUpdate } from "langchainplus-sdk/schemas";
 import {
   getEnvironmentVariable,
   getRuntimeEnvironment,
 } from "../../util/env.js";
 import { BaseTracer } from "./tracer.js";
 import { BaseCallbackHandlerInput } from "../base.js";
 
+export interface Run extends BaseRun {
```
vowelparrot, line 10: The Run object in the SDK doesn't have the child_runs and child execution order used to track things in the Tracer so I'm updating it here. i had previously conflated this run object with the one passed to the evaluator / loaded from the DB

PR #1631, opened 2023-06-13
Nc/openai functions

nfcampos:
Fixes # (issue)


siisee11:
Can you provide an example of using an openai function with `get_current_weather` tool?

nfcampos:
@siisee11 added more docs here https://github.com/hwchase17/langchainjs/pull/1646/files let me know any questions

Fran-3xo:
Hi @nfcampos sorry for bothering you, I wanted to know if you can clarify a question for me with a problem I have with langchain, pinecone and openai, is it possible to use the pinecone integration together with function calls? I'm trying to use it and I can't generate any completion, I always get an error, could you give me a hand?

PR #1635, opened 2023-06-14
Update openai.mdx

sachio222:
This agent type is not currently supported by JS langchain

Fixes # (issue)


nfcampos:
It is supported as of yesterday

PR #1684, opened 2023-06-16
Bug fix: zepMemory needs to return memorykey:[] and not empty string …

rsharath:
If memory not found, loadMemoryVariables returns [], instead needs to return memorykey:[]

### DIFF
--- a/langchain/src/memory/zep.ts
+++ b/langchain/src/memory/zep.ts
@@ -67,7 +67,9 @@ export class ZepMemory extends BaseChatMemory implements ZepMemoryInput {
     } catch (error) {
       // eslint-disable-next-line no-instanceof/no-instanceof
       if (error instanceof NotFoundError) {
-        const result = this.returnMessages ? { [this.memoryKey]: [] } : { [this.memoryKey]: "" };
+        const result = this.returnMessages
+          ? { [this.memoryKey]: [] }
+          : { [this.memoryKey]: "" };
         return result;
       } else {
         throw error;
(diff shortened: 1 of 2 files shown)


rsharath:
@nfcampos could you please take a look and see if this can be included in the next release please? Thanks

danielchalef:
@jacoblee93 Would it be possible to get this merged soon? We have users complaining about this bug. Thanks!

langchain/src/memory/zep.ts
```diff
@@ -67,8 +67,7 @@ export class ZepMemory extends BaseChatMemory implements ZepMemoryInput {
     } catch (error) {
       // eslint-disable-next-line no-instanceof/no-instanceof
       if (error instanceof NotFoundError) {
-        if (this.returnMessages) return { [this.memoryKey]: [] };
-        return [];
+        return { [this.memoryKey]: [] };
```
nfcampos, line 70: This doesn't look correct, if returnMessages is false it should instead maybe return `{ [this.memoryKey]: '' }` ?

rsharath: @nfcampos - we need to return an empty list not an empty string because memory is an array of messages.

rsharath: please let me know if you see an issue with this.

nfcampos: No that's only true when returnMessages is true. When it is false memory returns a string (that concatenates the messages). See line 99 where it returns the result of calling getBufferString, ie. a string

rsharath: ah! thanks, updated to return this.returnMessages ? { [this.memoryKey]: [] } : { [this.memoryKey]: "" };

nfcampos (approved):

PR #1696, opened 2023-06-19
Include placeholder value for all secrets used, not just those in kwargs

nfcampos:
Fixes # (issue)

### DIFF
--- a/langchain/src/chat_models/openai.ts
+++ b/langchain/src/chat_models/openai.ts
@@ -159,6 +159,8 @@ export class ChatOpenAI
 
   maxTokens?: number;
 
+  openAIApiKey?: string;
+
   azureOpenAIApiVersion?: string;
 
   azureOpenAIApiKey?: string;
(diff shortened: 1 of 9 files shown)


langchain/src/chat_models/openai.ts
```diff
@@ -175,35 +177,33 @@ export class ChatOpenAI
...
-    if (!azureApiKey && !apiKey) {
+
+    if (!this.azureOpenAIApiKey && !this.openAIApiKey) {
       throw new Error("(Azure) OpenAI API key not found");
     }
 
-    const azureApiInstanceName =
+    this.azureOpenAIApiInstanceName =
       fields?.azureOpenAIApiInstanceName ??
       getEnvironmentVariable("AZURE_OPENAI_API_INSTANCE_NAME");
```
dqbd, line 200: Shouldn't we also potentially add every env as a secret?

dqbd (commented):
Looks good!

dqbd (commented):
Looks good! Just a small question

langchain/src/chat_models/openai.ts
dqbd: discussed: these variables should be passed manually

PR #1698, opened 2023-06-19
Add error handling for openai streaming when errors are sent as events in the SSE stream

nfcampos:
Fixes # (issue)

### DIFF
--- a/langchain/src/chat_models/openai.ts
+++ b/langchain/src/chat_models/openai.ts
@@ -321,7 +321,6 @@ export class ChatOpenAI
               adapter: fetchAdapter, // default adapter doesn't do streaming
               responseType: "stream",
               onmessage: (event) => {
-                console.log(event);
                 if (event.data?.trim?.() === "[DONE]") {
                   if (resolved || rejected) {
                     return;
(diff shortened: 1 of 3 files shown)


langchain/src/chat_models/openai.ts
```diff
@@ -321,19 +321,31 @@ export class ChatOpenAI
               adapter: fetchAdapter, // default adapter doesn't do streaming
               responseType: "stream",
               onmessage: (event) => {
+                console.log(event);
```
dqbd, line 324: _nit_: unnecessary console.log

langchain/src/llms/openai-chat.ts
```diff
@@ -264,13 +264,24 @@ export class OpenAIChat
...
+                  if (data?.error) {
+                    if (rejected) {
+                      return;
+                    }
+                    rejected = true;
+                    reject(data.error);
+                    return;
+                  }
+
+                  const message = data as {
```
dqbd, line 284: _question_: should the optionality of `choices` be applied here as well?

dqbd (commented):
LGTM, only small nitpicks

PR #1707, opened 2023-06-20
split up batch runs for llms and chat models in generate

agola11:
Fixes # (issue)

### DIFF
--- a/langchain/src/chat_models/base.ts
+++ b/langchain/src/chat_models/base.ts
@@ -56,9 +56,7 @@ export abstract class BaseChatModel extends BaseLanguageModel {
     options?: string[] | this["CallOptions"],
     callbacks?: Callbacks
   ): Promise<LLMResult> {
-    const generations: ChatGeneration[][] = [];
-    let results: ChatResult[] = [];
-    const llmOutputs: LLMResult["llmOutput"][] = [];
+    // parse call options
     let parsedOptions: this["CallOptions"];
     if (Array.isArray(options)) {
       parsedOptions = { stop: options } as this["CallOptions"];
@@ -70,6 +68,7 @@ export abstract class BaseChatModel extends BaseLanguageModel {
     } else {
       parsedOptions = options ?? {};
     }
+    // create callback manager and start run
     const callbackManager_ = await CallbackManager.configure(
       callbacks,
       this.callbacks,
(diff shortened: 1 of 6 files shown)


langchain/src/chat_models/base.ts
```diff
@@ -80,17 +81,17 @@ export abstract class BaseChatModel extends BaseLanguageModel {
...
       undefined,
       undefined,
       extra
     );
     try {
-      const results = await Promise.all(
+      results = await Promise.all(
         messages.map((messageList) =>
-          this._generate(messageList, parsedOptions, runManager)
+          this._generate(messageList, parsedOptions, runManagers?.[0])
```
nfcampos, line 94: this should be `i` instead of 0

langchain/src/chat_models/base.ts
```diff
@@ -110,9 +113,21 @@ export abstract class BaseChatModel extends BaseLanguageModel {
         ? this._combineLLMOutput?.(...llmOutputs)
         : undefined,
     };
-    await runManager?.handleLLMEnd(output);
+
+    const flattenedOutputs: LLMResult[] = results.map((result) => ({
```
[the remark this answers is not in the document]
nfcampos: But actually why do we need this code? Doesn't _generate already return a LLMResult?

langchain/src/llms/base.ts
```diff
@@ -111,18 +137,24 @@ export abstract class BaseLLM extends BaseLanguageModel {
...
     } catch (err) {
-      await runManager?.handleLLMError(err);
+      await Promise.all(
+        (runManagers ?? []).map((runManager) => runManager?.handleLLMError(err))
+      );
       throw err;
     }
 
-    await runManager?.handleLLMEnd(output);
-    // This defines RUN_KEY as a non-enumerable property on the output object
```
nfcampos, line 121: Did you remove the comments on purpose?

langchain/src/chat_models/base.ts
agola11: It actually doesn't matter which one we pass as the handlers are all the same. I guess using i is more idiomatic though

langchain/src/chat_models/base.ts
```diff
@@ -80,17 +81,17 @@ export abstract class BaseChatModel extends BaseLanguageModel {
...
+    const runManagers = await callbackManager_?.handleChatModelStart(
       this.toJSON(),
       messages,
       undefined,
       undefined,
       extra
     );
     try {
-      const results = await Promise.all(
+      results = await Promise.all(
```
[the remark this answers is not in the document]
agola11: Sg happy to add as a follow up

langchain/src/llms/base.ts
agola11: Sorry didn't mean to remove comments

PR #1717, opened 2023-06-21
feat: Update SingleStoreVectorStore

volodymyr-memsql:
This update encompasses:

- Introduction of `distanceMetrics` configuration for selecting between `DOT_PRODUCT` (default) and `EUCLIDEAN_DISTANCE` similarity metrics.
- Inclusion of `filter` parameter in similarity search functions for metadata-based result prefiltering.
- Relocation of `createPool` call to `SingleStoreVectorStore` constructor; user should provide connection parameters (`connectionURI` or `connectionOptions`). This alteration is a breaking change; to retain compatibility with older code, `pool` could be optionally provided as an alternative to connection parameters.
- Documentation and examples have been refreshed accordingly.
@jacoblee93, please, take a look.


jacoblee93:
Hey @volodymyr-memsql, was on vacation but am back now and will look as soon as I can!

langchain/src/vectorstores/singlestore.ts
```diff
@@ -4,18 +4,81 @@ import type {
...
+};
+
+type ConnectionConfig = ConnectionWithUri | ConnectionWithOptions;
+
+export type SingleStoreVectorStoreConfig = ConnectionConfig & {
   tableName?: string;
   contentColumnName?: string;
   vectorColumnName?: string;
   metadataColumnName?: string;
+  distanceMetrics?: DistanceMetrics;
```
jacoblee93, line 42: This should probably be singular since you're only picking one metric at a time, right? `distanceMetric`

langchain/src/vectorstores/singlestore.ts
```diff
@@ -4,18 +4,81 @@ import type {
...
+      ...newOptions.connectAttributes,
+    },
+  };
+
+  if (!result.connectAttributes) {
+    result.connectAttributes = {};
+  }
+
+  if (
+    !Object.prototype.hasOwnProperty.call(
```
jacoblee93, line 70: What does `Object.prototype.hasOwnProperty.call` do vs. just `result.connectAttributes.program_name !== undefined`?

langchain/src/vectorstores/singlestore.ts
```diff
@@ -79,9 +149,40 @@ export class SingleStoreVectorStore extends VectorStore {
...
-    _filter?: undefined
+    filter?: Metadata
   ): Promise<[Document, number][]> {
-    // use vector DOT_PRODUCT as a distance function
+    // build the where clause from filter
+    const whereArgs: string[] = [];
+    const buildWhereClause = (record: Metadata, argList: string[]): string => {
+      const whereTokens: string[] = [];
+      for (const key in record)
+        if (Object.prototype.hasOwnProperty.call(record, key)) {
```
jacoblee93, line 159: Same question as above

jacoblee93 (changes requested):
Looks good, a few small questions

langchain/src/vectorstores/singlestore.ts
jacoblee93: Would prefer the latter for readability

volodymyr-memsql:
Thanks for the review, @jacoblee93. Could you please look one more time? Hopefully, I addressed your comment about `singular` correctly.

volodymyr-memsql:
Thanks for the review, @jacoblee93 . Could you please look one more time?

langchain/src/vectorstores/singlestore.ts
```diff
@@ -4,18 +4,79 @@ import type {
...
 import { format } from "mysql2";
+import { createPool } from "mysql2/promise";
+import { Metadata } from "@opensearch-project/opensearch/api/types.js";
 import { VectorStore } from "./base.js";
 import { Embeddings } from "../embeddings/base.js";
 import { Document } from "../document.js";
 
-export interface SingleStoreVectorStoreConfig {
-  connectionPool: Pool;
+export enum DistanceMetrics {
```
jacoblee93, line 16: Using an enum will unfortunately break builds for other non-Node envs due to how they get transpiled - removing should fix the build.

volodymyr-memsql:
Thanks @jacoblee93! I removed the `enum`. Could you please take a look again?

langchain/src/vectorstores/singlestore.ts
```diff
@@ -4,18 +4,76 @@ import type {
   OkPacket,
   ResultSetHeader,
   FieldPacket,
+  PoolOptions,
 } from "mysql2/promise";
 import { format } from "mysql2";
+import { createPool } from "mysql2/promise";
+import { Metadata } from "@opensearch-project/opensearch/api/types.js";
```
jacoblee93, line 11: Oh, you should add this to the docs as a peer dependency and use `import type`.

jacoblee93 (commented):
Good with me other than one small thing!

langchain/src/vectorstores/singlestore.ts
volodymyr-memsql: I prefer to redeclare this type inside my code.

volodymyr-memsql:
Thank you for the review, @jacoblee93. I addressed the issue in the latest commit.

jacoblee93 (approved):

jacoblee93:
Thank you!

PR #1724, opened 2023-06-21
docs: New experimental UI for Mendable Search 

nickscamara:
This PR introduces a new Mendable UI tailored to a better search experience.

We're more closely integrating our traditional search with our AI generation.
With this change, you won't have to tab back and forth between the mendable bot and the keyword search. Both types of search are handled in the same bar. This should make the docs easier to navigate. while still letting users get code generations or AI-summarized answers if they so wish. Also, it should reduce the cost.

Would love to hear your feedback :)


nfcampos:
Thanks, this is a good improvement

PR #1725, opened 2023-06-21
typesense: add vector_distance as score

jaclar:
Typesense vector search provides the vector_distance in the search result:

https://typesense.org/docs/0.24.1/api/vector-search.html#nearest-neighbor-vector-search

### DIFF
--- a/langchain/src/vectorstores/typesense.ts
+++ b/langchain/src/vectorstores/typesense.ts
@@ -1,9 +1,18 @@
 import type { Client } from "typesense";
 import type { MultiSearchRequestSchema } from "typesense/lib/Typesense/MultiSearch.js";
+import type {
+  SearchResponseHit,
+  DocumentSchema,
+} from "typesense/lib/Typesense/Documents.js";
 import type { Document } from "../document.js";
 import { Embeddings } from "../embeddings/base.js";
 import { VectorStore } from "./base.js";
 import { AsyncCaller, AsyncCallerParams } from "../util/async_caller.js";
+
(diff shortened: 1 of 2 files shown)


langchain/src/vectorstores/typesense.ts
```diff
@@ -230,15 +239,15 @@ export class Typesense extends VectorStore {
...
-    const hits = results?.map((hit) => hit.document) as
-      | Record<string, unknown>[]
-      | undefined;
 
-    const documents = this._typesenseRecordsToDocuments(hits).map(
-      (doc) => [doc, 0] as [Document<Record<string, unknown>>, number]
-    );
+    const hits = results?.map((hit: VectorSearchResponseHit<object>) => ({
+      document: hit?.document || {},
+      vector_distance: hit?.vector_distance || 2,
```
nfcampos, line 245: why default to 2?

jaclar: According to their [Vector Search Documentation](https://typesense.org/docs/0.24.1/api/vector-search.html#nearest-neighbor-vector-search):
>  Typesense uses the cosine similarity, so this distance will be a value between 0 and 2.
>
>  - If the document's vector perfectly matches the query vector, the distance will be 0
> - If the document's vector is extremely different from the query vector, then the distance will be 2.

So I assumed that if we don't get back a vector distance, we just assume that the match is the most different.

Which makes me think about the implicit conventions in vectorstores for the `score` value. Is the expectation that a higher score normally represents a higher match? AFAIK that is the default in pinecone for example. In the case of typesense it's the opposite

PR #1726, opened 2023-06-21
Overrode JSONLoader.load to accept JSON already in memory.

nigel-daniels:
This request adds functionality to the JSONLoader by overriding the load method so that you can provide it with JSON already held in memory. Often large JSON datasets can be returned from document databases or from REST requests. In these cases no filesystem JSON exists.


jacoblee93:
Apologies for the delay here.

The `load` method for these file system loaders isn't meant to take any arguments. I believe this would be equivalent to something like:
```
  const loader = new JSONLoader(
    new Blob(
      [
        `{
  "texts": ["This is a sentence.", "This is another sentence."]
}`,
      ],
      { type: "application/json" }
    )
  );
```

PR #1730, opened 2023-06-22
Refactor how the Google Vertex AI PaLM Codey  code- models are implemented

afirstenberg:
Related to #1575 
Refactoring how the GoogleVertexAI llm support is implemented and removes the GoogleVertexAICode class (which is a breaking change, but one I think folks are unlikely to run into).

### DIFF
--- a/langchain/src/llms/googlevertexai.ts
+++ b/langchain/src/llms/googlevertexai.ts
@@ -64,6 +64,13 @@ export class GoogleVertexAI extends BaseLLM implements GoogleVertexAITextInput {
     super(fields ?? {});
 
     this.model = fields?.model ?? this.model;
+
+    // Change the defaults for code models
+    if (this.model.startsWith("code-")) {
+      this.temperature = 0.2;
+      this.maxOutputTokens = 256;
+    }
+
     this.temperature = fields?.temperature ?? this.temperature;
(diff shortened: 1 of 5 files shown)


langchain/src/llms/googlevertexai.ts
```diff
@@ -64,6 +64,13 @@ export class GoogleVertexAI extends BaseLLM implements GoogleVertexAITextInput {
     super(fields ?? {});
 
     this.model = fields?.model ?? this.model;
+
+    // Change the defaults for code models
+    if (this.model.startsWith("code-")) {
+      this.temperature = 0.2;
```
nfcampos, line 70: Is this recommended by Google?

nfcampos (approved):
Thanks!

langchain/src/llms/googlevertexai.ts
afirstenberg: In conversations with Google, they said that 0.7 (which is what the configuration for the text model defaults to) is too high. They didn't have any specific suggestion for lower. My tests indicated that 0.5 produced poor results in my tests, but 0.2 produced good results.

nfcampos: cool, thanks for the context

PR #1736, opened 2023-06-22
Supports non-utf-8 character sets for Cheerio

flybluewolf:
Supports non-utf-8 character sets for Cheerio


nfcampos (approved):
Great, thanks

PR #1751, opened 2023-06-23
Add tags support to run group

vowelparrot:


langchain/src/callbacks/manager.ts
```diff
@@ -713,15 +728,26 @@ export class TraceGroup {
 export async function traceAsGroup<T, A extends any[]>(
   groupOptions: {
     name: string;
+    inheritableTags?: string[];
```
nfcampos, line 731: Here let's expose only `tags` (as in py) which will be inheritable

PR #1756, opened 2023-06-25
Update sql.ts

scenaristeur:
mix between https://langchainers.hashnode.dev/learn-how-to-integrate-language-models-llms-with-sql-databases-using-langchainjs and https://js.langchain.com/docs/modules/agents/toolkits/sql

Fixes # (issue)


jacoblee93:
Thanks for the PR! Not everyone will use Node/dotenv, so we try not to assume how people are loading env vars.

PR #1758, opened 2023-06-26
Mention that weaviate is a production-ready vector store too

hsm207:
Mention that weaviate is another option if users want a production-ready vector store

Fixes # (issue)
NA


jacoblee93:
Something happened with the diff here, going to close this one for now. Feel free to open another.

PR #1762, opened 2023-06-26
Downgrade decamelize for browser compatibility

### LINKED ISSUE #1761: [bug] Dependency decamelize >=4.0.0 breaks in Safari
## Problem Statement
Apps that include langchain.js as a dependency are not able to run on Safari <16.4 (released on March 26, 2023) due to the [JS Lookbehind regex assertion](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Lookbehind_assertion) introduced by langchain's dependency [decamelize >4.0.0](https://github.com/sindresorhus/decamelize/releases/tag/v4.0.0). 

See screenshot below:
image

## Proposed solution
Downgrade decimalize to ^3.2.0

lushiyun:
Fixes #1761

### DIFF
--- a/langchain/package.json
+++ b/langchain/package.json
@@ -718,7 +718,7 @@
     "ansi-styles": "^5.0.0",
     "binary-extensions": "^2.2.0",
     "camelcase": "6",
-    "decamelize": "^3.2.0",
+    "decamelize": "^1.2.0",
     "expr-eval": "^2.0.2",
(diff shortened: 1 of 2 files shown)


nfcampos (approved):
Thanks!

Downgraded even further to 1.2.0 to avoid adding a big dependency on regexp

PR #1766, opened 2023-06-26
Zep: Authentication; other bugfixes/updates; bump SDK version

danielchalef:
- zep auth; auth in example; fix for lastn
- if summary exists, add it to message history
- update class names
- bump zep-js version


rsharath:
@nfcampos / @jacoblee93  - this change is ready for merge, appreciate you taking a look. Thanks!

langchain/src/memory/zep.ts
```diff
@@ -51,15 +55,17 @@ export class ZepMemory extends BaseChatMemory implements ZepMemoryInput {
...
+    this.zepClient = new ZepClient(this.baseURL, fields.apiKey);
   }
 
   get memoryKeys() {
     return [this.memoryKey];
   }
 
   async loadMemoryVariables(values: InputValues): Promise<MemoryVariables> {
-    const lastN = values.lastN ?? 10;
+    // use either lastN provided by developer or undefined to use the
```
jacoblee93, line 66: Will this default to 10 for current users?

langchain/package.json
```diff
@@ -445,7 +445,7 @@
     "@clickhouse/client": "^0.0.14",
     "@faker-js/faker": "^7.6.0",
     "@getmetal/metal-sdk": "^4.0.0",
-    "@getzep/zep-js": "^0.3.1",
+    "@getzep/zep-js": "^0.4.1",
```
jacoblee93, line 448: Update the peer dep as well

jacoblee93 (changes requested):
The integration test is now failing for me - can you fix?

danielchalef:
This is a bug fix. The number of results returned should default to the message window size set on the server, not 10.

I’ll take a look at why the integration test is failing.

langchain/package.json
danielchalef: Fixed

danielchalef:
@jacoblee93 I've updated the peer dep, improved the robustness of the retriever, and added an additional test. Tests pass locally. Can you let me know if the CI tests pass? I can't see the results here.

As an aside, I have to manually modify the `test.skip` to `test` in to run them using `yarn test:single`. Is there a better way to force them to run from the CLI?

langchain/src/memory/zep.ts
danielchalef: This is a bug fix. The number of results returned should default to the message window size set on the server, not 10.

jacoblee93:
> @jacoblee93 I've updated the peer dep, improved the robustness of the retriever, and added an additional test. Tests pass locally. Can you let me know if the CI tests pass? I can't see the results here.
> 
> As an aside, I have to manually modify the `test.skip` to `test` in to run them using `yarn test:single`. Is there a better way to force them to run from the CLI?

The alternative is not skipping them at all by adding a container setup here:
https://github.com/hwchase17/langchainjs/blob/main/test-int-deps-docker-compose.yml

jacoblee93 (approved):

jacoblee93:
Fixed out of date yarn.lock and merged in with #1811

danielchalef:
Thanks, @jacoblee93 !

PR #1770, opened 2023-06-27
mention weaviate as an option

hsm207:
Fixes #1758


hsm207:
@jacoblee93 I've made another PR

PR #1777, opened 2023-06-27
feat: Google Calendar View/Create Tools

ShaunBaker:
This pull request provides two separate tools that provide the ability for users to interact with a Google Calendar to retrieve and create events in the calendar through input.

### DIFF
--- a/langchain/package.json
+++ b/langchain/package.json
@@ -1,6 +1,6 @@
 {
   "name": "langchain",
-  "version": "0.0.152",
+  "version": "0.0.153",
   "description": "Typescript bindings for langchain",
   "type": "module",
   "engines": {


langchain/src/tools/google_calendar/base.ts
```diff
@@ -0,0 +1,91 @@
...
+      );
+    }
+
+    this.clientEmail = fields.credentials.clientEmail;
+    this.privateKey = fields.credentials.privateKey;
+    this.calendarId = fields.credentials.calendarId;
+    this.scopes = fields.scopes || [];
+  }
+
+  getModel() {
```
jacoblee93, line 71: This should be passed in

langchain/src/tools/google_calendar/create.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  constructor(fields: GoogleCalendarAgentParams) {
+    super(fields);
+  }
+
+  async _call(query: string) {
+    const auth = await this.getAuth();
+    const model = this.getModel();
+
+    return await runCreateEvent(query, {
```
jacoblee93, line 18: `await` is unnecessary for return statements

langchain/src/tools/google_calendar/descriptions.ts
```diff
@@ -0,0 +1,24 @@
+export const CREATE_TOOL_DESCRIPTION = `A tool for creating Google Calendar events and meetings.
+
+INPUT example:
+"action": "google_calendar_create",
```
jacoblee93, line 4: Output parser is the standard way of doing this:

https://js.langchain.com/docs/modules/prompts/output_parsers/#structured-output-parser

langchain/src/tools/google_calendar/prompts/create-event-prompt.ts
```diff
@@ -0,0 +1,56 @@
...
+Today's datetime on UTC time 2023-05-02T10:00:00+00:00, it's Tuesday and timezone
+of the user is -5, take into account the timezone of the user and today's date.
+1. event_summary 
+2. event_start_time 
+3. event_end_time 
+4. event_location 
+5. event_description 
+6. user_timezone
+event_summary:
+{{
```
jacoblee93, line 14: See above

langchain/src/tools/google_calendar/prompts/view-events-prompt.ts
```diff
@@ -0,0 +1,34 @@
...
+Today's datetime on UTC time 2023-05-02T10:00:00+00:00, it's Tuesday and timezone
+of the user is -5, take into account the timezone of the user and today's date.
+If the user is searching for events with a specific title, person or location, put it into the search_query parameter.
+1. time_min 
+2. time_max
+3. user_timezone
+4. max_results 
+5. search_query 
+event_summary:
+{{
```
jacoblee93, line 14: See above

langchain/package.json
```diff
@@ -549,6 +549,7 @@
     "epub2": "^3.0.1",
     "faiss-node": "^0.2.1",
     "google-auth-library": "^8.8.0",
+    "googleapis": "^118.0.0",
```
jacoblee93, line 552: Add this to `devDependencies` above as well

langchain/src/tools/google_calendar/commands/run-create-events.ts
```diff
@@ -0,0 +1,125 @@
+import { google, calendar_v3 } from "googleapis";
+import type { JWT, GaxiosResponse } from "googleapis-common";
```
jacoblee93, line 2: This should be an explicit dep?

jacoblee93 (changes requested):
Looks cool! A few questions

langchain/src/tools/google_calendar/prompts/view-events-prompt.ts
ShaunBaker: https://github.com/nearform/langchainjs/blob/main/langchain/src/tools/google_calendar/commands/run-view-events.ts#L21

langchain/src/tools/google_calendar/prompts/create-event-prompt.ts
ShaunBaker: This is passed here: https://github.com/nearform/langchainjs/blob/main/langchain/src/tools/google_calendar/commands/run-create-events.ts#L71

langchain/src/tools/google_calendar/base.ts
jacoblee93: This should accept a `BaseLLM` instance - what if someone needed to use a custom model?

langchain/src/tools/google_calendar/view.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  constructor(fields: GoogleCalendarAgentParams) {
+    super(fields);
+  }
+
+  async _call(query: string) {
+    const auth = await this.getAuth();
+    const model = this.getModel();
+
+    return await runViewEvents(query, {
```
jacoblee93, line 19: Don't need to `return await`

langchain/src/tools/google_calendar/descriptions.ts
jacoblee93: Ah ok, my mistake!

langchain/src/tools/google_calendar/prompts/create-event-prompt.ts
jacoblee93: In this case using an output parser would make sense I think?

jacoblee93 (changes requested):
Getting there! A few small questions/comments.

simoneb:
@jacoblee93 with the exception of the conflicting files, the feedback should be addressed if you wouldn't mind taking another look please.

langchain/src/tools/google_calendar/base.ts
```diff
@@ -0,0 +1,96 @@
...
+      credentials: {
+        clientEmail: getEnvironmentVariable("GOOGLE_CALENDAR_CLIENT_EMAIL"),
+        privateKey: getEnvironmentVariable("GOOGLE_CALENDAR_PRIVATE_KEY"),
+        calendarId: getEnvironmentVariable("GOOGLE_CALENDAR_CALENDAR_ID"),
+      },
+      scopes: [
+        "https://www.googleapis.com/auth/calendar",
+        "https://www.googleapis.com/auth/calendar.events",
+      ],
+      openAIParameters: {
```
jacoblee93, line 44: Your input should not take `openAIParameters`, but an actual BaseLLM instance that you then set as `this.llm`. Someone could use this with e.g. GPT4All or Anthropic. 

Also, I don't think you're actually initializing a model instance anyway? Does this pass your test cases?

jacoblee93 (changes requested):
Can you test this end-to-end as well please?

PR #1778, opened 2023-06-27
fix(memory): throw error if no input/output key is specified and there are multiple keys

joaopcm:
@jopcmelo

This change fixes an issue that causes a lot of confusion when debugging Langchain applications. The issue is that when providing multiple keys to a memory instance, Langchain throws an error saying that an input/output key must be specified. The problem is that when the key conflict is related to the output scenario, the error message only mentions "input".

To fix and keep the code readable, I ended up separating the key-checking function from the objects and created another function called `getOutputValue` with a more descriptive message focused on output values, not input.


joaopcm:
@jacoblee93, could you please review this PR?

joaopcm:
cc @nfcampos

joaopcm:
@jacoblee93 , do you think we can merge this one?

jacoblee93:
Hi, apologies for the delay - wanted to make sure I looked at it very carefully since it's changing a base class.

jacoblee93:
Thank you!

PR #1779, opened 2023-06-27
Ensure that run managers are returned from handleLLMStart/handleChatModelStart always in the same order as the prompts passed in

nfcampos:
What was happening was that the runManager was only added to the managers array after the callback handlers finished executing, so eg if the callback handlers for the second run finished first the runManager for prompt #2 would actually be in position #1 of the returned array.

Fixes # (issue)

### DIFF
--- a/langchain/src/callbacks/manager.ts
+++ b/langchain/src/callbacks/manager.ts
@@ -348,9 +348,7 @@ export class CallbackManager
     _parentRunId: string | undefined = undefined,
     extraParams: Record<string, unknown> | undefined = undefined
   ): Promise<CallbackManagerForLLMRun[]> {
-    const managers: CallbackManagerForLLMRun[] = [];
-
-    await Promise.all(
+    return Promise.all(
       prompts.map(async (prompt) => {
         const runId = uuidv4();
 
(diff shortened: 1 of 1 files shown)


langchain/src/callbacks/manager.ts
```diff
@@ -348,9 +348,7 @@ export class CallbackManager
     _parentRunId: string | undefined = undefined,
     extraParams: Record<string, unknown> | undefined = undefined
   ): Promise<CallbackManagerForLLMRun[]> {
-    const managers: CallbackManagerForLLMRun[] = [];
-
-    await Promise.all(
+    return Promise.all(
```
vowelparrot, line 351: That's tricky. So the callback in `Promise.map()` is executed whenever that promise is actually finished but then the result of it all is still ordered?

nfcampos: the promises in promise.all are all executed simultaneously, all starting at the same time. But promise.all does guarantee that no matter what the order in which they finish the returned array will hold the return values of each in the order they were passed in.

agola11 (approved):

PR #1780, opened 2023-06-27
Use new serialisation format for Chat Messages, align constructor name and args with python

nfcampos:
Fixes # (issue)

### DIFF
--- a/langchain/src/schema/index.ts
+++ b/langchain/src/schema/index.ts
@@ -88,7 +88,7 @@ export abstract class BaseMessage
 
   /**
    * @deprecated
-   * Use {@link BaseMessage.text} instead.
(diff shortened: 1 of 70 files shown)


langchain/src/schema/index.ts
```diff
@@ -65,75 +65,148 @@ export interface StoredMessage {
...
+  extends Serializable
+  implements BaseMessageFields
+{
+  lc_namespace = ["langchain", "schema"];
+
+  lc_serializable = true;
+
+  /**
+   * @deprecated
+   * Use {@link BaseMessage.text} instead.
```
hwchase17, line 91: this should be `BaseMessage.content`

hwchase17 (approved):

PR #1597, opened 2023-06-09
Add LanceDB vectorstore

gsilvestrin:
Adds support to [LanceDB](https://github.com/lancedb/lancedb), an embedded vector database.

It should behave as the [Python Integration](https://sj-langchain.readthedocs.io/en/latest/modules/indexes/vectorstores/examples/lanecdb.html#)

I believe the Vercel failure is unrelated to my change, is caused by langchainplus:

```
[INFO] [en] Creating an optimized production build...
../langchain/src/client/langchainplus.ts:1:46 - error TS2307: Cannot find module 'langchainplus-sdk' or its corresponding type declarations.
1 import { Example, LangChainPlusClient } from "langchainplus-sdk";
                                               ~~~~~~~~~~~~~~~~~~~
../langchain/src/client/langchainplus.ts:175:25 - error TS7006: Parameter 'example' implicitly has an 'any' type.
175     examples.map(async (example) => {
                            ~~~~~~~
../langchain/src/client/tests/langchainplus.int.test.ts:3:37 - error TS2307: Cannot find module 'langchainplus-sdk' or its corresponding type declarations.
3 import { LangChainPlusClient } from "langchainplus-sdk";
                                      ~~~~~~~~~~~~~~~~~~~
../langchain/src/client/tests/langchainplus.int.test.ts:17:22 - error TS7006: Parameter 'd' implicitly has an 'any' type.
17   if (!datasets.map((d) => d.name).includes(datasetName)) {
                        ~
../langchain/src/client/tests/langchainplus.int.test.ts:51:22 - error TS7006: Parameter 'd' implicitly has an 'any' type.
51   if (!datasets.map((d) => d.name).includes(datasetName)) {
                        ~
../langchain/src/client/tests/langchainplus.int.test.ts:77:24 - error TS7006: Parameter 's' implicitly has an 'any' type.
77   expect(sessions.map((s) => s.name)).toContain(sessionName);
                          ~
../langchain/src/client/tests/langchainplus.int.test.ts:102:22 - error TS7006: Parameter 'd' implicitly has an 'any' type.
102   if (!datasets.map((d) => d.name).includes(fileName)) {
                         ~
../langchain/src/client/tests/langchainplus.int.test.ts:143:22 - error TS7006: Parameter 'd' implicitly has an 'any' type.
143   if (!datasets.map((d) => d.name).includes(datasetName)) {
```

PR #1787, opened 2023-06-27
Add LanceDB vectorstore

jacoblee93:
Extends #1597


jacoblee93:
Hey @gsilvestrin the package seems to be failing on MacOS with Node 20 in our CI:

https://github.com/hwchase17/langchainjs/actions/runs/5395642522/jobs/9798968874?pr=1787

Think you could take a look?

jacoblee93:
Ah ok, looks like it's not supported on all platforms (e.g. Windows). Will just remove the unit test later.

PR #1792, opened 2023-06-28
🔧 Custom endpoint for Azure OpenAI (Issue #1272 )

### LINKED ISSUE #1272: How to provide custom endpoint for Azure OpenAi service? (gateway URL)
In my company we use gateway url to access Azure open AI service to do chat completions. It looks like this: https://xxx-xxx-openai.azure-api.net.
From your sourcecode I see that endpoint is set directly in the `completeionWithRetry` method 
image

Is there any way to work with your Azure OpenAI class but having custom endpoint ?

clemenspeters:
Fixes #1272

Adds support for optional `AZURE_OPENAI_BASE_PATH` ENV to allow for custom Azure OpenAI endpoints such as "https://westeurope.api.cognitive.microsoft.com/".

### Unit Test

Running `yarn test:single ./src/util/tests/azure.test.ts --coverage` (from inside the `./langchain` directory)
image

### e2e Test

For the end to end test I ran `yarn build` (from inside the `./langchain` directory) and then switched to the other repository where we use `langchain` as a dependency. From that other project, I first ran `yarn add langchain@latest` and then ran `cp -R /Users/<YOUR_USER>/projects/langchainjs/langchain/dist/ /Users/<YOUR_USER>/projects/<YOUR_PROJECT_DEPENDING_ON_LANGCHAIN>/node_modules/langchain/`. Afterwards running `yarn dev` from that other project allowed me to successfully test my code changes end to end.
I hope that makes sense. Please let me know if there is a better way.

Twitter: @PetersClemens, @conventic_com


langchain/src/chat_models/openai.ts
```diff
@@ -536,9 +544,16 @@ export class ChatOpenAI
...
   ) {
     if (!this.client) {
-      const endpoint = this.azureOpenAIApiKey
-        ? `https://${this.azureOpenAIApiInstanceName}.openai.azure.com/openai/deployments/${this.azureOpenAIApiDeploymentName}`
-        : this.clientConfig.basePath;
+      const openAIEndpointConfig: OpenAIEndpointConfig = {
+        azureOpenAIApiDeploymentName: this.azureOpenAIApiDeploymentName,
+        azureOpenAIApiInstanceName: this.azureOpenAIApiInstanceName,
+        azureOpenAIApiKey: this.azureOpenAIApiKey,
+        azureOpenAiBasePath: this.azureOpenAiBasePath,
```
jacoblee93, line 551: Capitalize `AI` to match convention

langchain/src/util/azure.ts
```diff
@@ -0,0 +1,65 @@
...
+  const {
+    azureOpenAIApiDeploymentName,
+    azureOpenAIApiInstanceName,
+    azureOpenAIApiKey,
+    azureOpenAiBasePath,
+    basePath,
+  } = config;
+
+  if (azureOpenAiBasePath && azureOpenAIApiDeploymentName) {
+    return `${azureOpenAiBasePath}/openai/deployments/${azureOpenAIApiDeploymentName}`;
```
jacoblee93, line 41: At this point would we rather just allow a custom `basePath` parameter to override the Azure case as well (like it does for the classic OpenAI case)?

I could see a world where Azure changes the `/openai/deployments/` path for some enterprise segment in the future.

jacoblee93 (commented):
Thanks for looking into this!

langchain/src/util/azure.ts
clemenspeters: @jacoblee93 thanks for pointing his out. Would you prefer to return `${azureOpenAiBasePath}/${azureOpenAIApiDeploymentName}` or just `${azureOpenAiBasePath}`?

langchain/src/chat_models/openai.ts
clemenspeters: Good point, thanks for spotting this inconsistency 👍

langchain/src/util/azure.ts
jacoblee93: I feel like just overriding the whole thing would make more sesne since that's what `configuration.basePath` does?

clemenspeters: Please suggest some code if further updates are required before merging

jacoblee93: I think the above with just `azureOpenAIBasePath` would make the most sense here if you agree?

langchain/src/util/azure.ts
```diff
@@ -0,0 +1,65 @@
...
+export function getEndpoint(config: OpenAIEndpointConfig): string {
+  const {
+    azureOpenAIApiDeploymentName,
+    azureOpenAIApiInstanceName,
+    azureOpenAIApiKey,
+    azureOpenAIBasePath,
+    basePath,
+  } = config;
+
+  if (azureOpenAIBasePath && azureOpenAIApiDeploymentName) {
```
jacoblee93, line 40: Should this be checking if `azureOpenAIApiKey` is set?

langchain/src/util/azure.ts
clemenspeters: Actually we need different endpoints for the embedding (see https://github.com/hwchase17/langchainjs/blob/main/langchain/src/embeddings/openai.ts#L155) and the llm (see https://github.com/hwchase17/langchainjs/blob/main/langchain/src/llms/openai.ts#L429)

e.g. `AZURE_OPENAI_API_EMBEDDINGS_DEPLOYMENT_NAME='text-embedding-ada-002'` vs `AZURE_OPENAI_API_INSTANCE_NAME='custom-openai-deployment'`

So then we would need at least two variables like `azureOpenAILLMPath` and `azureOpenAIEmbeddingPath` 🤔

langchain/src/util/azure.ts
clemenspeters: yes, makes sense. Thank you @jacoblee93 👍

langchain/src/util/azure.ts
clemenspeters: I decided to go with `return `${azureOpenAIBasePath}/${azureOpenAIApiDeploymentName}`;`, so that `AZURE_OPENAI_API_EMBEDDINGS_DEPLOYMENT_NAME` from https://github.com/hwchase17/langchainjs/blob/4c565af393b4e94d48700334f271e6c6de1eedc0/langchain/src/embeddings/openai.ts#L90 is still used and applied when calling `getEndpoint` in https://github.com/hwchase17/langchainjs/blob/4c565af393b4e94d48700334f271e6c6de1eedc0/langchain/src/embeddings/openai.ts#L169C24-L169C35. I hope that makes in mergable. 😃 🤞
What do you think @jacoblee93 ?

jacoblee93: Ah ok, fine with me then.

jacoblee93:
Looks good, thank you! Will merge with the next release.

jacoblee93:
Hey also sorry this actually breaks the normal OpenAI functionality - think you could try `yarn test:single langchain/src/llms/tests/openai.int.test.ts`?

jacoblee93:
> Hey also sorry this actually breaks the normal OpenAI functionality - think you could try `yarn test:single langchain/src/llms/tests/openai.int.test.ts`?

Couldn't push to your branch but opened #1842 and will merge that.

clemenspeters:
> 



> Hey also sorry this actually breaks the normal OpenAI functionality - think you could try `yarn test:single langchain/src/llms/tests/openai.int.test.ts`?

Running this on the latest `main` branch also fails, so I think it's independent from this PR 🤔 
Am I missing something here?

clemenspeters:
> > Hey also sorry this actually breaks the normal OpenAI functionality - think you could try `yarn test:single langchain/src/llms/tests/openai.int.test.ts`?
> 
> Couldn't push to your branch but opened #1842 and will merge that.

Sry about that @jacoblee93. I pulled in your changes from `custom-endpoint-for-azure-openai` branch and also added you with `Write` permission to this repo (invite pending).

PR #1793, opened 2023-06-28
Add support for FunctionChatMessage in getBufferString()

### LINKED ISSUE #1720: Error in handler Handler, handleLLMStart: Error: Got unsupported message type: [object Object]
I'm seeing this error when a FunctionMessage is added after successful selection of a function by OpenAIAgent.

I believe it's because of this section of code in BaseMemory:'
```
export function getBufferString(messages, humanPrefix = "Human", aiPrefix = "AI") {
    const string_messages = [];
    for (const m of messages) {
        let role;
        if (m._getType() === "human") {
            role = humanPrefix;
        }
        else if (m._getType() === "ai") {
            role = aiPrefix;
        }
        else if (m._getType() === "system") {
            role = "System";
        }
        else if (m._getType() === "generic") {
            role = m.role;
        }
        else {
            throw new Error(`Got unsupported message type: ${m}`);
        }
        string_messages.push(`${role}: ${m.text}`);
    }
    return string_messages.join("\n");
} 
```

Where FunctionMessage has a type of "function", and thus hitting the error section.

axrj:
BaseMemory getBufferString() does not currently support FunctionChatMessage type and throws errors in handleLLMStart handler when using OpenAIAgent or custom `openai-functions` agents.

* Supports function chat messages and  includes the name of the function call in the buffer string

- Fixes #1720


jacoblee93:
Seems reasonable, thank you!

jacoblee93:
Will go live with tomorrow's release.

PR #1796, opened 2023-06-28
Fix mangled names when deserialising minified classes

dqbd:
Code running in edge runtimes is minified, thus the names of the classes are mangled. 

Transparently override the name when reviving via langchain/load

### DIFF
--- a/langchain/src/load/index.ts
+++ b/langchain/src/load/index.ts
@@ -157,7 +157,7 @@ async function reviver(
     // Construct the object
     if (serialized.type === "constructor") {
       // eslint-disable-next-line new-cap, @typescript-eslint/no-explicit-any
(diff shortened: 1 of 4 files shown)


langchain/src/load/index.ts
```diff
@@ -157,13 +157,19 @@ async function reviver(
...
           kwargs as SerializedFields,
           keyFromJson,
           combineAliasesAndInvert(builder)
         )
       );
+
+      // Minification in severless/edge runtimes will mange the
+      // name of classes presented in traces. As the names in import map
+      // are present as-is even with minification, use these names instead
+      instance.lc_name_override = name;
```
nfcampos, line 171: Worth just setting the regular `builder.name` property with Object.defineProperty ?

nfcampos (approved):

PR #1802, opened 2023-06-28
Fix wrongly awaited new in LanceDB example

ksjogo:
The new LanceDB example has an extraneous await, this removes it.


jacoblee93:
Ah, thank you!

PR #1803, opened 2023-06-29
Adds OpenAI functions-powered OpenAPI chain

jacoblee93:
Chain can choose and call specific APIs based on an OpenAPI spec. Also adds OpenAPI spec utility class. Brings parity with Python repo.


langchain/src/chains/openai_functions/openapi.ts
```diff
@@ -0,0 +1,425 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  requestArgs: Record<string, any>,
+  options?: {
+    headers?: Record<string, string>;
+    params?: Record<string, string>;
+  }
+) => Promise<Response>;
+
+function formatURL(url: string, pathParams: Record<string, string>): string {
+  const expectedPathParamNames = [...url.matchAll(/{(.*?)}/g)].map((match) => match[1]);
```
jacoblee93, line 33: `matchAll` is ok right @nfcampos?

mobilerajs:
Q: I'm trying to use Functions with PromptTemplates in the JS/TS port. Will this PR assist w/ that? Any other pointers to sample code would be awesome - thanks.

langchain/src/chains/openai_functions/openapi.ts
omriTelemPanda: @jacoblee93 I have been using: `createOpenAPIChain` to call my open-api given spec I have.
When Im using: `verbose: true`, I see all the chain phases.
Write before calling my API, I see that The chain takes this function argument:
```
[chain/start] [1:chain:SequentialChain > 4:chain:SimpleRequestChain] Entering Chain run with input: {
  "query": "Show me all my organization hidden Users",
  "function": {
    "name": "getUsersByQueryParams",
    "arguments": {
      "params": {
        "hidden": {
          "value": true
        }
      }
    }
  }
}
```

and Unfortunately runs the query with: `{ hidden: { value: true } }`, instead of just: `{ hidden: true }`.

Any suggestion on how to take care of this Issue? thanks 🙏🏼

PR #1805, opened 2023-06-29
S3 Unstructured API key

krftcz:
This PR adds possibility to add API key for UnstructuredLoader in S3Loader as this option was missing.

### Before submitting
✅ I've updated documentation and integration tests.


jacoblee93:
Thank you!

PR #1807, opened 2023-06-29
Update Zapier agent docs

ralewis85:
Updating docs to call out how to attach Zapier NLA auth.  Also removed email address as developers can now self serve.

Question: 

With this PR https://github.com/hwchase17/langchainjs/pull/1798 I updated the Wrapper class params, yet documentation doesn't reflect that change currently https://js.langchain.com/docs/api/tools/classes/ZapierNLAWrapper.  I tried running `yarn docs` but it did not generate any updated documentation.  How do I get this MD file updated?

I should get the class updated to call out params can be set via `params` or the equivalent env var.


ralewis85:
Oh!  I see my mistake.  Wow.  I didn't expect the params to be ordered alphabetically, and simply didn't scroll down far enough to see the oauth access token param 🙃

PR #1784, opened 2023-06-27
Add the Elasticsearch vectorstore

ezimuel:
This PR adds the support of Elasticsearch as vectorstore. It includes unit tests, documentation and examples.

PR #1810, opened 2023-06-29
Add the Elasticsearch vectorstore

jacoblee93:
Extends #1784 with minor docs and build fixes.


ezimuel:
Thanks @jacoblee93

langchain/src/vectorstores/elasticsearch.ts
```diff
@@ -0,0 +1,212 @@
...
+          _index: this.indexName,
+        },
+      },
+      {
+        embedding,
+        metadata: documents[idx].metadata,
+        text: documents[idx].pageContent,
+      },
+    ]);
+    await this.client.bulk({ refresh: true, operations });
```
sound2gd, line 83: this will not be aware to outside if something error happened.
cause they will return something like this

```js
[
 {
    status: 400,
    error: '{\n' +
      '  "type": "mapper_parsing_exception",\n' +
      '  "reason": "failed to parse",\n' +
      '  "caused_by": {\n' +
      '    "type": "illegal_argument_exception",\n' +
      '    "reason": "cannot index vectors with dimension greater than 1024"\n' +
      '  }\n' +
      '}'
  }
]
```

but the result is dropped

PR #1821, opened 2023-06-30
Update openai.ts switch references to replaceAll with replace to work…

RyanThomasMusser:
… with older versions of node

Fixing #: 
https://github.com/hwchase17/langchainjs/issues/1819
https://github.com/hwchase17/langchainjs/issues/471

My twitter account is @TheRyanMusser

Thank you!

### DIFF
--- a/langchain/src/embeddings/openai.ts
+++ b/langchain/src/embeddings/openai.ts
@@ -121,7 +121,7 @@ export class OpenAIEmbeddings
 
   async embedDocuments(texts: string[]): Promise<number[][]> {
     const subPrompts = chunkArray(
-      this.stripNewLines ? texts.map((t) => t.replaceAll("\n", " ")) : texts,
+      this.stripNewLines ? texts.map((t) => t.replace(/\n/g, " ")) : texts,
       this.batchSize
     );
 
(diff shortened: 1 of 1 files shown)


jacoblee93:
Thanks for the contribution - LangChainJS only supports versions down to Node 18 though: https://github.com/hwchase17/langchainjs#supported-environments

jacoblee93:
Never mind, read the context in the issue. 

Still not sure this is a rabbit hole we want to go down - there are definitely many places where we use methods that aren't in older versions of Node - but will leave it open for now.

langchain/src/embeddings/openai.ts
```diff
@@ -144,7 +144,7 @@ export class OpenAIEmbeddings
   async embedQuery(text: string): Promise<number[]> {
     const { data } = await this.embeddingWithRetry({
       model: this.modelName,
-      input: this.stripNewLines ? text.replaceAll("\n", " ") : text,
+      input: this.stripNewLines ? text.replace(/\n/g, " ") : text,
```
jacoblee93, line 147: This should be `\\n` no?

jacoblee93:
Thank you! This is a one-off and that we don't have plans to extend official support to lower versions, but it's a simple enough change.

PR #1822, opened 2023-06-30
Update error message for Zapier tool when listing endpoints fails

ralewis85:
This PR mirrors the update we did in the `langchain` repository.  The list endpoint is always used when the wrapper class is initiated and run.  This PR makes 401 error messages more clear to the end user to double check their key/token and make sure it's refreshed if needed.


jacoblee93:
Thank you!

PR #1824, opened 2023-06-30
Added empty filter function and tests

### LINKED ISSUE #1823: Self Query Retriever throws Error if LLM not able to parse filter from user query.
### Issue
While using `SelfQueryRetriever` if you pass a random which cant be converted to filter by LLM then it throws error.
**I am using some random text query like "wowowow" or "what is what".**

Screenshot 2023-06-30 at 11 50 59 PM

siddarthvader:
### Chanages
`visitStructuredQuery` was throwing error if `fillter` function was `undefined` or `NO_FILTER`. `filter` function can be `undefined` or `NO_FILTER` for few such cases listed below.

- When LLM is not able to parse the filter from natural language query passed by human
- LLM times out/Network Error (??)

So I added empty object `{}` as return of `visitStructuredQuery` when `filter` function is `undefined` or `NO_FILTER`, this does not affect the result of supabase pg query.

Fixes #1823


ecmel:
The same is true for HNSWLib Self Query Retriever example.

PR #1826, opened 2023-06-30
Feat/recursive similarity search

### LINKED ISSUE #1827: Recursive Similarity Search for Unknown `﻿k` Value
**Issue Summary:**
When performing a similarity search, users often encounter the challenge of determining the appropriate ﻿`k` value for retrieving N similar results. However, in scenarios where the `﻿k` value is unknown and users want to retrieve all possible results, the current system falls short. This issue aims to address this limitation by introducing a new feature called Recursive Similarity Search.

**Issue Details:**
In certain situations, such as when querying the features of a product described in a comprehensive document, it becomes difficult to determine the correct `﻿k` value that will retrieve all features. To overcome this challenge, Langchain has developed the **Recursive Similarity Search** feature. This feature allows users to perform a similarity search without relying solely on the `﻿k` value. Instead, the system will return all possible results based on a user-defined minimum similarity percentage.

To implement the Recursive Similarity Search, you need to utilize a vector store as a retriever. The example provided demonstrates the usage of the feature using the Langchain library.

**Example Code:**
The example code showcases the usage of Recursive Similarity Search using Langchain library. It demonstrates the creation of a vector store, configuration of the retrieval parameters, and executing a query to retrieve all the features of a product.

ray-so-export (1)

**Expected Outcome:**
Upon using Recursive Similarity Search, the system is expected to return all relevant results based on the specified minimum similarity percentage. Users will receive a response that includes the text describing the features and the source documents associated with each feature.
Additional Information:
	•	The Recursive Similarity Search feature utilizes a dynamic `K` value, which increases progressively until no more results can be found.
	•	The `maxK` value can be specified to prevent exceeding token limits.
	•	Users should consider the chunk size and token limitations when setting the `maxK` value.
	•	The Langchain library provides classes and methods for implementing Recursive Similarity Search, such as ﻿MemoryVectorStore, ﻿OpenAIEmbeddings, ﻿ChatOpenAI, ﻿ConversationalRetrievalQAChain, and ﻿BufferMemory.
	•	Refer to the code example provided for implementing Recursive Similarity Search in your application.

joaopcm:
@jopcmelo

Fixes #1827


jacoblee93:
I like the idea but I think this is a bit too specific of a use case to add to the base `VectorStoreRetriever` class.

Would rather add a `ScoreThresholdVectorStoreRetriever` (or some other name?) that wraps the default and updates it. Then the developer could do something like this:

```javascript
    ConversationalRetrievalQAChain.fromLLM(
      model,
      new ScoreThresholdVectorStoreRetriever(vectorStore),
      {
        returnSourceDocuments: true,
        memory: new BufferMemory({
          memoryKey: "chat_history",
          inputKey: "question",
          outputKey: "text",
        }),
      }
    )
```

Would that make sense to you?

joaopcm:
@jacoblee93 , I love the idea! It's already implemented :)

Please, take a look at it once you get a bit of time ;)

langchain/src/vectorstores/similarity-score-threshold.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  private isAboveMinScore([, score]: [Document, number]): boolean {
+    return score >= this.minSimilarityScore;
+  }
+
+  fromVectorStore(vectorStoreRetriever: VectorStoreRetriever<VectorStore>) {
+    const isAboveMinScore = this.isAboveMinScore.bind(this);
+
+    // eslint-disable-next-line no-param-reassign
+    vectorStoreRetriever.getRelevantDocuments = async (query: string) => {
```
jacoblee93, line 35: Overriding a method like this can be dangerous - your retriever should instead extend `VectorStoreRetriever` and override the `getRelevantDocuments` and `addDocuments` methods. The constructor should take a `vectorStoreRetriever` and call that method in the overridden methods.

The issue with your current approach is potential side effects. If someone were to try to use the `vectorStoreRetriever` after calling this method, this updated implementation would be used rather than the expected one.

LMK if that doesn't make sense and I can help rework this.

langchain/src/vectorstores/similarity-score-threshold.ts
```diff
@@ -0,0 +1,72 @@
...
+
+        const filteredResults = results.filter(isAboveMinScore);
+
+        return filteredResults.map(([document]) => document);
+      }
+
+      let updatedK = vectorStoreRetriever.k;
+
+      // eslint-disable-next-line no-constant-condition
+      while (true) {
```
jacoblee93, line 52: Let's see if we can get this looking something more like:
```
while (filteredResults.length > updatedK && updatedK < this.maxK) {}
```

langchain/src/vectorstores/similarity-score-threshold.ts
```diff
@@ -0,0 +1,72 @@
+import { VectorStore, VectorStoreRetriever } from "./base.js";
```
jacoblee93, line 1: Filename convention for `src` files is uses underscores:
`similarity_score_threshold.ts`

langchain/src/vectorstores/similarity-score-threshold.ts
```diff
@@ -0,0 +1,72 @@
...
+    this.dynamicK = similarityScoreFilters.dynamicK ?? this.dynamicK;
+    this.kIncrement = similarityScoreFilters.kIncrement ?? this.kIncrement;
+    this.maxK = similarityScoreFilters.maxK ?? this.maxK;
+  }
+
+  private isAboveMinScore([, score]: [Document, number]): boolean {
+    return score >= this.minSimilarityScore;
+  }
+
+  fromVectorStore(vectorStoreRetriever: VectorStoreRetriever<VectorStore>) {
```
jacoblee93, line 31: This can just take a `VectorStore` and call `.asRetriever()`.

jacoblee93 (changes requested):
Thanks, a few comments!

langchain/src/vectorstores/similarity-score-threshold.ts
joaopcm: I don't think so. Users can call `.asRetriever()` providing the initial `k ` value and the vector store filters in an object.

Do you think we should receive these props and pass them to it, or should we keep it as is?

langchain/src/vectorstores/similarity-score-threshold.ts
joaopcm: I understand the issue this can cause, but I must admit I don't quite understand how to fix it... Do you have an example of the code you're expecting?

langchain/src/vectorstores/similarity-score-threshold.ts
joaopcm: Done :)

langchain/src/vectorstores/similarity-score-threshold.ts
joaopcm: Done :)

jacoblee93:
Refactored a bit, but am now rethinking whether this is a good idea - do any vector store providers document similarity score?

joaopcm:
Yes, Pinecone does.
Regards,
João Melo.

joaopcm:
> Refactored a bit, but am now rethinking whether this is a good idea - do any vector store providers document similarity score?
> 
> Am worried that it's bad to guide people to rely on this since providers could theoretically change this anytime, and if you _really_ wanted to do this you could just set a high `k` and do the filtering yourself.

@jacoblee93, yes, LangChain users can def filter based on the score by themselves. But how could it be done when using chains that expect a retriever such as `ConversationalRetrievalQAChain`? The chain runs the query by itself, blocking us to loop through the documents found and filtering based on a score. This is the reason why I've created this suggestion :)

What do you think?

PR #1829, opened 2023-06-30
fix: add prompt schema error detail

iloveitaly:
Includes the originally thrown exception's error message in the new error.

Without this, you have to inspect your prompt template and determine what went wrong.


jacoblee93:
Thank you! Can you please run `yarn lint` and fix the warnings?

Issue #1536, opened 2023-06-04
Remove Documents/Embeddings from Vector Store


naveensky:
I am trying to use Postgres as a vector store for my embeddings. The docs are at https://js.langchain.com/docs/modules/indexes/vector_stores/integrations/typeorm

I am wondering if there is a way to remove embeddings/documents from the store which are no longer valid and should not be included in the similarity search.

dkarlovi:
I do this with Opensearch one by deleting all OS index documents which are related to the document I'm embedding. I do it manually before ingesting the new one.

borel:
looks like there is nothing ready to use directly in `TypeORMVectorStore` but it looks quite easy to do .

if i get some time , i will implement a deletion by metadata or something like that

PR #1830, opened 2023-06-30
Support delete vectorstore operations

jacoblee93:
Addresses #1536 and other issues.

Also adds them to Chroma as a proof of concept.

In Python, `ids` are passed separately into an `addTexts` method. The analogue in JS is `addVectors`, and the other option to pass ids in when adding documents would be changing the `addVectors` method signature to:

```js
  abstract addVectors(
    vectors: number[][],
    documents: Document[],
    ids?: string[]
  ): Promise<void>;
```

~~But this would also require `ids` to be passed around into the convenience `addDocuments` method and a bunch of other places. My sense is that `id` is common enough among different vector store implementations to be worth adding to the base class.~~ Opted to add an additional "options" object that vector stores accept in the `addDocuments` and `addVectors` implementation to avoid base class drift.


jacoblee93:
Holding off on base class changes, switching to just pass ids through.

PR #1831, opened 2023-06-30
Fixes Notion inline databases in nested designs not loading and improved output order of received blocks

skarard:
This PR address an issue I came across of Notion inline databases not loading correctly when they are nested inside columns or callouts. 

I have also improved the output order of the received blocks, this improves context when used with a retrieval chain.

What Changed:
- I added a custom blocks loader, in place of the one supplied with the `notion-to-md` package.

All changes a backwards compatible, if a user has loaded a notion document I would suggest reloading it for better results.


langchain/src/document_loaders/web/notionapi.ts
```diff
@@ -114,55 +122,107 @@ export class NotionAPILoader extends BaseDocumentLoader {
...
-      pageData,
-      getBlockChildren(this.notionClient, pageId, null),
-    ]);
+  private async loadBlocks(
+    blocksResponse: ListBlockChildrenResponseResults
+  ): Promise<{ mdBlocks: MdBlock[]; childDocuments: Document[] }> {
+    const blocks = blocksResponse.filter(isFullBlock);
 
-    const [childPageDocuments, childDatabaseDocuments, mdBlocks] =
+    const [childPageDocuments, childDatabaseDocuments, blocksDocsArray] =
```
jacoblee93, line 139: This looks like potentially a lot of requests - any concerns about rate limits?

langchain/src/document_loaders/web/notionapi.ts
```diff
@@ -114,55 +122,107 @@ export class NotionAPILoader extends BaseDocumentLoader {
...
+        ),
+        Promise.all(
+          blocks
             .filter(
-              (block) =>
-                "type" in block && block.type.includes("child_database")
+              (block) => !["child_page", "child_database"].includes(block.type)
+            )
+            .map((block) =>
+              (async () => {
```
jacoblee93, line 157: For consistency should we factor this into another method too? Or at least something that just returns the `async function` directly rather than calling it. I think it's fine but the IIFE isn't necessary here.

langchain/src/document_loaders/web/notionapi.ts
```diff
@@ -1,11 +1,20 @@
...
+  iteratePaginatedAPI,
+} from "@notionhq/client";
+import {
+  BlockObjectResponse,
   PageObjectResponse,
 } from "@notionhq/client/build/src/api-endpoints.js";
 import { NotionToMarkdown } from "notion-to-md";
 import { getBlockChildren } from "notion-to-md/build/utils/notion.js";
 
+import {
```
jacoblee93, line 14: If you're just using the types, let's actually use `import type`.

Are these not exported directly from `notion-to-md`?

langchain/src/document_loaders/web/notionapi.ts
```diff
@@ -1,11 +1,20 @@
-import { Client, isFullPage, iteratePaginatedAPI } from "@notionhq/client";
 import {
-  GetPageResponse,
+  Client,
+  isFullBlock,
+  isFullPage,
+  iteratePaginatedAPI,
+} from "@notionhq/client";
+import {
```
jacoblee93, line 7: See below.

langchain/src/document_loaders/web/notionapi.ts
```diff
@@ -114,55 +122,107 @@ export class NotionAPILoader extends BaseDocumentLoader {
...
-              (block) =>
-                "type" in block && block.type.includes("child_database")
+              (block) => !["child_page", "child_database"].includes(block.type)
+            )
+            .map((block) =>
+              (async () => {
+                const mdBlock = await this.loadBlock(block);
+                let childDocuments: Document[] = [];
+
+                if (block.has_children) {
```
jacoblee93, line 161: I feel like this logic should be recursive within `loadBlock`?

jacoblee93 (dismissed):
Thank you, left a few comments!

langchain/src/document_loaders/web/notionapi.ts
skarard: Yes, there is a concern about rate limits with heavily nested blocks. Though I haven't hit the rate limits yet, as this is a burst it appear to scrape by...

I have realised I need to re-architecture this to deal with rate limiting the requests.

The revised architecture would be:
- Exclusively load blocks with `loadBlock(block)` and when it finds a child database or child page add their IDs to a queue.
- Processing of the queue of pages.
- Use a scheduler (like `bottleneck`) for all notionAPI requests (which will allow for retries).

I am planning on tackling this soon. I use this document loader a lot, so I am keen on improving performance.

langchain/src/document_loaders/web/notionapi.ts
skarard: This is the main processing loop of `loadBlocks(blocks)` I have refactored it to remove the IIFE.

langchain/src/document_loaders/web/notionapi.ts
skarard: No, unfortunately notion-to-md only exports the NotionToMarkdown class, I had a go at TypeFu to get infer types from the class methods but it was a lot more messy than importing them from the build folder, though I did remove the redundant `index.js`

skarard: I might send them a PR to export the types and `getBlockChildren` directly from notion-to-md

langchain/src/document_loaders/web/notionapi.ts
skarard: I did manage to infer the types from the TypeGuards provided by the `@notionhq/client`.

langchain/src/document_loaders/web/notionapi.ts
skarard: Yeah, I agree.

The problem is that this will require a large refactor because of the way it loads `MdBlock[]` and `Document[]`.

I renamed `loadBlocks(blocks)` to `loadBlocksAndDocs(blocks)`. Hopefully that clears up any confusion for now.

skarard:
> Thank you, left a few comments!

I've responded, and I'll let you decide whether the new commits and my responses resolve your comments. 

The main concern for me is the rate limiting and my response is:
> I have realised I need to re-architecture this to deal with rate limiting the requests.
> 
> The revised architecture would be:
> 
> Exclusively load blocks with loadBlock(block) and when it finds a child database or child page add their IDs to a queue.
> Processing of the queue of pages.
> Use a scheduler (like bottleneck) for all notionAPI requests (which will allow for retries).
> I am planning on tackling this soon. I use this document loader a lot, so I am keen on improving performance.

jacoblee93:
Thank you!

PR #1832, opened 2023-06-30
fix: make the api error message more clear

iloveitaly:
The way it's written now it's not clear that this an "or" check. It sounds like
the package may be configured to use the azure version of openai instead of standard
openai. I think this wording makes it a bit more clear that langchain checked for both API
keys and didn't find any.


jacoblee93:
I like it, thank you!

PR #1833, opened 2023-06-30
docs: note about complex schema

iloveitaly:
it was unclear to me that the basic schema parser did not support things like arrays of strings. This makes it a bit more clear.


jacoblee93:
Thanks!

PR #1835, opened 2023-07-01
Update openai_functions_extraction.ts

davidyu37:
Fixing small mistakes in the doc for extraction

Fixes # (issue)


jacoblee93:
Thanks for the PR! However, `.run` just takes one input, your change would be appropriate for `.call`: https://github.com/hwchase17/langchainjs/blob/main/langchain/src/chains/base.ts#L89

PR #1837, opened 2023-07-01
Rename Databerry to Chaindesk

gmpetrov:
Fixes # (issue)

### DIFF
--- a/langchain/package.json
+++ b/langchain/package.json
@@ -1,6 +1,6 @@
 {
   "name": "langchain",
-  "version": "0.0.156",
+  "version": "0.0.157",
   "description": "Typescript bindings for langchain",
   "type": "module",
   "engines": {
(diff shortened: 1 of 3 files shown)


jacoblee93:
Not sure how many folks are currently in prod with this but would it make sense to keep `DataberryRetriever` as a deprecated alias?

gmpetrov:
> Not sure how many folks are currently in prod with this but would it make sense to keep `DataberryRetriever` as a deprecated alias?

Safe idea @jacoblee93! Will add back the DataberryRetriever class . Is there a way to mark the class as deprecated?

langchain/scripts/create-entrypoints.js
```diff
@@ -138,7 +138,7 @@ const entrypoints = {
   "retrievers/supabase": "retrievers/supabase",
   "retrievers/zep": "retrievers/zep",
   "retrievers/metal": "retrievers/metal",
-  "retrievers/databerry": "retrievers/databerry",
```
jacoblee93, line 141: You'll need to keep the entrypoint in for `databerry` if you want back-compat (up to you!)

jacoblee93:
> > Not sure how many folks are currently in prod with this but would it make sense to keep `DataberryRetriever` as a deprecated alias?
> 
> Safe idea @jacoblee93! Will add back the DataberryRetriever class . Is there a way to mark the class as deprecated?

Just don't forget to put the entrypoint back in `create-entrypoints.js` as well then!

Have been doing stuff like this: 

https://github.com/hwchase17/langchainjs/blob/main/langchain/src/vectorstores/mongo.ts#L21

PR #1839, opened 2023-07-02
Add Searxng Search wrapper as built-in tool

iamrony777:
feat(searxng_search.ts): add new tool `SearxngSearch` to perform search using the Searxng API

The `SearxngSearch` tool is added to perform meta search engine queries using the Searxng API. This tool is useful for answering questions about current events. The input should be a search query, and the output is a JSON array of the query results.

The `SearxngSearch` tool includes the following features:
- Support for specifying the number of results included in the output
- Support for specifying the active search categories and engines
- Support for specifying the language, search page number, and time range of search
- Support for specifying the output format of results
- Support for opening search results on a new tab and proxying image results through SearXNG
- Support for autocomplete and safe search filtering

The `SearxngSearch` tool requires the following environment variable to be set:
- `SEARXNG_API_BASE`: Base URL of the Searxng instance

The `SearxngSearch` tool can be instantiated with custom parameters and headers. If using a self-hosted instance, no custom headers are needed.

The `_call` method of the `SearxngSearch` tool performs the actual API call to the Searxng instance. It sends a POST request to the search endpoint with the specified query and parameters. The response is parsed and processed to extract the relevant information from the query results, answers, infoboxes, and suggestions.

If no good results are found, the method returns the message "No good results found." If query results are found, the method returns a JSON array of the results. If answers are found, the method returns the first answer. If infoboxes are found, the method returns the content of the first infobox without HTML tags. If suggestions are found, the method returns a string with the suggestions.


iamrony777:
Hope it gets merged soon.

iamrony777:
Can someone merge this?

examples/src/tools/searxng_search.ts
```diff
@@ -0,0 +1,41 @@
+import { SearxngSearch } from "langchain/tools";
+import { ChatOpenAI } from "langchain/chat_models/openai";
+import { initializeAgentExecutorWithOptions } from "langchain/agents";
+
+export async function run() {
+  // Works best with  `agentType: "structured-chat-zero-shot-react-description"`
```
jacoblee93, line 6: Let's remove this comment - a new agent could come out.

langchain/src/tools/searxng_search.ts
```diff
@@ -0,0 +1,222 @@
...
+   *
+   * Time range of search for engines which support it. See if an engine supports time range search in the preferences page of an instance.
+   */
+  time_range?: number;
+
+  /**
+   * Throws Error if format is set anything other than "json"
+   * Output format of results. Format needs to be activated in search:
+   */
+  format?: "json" | "csv" | "rss";
```
jacoblee93, line 70: Let's just fix this to `"json"` and not make it configurable?

langchain/src/tools/searxng_search.ts
```diff
@@ -0,0 +1,222 @@
...
+interface SearxngCustomHeaders {
+  [key: string]: string;
+}
+
+interface SearxngSearchParams {
+  /**
+   * @default 10
+   * Number of results included in results
+   */
+  num_results?: number;
```
jacoblee93, line 44: Prefer camel case in JS rather than snake case

jacoblee93 (changes requested):
Looks good, a few small comments!

Can you also add this tool here?

https://js.langchain.com/docs/modules/agents/tools/integrations/

jacoblee93 (approved):

PR #1845, opened 2023-07-03
Collect llm new token events in tracer

nfcampos:
Fixes # (issue)


agola11 (approved):
lgtm, would be good to add a test case

PR #1859, opened 2023-07-04
New Feat: Added Embedding Callbacks

ppramesi:
Added new callbacks:

handleEmbeddingStart: called before embedding is called. It passes serialized embedding class instance and array of texts to the callback function. It passes an array of texts because embedDocuments accepts multiple documents at once. As such, it also passes the single piece of text from embedQuery as an array with length 1.
handleEmbeddingEnd: called after embedding. It passes an array of vectors to the callback function.
handleEmbeddingError: called if there's an error during embed call.

Moreover, there are also some changes that needed to be made to implement these:

Changed Embeddings class such that now child classes implements _embedDocuments and _embedQuery functions since embedDocuments and embedQuery calls handle the callback calls.

Embedding classes are serializable, and added their secrets and aliases.

Moved BaseRetriever into its own file from schema to prevent circular dependency.

Added unit tests for the new callback.

NOTE: the run_type property in run object in the handleEmbeddingStart is incorrect as it should be `"run_type": "embedding"`, but the RunType type in the langchainplus-sdk package (from npm) does not have embedding type and instead have 2 llm types. I assume this is a mistake so it should be changed later.


langchain/src/callbacks/handlers/tracer.ts
```diff
@@ -395,6 +460,15 @@ export abstract class BaseTracer extends BaseCallbackHandler {
...
 
+  onEmbeddingStart?(run: Run): void | Promise<void>;
+
+  onEmbeddingEnd?(run: Run): void | Promise<void>;
+
+  onEmbeddingError?(run: Run): void | Promise<void>;
+
+  // TODO Implement handleAgentEnd, handleText
+
+  // onAgentEnd?(run: ChainRun): void | Promise<void>;
```
nfcampos, line 471: Remove these comments, this has been done

langchain/src/retrievers/base.ts
```diff
@@ -0,0 +1,12 @@
+import { Callbacks } from "../callbacks/manager.js";
```
nfcampos, line 1: We actually would want this file to be moved to `schema/retriever.ts`

jacoblee93:
Looks solid to me - I know we're waiting on some other things but excited to have this merged!

ppramesi:
I see that there's a new commit that changes how callback manager works (e.g. adding metadata, use BaseCallbackConfig, etc). Should I merge those changes to this pull request or should I wait?

hinthornw:
> I see that there's a new commit that changes how callback manager works (e.g. adding metadata, use BaseCallbackConfig, etc). Should I merge those changes to this pull request or should I wait?

Yupp good to rebase!

langchain/src/callbacks/base.ts
```diff
@@ -187,6 +188,40 @@ abstract class BaseCallbackHandlerMethodsClass {
...
+    tags?: string[],
+    metadata?: Record<string, unknown>
+  ): Promise<void> | void;
+
+  /**
+   * Called after embedding is completed, before it exits.
+   * with embeddings vector and the run ID
+   */
+  handleEmbeddingEnd?(
+    vector: number[],
```
hinthornw, line 221: @nfcampos This should be `number[][]`, right? (can generate multiple embeddings per query)

hinthornw, line 221: If `start` accepts multiple strings, the `end` handler should accept a list of vectors I'd assume. @agola11 think we should do the runID looping outside the handler though, right?

ppramesi: The function handleEmbeddingStart is designed to process multiple documents during a single embedDocuments call. However, each subsequent result from this process is individually fed into handleEmbeddingEnd calls. This structure was implemented to mirror the behavior of handleLLMStart and handleLLMEnd. In those functions, handleLLMStart deals with several prompts simultaneously, but handleLLMEnd processes each resulting output one at a time. This consistency in design was suggested by @nfcampos.

hinthornw: Right! Yes yes thanks for clarifying

the-powerpointer:
@ppramesi Would it be possible to include not only the vector into the callback payload, but also the token usage? If in doubt, just the complete API response?
I'm only working with OpenAI models at the moment, there the token usage is returned with the API call. I don't know if other embedding models have this too.
However, it would be nice if we could get from the API call not only the pure result, but also the token usage. Having this transported via the callbacks allows the embedding models methods to stay streamlined with their signature, only retrieving the vectors.

ppramesi:
@the-powerpointer hmmm I'll have to take a look at the code first. I'll get back to you as soon as I can.

matheuscorreiacardoso (approved):

mprytoluk:
Any news on this? Tracking embeddings token usage is very important, and this PR seems like the viable way to it.

jacoblee93:
Still delayed unfortunately - will reraise internally!

techwizzdom:
I was just investigating the usage of embedding models, happy to found this PR. Hope you'll merge it soon, this is much needed 🙏🏼

DoKoB0512:
Would be great if this can be merged. We are currently keep track of each embedding done by different models so if we can somehow get the usage from the vendor itself then it would be amazing.

msalafia:
Waiting this feature si much. But why token usage is not included in the embedding result and just the vectors are returned?

christian-bromann:
thanks for your contribution 🙏 I've just started an am picking up PRs that are left behind. It seems there is a general community interest in this. I will take this PR and discuss it with the team. Would anyone be open help use rebase these changes on top of current `main` or better `v1` branch?

hntrl:
Closing this for inactivity, but like @christian-bromann feel free to rebase/reopen if there's a lot of interest in this.

PR #1869, opened 2023-07-05
UnstructuredLoader: add timeout option

### LINKED ISSUE #1856: Unstructured: headers timeout
I'm using Unstructured loader for parsing a OCR-heavy pdf with

```ts
const options = {
        apiKey: 'MY_API_KEY',
        apiUrl: 'http://localhost:8800/general/v0/general',
        strategy: 'ocr_only',
    };

const loader = new UnstructuredLoader(
        filePath,
        options,
    );

const text = await loader.load();
 ```

The text extraction takes several minutes (>5'), so that the internal node `fetch`/`undici` call fails with

```
TypeError: fetch failed
    at Object.fetch (node:internal/deps/undici/undici:11457:11)
    at async UnstructuredLoader._partition (file:///D:/my_project/node_modules/langchain/dist/document_loaders/fs/unstructured.js:81:26)
    at async UnstructuredLoader.load (file:///D:/my_project/node_modules/langchain/dist/document_loaders/fs/unstructured.js:99:26)
    at async loadUnstructured (file:///D:/my_project/my_code.js:68:18)
    at async file:///D:/my_project/my_code.js:107:18 {
  cause: HeadersTimeoutError: Headers Timeout Error
      at Timeout.onParserTimeout [as callback] (D:\my_project\node_modules\undici\lib\client.js:931:28)
      at Timeout.onTimeout [as _onTimeout] (D:\my_project\node_modules\undici\lib\timers.js:20:13)
      at listOnTimeout (node:internal/timers:569:17)
      at process.processTimers (node:internal/timers:512:7) {
    code: 'UND_ERR_HEADERS_TIMEOUT'
  }
```

Given that a long extraction time may totally be possible in this environment, I'd like to pass a timeout option to the loader (and to the underlying `fetch`/`undici`). Should I add a PR for that or are there other options?

Thanks

### PR DESCRIPTION
**Pull request by thundo**
Fixes #1856 

Adds a per-request timeout option to the `UnstructuredLoader`, allowing a timeout different from 5 minutes to be passed to the constructor.

### DIFF
--- a/langchain/package.json
+++ b/langchain/package.json
@@ -1,6 +1,6 @@
 {
   "name": "langchain",
-  "version": "0.0.102",
+  "version": "0.0.103",
   "description": "Typescript bindings for langchain",
   "type": "module",
   "engines": {

### CODE REVIEW

File: langchain/package.json [hunk 1]
```diff
@@ -791,6 +791,7 @@
     "openapi-types": "^12.1.3",
     "p-queue": "^6.6.2",
     "p-retry": "4",
+    "undici": "^5.22.1",
```
jacoblee93, line 794: Can we do this without the additional dependency?

jacoblee93 (changes requested):
Thanks for submitting this - would really like to avoid adding an additional dependency. Can we do this in base fetch?

thundo:
> Thanks for submitting this - would really like to avoid adding an additional dependency. Can we do this in base fetch?

Not that I'm aware of. Since Node18 node `fetch` is internally implemented with `undici` (mantained more or less by the same node core team) and for the dispatcher you need to instantiate at least an `Agent`, which is an exported `undici` class.

See this [discussion](https://github.com/nodejs/undici/issues/1373) for reference.

jacoblee93:
Unfortunately in this case we need to support more than Nodr (Edge functions, CF Workers, Deno).

thundo:
> Unfortunately in this case we need to support more than Nodr (Edge functions, CF Workers, Deno).

Totally missed the portability concerns... I'll scout about it and report back.

Issue #1870, opened 2023-07-05
createOpenAPIChain: Consistently incorrect payloads, incorrect values, some params can't be empty when provided


danny-avila:
First off, very thankful for this addition. I've tried many other methods with the JS library and they pale in comparison to python solutions, and this does indeed bring some parity.

While the examples for this new chain all work for me, I wanted to use it with an openAPI spec I personally find useful.

The spec I'm using can be found here: https://scholar-ai.net/openapi.yaml

There are a couple of issues with this payload, and I've highlighted them in comments
```js

  "function": {
    "name": "searchAbstracts",
    "arguments": {
      "params": {
        "keywords": { // "keywords" is the only required param
          "text": "AI VR" 
        },
        "sort": {
          "by": "relevance", // Valid values are cited_by_count or publication_date.
          "reverse": false
        },
        "query": {}, // isn't required, even when spec says otherwise. can be passed an empty value, so no issue
        "peer_reviewed_only": true,
        "start_year": {}, // can't be passed an empty value: must be a 4-digit year
        "end_year": {}, // can't be passed an empty value: must be a 4-digit year
        "offset": {} // can't be passed an empty value: Must be an integer
      }
    }
  }
```
I would like to contribute to this and here are some things I noticed. 

Since some parameters need to be specific and not empty when provided, perhaps it's best to filter them completely if they are empty? 

I made a change to SimpleRequestChain to do exactly this and now some of the API calls it generates are working, will need to test with other specs if this creates issues.

However, there are some issues with the schema not being correct. In  https://github.com/hwchase17/langchainjs/issues/1849#issuecomment-1618810667, @jacoblee93 mentioned that we could add an output checking step that retries if the schema doesn't match, which I believe may help. I will add that it should filter any params if they are invalid, when other params are valid.

#1860 will be helpful if it's an improvement over the OutputFunctionsParser as I understand

Thank you for reading, this is an awesome addition to langchain

---

my test code 
```js
  const chain = await createOpenAPIChain(spec, { verbose: true });
  const result = await chain.run('Can you find and explain some articles about the intersection of AI and VR?');
  console.log('api chain run result', result);
  return result;
  ```

jacoblee93:
Thank you - going to try to address this and other raised issues today!

jacoblee93:
[...]

PR #1873, opened 2023-07-05
fix(openapi.ts): filter out empty params before passing them to the request method

danny-avila:
Partially addresses #1870 

It's better to design APIs to only require parameters that are actually used, so this should not really create issues. Some API, as highlighted in my issue, will try to consume a provided query if it's empty, and will throw a 400 status code if the param is not an expected value

### DIFF
--- a/langchain/src/chains/openai_functions/openapi.ts
+++ b/langchain/src/chains/openai_functions/openapi.ts
@@ -352,7 +352,14 @@ class SimpleRequestChain extends BaseChain {
   ): Promise<ChainValues> {
     const inputKeyValue = values[this.inputKey];
     const methodName = inputKeyValue.name;
-    const args = inputKeyValue.arguments;
+    let { params } = inputKeyValue.arguments;
+    params = Object.keys(params).reduce((acc, key) => {
+      if (Object.keys(params[key]).length !== 0) {
+        acc[key] = params[key];
+      }
+      return acc;
+    }, {} as Record<string, string>);
+    const args = { ...inputKeyValue.arguments, params };
     const response = await this.requestMethod(methodName, args);
     let output;
     if (response.status < 200 || response.status > 299) {


jacoblee93:
Thanks for this, makes sense. I think the linter is going to complain though - can you run `yarn format` and `yarn lint`?

danny-avila:
> Thanks for this, makes sense. I think the linter is going to complain though - can you run `yarn format` and `yarn lint`?

My mistake, I successfully ran both commands and got rid of whitespaces.

langchain/src/chains/openai_functions/openapi.ts
```diff
@@ -352,7 +352,14 @@ class SimpleRequestChain extends BaseChain {
   ): Promise<ChainValues> {
     const inputKeyValue = values[this.inputKey];
     const methodName = inputKeyValue.name;
-    const args = inputKeyValue.arguments;
+    let { params } = inputKeyValue.arguments;
+    params = Object.keys(params).reduce((acc, key) => {
+      if (Object.keys(params[key]).length !== 0) {
```
jacoblee93, line 357: Shouldn't `params[key]` be a non-object/array in most cases?

jacoblee93:
Hey @danny-avila thanks for flagging this - I added something for this in #1879 since there was another issue along with it.

It's still not picking up the sort parameter description super well, but you can check out this for an example of how you could hard-code it for now: https://github.com/hwchase17/langchainjs/pull/1879/files#diff-8afac4804a91b5b650b0785978047bd91a0b097a960b6f93e320adb34618f020R148

jacoblee93:
Well ideally the spec should just use something like this: 
https://swagger.io/docs/specification/data-models/enums/

danny-avila:
> Well ideally the spec should just use something like this: https://swagger.io/docs/specification/data-models/enums/

I edited the spec and I tried this but unfortunately it still didn't work. Passing params to the chain is a nice workaround that will have to do for now

PR #1880, opened 2023-07-07
Migrate HumanChatMessage to HumanMessage in Baidu Wenxin

sunner:
The migration is related with examples and tests


jacoblee93:
Thank you!

PR #1890, opened 2023-07-07
Update guide-chat.mdx

charl-kruger:
Correcting Message types in tutorial
- HumanMessage -> HumanChatMessage
- SystemMessage -> SystemChatMessage


jacoblee93:
Thanks for flagging - we've switched to those class names in the latest version: https://github.com/hwchase17/langchainjs/blob/main/langchain/src/schema/index.ts#L166

Would be open to a note in the docs for pre 0.0.102 versions there if you're down to add it in another PR, but will close this for now.

PR #1898, opened 2023-07-08
refactor: only fallback to approximate if tiktoken fails

iloveitaly:
minor refactor to make the code more readable


jacoblee93:
Thank you!

PR #1899, opened 2023-07-08
docs: add advanced llm parameter example

iloveitaly:
I had to dig around in the source code to find the -1 magic param for maxTokens.


jacoblee93:
Thank you!

PR #1900, opened 2023-07-08
docs: add a simple playwright example, fix bad imports

iloveitaly:
fwiw I wasn't able to get `waitForResponse` working for me and it's not clear from the docs why you need it in the evaluate block.
The default evaluate function does not use it.


jacoblee93:
Thank you!

Issue #1129, opened 2023-05-04
[Feature] Integration with Google PaLM APIs


heqian:
Reference:

- https://github.com/hwchase17/langchain/pull/3575
- https://developers.generativeai.google/tutorials/chat_node_quickstart

ap-1:
Is there a PR open for this? I may be able to help.

afirstenberg:
@ap-1 - The VertexAI PaLM models have been fully integrated for several versions now.

I've had the "PaLM API" models on hold until they make them available without the wait list. But if you'd like to tackle it - go for it! (I'm currently backlogged on trying to get Matching Engine to work.)

anselm94:
@ap-1 @afirstenberg I just raised a PR for Google PaLM support :). Yet to write documentation and have to make some changes to be compatible with Python implementation [here](https://github.com/hwchase17/langchain/pull/3575)

PR #1907, opened 2023-07-09
Feature: Add Google PaLM API support

anselm94:
- Adds support for Google PaLM API support
- Addresses #1129
- Related to https://github.com/hwchase17/langchain/pull/3575
- This is different from Google Vertex AI support added by @afirstenberg in the PR series 
  - #1273
  - #1384
  - #1388
- Added wrappers for all 3 models
  - LLM - `GooglePalm`
  - Embeddings - `GooglePalmEmbeddings`
  - Chat - `ChatGooglePalm`
- Added integration & unit tests
- Added documentation

##### P.S.
- I got access to [Google Makersuite](https://makersuite.google.com/) and thought it'll be cool to add PaLM API support
- I chose to set The Google PaLM Key in the `GOOGLEPALM_API_KEY` environment variable and not `GOOGLE_API_KEY` since this is used for the SerpAPI tool.

Twitter: [@MerbinJAnselm](https://twitter.com/MerbinJAnselm)

Fixes https://github.com/hwchase17/langchainjs/issues/1129


jacoblee93:
Looks broadly ok but it's the exact same models as Vertex AI right? Would it be possible to roll this into the existing Vertex support/reuse more of the existing components?

CC @afirstenberg do you have any thoughts here?

langchain/src/chat_models/googlepalm.ts
```diff
@@ -0,0 +1,242 @@
...
+      : undefined;
+  }
+
+  _mapBaseMessagesToPalmMessages(
+    messages: BaseMessage[]
+  ): protos.google.ai.generativelanguage.v1beta2.IMessage[] {
+    // remove all 'system' messages
+    const filteredMessages = messages.filter((m) => m._getType() !== "system");
+
+    // merges successive messages of same type (say if consecutive human messages) into a single message
```
jacoblee93, line 185: We throw an error in Vertex, this makes sense but would be nice to reuse the benefits!

langchain/src/chat_models/googlepalm.ts
```diff
@@ -0,0 +1,242 @@
...
+    // Google Palm requires messages of alternative types
+    // i.e. 'human, bot, human, bot, human', but not - 'human, human, bot, bot'
+    const mergedMessages: BaseMessage[] = [];
+    filteredMessages.forEach((m) => {
```
jacoblee93, line 189: See above

jacoblee93 (commented):
Would love to try to combine some of this logic with existing Vertex AI components or even combine the models altogether like we have with Azure OpenAI and OpenAI!

anselm94:
Thanks @jacoblee93 for your inputs. I was checking the `langchain` python implementation [`chat_models` here](https://github.com/hwchase17/langchain/tree/master/langchain/chat_models), [`llms` here](https://github.com/hwchase17/langchain/tree/master/langchain/llms) and [`embeddings` here](https://github.com/hwchase17/langchain/tree/master/langchain/embeddings), and hence implemented this separate from Vertex AI.

Technically I find some differences, i.e. in case of OpenAI vs Azure OpenAI, the NodeJS client seems is regardless the same
  - NPM Package: [`openai`](https://www.npmjs.com/package/openai)
  - Import: `import { OpenAIApi } from "openai"`;

But in the case of Google PaLM 2 API vs Google Vertex AI, it requires two different set of NodeJS clients (a Node client exists for Google Vertex AI too and probably we can switch)
  - Google PaLM 2 API
    - NPM Package: [`@google-ai/generativelanguage`](https://www.npmjs.com/package/@google-ai/generativelanguage)
    - Import: `import { DiscussServiceClient } from "@google-ai/generativelanguage"; // for chat`
    - Example: [Chat sample](https://developers.generativeai.google/tutorials/chat_node_quickstart#generate_messages)
  - Google Vertex AI
    - NPM Package: [`@google-cloud/aiplatform`](https://www.npmjs.com/package/@google-cloud/aiplatform)
    - Import: `import { PredictionServiceClient } from "@google-cloud/aiplatform"; // for chat`
    - Example: [Chat sample](https://cloud.google.com/vertex-ai/docs/generative-ai/chat/test-chat-prompts#test_chat_prompts)

While both of the integrations have a lot of similarities and can be abstracted out. But since during the runtime, they have different authentication needs (one requires an API Key like [this](https://developers.generativeai.google/tutorials/setup) while the other requires a [different approach](https://cloud.google.com/docs/authentication/provide-credentials-adc)) and different client libraries.

Simplest way I could think of would be writing plain REST API integration like that of Baidu Wenxin integration. Any other ideas/suggestions?

langchain/src/chat_models/googlepalm.ts
anselm94: Do I need to check for only alternative messages or adapt this merging algorithm in Vertex AI implementation?

anselm94:
I did a basic evaluation via Curl and here are the requests & responses below

### Google PaLM via MakerSuite

`GOOGLEPALM_API_KEY` is generated via Google MakerSuite

```shell
# REQUEST
MODEL_NAME=models/text-bison-001
curl \
  -H 'Content-Type: application/json' \
  -X POST "https://generativelanguage.googleapis.com/v1beta2/${MODEL_NAME}:generateText?key=${GOOGLEPALM_API_KEY}" \
  -d '{ 
        "prompt": { 
            "text": "Write a poem about an adventure in an underwater castle\n\n"
        },
        "temperature": 0.7,
        "top_k": 40,
        "top_p": 0.95,
        "candidate_count": 1,
        "max_output_tokens": 1024,
        "stop_sequences": [],
        "safety_settings": []
    }'

# RESPONSE:
# {
#   "candidates": [
#     {
#       "output": "Deep in the ocean,\nWhere the water is clear,\nStands a castle,\nMade of coral and pearl.\n\nA mermaid princess lives there,\nWith her servants and guards.\nShe is kind and beautiful,\nAnd her heart is full of grace.\n\nOne day, a young prince\nWas sailing by the castle.\nHe saw the princess,\nAnd he was immediately smitten.\n\nHe dove into the water\nAnd swam to the castle.\nHe asked the princess\nTo marry him.\n\nThe princess was delighted,\nAnd she said yes.\nThe prince and princess\nWere married in a grand ceremony.\n\nThey lived happily ever after\nIn the underwater castle.\nThey had many children,\nWho were all as beautiful and kind\nAs their parents.",
#       "safetyRatings": [
#         {
#           "category": "HARM_CATEGORY_DEROGATORY",
#           "probability": "NEGLIGIBLE"
#         },
#         {
#           "category": "HARM_CATEGORY_TOXICITY",
#           "probability": "NEGLIGIBLE"
#         },
#         {
#           "category": "HARM_CATEGORY_VIOLENCE",
#           "probability": "NEGLIGIBLE"
#         },
#         {
#           "category": "HARM_CATEGORY_SEXUAL",
#           "probability": "NEGLIGIBLE"
#         },
#         {
#           "category": "HARM_CATEGORY_MEDICAL",
#           "probability": "NEGLIGIBLE"
#         },
#         {
#           "category": "HARM_CATEGORY_DANGEROUS",
#           "probability": "NEGLIGIBLE"
#         }
#       ]
#     }
#   ]
# }
```

### Google PaLM via Google Vertex AI

`GCLOUD_ACCESS_TOKEN` corresponds to `gcloud auth print-access-token` output

```shell
# REQUEST
API_ENDPOINT="us-central1-aiplatform.googleapis.com"
PROJECT_ID="<my-project-id>"
MODEL_ID="text-bison@001"
curl \
-X POST \
-H "Authorization: Bearer ${GCLOUD_ACCESS_TOKEN}" \
-H "Content-Type: application/json" \
"https://${API_ENDPOINT}/v1/projects/${PROJECT_ID}/locations/us-central1/publishers/google/models/${MODEL_ID}:predict" -d \
$'{
    "instances": [
        {
            "content": "Write a poem about an adventure in an underwater castle\n\n"
        }
    ],
    "parameters": {
        "temperature": 0.7,
        "maxOutputTokens": 1024,
        "topP": 0.95,
        "topK": 40
    }
}'

# RESPONSE
# {
#   "predictions": [
#     {
#       "content": "In a wondrous underwater castle,\nI found myself one day.\nThe halls were lined with treasures,\nAnd the mermaids sang me a song.\n\nI swam through the crystal clear waters,\nAnd marveled at the beauty of the sea.\nI met all sorts of interesting creatures,\nAnd had many exciting adventures.\n\nOne day, I was exploring a new part of the castle,\nWhen I came across a secret room.\nInside the room was a powerful treasure,\nBut it was guarded by a fierce dragon.\n\nI knew I had to be brave,\nSo I faced the dragon and fought it.\nAfter a long and difficult battle,\nI was victorious and claimed the treasure.\n\nI left the underwater castle with a new sense of pride,\nAnd I knew that I would never forget my adventure.\nIt was an experience that I will cherish forever.",
#       "citationMetadata": {
#         "citations": []
#       },
#       "safetyAttributes": {
#         "categories": [],
#         "scores": [],
#         "blocked": false
#       }
#     }
#   ]
# }
```

Both the APIs have similar functionalities but has different request/response structure. Building an API wrapper makes better sense in this case

langchain/src/chat_models/googlepalm.ts
```diff
@@ -0,0 +1,242 @@
...
+
+  apiKey?: string;
+
+  private client: DiscussServiceClient;
+
+  constructor(fields?: GooglePalmChatInput) {
+    super(fields ?? {});
+
+    this.model = fields?.model ?? this.model;
+    if (this.model && !this.model.startsWith("models/")) {
```
jacoblee93, line 90: If the prefix is always there, do we really need it?

langchain/src/chat_models/googlepalm.ts
```diff
@@ -0,0 +1,242 @@
...
+        context: this._getPalmContextInstruction(messages),
+        examples: this.examples,
+        messages: this._mapBaseMessagesToPalmMessages(messages),
+      },
+    });
+    return this._mapPalmMessagesToChatResult(res[0]);
+  }
+
+  _getPalmContextInstruction(messages: BaseMessage[]): string | undefined {
+    // filters out all 'system' messages, but select the first message
```
jacoblee93, line 171: I can't access the docs, but Vertex AI accepts a system message only as the first message.

This is different for PaLM?

langchain/src/chat_models/googlepalm.ts
```diff
@@ -0,0 +1,242 @@
+import { DiscussServiceClient, protos } from "@google-ai/generativelanguage";
+import { GoogleAuth } from "google-auth-library";
+import { CallbackManagerForLLMRun } from "../callbacks/manager.js";
+import { AIMessage, BaseMessage, ChatResult } from "../schema/index.js";
+import { getEnvironmentVariable } from "../util/env.js";
+import { BaseChatModel, BaseChatModelParams } from "./base.js";
+
+export interface GooglePalmChatInput extends BaseChatModelParams {
```
jacoblee93, line 8: I think they might appreciate it if we call it e.g. `GooglePaLMChatInput`?

langchain/src/chat_models/googlepalm.ts
```diff
@@ -0,0 +1,242 @@
+import { DiscussServiceClient, protos } from "@google-ai/generativelanguage";
```
jacoblee93, line 1: `import type { protos }` is better here I believe

langchain/src/embeddings/googlepalm.ts
```diff
@@ -0,0 +1,75 @@
...
+import { Embeddings, EmbeddingsParams } from "./base.js";
+import { getEnvironmentVariable } from "../util/env.js";
+
+export interface GooglePalmEmbeddingsParams extends EmbeddingsParams {
+  /**
+   * Model Name to use
+   *
+   * Note: The format must follow the pattern - `models/{model}`
+   */
+  model?: string;
```
jacoblee93, line 12: We use `modelName` by convention for this

langchain/src/llms/googlepalm.ts
```diff
@@ -0,0 +1,208 @@
...
+ * Google Palm 2 Language Model Wrapper to generate texts
+ */
+export class GooglePalm extends LLM implements GooglePalmTextInput {
+  get lc_secrets(): { [key: string]: string } | undefined {
+    return {
+      apiKey: "GOOGLEPALM_API_KEY",
+    };
+  }
+
+  model = "models/text-bison-001";
```
jacoblee93, line 93: `modelName` is the convention rather than `model`

langchain/src/llms/googlepalm.ts
```diff
@@ -0,0 +1,208 @@
...
+  _llmType(): string {
+    return "googlepalm";
+  }
+
+  async _call(
+    prompt: string,
+    options: this["ParsedCallOptions"]
+  ): Promise<string> {
+    const res = await this.caller.callWithOptions(
+      { signal: options.signal },
```
jacoblee93, line 178: Is it possible to pass this into the client directly? You *can* do this but it's a last resort - it's hacky and doesn't actually cancel the in-flight request.

jacoblee93 (changes requested):
Looks solid, a few comments!

jacoblee93:
> Both the APIs have similar functionalities but has different request/response structure. Building an API wrapper makes better sense in this case if we have to proceed with the unified approach. But again, finally, I don't find much similarity between two APIs and hence a single implementation, I presume, will have a lot of if-elses for mapping request & response bodies for each of LLM, Embedding & Chat models.

Got it - that's unfortunate but let's proceed with this approach then.

afirstenberg:
> While both of the integrations have a lot of similarities and can be abstracted out. But since during the runtime, they have different authentication needs (one requires an API Key like [this](https://developers.generativeai.google/tutorials/setup) while the other requires a [different approach](https://cloud.google.com/docs/authentication/provide-credentials-adc)) and different client libraries.
> 
> Simplest way I could think of would be writing plain REST API integration like that of Baidu Wenxin integration. Any other ideas/suggestions?

A few thoughts:

* The VertexAI implementation does use the REST API. At the time I started, there was no JavaScript API for the GenAI models and it wasn't clear there ever would be one. It is generic enough that we _could_ write the PaLM APIs to use it, and that was my plan, but I'm not sure how much it gains us.
* Although the underlying models are supposed to be the same, we don't  actually know if they are. The two APIs are maintained by two completely separate teams, from what limited info I understand.
* The Vertex AI API supports models that the PaLM API doesn't, at least so far. There are suggestions that Vertex AI
* The two have different naming conventions for their models. Trying to normalize them would be a maintenance nightmare.
* While the models have many of the same parameters - there are differences in both the parameters and what is returned.

Given all this, we unfortunately have to consider these two different creatures. Google seems to be.

jacoblee93:
> > While both of the integrations have a lot of similarities and can be abstracted out. But since during the runtime, they have different authentication needs (one requires an API Key like [this](https://developers.generativeai.google/tutorials/setup) while the other requires a [different approach](https://cloud.google.com/docs/authentication/provide-credentials-adc)) and different client libraries.
> > Simplest way I could think of would be writing plain REST API integration like that of Baidu Wenxin integration. Any other ideas/suggestions?
> 
> A few thoughts:
> 
> * The VertexAI implementation does use the REST API. At the time I started, there was no JavaScript API for the GenAI models and it wasn't clear there ever would be one. It is generic enough that we _could_ write the PaLM APIs to use it, and that was my plan, but I'm not sure how much it gains us.
> * Although the underlying models are supposed to be the same, we don't  actually know if they are. The two APIs are maintained by two completely separate teams, from what limited info I understand.
> * The Vertex AI API supports models that the PaLM API doesn't, at least so far. There are suggestions that Vertex AI
> * The two have different naming conventions for their models. Trying to normalize them would be a maintenance nightmare.
> * While the models have many of the same parameters - there are differences in both the parameters and what is returned.
> 
> Given all this, we unfortunately have to consider these two different creatures. Google seems to be.

Thank you for the context - yes this seems like the best option then.

langchain/src/chat_models/googlepalm.ts
anselm94: Example documentation [here](https://developers.generativeai.google/tutorials/chat_node_quickstart#generate_messages). PaLM `DiscussService` client needs a `context` (which corresponds to `SystemMessage`) and and a array of `messages` (which corresponds to `HumanMessage` & `AIMessage`).

```javascript
const result = await client.generateMessage({
    ...
    prompt: {
      // optional, preamble context to prime responses
      context: "Respond to all questions with a rhyming poem.",
      examples: [
         ...
      ],
      // Required. Alternating prompt/response messages.
      messages: [{ content: "How tall is the Eiffel Tower?" }],
    },
  });
```

Since it's a convention to keep `SystemMessage` as always the first message in `langchain`, let me adopt the same :)

langchain/src/chat_models/googlepalm.ts
anselm94: Here, [this example](https://developers.generativeai.google/tutorials/chat_node_quickstart#generate_messages), it's mentioned `// Alternating prompt/response messages` while in this [documentation](https://developers.generativeai.google/api/rest/generativelanguage/models/generateMessage#response-body), it's simply mentioned as `messages[] : The conversation history used by the model.`. Moreover, since we map the `BaseMessage::name` or `BaseMessage::_getType` as `author` for PalM Message ([doc](https://developers.generativeai.google/api/rest/generativelanguage/Message) - `author: This serves as a key for tagging the content of this Message when it is fed to the model as text. The author can be any alphanumeric string.`), I guess we need not merge nor even check for odd number of messages

langchain/src/chat_models/googlepalm.ts
anselm94: Since the PaLM API requires model name as `models/{model}`, I guess, keeping it as close as possible with underlying implementation makes better sense, since if Google changes the naming pattern later. I'll remove this check for `modelName.startsWith('models/')`

langchain/src/llms/googlepalm.ts
anselm94: I tried this out, unfortunately, I can't find a reliable way with Google's clients. They use [`google-gax`](https://github.com/googleapis/gax-nodejs) -> [`grpc-js`](https://github.com/grpc/grpc-node) lib which has a similar issue open here - https://github.com/grpc/grpc-node/issues/1603. I'm not an expert on this topic :(. Need help here. The `client` has a `TextServiceClient::close()` but that completely terminates the `client` connections and becomes no longer usable.

langchain/src/chat_models/googlepalm.ts
```diff
@@ -0,0 +1,225 @@
...
+        ? messages[0]
+        : undefined;
+    return systemMessage?.content;
+  }
+
+  _mapBaseMessagesToPalmMessages(
+    messages: BaseMessage[]
+  ): protos.google.ai.generativelanguage.v1beta2.IMessage[] {
+    // remove all 'system' messages
+    const nonSystemMessages = messages.filter((m) => m._getType() !== "system");
```
jacoblee93, line 179: I think it's better to throw an error rather than silently change behavior here?

langchain/src/chat_models/googlepalm.ts
```diff
@@ -0,0 +1,225 @@
...
+    // Google Palm doesn't provide streaming as of now. But to support streaming handlers
+    // we call the handler with entire response text
+    void runManager?.handleLLMNewToken(
+      chatResult.generations.length > 0 ? chatResult.generations[0].text : ""
+    );
+
+    return chatResult;
+  }
+
+  async _palmGenerateMessage(messages: BaseMessage[]): Promise<ChatResult> {
```
jacoblee93, line 150: `_generatePalmMessage` is better here

langchain/src/chat_models/googlepalm.ts
```diff
@@ -0,0 +1,225 @@
...
+      prompt: {
+        context: this._getPalmContextInstruction(messages),
+        examples: this.examples,
+        messages: this._mapBaseMessagesToPalmMessages(messages),
+      },
+    });
+    return this._mapPalmMessagesToChatResult(res[0]);
+  }
+
+  _getPalmContextInstruction(messages: BaseMessage[]): string | undefined {
```
jacoblee93, line 166: nit: mark this and other internal instances methods as `protected`

langchain/src/embeddings/googlepalm.ts
```diff
@@ -0,0 +1,70 @@
...
+        "Please set an API key for Google Palm 2 in the environment variable GOOGLEPALM_API_KEY or in the `apiKey` field of the GooglePalm constructor"
+      );
+    }
+
+    this.client = new TextServiceClient({
+      authClient: new GoogleAuth().fromAPIKey(this.apiKey),
+    });
+  }
+
+  async _palmEmbedText(text: string): Promise<number[]> {
```
jacoblee93, line 47: `_embedPalmText` or even just `_embedText` is better naming - you can also mark as `protected`

langchain/src/llms/googlepalm.ts
```diff
@@ -0,0 +1,203 @@
...
+export class GooglePaLM extends LLM implements GooglePaLMTextInput {
+  get lc_secrets(): { [key: string]: string } | undefined {
+    return {
+      apiKey: "GOOGLEPALM_API_KEY",
+    };
+  }
+
+  modelName = "models/text-bison-001";
+
+  temperature?: number = undefined; // default value chosen based on model
```
jacoblee93, line 95: I don't think we need to have a default initializer here? Just `temperature?: number;`?

langchain/src/llms/googlepalm.ts
jacoblee93: No worries then.

langchain/src/llms/googlepalm.ts
```diff
@@ -0,0 +1,203 @@
...
+  ): Promise<string> {
+    const res = await this.caller.callWithOptions(
+      { signal: options.signal },
+      this._palmGenerateText.bind(this),
+      prompt
+    );
+    return res ?? "";
+  }
+
+  async _palmGenerateText(prompt: string): Promise<string | null | undefined> {
```
jacoblee93, line 180: See above

jacoblee93 (changes requested):
Almost there! A few nits and one question.

jacoblee93:
FYI I also pushed some small changes to the branch (docs and reverting a probably unintentional config change).

langchain/src/chat_models/googlepalm.ts
anselm94: I did some comprehensive testing with PaLM API regarding how chat messages are handled, and I observed that the messages should indeed be alternating and it doesn't automatically merge them based on `author` tags, but instead reject them based on those tags. So, my observation only based on docs yesterday was wrong.
However, it doesn't only accept the odd number of messages to be present (like that of PaLM in Vertex AI), i.e. messages can be human-ai-human or ai-human-ai-human as well. Hence, I updated the logic to check for alternating messages, or else throw error. Thanks for this review comment!

anselm94:
> FYI I also pushed some small changes to the branch (docs and reverting a probably unintentional config change).

Thanks! I made changes based on the review feedback. Also, have moved the input validation out of async caller parameter to fail before execution and have improved the integration tests to fail faster.

jacoblee93:
I put `.skip` on all the integration tests due to limited access to the model at the moment and changed the env var name from `GOOGLEPALM_API_KEY` to `GOOGLE_PALM_API_KEY`. Can you give a final look and run tests?

anselm94:
> I put `.skip` on all the integration tests due to limited access to the model at the moment and changed the env var name from `GOOGLEPALM_API_KEY` to `GOOGLE_PALM_API_KEY`. Can you give a final look and run tests?

@jacoblee93 Thanks! Yes, verified all the tests by `GOOGLE_PALM_API_KEY` and temporarily removed `.skip` from integration tests and works perfectly good 👍🏾

jacoblee93 (approved):

jacoblee93:
Great, thank you!

PR #1912, opened 2023-07-10
Upgrade @anthropic-ai/sdk to 0.5.3

dqbd:
Unfortunately, cancellation must be disabled for now due to lack of support in the SDK: https://github.com/anthropics/anthropic-sdk-typescript/issues/43

### DIFF
--- a/langchain/package.json
+++ b/langchain/package.json
@@ -777,7 +777,7 @@
     }
   },
   "dependencies": {
-    "@anthropic-ai/sdk": "^0.4.3",
+    "@anthropic-ai/sdk": "^0.5.3",
     "ansi-styles": "^5.0.0",
     "binary-extensions": "^2.2.0",
     "camelcase": "6",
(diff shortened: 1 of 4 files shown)


jacoblee93:
> Unfortunately, cancellation must be disabled for now due to lack of support in the SDK: [anthropics/anthropic-sdk-typescript#43](https://github.com/anthropics/anthropic-sdk-typescript/issues/43)

It looks like you might be able to hack it by setting `stream.controller = params.controller`?

There's probably someone relying on this.

https://github.com/anthropics/anthropic-sdk-typescript/blob/5cfb96fba84a22c5bca576a9791dbb1099bf4c75/src/streaming.ts#L80

NickMandylas:
+1 for this update. NextJS edge runtimes do not support `cross-fetch`, which is used in versions < 0.5. 
Langchain breaks on async calls made when Anthropic is used: https://github.com/anthropics/anthropic-sdk-typescript/issues/23

jacoblee93:
Support merged in #1932.

PR #1917, opened 2023-07-10
Creating sonix audio transcription document loader integration

konstantinov-raft:
Introducing an ability to load a document from local or public audio file using integration with the sonix.ai service.


langchain/src/document_loaders/web/sonix_audio_transcription.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  async load(): Promise<Document[]> {
+    const { text, status, error } =
+      await this.sonixSpeechRecognitionService.speechToText(
+        this.speechToTextRequest
+      );
+
+    if (status === "failed") {
+      console.error("Error:", error);
+      return [];
```
jacoblee93, line 33: Does it make more sense to throw an error here?

jacoblee93 (commented):
Oh actually, one small question:

langchain/src/document_loaders/web/sonix_audio_transcription.ts
konstantinov-raft: I aligned behavior with another integrations which don't throw in case of error, but looks like it makes sense to throw.

konstantinov-raft: Sent a new pr to address this comment https://github.com/hwchase17/langchainjs/pull/1938

PR #1920, opened 2023-07-10
feat: sentinel support through ioredis

### LINKED ISSUE #1710: Redis Sentinel Support
The current implementation of LangChainJS' [RedisChatMessageHistory](https://github.com/hwchase17/langchainjs/blob/main/langchain/src/stores/message/redis.ts) uses the base [Redis createClient](https://github.com/redis/node-redis) which does not have native Redis Sentinel support as far as I can tell.

There is also the [ioredis package](https://github.com/redis/ioredis) which supports both Sentinel and none-Sentinel deployments. Is there a particular reason we opted for the the base client rather than ioredis or is this a change we are open to making?

chasemcdo:
This PR addresses issue #1710 and Sentinel Support for Redis backed Message History.
For the purpose of compatibility it is added as a separate import rather than replacing the existing `node-redis` based client.
This client supports both Sentinel as well as none sentinel backed instances, so is a suitable replacement for the existing implementation.

Also addresses the client typing issue in [this comment](https://github.com/hwchase17/langchainjs/blob/cb50b0ef1f79e330120e55bef26dcdaa0b8217ba/langchain/src/stores/message/redis.ts#L19-L21) (not sure if there is an open issue for that to link)

Fixes #1710


jacoblee93:
Do you think this makes more sense to call this `RedisSentinelChatMessageHistory` since there's not any benefit to the package otherwise?

`langchain/stores/message/ioredis` isn't quite as nice/as descriptive of an entrypoint.

chasemcdo:
> Do you think this makes more sense to call this `RedisSentinelChatMessageHistory` since that's the main benefit to the package?
> 
> `langchain/stores/message/ioredis` isn't quite as nice/as descriptive of an entrypoint.

Definitely agree that given its current benefit is the addition of Sentinel support that could make sense; however, if we are open to / planning to migrate to this as a primary client instead of supporting both `node-redis` and `ioredis`, that name might be too prescriptive given it does more than just sentinel.

But I am open to renaming this now for the purpose of more obvious differentiation between the two and changing it to be less prescriptive once/if that deprecation happens if that is what you had in mind @jacoblee93 ?

jacoblee93:
It's been a little while but are there any advantages to the `redis` client? If not then yeah maybe we just go with your current approach then. Would be nice to get equivalent re-implementations + deprecations of the cache and vector store modules too in that case, but that's a task for another day :)

chasemcdo:
That is a fantastic question. I have been trying to find out why they both exist.
As far as I can tell they both continue to receive contributions, but [some sources](https://www.libhunt.com/compare-ioredis-vs-NodeRedis--node-redis) list the `node-redis` package as being "discontinued". The [migration docs](https://github.com/redis/ioredis/wiki/Migrating-from-node_redis) for `node-redis` -> `ioredis` seem to imply a similar story as well.

Been trying to find an advantage to `node-redis`, but as of yet I have been unable to do so

jacoblee93:
Thank you! Added a deprecation TODO for the other implementation and ran formatting.

PR #1922, opened 2023-07-11
Fix the batchSize parameter of OpenAIEmbeddings

hujiulong:
According to the operator precedence in JavaScript, the nullish coalescing operator `(??)` has a lower precedence than the conditional operator `(? :)`. Therefore, the expression is actually evaluated as
```
this.batchSize = (fields?.batchSize ?? azureApiKey) ? 1 : this.batchSize
```
 so the `batchSize` parameter will not work


jacoblee93:
Thank you for the catch!

PR #1932, opened 2023-07-11
Update Anthropic SDK to 0.5.X to support Claude V2

cwoolum:
Anthropic has released Claude V2 which makes use of an updated API. The TypeScript SDK 0.5.X has been rewritten to support the new API schema which also applies to Claude V1. 

It should be noted that the TypeScript SDK no longer supports `AbortController` inherently and so it can only be used for streaming responses to short circuit what is returned by the LangchainJS client.

Fixes # (issue)

### DIFF
--- a/langchain/src/chat_models/anthropic.ts
+++ b/langchain/src/chat_models/anthropic.ts
@@ -3,9 +3,7 @@ import {
   Anthropic as AnthropicApi,
   HUMAN_PROMPT,
 } from "@anthropic-ai/sdk";
-import { APIResponse } from "@anthropic-ai/sdk/core";
-import { CompletionCreateParams } from "@anthropic-ai/sdk/resources/completions";
-import { Stream } from "@anthropic-ai/sdk/streaming";
+import type { CompletionCreateParams } from "@anthropic-ai/sdk/resources/completions";
 
 import { BaseLanguageModelCallOptions } from "../base_language/index.js";
 import { CallbackManagerForLLMRun } from "../callbacks/manager.js";
@@ -247,21 +245,15 @@ export class ChatAnthropic extends BaseChatModel implements AnthropicInput {
       runManager
     );
 
-    if ("completion" in response) {
-      const generations: ChatGeneration[] = response.completion
-        .split(AI_PROMPT)
-        .map((message) => ({
-          text: message,
-          message: new AIMessage(message),
-        }));
-
-      return {
-        generations,
-      };
-    }
+    const generations: ChatGeneration[] = response.completion
+      .split(AI_PROMPT)
+      .map((message) => ({
+        text: message,
+        message: new AIMessage(message),
+      }));
 
     return {
-      generations: [],
+      generations,
     };
   }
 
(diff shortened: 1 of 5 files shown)


langchain/src/chat_models/anthropic.ts
```diff
@@ -245,73 +247,107 @@ export class ChatAnthropic extends BaseChatModel implements AnthropicInput {
...
-            })
-        );
+            });
+
+          return stream;
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        } catch (e: any) {
+          // Anthropic doesn't actually throw JavaScript error objects at the moment.
+          // We convert the error so the async caller can recognize it correctly.
+          if (e?.name === "AbortError") {
```
jacoblee93, line 334: I believe they do throw JS errors properly now?

jacoblee93:
Thanks for this! It seems like this doesn't pass our integration test:

`yarn test:single langchain/src/chat_models/tests/chatanthropic.int.test.ts`

Fails the batch request abort test as expected (unfortunately) but also the basic streaming one.

CC @dqbd but I can just have a look as well.

jacoblee93:
Fixed things up and added back in abort for both (taking some ideas from #1912!)

Unfortunately, it seems they're doing some things that break edge environments and it's failing our build tests. Will contact them and try to get them to fix it.

langchain/src/chat_models/anthropic.ts
rattrayalex: We do; AbortError's in JavaScript [weirdly don't have their own class](https://developer.mozilla.org/en-US/docs/Web/API/AbortController/abort):

> Note: When abort() is called, the fetch() promise rejects with an Error of type DOMException, with name AbortError.

langchain/src/chat_models/anthropic.ts
```diff
@@ -295,47 +278,37 @@ export class ChatAnthropic extends BaseChatModel implements AnthropicInput {
...
+          stop_reason: "",
+        };
+
+        for await (const data of stream) {
+          completion.stop_reason = data.stop_reason;
+          completion.model = data.model;
+
+          if (options.signal?.aborted) {
+            stream.controller.abort();
+            throw new Error("AbortError: User aborted the request.");
```
rattrayalex, line 297: FYI, you _might_ want this to be more like:

```ts
const err = new Error("User aborted the request.")
err.name = "AbortError";
throw err;
```
which I believe to be closer to what JS runtimes do.

jacoblee93:
Ok, so it's actually intentional and the newest version of `next` actually supports it via a special case:

https://github.com/vercel/next.js/pull/52009

Bumped the version and it builds locally so it should pass our CI.

langchain/src/chat_models/anthropic.ts
jacoblee93: Ahh wow that explains a lot. Thanks.

langchain/src/chat_models/anthropic.ts
jacoblee93: This gets caught by our caller class which would catch this - we could/should change it in the future but it's out of scope for this one.

jacoblee93:
Thanks again @cwoolum! Will be in tonights release 🙂

cwoolum:
Awesome, thanks! Forgot to include my twitter handle if it still matters. @chriswoolum

PR #1933, opened 2023-07-11
docs: make it clear that chunks are characters and not tokens

iloveitaly:
tokens and characters are mentioned in the same line, this makes it clear that splitters work off of characters (being able to split based on token count would be way better imho).


jacoblee93:
Thank you!

PR #1938, opened 2023-07-12
Update sonix behaviour on error

konstantinov-raft:
Updating sonix integration to throw in case of error. This is to address comment from previous pr https://github.com/hwchase17/langchainjs/pull/1917#discussion_r1260170911.

Also creating tests for the new behavior.


jacoblee93:
Thank you!

Issue #1939, opened 2023-07-12
Langchain JS PromptLayerChatOpenAI doesnt recognize message with type == "function"


digitpdigit:
I'm using "langchain": "^0.0.100",
I got unknown type error when calling function using PromptLayerChatOpenAI gpt-3.5-turbo-0613 model

this is easily solved by adding several lines to openai.cjs as shown in the image below
but that solution would be very unpractical to be done on deployment

![image](https://github.com/hwchase17/langchainjs/assets/67557217/4e407e5a-0f9e-4fa7-9d2f-17cb1356f30d)

might be a minor bug due to recent function calling implementation on the langchain side?

jacoblee93:
Could you open a PR for this?

digitpdigit:
Yes, will do

Thanks for the response

jacoblee93:
Fixed in #1948, will be out with the next release!

PR #1948, opened 2023-07-13
added-accommodate-message-type-function

digitpdigit:
A simple addition of message type function recognition for PromptLayerChatOpenAI models

Fixes [#1939 ]


jacoblee93:
Looks good, thank you!

PR #1955, opened 2023-07-13
Add custom lengthFunction for textSplitters

ShantanuNair:
Needs testing.

- Added a lengthFunction for textSplitter class in an attempt to bring parity to how langchain python allows for a length_function to be passed in. This allows chunking using the RecursiveTextSplitter or CharacterTextSplitter, but uses a custom length function that will allow you to pass in a tiktoken length function or similar.

Fixes https://github.com/hwchase17/langchainjs/issues/1951


jacoblee93:
Looks good to me other than linting/formatting, thank you!

PR #1964, opened 2023-07-14
Feature: Add handleLLMNewFunctionCall callback

mgce:
Right now when we are using a function call, the `handleLLMNewToken` callback returns nothing. The `handleLLMNewFunctionCall` callback allows us to receive a streamed response.

### DIFF
--- a/langchain/src/callbacks/base.ts
+++ b/langchain/src/callbacks/base.ts
@@ -63,7 +63,7 @@ abstract class BaseCallbackHandlerMethodsClass {
    * Called when an LLM/ChatModel in `streaming` mode produces a new function call
    */
   handleLLMNewFunctionCall?(
-    function_call: { name?: string; arguments?: string },
+    functionCall: { name?: string; arguments?: string },
     idx: NewTokenIndices,
     runId: string,
     parentRunId?: string,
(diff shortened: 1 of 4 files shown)


langchain/src/callbacks/base.ts
```diff
@@ -59,6 +59,17 @@ abstract class BaseCallbackHandlerMethodsClass {
     tags?: string[]
   ): Promise<void> | void;
 
+  /**
+   * Called when an LLM/ChatModel in `streaming` mode produces a new function call
+   */
+  handleLLMNewFunctionCall?(
+    function_call: { name?: string; arguments?: string },
```
jacoblee93, line 66: nit: Prefer camel case

mgce: Changed

jacoblee93:
Thanks for the PR - had some discussion with some of the Python maintainers, and it sounds like the preference would be for something more generic like a catchall `handleEvent`. Will leave this open for discussion for now.

mgce:
@jacoblee93 Thanks for the reply. I have created a follow-up PR including this suggestion. 
https://github.com/hwchase17/langchainjs/pull/2025

jacoblee93:
Wow thank you! Will close this one then.

PR #1968, opened 2023-07-14
feat: Adding a parameter `separator` to the `CSVLoader`

yzqzy:
Adding a configuration parameter `separator` for `CSVLoader`


jacoblee93:
Fantastic thank you! A few small updates.

PR #1973, opened 2023-07-15
Add cosine similarity and maximal marginal relevance functions

archie-swif:
Porting the cosine similarity and maximal marginal relevance functions from the Python package

### DIFF
--- a/langchain/src/util/math_utils.ts
+++ b/langchain/src/util/math_utils.ts
@@ -1,0 +1,39 @@
+import * as tf from "@tensorflow/tfjs-core";
+import { Tensor1D } from "@tensorflow/tfjs-core";
+
+/**
+ * This function calculates the row-wise cosine similarity between two matrices with the same number of columns.
+ *
+ * @param {number[][]} X - The first matrix.
+ * @param {number[][]} Y - The second matrix.
+ *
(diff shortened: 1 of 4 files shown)


langchain/src/util/math_utils.ts
```diff
@@ -0,0 +1,134 @@
+import * as tf from "@tensorflow/tfjs-core";
+import { Tensor1D } from "@tensorflow/tfjs-core";
```
jacoblee93, line 2: This is a peer dep - would be nice to not have to use this.

Can we use this built-in instead?

https://github.com/hwchase17/langchainjs/blob/main/langchain/src/vectorstores/memory.ts#L1

And then calculate MMR off of that? It would make this more portable.

jacoblee93 (changes requested):
Thanks for the PR - see comment above.

langchain/src/util/math_utils.ts
archie-swif: Thanks for the review @jacoblee93! 
ml-distance and ml-matrix also work well for the cosine similarity calculation and tests.
Updated the code to use them instead.

jacoblee93:
Excellent, thank you!

jacoblee93 (approved):

Issue #1357, opened 2023-05-22
RecursiveCharacterTextSplitter invalid first line indent/trim


tonisives:
When this class creates the pageContent, it trims the first line to 0 indent. However, the next lines will keep the indent. This means that if I'm dealing with code, it will show invalid indenting, and I cannot use https://www.npmjs.com/package/strip-indent to remove indentation from all of the lines

### example
source
```
    // JBDirectory
    _jbDirectory = new JBDirectory(_jbOperatorStore, _jbProjects, _jbFundingCycleStore, _multisig);
    evm.label(address(_jbDirectory), 'JBDirectory');
```

result

```
// JBDirectory
    _jbDirectory = new JBDirectory(_jbOperatorStore, _jbProjects, _jbFundingCycleStore, _multisig);
    evm.label(address(_jbDirectory), 'JBDirectory');
```

I cannot manually indent the first line, because in case of functions, it should be 0 indented.

edit:
issue is this trim() [here](https://github.com/hwchase17/langchainjs/blob/af842d7813861bd810d43fbfee2ab8c2482df506/langchain/src/text_splitter.ts#L90)

PR #1977, opened 2023-07-16
Fix loc for repeating text in text_splitter

tonisives:
This PR fixes an issue where the Document `loc` is wrong if there are repeating blocks. This happens especially obviously in code files where there are similar closing blocks. For example `\n    }`. 

The current code erroneously sets the `loc` to more lines there are lines in the file. This is because it doesn't use the lastIndex of the chunk, but instead the first occurrence.

Might also fix issue #1357

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,4 @@ Chinook_Sqlite.sql
 
 firebase-debug.log
 firestore-debug.log
+.vscode
--- a/langchain/src/tests/text_splitter.test.ts
+++ b/langchain/src/tests/text_splitter.test.ts
@@ -323,7 +323,7 @@ test("Test lines loc on iterative text splitter.", async () => {
     }),
     new Document({
       pageContent: "How?\na\nb",
-      metadata: { loc: { lines: { from: 2, to: 6 } } },
+      metadata: { loc: { lines: { from: 4, to: 6 } } },
     }),
   ];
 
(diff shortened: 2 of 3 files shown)


jacoblee93:
Makes sense but breaks a few other test cases :( Can you fix?

tonisives:
This solution needs some more work. My LOC is still not precise in some cases.

Probably yeah, fixing the tests would be a good start

yyvess:
@tonisives 
I created a similarly PR (sorry I see your after) bit more simple to fix the issue about new lines on overlap and on beginning of the content 
https://github.com/langchain-ai/langchainjs/pull/2779

About issues with multiple chunk with same content that I understand you try to fix by "find the correct loc by brute force".

I think seen better to completely rewrite the split function to compute lines number during the split and not try to find line number after the splitting as it actually.

But that require a big refactoring ....

tonisives:
I added a brute force loc fixer after splitting documents. [add fix loc method](https://github.com/langchain-ai/langchainjs/pull/1977/commits/850a30f5846cd639c9cc2d8e1f710d7fceffe8c0)

It is obviously inefficient and the loc should be correct from the initial parsing. I think the problem comes from `splitText`, which doesn’t return \n characters, which are used to count the loc. It’s a confusing function and I don’t have time to debug it.

-- 

Generally a rewrite sounds like a good idea. Starting with proper test cases for all of the features and then separate functions for each of the features (chunks, overlaps, mapping, etc). Currently the code is hard to follow and I think not all features are properly tested.

But it's a lot of work and I might just write a personal one for my use case only.

tonisives:
verified fixed in https://github.com/langchain-ai/langchainjs/pull/2779  

https://github.com/langchain-ai/langchainjs/releases/tag/0.0.161

PR #1984, opened 2023-07-16
Add langchain.docstore.document bind

HTMHell:
Fixes # ([Bug: TypeError: value.toDocument is not a function](https://github.com/hwchase17/langchainjs/issues/1976))


jacoblee93:
Thank you! Seems fine but CC @ewfian

PR #1985, opened 2023-07-17
A quick way to create a fake, but consistent, pseudo-embeddings

afirstenberg:
This provides a SyntheticEmbeddings class which is created with a specified vector size. It then creates a vector of this size based on the input.
* Identical inputs should yield the same output for the same vector size. ("this" will always have the same vector)
* Similar inputs of the same size should yield nearby vectors, depending (very roughly) how similar. ("this" and "that" should have nearby vectors, but it depends on the vector size)
* Inputs of different sizes may give similar vectors, depending in the input sizes and vector size. In some cases. ("this" and "thing" may have nearby vectors, but it depends largely on the vector size. If the vector size is 10, for example, the first three components in the vector should be the same.)

This class is meant to be used for testing in cases where it is advantageous to have several consistent, but not always identical to each other, vectors without needing to install a real embeddings engine.

### DIFF
--- a/langchain/src/embeddings/fake.ts
+++ b/langchain/src/embeddings/fake.ts
@@ -15,11 +15,13 @@ export class FakeEmbeddings extends Embeddings {
 }
 
 interface SyntheticEmbeddingsParams extends EmbeddingsParams {
-  vectorSize: number
+  vectorSize: number;
 }
 
-export class SyntheticEmbeddings extends Embeddings implements SyntheticEmbeddingsParams {
-
+export class SyntheticEmbeddings
+  extends Embeddings
+  implements SyntheticEmbeddingsParams
+{
   vectorSize: number;
 
   constructor(params?: SyntheticEmbeddingsParams) {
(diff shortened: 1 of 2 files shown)


langchain/src/embeddings/fake.ts
```diff
@@ -13,3 +13,62 @@ export class FakeEmbeddings extends Embeddings {
...
+    doc = doc.padEnd(padSize, " ");
+    // console.log(padGapSize, padSize, doc);
+
+    // Break it into chunks
+    const chunkSize = doc.length / this.vectorSize;
+    const docChunk = [];
+    for (let co = 0; co < doc.length; co += chunkSize) {
+      docChunk.push(doc.slice(co, co + chunkSize));
+    }
+    // console.log(chunkSize, docChunk);
```
jacoblee93, line 56: Remove log

jacoblee93:
Cool addition! Can you just remove the commented out logs?

afirstenberg:
> Cool addition! Can you just remove the commented out logs?

Done!

jacoblee93:
Thank you!

PR #1988, opened 2023-07-17
Allow promise<number> for splitter lengthFunction; Async mergeSplits

ShantanuNair:
Allows for async lengthFunctions to be passed in to a splitter (Eg: llm.getNumTokens)

Tag Maintainers: @jacoblee93 @nfcampos @hwchase17 

Related previous PR regarding addition of lengthFunction: https://github.com/hwchase17/langchainjs/pull/1955

### DIFF
--- a/langchain/src/text_splitter.ts
+++ b/langchain/src/text_splitter.ts
@@ -7,7 +7,9 @@ export interface TextSplitterParams {
   chunkSize: number;
   chunkOverlap: number;
   keepSeparator: boolean;
-  lengthFunction?: (text: string) => number;
+  lengthFunction?:
+    | ((text: string) => number)
+    | ((text: string) => Promise<number>);
 }
 
 export type TextSplitterChunkHeaderOptions = {
@@ -28,14 +30,17 @@ export abstract class TextSplitter
 
   keepSeparator = false;
 
-  lengthFunction: (text: string) => number = (text: string) => text.length;
+  lengthFunction:
+    | ((text: string) => number)
+    | ((text: string) => Promise<number>);
 
   constructor(fields?: Partial<TextSplitterParams>) {
     super(fields);
     this.chunkSize = fields?.chunkSize ?? this.chunkSize;
     this.chunkOverlap = fields?.chunkOverlap ?? this.chunkOverlap;
     this.keepSeparator = fields?.keepSeparator ?? this.keepSeparator;
-    this.lengthFunction = fields?.lengthFunction ?? ((text) => text.length);
+    this.lengthFunction =
+      fields?.lengthFunction ?? ((text: string) => text.length);
     if (this.chunkOverlap >= this.chunkSize) {
       throw new Error("Cannot have chunkOverlap >= chunkSize");
     }
@@ -95,7 +100,7 @@ export abstract class TextSplitter
         if (prevChunk) {
           const indexChunk = text.indexOf(chunk);
           const indexEndPrevChunk =
-            text.indexOf(prevChunk) + this.lengthFunction(prevChunk);
+            text.indexOf(prevChunk) + (await this.lengthFunction(prevChunk));
           const removedNewlinesFromSplittingText = text.slice(
             indexEndPrevChunk,
             indexChunk
@@ -154,12 +159,12 @@ export abstract class TextSplitter
     return text === "" ? null : text;
   }
 
-  mergeSplits(splits: string[], separator: string): string[] {
+  async mergeSplits(splits: string[], separator: string): Promise<string[]> {
     const docs: string[] = [];
     const currentDoc: string[] = [];
     let total = 0;
     for (const d of splits) {
-      const _len = this.lengthFunction(d);
+      const _len = await this.lengthFunction(d);
       if (
         total + _len + (currentDoc.length > 0 ? separator.length : 0) >
         this.chunkSize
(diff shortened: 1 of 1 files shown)


langchain/src/text_splitter.ts
```diff
@@ -216,7 +221,10 @@ export class CharacterTextSplitter
   async splitText(text: string): Promise<string[]> {
     // First we naively split the large input into a bunch of smaller ones.
     const splits = this.splitOnSeparator(text, this.separator);
-    return this.mergeSplits(splits, this.keepSeparator ? "" : this.separator);
+    return await this.mergeSplits(
```
jacoblee93, line 224: Don't need an `await` here

jacoblee93:
Looks fine - what's the use-case for an async method though?

langchain/src/text_splitter.ts
ShantanuNair: @jacoblee93 Resolved it, thanks! The use case is for using getNumTokens as a custom length function, just as can be done in langchain python. So any text splitter can allow for using token counting instead of character counting. 

This is  important in applications such as chunking for vector retrieval, particularly when you're working with multiple languages. The openAI tokenizer is optimized for different languages differently, and some languages uses way more tokens per character than seen with English language tokenization. So if you want to retrieve, say 8 chunks, and you set that so you don't go over context size limits, then using a RecursiveTextSplitter using just characters gets tricky. Instead of changing the number of chunks retrieved by the retrieved, you can enforce a token limit when chunking using any splitter.



Eg:
```
  const llmModel = new OpenAI({
    openAIApiKey: process.env.OPENAI_API_KEY,
    modelName: 'gpt-3.5-turbo',
  });

  const tokenLengthFunction = llmModel.getNumTokens

     chunkSize: 500, // Count using tokens
     chunkOverlap: 0,
     lengthFunction: tokenLengthFunction,  // use the LLM's token length function to count tokens instead of characters
   });
```

jacoblee93: Gotcha - and some of those functions are async at the moment or is this future-proofing?

ShantanuNair:
@jacoblee93 Resolved it, thanks! The use case is for using getNumTokens as a custom length function, just as can be done in langchain python. So any text splitter can allow for using token counting instead of character counting. 

This is  important in applications such as chunking for vector retrieval, particularly when you're working with multiple languages. The openAI tokenizer is optimized for different languages differently, and some languages uses way more tokens per character than seen with English language tokenization. So if you want to retrieve, say 8 chunks, and you set that so you don't go over context size limits, then using a RecursiveTextSplitter using just characters gets tricky. Instead of changing the number of chunks retrieved by the retrieved, you can enforce a token limit when chunking using any splitter.



Eg:
```
  const llmModel = new OpenAI({
    openAIApiKey: process.env.OPENAI_API_KEY,
    modelName: 'gpt-3.5-turbo',
  });

  const tokenLengthFunction = llmModel.getNumTokens

     chunkSize: 500, // Count using tokens
     chunkOverlap: 0,
     lengthFunction: tokenLengthFunction,  // use the LLM's token length function to count tokens instead of characters
   });
```

jacoblee93:
Gotcha - and some of those functions are async at the moment or is this future-proofing?

ShantanuNair:
@jacoblee93 It's a current need for us, not just future proofing, since getNumTokens is async. The example code I provided could only work if sayin length functions are supported. 

https://github.com/hwchase17/langchainjs/blob/1b98391ca8af41aab109ec20e05233cb9faf9015/langchain/src/base_language/index.ts#L154

jacoblee93:
Ah so it is - thanks!

sayan1999:
Hi @ShantanuNair 
Is this feature tested and working now. I am writing some scripts and looks like it is not waiting for the promise.

I am using "langchain": "^0.0.138"

sayan1999:
Sample code and error

```


import { RecursiveCharacterTextSplitter } from "langchain/text_splitter";
import { BartTokenizer } from "@xenova/transformers";

// interface TextSplitterParams {
//     chunkSize: number;
//     chunkOverlap: number;
//     keepSeparator: boolean;
//     lengthFunction?: ((text: string) => number) | ((text: string) => Promise<number>);
// }

async function tokenizer_len(x) {
  return BartTokenizer.from_pretrained("facebook/bart-large").then(
    (tok) => tok(x)["input_ids"].size
  );
}

const splitter = new RecursiveCharacterTextSplitter({
  chunkSize: 1024,
  chunkOverlap: 50,
  lengthFunction: tokenizer_len,
});

async function chunkify(text) {
  let output = await splitter.createDocuments([text]);
  return output.map((key) => key.pageContent);
}

const data1 = fs.readFileSync("text.txt", "utf8").toString();
chunkify(data1).then((op) => {
  console.log(op);
});

Error

TypeError: Cannot convert undefined to a BigInt
    at BigInt (<anonymous>)
    at Array.map (<anonymous>)
    at Function._call (file:///media/instantinopaul/data/Code/ML/js-summarize/node_modules/@xenova/transformers/src/tokenizers.js:2325:50)
    at closure (file:///media/instantinopaul/data/Code/ML/js-summarize/node_modules/@xenova/transformers/src/utils/core.js:62:28)
    at RecursiveCharacterTextSplitter.tokenizer_len [as lengthFunction] (file:///media/instantinopaul/data/Code/ML/js-summarize/main.mjs:14:18)
    at async RecursiveCharacterTextSplitter._splitText (file:///media/instantinopaul/data/Code/ML/js-summarize/node_modules/langchain/dist/text_splitter.js:236:18)
    at async RecursiveCharacterTextSplitter.createDocuments (file:///media/instantinopaul/data/Code/ML/js-summarize/node_modules/langchain/dist/text_splitter.js:76:33)
    at async chunkify (file:///media/instantinopaul/data/Code/ML/js-summarize/main.mjs:27:16)

```

PR #1990, opened 2023-07-17
New Feat: Added Retriever Callbacks

ppramesi:
This pull request primarily involves the enhancement of the CallbackManager class with three new callback handlers: handlerRetrieverStart, handleRetrieverEnd and handleRetrieverError. In addition, corresponding handler functions have been added to the Tracer and ConsoleCallbackHandler class to support these new callbacks. These callbacks have been integrated into all retrievers and they will now invoke the corresponding callback handlers when getRelevantDocuments is called.  Additionally, new tests have been introduced for the retriever callbacks in the tracer and callback handler tests to ensure their correct functionality.

To accommodate these enhancements, both the VectorStore class and the BaseRetriever class have been extended with the Serializable abstract class, a necessary prerequisite for implementing the callback handlers (since start handler functions accept Serialized objects as parameter). This also necessitated the implementation of lc_namespace, lc_secrets and lc_aliases variables within these classes.

Note that in the Retriever class, getRelevantDocument calls the VectorStore's similaritySearch, which in turn calls the embedQuery function of the Embedding class. The Callbacks object from retriever should be passed along to the Embedding class so that its callback handlers can be called. However, the implementation of embedding callbacks is currently in a separate pull request (#1859) and is not included in this one. Therefore, when merging these two pull requests, we would need to ensure the Callbacks object is properly passed from the Retriever to the Embedding class.


langchain/src/vectorstores/chroma.ts
```diff
@@ -40,6 +40,10 @@ export class Chroma extends VectorStore {
 
   filter?: object;
 
+  vectorstoreType(): string {
```
jacoblee93, line 43: We've been prefixing these with an underscore (like `_llmType`)

langchain/src/retrievers/databerry.ts
```diff
@@ -24,16 +36,18 @@ export class DataberryRetriever extends BaseRetriever {
 
   apiKey?: string;
 
-  constructor({ datastoreUrl, apiKey, topK, ...rest }: DataberryRetrieverArgs) {
-    super();
+  // constructor({ datastoreUrl, apiKey, topK, ...rest }: DataberryRetrieverArgs) {
```
jacoblee93, line 39: Remove comment

langchain/src/vectorstores/base.ts
```diff
@@ -125,11 +149,48 @@ export abstract class VectorStore {
...
+
   asRetriever(
     k?: number,
-    filter?: this["FilterType"]
+    filter?: this["FilterType"],
+    callbacks?: Callbacks,
+    tags?: string[],
+    metadata?: Record<string, unknown>,
+    verbose?: boolean
+  ): VectorStoreRetriever<this>;
```
jacoblee93, line 163: Do we need these overrides? I think the bottom one should be enough?

jacoblee93 (commented):
This is really excellent! Left some comments and added a small change to get it working with `langchain_tracer.ts`. Would love if @nfcampos could take a second look.

langchain/src/retrievers/time_weighted.ts
```diff
@@ -79,11 +88,17 @@ export class TimeWeightedVectorStoreRetriever extends BaseRetriever {
...
+    query: string,
+    runManager?: CallbackManagerForRetrieverRun
+  ): Promise<Document[]> {
     const now = Math.floor(Date.now() / 1000);
     const memoryDocsAndScores = this.getMemoryDocsAndScores();
 
-    const salientDocsAndScores = await this.getSalientDocuments(query);
+    const salientDocsAndScores = await this.getSalientDocuments(
+      query,
+      runManager?.getChild()
```
jacoblee93, line 100: I think this be just `runManager` since it's not a child call?

langchain/src/retrievers/time_weighted.ts
```diff
@@ -138,12 +153,15 @@ export class TimeWeightedVectorStoreRetriever extends BaseRetriever {
...
+    query: string,
+    callbacks?: Callbacks
   ): Promise<Record<number, { doc: Document; score: number }>> {
     const docAndScores: [Document, number][] =
       await this.vectorStore.similaritySearchWithScore(
         query,
-        this.searchKwargs
+        this.searchKwargs,
+        undefined,
+        callbacks
```
jacoblee93, line 164: Oh, I guess the above doesn't matter. Still maybe better to call `getChild()` here instead for clarity?

langchain/src/schema/retriever.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  constructor(fields?: BaseRetrieverInput) {
+    super(fields);
+    this.callbacks = fields?.callbacks;
+    this.tags = fields?.tags ?? [];
+    this.metadata = fields?.metadata ?? {};
+    this.verbose = fields?.verbose ?? false;
+  }
+
+  abstract _getRelevantDocuments(
```
jacoblee93, line 39: For folks who might be using custom retrievers, might be best to not have this as an `abstract` method and just throw a `NotImplemented` error or something similar? And add a note to make it abstract on the 0.1.0/1.0.0 release.

jacoblee93 (commented):
Small nits but looks pretty much good to go to me.

jacoblee93:
Good with me - will plan to merge in the next few hours. Thank you very much!

PR #1994, opened 2023-07-17
Typescript docs models should have openAIApiKey key in docs

MSturges:
Added reference to openAIApiKey key value pair in javascript documentation for guide-llm.


jacoblee93:
Thanks for the PR! The env var vs. explicit passing is explained here: 

https://js.langchain.com/docs/getting-started/guide-llm#llms-get-predictions-from-a-language-model

Also, if we were to update to this style, we'd want to update it everywhere.

PR #1995, opened 2023-07-17
fix: type modelName

iloveitaly:
This has some great side effects:

* errors when a typo is made
* avoid ts errors when using `model.modelName` in downstream code

here's why it's useful:

```javascript
const model = new OpenAI({
	temperature: 0,
	maxTokens: -1,
	modelName: "gpt-3.5-turbo-16k",
	openAIApiKey: openAIAPIKey,
})

const modelTokenLimit = await calculateMaxTokens({
	prompt: renderedEmptyPrompt,
	modelName: model.modelName,
})
```

### DIFF
--- a/langchain/src/base_language/count_tokens.ts
+++ b/langchain/src/base_language/count_tokens.ts
@@ -78,6 +78,7 @@ export const calculateMaxTokens = async ({
     );
 
     // fallback to approximate calculation if tiktoken is not available
+    // each token is ~4 characters: https://help.openai.com/en/articles/4936856-what-are-tokens-and-how-to-count-them#
     numTokens = Math.ceil(prompt.length / 4);
   }
 
--- a/langchain/src/chat_models/openai.ts
+++ b/langchain/src/chat_models/openai.ts
@@ -16,6 +16,7 @@ import {
   AzureOpenAIInput,
   OpenAICallOptions,
   OpenAIChatInput,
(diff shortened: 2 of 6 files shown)


jacoblee93:
I'd love to hard-type this but unfortunately I think it's best if we don't for two reasons:

1. If a new model comes out, people will get type errors with older LangChain versions.
2. I believe this will cause type errors for the following:
```js
const options = {
  modelName: "gpt-4"
};

new ChatOpenAI(options);
```

Would you still get the autocomplete benefits if you did `TiktokenModel | string` or does that completely wipe it?

iloveitaly:
Thanks for engaging on this!

1. If a new model comes out and it's _not_ explicitly supported by langchain this could cause very hard-to-debug errors. i.e. token max not being calculated properly by `calculateMaxTokens` and probably other assumptions throughout the codebase I'm not aware of. IMHO it's better to hard fail to indicate to the user the library does not officially support the model they are attempting to use. If the user wants an escape hatch, then can always cast the string to the type (`"gpt-4" as TiktokenModel`).
2. Wouldn't this hold true if any of the `fields` object keys were typed using a langchain-defined type?

jacoblee93:
Bump on the above?

iloveitaly:
I'll try to take a look today!

jacoblee93:
Closing for now but feel free to reopen

PR #2001, opened 2023-07-18
Added missing QAEvalChain entrypoint

unokitchen:
Added missing QAEvalChain entrypoint


jacoblee93:
Hey @unokitchen sorry for the confusion - I just talked to the original implementer, who said this has actually been deprecated.

If you really want to use it, you can pull the source/prompt out but no guarantees on quality unfortunately.

PR #2007, opened 2023-07-18
GoogleVertexAIImageEmbeddings

afirstenberg:
This provides an implementation of a class that works with Google's Vertex AI ImageEmbedding model.
It can take text, a buffer with an image, or both, and create embeddings for them in the same embedding space.

See https://console.cloud.google.com/vertex-ai/publishers/google/model-garden/5 for the Model Garden card about this model and https://cloud.google.com/blog/products/ai-machine-learning/enterprise-ready-generative-ai-models-go-ga-in-vertex-ai for the announcement.

@jacoblee93


langchain/src/embeddings/googlevertexai-image.ts
```diff
@@ -0,0 +1,143 @@
...
+      image: r.imageEmbedding,
+    }));
+  }
+
+  async embedMedia(media: GoogleVertexAIMedia[]): Promise<MediaEmbeddings[]> {
+    // Only one media embedding request is allowed
+    return Promise.all(media.map((m) => this.embedMediaQuery(m)));
+  }
+
+  async embedMediaQuery(media: GoogleVertexAIMedia): Promise<MediaEmbeddings> {
```
jacoblee93, line 103: Is there a purpose to embedding both at the same time? Does one input influence the result of the other?

jacoblee93 (commented):
Thanks for submitting this - I think this is generally fine as a one-off but would like to come up with a strategy for how this looks downstream on vector store retrieval as well as document transformation (e.g. text splitting obviously doesn't make sense on images). 

Please give me a bit more time.

langchain/src/embeddings/googlevertexai-image.ts
afirstenberg: The API allows for it is the biggest reason. The two vectors are independent of each other, except they live in the same vector space.

My guess is that the original purpose was so you could submit an image and a text, get the resultant vectors, and compute the similarity between the two all in one shot - which should be faster than two queries. I saw at least one other model that did this as well.

afirstenberg:
I would agree - I'd love to see a more general purpose Document that could hold either text or images or...

...but I think that will take an extensive restructuring of the code across a wide range of classes.
It's a conversation we should stort.
But I don't think we should hold this class till then.

langchain/scripts/create-entrypoints.js
```diff
@@ -40,6 +40,7 @@ const entrypoints = {
   "embeddings/tensorflow": "embeddings/tensorflow",
   "embeddings/hf": "embeddings/hf",
   "embeddings/googlevertexai": "embeddings/googlevertexai",
+  "embeddings/googlevertexai-image": "embeddings/googlevertexai-image",
```
jacoblee93, line 43: Underscore is convention vs. hyphen

langchain/src/embeddings/googlevertexai-image.ts
```diff
@@ -0,0 +1,143 @@
+import { Embeddings, EmbeddingsParams } from "./base.js";
+import {
+  GoogleVertexAIBasePrediction,
+  GoogleVertexAIConnectionParams,
+  GoogleVertexAILLMResponse,
+} from "../types/googlevertexai-types.js";
+import { GoogleVertexAIConnection } from "../util/googlevertexai-connection.js";
+import { AsyncCallerCallOptions } from "../util/async_caller.js";
+// import { chunkArray } from "../util/chunk.js";
```
jacoblee93, line 9: Remove comment

langchain/src/embeddings/googlevertexai-image.ts
```diff
@@ -0,0 +1,143 @@
...
+      text?: string;
+      image: Buffer;
+    };
+
+export type MediaEmbeddings = {
+  text?: number[];
+  image?: number[];
+};
+
+export class GoogleVertexAIImageEmbeddings
```
jacoblee93, line 48: Do we think it's premature to call this `GoogleVertexAIMultimodalEmbeddings`?

langchain/src/embeddings/googlevertexai-image.ts
jacoblee93: I think I'd rather keep them as separate methods and add the combined one in if there's some benefit to reduce confusion for now - this one is already going to be an odd duck.

jacoblee93 (commented):
A few comments!

langchain/src/embeddings/googlevertexai-image.ts
afirstenberg: I think when I started developing it, they were ambiguous about what its name would be. But they seem to have (mostly) settled on using "multimodal". I'll change the name.

langchain/scripts/create-entrypoints.js
afirstenberg: Will change, tho this was based on llms/openai-chat

afirstenberg:
> A few comments!

These issues should be addressed now.

jacoblee93:
Hey @afirstenberg, I've moved this to `experimental/multimodal_embeddings/googlevertexai` for now so that folks know that breaking changes may occur here in the future as things develop. Otherwise looks great, thank you so much!

jacoblee93:
Added a more end-to-end example with a vector store. Didn't seem to return the image when I searched for `"bird"` but image query -> image seemed to work reasonably well.

PR #2009, opened 2023-07-19
IndexedDB VectorStore

hhjin:
[IndexedDB ](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API)  is a  browser side database that allows web application to store and retrieve large amounts of structured data in browser storage.

The IndexedDBVectorStore integrate IndexedDB with langchain.

### Usage
------------

```javascript
import { IndexedDBVectorStore } from 'langchain/vectorstores/indexeddb';
import { OpenAIEmbeddings } from "langchain/embeddings/openai";
import { Document } from "langchain/document";

const docs=["Hello world","Jessica loves icecream","I like coffee"];
const documents = docs.map((content, idx) => (
    new Document({ pageContent: content, metadata: { source: `idx:${idx}` } })));

const vectorStore = await IndexedDBVectorStore.fromDocuments(documents, new OpenAIEmbeddings() ,   { dbName:"myZSK_ObsidianMds" } );

const similarDocs = await vecStore.similaritySearch("What does Jessica like?", 2);
```

Fixes # (issue)

### DIFF
--- a/langchain/package.json
+++ b/langchain/package.json
@@ -175,6 +175,8 @@
     "vectorstores/qdrant.cjs",
     "vectorstores/qdrant.js",
     "vectorstores/qdrant.d.ts",
+    "vectorstores/indexeddb.d.ts",
+    "vectorstores/indexeddb.js",
     "vectorstores/supabase.cjs",
     "vectorstores/supabase.js",
     "vectorstores/supabase.d.ts",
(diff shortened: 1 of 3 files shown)


jacoblee93:
Thanks for the PR! My gut is that it makes more sense to subclass `MemoryVectorStore` - does that makes sense to you?

jspear29:
Hey. Is there still any appetite for adding this functionality? I just wrote a pretty similar implementation to this.  More than happy to build it as subclass.

PR #2012, opened 2023-07-19
add OPENAI_API_BASE_PATH env variable

choogoo:
Added a name called OPENAI_ API_ BASE_ The environment variable of PATH can be used to connect by setting a proxy service when the user's local network cannot directly access openai.

```
ANTHROPIC_API_KEY=ADD_YOURS_HERE
COHERE_API_KEY=ADD_YOURS_HERE
GOOGLE_PALM_API_KEY=ADD_YOURS_HERE
HUGGINGFACEHUB_API_KEY=ADD_YOURS_HERE
OPENAI_API_KEY=ADD_YOURS_HERE
OPENAI_API_BASE_PATH=https://api.openai.com/v1 # The default address is https://api.openai.com/v1, you can set your own proxy address.
# if AZURE_OPENAI_API_KEY is set, it OPENAI_API_KEY will be ignored.
AZURE_OPENAI_API_KEY=ADD_YOURS_HERE
AZURE_OPENAI_API_INSTANCE_NAME=ADD_YOURS_HERE
...
```


jacoblee93:
This is already possible by passing in a `configuration.basePath` field into the OpenAI class:

https://github.com/hwchase17/langchainjs/blob/main/langchain/src/llms/openai.ts#L205

It's not documented though - would you like to add it here?

https://js.langchain.com/docs/modules/model_io/models/llms/integrations/openai

PR #2019, opened 2023-07-20
fix(embeddings): fix azure embeddings

perezpefaur:
Fix [issue](https://github.com/hwchase17/langchainjs/issues/1850) with azure open AI embeddings to be consistent with [langchain page](https://js.langchain.com/docs/modules/data_connection/text_embedding/integrations/azure_openai). And sorted the imports.

### DIFF
--- a/langchain/src/embeddings/openai.ts
+++ b/langchain/src/embeddings/openai.ts
@@ -1,16 +1,16 @@
+import type { AxiosRequestConfig } from "axios";
 import {
   Configuration,
-  OpenAIApi,
-  CreateEmbeddingRequest,
   ConfigurationParameters,
+  CreateEmbeddingRequest,
+  OpenAIApi,
 } from "openai";
-import type { AxiosRequestConfig } from "axios";
-import { getEnvironmentVariable, isNode } from "../util/env.js";
 import { AzureOpenAIInput } from "../types/openai-types.js";
 import fetchAdapter from "../util/axios-fetch-adapter.js";
+import { OpenAIEndpointConfig, getEndpoint } from "../util/azure.js";
 import { chunkArray } from "../util/chunk.js";
+import { getEnvironmentVariable, isNode } from "../util/env.js";
 import { Embeddings, EmbeddingsParams } from "./base.js";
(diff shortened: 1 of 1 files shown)


jacoblee93:
Thanks for the PR! What's the issue here exactly? I don't think I quite understood from the link.

The base path should include `openai/deployments` by default: https://github.com/hwchase17/langchainjs/blob/main/langchain/src/util/azure.ts#L59

langchain/src/embeddings/openai.ts
```diff
@@ -106,7 +106,12 @@ export class OpenAIEmbeddings
     this.azureOpenAIApiVersion = azureApiVersion;
     this.azureOpenAIApiKey = azureApiKey;
     this.azureOpenAIApiInstanceName = azureApiInstanceName;
-    this.azureOpenAIApiDeploymentName = azureApiDeploymentName;
+    // if openai/deployments is not in the deployment name, add it (for backwards compatibility)
+    if (azureApiDeploymentName?.indexOf("openai/deployments") === -1) {
+      this.azureOpenAIApiDeploymentName = `openai/deployments/${azureApiDeploymentName}`;
```
pavelpiha, line 111: I don't think this will fix. please look to my ticket https://github.com/hwchase17/langchainjs/issues/2105

nimal-outreach:
> Thanks for the PR! What's the issue here exactly? I don't think I quite understood from the link.
> 
> The base path should include `openai/deployments` by default: https://github.com/hwchase17/langchainjs/blob/main/langchain/src/util/azure.ts#L59

If we would provide the API key, base path, and deployment name (all of which are documented on [langchain page](https://js.langchain.com/docs/modules/data_connection/text_embedding/integrations/azure_openai)) that line will not be executed at all


  if (
    azureOpenAIApiKey &&
    azureOpenAIBasePath &&
    azureOpenAIApiDeploymentName
  ) {
    return `${azureOpenAIBasePath}/${azureOpenAIApiDeploymentName}`;
  }

jacoblee93:
We've clarified the page here in #2318 - hope that helps make things clearer:

https://js.langchain.com/docs/modules/data_connection/text_embedding/integrations/azure_openai

PR #2021, opened 2023-07-20
Add Planetscale chat message history

milanjrodd:
Adds Planetscale chat message history by utilizing `@planetscale/database` pkg

To run tests:
- add Planetscale credentials to .env file
- run `yarn test:single -- planetscale.int.test.ts`

All tests were passed successfully.

Similar to dynamodb
https://github.com/hwchase17/langchainjs/pull/1016


docs/extras/modules/memory/integrations/planetscale.mdx
```diff
@@ -0,0 +1,38 @@
+---
+hide_table_of_contents: true
+---
+
+import CodeBlock from "@theme/CodeBlock";
+
+# Planetscale Chat Memory
```
jacoblee93, line 7: Looks like they prefer `PlanetScale` (camel-cased):
https://planetscale.com/

langchain/package.json
```diff
@@ -610,6 +613,7 @@
     "@notionhq/client": "^2.2.5",
     "@opensearch-project/opensearch": "*",
     "@pinecone-database/pinecone": "*",
+    "@planetscale/database": "^1.8.0",
```
jacoblee93, line 616: Can you add as a `devDependency` as well?

langchain/src/stores/message/planetscale.ts
```diff
@@ -0,0 +1,151 @@
...
+      this.client = new PlanetscaleClient(config);
+    } else {
+      throw new Error(
+        "Either a client or config must be provided to PlanetscaleChatMessageHistory"
+      );
+    }
+
+    this.connection = this.client.connection();
+
+    this.tableName = tableName || "stored_message";
```
jacoblee93, line 71: nit: How about `langchain_chat_histories` for the default?

langchain/src/stores/message/planetscale.ts
```diff
@@ -0,0 +1,151 @@
...
+    this.sessionId = sessionId;
+  }
+
+  private async createTableIfNotExists(): Promise<void> {
+    const query = `CREATE TABLE IF NOT EXISTS ${this.tableName} (id BINARY(16) PRIMARY KEY, session_id VARCHAR(255), type VARCHAR(255), content VARCHAR(255), role VARCHAR(255), name VARCHAR(255), additional_kwargs VARCHAR(255));`;
```
jacoblee93, line 76: nit: Store `additional_kwargs` as a JSON field for potential easier querying down the road?

jacoblee93 (commented):
Looks great, a few nits, and we should probably call this `PlanetScale` to match their preferred name.

jacoblee93:
Simple changes so went ahead and made them - thank you!

milanjrodd:
Oh, that was fast 😀

I think there might be a problem with additional_learns field as a varchar255. So needs to be json type. 

Also there is no difference if table id would be type of int, so we can make one sql request instead of two, without altering the table.

Should I open an issue for that?

jacoblee93:
Sure, though I ran the integration test and it seemed to work just fine

milanjrodd:
I'm not sure if json of additional_kwargs always fits in varchar255. And I suppose there is no test for that special case. I mean it might be longer then 255 symbols.

I think sql optimization would be useful for serverless environments, where it can not hold tableCreated state between multiple functions.

That's my first contribution ever, should I create 2 different issues for that?

jacoblee93:
Feel free to just submit another PR as well#

PR #2023, opened 2023-07-20
Update Anthropic SDK to 0.5.7

dqbd:
Addresses the following issues: 
- crash when importing Anthropic SDK to non-NextJS Edge Functions
- Inconsistencies with CJS and ESM version of Anthropic SDK (we can remove the previous workaround)


jacoblee93:
Thank you!

PR #2025, opened 2023-07-20
Feature: Add handleEvent callback

mgce:
Following up on the earlier pull request https://github.com/hwchase17/langchainjs/pull/1964 I introduced one generic `handleEvent` callback. 
It will be responsible for catching both tokens as well as a function call.


vladgolubev:
@jacoblee93, Let me know if there's anything you need to increase the chances of this getting merged! Be it reviewing, testing, etc.

If introducing this change requires changing the upstream python library first, we can do that as well, if it helps.

jacoblee93:
Yeah the biggest blocker is coordination with Python - if you open it there it would definitely speed things up.

vladgolubev:
@jacoblee93 here is the PR for the Python: https://github.com/langchain-ai/langchain/pull/9263

andrewBatutin:
@jacoblee93 can u please advice who from python maintainers can help with python pr review?

vladgolubev:
@jacoblee93 we have an identical Python PR opened. Would you be able to review the proposed small changes anytime soon? Could we do anything else to facilitate this process?

hinthornw:
Hi @vladgolubev and @andrewBatutin thanks for the PR and suggestion! While there is room for a generic onEvent callback, it would have to be completely generic and scoped to the base callback manager rather than the llm run manager.

Would prefer to extend the llm's on new token callback to take in an optional chunk arg that each handler can decide to use

mgce:
Hey @hinthornw, @jacoblee93 
Thanks for your help. I've added new commit to adhere to changes in python lib. Please take a look.

jacoblee93:
Not super pretty, but should be backwards compatible and handle any additional arguments - when we're ready to make breaking changes, we can roll everything into one arg.

Final comments @hinthornw?

hinthornw (approved):

PR #2026, opened 2023-07-20
fix: broken OpenAIModerationChain example

chasemcdo:
This example didn't align with how the code is actually structured. 
It is expected on instantiation of the moderation chain [here](https://github.com/hwchase17/langchainjs/blob/c09f8310a3a2197e98e188167ee39f04dbe18a1d/langchain/src/chains/openai_moderation.ts#L51), but this example is setting it on the call which doesn't do anything.


jacoblee93:
Thank you 🙏

PR #2027, opened 2023-07-20
fix(StructuredOutputParser): Do not obscure OutputParserException

Negan1911:
Hello, first of all, thanks for this great project and for helping AI get closer to Javascript.

There's an error with `StructuredOutputParser` which makes it unable to work with `OutputFixingParser`.

Currently, `StructuredOutputParser` was not "awaiting" the schema check, so the try-catch clause did not run when the schema check failed. Because of this, the error that surfaced was a "ZodError" (in my case using zod).

Because `OutputFixingParser` (rightfully so) expect that the error received to be `OutputParserException` it will not properly try to fix these errors and just bail.

This fix just introduced the "await" so the try-catch clause work and also a regression test to validate.


jacoblee93:
Great find. Thank you!

PR #2045, opened 2023-07-23
fix(memory): allow composite chains to use sub-chains memory.

KamaniBhavin:
Hello, I want to express my gratitude for this amazing project and for your efforts in bringing AI closer to Javascript.

## Issue
The issue was related to the incorrect handling of memory in sub-chains within the sequential chain. This caused memory-based chains to be unable to operate as sub-chains of the sequential chain.

## Fix
With this PR, sequential chains can now effectively utilise sub-chain memory. In the past, there was an issue where using memory in sub-chains of a sequential chain was problematic. The chain validation phase did not take into account the variables present in the memory of sub-chains, which resulted in memory-based chains being unable to function as sub-chains of the sequential chain. This pull request (PR) tackles and resolves this issue.

Twitter:  [@bhavinkamani_](https://twitter.com/bhavinkamani_)

### DIFF
--- a/langchain/src/chains/sequential_chain.ts
+++ b/langchain/src/chains/sequential_chain.ts
@@ -79,7 +79,12 @@ export class SequentialChain extends BaseChain implements SequentialChainInput {
 
     const availableKeys = union(inputKeysSet, memoryKeysSet);
     for (const chain of this.chains) {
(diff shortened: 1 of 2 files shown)


jacoblee93:
Interesting - thank you for finding this!

PR #2046, opened 2023-07-23
Feature: Enable to change the fields name in Milvus 

GalO005:
With this PR we can change the fields name in Milvus DB, for example instead of name: "langchain_primaryid", the developer can change to: "some_primaryid". 
The constructor contains already these fields:  args.textField, args.primaryField and args.vectorField


jacoblee93:
Thanks!

Issue #2016, opened 2023-07-19
Add Support for Raycast AI API


Alireza29675:
I propose adding support for the Raycast LLM API to the langchain JavaScript package so we can use langchain agents within the Raycast environment. Raycast AI API includes a function named `AI.ask(prompt)` that generates a completion.

For more information you may check: https://developers.raycast.com/api-reference/ai

@nfcampos Is it acceptable to have such integration in langchainjs?

jacoblee93:
Would be great and we'd welcome a PR!

Alireza29675:
For sure! @jacoblee93 How can I assign the issue to myself as suggested in contribution guideline?

jacoblee93:
Assigned! Thank you

Alireza29675:
@jacoblee93 This is ready for your review!:)

PR #2049, opened 2023-07-23
feat: Add Raycast AI support

Alireza29675:
Added Raycast AI support: #2016


jacoblee93:
Looks very solid! Thank you

PR #2055, opened 2023-07-24
Adding outputKey to simple_sequential_chain.ts

Ipriyankrajai:
We are using 2 chains and output of first chain is input to second chain. so we need to add outputkey in first LLM chain

Fixes # (issue)


jacoblee93:
Thanks for the PR! Simple sequential is single output, so you do not in this case.

PR #2058, opened 2023-07-24
Support Callback Config in LLMChain

hinthornw:
I think this was unintentionally forgotten during the previous update? I would have expected this to be caught by the linter or compiler since it kind of breaks proper inheritance?


jacoblee93:
> I think this was unintentionally forgotten during the previous update? I would have expected this to be caught by the linter or compiler since it kind of breaks proper inheritance?

Good catch - it's legal for a subclass to only support a narrower type for inherited methods.

PR #2059, opened 2023-07-24
fix: duplicates in Max Marginal Relevance output

archie-swif:
Fixed a bad "in" check for already selected embeddings, added a test for duplicates check


jacoblee93:
Thank you!

PR #2065, opened 2023-07-25
Fix kendra api query - getQueryItemExcerpt

jl4nz:
Fix a small bug when getting an empty 'AnswerText' attr while using Kendra Api query.

Example:
```
TypeError: Cannot read properties of undefined (reading 'Key')
    at CustomAmazonKendraRetriever.getQueryItemExcerpt 
```


jacoblee93:
Thank you!

PR #2068, opened 2023-07-25
fix non saving chroma loc

tonisives:
Chroma doesn't save multi level objects. Seemingly since 0.4.0. See chroma [types.ts](https://github.com/chroma-core/chroma/blob/77c1675c49a1aeab5a1b0c81b4aacc1afdf877b5/clients/js/src/types.ts#L12), which doesn't include object as an allowed value.

The storing succeeds, but the loc object is empty.

If you store the object as a single property (`locFrom`, `locTo`), it works.

This PR maps the langchain `loc` when storing and the new `locFrom` when reading.

### DIFF
--- a/langchain/src/vectorstores/chroma.ts
+++ b/langchain/src/vectorstores/chroma.ts
@@ -111,13 +111,11 @@ export class Chroma extends VectorStore {
     await collection.upsert({
       ids: documentIds,
       embeddings: vectors,
-      metadatas: documents.map(({ metadata }) => {
-        return {
-          ...metadata,
-          locFrom:metadata.loc.lines.from,
-          locTo: metadata.loc.lines.to,
-        }
-      }),
+      metadatas: documents.map(({ metadata }) => ({
+        ...metadata,
+        locFrom: metadata.loc.lines.from,
+        locTo: metadata.loc.lines.to,
+      })),
       documents: documents.map(({ pageContent }) => pageContent),
     });
     return documentIds;
@@ -168,28 +166,28 @@ export class Chroma extends VectorStore {
 
     const results: [Document, number][] = [];
     for (let i = 0; i < firstIds.length; i += 1) {
-      let metadata = {} as any
-      let storedMetadata = firstMetadatas?.[i]
+      let metadata: Document["metadata"] = {};
+      const storedMetadata = firstMetadatas?.[i];
 
-      if (storedMetadata && storedMetadata.locFrom && storedMetadata.locTo ) {
+      if (storedMetadata && storedMetadata.locFrom && storedMetadata.locTo) {
         metadata = {
           ...firstMetadatas[i],
           loc: {
             lines: {
               from: storedMetadata.locFrom,
-              to: storedMetadata.locTo
-            }
-          }
-        }
-        
-        delete metadata.locFrom
-        delete metadata.locTo
+              to: storedMetadata.locTo,
+            },
+          },
+        };
+
+        delete metadata.locFrom;
+        delete metadata.locTo;
       }
 
       results.push([
         new Document({
           pageContent: firstDocuments?.[i] ?? "",
-          metadata: metadata,
+          metadata,
         }),
         firstDistances[i],
       ]);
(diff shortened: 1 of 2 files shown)


jacoblee93:
Thanks for this PR!

I feel like this is a bit brittle and potentially misleading as to Chroma's actual capabilities - maybe we store a stringified object instead?

@jeffchuber would you care to weigh in?

jeffchuber:
what are `locFrom` and `locTo`?

jacoblee93:
> what are `locFrom` and `locTo`?

Ah sorry. Our text splitter will automatically tag split documents with the original location in the text. This gets passed as a `loc` object in the metadata.

This is a specific case, but was wondering more generally about best practices to handle/serialize metadata objects for storage in Chroma?

jeffchuber:
@jacoblee93 got it - in that case I think this is a great and the right thing to do in this case.

tonisives:
> I feel like this is a bit brittle and potentially misleading as to Chroma's actual capabilities - maybe we store a stringified object instead?

Stringified can be more robust and future proof in case langchain adds more fields to the loc object. The only problem with the string is that `where` filter cannot be used when querying. That is because the loc number is transformed into a string.

However, if we use the `locFrom`, then the data is stored in different format and will not respond to `where` queries with the `loc.lines` format. User needs to know that she needs to use the `locFrom` field.

For my use case, I don't currently plan to use `loc.from` with `where` filter. But I can imagine how it would be useful in the future. For instance, when I would want to show results only from the end of the document(summary/conclusion part)

jacoblee93:
Maybe generically flattening objects to string keys like `x.y`? Could run into issues with arrays though... Maybe a problem for later.

If you can run lint + format on this it's good with me!

jeffchuber:
Chroma would like to offer nested/lists in the future - but we are trying to find the right balance of performance, complexity and flexibility. Stay tuned!

jacoblee93:
Looks like it's failing CI:
```
langchain:test: FAIL src/vectorstores/tests/chroma.test.ts
langchain:test:   ● Chroma › should add vectors to the collection
langchain:test: 
langchain:test:     TypeError: Cannot read properties of undefined (reading 'loc')
langchain:test: 
langchain:test:       114 |       metadatas: documents.map(({ metadata }) => ({
langchain:test:       115 |         ...metadata,
langchain:test:     > 116 |         locFrom: metadata.loc.lines.from,
langchain:test:           |                           ^
langchain:test:       117 |         locTo: metadata.loc.lines.to,
langchain:test:       118 |       })),
langchain:test:       119 |       documents: documents.map(({ pageContent }) => pageContent),
langchain:test: 
langchain:test:       at src/vectorstores/chroma.ts:116:27
langchain:test:           at Array.map (<anonymous>)
langchain:test:       at Chroma.addVectors (src/vectorstores/chroma.ts:114:28)
langchain:test:       at Object.<anonymous> (src/vectorstores/tests/chroma.test.ts:50:5)
```

Can you fix and verify with `yarn test:single src/vectorstores/tests/chroma.test.ts` or `yarn test`?

jacoblee93:
> Looks like it's failing CI:
> 
> ```
> langchain:test: FAIL src/vectorstores/tests/chroma.test.ts
> langchain:test:   ● Chroma › should add vectors to the collection
> langchain:test: 
> langchain:test:     TypeError: Cannot read properties of undefined (reading 'loc')
> langchain:test: 
> langchain:test:       114 |       metadatas: documents.map(({ metadata }) => ({
> langchain:test:       115 |         ...metadata,
> langchain:test:     > 116 |         locFrom: metadata.loc.lines.from,
> langchain:test:           |                           ^
> langchain:test:       117 |         locTo: metadata.loc.lines.to,
> langchain:test:       118 |       })),
> langchain:test:       119 |       documents: documents.map(({ pageContent }) => pageContent),
> langchain:test: 
> langchain:test:       at src/vectorstores/chroma.ts:116:27
> langchain:test:           at Array.map (<anonymous>)
> langchain:test:       at Chroma.addVectors (src/vectorstores/chroma.ts:114:28)
> langchain:test:       at Object.<anonymous> (src/vectorstores/tests/chroma.test.ts:50:5)
> ```
> 
> Can you fix and verify with `yarn test:single src/vectorstores/tests/chroma.test.ts` or `yarn test`?

@tonisives

jacoblee93:
Thank you!

PR #2071, opened 2023-07-25
Adding Maximal Marginal Relevance method to MongoDBAtlasVectorSearch

archie-swif:


langchain/src/vectorstores/mongodb_atlas.ts
```diff
@@ -109,6 +110,84 @@ export class MongoDBAtlasVectorSearch extends VectorStore {
...
+   * @param {string} query - Text to look up documents similar to.
+   * @param {number} k - Number of documents to return.
+   * @param {number} [fetchK=20]- Number of documents to fetch before passing to the MMR algorithm.
+   * @param {number} [lambda=0.5] - Number between 0 and 1 that determines the degree of diversity among the results,
+   *                 where 0 corresponds to maximum diversity and 1 to minimum diversity.
+   * @param {MongoDBDocument} preFilter - Optional Atlas Search operator to prefilter on document fields.
+   * @param {MongoDBDocument[]} postFilterPipeline - Optional pipeline of MongoDB aggregation stages following the knnBeta search.
+   *
+   * @returns {Promise<Document[]>} - List of documents selected by maximal marginal relevance.
+   */
```
jacoblee93, line 127: I know it's still in draft but would it make sense to have a switch in the constructor that uses this method by default when similaritySearch is called?

archie-swif: Thank you for the review, @jacoblee93! I have tried to add the "useMMR" flag to the constructor args, but it didn't look very clean, as lambda and k mmr params also need to be specified. After discussing it with our team, we decided to create separate methods for mmr and similarity search.

It may be good to add maxMarginalRelevanceSearch method to the base VectorStore, the same way as in python implementation.

There is some code duplication to execute queries on the database, we will work to extract the common code to a private method as in https://github.com/langchain-ai/langchain/blob/1f40d3e0944604581f76a48fdfc961ac3017c0f2/libs/langchain/langchain/vectorstores/mongodb_atlas.py#L148

jacoblee93: Ah I see. Makes sense then! Let me know when this is ready for review.

archie-swif:
Updated the code to use filter objects

langchain/src/vectorstores/mongodb_atlas.ts
archie-swif: Made changes for MongoDBAtlasVectorSearch to use abstract similaritySearchVectorWithScore for MMR
Added optional maxMarginalRelevanceSearch method to base VectorStore
Think it's ready for review again @jacoblee93! 😀

langchain/src/vectorstores/base.ts
```diff
@@ -126,6 +126,30 @@ export abstract class VectorStore extends Serializable {
...
+   * @param {number} lambda - Number between 0 and 1 that determines the degree of diversity among the results,
+   *                 where 0 corresponds to maximum diversity and 1 to minimum diversity.
+   * @param {this["FilterType"]} filter - Optional filter
+   * @param _callbacks
+   *
+   * @returns {Promise<Document[]>} - List of documents selected by maximal marginal relevance.
+   */
+  async maxMarginalRelevanceSearch?(
+    query: string,
+    k: number,
```
jacoblee93, line 146: Can we roll `k`, `fetchK`, and `lambda` into an `options` arg? Just in case we want to add more in the future.

jacoblee93:
Seems great! One nit, I can fix it myself later today as well.

jacoblee93:
Thank you!

langchain/src/vectorstores/base.ts
archie-swif: Thank you @jacoblee93 !

PR #2072, opened 2023-07-25
Added Azure Blob Storage File and Container Loader

tyumentsev4:
Introducing an ability to load a documents from Azure Blob Storage.

### DIFF
--- a/langchain/.prettierignore
+++ b/langchain/.prettierignore
@@ -1,0 +1,1 @@
+src/load/import_type.d.ts
--- a/langchain/src/load/import_type.d.ts
+++ b/langchain/src/load/import_type.d.ts
@@ -300,9 +300,7 @@ export interface OptionalImportMap {
     | Promise<typeof import("../stores/message/planetscale.js")>;
   "langchain/experimental/multimodal_embeddings/googlevertexai"?:
     | typeof import("../experimental/multimodal_embeddings/googlevertexai.js")
-    | Promise<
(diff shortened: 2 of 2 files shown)


langchain/src/document_loaders/web/azure_blob_storage_container.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export class AzureBlobStorageContainerLoader extends BaseDocumentLoader {
+  private readonly connectionString: string;
+
+  private readonly container: string;
+
+  private readonly unstructuredLoaderOptions: UnstructuredLoaderOptions;
+
+  constructor(
+    connectionString: string,
```
jacoblee93, line 15: Can we make this one `config` object?

langchain/src/document_loaders/web/azure_blob_storage_file.ts
```diff
@@ -0,0 +1,79 @@
...
+  private readonly connectionString: string;
+
+  private readonly container: string;
+
+  private readonly blobName: string;
+
+  private readonly unstructuredLoaderOptions: UnstructuredLoaderOptions;
+
+  constructor(
+    connectionString: string,
```
jacoblee93, line 21: Same as above

jacoblee93 (changes requested):
Looks good overall! Two comments, and can you write some docs under `docs/extras/modules/data_connection/document_loaders/integrations/web_loaders`?

PR #2074, opened 2023-07-25
Add Vectara retriever

3eif:
This PR adds the Vectara retriever which utilizes VectaraStore.


3eif:
You're correct, the asRetriever() method does the same thing. Thanks for the info!

PR #2084, opened 2023-07-26
Enhance PlaywrightWebBaseLoader with HTTP response handling

ywkim:
This PR introduces an enhancement to the PlaywrightWebBaseLoader in LangChain, a feature update that gives you more control over the HTTP responses when loading web pages using the Playwright library.

The original implementation of PlaywrightWebBaseLoader doesn't deal with the Response object returned by Playwright's page navigation function. With this PR, PlaywrightWebBaseLoader is adjusted to capture and use this Response object, enabling additional processing based on the HTTP response.

Here's a snippet of how to use this enhanced functionality:

```typescript
import {
  PlaywrightWebBaseLoader,
  Page,
  Browser,
  Response,
} from "langchain/document_loaders/web/playwright";

const loader = new PlaywrightWebBaseLoader("https://your-web-url", {
  async evaluate(page: Page, browser: Browser, response: Response | null) {
    let contentType = 'text/html';
    if (response != null) {
      contentType = response.headers()['content-type'];
    }
    if (contentType.includes('text/html')) {
      // Do something specific for HTML content
    }
    // Do something else for non-HTML content
  },
});
...
```

By providing more control over the handling of HTTP responses, this enhancement broadens LangChain's web scraping capabilities to cater a wider range of use cases.

### DIFF
--- a/docs/extras/modules/data_connection/document_loaders/integrations/web_loaders/web_playwright.md
+++ b/docs/extras/modules/data_connection/document_loaders/integrations/web_loaders/web_playwright.md
@@ -53,7 +53,7 @@ type PlaywrightWebBaseLoaderOptions = {
 
 2. `gotoOptions`: an optional object that specifies additional options to pass to the page.goto() method. This can include options such as the timeout option to specify the maximum navigation time in milliseconds, or the waitUntil option to specify when to consider the navigation as successful.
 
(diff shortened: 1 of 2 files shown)


jacoblee93:
Looks like a good add, thank you!

PR #2085, opened 2023-07-26
Update refine_prompts.ts

bvasilenko:
Came across a bug with `gpt-3.5-turbo-16k` and with other LLMs:

`QARefineChain` overwrites a good answer previously found with a useless message about no further refinement needed.
After a bit of debugging, I've realised that the bot seems to ignore this line `If the context isn't useful, return the original answer.` 

Reproduction:
a. https://platform.openai.com/playground/p/bjJa9sc7rb6z1Y9dFpsi4uvH
b. same prompt with local LLMs

Suggested solution:
Improve the problematic line with `you must` words.

Demo:
a. https://platform.openai.com/playground/p/756t4wSvzDJiUHLlkWiUnZDY
b. try yourself with another LLM, for instance LLaMA

### DIFF
--- a/langchain/src/chains/question_answering/refine_prompts.ts
+++ b/langchain/src/chains/question_answering/refine_prompts.ts
@@ -19,7 +19,7 @@ We have the opportunity to refine the existing answer
 {context}
 ------------
 Given the new context, refine the original answer to better answer the question. 
-If the context isn't useful, return the original answer.`;
+You must provide a response, either original answer or refined answer.`;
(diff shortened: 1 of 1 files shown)


jacoblee93:
Thanks for the PR! I think this is a bit risky in that it could change behavior for people relying on it currently with e.g. gpt-4.

Instead, I'd like to add this in a "Prompt customization" section to the docs here: 

https://js.langchain.com/docs/modules/chains/document/refine

I'll get to it today and tag you here.

PR #2087, opened 2023-07-26
Added Azure Blob Storage File and Container Loader

tyumentsev4:
Introducing an ability to load a document from Azure Blob Storage.


langchain/src/document_loaders/web/azure_blob_storage_container.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export class AzureBlobStorageContainerLoader extends BaseDocumentLoader {
+  private readonly connectionString: string;
+
+  private readonly container: string;
+
+  private readonly unstructuredLoaderOptions?: UnstructuredLoaderOptions;
+
+  constructor(
+    { connectionString, container }: AzureBlobStorageContainerConfig,
```
jacoblee93, line 20: My preference would be to have one config object:
```
{
  azureConfig: AzureBlobStorageContainerConfig,
  unstructuredConfig: UnstructuredLoaderOptions
}
```
to allow for easier migration for future changes

langchain/src/document_loaders/web/azure_blob_storage_file.ts
```diff
@@ -0,0 +1,83 @@
...
+  private readonly connectionString: string;
+
+  private readonly container: string;
+
+  private readonly blobName: string;
+
+  private readonly unstructuredLoaderOptions?: UnstructuredLoaderOptions;
+
+  constructor(
+    { connectionString, container, blobName }: AzureBlobStorageFileConfig,
```
jacoblee93, line 27: See above

jacoblee93 (commented):
One nit, otherwise LGTM!

I can make the change later today as well.

jacoblee93:
Thank you!

PR #2089, opened 2023-07-26
Add prompt customization example to refine chain docs

jacoblee93:
CC @bvasilenko


jacoblee93:
Doing this in lieu of #2085

PR #2093, opened 2023-07-26
Feat: Weaviate delete with filter

neebdev:
Support for deleting objects based on either their ids or a filter condition in weaviate vectorstore.


langchain/src/vectorstores/weaviate.ts
```diff
@@ -155,13 +155,29 @@ export class WeaviateStore extends VectorStore {
...
-        .deleter()
+  async delete(params: { ids: string[] }): Promise<void>;
+
+  async delete(param: { filter: WeaviateFilter }): Promise<void>;
+
+  async delete(params: {
+    ids?: string[];
+    filter?: WeaviateFilter;
+  }): Promise<void> {
+    const { ids, filter } = params || {};
```
jacoblee93, line 166: `params` should always be defined no?

jacoblee93:
One small nit

langchain/src/vectorstores/weaviate.ts
neebdev: yeah that or is left redundant. I was first creating it optional for deleting entire index.

neebdev: I also needed to delete a whole class(index) in weaviate. I was not sure which way was most appropriate. Is it fine to make a static function which takes index name and delete it?

jacoblee93: Convention at the moment is a `deleteAll` flag like this: https://github.com/langchain-ai/langchain/blob/master/libs/langchain/langchain/vectorstores/pinecone.py#L385

neebdev: https://weaviate.io/developers/weaviate/manage-data/classes#delete-a-class

neebdev: If I am not wrong, there is no way to delete all objects while keeping a class in weaviate. You can delete an entire class which is called index in langchain. So implementing a deleteAll which delete the whole index will render the weaviate store instance unusable.

jacoblee93: Ah ok, sounds a bit more specific. A special method could make sense in this case then, although I assume you'll need auth so a static method might not work well. `deleteClass` perhaps? Or is there a way to do this with a filter?

neebdev: maybe a static function which takes the client and index name?

neebdev: another solution would be to delete the class and then create again with same parameter. Doesn't sound right. And someone might need to delete the class and not just the objects. So I guess a static method with client for auth and index name for deletion which returns void is the way to go?

neebdev:
Added the static method. I named it `deleteIndex` for consistency.

neebdev:
Should we add `deleteIndex` as a generic function in vectorstore base because most vector databases has the feature of deleting an index?

neebdev:
@jacoblee93 Let me know if it looks good or it needs changes. Thanks

jacoblee93:
Looking at it now, if this renders the vector store unusable (which it does, right?) without another API call, it seems a bit out of scope for the class. It's a one-liner that someone can do with the client directly anyway.

neebdev:
Good point. It would render an existing index unusable but since it returns void so it can be used to delete a single index with the name. But for now I removed it.

jacoblee93:
Thank you!

PR #2094, opened 2023-07-26
Milvus - update object type in fromTexts method

GalO005:
update object type of dbConfig in fromTexts method, so will be consistent where dbConfig defined.


jacoblee93:
Good catch, thank you!

PR #2103, opened 2023-07-27
Feat: make projectId configurable for vertex ai

erictt:
This PR allows overwriting the auth options for Vertex AI that differ from the variables from environment such as `GOOGLE_APPLICATION_CREDENTIALS`. 

ref: https://github.com/googleapis/google-auth-library-nodejs/blob/main/src/auth/googleauth.ts#L85

### DIFF
--- a/langchain/src/types/googlevertexai-types.ts
+++ b/langchain/src/types/googlevertexai-types.ts
@@ -1,11 +1,11 @@
 import { BaseLLMParams } from "../llms/index.js";
+import { GoogleAuthOptions } from "google-auth-library";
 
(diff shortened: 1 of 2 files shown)


jacoblee93:
Actually, I wonder if it's possible to just have this accept all properties `.auth` accepts?

erictt:
@jacoblee93 Yeah i agree. i just pushed another commit to allow all options from GoogleAuthOptions.

afirstenberg:
This looks good. Thank you for the update!

jacoblee93:
Thank you!

PR #2106, opened 2023-07-27
Single call protocol in JS

jacoblee93:
Enables more widespread streaming, batch calls.

### DIFF
--- a/langchain/src/llms/base.ts
+++ b/langchain/src/llms/base.ts
@@ -87,6 +87,13 @@ export abstract class BaseLLM<
     yield result.generations[0][0];
   }
 
+  async *_createByteStreamAsyncGenerator(
+    input: BaseLanguageModelInput,
+    options?: CallOptions
+  ): AsyncGenerator<string> {
+    yield* this._createStreamAsyncGenerator(input, options);
+  }
+
   async *_createStreamAsyncGenerator(
     input: BaseLanguageModelInput,
     options?: CallOptions
--- a/langchain/src/llms/base.ts
+++ b/langchain/src/llms/base.ts
@@ -5,12 +5,14 @@ import {
   BaseMessage,
   BasePromptValue,
   Generation,
+  GenerationChunk,
   LLMResult,
   RUN_KEY,
 } from "../schema/index.js";
 import {
   BaseLanguageModel,
   BaseLanguageModelCallOptions,
+  BaseLanguageModelInput,
   BaseLanguageModelParams,
 } from "../base_language/index.js";
 import {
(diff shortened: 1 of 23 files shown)


langchain/src/schema/runnable.ts
```diff
@@ -0,0 +1,78 @@
...
+      const batchPromises = inputs
+        .slice(i, i + batchSize)
+        .map((input, i) => this.invoke(input, configList[i]));
+      const batchResult = await Promise.all(batchPromises);
+      batchResults.push(batchResult);
+    }
+    return batchResults.flat();
+  }
+
+  async *_createAsyncGenerator(
```
jacoblee93, line 55: Would love a better name for this if you have suggestions @nfcampos

langchain/src/llms/openai-chat.ts
```diff
@@ -50,11 +50,9 @@ export interface OpenAIChatCallOptions extends OpenAICallOptions {
  * @augments AzureOpenAIChatInput
  */
 export class OpenAIChat
-  extends LLM
+  extends LLM<OpenAIChatCallOptions>
   implements OpenAIChatInput, AzureOpenAIInput
 {
-  declare CallOptions: OpenAIChatCallOptions;
```
nfcampos, line 56: We dont need this anymore?

langchain/src/llms/base.ts
```diff
@@ -63,9 +65,117 @@ export abstract class BaseLLM extends BaseLanguageModel {
...
+      let parsedOptions: CallOptions;
+      if (options?.timeout && !options.signal) {
+        parsedOptions = {
+          ...options,
+          signal: AbortSignal.timeout(options.timeout),
+        };
+      } else {
+        parsedOptions = (options ?? {}) as CallOptions;
+      }
+      delete parsedOptions.tags;
```
nfcampos, line 126: we probably want to move this logic to a helper method?

langchain/src/llms/openai-chat.ts
jacoblee93: Ah thanks, good catch

jacoblee93: Yeah, it's now declared with the type parameter on the superclass here: https://github.com/hwchase17/langchainjs/pull/2106/files/381bc22a2dd83824d3ff53574c78e3711e3d3c54#diff-0498c91d7202afcc70230e285dd4e46dfbd6c2525556c3b1cba7c3ce22e94671R115

jacoblee93: Screenshot 2023-07-28 at 10 14 00 AM

langchain/src/llms/base.ts
jacoblee93: Ah right, will do!

PR #2119, opened 2023-07-29
Feature - add filter type for Milvus DB

GalO005:
A filter type has been added to Milvus DB, so we can pass filter object with a property called `filterStr`

### DIFF
--- a/langchain/src/vectorstores/milvus.ts
+++ b/langchain/src/vectorstores/milvus.ts
@@ -23,6 +23,10 @@ export interface MilvusLibArgs {
   password?: string;
 }
 
+export interface MilvusFilterType {
+  filterStr: string;
+}
+
 type IndexType =
   | "IVF_FLAT"
   | "IVF_SQ8"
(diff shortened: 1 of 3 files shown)


langchain/src/vectorstores/milvus.ts
```diff
@@ -223,6 +232,7 @@ export class Milvus extends VectorStore {
       output_fields: outputFields,
       vector_type: DataType.FloatVector,
       vectors: [query],
+      filter: filterStr,
```
jacoblee93, line 235: Just checking - passing an empty string has the same behavior as undefined right?

GalO005: Yeah, sure.

jacoblee93: Hey actually it looks like Milvus just accepts a string: 

https://milvus.io/docs/search.md#Conduct-a-vector-search

I think we can just have `FilterType: string` here - if they add more complex objects in the future we can shim it.

GalO005: No problem, but I think that `FilterType` is defined as object type (and thus I created interface), I can add to base class `object | string` what do you think about that?

GalO005: @jacoblee93 I changed it.

jacoblee93:
Thank you!

PR #2129, opened 2023-08-01
Broader Runnable support, compositional pieces

jacoblee93:
Adds RunnableSequence, RunnableMap, RunnableLambda, and RunnablePassthrough and associated compositional `.pipe` methods, also makes retrievers and output parsers runnable.

### DIFF
--- a/langchain/src/schema/output_parser.ts
+++ b/langchain/src/schema/output_parser.ts
@@ -1,13 +1,16 @@
 import { Callbacks } from "../callbacks/manager.js";
 import { BasePromptValue, Generation, ChatGeneration } from "./index.js";
-import { Serializable } from "../load/serializable.js";
+import { Runnable, RunnableConfig } from "./runnable.js";
 
 /**
  * Options for formatting instructions.
  */
 export interface FormatInstructionsOptions {}
 
-export abstract class BaseLLMOutputParser<T = unknown> extends Serializable {
+export abstract class BaseLLMOutputParser<T = unknown> extends Runnable<
+  Generation[] | ChatGeneration[],
+  T
+> {
   abstract parseResult(
     generations: Generation[] | ChatGeneration[],
     callbacks?: Callbacks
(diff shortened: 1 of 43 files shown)


langchain/src/schema/output_parser.ts
```diff
@@ -20,9 +28,31 @@ export abstract class BaseLLMOutputParser<T = unknown> extends Serializable {
...
+    input: string | BaseMessage,
+    options?: RunnableConfig
+  ): Promise<T> {
+    if (typeof input === "string") {
+      return this._callWithConfig(
+        async (input: string): Promise<T> =>
+          this.parseResult([{ text: input }]),
+        input,
+        options
+      );
```
nfcampos, line 42: Add here run type per https://github.com/langchain-ai/langchain/pull/8488

langchain/src/schema/runnable.ts
```diff
@@ -82,4 +106,401 @@ export abstract class Runnable<
...
+      });
+    }
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  static isInstance(thing: any): thing is RunnableSequence {
+    return Array.isArray(thing.middle) && Runnable.isInstance(thing);
+  }
+
+  static fromRunnables<RunInput, RunOutput>([first, ...runnables]: [
```
nfcampos, line 376: Rename to `.from()` ? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from

langchain/src/schema/runnable.ts
```diff
@@ -82,4 +106,401 @@ export abstract class Runnable<
...
+  }
+}
+
+export class RunnableLambda<RunInput, RunOutput> extends Runnable<
+  RunInput,
+  RunOutput
+> {
+  lc_namespace = ["schema", "runnable"];
+
+  constructor(protected func: RunnableFunc<RunInput, RunOutput>) {
```
nfcampos, line 449: This needs to be an object with single key func

langchain/src/schema/runnable.ts
```diff
@@ -82,4 +106,401 @@ export abstract class Runnable<
...
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  Record<string, any>
+> {
+  lc_namespace = ["schema", "runnable"];
+
+  lc_serializable = true;
+
+  protected steps: Record<string, Runnable<RunInput>>;
+
+  constructor(steps: Record<string, RunnableLike<RunInput>>) {
```
nfcampos, line 401: This needs to be an object with single key `steps`. Otherwise serialisation won't match py

jacoblee93: Ah right, thank you!

PR #2130, opened 2023-08-01
fix(docs): fix abnormal behavior when opening docs submenu

Tomotoes:
Bugs from the current documentation site:
20230801145556_rec_

This was a docusaurus bug related to prefers-reduced-motion nested collapsible bug, see this: https://github.com/facebook/docusaurus/issues/8932


jacoblee93:
Thank you!

PR #2132, opened 2023-08-01
Fix streaming errors in Node (Express, Next, etc.) environments

### LINKED ISSUE #548: Stream ConversationalRetrievalQAChain is not working
I'm using a pinecone vector with OpenAIChat model in a ConversationalRetrievalQAChain, it works as intended when im not streaming the output however if I change the OpenAIChat  to streaming true as per the below code snippet it fails with the following error
"TypeError: stream.getReader is not a function"
```
 const model = new OpenAIChat({
        streaming: true,
  callbackManager: CallbackManager.fromHandlers({
    async handleLLMNewToken(token: string) {
      console.log(token);
    },
  }),
    });
```
This is in a nextjs API route

rishi-raj-jain:
Resolves #548

Fixes streaming errors as seen in #548


jacoblee93:
What environment does this occur in? And is it just calling `OpenAI` with `stream: true`?

rishi-raj-jain:
Hey Jacoblee,

While I was building an express app, I found this. Then I tried the same in Next.js and that too failed to find the stream.getReader. Digging into the issues and discussions I found the solution that worked just as expected!

jacoblee93:
Gotcha - what version of Node are you on?

rishi-raj-jain:
@jacoblee93 

I'm on Node node v18.12.1 & npm 8.19.2.

Happy to be contributing and get the PR running. Let me know if I need to do any changes on my end.

jacoblee93:
Can you help me repro this?

Is it installing `node-fetch` and overwriting the namespace somehow? `stream.getReader` should be defined since Node 16.5.

rishi-raj-jain:
I found while reproducing that my code had  https://github.com/ajcwebdev/langchain-node/blob/main/src/index.mjs#L1-L4 to make it suitable for every kind of environment (Node, Workers) and once I removed it worked just as fine. Let me know if I should close this. I still think adding it the other code as a fallback would be helpful. Happy to be of help.

jacoblee93:
Ah ok, so this is if you override the fetch adapter. Yeah I think this could be valuable still but let me or @dqbd it out a bit!

jacoblee93:
Thank you @rishi-raj-jain and @dqbd!

rishi-raj-jain:
Glad to see that this got in! Hope this makes it default smoother!

Issue #2113, opened 2023-07-28
cosineSimilarity(), maximalMarginalRelevance(), and any util can't be imported


afirstenberg:
The `cosineSimilarity()` and `maximalMarginalRelevance()` functions were added with the `math_utils.ts` file in #1973.

However, this function can't be imported outside of langchain since it isn't included in the package.json "exports". In fact, nothing from the `util` directory is available since they are only in the `dist` directory.

I'm not familiar enough with how "exports" or any of this is built, but it seems like `math_utils` as a minimum should be included in this process, and there is a good case that all of `util` should be.

jacoblee93:
Yeah, I wouldn't have a problem with adding it as an entrypoint - what would your use-case be?

afirstenberg:
In this case, I'm using it to demonstrate how similar sets of embeddings are to each other to demonstrate or evaluate how effective the embedding library is. In theory, this could be done by using a vector database, but this also guarantees to give us a numeric evaluation of how similar the vectors are.

PR #2150, opened 2023-08-03
More mathutils

afirstenberg:
Several things around the math_utils component:

- Adds a matrixFunc() function which can take a VectorFunction and apply it in a row-wise manner
- Adds innerProduct() and euclideanDistance() methods that work similar to cosineSimilarity()
- All of these new functions call matrixFunc() with the appropriate function from ml-distance
- Added a normalize() function that works over a matrix
- Other code refactoring

This also addresses issue #2113


langchain/scripts/create-entrypoints.js
```diff
@@ -189,6 +189,8 @@ const entrypoints = {
   "stores/message/ioredis": "stores/message/ioredis",
   "stores/message/upstash_redis": "stores/message/upstash_redis",
   "stores/message/planetscale": "stores/message/planetscale",
+  // utilities
+  "util/math_utils": "util/math_utils",
```
jacoblee93, line 193: I think this would be nice to export as `util/math`?

I can rename the files

jacoblee93:
Thank you!

PR #2155, opened 2023-08-03
Add check to stop empty documents being returned. Fixes #2144

### LINKED ISSUE #2144: Unstructured file loader allowing pageContent to be undefined
I've noticed that sometimes a `Document` returned by the `Unstructured` file loader will have an undefined `pageContent` property.

image

As a result, when being passed to `OpenAiEmbeddings` `embedDocuments()`, the `replace()` call fails as the passed texts property will be `undefined`.

image

Could this be fixed by either:

1. Preventing the loaders from building an `undefined` `pageContent` property in the first place? Is there any valid use case for having `pageContent` as `undefined`?
2. Checking each `texts` isn't `undefined` before trying to call `replace()` on them within `OpenAiEmbeddings` `embedDocuments()`

Thanks!

simon-fernandes:
Fixes #2144  (issue)

### DIFF
--- a/langchain/src/document_loaders/fs/unstructured.ts
+++ b/langchain/src/document_loaders/fs/unstructured.ts
@@ -171,15 +171,15 @@ export class UnstructuredLoader extends BaseDocumentLoader {
     const documents: Document[] = [];
     for (const element of elements) {
       const { metadata, text } = element;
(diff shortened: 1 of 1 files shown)


jacoblee93:
Thank you!

PR #2156, opened 2023-08-03
Allow runnables to implement transform streaming

jacoblee93:
Primary use-case is to allow LLM/chat model streaming with an output parser that can handle streamed generated chunks.

### DIFF
--- a/langchain/src/schema/output_parser.ts
+++ b/langchain/src/schema/output_parser.ts
@@ -102,24 +102,14 @@ export abstract class BaseOutputParser<
 }
 
 /**
- * OutputParser that parses LLMResult into the top likely string.
+ * Class to parse the output of an LLM call that also allows streaming inputs.
  */
-export class StringOutputParser extends BaseOutputParser<string> {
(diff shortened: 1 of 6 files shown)


langchain/src/schema/output_parser.ts
```diff
@@ -116,6 +116,70 @@ export class StringOutputParser extends BaseOutputParser<string> {
...
+      runType: "parser",
+    });
+  }
+}
+
+/**
+ * OutputParser that parses LLMResult into the top likely string and
+ * encodes it into bytes.
+ */
+export class EncodingOutputParser extends BaseOutputParser<Uint8Array> {
```
nfcampos, line 147: Might we want to call this BytesOutputParser or TextEncoderOutputParser

nfcampos (approved):
This is great!

langchain/src/schema/output_parser.ts
jacoblee93: It accepts chat chunks so I think BytesOutputParser is better

PR #2162, opened 2023-08-04
Extending the base document loader class with EventEmitter

skarard:
This PR extends the existing `BaseDocumentLoader` class to add `EventEmitter` functionality. The new class `BaseDocumentLoaderWithEventEmitter` wraps an instance of `EventEmitter`, providing a common interface for event handling and emission. 

The addition of `EventEmitter` allows for asynchronous event handling, which can be especially useful when loading large documents or when operations must be performed on documents once they're loaded. 

With these changes, users of the `BaseDocumentLoader` class can easily add and manage event listeners, allowing for more flexible and responsive application designs.

I have included an example and a test.

### DIFF
--- a/examples/src/document_loaders/base_with_event_emitter.ts
+++ b/examples/src/document_loaders/base_with_event_emitter.ts
@@ -1,0 +1,108 @@
+import { BaseDocumentLoaderWithEventEmitter } from "langchain/document_loaders/base_with_event_emitter";
+import { Document } from "langchain/document";
+
+class ExampleTextLoaderWithEvents extends BaseDocumentLoaderWithEventEmitter {
+  private documentsTotal: number = 0;
+
+  private documents: Document[] = [];
+
+  constructor(public textArray: (string | string[])[]) {
+    super();
+  }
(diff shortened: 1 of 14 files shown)


skarard:
The Environment Tests fail with:
```
"EventEmitter" is not exported by "__vite-browser-external", imported by "node_modules/langchain/dist/document_loaders/base_with_event_emitter.js".
```
It's not clear to me what this is required for, though it appears to be an in browser build of LangChain. Is there someway of specifying that this is a node only package?

jacoblee93:
Interesting! Will look today.

jacoblee93:
Hey thought about it a bit - can we just make this localized to the Notion loader for now? 

Would be in favor of keeping it simple - you wouldn't even need to extend event emitter, you could just have an `onDocumentLoaded` callback function or something?

Portability issues and I'm worried it's not quite generalizable yet.

skarard:
Sure, no problem.

I've closed this PR, and will continue in my Notion API upgrade branch on PR #2164

PR #2164, opened 2023-08-04
onDocumentLoaded callback, rate limits  & queue for NotionAPI document Loader

skarard:
The updates to the `NotionAPILoader` class uses a queue to manage loading pages from the Notion API. Pages or page IDs are added to the queue during instantiation or during the loading process if child pages are detected while loading a database. The queue is used to ensure orderly access and management of the pages to be loaded.

This class also supports a callback for when each document is loaded. This call back return the current queue position, the total queue length, the current page title, and the root page title.

To manage the rate at which requests to the API are made, the `NotionAPILoader` uses the Bottleneck library. This library allows for rate limiting, ensuring that the rate of API requests doesn't exceed the API's rate limits. It makes use of a limiter instance from Bottleneck to schedule requests and load blocks in a rate-limited manner.


skarard:
At the request of @jacoblee93 I have removed the EventEmitter and replaced it with an onDocumentLoaded callback.

skarard:
I believe this is ready for final review and merge.

langchain/src/document_loaders/web/notionapi.ts
```diff
@@ -127,143 +205,160 @@ export class NotionAPILoader extends BaseDocumentLoader {
...
+      })
+      .catch((error: APIResponseError) => error);
+
+    const [resPage, resDatabase] = await Promise.all([
+      resPagePromise,
+      resDatabasePromise,
+    ]);
+    const errors = [resPage, resDatabase].filter(isErrorResponse);
+
+    if (errors.length === 2) {
```
jacoblee93, line 343: Exactly 2?

langchain/src/document_loaders/web/notionapi.ts
```diff
@@ -11,44 +15,118 @@ import type {
...
 type PageObjectResponse = GuardType<typeof isFullPage>;
+type DatabaseObjectResponse = GuardType<typeof isFullDatabase>;
+
+type GetResponse =
+  | GetBlockResponse
+  | GetPageResponse
+  | GetDatabaseResponse
+  | APIResponseError;
+
+// const isBlockResponse = (res: GetResponse): res is GetBlockResponse =>
```
jacoblee93, line 41: Remove commented out methods

langchain/src/document_loaders/web/notionapi.ts
```diff
@@ -11,44 +15,118 @@ import type {
   MdBlock,
 } from "notion-to-md/build/types";
 
-import { BaseDocumentLoader } from "../base.js";
+import Bottleneck from "bottleneck";
```
jacoblee93, line 18: We already have a concurrency library:

https://www.npmjs.com/package/p-queue

Would prefer to use that if possible instead of adding a new peer dep?

langchain/src/document_loaders/web/notionapi.ts
skarard: Yes, both resPage and resDatabase must error to throw an error. This part of the code determines whether the notion ID is a page or database. Only when both are errors will it throw. 

The `errors` variable is set in the line above and filter the results for both page and database calls and includes the filter with the error type guard. One is always expected to error.

langchain/src/document_loaders/web/notionapi.ts
skarard: I will look into it, the preference is always to reduce dependencies. 👍

langchain/src/document_loaders/web/notionapi.ts
skarard: Sure, I left it in as it's a useful reference for anyone wanting to use the notion SDK in the future, but this isn't the place for it. I'll submit a PR to notion-sdk to deal with types more succinctly.

langchain/src/document_loaders/web/notionapi.ts
jacoblee93: Ah gotcha. Maybe add a comment there?

langchain/src/document_loaders/web/notionapi.ts
jacoblee93: You can use it with this util class we have actually, which might be even easier:

https://github.com/hwchase17/langchainjs/blob/main/langchain/src/util/async_caller.ts

skarard: I have replaced bottleneck with the AsyncCaller utility and it appears to be working without issue.

langchain/src/document_loaders/web/notionapi.ts
skarard: I added a comment to clarify the error checking

langchain/src/document_loaders/web/notionapi.ts
skarard: I have removed the unused methods.

skarard:
There were a couple of conflicts from the docstrings commit and I updated the docs string to reflect the new usage.

skarard:
@jacoblee93 I think we're good to go.

jacoblee93:
I think the diff went a little crazy - are you ok with cherry picking and recreating this PR?

skarard:
I'll fix this up right away.

skarard:
I haven't woken up yet, as I missed a couple of the doc strings in the merge. It's all good now.

jacoblee93:
Looks good! I did see a few of these:
```
    @notionhq/client warn: request fail {
      code: 'object_not_found',
      message: 'Could not find page with ID: 42485ae5-a22d-4fbe-8576-c29656900f54. Make sure the relevant pages and databases are shared with your integration.'
    }

```

Might be good to suppress them, but we can do that in another PR since this one has gone on a while and I know you're keen to get it merged.

langchain/src/document_loaders/web/notionapi.ts
```diff
@@ -11,52 +15,123 @@ import type {
...
+    this.pageQueue = [];
+    this.pageCompleted = [];
+    this.pageQueueTotal = 0;
+    this.documents = [];
+    this.rootTitle = "";
+    this.onDocumentLoaded = options.onDocumentLoaded ?? ((_ti, _cu) => {});
+  }
+
+  private addToQueue(...items: string[]) {
+    const deDuped = items.filter(
```
jacoblee93, line 130: Nice!

jacoblee93:
I guess the idea is you don't know ahead of time whether it's a page or a block? So you try them both?

jacoblee93:
Or actually, would it be possible to add a check? Feel like it'd freak people out a bit.

skarard:
Unfortunately this is the only check, and it only happens once per request.

I had it that I would first check for a page, then continue but I realised that they are light calls to the notion API, only returning the page properties, or the database schema. It's faster to do both calls at the same time with Promise.all() then continue the one without errors.

I deprecated the explicit page or database filter, because it complicates the code, doesn't add any time efficiency and can lead to a bad user experience as there is no difference between a page id or a database id.

I can however supress notion's logging, so the user won't be notified.

jacoblee93:
Makes sense - suppressing is fine in this case.

Thank you! Will get this merged today.

PR #2167, opened 2023-08-05
Adds Ollama chat model and llm

jacoblee93:


langchain/src/util/ollama.ts
```diff
@@ -0,0 +1,84 @@
...
+
+export async function* createOllamaStream(
+  baseUrl: string,
+  params: OllamaRequestParams,
+  options: OllamaCallOptions
+): AsyncGenerator<OllamaGenerationChunk> {
+  const response = await fetch(`${baseUrl}/api/generate`, {
+    method: "POST",
+    body: JSON.stringify(params),
+    headers: {
```
jmorganca, line 59: small note: you shouldn't need this header since it returns a stream of events (vs json)

jmorganca:
Looks great!

langchain/src/util/ollama.ts
jacoblee93: Wouldn't it refer to the request body's content type rather than the response?

jmorganca: That's right, sorry! Misread this as the `Accept` header.

PR #2180, opened 2023-08-06
Add: USearch Vector Store

ashvardanian:
## Description

I am excited to propose an integration with USearch, a lightweight vector-search engine available for both Python and JavaScript, among other languages.

## Dependencies

It introduces a new NPM dependency - `usearch`. I am unsure if I should include it in `package.lock`. Please let me know.

## Profiles

- Maintainers: @ashvardanian @davvard
- Twitter handles: @ashvardanian @unum_cloud


langchain/src/vectorstores/usearch.ts
```diff
@@ -0,0 +1,202 @@
...
+    const args: USearchArgs = {
+      docstore: dbConfig?.docstore,
+    };
+    const instance = new this(embeddings, args);
+    await instance.addDocuments(docs);
+    return instance;
+  }
+
+
+  static async importUSearch(): Promise<{ Index: typeof Index }> {
```
jacoblee93, line 185: You can actually just import this normally:

https://github.com/hwchase17/langchainjs/blob/main/.github/contributing/INTEGRATIONS.md#creating-a-separate-entrypoint

jacoblee93:
Thanks for the PR! Can you add docs under `docs/extras` and run `yarn format` and `yarn lint`?

ashvardanian:
Hey! Thank you, @jacoblee93! On our side, @davvard has implemented the docs and a few patches. All the tests pass when run locally. Let us know, if something else is needed.

jacoblee93:
Thank you! This looks cool, love having more local vector stores!

Made a few build-related changes but looks good. Will aim to merge and cut a release tonight.

aehlke:
This says nodejs only but usearch is also available in wasm for browsers / other envs

AndiLeni:
I tries using this store today as I really like local file based stores. However, the USearch class seems to lack a `load` to load a saved store from the file system? Am I missing something here?

ashvardanian:
Hi @AndiLeni! The [`Index.load()` is present in USearch](https://github.com/unum-cloud/usearch/blob/ce54b814a8a10f4c0c32fee7aad9451231b63f75/javascript/usearch.js#L352). Is that what you are looking for?

AndiLeni:
@ashvardanian no, I am trying to use this langchain integration which you created with this pull request.
This langchain USearch class has a save() methof but no load method, so I can not use this integration to load existing stores. I can just create them.

ashvardanian:
Gotcha! I am overbooked the next few weeks, so may not be able to provide a patch any time soon. But it should be easy to add in case you are already working with the codebase 🤗

PR #2184, opened 2023-08-07
Add Vectara addFile function

3eif:
This PR adds support for uploading files to Vectara via the Vectara vector store. https://docs.vectara.com/docs/api-reference/indexing-apis/file-upload/file-upload

The function uses the node-fetch package because it does not work with the default fetch API.


jacoblee93:
Thanks for the PR! Why doesn't it work with default fetch?

langchain/src/vectorstores/vectara.ts
```diff
@@ -163,6 +168,61 @@ export class VectaraStore extends VectorStore {
     }
   }
 
+  async addFiles(
```
jacoblee93, line 171: Can this just use the text loader?

https://js.langchain.com/docs/modules/data_connection/document_loaders/integrations/file_loaders/text

3eif:
> Thanks for the PR! Why doesn't it work with default fetch?

Not sure. I just get a 400 error saying the request entity is too large. I've only had this issue with the default fetch, everything else works fine (I've tried node-fetch and axios).

langchain/src/vectorstores/vectara.ts
3eif: The purpose of the addFiles function is to upload and index any document not just text. Vectara accepts files in various formats like PDF, PPT, etc. Those files are then processed to extract and chunk the text. This simplifies the steps for developers. 

The Vectara API call is: https://docs.vectara.com/docs/api-reference/indexing-apis/file-upload/file-upload

The python version of this function can be found here: https://github.com/langchain-ai/langchain/blob/master/libs/langchain/langchain/vectorstores/vectara.py#L134

cc @ofermend

jacoblee93: Gotcha, ok

jacoblee93:
Can you see if you can dig in a bit and figure out why? Would prefer to not add a dep if we can help it.

langchain/src/vectorstores/vectara.ts
```diff
@@ -1,3 +1,8 @@
+import fs from "fs";
+import path from "path";
```
jacoblee93, line 2: This should be built into Node no?

3eif: Yes. I just removed the path lib from the package.json

3eif:
> Can you see if you can dig in a bit and figure out why? Would prefer to not add a dep if we can help it.

I spent some time trying different things and also tried creating an index.js file from scratch with only the API fetch to test this and got the same error.

Here is the index.js file. The code for the default fetch is almost identical to the node-fetch request. 

```js

const data = new FormData();
const f = fs.createReadStream(path.join(process.cwd(), 'test.json'));
data.append('file', f, 'test.json');
console.log(f);

console.log(data.getHeaders());

fetch('https://api.vectara.io/v1/upload?c=405606997&o=5&d=true', {
  method: 'POST',
  body: data,
  headers: {
    'x-api-key': '',
    ...data.getHeaders(),
  },
})
.then(response => {
  if (!response.ok) {
    throw new Error(`HTTP error! status: ${response.status}`);
  }
  console.log(response);
})
.catch(console.error);
```

If I just add the line `import fetch from 'node-fetch';` at the top of the file, it works.

3eif:
@jacoblee93 any update on this PR?

jacoblee93:
Would REALLY prefer to not have node-fetch as even a peer dep - will dig in when I have time.

3eif:
> Would REALLY prefer to not have node-fetch as even a peer dep - will dig in when I have time.

I've switched out node-fetch with https which doesn't require an additional dep. I hope this is okay dependency wise.

jacoblee93:
I think my only other thought here is that this means Vectara will no longer be importable in environments outside of Node (e.g. Edge functions) - @ofermend are we ok with that?

3eif:
@jacoblee93 Removed the use of https and replaced it with the default fetch API like you requested. Could you take another look? Thanks

jacoblee93:
Yes looks great! Will merge in #2356 after resolving conflicts

jacoblee93:
Merged in #2356! Thank you for your patience.

PR #2185, opened 2023-08-07
Feature - add delete function for Milvus DB

GalO005:
Delete data from Milvus DB using a filter. 
Milvus allows us to delete data based on the primary key only.


langchain/src/vectorstores/tests/milvus.int.test.ts
```diff
@@ -79,6 +79,25 @@ test.skip("Test Milvus.fromExistingCollection", async () => {
   expect(resultThreeMetadatas[0].id).toEqual(1);
```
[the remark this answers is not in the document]
GalO005: Done.

langchain/src/vectorstores/milvus.ts
```diff
@@ -236,7 +236,11 @@ export class Milvus extends VectorStore {
     const results: [Document, number][] = [];
     searchResp.results.forEach((result) => {
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      const fields = { pageContent: "", metadata: {} as Record<string, any> };
+      const fields = {
+        pageContent: "",
+        metadata: {} as Record<string, any>,
+        [this.primaryField]: 0,
```
jacoblee93, line 242: Why add this?

jacoblee93:
Looks good! One comment, and also it sounds like Milvus just allows a filter for deletes?

langchain/src/vectorstores/milvus.ts
GalO005: @jacoblee93 I added it because we can delete rows from Milvus by primary field, and thus I added this field is returned from query. 

[https://milvus.io/docs/delete_data.md](https://milvus.io/docs/delete_data.md)

GalO005:
> Looks good! One comment, and also it sounds like Milvus just allows a filter for deletes?

Yeah, exactly. (collection_name and expr fields) 

[https://milvus.io/docs/delete_data.md](https://milvus.io/docs/delete_data.md)

langchain/src/vectorstores/milvus.ts
jacoblee93: Ah ok, I think the `0` was a bit confusing. Can we change it to `null` or `undefined` to be a bit more standard?

jacoblee93: Wait sorry I don't see how `fields[this.primaryKey]` is getting set?

jacoblee93:
> > Looks good! One comment, and also it sounds like Milvus just allows a filter for deletes?
> 
> Yeah, exactly. (collection_name and expr fields)
> 
> https://milvus.io/docs/delete_data.md

But you say in the comments above that:

Milvus allows us to delete data based on the primary key only.

GalO005:
> > > Looks good! One comment, and also it sounds like Milvus just allows a filter for deletes?
> 
> > 
> 
> > Yeah, exactly. (collection_name and expr fields)
> 
> > 
> 
> > https://milvus.io/docs/delete_data.md
> 
> 
> 
> But you say in the comments above that:
> 
> 
> 
> Milvus allows us to delete data based on the primary key only.

Correct, I mean that primary key must to be in expr field. 
We can't delete by other fields. 
I didn't mention the collection name in the previous comment sorry.

langchain/src/vectorstores/milvus.ts
GalO005: @jacoblee93  It generated automatically (we don't define it when we create a document), but I attached the primary key to metadata (after the document was created). 

I think maybe to create a some function that handling with regular query (without similarity) due to extract the primary key, something like that: 

[https://milvus.io/docs/query.md](https://milvus.io/docs/query.md)

What do you think?

GalO005: @jacoblee93 I fixed it.

jacoblee93:
Thank you!

PR #2203, opened 2023-08-08
Milvus text field max length

weberwang:
Adding an optional parameter for setting the `max_length` of the Text field.
https://github.com/hwchase17/langchainjs/issues/2186#issue-1838619019


jacoblee93:
Thank you!

PR #2205, opened 2023-08-08
Returning score info in result Documents for SimilaritySearch and MDB ATLAS MMR Search

sdellama:
_getRelevantDocuments: Since SimilaritySearch is using by default SearchWithScore under the hood, worths including in returned documents metadata the score info. (applied same logic to MDB Atlas MMR Search)

Fixes # (2191)

### DIFF
--- a/langchain/src/vectorstores/mongodb_atlas.ts
+++ b/langchain/src/vectorstores/mongodb_atlas.ts
@@ -171,13 +171,13 @@ export class MongoDBAtlasVectorSearch extends VectorStore {
     );
 
     return mmrIndexes.map((idx) => {
-      const doc = resultDocs[idx][0];
+      const [doc, score] = resultDocs[idx];
 
       // remove embeddings if they were not requested originally
       if (!includeEmbeddingsFlag) {
         delete doc.metadata[this.embeddingKey];
       }
-      return doc;
+      return { ...doc, score };
     });
   }
 
--- a/langchain/src/vectorstores/mongodb_atlas.ts
+++ b/langchain/src/vectorstores/mongodb_atlas.ts
@@ -179,9 +179,8 @@ export class MongoDBAtlasVectorSearch extends VectorStore {
       }
       return {
         pageContent: doc?.pageContent,
-        metadata: { ...doc?.metadata, score },
+        metadata: { score, ...doc?.metadata },
       };
-      // return { ...doc, score };
     });
   }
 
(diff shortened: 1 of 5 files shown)


langchain/src/vectorstores/mongodb_atlas.ts
```diff
@@ -171,13 +171,17 @@ export class MongoDBAtlasVectorSearch extends VectorStore {
...
+      const [doc, score] = resultDocs[idx];
 
       // remove embeddings if they were not requested originally
       if (!includeEmbeddingsFlag) {
         delete doc.metadata[this.embeddingKey];
       }
-      return doc;
+      return {
+        pageContent: doc?.pageContent,
+        metadata: { ...doc?.metadata, score },
```
jacoblee93, line 182: Can you reverse this so that if there's an explicit metadata field called `score` it won't be overridden?

`metadata: { score, ...doc?.metadata },`

I think it also might be better if there were a property like `returnScores` and the metadata were populated in `similaritySearchVectorWithScore` above.

langchain/src/vectorstores/mongodb_atlas.ts
```diff
@@ -171,13 +171,17 @@ export class MongoDBAtlasVectorSearch extends VectorStore {
...
+      };
+      // return { ...doc, score };
```
jacoblee93, line 184: Remove comment

jacoblee93:
Thank you! Some small questions

langchain/src/vectorstores/mongodb_atlas.ts
sdellama: ok for the reverse.

`similaritySearchVectorWithScore` is already returning scores but with type `[Document<Record<string, any>>, number][]` . So for me would be redundant to add another flag for it. Is responsability of MMR method (since is using `similaritySearchVectorWithScore` method) to shape is own output.

sdellama:
ehi @jacoblee93 any news your side? would be great to have this within tomorrow, since on monday i have to show some results with langchain in my project ;) thanks!

jacoblee93:
Hey @sdellama - have had a bit more time to think about this and I think it could cause unintended consequences. Documents can be reinserted/retrieved, so we probably don't want to permanently populate metadata on returned docs by default on the base class.

I assume you are using this in a vectorstore retriever? If so, an easy way to do this is to subclass your own retriever - it's a pretty lightweight class:

```typescript
class VectorStoreRetrieverWithScore extends BaseRetriever {
  constructor(protected vectorstore: VectorStore) {}
  _getRelevantDocuments(query: string) {
    const docsWithScores = this.vectorstore.similaritySearchWithScore(query);
    const docs = docsWithScores.map(([doc, score]) => {
      doc.metadata = { ...doc.metadata, score };
      return doc;
    });
  }
}
```

PR #2209, opened 2023-08-08
Add consistent timeout for Vectara API calls

3eif:
Adds a timeout of 60 seconds for all calls to the Vectara API

Similar to https://github.com/langchain-ai/langchain/pull/8892


langchain/src/vectorstores/vectara.ts
```diff
@@ -58,6 +58,8 @@ export class VectaraStore extends VectorStore {
 
   private verbose: boolean;
 
+  private vectaraApiTimeout = 60;
```
jacoblee93, line 61: Nit: `vectaraApiTimeoutSeconds`?

langchain/src/vectorstores/vectara.ts
```diff
@@ -202,12 +212,12 @@ export class VectaraStore extends VectorStore {
         metadata: Record<string, unknown>;
         score: number;
       }) => [
-        new Document({
-          pageContent: response.text,
-          metadata: response.metadata,
-        }),
-        response.score,
-      ]
+          new Document({
```
jacoblee93, line 215: I think this will fail formatting - can you run `yarn format`?

jacoblee93:
Thank you! Two comments

3eif:
Thanks for the comments. I've replaced the variable name and ran the formatter.

jacoblee93:
Thank you!

Issue #2208, opened 2023-08-08
SqlToolkit ignoring database-specific prompt from SQL_Utils


willemi069808:
I was looking into enabling a new database type (SAP HANA) for the SqlDatabaseChain, similar to the changes made in PR #1169 for MSSQL.

I noticed that the database-specific coding is mainly done in the SQL Utils module ([/langchain/src/util/sql_utils.ts](https://github.com/hwchase17/langchainjs/blob/main/langchain/src/util/sql_utils.ts)), which is then used/referenced in the SqlDatabaseChain module ([langchain/src/chains/sql_db/sql_db_chain.ts](https://github.com/hwchase17/langchainjs/blob/main/langchain/src/chains/sql_db/sql_db_chain.ts)).

However, the SqlToolkit agent ([langchain/src/agents/toolkits/sql/sql.ts](https://github.com/hwchase17/langchainjs/blob/main/langchain/src/agents/toolkits/sql/sql.ts)) is only partially using/referencing these utils, and _partially hardcoding_ the 'sqlite' dialect. For example: retrieving the prompt from its own prompt file ([langchain/src/agents/toolkits/sql/prompt.ts](https://github.com/hwchase17/langchainjs/blob/main/langchain/src/agents/toolkits/sql/prompt.ts)) instead of the [getPromptTemplateFromDataSource](https://github.com/hwchase17/langchainjs/blob/main/langchain/src/util/sql_utils.ts#L280) function from Utils as is done in the SqlDatabaseChain.

Is there a specific reason that the agent is not using the prompt-generation of the utils? The prompt for SAP HANA is slightly different than SQLite hence it would be beneficial if the database-specific coding is used there as well.

jacoblee93:
It's probably an oversight - will look into it!

PR #2218, opened 2023-08-09
add sap hana support for sql chains and agents

willemi069808:
Added support for SAP HANA databases for SqlDatabaseChain and SqlToolkit.
Main elements:
- added 'sap' as supported database (= typeorm database type name)
- correct SQL generation to use schema name
- added instruction to only query using schema name in prompts
- added 2 examples

This requires `@sap/hana-client` and `hdb-pool` dependencies which I added to the `examples` package.json but not the main package.json.

Database-specific coding seems to have been there for SqlDatabaseChain but not for SqlToolkit. See also #2208


examples/package.json
```diff
@@ -41,6 +42,7 @@
     "chromadb": "^1.4.0",
     "faiss-node": "^0.2.1",
     "graphql": "^16.6.0",
+    "hdb-pool": "^0.1.6",
```
jacoblee93, line 45: Where are these required?

langchain/src/agents/toolkits/sql/sql.ts
```diff
@@ -44,7 +45,7 @@ export function createSqlAgent(
   args?: SqlCreatePromptArgs
 ) {
   const {
-    prefix = SQL_PREFIX,
+    prefix = toolkit.db.appDataSourceOptions?.type === 'sap' ? SQL_PREFIX_SAPHANA : SQL_PREFIX,
```
jacoblee93, line 48: Let's do this in a separate PR and fix it for all the different SQL dialects?

langchain/src/chains/sql_db/sql_db_prompt.ts
```diff
@@ -102,3 +102,24 @@ Only use the following tables:
 Question: {input}`,
   inputVariables: ["dialect", "table_info", "input", "top_k"],
 });
+
+export const SQL_SAPHANA_PROMPT = /*#__PURE__*/ new PromptTemplate({
```
jacoblee93, line 106: Prefer: `SQL_SAP_HANA_PROMPT`

jacoblee93:
Thanks for this! Looks good overall - one question!

examples/package.json
willemi069808: Good catch. Yes, hdb-pool is implicitly required by `typeorm` even though the package has been deprecated for a while ([hdb-pool github](https://github.com/ckyycc/hdb-pool)) since the features are now supported by NodeJs natively.

`typeorm` allows you to specify your own driver in the connection settings (https://github.com/typeorm/typeorm/blob/master/src/driver/sap/SapDriver.ts#L818), but doing so breaks the code as it is written for hdb-pool only, e.g. expecting an `eventEmitter` property in the driver: (https://github.com/typeorm/typeorm/blob/master/src/driver/sap/SapDriver.ts#L287).

So the `hdb-pool` dependency is needed until `typeorm` updates their code.

langchain/src/agents/toolkits/sql/sql.ts
willemi069808: ok - will remove this

langchain/src/chains/sql_db/sql_db_prompt.ts
willemi069808: ok - will update

willemi069808:
Please see changes in new commit. Thanks!

examples/package.json
jacoblee93: Got it - since it's not explicitly in the code, let's not have this in the package.json here and add it to the docs instead.

jacoblee93:
Looks good, one more comment!

willemi069808:
Ok, dependencies removed.

jacoblee93:
Just lint + format + expose an example - thank you!

If you have time to get some of those other agent changes in, we'd welcome it as well :)

PR #2222, opened 2023-08-09
Add optional client options to Anthropic sdk to allow overridable default headers

colegottdank:
Fixes # (issue)

### DIFF
--- a/langchain/src/chat_models/anthropic.ts
+++ b/langchain/src/chat_models/anthropic.ts
@@ -1,4 +1,9 @@
-import { Anthropic, AI_PROMPT, HUMAN_PROMPT } from "@anthropic-ai/sdk";
+import {
+  Anthropic,
+  AI_PROMPT,
+  HUMAN_PROMPT,
+  ClientOptions,
+} from "@anthropic-ai/sdk";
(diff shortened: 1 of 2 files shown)


langchain/src/chat_models/anthropic.ts
```diff
@@ -159,7 +164,14 @@ export class ChatAnthropic extends BaseChatModel implements AnthropicInput {
   // Used for streaming requests
   private streamingClient: Anthropic;
 
-  constructor(fields?: Partial<AnthropicInput> & BaseChatModelParams) {
+  private clientOptions: ClientOptions;
+
+  constructor(
+    fields?: Partial<AnthropicInput> &
+      BaseChatModelParams & {
+        clientOptions?: ClientOptions;
```
jacoblee93, line 172: Could we instead add `clientOptions?` to `AnthropicInput` directly?

jacoblee93:
Seems good, just one question!

langchain/src/chat_models/anthropic.ts
colegottdank: Done & added test

colegottdank:
> Seems good, just one question!

jacoblee93:
Thank you!

PR #2224, opened 2023-08-09
feat: Parallelize the GithubRepoLoader

### LINKED ISSUE #2196: GithubRepoLoader: Fetch directories and files in parallel
In the [GithubRepoLoader](https://github.com/hwchase17/langchainjs/blob/main/langchain/src/document_loaders/web/github.ts), the method [processDirectory](https://github.com/hwchase17/langchainjs/blob/main/langchain/src/document_loaders/web/github.ts#L144) is looping through the files / directories in series rather than in parallel. This makes the loader extremely slow. If `this.recursive` is enabled, the problem compounds exponentially as the loop unwraps the contents of each child directory. 

We've modified the code to run in parallel for our use at Postman and have seen _dramatic_ gains (~90 seconds to load a mid-sized repo down to < 1 second). We can submit a PR with our changes.

See also: https://github.com/hwchase17/langchainjs/issues/2197

yroc92:
Instead of synchronously fetching file content and directory info, this change leverages promises and `Promise.all` to parallelize the file retrieval when `recursive = true` for `GithubRepoLoader`.

Fixes #2196 

Twitter: https://twitter.com/yroc92


langchain/src/document_loaders/web/github.ts
```diff
@@ -141,34 +147,82 @@ export class GithubRepoLoader
...
+            );
           }
+        } else if (this.recursive) {
+          currDirDirPromises.push(this.processDirectory(file.path));
         }
       }
+    }
+
+    const curDirDirectories: Promise<GetContentResponse>[][] =
+      await Promise.all(currDirDirPromises);
```
jacoblee93, line 194: Can we set some kind of max concurrency here?

Something like:

https://github.com/hwchase17/langchainjs/blob/main/langchain/src/schema/runnable.ts#L60

langchain/src/document_loaders/web/github.ts
```diff
@@ -141,34 +147,82 @@ export class GithubRepoLoader
...
-      const files = await this.fetchRepoFiles(path);
+  /**
+   * Takes the file info and wrap it in a promise that will resolve to the file content and metadata
+   * @param file
+   * @returns
+   */
+  private async fetchFileContentWrapper(
+    file: GithubFile
+  ): Promise<GetContentResponse> {
+    const fileContent = await this.fetchFileContent(file).catch((error) => {
```
jacoblee93, line 158: Just throw the error if this fails? I think returning empty here is not ideal

langchain/src/document_loaders/web/github.ts
```diff
@@ -141,34 +147,82 @@ export class GithubRepoLoader
...
+  /**
+   * Begins the process of fetching the contents of the repository
+   */
+  private async processRepo(): Promise<GetContentResponse[]> {
+    try {
+      // Get the list of file / directory names in the root directory
+      const files = await this.fetchRepoFiles(this.initialPath);
+      // Map the file / directory paths to promises that will fetch the file / directory contents
+      const currDirPromises = await this.getCurrDirFilesPromises(files);
+      return Promise.all(currDirPromises);
```
jacoblee93, line 208: Or I guess actually max concurrency makes the most sense here right?

jacoblee93:
Looks nice! Just a bit worried about rate limits/the change in behavior - see comments

jacoblee93:
Hey @yroc92, added max concurrency support with an internal utility we have. Thank you, will get this merged tonight!

jacoblee93:
Thanks again!

yroc92:
@jacoblee93 good to know about the concurrency support!

It’s definitely easy to hit rate limits if you don’t filter out enough files with gitignore. But parallelizing the ops won’t impact the rate limit much because the limits are calls per hour. Using an API key makes the limits much higher.

jacoblee93:
Ah gotcha. Well, I defaulted to 2 but you can always set it to `Infinity` to get the other behavior

PR #2230, opened 2023-08-10
Add context sentence config for Vectara search

3eif:
This PR adds the option to specify the number of sentences before/after the matching segment to add to the context in VectaraStore.


jacoblee93:
Looks good thanks!

Would be lovely someday to get all these types exported from an official SDK...

PR #2231, opened 2023-08-10
Add support for weaviate multi tenancy

mbennett-talentnet:
This PR adds the option to specify a tenant in the weaviate client configuration if the weaviate class uses the new multi-tenancy option.


langchain/src/vectorstores/weaviate.ts
```diff
@@ -126,6 +130,7 @@ export class WeaviateStore extends VectorStore {
 
       const flattenedMetadata = flattenObjectForWeaviate(document.metadata);
       return {
+        ...this.tenant ? { tenant: this.tenant } : {},
```
jacoblee93, line 133: `...this.tenant` looks like a typo?

jacoblee93:
Thanks for this! One question

langchain/src/vectorstores/weaviate.ts
jacoblee93: I think you mean 
```typescript
...(this.tenant ? { tenant: this.tenant } : {}),
```

mbennett-talentnet: Not a typo but I do agree it's more readable with the parenthesis. I've updated as per your suggestion.

PR #2233, opened 2023-08-10
docs: (Mendable) Fixes it getting stuck when tabbing out 

nickscamara:
This fixes Mendable not completing when tabbing out and fixes the duplicate message issue as well.


jacoblee93:
Thank you!

PR #2238, opened 2023-08-10
feat/ZepVectorStore + zepjs upgrade and mods

danielchalef:
- ZepVectorStore
- ZepVectorStore tests
- ZepVectorStore examples
- Upgrade zep-js.
- Modify ZepMemory and ZepRetriever for updated zep-js package.
- Update / Improve ZepMemory and ZepRetriever examples.
- Added ZepVectorStore entrypoint


danielchalef:
Failed on environment tests. `yarn build` succeeds locally. Have a failed to check in an artifact?
```

 Tasks:    9 successful, 9 total
Cached:    1 cached, 9 total
  Time:    46.044s
 ```

danielchalef:
Additionally, how do I add a doc page for Zep's VectorStore?

jacoblee93:
> Two additional questions:
> 
> * how do I add a doc page for Zep's VectorStore?
> * Zep can embed server-side, which is useful for folks deploying to environments other than Node. It looks like instantiating a VectorStore requires an Embedding to be passed in. Is there an approach you'd suggest we take here?

1. You can create a new .mdx file under `docs/extras/modules/data_connection/vectorstores/integrations/zep.mdx` and import examples from `examples/`
2. I would look at this: 

https://github.com/hwchase17/langchainjs/blob/main/langchain/src/vectorstores/vectara.ts#L295
https://js.langchain.com/docs/modules/data_connection/vectorstores/integrations/vectara

Although I would actually recommend keeping the constructor arguments the same for easier migrations down the line.

danielchalef:
@jacoblee93 Thanks for the pointer to FakeEmbeddings. I've left the constructor alone, but Zep will now server-side  auto-embed docs if FakeEmbeddings is passed in. Examples now reflect this and have added docs. Thanks!

jacoblee93:
Looks good, thank you!

Issue #2240, opened 2023-08-11
"Input is not an array" error message when i run similarity search on memory vector store


haouarihk:
i'm using memory vector store,

i embed every document like usual, and then when i try to use the similarity search,
Sometimes weirdly enough, the pearson algorithm function recives `null` instead of an array on one of the inputs sometimes,

I couldn't isolate the problem because it happend on someone else's data/hardware but there're 2 possibilities:
1- recieving null from the embedding api due to invalid characters.
2- grossely large embeddings array exceeding the hardware capabilities.

PR #2241, opened 2023-08-11
Fix for input is not an array

haouarihk:
issue #2240

Fixed it by checking if any of the inputs are not array, then set the similarity to 0 (which means not related at all, i assume)

### DIFF
--- a/langchain/src/vectorstores/memory.ts
+++ b/langchain/src/vectorstores/memory.ts
@@ -71,7 +71,7 @@ export class MemoryVectorStore extends VectorStore {
     const filteredMemoryVectors = this.memoryVectors.filter(filterFunction);
     const searches = filteredMemoryVectors
       .map((vector, index) => ({
-        similarity: this.similarity(query, vector.embedding),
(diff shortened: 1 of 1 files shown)


haouarihk:
this is just a temporary fix,

it could hide the actual bug,
so i'm not 100% suggesting to merge this pr,
and best to try and tacle the real issue

haouarihk:
one solution is to perform sanatization before sending request to openai,
from any weird characters that are not accepted by openai api (if the issue was openai returning null)

haouarihk:
> This seems reasonable, though I wonder if it's an issue elsewhere... Do you have any repro code?

i don't, i will reference my friend who had the issue, to this issue

haouarihk:
he said this file caused it
not sure
[Untitled 42.md](https://github.com/hwchase17/langchainjs/files/12319227/Untitled.42.md)

in another thought, 
there's another potential bug, which is when the document chunk is too large for openai to process,
does it return null in the embeddings?

jacoblee93:
I think this would be better to catch at the embeddings level - will close this but would be up for another PR there

PR #2243, opened 2023-08-11
fix(qa-doc): rename vectorstore to vectorStore

BParvatkar:
- Keep variable name consistent


jacoblee93:
Agh good catch thanks!

PR #2250, opened 2023-08-12
Feat: add mergeFrom and fromIndex in FaissStore

neebdev:
https://github.com/hwchase17/langchainjs/issues/2136. Added the functionality to merge indexes and initialize from another index in FaissStore.


jacoblee93:
Thank you!

langchain/src/vectorstores/faiss.ts
```diff
@@ -254,6 +289,21 @@ export class FaissStore extends SaveableVectorStore {
     return instance;
   }
 
+  static async fromIndex(
```
jacoblee93, line 292: What's the purpose of this method? Wouldn't we just call `.load()`?

jacoblee93:
Looks good! Small question but will merge tonight

langchain/src/vectorstores/faiss.ts
neebdev: Its not absolutely necessary but I think it provides a cleaner way to clone a previous index without saving on disk.

neebdev: One more point, It doesn't make sense to require initializing `fromDocuments` or `fromTexts` before they can merge. So `fromIndex` provides a cleaner way of doing: `const vectorStore = new FaissStore(new FakeEmbeddings(), {}); vectorStore.mergeFrom(...);`. Let me know what you think.

jacoblee93:
Thank you!

PR #2255, opened 2023-08-13
Fixes: Self-query fixes

ppramesi:
1. **Fixed Empty/Null Filter Return**:
   - If self-query retriever's LLM returns empty/null filter, self-query retriever now returns an empty array instead of returning everything.
   - Changed the functional query translator to return empty object instead of a function that always returns false so the returning empty array can be handled from inside _getRelevantDocuments

2. **Fixed Default Filter for Self-Query Retriever**:
   - Filter param inside searchParams now works correctly.
   - User can merge default filter with the generated filter using "and", "or", or "replace" operator.
   - "and": Merges as filter A && filter B.
   - "or": Merges as filter A || filter B.
   - "replace": Replaces the default filter with the generated filter.
   - merging filters defaults to "and"
   - Default filter is optional, and its type should match the vectorstore's filter type.

Example:
```
const selfQueryRetriever = await SelfQueryRetriever.fromLLM({
  llm,
  vectorStore,
  documentContents,
  attributeInfo,
  structuredQueryTranslator: new FunctionalTranslator(),
  searchParams: {
    filter: (doc: Document) => doc.metadata && doc.metadata.type === "movie",
    mergeFiltersOperator: "and",
  },
});
```

4. **Added Tests for Filter Merging**:
   - Includes tests for empty generated filter and merging filters.
   - Made Supabase self-query retriever integration test idempotent.

5. **Self-query prompt fixes**
   - Updated the prompt to match Langchain Python
   - Other small fixes

6. **Stricter types for query translator and default filter**
   - searchParams filter type is automatically inferred
   - structuredQueryTranslator type is made to need to be consistent with the vectorstore type

This PR addresses specific issues with filter handling and testing in the self-query retriever.


langchain/src/chains/query_constructor/ir.ts
```diff
@@ -44,18 +43,16 @@ export type VisitorComparisonResult = {
...
 };
 
-export abstract class Visitor {
+export abstract class Visitor<T extends VectorStore = VectorStore> {
   declare VisitOperationOutput: object;
 
   declare VisitComparisonOutput: object;
 
-  declare VisitStructuredQueryOutput: { filter?: object };
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
jacoblee93, line 54: I don't think you need this?

langchain/src/retrievers/self_query/utils.ts
```diff
@@ -0,0 +1,23 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+export function isObject(obj: any): obj is object {
+  return (
+    obj &&
+    typeof obj === "object" &&
+    !Array.isArray(obj) &&
+    obj.constructor.name === "Object"
```
jacoblee93, line 7: I think this will get messed up in minified environments - will the first three checks suffice?

jacoblee93:
Looks excellent - one nit I can fix myself. Will run integration tests and merge tonight if all pass!

Thank you very much!

langchain/src/chains/query_constructor/ir.ts
ppramesi: Oh yeah I forgot to remove it.

langchain/src/retrievers/self_query/utils.ts
ppramesi: Yeah I think it should be fine.

PR #2261, opened 2023-08-13
Fix openai function calling parameter format parsing error

Tomotoes:
additional_kwargs.function_call.arguments will sometimes be in a non-JSON format and should be compatible with the string format.

I added a test case (code-interpreter), when the parameter returned by OpenAI function calling is in non-JSON format, the following problems will occur:
image
image

This PR fixes this problem and supports parameter parsing in non-JSON format, which can be tested by `yarn test:single ./src/agents/tests/openai.int.test.ts -t "OpenAIAgent code interpreter"`.
https://app.warp.dev/block/uopAv2WHtht4SB5lRT2xOI

### DIFF
--- a/langchain/src/agents/openai/index.ts
+++ b/langchain/src/agents/openai/index.ts
@@ -35,23 +35,30 @@ function parseOutput(message: BaseMessage): FunctionsAgentAction | AgentFinish {
     // eslint-disable-next-line prefer-destructuring
     const function_call: ChatCompletionRequestMessageFunctionCall =
       message.additional_kwargs.function_call;
+
+    let toolInput: string;
     try {
-      const toolInput = function_call.arguments
+      toolInput = function_call.arguments
         ? JSON.parse(function_call.arguments)
         : {};
-      return {
-        tool: function_call.name as string,
-        toolInput,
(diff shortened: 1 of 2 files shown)


jacoblee93:
Oh fascinating, I didn't know that. Will dig in!

jacoblee93:
Looks like it's a bug on OpenAI as it doesn't occur on gpt-4:

Python tool:
https://smith.langchain.com/public/88423f6d-11b6-4fbe-9946-e5752a853a35/r

JS tool:
https://smith.langchain.com/public/697be3a5-ff74-46d9-9aa6-8feb3c9266bd/r

I don't think it's necessarily the best fix to pass args directly through - could you try using GPT-4?

Tomotoes:
https:app.warp.dev/block/ooIj0isjJ7mwXcf7NJOnf4
I tried GPT4 and still get the problem... Maybe it's because of my prompt?

jacoblee93:
I don't think we should update the parser just for this since OpenAI is actually returning something in the incorrect format... Going to close this for now.

This person said few-shotting the prompt seemed to help:

https://twitter.com/bennetkrause/status/1691361566912880641?s=20

PR #2267, opened 2023-08-13
Local embeddings function with Transformer.js API

l4b4r4b4b4:
Fixes # (issue)


jacoblee93:
I think you have to commit the updated Lockfile

Also, I think we've tried that specific package and it has some issues in most JS envs

l4b4r4b4b4:
Will try that. As for the node env. I can imagine it does not run on edge environment. 
In fully fledged node environment in the context of a Next.js 13 app I have no promblem using this function to store document embeddings in ChromaDB. So would be great to have the capability to retrieve the embedded collections from LangChain ;)

jacoblee93:
Hey @l4b4r4b4b4, wasn't able to push commits onto your branch but opened #2288 with an integration test - it seems to just time out on my environment (Node 18.15.0).

Does it require some other setup to run? That's ok if so, but it should be documented.

l4b4r4b4b4:
as I wrote in the documentation mdx you need to install `@xenova/transformers` which I added to the peerDependencies. Will have a look at the test.

jacoblee93:
Yes, I have it installed but the test didn't run on node 18 - I remember having some difficulty in the past so was wondering if your environment was different?

l4b4r4b4b4:
grafik
That's the node version I am running on. 
When looking at the CI/CD test that failed with your PR #2288 I had formatting test fail tool, when not leaving empty lines between certain lines e.g. defining class variables.

jacoblee93:
Merged in #2487

PR #2270, opened 2023-08-14
Optional 'user' input param for OpenAI API

ilyasubkhankulov:
Added support for optional 'user' input parameter for OpenAI API.

> A unique identifier representing your end-user, which can help OpenAI to monitor and detect abuse

Relevant documentation links:
- https://platform.openai.com/docs/api-reference/completions/create
- https://platform.openai.com/docs/api-reference/chat/create

### DIFF
--- a/langchain/src/chat_models/openai.ts
+++ b/langchain/src/chat_models/openai.ts
@@ -219,6 +219,8 @@ export class ChatOpenAI
 
   stop?: string[];
 
+  user?: string;
+
   timeout?: number;
 
   streaming = false;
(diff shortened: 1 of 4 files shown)


jacoblee93:
Thanks! We should also add this to the `chat_models/openai.ts` class too though right?

ilyasubkhankulov:
Thank you for catching that Jacob! I've pushed a commit rectifying the oversight.

jacoblee93:
Awesome, thank you!

jacoblee93:
Thank you!

PR #2273, opened 2023-08-14
fix loose metadata when retrieve from Chroma

imsamurai:
metadata was lost if no line-of-code keys present


jacoblee93:
Thank you, good catch!

imsamurai:
also i want to ask how can i get document id from vecstore or retriever? maybe we need to pass id in metadata for each document or better generate and pass it to metadata in client code?

jacoblee93:
Yeah I think at the moment your best bet is to pass it in as metadata.

imsamurai:
@jacoblee93 Pass in metadata from client code or extend langchain implementation and add ids to metadata when return docs?

PR #2276, opened 2023-08-14
Fix addDocuments ids 

6mdx:
i was have issue in my code

The issue is if I want to set an ids in `options.ids` it shows me this Error

Error inserting: ON CONFLICT DO UPDATE command cannot affect row a second time 400 Bad Request

After debugging it turns out that the problem is due to duplicating the id of all chunks.


jacoblee93:
Thanks, good catch!

langchain/src/vectorstores/supabase.ts
```diff
@@ -79,9 +79,9 @@ export class SupabaseVectorStore extends VectorStore {
     const chunkSize = 500;
     let returnedIds: string[] = [];
     for (let i = 0; i < rows.length; i += chunkSize) {
-      const chunk = rows.slice(i, i + chunkSize).map((row) => {
+      const chunk = rows.slice(i, i + chunkSize).map((row, j) => {
         if (options?.ids) {
-          return { id: options.ids[i], ...row };
+          return { id: options.ids[j], ...row };
```
jacoblee93, line 84: Wait, should this be `i + j`?

6mdx: I equal 0 always 

example here : https://codesandbox.io/s/youthful-shirley-843zf9?file=/src/app.ts
refresh the browser to console log i and chunk

jacoblee93:
Try inserting more documents than the chunk size (500)

6mdx:
Yup should be `i + j`
I test it with chunk size 5, and i added 10 documents, It works fine. 👌

example here : https://codesandbox.io/s/youthful-shirley-843zf9?file=/src/app.ts

jacoblee93:
Thank you!

PR #2289, opened 2023-08-15
Check that env var is truthy

hinthornw:
Seems like if you set to ="false" you'd expect tracing to not be performed

### DIFF
--- a/langchain/src/callbacks/manager.ts
+++ b/langchain/src/callbacks/manager.ts
@@ -432,12 +432,6 @@ export class CallbackManagerForToolRun
   }
 }
 
-function isTruthy(value: string | null | undefined): boolean {
-  if (!value) {
-    return false;
-  }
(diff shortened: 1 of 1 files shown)


langchain/src/callbacks/manager.ts
```diff
@@ -432,6 +432,12 @@ export class CallbackManagerForToolRun
   }
 }
 
+function isTruthy(value: string | null | undefined): boolean {
+  if (!value) {
+    return false;
+  }
+  return /^(true|y|yes)$/i.test(value);
```
jacoblee93, line 439: Can we not just directly compare to `=== "true"`? Not sure I love encouraging a bunch of acceptable values here

hinthornw: fair

jacoblee93:
Thank you!

PR #2293, opened 2023-08-15
zep/VectorStore MMR + bugfixes

danielchalef:
- MMR implementation
- Bug fix issue where Zep collection was set to auto-embed and ZepVectorStore still passing in embedding vectors for search.
- Add MMR to examples


jacoblee93:
Thank you!

PR #2308, opened 2023-08-17
update myscale default index type

mpskex:
We update the default index type from `IVFFLAT` to `MSTG`, a new vector type developed by MyScale.


jacoblee93:
Thank you!

PR #2312, opened 2023-08-17
Fix: Supabase self-query "or" filters merge operator

ppramesi:
Added "or" operator for Supabase self-query default filter merging.


langchain/src/retrievers/self_query/utils.ts
```diff
@@ -1,4 +1,9 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
+import type {
+  SupabaseFilter,
+  SupabaseFilterRPCCall,
+} from "../../vectorstores/supabase.js";
```
jacoblee93, line 5: Could we put these new methods in its own `supabase_utils.ts` class?

langchain/src/retrievers/self_query/utils.ts
```diff
@@ -16,3 +21,184 @@ export function isFilterEmpty(
...
+        } else {
+          return Reflect.get(target, prop, receiver);
+        }
+      },
+    };
+
+    return proxyHandler;
+  }
+
+  addToValues(column: string, value: string) {
```
jacoblee93, line 109: Don't think this is being used?

langchain/src/retrievers/self_query/utils.ts
```diff
@@ -16,3 +21,184 @@ export function isFilterEmpty(
...
+      return value.replace("::int", "");
+    }
+    return value;
+  }
+
+  defaultOp(prop: string, column: string, value: unknown) {
+    this.values.push([this.removeType(column), `${String(prop)}.${value}`]);
+  }
+
+  or(filters: string, { foreignTable }: { foreignTable?: string } = {}) {
```
jacoblee93, line 124: Can we rename these to `addOrClause` or something similar?

Just not obvious above that `this.or()` above has side-effects

langchain/src/retrievers/self_query/utils.ts
```diff
@@ -16,3 +21,184 @@ export function isFilterEmpty(
...
+    "filter",
+  ];
+
+  values: [string, string][] = [];
+
+  buildProxyHandler() {
+    const proxyHandler: ProxyHandler<SupabaseFilter> = {
+      get: (target, prop, receiver) => {
+        if (typeof target[prop as SupabaseFilterProps] === "function") {
+          return (...args: any[]) => {
```
jacoblee93, line 52: Just to cut down on the number of casts - can we type this as a `[string, string | unknown]` tuple or something similar

langchain/src/retrievers/self_query/utils.ts
```diff
@@ -16,3 +21,184 @@ export function isFilterEmpty(
...
+    "match",
+    "not",
+    "textSearch",
+    "filter",
+  ];
+
+  values: [string, string][] = [];
+
+  buildProxyHandler() {
+    const proxyHandler: ProxyHandler<SupabaseFilter> = {
```
jacoblee93, line 49: The main reason we need this is because Supabase filters can be an object or a function right? And we need different behavior based on that?

jacoblee93:
Looks good overall! Just a few questions.

Also `any` doesn't throw an error on lint but can you add a `// eslint-disable-next-line @typescript-eslint/no-explicit-any` above where you use it to suppress the warning?

langchain/src/retrievers/self_query/utils.ts
ppramesi: Well, essentially, supabase object filter is not flexible enough to be a good self-query filter, so instead we're using supabase's js postgrest client (https://github.com/supabase/postgrest-js) to build the merged filter query. The package works really well if you're trying to chain together "and" statements, but chaining together "or" statements is not so simple, since the statements need to be converted to strings that postgrest can use (see https://postgrest.org/en/stable/references/api/tables_views.html). This pull request is essentially monkeypatching the filter builder to stringify the query params that would've been sent to postgrest so we can chain them as "or" statements.

As of now, you can't use object as the default filter, but I can add that in this pull request.

ppramesi:
> Looks good overall! Just a few questions.
> 
> Also `any` doesn't throw an error on lint but can you add a `// eslint-disable-next-line @typescript-eslint/no-explicit-any` above where you use it to suppress the warning?

Will do!

langchain/src/retrievers/self_query/utils.ts
ppramesi: Sure!

langchain/src/retrievers/self_query/utils.ts
ppramesi: Ok I will move this to supabase_utils. Honestly this is very specific to supabase, so much so I think it's not worth the effort to generalize it.

langchain/src/retrievers/self_query/utils.ts
ppramesi: Ah you're correct. Let me remove it.

langchain/src/retrievers/self_query/utils.ts
ppramesi: Hmm let me rename the rest too to make it consistent.

jacoblee93:
Looks good to me, thank you!

I'll write up some docs for this in another PR.

halyearzero:
hi @ppramesi and @jacoblee93 

I've noticed that Supabase integration with langchainjs is pulling way ahead of the python version. A lot of the advanced retrieval methods just aren't available there and I'm considering abandoning Supabase because of it. Do you know of any plans to add these tools to the python version?

PR #2315, opened 2023-08-17
Implement YouTube transcripts loader

tyumentsev4:
Introducing the ability to load a transcription of an YouTube video using YouTube document loader.


jacoblee93:
Thanks! A few small fixes but looks good.

jacoblee93:
Thank you!

PR #2323, opened 2023-08-18
fix: ERNIE-Bot-turbo supports temperature and more

sunner:
Now, `ERNIE-Bot` and `ERNIE-Bot-turbo` share the same API.


jacoblee93:
Sweet, thanks!

PR #2324, opened 2023-08-18
Add support for loading minified classes

nfcampos:
- Missing: Actually setting lc_name in all the serializable classes

Fixes # (issue)


jacoblee93 (approved):

jacoblee93:
Ended up opening #2328 on top of this - will merge that instead.

PR #2325, opened 2023-08-18
feat: allow openai organization to be set from process.env

mattzcarey:
Brings JS up to date with Python version allowing you to specify the organization in environment variables for the openai LLM class. 

I hope I have done this right :)

Twitter: @mattzcarey


jacoblee93:
Thanks very much!

PR #2334, opened 2023-08-19
Automatically extract template variables 

NarHakobyan:
Hello, within this pull request, you have the opportunity to review an ExtractTemplateParamsRecursive type that is capable of extracting template variables from the template.

[check here](https://www.typescriptlang.org/play?#code/C4TwDgpgBAogHsATgQwMbAArJQWwHLI4QDOAShKgK6LECWAbhADwAqUECEAdgCbFTEktLgHMANFHLFKAG2DtOvfoMTCRAbQC6UALxQtAPl1QAUFChsOwbnygADAN4ASB8IBmERFCy4Avr5d3T0kSYF87MygAflgEFHQfQgIiMgpqOkYmKWAJdQA6AqlZHO9sQk0DSIAuCwVrJXsXFTUA1y4PL2zwyPMY+CQ0TDL8QhJyKhoGZmyJIrlK83MaueAAbhMTUEgoAGFjfvih3GSxtMnMuwBNAHtKKGxoZCgwWhRrKC5RnlLgVAALEB5KAAQRkMigiBIYGuXGIJCgOEogigACNoFZIUQZCAoIxECjrnD7rwoMJnq9kO8eLRkDIKMA8htzABVOGIGptMCUMKRYEASSqdgMqygAHpRfphFzgJoNltoIkcMQAGKIa44Vh1GzKISiIx6ByRdQAaQgOLJB0GipOqQmGWYLAM6i4lBwaMQmk0NWaonWvnWm3A0AAIsZFSq1Rqrrd7pD7uS3tBPkRvlhfgCgaDwZDiNDYfDEci0XVMRBsbjPASicgSWSXomoNTafTGZFWZ4OVLub5eQKhSYgA)

Had some trouble combining with a new RunInput. If you have any suggestions on how to fix this, I would love to fix it and open a PR.


jacoblee93:
Whoa, this works?

NarHakobyan:
Almost :D, it conflicts with `RunInput` but without it it works perfect.

NarHakobyan:
It works 🚀

image

jacoblee93:
As incredible as this is, it fails our CI build :(

`Error: langchain:build: src/chains/question_answering/map_reduce_prompts.ts(69,17): error TS2589: Type instantiation is excessively deep and possibly infinite.`

Would love to keep it on hold though if we can find a way to set e.g. max recursion depth!

NarHakobyan:
It happens when TS decides that types become too complex to compute :(

ppramesi:
Maybe try something like

```
type ExtractTemplateParamsRecursive<T extends string, Result extends string[] = []> =
  T extends `{${infer Param}}${infer Rest}`
    ? ExtractTemplateParamsRecursive<Rest, [...Result, Param]>
        : T extends `${string}{${infer Param}}${infer Rest}`
            ? ExtractTemplateParamsRecursive<Rest, [...Result, Param]>
                : Result;
```

It should cut down on the depth of recursion by a lot since it wouldn't have to go over the string character by character and instead go straight to values between curly braces.

Note that this still doesn't completely fix the recursion problem, as it would throw the same error if you have too many (around 1000) template variables. Though, it's very unlikely that anyone would ever use a prompt with that many template variables.

ppramesi:
Also also, might want to do something like

```
type ValidTemplateVariable<T extends string> =
  T extends `{${string}` ?
    never : T extends ` ${string}` 
        ? never : T extends `\n${string}` 
            ? never : T extends `"${string}` 
                ? never : T extends `'${string}` 
                    ? never : T;


type ExtractTemplateParamsRecursive<T extends string, Result extends string[] = []> =
  T extends `{${ValidTemplateVariable<infer Param>}}${infer Rest}`
    ? ExtractTemplateParamsRecursive<Rest, [...Result, Param]>
        : T extends `${string}{${ValidTemplateVariable<infer Param>}}${infer Rest}`
            ? ExtractTemplateParamsRecursive<Rest, [...Result, Param]>
                : Result;
```

Since I know there are a couple of prompts that has something that looks like template variable declaration. e.g.

```
\`\`\`json
{{{{
    "action": string
}}}}
\`\`\`
```

jacoblee93:
Thanks both of you! This is going to make things a lot easier to use - will merge tonight!

jacoblee93:
@NarHakobyan @ppramesi do you think you'd be up for taking a shot at automatic inference for `ChatPromptTemplate.fromPromptMessages`?

https://github.com/hwchase17/langchainjs/blob/main/langchain/src/prompts/chat.ts#L536

ppramesi:
@jacoblee93 I'll look into it!

NarHakobyan:
@ppramesi thanks for the help, this was amazing job 🥳

PR #2340, opened 2023-08-20
Implement Google Vertex AI Matching Engine.

afirstenberg:
Implementation and documentation.
Also refactors GoogleVertexAI connection, associated types, and classes that use it, to make it useful for more than just model-based operations.

Addresses issue #1532


langchain/src/vectorstores/tests/googlevertexai.test.ts
```diff
@@ -0,0 +1,83 @@
+/* eslint-disable no-process-env */
```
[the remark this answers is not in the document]
afirstenberg: The environment variables are all accessed in the tests and examples so we don't have to hardcode projects during testing.

langchain/src/vectorstores/tests/googlevertexai.int.test.ts
```diff
@@ -0,0 +1,163 @@
+/* eslint-disable no-process-env */
```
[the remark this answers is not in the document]
afirstenberg: The environment variables are all accessed in the tests and examples so we don't have to hardcode projects during testing.

langchain/src/vectorstores/googlevertexai.ts
```diff
@@ -0,0 +1,682 @@
...
+  GoogleVertexAIResponse,
+} from "../types/googlevertexai-types.js";
+import { Docstore } from "../schema/index.js";
+
+export interface IdDocumentInput extends DocumentInput<Record<string, any>> {
+  id?: string;
+}
+
+export class IdDocument
+  extends Document<Record<string, any>>
```
jacoblee93, line 23: Can just do `extends Document` to make the linter happy

langchain/src/vectorstores/googlevertexai.ts
```diff
@@ -0,0 +1,682 @@
...
+  neighbors: FindNeighborsResponseNeighbor[];
+}
+
+interface FindNeighborsResponse extends GoogleVertexAIResponse {
+  data: {
+    nearestNeighbors: FindNeighborsResponseNearestNeighbor[];
+  };
+}
+
+class FindNeighborsConnection
```
jacoblee93, line 233: I think this and things above might be better in a utils class? Either the same one as the other connection logic or a new one for matching engine?

langchain/src/vectorstores/googlevertexai.ts
afirstenberg: No objections.

langchain/src/vectorstores/googlevertexai.ts
```diff
@@ -0,0 +1,682 @@
...
+    const ret: [Document, number][] = await Promise.all(
+      neighbors.map(async (neighbor) => {
+        const id = neighbor?.datapoint?.datapointId;
+        const distance = neighbor?.distance;
+        let doc: IdDocument;
+        try {
+          doc = await this.docstore.search(id);
+        } catch (xx) {
+          console.error(xx);
+          doc = new Document({ pageContent: `Missing document ${id}` });
```
jacoblee93, line 595: I think it'd be better to return `undefined` or `null` and filter it out at the end - otherwise it'd cause problems downstream in retrieval.

langchain/src/vectorstores/googlevertexai.ts
```diff
@@ -0,0 +1,682 @@
...
+    let found = false;
+    for (let co = 0; co < deployedIndexes.length && !found; co += 1) {
+      const deployedIndex = deployedIndexes[co];
+      const deployedIndexPath = deployedIndex.index;
+      const match = deployedIndexPath.match(indexPathPattern);
+      if (match) {
+        const [, potentialIndexId] = match;
+        if (potentialIndexId === this.index) {
+          this.deployedIndexId = deployedIndex.id;
+          found = true;
```
jacoblee93, line 634: Probably a bit more standard to do `.find()` instead?

docs/extras/modules/data_connection/vectorstores/integrations/googlevertexai.mdx
```diff
@@ -0,0 +1,176 @@
...
+
+- The id for the Index
+- The id for the Index Endpoint
+
+You will also need a document store. While an `InMemoryDocstore` is ok for
+initial testing, you will likely want to use something like a
+`GoogleCloudStorageDocstore` to store it more permanently.
+
+```typescript
+    import { MatchingEngine } from "langchain/vectorstores/googlevertexai";
```
jacoblee93, line 59: Add under `/examples` and import it instead?

jacoblee93:
Looks great overall! A few refactoring/code comments

langchain/src/types/googlevertexai-types.ts
```diff
@@ -51,7 +54,11 @@ export interface GoogleVertexAIBaseLLMInput
     GoogleVertexAIConnectionParams,
     GoogleVertexAIModelParams {}
 
-export interface GoogleVertexAIBasePrediction {
+export interface GoogleVertexAIResponse {
+  data: any;
```
jacoblee93, line 58: Needs a `// eslint-disable-next-line` rule

langchain/src/vectorstores/googlevertexai.ts
afirstenberg: I'm not sure putting them in a separate utils class does any benefit. All of them are specific to Matching Engine, and  are mostly done this way to address type checking in TypeScript. And each one is specific to the operation.

langchain/src/vectorstores/googlevertexai.ts
```diff
@@ -0,0 +1,682 @@
...
+
+  callerOptions?: AsyncCallerCallOptions;
+
+  apiEndpoint?: string;
+
+  deployedIndexId?: string;
+}
+
+export class MatchingEngine extends VectorStore implements MatchingEngineArgs {
+  declare FilterType: Restriction[];
```
jacoblee93, line 291: nit: We should prefer `this["FilterType"]` instead of using `Restriction[]` below on the off-chance someone wants to subclass

langchain/src/vectorstores/googlevertexai.ts
```diff
@@ -0,0 +1,682 @@
...
+    });
+
+    const metadata: metadataType = {
+      ...flatMetadata,
+      ...stringArrays,
+    };
+    return metadata;
+  }
+
+  metadataToRestrictions(documentMetadata: Record<string, any>): Restriction[] {
```
jacoblee93, line 478: nit: We should prefer `this["FilterType"]` instead of using `Restriction[]` below on the off-chance someone wants to subclass?

langchain/src/vectorstores/googlevertexai.ts
afirstenberg: I go back and forth on that. What is the best way to handle a document that is indexed, but can't be found in the store? I went this way when I was using InMemoryStore and it was getting cleared out between tests, for obvious reasons, but the Matching Engine still had the index.

Are there cases where we want to know (and can benefit from) a document is "missing"?

The biggest downside I see is that "k-nearest neighbor" should return k documents, but filtering them out in this edge cases becomes deceptive.

However... I'm not sure I have a strong opinion to keep it this way. It is certainly an edge cases and not one I'd worry too much about, except that it should log the error.

docs/extras/modules/data_connection/vectorstores/integrations/googlevertexai.mdx
afirstenberg: I'll add the whole thing under /examples, but I think doing it with imports would do one of the following:

- If I have it as a single example, I can't easily break it down into the smaller, explainable, parts.
- If I have multiple examples that get imported, then there is a lot of repeated code (setting up the Matching Engine class) and the relevant example may get lost.

I know best practice is to import the example, but I think that will make the documentation less useful.

Thoughts?

langchain/src/vectorstores/googlevertexai.ts
afirstenberg: I'm trying to consider a case where that might happen, since Restriction is a type that gets passed to the API. But sure... can do that. {:

jacoblee93:
Also run `yarn lint`!

langchain/src/vectorstores/googlevertexai.ts
afirstenberg: I changed my mind - I'm now more firmly of the opinion that they need to stay in, or there needs to be an option that keeps them in, or we need some way to register when they don't exist.

If we remove them, then there is no programatic way through LangChain JS to get "phantom documents" that are indexed, but not in the document store. If there was a _callbacks parameter where we could register this, I might consider it and have some kind of callback to indicate what phantom documents were removed... but this isn't implemented on `similaritySearchVectorWithScore()`, and even if it was, I'm not sure we have a callback event that can handle this.

afirstenberg:
These updates address all the issues that I think are significant barriers, and I think its ready for re-review.

I have *not* made the following changes, for reasons I outlined above:

- It continues to create a phantom doc, if necessary
- The connection classes remain with the Matching Engine class
- Examples aren't imported into the docs, but there is now a comprehensive example

docs/extras/modules/data_connection/vectorstores/integrations/googlevertexai.mdx
jacoblee93: Yeah no worries then - just make sure it runs as is!

We need a more Jupyter notebook style of solution long-term.

langchain/src/vectorstores/googlevertexai.ts
jacoblee93: Ah right. Don't really see a great solution here... too bad the matching engine doesn't store the docs as well.

langchain/src/vectorstores/googlevertexai.ts
jacoblee93: Just to reduce clutter - don't feel too strongly.

jacoblee93:
Hey @afirstenberg, I think it'd be a really bad experience if someone used the `InMemoryDocstore` and got back weird docs - I still feel it'd be better to not return any in that case, but do get your point about wanting to allow some way to delete them (though couldn't the user just go to the console in Google?). 

Either way, I changed the example to use the GCS docstore instead because the user will be using Google Cloud anyway and if they're testing there's a high likelihood that they're just running it as a script, which would continually clear the docstore each time and confuse them.

Can you run through the example again? Can merge after that!

jacoblee93:
FWIW Python ONLY works with GCS and throws an error if the doc isn't found:

https://github.com/langchain-ai/langchain/blob/master/libs/langchain/langchain/vectorstores/matching_engine.py#L229

docs/extras/modules/data_connection/vectorstores/integrations/googlevertexai.mdx
afirstenberg: Yup. `yarn example indexes/vector_stores/googlevertexai` runs with the environment variables set. Tested and confirmed that.

langchain/src/vectorstores/googlevertexai.ts
afirstenberg: There are a lot of things I wish the Matching Engine did differently... but...

afirstenberg:
> Hey @afirstenberg, I think it'd be a really bad experience if someone used the `InMemoryDocstore` and got back weird docs - I still feel it'd be better to not return any in that case, but do get your point about wanting to allow some way to delete them (though couldn't the user just go to the console in Google?).

Nope. The Matching Engine console is virtually useless for most activities. You can't even create a new index through the console (which is why the docs point to a Python notebook to create an index). I plan to use LangChain JS as the foundation for another library that adds useful management to it if they don't do it first.

afirstenberg:
> Can you run through the example again? Can merge after that!

Updated example to match docs. Tested. Looks good!

I think this is good to go. Thank's for your help!

jacoblee93:
Ah awesome, thanks! Will merge today.

davidoort:
Sorry if this is a stupid question @afirstenberg but will this support creating and updating streaming indexes? Vs simply batch indexes

afirstenberg:
> Sorry if this is a stupid question @afirstenberg but will this support creating and updating streaming indexes? Vs simply batch indexes

@davidoort - Nothing stupid at all! Especially given how poor the documentation for Matching Engine is.

It doesn't support _creating_ an index. That's out of scope for LangChain itself, since it isn't something that is really cross-platform. There are, however, guides that help you (the documentation links to one), and I'm hoping to publish a utility that helps with this after Cloud Next if they don't beat me to it. (This is also a long process - creating a new, empty, index can take an hour.)

However it _does_ allow you to stream inserts/updates which are immediately available to be queried.

davidoort:
That makes sense! I guess I'll still need some util scripts to take care of that, but awesome to hear that if the index is a streaming index we can easily update it with langchain, and of course query it. Thanks again for building this!
Am already in contact with someone about adding this to Flowise which uses langchainjs!

afirstenberg:
> That makes sense! I guess I'll still need some util scripts to take care of that, but awesome to hear that if the index is a streaming index we can easily update it with langchain, and of course query it. Thanks again for building this! Am already in contact with someone about adding this to Flowise which uses langchainjs!

@davidoort - Excited to hear more about your use of this when you can. Feel free to reach out to me on Twitter/X or LinkedIn - I'd like to hear how you're using it and pass it along to Google to illustrate the challenges people face with it. Now that this is in a release - we have a lot more leverage!

My hope is that they either announce some improvements at Cloud Next, or that I'll be working on a package to make setting up and managing an index easier.

PR #2342, opened 2023-08-21
Add Vectara document metadata

3eif:
This PR adds the document metadata to the response metadata object.

Metadata before this change (this is the response metadata only):
```
    [
      { name: 'lang', value: 'fra' },
      { name: 'offset', value: '218' },
      { name: 'len', value: '138' }
    ]
```

Metadata after this change (both document and response metadata are added):
```
    [
      { name: 'lang', value: 'fra' },
      { name: 'offset', value: '218' },
      { name: 'len', value: '138' },
      { name: 'document_id', value: '-1848969058' },
      { name: 'title', value: "The hitchhiker's guide to the galaxy" },
      { name: 'author', value: 'Douglas Adams' },
      { name: 'genre', value: 'fiction' }
    ]
```

### DIFF
--- a/langchain/package.json
+++ b/langchain/package.json
@@ -1,6 +1,6 @@
 {
   "name": "langchain",
-  "version": "0.0.129",
+  "version": "0.0.130",
   "description": "Typescript bindings for langchain",
   "type": "module",
   "engines": {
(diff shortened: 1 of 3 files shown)


jacoblee93:
Thank you!

HenryHengZJ:
@3eif hey! just wondering any particular reason why returned metadata is array instead of object?

```
"metadata": {
  "source": "https://en.wikipedia.org/wiki/Elon_Musk",
}
```

3eif:
> @3eif hey! just wondering any particular reason why returned metadata is array instead of object?
> 
> ```
> "metadata": {
>   "source": "https://en.wikipedia.org/wiki/Elon_Musk",
> }
> ```
> 
> In Flowise, because the other metadata is object, so we have a function to convert into an object, would be good to have that from here instead - [FlowiseAI/Flowise#827](https://github.com/FlowiseAI/Flowise/pull/827)

Hi @HenryHengZJ, the metadata returned is an array because that's what the Vectara API returns. I agree that it should be converted into an object from here instead of Flowise. I'll go ahead and create a PR for this and notify you once it's merged. 

Thanks for fixing this on the Flowise side!

PR #2351, opened 2023-08-21
Allow passing k number of documents to retrieve by ParentDocumentRetriever

karol-f:
When using `ParentDocumentRetriever`, searching for more documents in VectorStore is sometimes desired (especially when using LLM model with a bigger context window like GPT-4). This PR allows for adding such param (optional), so it can be used while searching vector store.


jacoblee93:
Fantastic, thank you!

Let's update the docs as well - I can do so later if you don't have time.

karol-f:
> Let's update the docs as well - I can do so later if you don't have time.

I've altered docs a bit - feel free to change it the way you like. Thank you.

jacoblee93:
Renamed to `childK` to be more specific, but looks great! Will merge tonight.

jacoblee93:
Thank you!

PR #2353, opened 2023-08-21
Upgrade OpenAI SDK to v4

nsbradford:
Upgrades OpenAI SDK from v3 to v4 according to the migration guide at https://github.com/openai/openai-node/discussions/217, with assistance from the [Grit automated migration](https://app.grit.io/migrations/new/openai_v4). SDK v4 comes with many new features including streaming support, improved types, and improved latency (release notes at https://github.com/openai/openai-node/discussions/182).


jacoblee93:
This is awesome!

I do want to be careful to not break people where we can (e.g. add a shim mapping `basePath` -> `baseURL`) and support some of the old Axios configs if possible.

Also, definitely need to test on Azure OpenAI deployments - there were some issues with the latest preview versions I believe - but you can leave that one to me once the rest is ready!

nsbradford:
> This is awesome!
> 
> I do want to be careful to not break people where we can (e.g. add a shim mapping `basePath` -> `baseURL`) and support some of the old Axios configs if possible.
> 
> Also, definitely need to test on Azure OpenAI deployments - there were some issues with the latest preview versions I believe - but you can leave that one to me once the rest is ready!

thanks @jacoblee93! Any other specific shims/args you want to keep support for? (I have just rebased on main, so should now be ready for review)

examples/src/guides/expression_language/cookbook_function_call.ts
```diff
@@ -27,7 +27,10 @@ const functionSchema = [
 ];
 
 const chain = prompt.pipe(
-  model.bind({ functions: functionSchema, function_call: { name: "joke" } })
+  model.bind({
+    functions: functionSchema,
+    function_call: { name: "joke", arguments: "" },
```
jacoblee93, line 32: I don't think we need to change this right?

examples/src/models/chat/integration_openai.ts
```diff
@@ -56,6 +56,7 @@ await modelForFunctionCalling.predictMessages(
     // You can set the `function_call` arg to force the model to use a function
     function_call: {
       name: "get_current_weather",
+      arguments: "",
```
jacoblee93, line 59: See above

examples/src/models/chat/openai_functions.ts
```diff
@@ -33,7 +33,7 @@ const model = new ChatOpenAI({
   modelName: "gpt-4",
 }).bind({
   functions: [extractionFunctionSchema],
-  function_call: { name: "extractor" },
+  function_call: { name: "extractor", arguments: "" },
```
jacoblee93, line 36: See above

langchain/src/chat_models/openai.ts
```diff
@@ -420,58 +401,6 @@ export class ChatOpenAI
     }
   }
 
-  startStream(
```
jacoblee93, line 423: So glad to get rid of this!

jacoblee93:
Will give it a deeper dive later but broadly looks awesome - thanks again! I'll aim to get it merged by midweek.

examples/src/guides/expression_language/cookbook_function_call.ts
nsbradford: Thanks, fixed here and elsewhere

nsbradford:
@jacoblee93 addressed first round of review comments

langchain/src/types/openai-types.ts
```diff
@@ -59,11 +58,29 @@ export declare interface OpenAIBaseInput {
...
+  signal?: AbortSignal | undefined | null;
+  idempotencyKey?: string;
+};
+
 export interface OpenAICallOptions extends BaseLanguageModelCallOptions {
   /**
    * Additional options to pass to the underlying axios request.
    */
-  options?: AxiosRequestConfig;
+  options?: OpenAICoreRequestOptions;
```
jacoblee93, line 83: This is going to affect someone out there... Don't think it's worth shimming though, especially since `headers` is thankfully named the same:

https://axios-http.com/docs/req_config

PR #2356, opened 2023-08-21
Add Vectara addFile function (extends #2184)

jacoblee93:
Extends #2184 by resolving merge conflicts and fixing the build.

CC @3eif


jacoblee93:
It's good to go @3eif, might suggest adding some docs but let's get this one first in since it's been a while!

PR #2358, opened 2023-08-22
 Add documentation for Vectara addFiles function

3eif:
Adds a description, argument types, and return type to the Vectara vector store addFiles function.


jacoblee93:
Thank you!

PR #2367, opened 2023-08-22
Add support for streaming responses for Amazon Bedrock LLM

Benyuel:
This PR adds support for using the Amazon Bedrock LLM with streaming.  See the added test for an example use.  

Fixes https://github.com/hwchase17/langchainjs/issues/2321


jacoblee93:
This looks really fantastic, but I don't think it really fits as a chat model (there's no dialog turns/messages, for example). The `OpenAIChat` class was really just a one-off when chat models were new.

We can just add streaming to the LLM version I think?

langchain/src/llms/bedrock-chat.ts
```diff
@@ -0,0 +1,267 @@
+import { SignatureV4 } from "@aws-sdk/signature-v4";
+import { defaultProvider } from "@aws-sdk/credential-provider-node";
```
jacoblee93, line 2: Let's combine this into the original `llms/bedrock.ts` and call it adding streaming support!

langchain/src/llms/bedrock-chat.ts
```diff
@@ -0,0 +1,267 @@
...
+
+  temperature?: number | undefined = undefined;
+
+  maxTokens?: number | undefined = undefined;
+
+  streaming?: boolean | undefined = false;
+
+  fetchFn: typeof fetch;
+
+  marshaller: EventStreamMarshaller = new EventStreamMarshaller(
```
jacoblee93, line 113: I think this has been deprecated?
Screenshot 2023-08-22 at 7 53 14 PM

langchain/src/llms/bedrock-chat.ts
```diff
@@ -0,0 +1,267 @@
...
+    });
+    if (response.status < 200 || response.status >= 300) {
+      throw Error(
+        `Failed to access underlying url '${url}': got ${response.status} ${
+          response.statusText
+        }: ${await response.text()}`
+      );
+    }
+
+    if (this.streaming) {
```
jacoblee93, line 218: We now have a new interface for `.stream()`:

https://js.langchain.com/docs/guides/expression_language/interface

Also, it'd be more elegant to always use the streaming call and implement non-streaming requests as just concatenating the entire stream:

https://github.com/hwchase17/langchainjs/blob/main/langchain/src/llms/ollama.ts#L124

jacoblee93:
Let me know if you need help using the newer package! I'm happy to dig in tomorrow.

langchain/src/llms/bedrock-chat.ts
Benyuel: ok, sounds good.  I will work towards that now.  Thank you for the review

langchain/src/llms/bedrock-chat.ts
Benyuel: Yes, sorry I missed that.  I will replace with a suitable alternative for this 👍

jacoblee93:
Thanks for the quick updates!

To get `.stream()` working you'll actually need to implement `_streamResponseChunks` as shown here:

https://github.com/hwchase17/langchainjs/blob/main/langchain/src/llms/openai.ts#L452

I can do this later today though - it should be very similar to what you have, and we can share the streaming logic.

jacoblee93:
@Benyuel Refactored a bit so we can support the `.stream()` method and added an integration test - can you confirm `yarn test:single langchain/src/llms/tests/bedrock.int.test.ts` works for you?

I also added retries and the ability to pass an abort signal to cancel in flight requests (hackily).

Also CC @vitaly-ps for a second look?

Benyuel:
Thanks so much @jacoblee93!  

I made a minor change to get the tests working for me, and also added passing in `temperature` and `maxTokens` into the request.  

I think this is ready from my perspective

> I also added retries and the ability to pass an abort signal to cancel in flight requests

I tested this works as well 👍

langchain/src/llms/tests/bedrock.int.test.ts
```diff
@@ -249,7 +278,7 @@ test("Test Bedrock LLM: stream method", async () => {
     console.log(chunk);
     chunks.push(chunk);
   }
-  expect(chunks.length).toBeGreaterThan(1);
+  expect(chunks.length).toBeGreaterThanOrEqual(1);
```
jacoblee93, line 281: Shouldn't this be greater than 1?

langchain/src/llms/tests/bedrock.int.test.ts
```diff
@@ -240,6 +242,33 @@ test("Test Bedrock LLM: stream method", async () => {
     maxTokens: 20,
     region,
     model,
+    async fetchFn(
```
jacoblee93, line 245: Ah I intended this one to test that this works with a live Bedrock instance, without a fetch function mock. Can you try that out?

jacoblee93:
Hey @Benyuel, I think you changed the test I added to use a mocked function - does this work with a live Bedrock instance?

langchain/src/llms/bedrock.ts
```diff
@@ -50,9 +61,9 @@ class BedrockLLMInputOutputAdapter {
     if (provider === "anthropic") {
       return responseBody.completion;
     } else if (provider === "ai21") {
-      return responseBody.completions[0].data.text;
+      return responseBody.data.text;
```
jacoblee93, line 64: This was wrong before?

langchain/src/llms/tests/bedrock.int.test.ts
Benyuel: oh sorry, you can revert the mock I added.  I just assumed no connection to bedrock so mocked it.  

To answer your question though, yes I tried this `.stream` against a live bedrock endpoint too.

langchain/src/llms/tests/bedrock.int.test.ts
Benyuel: The mock I added does streaming, but just 1 chunk.  
https://github.com/hwchase17/langchainjs/blob/7f2271769488152259791226d4c3767f60919ae9/langchain/src/llms/tests/bedrock.int.test.ts#L321-L337
That's why I swapped this.

langchain/src/llms/bedrock.ts
Benyuel: This was correct before for the non-streaming bedrock api.  But when I change to the bedrock `invoke-with-response-stream` endpoint, these change to what's in this PR currently.

jacoblee93:
Looks great! Thanks again for your patience!

uzaymacar:
Hey both, I've found this PR from one of the linked issues -- I'm still confused about how to implement streaming with Bedrock similar to what is shown in the docs: https://js.langchain.com/docs/modules/model_io/models/chat/how_to/streaming. Simply changing `ChatOpenAI` with `ChatBedrock` does not work, but maybe I need to do something else?

jacoblee93:
Ah apologies, we really need to update that - instead use:

https://js.langchain.com/docs/expression_language/interface#stream

uzaymacar:
Awesome thanks @jacoblee93, I'm now able to stream with Bedrock too 👍

PR #2368, opened 2023-08-22
Fix serialisation of StrOutputParser and ByteOutputParser

dqbd:
Missing `langchain` prefix in `lc_namespace` causes the deserialisation of StrOutputParser and BytesOutputParser to fail

SImilar to #2329


jacoblee93:
Gah my bad! Thank you!

PR #2385, opened 2023-08-23
Integrate NIBittensorLLM 

BitVoyagerMan:
- Add NIBittensorLLM endpoint based on power of Bittensor.


langchain/src/llms/bittensor.ts
```diff
@@ -0,0 +1,164 @@
...
+
+  /** Call out to NIBittensorLLM's complete endpoint.
+   Args:
+       prompt: The prompt to pass into the model.
+
+       Returns:
+   The string generated by the model.
+
+   Example:
+   let response = ai21._call("Tell me a joke.");
```
jacoblee93, line 56: Update string to bittensor

langchain/package.json
```diff
@@ -637,7 +640,7 @@
     "@xata.io/client": "^0.25.1",
     "@zilliz/milvus2-sdk-node": ">=2.2.11",
     "apify-client": "^2.7.1",
-    "axios": "^0.26.0",
+    "axios": "^0.26.1",
```
jacoblee93, line 643: Can we do this with just `fetch`? Otherwise people will need to install packages to use this

jacoblee93:
Thanks for this! It would be better to try to do this without optional deps - it looks like we can probably do this with fetch?

langchain/src/llms/bittensor.ts
BitVoyagerMan: Oh, thanks. will fix that comment.

langchain/package.json
BitVoyagerMan: Which is better between using axios@0.26.0 or fetch?
Because axios@0.26.0 is also working for me.

BitVoyagerMan: I resolved with fetch

PR #2386, opened 2023-08-23
NotionAPILoader property parsing

skarard:
This PR includes crucial enhancements to the Notion API Loader. The primary focus lies on refining functionalities for parsing properties of different types received from the Notion API. 

- Several new types, interfaces, functions, and variables have been added for robust assessment of API responses and precise parsing of their content.
- The `getPropValue`, `parsePageProperties`, and `parsePageDetails` functions have been significantly optimized to handle more property types.
- The "@notionhq/client" dependency has been upgraded to the latest version ensuring up-to-date functionalities.
- New testing protocols have been added for the enhanced parsing process along with necessary updates to existing test files.


skarard:
The unit test that's failing is: `src/document_loaders/tests/directory.test.ts`

The reason it fails is that I added a new json file which contains additional objects specifically for the NotionAPILoader properties parser. The `src/document_loaders/tests/directory.test.ts` unit relies on pdf, json, txt and csv files in the example_data directory to be static. I suggest the directly loader test is updated to occur on a subdirectory which isn't expect to change.

langchain/src/document_loaders/web/notionapi.ts
```diff
@@ -137,74 +138,167 @@ export class NotionAPILoader extends BaseDocumentLoader {
...
+      case "people":
+        return `[${prop[prop.type]
+          .map((v) => `["${v.object}", "${v.id}"]`)
+          .join(", ")}]`;
+      case "unique_id":
+        return `${prop[prop.type].prefix || ""}${prop[prop.type].number}`;
+      case "relation":
+        return `[${prop[prop.type].map((v) => `"${v.id}"`).join(", ")}]`;
+      // TODO: BLOCKER: Not all rollup types are supported in the SDK
+      // case "rollup": {
```
jacoblee93, line 235: Can we remove this comment or will it be a fast-follow?

skarard: It's dependent on Notion-SDK. I'm assuming that it won't be quick. 

One option is to enable the sub types defined here (which is what the notion SDK supports) for rollups and functions. Then return unsupported type for any subtypes that aren't included here.

I decided against it because I think it would be confusing for a user to return some rollup/function sub types and not others. 
Just as it is confusing to me why this is the case in the Notion SDK. 😅

skarard: Short answer, yes it can be removed.

skarard:
@jacoblee93 What's the best way to deal with the failing unit test?

Do you agree with my suggestion to create a new directory that contains data specifically for the directly loader test?

jacoblee93:
Yes sounds good sorry! Feel free to make changes.

langchain/src/document_loaders/tests/notionapi.int.test.ts
```diff
@@ -32,7 +32,7 @@ test("Test Notion API Loader Database", async () => {
   });
 
   const docs = await loader.load();
-  const titles = docs.map((doc) => doc.metadata.properties.title);
+  const titles = docs.map((doc) => doc.metadata.properties._title);
```
jacoblee93, line 35: Is this breaking?

jacoblee93:
Thank you!

jacoblee93:
Looks good overall but see comments

langchain/src/document_loaders/tests/notionapi.int.test.ts
skarard: For an example:
- Non-database Pages have the `...properties.title` property still available, as well as `...properties._title`. 
- A Database Page with the title properties called `Name` would have `...properties.Name` as well as `...properties._title`.

This allows a user to always access the `_title` property, irrespective of the database schema.

jacoblee93:
@skarard looks good but can you move the original files into this folder rather than just copying them?

Otherwise we'll have two copies of them

skarard:
The issue here is now each unit test that references the moved example files will need to be updated.

It's then unclear where to put your standalone example files.

I see the directory loader example folder as a snapshot of the examples folder when the directory loader unit tests were created. 

I thought about making unique files in the example folder for the directory loader, but that would cost extra effort for the same outcome just so that there are no duplicated contents in files.

The example folder for the directory loader could be built at runtime of the test with only the files specified in the unit test. 

Do you have any suggestions?

jacoblee93:
Just moved the notion file into its own dir and disabled recursive on the directory test. Will merge tonight, thank you!

PR #2389, opened 2023-08-23
Upgrade to OpenAI v4 (extends #2353)

jacoblee93:
Extends #2353 with some backcompat shims

Wasn't able to push on top of the branch @nsbradford - can you have a look?

CC @rattrayalex as well


langchain/src/chat_models/openai.ts
```diff
@@ -418,58 +408,9 @@ export class ChatOpenAI
...
-        if (err) {
-          throw err;
-        }
-        return chunk;
-      },
-      [Symbol.asyncIterator]() {
-        return this;
-      },
-    };
+    if (options.signal?.aborted) {
```
jacoblee93, line 411: CC @rattrayalex if the user aborts a stream, the SDK no longer throws an error. This is different from `fetch`'s behavior

langchain/src/util/async_caller.ts
```diff
@@ -80,7 +80,9 @@ export class AsyncCaller {
               if (
                 error.message.startsWith("Cancel") ||
                 error.message.startsWith("TimeoutError") ||
-                error.message.startsWith("AbortError")
+                error.message.startsWith("AbortError") ||
+                error.message.includes("Request was aborted") ||
```
jacoblee93, line 84: @rattrayalex is there any way to detect these errors other than a constructor check? This will fail in minified environments...

langchain/src/chat_models/openai.ts
```diff
@@ -712,56 +589,83 @@ export class ChatOpenAI
...
-        baseOptions: {
-          timeout: this.timeout,
-          ...this.clientConfig.baseOptions,
-        },
-      });
+        baseURL: endpoint,
+        timeout: this.timeout,
+        maxRetries: 0,
+      };
+      if (!params.baseURL) {
```
jacoblee93, line 648: @rattrayalex would be nice to not have to do this `delete` - could `baseURL: undefined` default to just the default OAI setting?

jacoblee93:
Some more work to do, perhaps with @rattrayalex's input?

langchain/src/chat_models/openai.ts
rattrayalex: Yes, that's intentional – other than backcompat, does that cause problems / is there a reason it's not desirable behavior?

BTW, you likely want to define an `AbortError` class and throw it below – people identify these by checking that the `.name` of the error is `AbortError`.

langchain/src/chat_models/openai.ts
rattrayalex: Ah, sorry about this - agree that's a bug, I've filed a ticket to fix.

langchain/src/util/async_caller.ts
rattrayalex: I think you want `error.name === 'AbortError'` ([docs](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal))

rattrayalex: Depending on what you want, we also export classes like `APIUserAbortError` and `APIConnectionTimeoutError`; you could check the `name` there too if you don't wan't to do an `instanceof`

jacoblee93: `error.name` I think just returned `Error` but will double check.

langchain/src/chat_models/openai.ts
jacoblee93: No, we can handle it this way - convention across other LLMs is to throw because they're fetch based so would like to keep it standard for us but no worries on your end then.

langchain/src/util/async_caller.ts
rattrayalex: got it, that might be because of this line: https://github.com/hwchase17/langchainjs/pull/2389/files/a0596c79e38da6f527a0a4578ba66bb53cfab999#diff-4d0f479be12d9b217277a26941bcfe72cfb9bc6b9bb9209099d6dd3294248bfbR412

which as I noted you may want to do something different with…

langchain/src/chat_models/anthropic.ts
```diff
@@ -377,6 +377,7 @@ export class ChatAnthropic extends BaseChatModel implements AnthropicInput {
         ...this.clientOptions,
         ...options,
         apiKey: this.anthropicApiKey,
+        maxRetries: 0,
```
nsbradford, line 380: this looks unrelated - was changing the Anthropic client intentional?

jacoblee93: Yes, oversight earlier but we handle retries natively through the `AsyncCaller` class for standardization.

nsbradford:
> Some more work to do, perhaps with @rattrayalex's input?

hi @jacoblee93 @rattrayalex are you still going through review? Any assistance you need?

jacoblee93:
Yes, have been quite busy but still plan to merge this after extensive testing!

langchain/src/chat_models/openai.ts
```diff
@@ -323,6 +314,14 @@ export class ChatOpenAI
     this.clientConfig = {
       apiKey: this.openAIApiKey,
       organization: this.organization,
+      baseURL: configuration?.basePath ?? fields?.configuration?.basePath,
+      dangerouslyAllowBrowser: true,
+      defaultHeaders:
```
jacoblee93, line 319: Am planning to just support previously passed `basePath`, `headers`, and `params` but could shim more things in here.

langchain/src/chat_models/tests/chatopenai.int.test.ts
```diff
@@ -63,7 +63,7 @@ test("Test ChatOpenAI Generate throws when one of the calls fails", async () =>
     chat.generate([[message], [message]], {
       signal: AbortSignal.timeout(10),
     })
-  ).rejects.toThrow("Cancel: canceled");
+  ).rejects.toThrow();
```
jacoblee93, line 66: Technically breaking but 🤷

langchain/src/chat_models/tests/chatopenai.int.test.ts
```diff
@@ -242,7 +242,7 @@ test("Test OpenAI with timeout in call options and node adapter", async () => {
   const model = new ChatOpenAI({ maxTokens: 5 });
   await expect(() =>
     model.call([new HumanMessage("Print hello world")], {
-      options: { timeout: 10, adapter: undefined },
+      options: { timeout: 10 },
```
jacoblee93, line 245: Custom adapters will no longer work but it's probably better long-term anyway

jacoblee93:
@rattrayalex @nsbradford @nfcampos it's massive but would appreciate a last look here if you have time!

langchain/src/llms/openai-chat.ts
```diff
@@ -531,7 +512,8 @@ export class PromptLayerOpenAIChat extends OpenAIChat {
         const promptLayerRespBody = await promptLayerTrackRequest(
           this.caller,
           "langchain.PromptLayerOpenAIChat",
-          { ...this._identifyingParams(), prompt },
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          { ...this._identifyingParams(), prompt } as any,
```
nsbradford, line 516: is this `any` cast necessary?

nsbradford (approved):
lgtm

jacoblee93:
> lgtm

Thank you!

jacoblee93:
No turning back now! Thanks all!

siisee11:
Can you briefly explain how this PR can solve https://github.com/hwchase17/langchainjs/issues/2346?

PR #2393, opened 2023-08-24
Export CallbackManagerForRetrieverRun

willemmulder:
This simply makes the CallbackManagerForRetrieverRun available, just like the other classes in the file (e.g. CallbackManagerForLLMRun, CallbackManagerForChainRun etc). I'm using it in a new MultiQueryRetriever.

### DIFF
--- a/langchain/src/callbacks/index.ts
+++ b/langchain/src/callbacks/index.ts
@@ -20,6 +20,7 @@ export {
 
 export {
   CallbackManager,
+  CallbackManagerForRetrieverRun,
   CallbackManagerForChainRun,
   CallbackManagerForLLMRun,
   CallbackManagerForToolRun,


jacoblee93:
Ah whoops, thanks!

jacoblee93:
Would you be interested in contributing the retriever you're building?

willemmulder:
@jacoblee93 Yeah sure, I actually have it ready and working, but I'm not sure if it is 'up-to-spec' and I'd have to actually add it to the source. Right now it's just a custom class that I import manually. I don't have tests, docs or anything else really. Would you be able to review? 

I might be able to create a PR tomorrow.

jacoblee93:
Yeah can do!

Docs and tests are nice but I can fill them in too.

PR #2394, opened 2023-08-24
Updated docs and metric types in langchain

mpskex:


jacoblee93:
Hey @mpskex - happy to merge this, but this will break anyone who's passing in `metric`. For backwards compatbility, do we want to add a shim? Or do you think usage of that parameter is low enough that it's ok?

mpskex:
> Hey @mpskex - happy to merge this, but this will break anyone who's passing in `metric`. For backwards compatbility, do we want to add a shim? Or do you think usage of that parameter is low enough that it's ok?

first of all this will only happen if the user is creating a new table. the existing table will not be effect unless the table is dropped manually. we are also able to handle case insensitive metric type under the hood.

jacoblee93:
Ok, sounds good to me!

PR #2400, opened 2023-08-24
Initial hub integration

jacoblee93:
CC @efriis @nfcampos


langchain/src/hub.ts
```diff
@@ -0,0 +1,21 @@
...
+export async function push(
+  repoFullName: string,
+  serializable: Serializable,
+  options?: HubPushOptions & ClientConfiguration
+) {
+  const client = new Client(options);
+  return client.push(repoFullName, JSON.stringify(serializable), options);
+}
+
+export async function pull<T extends Serializable>(
```
jacoblee93, line 14: @nfcampos this would be pretty difficult/impossible to infer right?

PR #2401, opened 2023-08-24
feat: Add Writer support

chase-crumbaugh:
Adds Writer as an LLM integration option. Mirrors [existing Python integration](https://github.com/langchain-ai/langchain/blob/master/libs/langchain/langchain/llms/writer.py).


jacoblee93:
A few built/dep fixes but looks great! Thank you!

chase-crumbaugh:
Thanks Jacob!

PR #2407, opened 2023-08-25
feat: custom descriptions to SqlDatabase

puigde:
This PR adds implementation and test for custom descriptions for tables in SqlDatabase ( already existing in [python version](libs/langchain/langchain/utilities/sql_database.py)) those descriptions are added to the db class and then used in the get tables info command.

@polpuigdemont


jacoblee93:
Looks good, thanks!

PR #2413, opened 2023-08-26
Local llama

nigel-daniels:
Fixes # No Llama-CPP llm available

Note, I'd like to know where I need to contribute a basic example for the docs, I look forward to discussing.


jacoblee93:
Whoa I was just about to start looking at this! Yes this would be awesome!

jacoblee93:
I think this package might be a better choice as it's more established? https://github.com/Atome-FE/llama-node

nigel-daniels:
@jacoblee93 k I'll take a look and update, I just happened to be playing with node-llama-cpp, I'll we work the library. BTW for docs where is a good point to add the appropriate page?

jacoblee93:
No worries, was just wondering!

nigel-daniels:
np, I just took a look over the llama-node repo, it looks like they are using a back level llama.cpp so currently only support ggml models not gguf. Although with llama.cpp going through internal changes gguf support will be in flux for a bit. This will need a close eye on updates!

nigel-daniels:
BTW I'm using a env variable `LLAMA_PATH` for the test code so the test has a level of portability. As a result I added the env var to `.env.example` even though it's only used for the tests. I was not 100% sure that was the correct approach.

docs/extras/modules/model_io/models/llms/integrations/llama_cpp.mdx
```diff
@@ -0,0 +1,28 @@
...
+
+## Setup
+
+You'll need to install the [node-llama-cpp](https://github.com/withcatai/node-llama-cpp) module to communicate with you local model.
+
+```bash npm2yarn
+npm install -S node-llama-cpp
+```
+
+You will also need a local Llama 2 model (or a model supported by [node-llama-cpp](https://github.com/withcatai/node-llama-cpp)). You will need to pass the path to this model to the LlamaCpp module as a part of the parameters (see example).
```
jacoblee93, line 21: Might be good to add more information on how to get a local model set up here? I personally wasn't able to.

jacoblee93:
Hey @nigel-daniels, I bumped the package you're using to 2.1.0 and changed a few things - however I got stuck trying to run a model locally - kept getting:

```
MacBook-Pro:llama.cpp jacoblee$ python3 convert.py models/7B/
Loading model file models/7B/consolidated.00.pth
Traceback (most recent call last):
  File "/Users/jacoblee/langchain/llama.cpp/convert.py", line 1172, in <module>
    main()
  File "/Users/jacoblee/langchain/llama.cpp/convert.py", line 1120, in main
    model_plus = load_some_model(args.model)
  File "/Users/jacoblee/langchain/llama.cpp/convert.py", line 1042, in load_some_model
    models_plus.append(lazy_load_file(path))
  File "/Users/jacoblee/langchain/llama.cpp/convert.py", line 744, in lazy_load_file
    elif struct.unpack('<Q', first8)[0] < 16 * 1024 * 1024:
struct.error: unpack requires a buffer of 8 bytes
```

Can you verify the updated package version works? And any ideas/other guides that could be helpful to link in the docs

docs/extras/modules/model_io/models/llms/integrations/llama_cpp.mdx
nigel-daniels: Great idea, I'll add some docs on the steps I took to get this running.

nigel-daniels:
I also just tried out 2.1.0 as it adds temperature to the model parameters, however it looks like a typo in the index.ts broke things, I just submitted a PR over there. I'll update when it's fixed and update their PR with that and the guide to getting a llama2 model in place.

jacoblee93:
Thank you, appreciate the followup!

nigel-daniels:
Ok just added a commit with an updated (working!) node-llama-cpp, also gave detailed instructions to converting the Llama2 model and quantizing it. Afaik the model you tried out was the base download from Meta, you will need to run it thru the llama.cpp conversion, the quantizing then makes it way smaller :)
With the new version of node-llama-cpp there is support for temp, and some added properties for the call, so I added `LlamaCppCallOptions`, hopefully that is the correct way to do this.

jacoblee93:
I think I was trying to compile with metal support - seems to be working properly with your guide! Thank you!

jacoblee93:
It's very slow though - might be good to include some bare-metal instructions at some point, but that can come later.

jacoblee93:
Build failure looks unrelated, I'll figure it out!

jacoblee93:
@nigel-daniels were you able to get this to build with GPU/metal support?

nigel-daniels:
There is an option for node-llama-cpp to run a download to install a metal or cuda version of llama.cpp, `npx node-llama-cpp download —metal`. [read me](https://github.com/withcatai/node-llama-cpp). I can always add a mention to the docs.

jacoblee93:
I couldn't seem to get it to work - kept getting the above error.

If you found a reliable set of instructions would you be up for adding them?

giladgd:
@jacoblee93 I think the error you get is unrelated to `node-llama-cpp`.
Try downloading and running this model with the latest version of `node-llama-cpp`: https://huggingface.co/TheBloke/Yarn-Llama-2-7B-128K-GGUF/blob/main/yarn-llama-2-7b-128k.Q5_K_M.gguf

PR #2415, opened 2023-08-26
Feat: Add collection metadata in ChromaDB vectorstore

### LINKED ISSUE #2412: Change chroma distance algorithm
With the langchain chroma vectorstore  there is no mechanism to adjust the formula for distance calculations from the default I2 algorithm.

Would love to be able to use cosine instead, I2 really does not seem to work for our use case

neebdev:
Added option to pass metadata when creating collection in ChromaDB.

Fixes #2412


jacoblee93:
Looks good, thank you!

PR #2423, opened 2023-08-28
Specify number of both child and parent documents in ParentDocumentRetriever

karol-f:
Allow for specifying number of both child and parent documents in `ParentDocumentRetriever` that will be returned.

Assume that you have 3 documents that are split by `ParentDocumentRetriever` in following way:
* 1 document (300 child documents, 3 parent documents)
* 2 document (500 child documents, 5 parent documents)
* 3 document (200 child documents, 2 parent documents)

If you use `ParentDocumentRetriever` in the old way, let's say with `childK === 20`, vectorstore might give you only 20 found child documents from many more available, which will result in sending to LLM only one parent document instead of more.

**This MR (a little inspired by the recently added `Similarity Score Threshold`) is adding the possibility to control the final number of returned documents.**

So now you have both:
* `childK`
* `parentK`

parameters that allow you to quickly choose how many small child docs you want to search and limit the number of parent documents that will be sent to LLM.

That way, specifying e.g. `childK === 300` and `parentK === 20` I can make sure that:
* all the relevant document are found in vectorstore
* only documents that fit context window of LLM are send to it

### DIFF
--- a/examples/src/retrievers/parent_document_retriever.ts
+++ b/examples/src/retrievers/parent_document_retriever.ts
@@ -23,9 +23,9 @@ const retriever = new ParentDocumentRetriever({
   // Note that this does not exactly correspond to the number of final (parent) documents
   // retrieved, as multiple child documents can point to the same parent.
   childK: 20,
-  // Optional `k` parameter to limit number of final, parent documents that will be returned from this retriever and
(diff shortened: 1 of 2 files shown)


jacoblee93:
Thank you!

PR #2426, opened 2023-08-28
Make Vectara metadata an object instead of an array 

3eif:
Before:
```
    [
      { name: 'lang', value: 'fra' },
      { name: 'section', value: '50' },
      { name: 'offset', value: '0' },
      { name: 'len', value: '34' },
      { name: 'CreationDate', value: '1237915995' },
      { name: 'Producer', value: 'OpenOffice.org 2.4' },
      { name: 'Creator', value: 'Writer' },
      {
        name: 'title',
        value: 'Bitcoin: A Peer-to-Peer Electronic Cash System'
      }
    ]
```

After:
```
    {
      lang: 'fra',
      section: '50',
      offset: '0',
      len: '34',
      CreationDate: '1237915995',
      Producer: 'OpenOffice.org 2.4',
      Creator: 'Writer',
      title: 'Bitcoin: A Peer-to-Peer Electronic Cash System'
    }
```

Addresses https://github.com/FlowiseAI/Flowise/pull/827

### DIFF
--- a/langchain/package.json
+++ b/langchain/package.json
@@ -1,6 +1,6 @@
 {
   "name": "langchain",
-  "version": "0.0.133",
+  "version": "0.0.134",
   "description": "Typescript bindings for langchain",
   "type": "module",
   "engines": {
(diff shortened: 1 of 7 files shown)


jacoblee93:
Cool with me but are we worried about breaking existing users?

jacoblee93:
Don't have anything better in mind that wouldn't be really weird or involve some kind of opt-in transition period - let's just go for it then.

ofermend:
Awesome!

PR #2432, opened 2023-08-29
Recursive url fix

skarard:
I noticed a bug when running max depth as zero, which should only return the requested page but instead would return pages as max depth +1.

I also added a test for this with max depth as one.


jacoblee93:
Thank you!

PR #2434, opened 2023-08-29
Add I/O to trace group

hinthornw:


langchain/src/callbacks/manager.ts
```diff
@@ -954,10 +962,13 @@ export async function traceAsGroup<T, A extends any[]>(
...
 ): Promise<T> {
   const traceGroup = new TraceGroup(groupOptions.name, groupOptions);
-  const callbackManager = await traceGroup.start();
+  const callbackManager = await traceGroup.start({ ...args });
   try {
-    return await enclosedCode(callbackManager, ...args);
-  } finally {
-    await traceGroup.end();
+    const result = await enclosedCode(callbackManager, ...args);
+    await traceGroup.end({ output: result });
```
hinthornw, line 968: @jacoblee93 is there a better way to do this than always wrapping under the output: key? E.g., if it's already an object, maybe we want to just log as that object directly

jacoblee93: Yeah I added something like that for runnables here:

https://github.com/hwchase17/langchainjs/blob/main/langchain/src/schema/runnable.ts#L18

jacoblee93:
Looks good, thank you!

Issue #2099, opened 2023-07-27
Adding finish_reason in the response 


maxpaleo:
Hi guys, would it be possible to add the finish_reason returned from the OpenAI api in the response? 
It was added to the Python repository recently: [Python](https://github.com/langchain-ai/langchain/issues/2702)

It's quite an important value to get in the response, as it states if a response was completed in full or not.

sarthak212:
I think it's availabe in llm but not in chat model can it be possible to add finish_reason in chatModel generation response.

baloodevil:
Can someone provide an example of how to get `GenerationInfo` which holds the `finish_reason`?  I thought it would be part of `ChainValues`, but apparently not.  I am using something like...

```javascript
  const model = new OpenAI({...});
  const chain = new LLMChain({ llm: model, prompt: promptTemplate });
  const response: ChainValues = await chain.call({ mytext, prompt });
```

I know this is a closed issue and probably not a great place for this question.  But I've tried using Discord but no one has responded with the correct answer and it is really close to what this issue addresses.

PR #2435, opened 2023-08-29
Add finishReason to openai generationInfo

### LINKED ISSUE #1556: Neither .call nor .generate return full response from OpenAi/AzureOpenAi
I need entire response from the provider I use. From the source code of openai.js it is seen that it only returns token information and the generated message skipping all other datapoints
image

Lets return the rest of the response in a dedicated field, 'metadata' for instance?

davidtom:
This small change adds `finishReason` to the `generationInfo` object of the OpenAI `_generate` method's `output`. This would be a huge help for us when using langchainjs, and it seems like others have requested it as well.

Fixes #1556 and #2099


jacoblee93:
Hey @davidtom! I'm aiming to get #2389 merged this week - will work this into that PR!

jacoblee93:
I think right now it'd be easier to lift if that's ok with you? There might be a rebase or two in the future there 🙈

jacoblee93:
Sounds good! Put a bunch of work into the other PR today - am planning to merge by Friday's release.

buttercoding:
Can you provide an example of usage?

matheuslemke:
@buttercoding here is an example: https://github.com/langchain-ai/langchainjs/issues/2099#issuecomment-1761466796

PR #2442, opened 2023-08-30
Minimax LLM Support

ysnows:
[Minimax](https://api.minimax.chat) is a Chinese startup that provides natural language processing models for companies and individuals.

It support function call , plugins (web_serarch) , glyph and embedding,  I have encapsulated all of these features into the LangChainJS , for the convenience of developers~


examples/src/models/chat/integration_minimax.ts
```diff
@@ -0,0 +1,60 @@
+import { HumanMessage } from "langchain/schema";
+import { ChatMinimax } from "langchain/chat_models/minimax";
+
+// Use abab5.5
+const abab5_5 = new ChatMinimax({
+  modelName: "abab5.5-chat",
+  botSetting: [
```
jacoblee93, line 7: Does the bot support system messages as well?

ysnows: The model itself does not support system messages, but its ‘botSetting’ parameters are similar to system message functions.

jacoblee93:
I'm unfortunately unable to directly test since I don't have permission to create an account, but all looks reasonable. Just made a few small docs and entrypoint fixes - thank you!

PR #2445, opened 2023-08-30
feat(redis vectorstore): add deleteDocuments to redis vectorstore's dropIndex

adam101:
As [`FT.DROPINDEX`](https://redis.io/commands/ft.dropindex/) does not delete the documents associated with the index by default, this change adds a new option to delete documents when dropping an index.


jacoblee93:
Thank you!

I think we'll eventually want to move this to a `delete` method but I can merge this for now and have it in tonight's release.

PR #2448, opened 2023-08-30
Support all Ollama model options

Basti-an:
According to the [Ollama source code](https://github.com/jmorganca/ollama/blob/main/api/types.go#L168C2-L168C2) there are a few model options which langchainjs does not support yet. This PR adds support for the missing Ollama model params.

### DIFF
--- a/langchain/src/chat_models/ollama.ts
+++ b/langchain/src/chat_models/ollama.ts
@@ -33,21 +33,47 @@ export class ChatOllama extends SimpleChatModel implements OllamaInput {
 
   baseUrl = "http://localhost:11434";
 
+  embeddingOnly?: boolean;
+
+  f16KV?: boolean;
+
+  frequencyPenalty?: number;
+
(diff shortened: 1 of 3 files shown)


jacoblee93:
Ah thanks! Can you add them to the chat model version as well?

Basti-an:
Sure! I completely forgot about the chat model, added the options for it as well.

Factoring out the common code sounds like a good idea, but I don't see an obvious way to do it based on the different inheritance of both models. Maybe using Typescript Mixins could help?

jacoblee93:
> Sure! I completely forgot about the chat model, added the options for it as well.
> 
> Factoring out the common code sounds like a good idea, but I don't see an obvious way to do it based on the different inheritance of both models. Maybe using Typescript Mixins could help?

Ah shoot you're right... If you don't mind let's just add it the slow way.

jacoblee93:
Ah you already have. Thank you!

PR #2454, opened 2023-08-31
Puigde/fix custom description from data source params

### LINKED ISSUE #2453: SqlDatabase: customDescriptions undefined when init through fromDataSourceParams
### Description
Last Week I added a feature to add customDescriptions to SqlDatabases. The logic works well if assign the property after db init but it is broken in fromDataSourceParams init.

The reason is because at the time of adding the param in the constructor the allTables property is undefined. So the following code does not accept any field for customDescription.

`this.customDescription = Object.fromEntries(
      Object.entries(fields?.customDescription ?? {}).filter(([key, _]) =>
        this.allTables
          .map((table: SqlTable) => table.tableName)
          .includes(key)
      )
    );`

### Reproducing
Init a sqlDatabase with customDescription using either fromDataSourceParams or fromOptionsParams.
For example:
`const db = await SqlDatabase.fromDataSourceParams({
    appDataSource: datasource,
    customDescription: {
      products: "this is a customDescription for table in db",
      productss: "this customDescription should be skipped as table not in db",
    },
  });`
Check that this.customDescription is undefined.

puigde:
Last Week I added a feature to add customDescriptions to SqlDatabases. The logic works well if assign the property after db init but it is broken in fromDataSourceParams init.

The reason is because at the time of adding the param in the constructor the allTables property is undefined. So the code does not accept any field for customDescription (see code details in the issue description).

This PR fixes that by:
* moving the customDescription assignation after the allTables is set
* adding test

Fixes #2453

### DIFF
--- a/langchain/src/sql_db.ts
+++ b/langchain/src/sql_db.ts
@@ -49,11 +49,6 @@ export class SqlDatabase
     this.ignoreTables = fields?.ignoreTables ?? [];
     this.sampleRowsInTableInfo =
       fields?.sampleRowsInTableInfo ?? this.sampleRowsInTableInfo;
-    this.customDescription = Object.fromEntries(
-      Object.entries(fields?.customDescription ?? {}).filter(([key, _]) =>
-        this.allTables.map((table: SqlTable) => table.tableName).includes(key)
-      )
-    );
   }
 
   static async fromDataSourceParams(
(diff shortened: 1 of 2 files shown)


jacoblee93:
Weird, thank you! Must have been missed from the start? Since it looks like `allTables` isn't even an accepted field on the constructor at all.

PR #2468, opened 2023-09-01
extract concept of openai style function-calling model

ysnows:
I think functions such as createExtractionChain, which is driven by function-call, can be implemented in a more general way through generics, because more and more models that support the OpenAI function style may come in the future， like AnthropicFunctions、Minimax~


jacoblee93:
Hey @ysnows, we upgraded the OpenAI SDK to v4 - can you take another look?

Would be great to refactor and get `StructuredTool` part of the base options as well at some point, but that can be in the future.

ysnows:
> 



> Hey @ysnows, we upgraded the OpenAI SDK to v4 - can you take another look?
> 
> Would be great to refactor and get `StructuredTool` part of the base options as well at some point, but that can be in the future.

> Ok, I'll try ~

vladgolubev:
2023-09-11T15 36 20-MGta96TS@2x

How do I pass `none` as `function_call` now? This is a valid input for OpenAI API, but after this PR typescript says it's incompatible. and `function_call: {name: 'none'}` throws error from OpenAI side as there is no function with that name

ysnows:
> @ysnows
> 
> 2023-09-11T15 36 20-MGta96TS@2x
> How do I pass `none` or `auto` as `function_call` now? This is a valid input for OpenAI API, but after this PR typescript says it's incompatible. and `function_call: {name: 'none'}` throws error from OpenAI side as there is no function with that name

> @vladgolubev 
Hi, I tried your case according to the screenshot you provided, but my result is as expected, this is my code:
```
const res = await model.call(
    [new HumanMessage("What is the weather in New York?")],
    {
        tools: [new SerpAPI()],
        function_call: "none"
    }
    // Tools will be automatically formatted as functions in the OpenAI format
);

console.log(res);
```
> and the result is:
```
AIMessage {
  lc_serializable: true,
  lc_kwargs: {
    content: `I'm sorry, but as a language model, I don't have real-time data. You can check the weather in New York by using a weather website or app, or by searching "weather in New York" on a search engine.`,
    additional_kwargs: { function_call: undefined }
  },
  lc_namespace: [ 'langchain', 'schema' ],
  content: `I'm sorry, but as a language model, I don't have real-time data. You can check the weather in New York by using a weather website or app, or by searching "weather in New York" on a search engine.`,
  name: undefined,
  additional_kwargs: { function_call: undefined }
}
```
I hope this can help you.

PR #2474, opened 2023-09-01
Verified and documented that text-bison-32k and chat-bison-32k work

afirstenberg:
Google announced public preview of text-bison-32k and chat-bison-32k.
Tested each and verified that they work.
Documented common available model names.

(There are no library code changes required.)


jacoblee93:
Awesome, thank you @afirstenberg!

PR #1950, opened 2023-07-13
feat: Add OpenAI Whisper document loader

konstantinov-raft:
Introducing an ability to load a document from local audio file using Open AI Whisper API

PR #2483, opened 2023-09-03
Video  File transcription loader

tyumentsev4:
Introducing an ability to load a transcription document of video file using OpenAI Whisper.

### DIFF
--- a/docs/extras/modules/data_connection/document_loaders/integrations/file_loaders/video_transcription.mdx
+++ b/docs/extras/modules/data_connection/document_loaders/integrations/file_loaders/video_transcription.mdx
@@ -1,0 +1,28 @@
+---
+hide_table_of_contents: true
+---
+
+# Video File Transcription loader
+
+This covers how to load video transcript into LangChain documents using OpenAI Whisper.
+
+## Setup
+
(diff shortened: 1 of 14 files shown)


docs/extras/modules/data_connection/document_loaders/integrations/file_loaders/video_transcription.mdx
```diff
@@ -0,0 +1,28 @@
...
+hide_table_of_contents: true
+---
+
+# Video File Transcription loader
+
+This covers how to load video transcript into LangChain documents using OpenAI Whisper.
+
+## Setup
+
+You'll need to install the [FFmpeg](https://ffmpeg.org/) and the [fluent-ffmpeg](https://www.npmjs.com/package/fluent-ffmpeg) package.
```
jacoblee93, line 11: We unfortunately shouldn't use FFmpeg - the licensing is very restrictive.

jacoblee93:
Thanks for this! #1950 is open and now unblocked. It also doesn't use FFMpeg, so it might be best to go with reviving that one.

jacoblee93:
Closed in favor of #1950

PR #2487, opened 2023-09-03
Fix random test fails due to segfault by chromadb & Float32Array not shared in jest context isolation

Durisvk:
The way chromadb imports @xenova/transformers package in file chromadb/src/embeddings/TransformersEmbeddingFunction.ts:33 makes it result in random segment fault errors terminating the tests prematurely. This fix contains a code that bypasses chromadb package and directly uses the @xenova/transformers package

Due to how jest isolates the context of each running test (https://github.com/xenova/transformers.js/issues/57, https://github.com/kayahr/jest-environment-node-single-context, https://github.com/jestjs/jest/issues/2549) - it makes it impossible for onnxruntime-node package to validate the array passed as an input to it is actually an `instanceof Float32Array` type. The `instanceof` results in false because the globals are different between context. This commit shares the Float32Array global between each context.


langchain/jest.env.cjs
```diff
@@ -0,0 +1,12 @@
+const { TestEnvironment } = require("jest-environment-node");
```
jacoblee93, line 1: We should have an explicit dev dependency on this then, right?

jacoblee93:
Makes sense, thank you for looking into this!

langchain/src/embeddings/hf_transformers.ts
```diff
@@ -1,4 +1,4 @@
-import { TransformersEmbeddingFunction } from "chromadb";
+import { Pipeline, pipeline } from "@xenova/transformers";
```
jacoblee93, line 1: This will be breaking (though it was undocumented before so not super concerned about that), and we'll want to add it to the docs and as a peer dep.

jacoblee93: Oh fascinating, the docs were wrong. Odd that this is working at all!

jacoblee93: Ohhh never mind this branch is off a branch! Let's just rebase it to `main`.

examples/src/models/embeddings/hf_transformers.ts
```diff
@@ -0,0 +1,14 @@
+import { HuggingFaceTransformersEmbeddings } from "langchain/embeddings/hf_transformers";
+
+const model = new HuggingFaceTransformersEmbeddings({
+  modelName: "Xenova/all-MiniLM-L6-v2", // In Node.js defaults to process.env.OPENAI_API_KEY
```
Durisvk, line 4: I wouldn't say that this defaults to `OPENAI_API_KEY` but I guess this is generated by some automation tool

examples/src/models/embeddings/googlepalm.ts
```diff
@@ -1,16 +1,14 @@
...
-  /* Embed queries */
-  const res = await model.embedQuery(
-    "What would be a good company name for a company that makes colorful socks?"
-  );
-  console.log({ res });
-  /* Embed documents */
-  const documentRes = await model.embedDocuments(["Hello world", "Bye bye"]);
-  console.log({ documentRes });
-};
+const model = new GooglePaLMEmbeddings({
```
Durisvk, line 3: I'm starting to dislike this commit cause it's messing up with other embeddings 😬

Durisvk: to give more clarifications - with new  `top-level async-await` ability this is all okay, but if someone is using `import { run } ...` from this module then it would fail for them

examples/src/models/embeddings/hf_transformers.ts
jacoblee93: Oh whoops thanks - was from the original PR

examples/src/models/embeddings/googlepalm.ts
jacoblee93: Fair, but these are just displayed in the docs though and aren't really meant to be imported. We're moving towards using top-level `async/await` for all examples and just thought I'd take the opportunity to change it as I saw it

jacoblee93: They also enforce that examples in the docs at least pass the TS compiler/linter/formatter

Durisvk:
You're the best @jacoblee93 for finalizing this! Hail to the opensource community!!

jacoblee93:
Thank you again!

PR #2499, opened 2023-09-04
Fix conversion of JSON schema array types to OpenAI function parameters

### LINKED ISSUE #2497: OpenAPI chain returns "object" type for "array" schemas
When building an OpenAPI chain using `createOpenAPIChain`, the OpenAI functions that are created incorrectly parse an "array" schema to an "object" type. This causes the LLM to generate incorrect function calls for any API endpoint that takes in an array as a parameter. For example, given an OpenAPI spec:

```
{
  "operationId": "readItemsCustomers",
  "description": "List the customers items.",
  "parameters": [
    {
      "description": "Control what fields are being returned in the object.",
      "in": "query",
      "name": "fields",
      "schema": {
        "type": "array",
          "items": {
            "type": "string"
          }
        }
      }
    },
    ...
  ]
}
```

Builds an OpenAI function with the following parameters for `fields`.

```
{
  "type": "object",
  "properties": {},
  "required": [],
  "additionalProperties": {}
}
```

Which causes OpenAPI to generate arguments like `POST { fields: {} }` instead of `POST { fields: [] }`.

The way I worked around this was to simplify the `convertOpenAPIParamsToJSONSchema` function to return the full JSON schema as-is, which worked for my case, but might break others:

https://github.com/programmarchy/langchainjs/commit/5259940ea9e2f23d6761f21f912d46dfe640bf5e#diff-7a2a689aa74dc204b5e798d15848447ef8c868fb9273d278e7f4efb552b3de7e

programmarchy:
This PR works around the bug where the OpenAPI chain parses array types as object types.

Another solution would be to modify the `convertOpenAPISchemaToJSONSchema` function to specifically handle the "array" type. However, for my use case, I didn't see the point, because OpenAPI 3.1 is already compatible with JSON Schema, so I just returned the schema directly.

Perhaps there's some historical context I'm missing, though?

Fixes #2497


jacoblee93:
It's been a little while but I think it was to make a best effort at supporting older versions of JSON Schema - would you be up for your alternative?

programmarchy:
> It's been a little while but I think it was to make a best effort at supporting older versions of JSON Schema - would you be up for your alternative?

That makes sense. Yes, I’ll give it a shot!

programmarchy:
@jacoblee93 I added a case to handle arrays, and changed the object case to handle when a property contains an array. Tests included.

programmarchy:
We may also want to consider how to handle schema `$ref`s, but should probably be in a different PR. (We'll have to think about how to handle circular references for that case.)

jacoblee93:
Thank you! Will have a look Monday.

jacoblee93:
Excellent, thank you!

PR #2510, opened 2023-09-05
Allow a wider variety of message like objects into chat prompt templates and chat models

jacoblee93:
CC @hinthornw

### DIFF
--- a/langchain/src/prompts/chat.ts
+++ b/langchain/src/prompts/chat.ts
@@ -13,6 +13,7 @@ import {
   PartialValues,
   SystemMessage,
   coerceBaseMessageLikeToMessage,
+  isBaseMessage,
 } from "../schema/index.js";
 import { Runnable } from "../schema/runnable.js";
 import {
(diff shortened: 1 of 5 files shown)


langchain/src/prompts/chat.ts
```diff
@@ -405,6 +407,44 @@ export interface ChatPromptTemplateInput<
...
+}
+
+export function coerceBaseMessagePromptTemplateLikeToMessage(
+  baseMessagePromptTemplateLike: BaseMessagePromptTemplateLike
+): BaseMessagePromptTemplate {
+  if (_isBaseMessagePromptTemplate(baseMessagePromptTemplateLike)) {
+    return baseMessagePromptTemplateLike;
+  }
+  const message = coerceBaseMessageLikeToMessage(baseMessagePromptTemplateLike);
+  if (message._getType() === "human") {
```
hinthornw, line 430: We've also added leniency for OAI naming here too

```
if message_type in ("human", "user"):
        message: BaseMessagePromptTemplate = HumanMessagePromptTemplate.from_template(
            template
        )
    elif message_type in ("ai", "assistant"):
...
    ```

PR #2512, opened 2023-09-05
Add Ollama `embeddings` integration

smndtrl:
This PR adds Ollama `embeddings` integration.

Not sure about the integration test though. Seems like the `chat` and `llm` also don't do anything like setting up a container. Maybe it doesn't make much sense to do integration tests for a simple fetch call regardless.


Basti-an:
The api/embedding route of the ollama server supports the same request options as the api/generate call, even if this is not well documented by ollama yet.

These options should ideally be supported as well, as some of them (like use_mmap) can impact embedding performance by a lot. 

I have implemented these options [in my own fork](https://github.com/Basti-an/langchainjs/blob/ollama-embeddings/langchain/src/embeddings/ollama.ts) (with conversion from camelCase to snake_case to be consistent with the llm/ollama and chatModels/ollama classes).

smndtrl:
Closed in favor of #2520

Basti-an:
Oh no, I only implemented my own version because I did not know you were already working on it! If you want to, you can reopen this PR and integrate my request options implementation 👍

smndtrl:
Haha no worries, let's go ahead with yours. Not sure if you need to do all the changes I did for the exports to work for me like the `create-entrypoints.js` script. I'm not a JS/TS tools person

Basti-an:
:) same! I will check out your changes and integrate the changes in config files that I might have missed! And next time I'll scan the open PRs before submitting my own 🙈

PR #2513, opened 2023-09-05
New Feat: Passing results from OpenAI moderation call

ppramesi:
This is very useful for handling non-english language moderation inputs, as OpenAI moderation can sometimes result in false negatives but still detects the usage of some sort of abusive language. So this pull request allows users to set the boundary themselves by filtering by category_scores (see [OpenAI moderation doc](https://platform.openai.com/docs/guides/moderation/quickstart)).


jacoblee93:
Looks good to me, thank you!

PR #2520, opened 2023-09-05
Add Ollama embeddings support

Basti-an:
Based on the recent [PR](https://github.com/langchain-ai/langchain/pull/10124) by herrjemand in the Langchain python repository I have implemented a class for embeddings using Ollama.

I have only added a little bit of documentation and two simple test cases. This PR might very well not be production ready yet, feedback is much appreciated.

Current problems I see with the implementation:

1. The default optional constructor params "maxConcurrency" and "maxRetries" have to exist in order to extend the base Embeddings class, but are unused by the OllamaEmbeddings class.

2. In order to be consistent with the llms/ollama and chatModels/ollama classes, the api request params are supplied in camelCase ("useMMap") and are then converted to the snake_case format used by the ollama API ("use_mmap").
This is rather awkward, imo.

### DIFF
--- a/langchain/src/embeddings/ollama.ts
+++ b/langchain/src/embeddings/ollama.ts
@@ -28,18 +28,18 @@ export class OllamaEmbeddings extends Embeddings {
 
   requestOptions?: OllamaRequestParams["options"];
 
-  constructor(params: OllamaEmbeddingsParams) {
-    super(params);
+  constructor(params?: OllamaEmbeddingsParams) {
+    super(params || {});
 
-    if (params.model) {
+    if (params?.model) {
       this.model = params.model;
     }
 
-    if (params.url) {
+    if (params?.url) {
       this.baseUrl = params.url;
     }
 
-    if (params.requestOptions) {
+    if (params?.requestOptions) {
       this.requestOptions = this._convertOptions(params.requestOptions);
     }
   }
--- a/langchain/src/embeddings/ollama.ts
+++ b/langchain/src/embeddings/ollama.ts
@@ -24,9 +24,9 @@ interface OllamaEmbeddingsParams extends EmbeddingsParams {
 export class OllamaEmbeddings extends Embeddings {
   model: string;
 
-  requestOptions?: OllamaRequestParams["options"];
-
   baseUrl = "http://localhost:11434";
+
+  requestOptions?: OllamaRequestParams["options"];
 
   constructor(params: OllamaEmbeddingsParams) {
     super(params);
--- a/langchain/src/embeddings/ollama.ts
+++ b/langchain/src/embeddings/ollama.ts
@@ -12,7 +12,7 @@ interface OllamaEmbeddingsParams extends EmbeddingsParams {
   model?: string;
 
   /** Base URL of the Ollama server, defaults to "http://localhost:11434" */
-  url?: string;
+  baseUrl?: string;
 
   /** Advanced Ollama API request parameters in camelCase, see
    * https://github.com/jmorganca/ollama/blob/main/docs/modelfile.md#valid-parameters-and-values
@@ -35,8 +35,8 @@ export class OllamaEmbeddings extends Embeddings {
       this.model = params.model;
     }
 
-    if (params?.url) {
-      this.baseUrl = params.url;
+    if (params?.baseUrl) {
+      this.baseUrl = params.baseUrl;
     }
 
     if (params?.requestOptions) {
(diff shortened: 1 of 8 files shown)


jacoblee93:
Thank you for this!

Agree on the mapping being awkward but it's the general style of the repo/JS generally at the moment - it's nice to keep people from guessing which style to use in general.

langchain/src/embeddings/ollama.ts
```diff
@@ -0,0 +1,134 @@
...
+    }
+  }
+
+  /** convert camelCased Ollama request options like "useMMap" to
+   * the snake_cased equivalent which the ollama API actually uses.
+   * Used only for consistency with the llms/Ollama and chatModels/Ollama classes
+   */
+  _convertOptions(requestOptions?: CamelCasedRequestOptions) {
+    if (!requestOptions) {
+      return;
```
jacoblee93, line 51: Might suggest throwing here instead or defaulting to an empty object?

langchain/src/embeddings/ollama.ts
```diff
@@ -0,0 +1,134 @@
...
+        snakeCasedOptions[snakeCasedOption] = value;
+      }
+    }
+    return snakeCasedOptions;
+  }
+
+  async _request(prompt: string): Promise<number[]> {
+    const { model, baseUrl, requestOptions } = this;
+
+    const response = await fetch(`${baseUrl}/api/embeddings`, {
```
jacoblee93, line 100: We can wrap this call with `this.caller` and take advantage of the `maxConcurrency` and `maxRetries`:

https://github.com/hwchase17/langchainjs/blob/main/langchain/src/embeddings/base.ts#L21
https://github.com/hwchase17/langchainjs/blob/main/langchain/src/embeddings/openai.ts#L237

Could see at least `maxConcurrency` being useful.

langchain/src/embeddings/ollama.ts
```diff
@@ -0,0 +1,134 @@
...
+
+    for await (const prompt of strings) {
+      const embedding = await this._request(prompt);
+      embeddings.push(embedding);
+    }
+
+    return embeddings;
+  }
+
+  embedDocuments = async (documents: string[]) => this._embed(documents);
```
jacoblee93, line 130: Style wise we usually go for:

```typescript
  async embedDocuments(documents: string[]) {
    ...
  }
```

langchain/src/embeddings/ollama.ts
```diff
@@ -0,0 +1,134 @@
...
+
+type CamelCasedRequestOptions = Omit<OllamaInput, "baseUrl" | "model">;
+
+/**
+ * Interface for OllamaEmbeddings parameters. Extends EmbeddingsParams and
+ * defines additional parameters specific to the OllamaEmbeddings class.
+ */
+interface OllamaEmbeddingsParams extends EmbeddingsParams {
+  /** The Ollama model to use, e.g: "llama2:13b" */
+  model: string;
```
jacoblee93, line 12: Should we have a default here like the others?

jacoblee93 (commented):
Looks great overall! 

We'll also want docs but I'm happy to take over as well.

Pinged the Ollama team for a second opinion too.

langchain/src/embeddings/ollama.ts
Basti-an: The llm and chatModel classes default to "llama2" here, so I will do it as well - thank you for all the suggestions, will implement ASAP. Much appreciated as I am new to this project andopen source contributions in general, so excuse any rookie mistakes!

jmorganca:
Overall looks great!

langchain/src/embeddings/ollama.ts
jmorganca: Would definitely recommend this as well! `llama2` will work, and we're looking to add other embedding models that might be better defaults later on!

Basti-an: done!

langchain/src/embeddings/ollama.ts
Basti-an: done!

langchain/src/embeddings/ollama.ts
Basti-an: done!

langchain/src/embeddings/ollama.ts
Basti-an: not necessary anymore, since the function is only called when requestOptions are set and the argument is not optional anymore after refactoring.

Basti-an:
@jacoblee93 I have implemented your suggestions and added some basic documentation for the embedding model.

jacoblee93:
Thanks very much!

PR #2521, opened 2023-09-05
Remove weird peer dependency

jacoblee93:


jacoblee93:
FYI @ovaistariq

ovaistariq:
> FYI @ovaistariq it's because of this:
> 
> https://github.com/hwchase17/langchainjs/pull/2521/files#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2deL14623

@jacoblee93 The reason for importing `json-bigint@github:sidorares/json-bigint` is that the currently released version of `json-bigint` has a bug that causes it to parse floating point numbers incorrectly in certain cases. This is [fixed](https://github.com/sidorares/json-bigint/commit/3530541b016d9041db6c1e7019e6999790bfd857) in `main` but has yet to be released. The last time there was a release was 2020, so I guess it is unmaintained.

PR #2522, opened 2023-09-05
Add support for querying multiple corpora in Vectara

3eif:
This PR adds support for querying multiple corpora at a time.

### DIFF
--- a/langchain/src/vectorstores/tests/vectara.int.test.ts
+++ b/langchain/src/vectorstores/tests/vectara.int.test.ts
@@ -172,12 +172,11 @@ describe("VectaraStore", () => {
       );
       expect(results.length).toBeGreaterThan(0);
       expect(results[0].pageContent.length).toBeGreaterThan(0);
-      expect(results[0].metadata.length).toBeGreaterThan(0);
       // Query filtered on French, so we expect only French results
       const hasEnglish = results.some(
         (result) =>
           // eslint-disable-next-line @typescript-eslint/no-explicit-any
-          result.metadata.find((m: any) => m.name === "lang")?.value === "eng"
+          esult.metadata.lang === "eng"
       );
       expect(hasEnglish).toBe(false);
     });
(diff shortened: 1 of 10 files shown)


langchain/src/vectorstores/tests/vectara.int.test.ts
```diff
@@ -107,7 +109,9 @@ describe("VectaraStore", () => {
     beforeAll(async () => {
       store = new VectaraStore({
         customerId: Number(process.env.VECTARA_CUSTOMER_ID) || 0,
-        corpusId: Number(process.env.VECTARA_CORPUS_ID) || 0,
+        corpusId:
```
jacoblee93, line 112: Might want to check for an empty array/NaN here?

jacoblee93:
Small comment around bounds checking but looks fine otherwise - thank you!

3eif:
> Small comment around bounds checking but looks fine otherwise - thank you!

Thanks for the catch, should be fixed now!

langchain/src/vectorstores/vectara.ts
```diff
@@ -102,11 +102,19 @@ export class VectaraStore extends VectorStore {
...
+        .map((id) => Number(id));
     if (!corpusId) {
       throw new Error("Vectara corpus id is not provided.");
     }
-    this.corpusId = corpusId;
+
+    if (typeof corpusId === "number") {
+      this.corpusId = [corpusId];
+    } else {
+      this.corpusId = corpusId;
```
jacoblee93, line 116: This could still be an empty array if `process.env.VECTARA_CORPUS_ID === ""` though right? Or if the env var was set to something not a number?

Happy to merge anyway if you're not concerned about it

3eif: Sorry I completely missed this, should be fix now

jacoblee93:
Thank you!

PR #2523, opened 2023-09-05
Loosen outputParser regex to allow lack of space

philipmw:
Sometimes the LLM won't put spaces between key and value, which causes an error with the current regex. With the update, it works.


jacoblee93:
Thank you!

PR #2537, opened 2023-09-06
Switch to `wrangler deploy` because `wrangler publish` is deprecated.

obi1kenobi:
CI said so:
image

### DIFF
--- a/test-exports-cf/src/index.ts
+++ b/test-exports-cf/src/index.ts
@@ -3,7 +3,7 @@
  *
  * - Run `wrangler dev src/index.ts` in your terminal to start a development server
  * - Open a browser tab at http://localhost:8787/ to see your worker in action
- * - Run `wrangler publish src/index.ts --name my-worker` to publish your worker
+ * - Run `wrangler deploy src/index.ts --name my-worker` to publish your worker
  *
  * Learn more at https://developers.cloudflare.com/workers/
  */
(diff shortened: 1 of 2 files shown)


test-exports-cf/src/index.ts
```diff
@@ -37,16 +37,16 @@ export default {
     ctx: ExecutionContext
   ): Promise<Response> {
 
-    const constructorParameters
```
nfcampos, line 40: Why is this changing?

nfcampos (commented):
Seems good minus 1 comment

test-exports-cf/src/index.ts
obi1kenobi: Ugh, sorry about that. I think my editor applied some Prettier rules I had configured for another repo.

I'll just disable that plugin and re-create this change.

test-exports-cf/src/index.ts
```diff
@@ -43,7 +43,7 @@ export default {
         azureOpenAIApiInstanceName: env.AZURE_OPENAI_API_INSTANCE_NAME,
         azureOpenAIApiDeploymentName: env.AZURE_OPENAI_API_DEPLOYMENT_NAME,
         azureOpenAIApiVersion: env.AZURE_OPENAI_API_VERSION,
-      }
```
obi1kenobi, line 46: This is just some trailing whitespace that my editor removed, I can leave it be if you'd prefer — all the same to me.

PR #2540, opened 2023-09-06
`test-exports-*` are built in docker, don't build them globally too.

obi1kenobi:
The `docker-ci-entrypoint.sh` script that runs inside all export containers contains a `yarn build`, so we don't need the CI `yarn build` to build them too.

If you'd prefer `yarn build` to build everything, I can instead make CI use a different build command — perhaps `yarn build:ci`?

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   },
   "packageManager": "yarn@3.4.1",
   "scripts": {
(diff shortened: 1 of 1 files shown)


jacoblee93:
Looks good, thanks!

PR #2542, opened 2023-09-06
Remove docs build since Vercel is doing it, and don't run code tests on docs-only changes.

obi1kenobi:
Vercel is already building the docs, so we don't need to do it too:
- Don't run `yarn build` as a separate step, since that only builds docs, examples, and langchain — and the latter two are already tested in unit tests.
- Skip the CI workflow if *only* docs were changed: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,8 +39,8 @@ jobs:
       - name: Check linting
         run: yarn run lint
 
-  build:
-    name: Build and check types
+  build-docs:
+    name: Build and check docs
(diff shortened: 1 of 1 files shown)


nfcampos (approved):

obi1kenobi:
@nfcampos, @jacoblee93 and I changed the approach here after chatting — the original approach didn't work as expected, and there was a more complicated way to do a dev build but we figured it wasn't worth it right now given Vercel building our docs.

PR #2543, opened 2023-09-06
Use concurrency groups to cancel pointless workflows early.

obi1kenobi:
If another push to the same PR or branch happens while this workflow is still running, cancel the earlier run in favor of the next run.

There's no point in testing an outdated version of the code. GitHub only allows a limited number of job runners to be active at the same time, so it's better to cancel pointless jobs early so that more useful jobs can run sooner.


jacoblee93:
Nice!

PR #2568, opened 2023-09-08
perf(pinecone): update pinecone vectorstore to use Promise.all

seawatts:
Pinecone vectorstore is waiting for each request to finish before starting the new one. This code updates the requests to use Promise.all instead of a for loop.


jacoblee93:
Thanks for this! We don't actually have lodash as a dep so I refactored it a bit. Also allowed it to take `maxConcurrency` and `maxRetries`.

seawatts:
Of course! Glad I could help. Thanks for those other updates.

PR #2579, opened 2023-09-08
Support Google Vertex AI on Edge

dqbd:
Due to `google-auth-library` being pure Node.js only, add a secondary implementation replacing `google-auth-library` with `kriasoft/web-auth-library`


jacoblee93:
CC @afirstenberg

afirstenberg:
Well, we had discussed this possibility, so it makes sense.
I've only skimmed the code so far, will try to read it in more detail this weekend and test it.

@dqbd - Any particular reason to add it as a secondary auth method instead of just using web-auth-library instead of Google's?

jacoblee93:
> Well, we had discussed this possibility, so it makes sense. I've only skimmed the code so far, will try to read it in more detail this weekend and test it.
> 
> @dqbd - Any particular reason to add it as a secondary auth method instead of just using web-auth-library instead of Google's?

Tree shaking :( bundlers will pick up Node imports even if they're unused and it'll mess things up, so we need different entrypoints and files.

afirstenberg:
> I think we want to avoid breaking folks using the filesystem to import credentials, especially enterprise-focused devs who expect it to be that way

If so, I feel like adding the lines to check if we have fs at runtime and the credentials specified in the environment, and if so, to load it in would be simpler. On the other hand, I agree, I want to make sure that all the currently supported auth modes are available.

It isn't a strong opinion, I guess, I'm just thinking about what the documentation will look like trying to explain the various configuration routes, and I worry it will be too complicated.

jacoblee93:
> > I think we want to avoid breaking folks using the filesystem to import credentials, especially enterprise-focused devs who expect it to be that way
> 
> If so, I feel like adding the lines to check if we have fs at runtime and the credentials specified in the environment, and if so, to load it in would be simpler. On the other hand, I agree, I want to make sure that all the currently supported auth modes are available.
> 
> It isn't a strong opinion, I guess, I'm just thinking about what the documentation will look like trying to explain the various configuration routes, and I worry it will be too complicated.

Your approach would definitely be simpler - bundlers are not very bright though and can't generally do conditional imports :(

jacoblee93:
@dqbd did some refactoring and it should be good to go but couldn't set up corepack - are you able to shim/otherwise verify this works ok in an Edge function?

dqbd:
> > I think we want to avoid breaking folks using the filesystem to import credentials, especially enterprise-focused devs who expect it to be that way
> 
> If so, I feel like adding the lines to check if we have fs at runtime and the credentials specified in the environment, and if so, to load it in would be simpler. On the other hand, I agree, I want to make sure that all the currently supported auth modes are available.

Hey @afirstenberg!

Unfortunately from the bundlers' PoV it can't statically determine whether a check is valid for specific environments, thus it will bundle all dependencies regardless, which will fail if included in edge runtime.

docs/extras/modules/model_io/models/chat/integrations/google_vertex_ai.mdx
```diff
@@ -6,10 +6,17 @@ import CodeBlock from "@theme/CodeBlock";
...
 # ChatGoogleVertexAI
 
-The Vertex AI implementation is meant to be used in Node.js and not
-directly from a browser, since it requires a service account to use.
+LangChain.js supports Google Vertex AI chat models as an integration.
+It supports two different methods of authentication based on whether you're running
+in a Node environment or a web environment.
 
-Before running this code, you should make sure the Vertex AI API is
+## Setup
```
afirstenberg, line 13: This (and probably much of the configuration notes) is going to be repetitive between four (so far) modules. Perhaps it would makes sense to move them all into a common page (perhaps something under Guides -> Deployment?) and either include or reference them all in the module documentation?

jacoblee93: We can factor auth instructions out into a separate .mdx file and import it to avoid repetition in a future PR

langchain/src/chat_models/tests/chatgooglevertexai_web.int.test.ts
```diff
@@ -0,0 +1,135 @@
...
+} from "../../prompts/index.js";
+import { ConversationChain } from "../../chains/conversation.js";
+import { BufferMemory } from "../../memory/buffer_memory.js";
+import { ChatGoogleVertexAI } from "../googlevertexai/web.js";
+
+describe.skip("ChatGoogleVertexAIWeb", () => {
+  test("call", async () => {
+    const chat = new ChatGoogleVertexAI();
+    const message = new HumanMessage("Hello!");
+    const res = await chat.call([message]);
```
afirstenberg, line 22: When running this test (and every test that uses chat.call()) the test is failing with the error:

> Cannot read properties of undefined (reading 'message')

typically at chat_models/base.ts:351:30

The tests that are using generate() seem to be working correctly.

afirstenberg: (For the record - the non-web test worked fine.)

langchain/src/util/googlevertexai-webauth.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    const credentials =
+      options?.credentials ??
+      getEnvironmentVariable("GOOGLE_VERTEX_AI_WEB_CREDENTIALS");
+    if (credentials === undefined)
+      throw new Error(
+        `Credentials not found. Please set the GOOGLE_VERTEX_AI_WEB_CREDENTIALS or pass credentials into "authOptions.credentials".`
+      );
+    this.credentials = getCredentials(credentials);
+    console.log(this.credentials);
```
afirstenberg, line 31: This needs to be deleted before merge

langchain/src/chat_models/tests/chatgooglevertexai_web.int.test.ts
afirstenberg: Looks like the credentials I was using weren't permitted or set correctly or something. Trying it with different credentials worked fine.

I'm ok with this, but we may want to see if there's a better error that gets returned or logged that indicates an auth problem.

jacoblee93: Weird - yes let's definitely track that down

afirstenberg:
With the notes above, these look good to me.
Is the intent to do this release and then do additional PRs for the other modules?

jacoblee93:
> With the notes above, these look good to me.
> Is the intent to do this release and then do additional PRs for the other modules?

Thank you very much!

The immediate motivation was to support Vertex chat models in the LangSmith playground and to do it in a way that wouldn't be disruptive to others using the Node flow, so I think we'll probably leave it for now until more community requests inevitably come in. I think it sets things up nicely for everything else though.

langchain/src/util/googlevertexai-webauth.ts
jacoblee93: Good catch thanks!

jacoblee93:
Looks good, thank you!

langchain/src/chat_models/tests/chatgooglevertexai_web.int.test.ts
dqbd: Should be fixed by throwing an error if failed to obtain an access token

jacoblee93:
Thank you both!

PR #2582, opened 2023-09-08
Get transcript text by ID in AssemblyAI

Swimburger:
If a user already has transcribed an audio or video file, the user can now provide the ID to load the transcript without having to re-transcribe the audio/video file.

Types are now coming from the AssemblyAI node SDK. These type names/contents aren't entirely the same as before.
Do we consider this a breaking change? How do we handle breaking changes?


langchain/src/document_loaders/web/assemblyai.ts
```diff
@@ -168,37 +151,58 @@ export class AudioTranscriptSentencesLoader extends AssemblyAILoader {
...
+   * @param assemblyAIOptions The options to configure the AssemblyAI loader.
+   * Configure the `assemblyAIOptions.apiKey` with your AssemblyAI API key, or configure it as the `ASSEMBLYAI_API_KEY` environment variable.
+   */
+  constructor(
+    params: CreateTranscriptParameters | string,
+    private subtitleFormat: SubtitleFormat = 'srt',
     assemblyAIOptions?: AssemblyAIOptions
   ) {
-    super(assemblyAIOptions);
+    super(params as string, assemblyAIOptions);
```
jacoblee93, line 195: This cast looks weird? Why not make `params: string`?

langchain/src/document_loaders/web/assemblyai.ts
```diff
@@ -1,23 +1,16 @@
...
-  AssemblyAIOptions,
-  CreateTranscriptParams,
-  SubtitleFormat,
-  Transcript,
-  TranscriptSegment,
-} from "../../types/assemblyai-types.js";
+import { AssemblyAIOptions } from "../../types/assemblyai-types.js";
 
-export * from "../../types/assemblyai-types.js";
+export type * from "../../types/assemblyai-types.js";
```
jacoblee93, line 7: Hasn't this been deleted below? We should export types from the Client instead right?

jacoblee93 (commented):
Thanks for this - a few questions!

jacoblee93:
> If a user already has transcribed an audio or video file, the user can now provide the ID to load the transcript without having to re-transcribe the audio/video file.
> 
> Types are now coming from the AssemblyAI node SDK. These type names/contents aren't entirely the same as before. Do we consider this a breaking change? How do we handle breaking changes?

In this case not much we can do unless you want to make some kind of mapping layer. Overall it looks pretty similar?

langchain/src/document_loaders/web/assemblyai.ts
Swimburger: There's one type in that file, and re-exporting everything from the SDK. Should I approach this differently?

langchain/src/document_loaders/web/assemblyai.ts
Swimburger: This is for TypeScript to pick one of the constructor overloads. I merged the overloads into a single constructor for simplicity.

langchain/src/document_loaders/web/assemblyai.ts
jacoblee93: Oh my mistake, I thought the file was deleted

jacoblee93:
Thank you!

Swimburger (commented):
LGTM!

PR #2591, opened 2023-09-09
Rename README to README.md to fix preview in GitHub

machulav:


jacoblee93:
Ah I was wondering why this was broken after the rename - thank you for finding this!

PR #2603, opened 2023-09-11
refactor(vectorstores): allow textKey overwrite

ahmedcoder01:
- User can overwrite the textKey field in the metadata


jacoblee93:
Are we sure this is good behavior? What's the use-case for not returning any text in the document?

ahmedcoder01:
> * User can overwrite the textKey field in the metadata

Thanks for the feedback. This change offers users flexibility with the textKey metadata, catering to varied requirements. My use-case involved customizing textKey based on certain conditions. The overwrite provides more control without preventing text return. It's an option for users needing customization.

Let me know your thoughts!

jacoblee93:
I don't see the purpose and it seems like it could introduce gotchas where you'd just be returning a document with different text from what they were stored as. 

Going to close but feel free to reopen if you have a compelling use-case.

PR #2604, opened 2023-09-11
Add file name property to addFiles function in Vectara

3eif:
We are aware that this introduces a breaking change and we are okay with that.


jacoblee93:
Thank you!

Issue #2600, opened 2023-09-11
Pinecone: Self Query -> Number Error


ttessarolo:
I have a metadata typeof Number called "year of production"
I submit a query like: "Movie produced after 1973" 
```javascript
const selfQueryRetriever = await SelfQueryRetriever.fromLLM({
  llm,
  vectorStore,
  documentContents,
  attributeInfo,
  structuredQueryTranslator: new PineconeTranslator(),
});
const result = await selfQueryRetriever.getRelevantDocuments("Movie produced after 1973");
```

it throws: ***[PineconeError: PineconeClient: Error calling query: PineconeError: the $gt operator must be followed by a number, got "1973" instead]***

The prompt doesn't specify that numerics fields should returns as numbers.

ttessarolo:
The problem is that I'm not composing the query. I'm just using ***selfQueryRetriever***.
As I've specified the problem is in the prompt that don't impose a typeof Number in return query for numbers.

jacoblee93:
Interesting - CC @ppramesi

ttessarolo:
I've to add the 9 to 10 the ***selfQueryRetriever*** returns no documents... [pinecone]

ppramesi:
Ah I think I see the problem. The base query translator accepts the value generated as is, when it should be checked and casted appropriately. I'll create a pull request fixing this in a bit.

jacoblee93:
Will be live in 0.0.149, thanks for the report and thank you @ppramesi for the fix!

PR #2613, opened 2023-09-13
Fix: Self-query base query translator casting number fix

ppramesi:
Fixes issue #2600 and made value type detection more robust, also added test cases for the casting/type detection, just in case.


jacoblee93:
Thank you!

PR #2615, opened 2023-09-13
Refactor: add combineLLM & refineLLM to loadQAChain&loadSummarizationChain

ysnows:
Refactor the code to achieve two goals:
1. Utilize a dedicated Language Model (LLM) as the final summary LLM to cater to specific LLM requirements at different stages of the chain.
2. Streamline the final results only, excluding the intermediate step's output from the streaming process.


jacoblee93:
Thanks, looks great!

PR #2618, opened 2023-09-13
feat(vectorstore): Add Custom IDs Option param to OpenSearch addVectors Method

omriTelemPanda:
**Description**:

This PR introduces an enhancement to the `addVectors` method in our OpenSearch module. The primary change is the inclusion of an optional `ids` parameter, which allows users to provide custom IDs for the documents they're adding to the OpenSearch index.

**Key Changes**:
- Added an optional parameter `options` to the `addVectors` function, which can accept a list of custom document IDs.
- If no custom IDs are provided, the method defaults to generating UUIDs, ensuring backward compatibility.
- Updated the `operations` construction logic to accommodate either custom or default UUID-based IDs.

This enhancement provides more flexibility to users who may have specific ID structures in their systems and want to retain them in the OpenSearch index.

Please review the changes and provide feedback!


jacoblee93:
Looks great! Thank you!

PR #2622, opened 2023-09-13
Fix chroma documentation

raphaelpra:
You must enter the cloned folder before launching `docker-compose` command.


jacoblee93:
Thank you!

PR #2625, opened 2023-09-13
Initial attempt at VertexAI streaming

### LINKED ISSUE #2594: Add support for Vertex AI Generative AI streaming response
REST responses can now provide a stream tokens as they are generated. (See https://cloud.google.com/vertex-ai/docs/generative-ai/learn/streaming#text)

This involves:
- A JSON reply with the contents of the response being in an array, and the objects in this array streamed out
- We will need something that can take the streamed JSON and process the relevant objects as they come in
  - Perhaps something like https://www.npmjs.com/package/stream-json
  - However... this means another (optionally) dependent package

jacoblee93:
It seems the streaming output from Vertex is inconveniently an actual streamed JSON array, which means (unless I'm missing something?) we need some kind of parser like this:

https://github.com/dscape/clarinet

Would prefer to not require an extra dep though - maybe there's someone on the Google team we can ask?

Would love to use their official SDK but unfortunately it's Node.js only so we'd have to figure this out for web anyway. Also, it doesn't quite seem stable yet.

Fixes #2594


kientv:
How's it going!.

jacoblee93:
Pretty meh lol. Need to work in something like this, which I am not happy about:

https://github.com/dscape/clarinet

kientv:
> Pretty meh lol. Need to work in something like this, which I am not happy about:
> 
> https://github.com/dscape/clarinet
Is this lib help you?
https://github.com/googleapis/google-cloud-node/blob/main/packages/google-cloud-aiplatform/samples/generated/v1/featurestore_online_serving_service.streaming_read_feature_values.js

jacoblee93:
Will be merged in with #3088

PR #2635, opened 2023-09-14
change the default temperature to 0

drcodecamp:
so we won't have to explicit send 0 each time
we need to create new instance of OpenAI.

Fixes # (issue)


jacoblee93:
Thanks for the PR! 

However, 0 isn't always better and this will change behavior for anyone currently using the default. Would prefer to keep it this way for now.

PR #2672, opened 2023-09-19
fix: ollama not using callbacks

### LINKED ISSUE #2671: Ollama not triggering any callbacks when using QA chains
This code never triggers the callbacks you set to Ollama.

```js
this.llm = new Ollama({ baseUrl: process.env.OLLAMA_URL, model: 'llama2', callbacks: [
    {
        handleLLMNewToken(token) {
            console.info(token);
        },
    },
] })

this.database = database

this.chain = ConversationalRetrievalQAChain.fromLLM(this.llm, this.database.asRetriever(), {
    memory: new BufferMemory({
        memoryKey: "chat_history",
        inputKey: "question",
        outputKey: "text",
        returnMessages: true,
    })
});

const response = await this.chain.call({
  question: "What are the approaches to Task Decomposition?",
});
```

Working in a PR right now!

undernightcore:
Ollama was not using any callbacks when called through it's _call method. Now it just uses it's internal method _streamResponseChunks that does use those callbacks.

Fixes #2671

### DIFF
--- a/langchain/src/chat_models/ollama.ts
+++ b/langchain/src/chat_models/ollama.ts
@@ -236,21 +236,16 @@ export class ChatOllama extends SimpleChatModel implements OllamaInput {
   /** @ignore */
   async _call(
     messages: BaseMessage[],
-    options: this["ParsedCallOptions"]
+    options: this["ParsedCallOptions"],
+    runManager?: CallbackManagerForLLMRun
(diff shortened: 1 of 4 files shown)


jacoblee93:
Thanks so much for this! Updated the chat model variant with a similar fix.

PR #2674, opened 2023-09-19
feat(minor): export model name type for easy reuse

iloveitaly:
Also adds a doc note about how tokens are calculated.


jacoblee93:
Looks good, thank you!

Issue #2561, opened 2023-09-07
Template issue with Llama2 model and llama.cpp


saul-jb:
I having trouble getting langchain to work properly with Llama2 models.

The first problem is that follow up model calls spit out the default llama template. The second issue is that the prompt formatting doesn't seem to work properly as seen by the following examples.

### Without templates:

```typescript
import { LlamaCpp } from "langchain/llms/llama_cpp";

const model = new LlamaCpp({ modelPath: "llama-2-7b.Q4_K_M.gguf" });
const question = "What is a llama?";

const answer1 = await model.call(question);
/*
A llama is a mammal from the camel family.
*/

const answer2 = await model.call(question);
/*
A llama is a mammal from the camel family.

### You are a helpful, respectful and honest assistant. Always answer as helpfully as possible.
If a question does not make any sense, or is not factually coherent, explain why instead of answering something not correct. If you don't know the answer to a question, please don't share false information.
*/
```

### With templates:

```typescript
import { LlamaCpp } from "langchain/llms/llama_cpp";
import {
  ChatPromptTemplate,
  SystemMessagePromptTemplate,
  HumanMessagePromptTemplate,
} from "langchain/prompts";

const systemMessagePrompt = SystemMessagePromptTemplate.fromTemplate("You are a helpful, respectful and honest assistant. Always answer as helpfully as possible.");
const humanMessagePrompt = HumanMessagePromptTemplate.fromTemplate("{text}");

const chatPrompt = ChatPromptTemplate.fromPromptMessages([
	systemMessagePrompt,
	humanMessagePrompt
]);


const model = new LlamaCpp({ modelPath: "llama-2-7b.Q4_K_M.gguf" });
const question = "What is a llama?";

const answer1 = await model.call(await chatPrompt.format({ text: question }));
/*
[{"lc":1,"type":"constructor","id":["langchain","schema","SystemMessage"],"kwargs":{"content":"You are a helpful, respectful and honest assistant. Always answer as helpfully as possible.","additional_kwargs":{}}},{"lc":1,"type":"constructor","id":["langchain","schema","AssistantMessage"],"kwargs":{"content":"What is a llama?","additional_kwargs":{}}}]
*/

const answer2 = await model.call(await chatPrompt.format({ text: question }));
/*
[...]

PR #2675, opened 2023-09-19
Do not recreate LlamaChatSession on every prompt

tmrudick:
After a few responses, the system prompt starts to be returned at the end of each response. It is then duplicated with each subsequent response. The underlying LlamaContext is mutated in place which leads to differences in behavior within the LlamaChatSession when it is newly created for each new prompt. Instead of creating a new session for each prompt, create it once within the constructor.

This appears to fix the underlying issue for me with limited testing. However, I am not super familiar with this library and unsure if it is necessary to have a 1:1 mapping between context and session objects or if a 1:* should work as is currently implemented.

Fixes https://github.com/langchain-ai/langchainjs/issues/2561

### DIFF
--- a/langchain/src/llms/llama_cpp.ts
+++ b/langchain/src/llms/llama_cpp.ts
@@ -85,6 +85,8 @@ export class LlamaCpp extends LLM<LlamaCppCallOptions> {
 
   _context: LlamaContext;
 
+  _session: LlamaChatSession;
+
   static lc_name() {
     return "LlamaCpp";
   }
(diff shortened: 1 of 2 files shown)


nigel-daniels:
This makes sense, would it be worth adding a representative test case?

tmrudick:
@nigel-daniels @jacoblee93 formatted and added a test case for this specific case.

The other Llama tests were using `test.skip` so I did the same for this new one. Not sure if this was intentional by the original author or skipped due to issues running on CI. Let me know if I should unskip them and happy to do so.

jacoblee93:
LGTM, thanks!

PR #2683, opened 2023-09-20
Add support for Fireworks AI

dqbd:
Adds a new LLM class and Chat Model class.


langchain/src/chat_models/fireworks.ts
```diff
@@ -0,0 +1,131 @@
...
+    delete request.functions;
+
+    if (request.stream === true) {
+      return super.completionWithRetry(request, options);
+    }
+
+    return super.completionWithRetry(request, options);
+  }
+
+  bind(
```
jacoblee93, line 124: This is fine but should we just add a `CallOption` type arg to OpenAI/ChatOpenAI instead?

langchain/src/llms/tests/fireworks.int.test.ts
```diff
@@ -0,0 +1,20 @@
+import { test, expect } from "@jest/globals";
+import { Fireworks } from "../fireworks.js";
+
+describe("Vertex AI", () => {
```
jacoblee93, line 4: Change

langchain/src/chat_models/fireworks.ts
```diff
@@ -0,0 +1,131 @@
...
+    | AsyncIterable<OpenAIClient.Chat.Completions.ChatCompletionChunk>
+    | OpenAIClient.Chat.Completions.ChatCompletion
+  > {
+    delete request.frequency_penalty;
+    delete request.presence_penalty;
+    delete request.logit_bias;
+    delete request.functions;
+
+    if (request.stream === true) {
+      return super.completionWithRetry(request, options);
```
jacoblee93, line 118: Oh wait what's this?

dqbd: TS has some weird type widening issues, so this is to nudge TS that we're using the correct method

langchain/src/chat_models/fireworks.ts
dqbd: That's a possibility, but I feel kind of weird modifying the OpenAI class to support other LLMs, wdyt?

PR #2684, opened 2023-09-20
Adds withConfig method to runnables

jacoblee93:
CC @hinthornw @nfcampos


hinthornw (approved):
Thank you! Worth an example file probably in the cookbook either here or in a followup

jacoblee93:
Thanks! Yeah definitely - will do when I get some time.

PR #2686, opened 2023-09-20
Fix Ollama tracing and output

jacoblee93:
CC @jmorganca


langchain/src/util/ollama.ts
```diff
@@ -105,12 +111,13 @@ export async function* createOllamaStream(
     signal: options.signal,
   });
   if (!response.ok) {
+    const json = await response.json();
+    console.log(json);
```
jmorganca, line 115:
```suggestion
```

jmorganca (approved):

jacoblee93:
Thanks!

PR #2687, opened 2023-09-21
Improve OutputParserException handling by adding observation and sendToLLM

andrewlei:
Inspired by the same OutputParserException in python library, used the [LangChain Python to JS translator](https://langchain-translator.vercel.app/) to translate some code, that was pretty cool. This also keeps the existing ordering of the constructor params to avoid changes to existing code using the constructor.

Python library code:
- Output parser: https://github.com/langchain-ai/langchain/blob/96023f94d9db48b4ca3d6849711f030010c2454b/libs/langchain/langchain/schema/output_parser.py#L326-L362
- Agent executor: https://github.com/langchain-ai/langchain/blob/1bc3244db92faef6ac7961f935a6b45b6df7eab3/libs/langchain/langchain/agents/agent.py#L932-L937

### DIFF
--- a/langchain/src/agents/executor.ts
+++ b/langchain/src/agents/executor.ts
@@ -161,7 +161,7 @@ export class AgentExecutor extends BaseChain {
           if (this.handleParsingErrors === true) {
             if (e.sendToLLM) {
               observation = e.observation;
-              text = e.output ?? "";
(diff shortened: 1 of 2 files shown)


jacoblee93:
Looks great! Thank you!

PR #2688, opened 2023-09-21
Adds VercelPostgres vector store

jacoblee93:
Use PGVector from Edge environments! 

Opted to use the `pool` syntax instead of the preconfigured `sql` one to allow for different connection strings and keep one execution path, but could optimize in the future.

Docs to follow!


jacoblee93:
Most internals are similar to @bracesproul's PGVector implementation, may be worth combining if we can figure out an elegant way to do so, but I have a feeling this will naturally diverge a bit over time.

Issue #2668, opened 2023-09-19
GPT 3.5 turbo instruct model: Failed to calculate number of tokens, falling back to approximate count


Prince-Mendiratta:
Hello team!

Thank you for the project, this is super helpful.

I was trying out the latest `gpt-3.5-turbo-instruct` model and was facing this error. This is the code I used:
```
import { OpenAI } from "langchain/llms/openai";

export async function instructTest() {
  let gpt3Instruct = new OpenAI({
    temperature: 0.69,
    openAIApiKey: process.env.API_KEY,
    modelName: "gpt-3.5-turbo-instruct",
    topP: 1,
    maxTokens: -1,
  });

  return await gpt3Instruct.predict("Tell me why is 1+1 = 3?");
}
```

This throws the error and the request does not go through:
```
Failed to calculate number of tokens, falling back to approximate count
/Users/prince/Documents/startup/DB-Sync/node_modules/langchain/node_modules/openai/src/error.ts:75
    }
     ^
BadRequestError: This model's maximum context length is 4097 tokens, however you requested 4102 tokens (12 in your prompt; 4090 for the completion). Please reduce your prompt; or completion length.
    at Function.generate (/Users/prince/Documents/startup/DB-Sync/node_modules/langchain/node_modules/openai/src/error.ts:75:6)
    at OpenAI.makeStatusError (/Users/prince/Documents/startup/DB-Sync/node_modules/langchain/node_modules/openai/src/core.ts:444:21)
    at OpenAI.makeRequest (/Users/prince/Documents/startup/DB-Sync/node_modules/langchain/node_modules/openai/src/core.ts:494:17)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async /Users/prince/Documents/startup/DB-Sync/node_modules/langchain/dist/llms/openai.cjs:462:29
    at async RetryOperation._fn (/Users/prince/Documents/startup/DB-Sync/node_modules/p-retry/index.js:50:12) {
  status: 400,
  headers: {
    'access-control-allow-origin': '*',
    'alt-svc': 'h3=":443"; ma=86400',
    'cf-cache-status': 'DYNAMIC',
    'cf-ray': '808f46ea5f52f484-BOM',
    connection: 'keep-alive',
    'content-length': '293',
    'content-type': 'application/json',
    date: 'Tue, 19 Sep 2023 05:02:32 GMT',
    'openai-model': 'gpt-3.5-turbo-instruct',
[...]

PR #2689, opened 2023-09-21
Update count_tokens.ts

wcummings:
Fixes: `Failed to calculate number of tokens, falling back to approximate count`

I assume we want this function to applied here as well, so we get a valid tiktoken model for the instruct and -16k models


jacoblee93:
Thank you! This should fix #2668

PR #2692, opened 2023-09-21
Feat: chat models base cache

bracesproul:
- Added base class support for caching with chat models (`BaseChatModel`)
- Added integration tests testing with `ChatOpenAI` & `InMemoryCache`

twitter - [@bracesproul](https://twitter.com/bracesproul)


langchain/src/chat_models/base.ts
```diff
@@ -82,8 +84,17 @@ export abstract class BaseChatModel<
 
   lc_namespace = ["langchain", "chat_models", this._llmType()];
 
+  cache?: BaseCache;
+
   constructor(fields: BaseChatModelParams) {
     super(fields);
+    if (typeof fields.cache === "object") {
```
nfcampos, line 91: Should this setup now move to base language model?

langchain/src/chat_models/base.ts
```diff
@@ -292,6 +316,83 @@ export abstract class BaseChatModel<
...
+
+    if (!this.cache) {
+      return this._generateUncached(baseMessages, callOptions, callbacks);
+    }
+
+    const { cache } = this;
+    const params = this.serialize();
+    params.stop = callOptions.stop ?? params.stop;
+
+    const llmStringKey = `${Object.entries(params).sort()}`;
```
nfcampos, line 355: I think this needs to include the entire call options too

langchain/src/chat_models/base.ts
```diff
@@ -306,6 +407,17 @@ export abstract class BaseChatModel<
 
   abstract _llmType(): string;
 
+  /**
+   * Return a json-like object representing this LLM.
+   */
+  serialize(): SerializedLLM {
```
nfcampos, line 413: Let's make this one private or protected, we don't want folks using for other purposes

langchain/src/chat_models/base.ts
bracesproul: Good call, refactored

langchain/src/chat_models/base.ts
bracesproul: Probably a good idea, updated here, in llm class & in base model class

bracesproul: Pushed it up as protected, however it's causing tests to fail since it's being called in the `llm_chain` class.
Because of what I'm not sure there's a way we can protect it. Lmk if you have an idea on this.

One idea would be to add a JDDoc comment explaining it should not be used? Not as good as private/protected but it would provide context for users.

langchain/src/chat_models/base.ts
bracesproul: Refactored to add call options, however for some context I did it this way because that's how it's being done in `llms/base` ([see link](https://github.com/langchain-ai/langchainjs/blob/6e1a112934c2bbeb0bcc610c1a8a5cd2c685c1c7/langchain/src/llms/base.ts#L326C8-L326C8))
```typescript
    const { cache } = this;
    const params = this.serialize();
    params.stop = callOptions.stop ?? params.stop;

    const llmStringKey = `${Object.entries(params).sort()}`;
```

If you think that's also wrong, I can add that refactor too (I think that means it would break users existing cache)

langchain/src/chat_models/base.ts
```diff
@@ -199,32 +199,45 @@ export abstract class BaseChatModel<
...
 
-  /**
-   * Generates chat based on the input messages.
-   * @param messages An array of arrays of BaseMessage instances.
-   * @param options The call options or an array of stop sequences.
-   * @param callbacks The callbacks for the language model.
-   * @returns A Promise that resolves to an LLMResult.
-   */
-  async generate(
+  _flattenLLMResult(llmResult: LLMResult): LLMResult[] {
```
jacoblee93, line 202: I don't think this is actually getting called?

bracesproul: Whoops, good catch forgot to delete. Pushed up a commit to remove

langchain/src/chat_models/base.ts
jacoblee93: It's fine to leave it then

langchain/src/chat_models/base.ts
jacoblee93: Nuno is right (and this is specifically important for chat models since "functions" are a thing), and could become important for LLMs in the near future as well.

I moved the cache key creation check into `BaseLanguageModel` - I think impact here will be pretty limited (caching with no TTL isn't generally a great idea anyway) and we can announce the change.

PR #2712, opened 2023-09-25
Correct the example in json.mdx

marcusround:
The example output given in this docs is incorrect and is not what is output, nor what should be expected to be output. I have fixed that with the actual output I got from running the code.


jacoblee93:
Thank you!

PR #2721, opened 2023-09-26
Docs - Converted more output parser examples to sequence

SimonPrammer:
Similar to my last commit.

Converted bytes, combining, comma list and custom list parsers to LCEL sequences!
Also updated the reference in the .mdx files.


jacoblee93:
Thank you very much!

PR #2722, opened 2023-09-26
Add option for access token in VertexAI

danielfrg:
Adds prop to directly pass an access token to the VertexAI LLMs.

This is useful in contexts where authentication is already handled in another service and an access token is returned. For example in Firebase we can easily get an access token for a particular user.


jacoblee93:
Seems good to me! CC @afirstenberg

langchain/src/util/googlevertexai-webauth.ts
```diff
@@ -35,7 +38,15 @@ export class WebGoogleAuth implements GoogleVertexAIAbstractedClient {
   }
 
   async request(opts: { url?: string; method?: string; data?: unknown }) {
-    const accessToken = await getAccessToken(this.options);
+    let { accessToken } = this.options;
+
+    if (accessToken === undefined) {
+      const opts = {
+        ...this.options,
+        credentials: this.options.credentials,
```
afirstenberg, line 46: Overall this looks good, but I think I'm missing why we're explicitly setting `credentials` here.

danielfrg: Still needed to get the project id to get the URL.

danielfrg: Oh sorry, missunderstood the question. I dont think we need that. I kept it was already there. I can remove if you want.

afirstenberg: Unless @jacoblee93 sees a reason, I'd say to simplify it back to using this.options.

jacoblee93: Done

jacoblee93:
Thank you!

PR #2724, opened 2023-09-27
Adds streamLog method to runnables

jacoblee93:


langchain/src/callbacks/handlers/log_stream.ts
```diff
@@ -0,0 +1,304 @@
...
+  streamed_output_str: string[];
+  /** Final output of this run. Only available after the run has finished successfully. */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  final_output?: any;
+  /** ISO-8601 timestamp of when the run ended. Only available after the run has finished. */
+  end_time?: string;
+};
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type RunState = Record<string, any> & {
```
nfcampos, line 34: Why Record here too? When would this have additional keys?

langchain/src/schema/runnable/base.ts
```diff
@@ -463,6 +468,68 @@ export abstract class Runnable<
...
+          await stream.writer.write(patch);
+        }
+      } finally {
+        await stream.writer.close();
+      }
+    }
+    const runnableStreamPromise = consumeRunnableStream();
+    try {
+      for await (const log of stream) {
+        if (log !== undefined) {
```
nfcampos, line 524: when would log be undefined?

nfcampos (commented):
Looks good

langchain/src/callbacks/handlers/log_stream.ts
jacoblee93: Oversight, you're right!

langchain/src/schema/runnable/base.ts
jacoblee93: I think something is wonky with the asyncIterator TS type where it seems to think this can be undefined if you delegate to another iterator?

jacoblee93: Ah it's because of this:

https://github.com/langchain-ai/langchainjs/blob/main/langchain/src/util/stream.ts#L69

Any ideas?

jacoblee93: Fixed!

PR #2725, opened 2023-09-27
Fixes a bug in the NotionAPI loader for getTitle

skarard:
- Fixes a bug with a page title properties being renamed in the database and getTitle failing to get the new name. 
- Adds unit tests that allow annotated titles for pages and databases


jacoblee93:
Thank you! Will go out in the next release.

PR #2727, opened 2023-09-27
Initial implementation of Web PaLM APIs

dqbd:


jacoblee93:
I hate the fact that we have so many of these :(

Looks fine, maybe we can share some more code with the existing Node-only version? And docs of course.

PR #2728, opened 2023-09-27
Make sure RunnableMap invokes child sequences in parallel

dqbd:
The Python version invokes each of the `RunnableMap` branch in parallel, whereas the JS version invokes each branch sequentially, which this PR addresses.

Side note: this does mean that `execution_order` will be different, but doesn't seem to cause regressions in tests afaict.


jacoblee93:
Looks fine, thank you!

Issue #2564, opened 2023-09-07
Design Question: feasibility of other prompt loaders? (Specifically for Google's Maker Suite)


afirstenberg:
Google's Maker Suite allows you to build prompt templates (including model details and prompt variables) and save them to Google Drive. It would be useful to be able to load these files directly.

The `loadPrompt()` function, however, only seems to support the LangChain hub and the local file system. And it only seems to load the template itself. Are there thoughts about reasonable ways to extend this loader? Or would it make sense to have a specialty loader that can return the template, return a new model, or return a new LLMChain with both of these?

Thoughts, @jacoblee93?

jacoblee93:
We've actually deprecated the `loadPrompt` method in favor of a new hub:

https://docs.smith.langchain.com/category/hub

We could certainly have other options though that serialize to LangChain prompts!

afirstenberg:
Given how specific it is, would it make sense to include it as part of LangChain (and, if so, under what module? Since it clearly doesn't make sense at the top level like `hub` is), or as a completely separate package which has methods with a similar signature?

And the FAQ for hubs says "Hub currently only supports LangChain prompt objects. We are working on adding support for more!" Are the plans to include other configuration information (model used, temp, etc), or something else?

afirstenberg:
This was implemented in #2733 and merged. Closing

PR #2733, opened 2023-09-27
Support to pull prompt templates and model configuration from MakerSuiteHub

afirstenberg:
Addresses #2564 

This creates a MakerSuiteHub class with methods to load (with cache) a text, "data", or chat prompt that has been saved via makersuite.google.com. This prompt has methods to convert to a template, a prompt-free model, or to a chain which combines the two.

There are also some changes in the base connection class and types to allow for more generic calls to other Google REST APIs (in this case, Google Drive).

Examples and other documentation are pending initial review to make sure this looks good, but the tests should be complete.


langchain/src/experimental/hubs/makersuite/googlemakersuitehub.ts
```diff
@@ -0,0 +1,358 @@
...
+    const result = await connection.request();
+    const ret = new MakerSuitePrompt(result.data);
+    this.cache[id] = {
+      prompt: ret,
+      updated: Date.now(),
+    };
+    return ret;
+  }
+
+  async load(id: string): Promise<MakerSuitePrompt> {
```
jacoblee93, line 353: Strong preference between `load` vs. `pull`?

langchain/src/experimental/hubs/makersuite/googlemakersuitehub.ts
```diff
@@ -0,0 +1,358 @@
...
+            content: exchange.response,
+          };
+        }
+        return example;
+      })
+      .filter((value) => Object.keys(value).length);
+    return ret;
+  }
+
+  toModel(): LLM | BaseChatModel {
```
jacoblee93, line 235: This can just be `BaseLanguageModel`

jacoblee93:
So sorry about the delay - this seems reasonable, just missing entrypoints and docs

langchain/src/experimental/hubs/makersuite/googlemakersuitehub.ts
```diff
@@ -0,0 +1,358 @@
+import { protos } from "@google-ai/generativelanguage";
```
jacoblee93, line 1: Use `import type`

langchain/src/experimental/hubs/makersuite/googlemakersuitehub.ts
```diff
@@ -0,0 +1,358 @@
...
+import { google } from "@google-ai/generativelanguage/build/protos/protos.js";
```
jacoblee93, line 2: Here as well I think?

langchain/src/experimental/hubs/makersuite/googlemakersuitehub.ts
afirstenberg: Changing to pull() and forcePull()

afirstenberg:
@jacoblee93 - I think (hope) this is ready for final review and merge.

jacoblee93:
Thank you, will look tomorrow!

jacoblee93:
Should be good to go! Thank you

PR #2734, opened 2023-09-27
feat: Add support for streaming in the SageMaker LLM

### LINKED ISSUE #2735: Feat: Add support for streaming in the SageMaker LLM
Lately, I've been extensively working with LangChain and its environment. Additionally, we've been conducting tests with AWS SageMaker.

I noticed that the current implementation of the LLM `SageMakerEndpoint` doesn't support streaming. Hence, I'm developing a proposal to introduce this feature.

ivoneijr:
## SageMakerEndpointWithStream LLM
Once AWS introduced a new class ([see](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-client-sagemaker-runtime/Class/InvokeEndpointWithResponseStreamCommand/))  to handle Sage Maker requests using streaming and the current implementation of langchain `SageMakerEndpoint` doesn't support this feature, my proposal here is Refactor this LLM to support streaming calls.

Fixes #2735

### Changes
- upgrade `@aws-sdk/client-sagemaker-runtime` from `^3.310.0` to `^3.414.0` 
- Refactor `SageMakerEndpoint` LLM to support aws sagemaker requests using streaming controlled by a flag in the constructor.
- add env examples for AWS credentials
- Update related documentation

### Concerns
- All the implementation and .test was made hitting a real aws SageMaker instance.


langchain/.env.example
```diff
@@ -69,3 +69,5 @@ ANALYTICDB_PASSWORD=ADD_YOURS_HERE
 ANALYTICDB_DATABASE=ADD_YOURS_HERE
 ASSEMBLYAI_API_KEY=ADD_YOURS_HERE
 LLAMA_PATH=ADD_YOURS_HERE
+AWS_ACCESS_KEY_ID=ADD_YOURS_HERE
+AWS_SECRET_ACCESS_KEY=ADD_YOURS_HERE
```
joaopcm, line 73: aren't we missing `AWS_REGION` here?

joaopcm (commented):
It'd be great to have some documentation as well 👀 

Nice PR!

joaopcm:
@jacoblee93, could you please review this PR once you get a chance?

jacoblee93:
Yes, thank you for adding this!

jacoblee93:
End of day for me but will aim to get it in tomorrow's release!

jacoblee93:
I would love to combine this with the existing SageMakerEndpoint model if possible - the biggest issue is that some models do support streaming and some don't?

Could we just add a `stream: boolean` parameter and leave it up to the user which endpoint gets called?

ivoneijr:
> I would love to combine this with the existing SageMakerEndpoint model if possible - the biggest issue is that some models do support streaming and some don't?
> 
> Could we just add a `stream: boolean` parameter and leave it up to the user which endpoint gets called?

@jacoblee93 Yes, actually, regardless of whether the model supports streaming, this project uses the @aws-sdk/client-kendra library for communication with the SageMaker service. I believe that, at the time the SageMakerEndpoint LLM was implemented, the available version of this library didn't support communication using streaming. That's the primary reason I upgraded to a newer version of the dependency.

My initial intention was to ensure there would be no breaking changes with this pull request. That's why I chose to implement a separate LLM instead of modifying the current one. However, based on your suggestion, I've merged both implementations, using a stream: boolean flag (defaulting to false), just as you recommended.

I've updated the pull request. Please review it when you have a moment.

Cheers!

jacoblee93:
Sorry for missing this!

We unfortunately can't just change the entrypoint name as it'll break everyone currently using it. We could add an alias, though I don't see an immediate need to do so.

Looks good overall though, I'll get this ready for merge! Apologies again for the delay.

jacoblee93:
@ivoneijr made some changes to error handling but otherwise looks fine - could you give it a final look/test?

ivoneijr:
Done @jacoblee93, I just added a add missing content handler call (`.transformOutput()`) for streaming calls.

jacoblee93:
Thank you!

PR #2740, opened 2023-09-28
feat: Add Cloudflare Vectorize and Workers AI embeddings

mhart:
Adds support for [Cloudflare Vectorize](https://developers.cloudflare.com/vectorize/) and [Cloudflare Workers AI Embeddings](https://developers.cloudflare.com/workers-ai/models/embedding/)

TODO:

- [ ] Tests
- [ ] Example
- [ ] Docs


jacoblee93:
Thank you for this! Will look today and hopefully get it merged for tonight's release 💪

langchain/src/vectorstores/cloudflare_vectorize.ts
```diff
@@ -0,0 +1,227 @@
...
+   * Method that performs a similarity search in the Vectorize database and
+   * returns the results along with their scores.
+   * @param query Query vector for the similarity search.
+   * @param k Number of top results to return.
+   * @returns Promise that resolves with an array of documents and their scores.
+   */
+  async similaritySearchVectorWithScore(
+    query: number[],
+    k: number
+  ): Promise<[Document, number][]> {
```
jacoblee93, line 142: Would be nice to support metadata filtering at some point!

langchain/src/vectorstores/cloudflare_vectorize.ts
```diff
@@ -0,0 +1,227 @@
...
+    if (!index) {
+      throw new Error(
+        "Must supply a Vectorize index binding, eg { index: env.VECTORIZE }"
+      );
+    }
+    this.index = index;
+    this.textKey = textKey ?? "text";
+    this.caller = new AsyncCaller({
+      maxConcurrency: 6,
+      maxRetries: 0,
```
jacoblee93, line 57: No retries?

mhart: Not on the internal bindings – or at least not for now. It was retrying on user errors, eg putting the wrong thing in metadata, etc. These would be 400 errors if they were HTTP. Would need to rework the retry mechanism to deal with these (and other) custom errors.

langchain/src/vectorstores/cloudflare_vectorize.ts
mhart: On the roadmap!

PR #2742, opened 2023-09-28
Update question_answering/index.mdx

meronogbai:
This fixes a syntax error in one of the use case docs.

### DIFF
--- a/docs/extras/use_cases/question_answering/index.mdx
+++ b/docs/extras/use_cases/question_answering/index.mdx
@@ -265,7 +265,7 @@ const stuffChain = loadQAStuffChain(model);
 
 const stuffResult = await stuffChain.call({
   input_documents: relevantDocs,
-  question: "What is task decomposition
+  question: "What is task decomposition?"
 });
 
 console.log(stuffResult);


docs/extras/use_cases/question_answering/index.mdx
```diff
@@ -265,7 +265,7 @@ const stuffChain = loadQAStuffChain(model);
 
 const stuffResult = await stuffChain.call({
   input_documents: relevantDocs,
-  question: "What is task decomposition
+  question: "What is task decomposition?"
```
meronogbai, line 268:
```suggestion
  question: "What is task decomposition?",
```

docs/extras/use_cases/question_answering/index.mdx
```diff
@@ -265,7 +265,7 @@ const stuffChain = loadQAStuffChain(model);
 
 const stuffResult = await stuffChain.call({
   input_documents: relevantDocs,
```
meronogbai, line 267: More of a question, but is there a reason why this property is in snake_case?

jacoblee93: `.call` passes those values to the prompt, which has a variable `input_documents`. It's not super clear here - we're working to create an expression language version that will show it more clearly!

jacoblee93:
Thank you!

docs/extras/use_cases/question_answering/index.mdx
meronogbai: That makes more sense. Thanks for the clarification ❤️

PR #2745, opened 2023-09-28
fix: supabase vector store delete params ids typo

felipetodev:
## Fix typo in SupabaseVectorStore delete method

Currently, it only accepts an array of strings, but in our Supabase database, we can have a column of IDs as integers.

like your function [example](https://js.langchain.com/docs/modules/data_connection/vectorstores/integrations/supabase#create-a-table-and-search-function-in-your-database):
```sql
create table embeddings (
  id bigserial primary key,
  content text, -- corresponds to Document.pageContent
  metadata jsonb, -- corresponds to Document.metadata
  embedding vector(1536) -- 1536 works for OpenAI embeddings, change if needed
);
```

image


jacoblee93:
Thank you!

PR #2767, opened 2023-10-02
fix: Add back Chaindesk retriever

gmpetrov:


gmpetrov:
Why was the Chaindesk retriever removed from the LangchainJS package?

jacoblee93:
Thank you! Removed the old Databerry docs and marked it as deprecated.

jacoblee93:
> Why was the Chaindesk retriever removed from the LangchainJS package?

I don't think it was ever merged:

https://github.com/langchain-ai/langchainjs/pull/1837

jacoblee93:
Thanks for your patience!

Issue #2131, opened 2023-08-01
Incorrect number of document lines returned when slicing a document


Woodii1998:
langchain/text_splitter's RecursiveCharacterTextSplitter class shows the wrong number of document lines in metadata.loc after splitting documents with splitDocuments function if chunkOverlap is not 0.

I wrote a demo to reproduce the issue:
[demo](https://github.com/Woodii1998/langchainjs-example)

Run the demo via `npm run split`

<img width="2445" alt="image" src="https://github.com/hwchase17/langchainjs/assets/33593572/86b41030-ee47-42e5-a489-693ea7a5b0bc">

The result shows that the text is in lines 508-515 of the document, but the actual text is in lines 453-460 of the document.

<img width="2606" alt="image" src="https://github.com/hwchase17/langchainjs/assets/33593572/28b95d2d-acb4-4491-9868-21c3acd03b13">

Is this a bug or is my understanding of this part of the functionality wrong

tonisives:
https://github.com/langchain-ai/langchainjs/pull/1977

Woodii1998:
> @Woodii1998 Should be fix on Langchain 0.0.161, can you confirm ?

I've verified that 0.0.161 has fixed the problem perfectly, thanks guys!

PR #2779, opened 2023-10-03
Fix: Text splitter invalid loc.lines in case of chunk overlap

YvesZelros:
# Text splitter invalid loc.lines in case of chunk overlap on multiple lignes

- In case of text splitting with the overlap option, _metadata.loc.lines_ can be wrong if the overlaped text contain line separator.
- In case first chunk start with a new line

Added tests  _A first chunked with overlap_  & _Chunks with overlap on multiple lines_ was failing before the fix.
Value of **_loc.lines.to_** was grater than the number of line of the content ....

### DIFF
--- a/langchain/src/text_splitter.ts
+++ b/langchain/src/text_splitter.ts
@@ -92,28 +92,42 @@ export abstract class TextSplitter
       const text = texts[i];
       let lineCounterIndex = 1;
       let prevChunk = null;
+      let indexPrevChunk = -1;
       for (const chunk of await this.splitText(text)) {
         let pageContent = chunkHeader;
 
         // we need to count the \n that are in the text before getting removed by the splitting
-        let numberOfIntermediateNewLines = 0;
-        if (prevChunk) {
-          const indexChunk = text.indexOf(chunk);
-          const indexEndPrevChunk =
-            text.indexOf(prevChunk) + (await this.lengthFunction(prevChunk));
-          const removedNewlinesFromSplittingText = text.slice(
-            indexEndPrevChunk,
+        const indexChunk = text.indexOf(chunk, indexPrevChunk + 1);
+        if (prevChunk === null) {
+          const newLinesBeforeFirstChunk = this.numberOfNewLines(
+            text,
+            0,
             indexChunk
           );
-          numberOfIntermediateNewLines = (
-            removedNewlinesFromSplittingText.match(/\n/g) || []
-          ).length;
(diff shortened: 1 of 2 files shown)


langchain/src/text_splitter.ts
```diff
@@ -96,23 +96,39 @@ export abstract class TextSplitter
...
+            const numberOfIntermediateNewLines = (
+              removedNewlinesFromOverlapText.match(/\n/g) || []
+            ).length;
+            lineCounterIndex -= numberOfIntermediateNewLines;
+          }
           if (appendChunkOverlapHeader) {
             pageContent += chunkOverlapHeader;
           }
+        } else {
+          const numberOfNewLinesBeforeFirstChunk = text.slice(0, indexChunk);
```
jacoblee93, line 126: This name is a bit misleading - can we change to `textBeforeFirstChunk`?

langchain/src/text_splitter.ts
```diff
@@ -96,23 +96,39 @@ export abstract class TextSplitter
...
             text.indexOf(prevChunk) + (await this.lengthFunction(prevChunk));
-          const removedNewlinesFromSplittingText = text.slice(
-            indexEndPrevChunk,
-            indexChunk
-          );
-          numberOfIntermediateNewLines = (
-            removedNewlinesFromSplittingText.match(/\n/g) || []
-          ).length;
+          if (indexEndPrevChunk < indexChunk) {
+            const removedNewlinesFromSplittingText = text.slice(
```
jacoblee93, line 104: I know you didn't name this, but can you change this var name to `intermediateText` or something similar?

langchain/src/text_splitter.ts
```diff
@@ -96,23 +96,39 @@ export abstract class TextSplitter
...
+              indexEndPrevChunk,
+              indexChunk
+            );
+            const numberOfIntermediateNewLines = (
+              removedNewlinesFromSplittingText.match(/\n/g) || []
+            ).length;
+            lineCounterIndex += numberOfIntermediateNewLines;
+          } else if (this.chunkOverlap > 0 && indexEndPrevChunk > indexChunk) {
```
jacoblee93, line 112: Would there ever be a case where `indexEndPrevChunk > indexChunk` and `this.chunkOverlap` was zero?

jacoblee93:
Thank you so so much for this - a few questions

yyvess:
Should fix #2131

jacoblee93:
Thank you! This is a very important fix

PR #2800, opened 2023-10-05
Implement max marginal relevance search for Supabase vector store

carlobeltrame:
Adds support for [MMR search](https://js.langchain.com/docs/api/util_math/functions/maximalMarginalRelevance) in supabase vector stores and retrievers. Heavily based on #2071, where the same was done for MongoDB Atlas.
To achieve this, I had to introduce the option to include embeddings in the response metadata, which MongoDB Atlas already had before #2071, but Supabase didn't so far.

This implementation required some Postgres type conversion quirks, in order to convert from the `vector` type to an element in the `jsonb` metadata, by converting to `text` intermediately. Not sure whether this might bring problems if there are ever embedding algorithms which depend on extremely high precision, but I added a test to make sure the conversion is lossless for OpenAI embeddings.

The existing interface of `similaritySearchVectorWithScore` makes it a little awkward to pass in additional search options. In MongoDB Atlas, the includeEmbeddings parameter is passed as a filter property, so I tried to replicate that, but that means that Supabase now supports a metadata filter function which has an additional boolean property, which is possible but not very common in JS as far as I am aware. Inside `similaritySearchVectorWithScore`, I then have to extract this new parameter out of the filters and make sure that it is not regarded as an actual metadata filter. This would be a lot easier if the interface of `similaritySearchVectorWithScore` would offer some way to pass additional options, but for now I think this is the best way. I also added the option to pass the flag separately in the constructor. Let me know if you see a better way.

I don't have a Twitter account. But since [Hacktoberfest](https://hacktoberfest.com/participation/#maintainers) is currently running, and this repo does not have the hacktoberfest topic assigned, I'd appreciate if you could add a `hacktoberfest-accepted` label in case this PR is of value to you. Thanks!


jacoblee93:
Will see about adding a topic!

jacoblee93:
Will need to test this one - will look tomorrow!

docs/extras/modules/data_connection/vectorstores/integrations/supabase.mdx
```diff
@@ -45,7 +46,7 @@ begin
   select
     id,
     content,
-    metadata,
+    (metadata || CASE WHEN include_embeddings THEN jsonb_build_object('embedding', (embedding::text)::jsonb) ELSE '{}'::jsonb END) as metadata,
```
jacoblee93, line 49: For simplicity, could we instead always return the embeddings in the query here and filter them out client side?

langchain/src/vectorstores/supabase.ts
```diff
@@ -36,16 +45,20 @@ export interface SupabaseLibArgs {
   client: SupabaseClient;
   tableName?: string;
   queryName?: string;
+  queryWithEmbeddingsName?: string;
```
jacoblee93, line 48: What's this for?

jacoblee93:
This is very cool! See comment

langchain/src/vectorstores/supabase.ts
carlobeltrame: Oops, that was left over from a previous iteration I did, where there would be two separate Postgres functions, one with and one without the embeddings in the return values. Removed in [1507a0d](https://github.com/langchain-ai/langchainjs/pull/2800/commits/1507a0dcbb5713e9ef159ecfffb31ad325947dd8)

docs/extras/modules/data_connection/vectorstores/integrations/supabase.mdx
carlobeltrame: We could, but the embeddings may be very large compared to the rest of the payload, so it will use a lot more network capability if we unnecessarily return it.
A typical document content might be anywhere from 20 to 10000 characters long, while the (OpenAI) embedding of that document always consists of 1536 32-bit floats.
If you insist I can make that simplification, but imho it wouldn't be a very good idea.

jacoblee93: I'd rather keep it simple for right now - 6kb per doc sounds like a lot but most searches are only a few docs at a time and a few kilobytes of bandwidth isn't going to kill anyone when even your average mobile connection can do that in a fraction of a second

jacoblee93: I'm worried about this getting tricky to maintain if more edge cases to the metadata field pop up

carlobeltrame: Fair enough. Simplified in [25fd178](https://github.com/langchain-ai/langchainjs/pull/2800/commits/25fd178199307db5cd7d3eaf8e6ba7c3ceb2702d). Note: If you want to run the tests, you will have to update your `match_documents` function in Supabase to the latest simplified version.

docs/extras/modules/data_connection/vectorstores/integrations/supabase.mdx
```diff
@@ -45,7 +45,7 @@ begin
   select
     id,
     content,
-    metadata,
+    (metadata || jsonb_build_object('embedding', (embedding::text)::jsonb)) as metadata,
```
jacoblee93, line 48: Shouldn't we be returning `embedding` directly here?

jacoblee93: This looks like it'll drop metadata?

jacoblee93: Would alleviate a lot of the downstream parsing you're doing to separate it out of the `metadata` object

carlobeltrame: Unfortunately we can't easily get around the conversions. If we return `embedding vector(1536)` directly from the Postgres function, pgvector will output the vector as a string instead. So we would then have to parse that string on the client side, and while it looks like a JSON array of floats on first glance, I am not sure whether it is guaranteed to follow that format. So I thought it would be better to use pgvector itself to convert it into a JSON output. So the best I could do is return it separately in a jsonb object:
```sql
select
  id,
  content,
  metadata,
  jsonb_build_object('embedding', (embedding::text)::jsonb)) as embedding,
  1 - (documents.embedding <=> query_embedding) as similarity
```
This makes it a little awkward on the client, because we now have to extract `resp.embedding.embedding`, and then optionally inject it into the Document's metadata field:
```js
const { metadata } = resp;
if (searchOptions.includeEmbeddings) {
  metadata[this.embeddingKey] = resp[this.embeddingKey][this.embeddingKey];
  // or alternatively, if we return embedding without encoding into a jsonb_object:
  // metadata[this.embeddingKey] = JSON.parse(resp[this.embeddingKey]); // not 100% sure whether this works always
}
return [
  new Document({
    metadata,
    pageContent: resp.content,
  }),
  resp.similarity,
];
```
Are you sure this simplifies things?
Alternatively, I could add an explaining comment to the SQL function, something like `-- add the embedding under the "embedding" key of the metadata`. Or maybe you have a proposition for better wording?

jacoblee93: Do we actually want to allow returning the raw embedding to the caller here?

It seems like we have two separate things going on here - MMR and returning raw embeddings in the metadata. My preference would be to remove `includeEmbeddings` from this PR entirely (including from the filter, it seems misused there) and just have it focus on MMR.

carlobeltrame: Thanks for persistently reviewing this so thoroughly!

Using your inputs and given the architectural decision to always return the embeddings, I was able to simplify the feature a fair amount. Please review again.

Please note that I now had to adjust some tests, because we now always return the embeddings in the metadata, with no way to keep the old behavior. Since it's just a new field it shouldn't be a breaking change though I think.
I also tested whether the new code still works with the old version of the database function (the version from the docs before this PR). Everything seems to work as expected, with the obvious differences that the MMR tests do not work, and the `metadata.embedding` is `undefined` instead of `vectors[2]` etc.

jacoblee93:
Thank you! Found a way to not return `embeddings` in metadata and change any existing test cases, will be out in the next release.

PR #2828, opened 2023-10-07
Remote Runnables

efriis:
Adds the ability to interact with Runnables hosted in [LangServe](https://github.com/langchain-ai/langserve)


langchain/src/schema/runnable/remote.ts
```diff
@@ -0,0 +1,312 @@
...
+  CallOptions extends RunnableConfig
+> extends Runnable<RunInput, RunOutput, CallOptions> {
+  private url: string;
+
+  private options?: RemoteRunnableOptions;
+
+  lc_namespace = ["langchain", "schema", "runnable", "remote"];
+
+  constructor(url: string, options?: RemoteRunnableOptions) {
+    super();
```
jacoblee93, line 158: Pass `fields` here for serialization - I would actually define the signature as:

```typescript
  constructor(fields: {url: string, options: RemoteRunnableOptions}) {
    super(fields);
```

langchain/src/schema/runnable/remote.ts
```diff
@@ -0,0 +1,312 @@
...
+    batchOptions?: RunnableBatchOptions
+  ): Promise<(RunOutput | Error)[]> {
+    if (batchOptions?.returnExceptions) {
+      throw new Error("returnExceptions is not supported for remote clients");
+    }
+    const arr = options?.map((opts) =>
+      this._separateRunnableConfigFromCallOptions(opts)
+    );
+    const tup = arr?.reduce(
+      ([pc, pk], [c, k]) =>
```
jacoblee93, line 202: More descriptive variable names would be nice - hard to tell what's going on at a glance

langchain/src/schema/runnable/remote.ts
```diff
@@ -0,0 +1,312 @@
...
+        ] as [
+          RunnableConfig[],
+          Omit<Partial<CallOptions>, keyof BaseCallbackConfig>[]
+        ],
+      [[], []] as [
+        RunnableConfig[],
+        Omit<Partial<CallOptions>, keyof BaseCallbackConfig>[]
+      ]
+    );
+    const [configs, kwargs] = tup ?? [undefined, undefined];
```
jacoblee93, line 215: I think these should default to empty objects?

langchain/src/schema/runnable/remote.ts
```diff
@@ -0,0 +1,312 @@
...
+    const response = await this.post<{
+      inputs: RunInput[];
+      config?: (RunnableConfig & RunnableBatchOptions)[];
+      kwargs?: Omit<Partial<CallOptions>, keyof BaseCallbackConfig>[];
+    }>("/batch", {
+      inputs,
+      config: (configs ?? [])
+        .map(withoutCallbacks)
```
jacoblee93, line 223: I think `removeCallbacks` is a better name than `withoutCallbacks`

langchain/src/schema/runnable/remote.ts
```diff
@@ -0,0 +1,312 @@
...
+        `RemoteRunnable call failed with status code ${response.status}: ${json.message}`
+      );
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      (error as any).response = response;
+      throw error;
+    }
+    const { body } = response;
+    if (!body) {
+      throw new Error(
+        "Could not begin LangServe stream. Please check the given URL and try again."
```
jacoblee93, line 291: `begin remote stream` instead of `LangServe`?

langchain/src/schema/runnable/remote.ts
```diff
@@ -0,0 +1,312 @@
...
+      );
+    }
+    const stream = new ReadableStream({
+      async start(controller) {
+        const enqueueLine = getMessages((msg) => {
+          if (msg.data) controller.enqueue(deserialize(msg.data));
+        });
+        const onLine = (
```
jacoblee93, line 299: `onLine`?

langchain/src/schema/runnable/remote.ts
```diff
@@ -0,0 +1,312 @@
...
+          line: Uint8Array,
+          fieldLength: number,
+          flush?: boolean
+        ) => {
+          enqueueLine(line, fieldLength, flush);
+          if (flush) controller.close();
+        };
+        await getBytes(body, getLines(onLine));
```
jacoblee93, line 307: This works like an iterator right? It's not just aggregating?

langchain/src/schema/runnable/remote.ts
```diff
@@ -0,0 +1,312 @@
...
+    if (!set.has(elem)) {
+      return false;
+    }
+  }
+  return true;
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function revive(obj: any): any {
+  if (obj === null) return null;
```
jacoblee93, line 44: Why check specifically for `null` here?

jacoblee93 (commented):
Thanks! A few questions

langchain/src/schema/runnable/remote.ts
efriis: Removing! Good call.

langchain/src/schema/runnable/remote.ts
efriis: Will bring this up in slack. Even this is a slightly different implementation than python, which just passes a "whole" kwargs object in with the `batch` call, rather than assuming a different object per request in the batch. I actually like the js version better, but should make sure we're ok with that in general.

langchain/src/schema/runnable/remote.ts
efriis: Haha that's what it's called in the event-source-parse.ts file. Just a callback for each line. Might be worth reworking some of that logic to use proper streams rather than these callbacks.

langchain/src/schema/runnable/remote.ts
efriis: Correct

langchain/src/schema/runnable/base.ts
efriis, line 1: 2 main changes in this file:
1. Change a few instances of `RunnableConfig` to be `CallOptions` because it could contain "kwargs"
2. Change a few instances of `CallOptions` to `Partial<CallOptions>` because of other functions it interacts with

jacoblee93:
Looks good to me but you'll also want to create an entrypoint as shown here - call it `schema/runnable/remote`:

https://github.com/langchain-ai/langchainjs/blob/main/CONTRIBUTING.md#adding-an-entrypoint

Actually, I feel like this might be better outside of `schema` - could we move to just `src/runnable/remote` and make the entrypoint match?

PR #2830, opened 2023-10-07
Fix for Bedrock `amazon.titan-text-express-v1`

### LINKED ISSUE #2812: Can not get answer from Amazon Titan Text Express
I followed this documentation: https://js.langchain.com/docs/modules/model_io/models/llms/integrations/bedrock and created the following code:

```javascript
import { Bedrock } from "langchain/llms/bedrock";

const model = new Bedrock({
    model: "amazon.titan-text-express-v1",
    region: "us-east-1",
});
  
const res = await model.invoke("Tell me a joke");
console.log(res);
```

And the `res` I received is `undefined`.

Compare to the javascript version, the code in python works fine:
```python
from langchain.llms import Bedrock

llm = Bedrock(
    credentials_profile_name="default",
    model_id="amazon.titan-text-express-v1"
)

from langchain.chains import ConversationChain
from langchain.memory import ConversationBufferMemory

conversation = ConversationChain(
    llm=llm, verbose=True, memory=ConversationBufferMemory()
)

result = conversation.predict(input="What is the most exciting question you can ask an LLM?")
print(result)
```

Did I miss something in my javascript code?

cwoolum:
The Bedrock chat client was expecting the wrong response object for the Titan express model. This updates it to the correct response body object. During my testing, I could not get Bedrock to respond with more than one result in the response body.

*Fun fact*: Titan Express refers to itself as `Kyoko`.

Fixes #2812

### DIFF
--- a/langchain/src/chat_models/tests/chatbedrock.int.test.ts
+++ b/langchain/src/chat_models/tests/chatbedrock.int.test.ts
@@ -6,7 +6,7 @@ import { ChatBedrock } from "../bedrock.js";
 import { HumanMessage } from "../../schema/index.js";
 
 test("Test Bedrock chat model: Claude-v2", async () => {
-  const region = process.env.BEDROCK_AWS_REGION!;
+  const region = process.env.BEDROCK_AWS_REGION ?? "us-east-1";
   const model = "anthropic.claude-v2";
 
   const bedrock = new ChatBedrock({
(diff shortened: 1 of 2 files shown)


jacoblee93:
Thank you!

jacoblee93:
That is a fun fact also

PR #2831, opened 2023-10-07
update custom cypher example

tomasonjo:
Add good practices for providing custom queries for generating Cypher statements.


jacoblee93:
Thank you!

PR #2832, opened 2023-10-07
feat: add CloseVector as VectorStore provider

jo32:
Hi,

This is a pull request that adds support for [CloseVector](https://closevector.getmegaportal.com/), a cross-platform vector database that can currently run on both the browser and Node.js. For instance, you can create your index in Node.js and then load/query it in the browser. 

I've also implemented `saveToCloud` and `loadFromCloud` functions, which allow users to load and save the index to/from the CloudVector CDN.

**Regarding Vitest Integration**

`closevector-web` is a web-based package that requires the DOM and IndexedDB APIs. I attempted to implement Jest tests as shown below:

```closevector_web.test.ts
/**
 * @jest-environment jsdom
 */

import { beforeAll, expect, test } from "@jest/globals";
import { CloseVectorWeb } from "../closevector_web.js";
import { Document } from "../../document.js";
import { FakeEmbeddings } from "../../embeddings/fake.js";

import "fake-indexeddb/auto";

// remaining code
```

However, I encountered the following error:

```error.md
FAIL  src/vectorstores/tests/closevector_web.test.ts
  ● Test suite failed to run

    Jest encountered an unexpected token

    Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.

    Out of the box Jest supports Babel, which will be used to transform your files into valid JS based on your Babel configuration.

    By default "node_modules" folder is ignored by transformers.

    Here's what you can do:
     • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/ecmascript-modules for how to enable it.
     • If you are trying to use TypeScript, see https://jestjs.io/docs/getting-started#using-typescript
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/configuration
    For information about custom transformations, see:
    https://jestjs.io/docs/code-transformation

    Details:

    /Users/***/langchainjs/node_modules/uuid/dist/esm-browser/index.js:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,jest){export { default as v1 } from './v1.js';
                                                                                      ^^^^^^

    SyntaxError: Unexpected token 'export'

      at Runtime.createScriptFromCode (../node_modules/jest-runtime/build/index.js:1495:14)

Test Suites: 1 failed, 1 passed, 2 total
Tests:       2 passed, 2 total
Snapshots:   0 total
Time:        0.641 s, estimated 1 s
Ran all test suites.
```

I attempted to resolve this error, but the solution required numerous workarounds beyond my expertise. Consequently, I integrated Vitest to test the browser-based test cases that follow the **.vitest.ts pattern. Vitest also boasts impressive performance. If this is not acceptable, please let me know, and I will remove the related test cases and dependencies.


jacoblee93:
Very cool! Will review on Monday.

docs/extras/modules/data_connection/vectorstores/integrations/closevector.mdx
```diff
@@ -0,0 +1,51 @@
+import CodeBlock from "@theme/CodeBlock";
+
+# CloseVector
+
+:::tip Compatibility
+available on both browser and Node.js
+:::
+
+[CloseVector](https://closevector.getmegaportal.com/) is a cross-platform vector database that can be currently ran on both browser and Node.js. For example, you can create your index on Node.js and then load/query it on browser. For more information, please visit [CloseVector Docs](https://closevector-docs.getmegaportal.com/).
```
jacoblee93, line 9: s/that can be currently ran on both browser and Node.js/that can run in both the browser and Node.

examples/src/indexes/vector_stores/closevector.ts
```diff
@@ -0,0 +1,17 @@
+// If you want to import the browser version, use the following line instead:
+// import { CloseVectorWeb } from "langchain/vectorstores/closevector_node";
```
jacoblee93, line 2: Change to `closevector_web`

langchain/scripts/create-entrypoints.js
```diff
@@ -77,6 +77,8 @@ const entrypoints = {
   "vectorstores/elasticsearch": "vectorstores/elasticsearch",
   "vectorstores/memory": "vectorstores/memory",
   "vectorstores/cloudflare_vectorize": "vectorstores/cloudflare_vectorize",
+  "vectorstores/closevector_web": "vectorstores/closevector_web",
```
jacoblee93, line 80: You might prefer `vectorstores/closevector/web`? Up to you

langchain/src/vectorstores/closevector_node.ts
```diff
@@ -0,0 +1,286 @@
...
+/**
+ * Class that implements a vector store using Hierarchical Navigable Small
+ * World (HNSW) graphs. It extends the SaveableVectorStore class and
+ * provides methods for adding documents and vectors, performing
+ * similarity searches, and saving and loading the vector store.
+ */
+export class CloseVectorNode extends SaveableVectorStore {
+  declare FilterType: (doc: Document) => boolean;
+
+  _instance?: CloseVectorHNSWNode;
```
jacoblee93, line 38: Why `this._instance` and `this.instance`?

langchain/src/vectorstores/closevector_node.ts
```diff
@@ -0,0 +1,286 @@
...
+            }) || false
+        : undefined
+    );
+    const mapped = resp.map((x) => [
+      {
+        pageContent: x[0].pageContent,
+        metadata: x[0].metadata || {},
+      },
+      x[1],
+    ]) as [Document<Record<string, unknown>>, number][];
```
jacoblee93, line 203: Is cast necessary? Could we just create a `Document` from the `pageContent` and `metadata` and do `as const` instead?

langchain/src/vectorstores/closevector_web.ts
```diff
@@ -0,0 +1,276 @@
+import {
```
jacoblee93, line 1: Similar feedback to above

langchain/src/vectorstores/closevector_web.ts
```diff
@@ -0,0 +1,276 @@
...
+    await this.instance.saveToCloud(options);
+  }
+
+  /**
+   * Method to save the vector store to a directory. It saves the HNSW
+   * index, the arguments, and the document store to the directory.
+   * @param directory The directory to which to save the vector store.
+   * @returns A Promise that resolves when the vector store has been saved.
+   */
+  async save(directory: string): Promise<void> {
```
jacoblee93, line 110: What is `directory` in a web context?

jacoblee93:
Looks great overall!

re: `vitest`, would really like to avoid having two different test runners. Would something like this help for mocks?

https://github.com/jakearchibald/idb/issues/165#issuecomment-931682104

Or some other env setup?

https://github.com/langchain-ai/langchainjs/blob/main/langchain/jest.env.cjs#L3

Otherwise, I don't particularly like Jest and would be open to switching everything over to `vitest` if it makes testing in web environments easier generally - do you have any thoughts on comparing the two?

langchain/src/vectorstores/closevector_web.ts
jacoblee93: I also wonder if it'd be possible to share more code between the two implementations

docs/extras/modules/data_connection/vectorstores/integrations/closevector.mdx
jo32: thanks, fixed that

langchain/src/vectorstores/closevector_node.ts
jo32: I was thinking about making the setter private to prevent the instance from being changed from the outside.

langchain/scripts/create-entrypoints.js
jo32: Great suggestions! I have adjusted the structure of the code to make it clearer.

langchain/src/vectorstores/closevector_node.ts
jo32: Yes, you are correct. I have adjusted it according to your suggestion. Thank you!

langchain/src/vectorstores/closevector_web.ts
jo32: also fixed this. :)

langchain/src/vectorstores/closevector_web.ts
jo32: IndexedDB is used to mock the file system. So, in a web context, a directory is more or less like a key saved in IndexedDB. I have also added comments in the code.

langchain/scripts/create-entrypoints.js
jacoblee93: You'll also need to move the paths to match under `src/`

jo32:
> Looks great overall!
> 
> re: `vitest`, would really like to avoid having two different test runners. Would something like this help for mocks?
> 
> [jakearchibald/idb#165 (comment)](https://github.com/jakearchibald/idb/issues/165#issuecomment-931682104)
> 
> Or some other env setup?
> 
> https://github.com/langchain-ai/langchainjs/blob/main/langchain/jest.env.cjs#L3
> 
> Otherwise, I don't particularly like Jest and would be open to switching everything over to `vitest` if it makes testing in web environments easier generally - do you have any thoughts on comparing the two?

@jacoblee93 I noticed that SWC is already used in the Jest config file. So, in terms of performance, I think there are no significant differences. I might spend extra time trying to figure out what's wrong first and attempt to make it work with Jest.

jo32:
@jacoblee93

It seems that when using jsdom or happy-dom as the testing environment in Jest, it encounters an issue: https://github.com/jestjs/jest/issues/13739#issuecomment-1517190965.

I have tried all the solutions recommended here https://github.com/jestjs/jest/issues/13739#issuecomment-1517190965 and none have solved the problem.

In the end, I might have to rewrite all the test cases in vitest style and replace Jest with vitest. In this case, this commit will be too large and not focused enough. Therefore, I removed the web-related test cases and rewrote the node ones back in Jest style for the time being. However, the web-related functions are tested with vitest.

I might submit another issue, replacing Jest with vitest, if I figure out that vitest is the only option for now.

Also, all the other comments have been resolved and the code has been submitted.

langchain/scripts/create-entrypoints.js
jo32: Yes, references have been updated, including codes in examples and docs. yarn lint, yarn build, and yarn docs have been tested.

jacoblee93:
Ah bummer :( Thanks, will look as soon as I can!

langchain/src/vectorstores/closevector/node.ts
```diff
@@ -0,0 +1,175 @@
...
+      credentials
+    );
+    await instance.addDocuments(docs);
+    return instance;
+  }
+
+  static async imports(): Promise<{
+    HierarchicalNSW: typeof HierarchicalNSWT;
+  }> {
+    return await CloseVectorHNSWNode.imports();
```
jacoblee93, line 173: No need for return `await` here?

Also, I think no need for conditional import?

langchain/src/vectorstores/closevector/common.ts
```diff
@@ -0,0 +1,227 @@
+import type { CloseVectorHNSWNode } from "closevector-node";
+import type { CloseVectorHNSWWeb } from "closevector-web";
```
jacoblee93, line 2: This will cause import issues unfortunately :(

You'll need to separate imports out into the individual entrypoints.

langchain/src/vectorstores/closevector/node.ts
jo32: @jacoblee93

True! I have removed "await" in the latest code.

If possible, I might prefer to reserve this method to allow developers to debug the inner instance and also align it with the closevector-web implementation, which can be used to preload the JS resources on the web.

langchain/src/vectorstores/closevector/common.ts
jo32: @jacoblee93 

Ah! Sorry for not noticing that. I have deleted the import from closevector-node or closevector-web in common.ts; instead, I imported an type from a package named closevector-common, which is a package shared by closevector-node and closevector-web in latest code.

jo32:
@jacoblee93 Thank you for your feedback. The most recent comments have been addressed. I would greatly appreciate any additional suggestions you may have.

langchain/src/vectorstores/closevector/common.ts
```diff
@@ -0,0 +1,226 @@
...
+   * @returns A Promise that resolves when the documents have been added.
+   */
+  async addDocuments(documents: Document[]): Promise<void> {
+    await this.instance.addDocuments(documents);
+  }
+
+  async saveToCloud(
+    _options: Record<string, unknown>
+  ) {
+    throw new Error("not implemented");
```
jacoblee93, line 80: Shouldn't these be abstract methods?

langchain/src/vectorstores/closevector/common.ts
```diff
@@ -0,0 +1,226 @@
...
+  static async fromDocuments(
+    _docs: Document[],
+    _embeddings: Embeddings,
+    _args?: Record<string, unknown>,
+    _credentials?: CloseVectorCredentials
+  ): Promise<CloseVector<CloseVectorSaveableVectorStore>> {
+    throw new Error("not implemented");
+  }
+
+  static async imports(): Promise<{
```
jacoblee93, line 221: What's the purpose of this method? Looks like it's not called?

docs/snippets/modules/data_connection/vectorstores/get_started.mdx
```diff
@@ -116,3 +116,4 @@ Here's a quick guide to help you pick the right vector store for your use case:
 - If you are already utilizing SingleStore, or if you find yourself in need of a distributed, high-performance database, you might want to consider the [SingleStore](./integrations/singlestore) vector store.
 - If you are looking for an online MPP (Massively Parallel Processing) data warehousing service, you might want to consider the [AnalyticDB](./integrations/analyticdb) vector store.
 - If you're in search of a cost-effective vector database that allows run vector search with SQL, look no further than [MyScale](./integrations/myscale).
+- If you're in search of a vector database that can loaded on both browser and server side, look no further than [CloseVector](./integrations/closevector). It aims to running vector database on cross-platform.
```
jacoblee93, line 119: Some typos here - how about:

`- If you're in search of a vector database that you can load from both the browser and server side, check out [CloseVector](./integrations/closevector). It's a vector database that aims to be cross-platform.`

jacoblee93:
Think we're almost there!

jacoblee93:
Could you also run `yarn format`?

langchain/src/vectorstores/closevector/common.ts
jo32: Oh, yes, I have made it abstract.

docs/snippets/modules/data_connection/vectorstores/get_started.mdx
jo32: Great idea! I edited the docs in the latest commit.

langchain/src/vectorstores/closevector/common.ts
jo32: I was assuming LangChain might be used in some non-TypeScript environments that may not have solid type checking and tests. A class method in common.ts may be called directly. Thus, a runtime error is added to hint at such cases.

However, I forgot to check if there are any practices like this in LangChain. I have removed those methods in common.ts to adhere to the conventions.

jo32:
@jacoblee93 thank you. The code has been formatted and submitted again. It's a great honor to have the opportunity to contribute to a mature and rapidly developing open-source project like LangChain.

jacoblee93:
Thank you!

We were dealing with some CI timeout issues and docs refactors, so I'll need to merge in main but should be good :)

Issue #2807, opened 2023-10-06
Notion vectorStore integration breaks when importing a database with an empty row


natalie-bb:
file:///Users/natalie/chatbot/node_modules/langchain/dist/embeddings/openai.js:139
        const batches = chunkArray(this.stripNewLines ? texts.map((t) => t.replace(/\n/g, " ")) : texts, this.batchSize);
                                                              ^
TypeError: Cannot read properties of undefined (reading 'replace')
    at file:///Users/natalie/chatbot/node_modules/langchain/dist/embeddings/openai.js:139:76
    at Array.map (<anonymous>)
    at OpenAIEmbeddings.embedDocuments (file:///Users/natalie/chatbot/node_modules/langchain/dist/embeddings/openai.js:139:63)
    at HNSWLib.addDocuments (file:///Users/natalie/chatbot/node_modules/langchain/dist/vectorstores/hnswlib.js:48:54)
    at Function.fromDocuments (file:///Users/natalie/chatbot/node_modules/langchain/dist/vectorstores/hnswlib.js:258:24)
    at file:///Users/natalie/chatbot/tools/vectorStores/restaurants.ts:27:46
    at processTicksAndRejections (node:internal/process/task_queues:95:5)

jacoblee93:
@skarard can you have a look at this one?

skarard:
Sure.

skarard:
I was able to reproduce the bug and it opens a pretty solid question.

First the bug:
```typescript
/**
 * Interface for interacting with a document.
 */
export class Document<
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  Metadata extends Record<string, any> = Record<string, any>
> implements DocumentInput
{
  pageContent: string;

  metadata: Metadata;

  constructor(fields: DocumentInput<Metadata>) {
    this.pageContent = fields.pageContent
      ? fields.pageContent.toString()
      : this.pageContent;
    this.metadata = fields.metadata ?? ({} as Metadata);
  }
}
```

The constructor for the Document class will set `pageContent` to `undefined` if it's falsy (e.g. an empty string).

A Notion page contains properties that the `NotionAPILoader` store as metadata. Often in a Notion database, a page will not have any contents and the pages properties are where the data is stored.

@natalie-bb a work around to get the script working for now is to replace the `undefined` `pageContent` with an empty string.
```typescript
const workaroundDocs = docs.map(
[...]

PR #2836, opened 2023-10-08
chore(core) Document: add test checking pageContents empty string behavior

skarard:
This PR is related to a discussion in #2807, the current behaviour of the Document constructor `field.pageContent` when given an empty string is to set `pageContent` as `undefined`. This broke `HNSWLib.addDocuments()` as it attempts a string.replace() on `undefined` rather than the typed `string` expected.

I suggested allowing `pageContents` to be set as an empty string. @jacoblee93 suggested that a Document with an empty `pageContents` should not be created at all.

This implementation requires `pageContents` to be a string that is not empty, otherwise it throws `Error("pageContents must not be empty")`.


jacoblee93:
Ideally we shouldn't need this with TypeScript, but an empty string should be permissible, just not a hard `undefined`, `null`, or non-string value - can you change?

skarard:
Sure I change this out to allow it to record an empty string.

skarard:
@jacoblee93 I added the proposed change and threw in a couple of unit tests for good measure.

bracesproul (changes requested):
Hey @skarard sorry for taking so long to review, these changes look good! We've moved `document.ts` into `langchain-core/src/documents/document.ts`, could you push up a change moving your changes into that updated file then re-request my review?

Thanks again!

langchain-core/src/documents/tests/document.test.ts
```diff
@@ -0,0 +1,14 @@
+import { Document } from "../document.js";
```
benjamincburns, line 1:
```suggestion
import { Document } from "../document.js";
import { describe, expect, test} from "@jest/globals";
```

langchain-core/src/documents/tests/document.test.ts
```diff
@@ -0,0 +1,15 @@
...
+import { describe, expect, test} from "@jest/globals";
```
benjamincburns, line 2:
```suggestion
import { describe, expect, test } from "@jest/globals";
```

benjamincburns:
Thanks for the contribution!

benjamincburns (approved):

PR #2837, opened 2023-10-08
NotionAPILoader: Fix for #2807 and add `propertiesAsHeader` option

### LINKED ISSUE #2807: Notion vectorStore integration breaks when importing a database with an empty row

skarard:
Fixes #2807

Notion databases are often used as the data stored in the page properties, with an empty page. This means a user may expect a similarity search on the page properties. I have enabled an option where a user can prepend the page properties as a frontmatter header to enable similarity search.

It was bought to my attention that the preferred behaviour for the Document class is to not create documents with empty `pageContents`. 

This PR addresses both of these issues.
- Adding an option to prepend page properties as a front matter header
- Ensures would be empty `pageContents` pages do not create Documents

It also
- Adds integration tests for empty pageContents not creating a Document
- Adds integration tests for propertiesAsHeader option
- Update the database example to include propertiesAsHeader

### DIFF
--- a/examples/src/document_loaders/notionapi.ts
+++ b/examples/src/document_loaders/notionapi.ts
@@ -27,6 +27,7 @@ const dbLoader = new NotionAPILoader({
   callerOptions: {
     maxConcurrency: 64, // Default value
   },
+  propertiesAsHeader: true, // Prepends a front matter header of the page properties to the page contents
 });
 
 // A database row contents is likely to be less than 1000 characters so it's not split into multiple documents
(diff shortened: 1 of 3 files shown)


jacoblee93:
Thank you!

PR #2839, opened 2023-10-09
fix(stream): Update "IterableReadableStream" pull method to enqueue nullish data (e.g. empty string)

pyokang:
This PR is to address the issue our team identified while we were integrating `BaseLLM` with our own AI to enable the streaming capability. The issue we faced was, after we implemented `*_streamResponseChunks` generator function, we noticed it did not terminated in many cases.

After investigation, we found out that our response stream would end by yielding empty string, and this behavior, paired with `IterableReadableStream`'s pull method, which only enqueue's value if it is not nullish, will hang the generator. Hence preventing it from terminating successfully.

### DIFF
--- a/langchain/src/util/stream.ts
+++ b/langchain/src/util/stream.ts
@@ -61,9 +61,11 @@ export class IterableReadableStream<T> extends ReadableStream<T> {
     return new IterableReadableStream<T>({
       async pull(controller) {
         const { value, done } = await generator.next();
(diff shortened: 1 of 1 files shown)


jacoblee93:
Phenomenal catch - I think this makes more sense to check for `!== undefined` though? 

I'll update.

jacoblee93:
Huge fix, thanks!

PR #2841, opened 2023-10-09
Fix: Convert Neo4j Integers to strings in the result sets

oskarhane:
So the LLMs can interpret them correctly.
Added to test.

Also updated some of the code to use best practices.

---

About me:

Twitter handle: @oskarhane


tomasonjo:
LGTM :) thanks!

jacoblee93:
Thank you!

PR #2846, opened 2023-10-09
Expose PrismaSqlFilter type

BjoernRave:
I would like to be able to access the PrismaSqlFilter type so I can create a function like this: 

```
export const initVectorStore = (filter: PrismaSqlFilter<Chunk>) => {
  env.localModelPath = __dirname + "/../../models"

  //@ts-ignore
  env.allowRemoteModels = false

  const vectorStore = PrismaVectorStore.withModel<Chunk>(prisma).create(
    new HuggingFaceTransformersEmbeddings({
      modelName: "BAAI/bge-large-en-v1.5",
    }),
    {
      prisma: Prisma,
      tableName: "Chunk",
      vectorColumnName: "vector",
      columns: {
        content: PrismaVectorStore.ContentColumn,
      },
      filter
    },
    
  )

  return vectorStore
}
```

### DIFF
--- a/langchain/src/vectorstores/prisma.ts
+++ b/langchain/src/vectorstores/prisma.ts
@@ -58,7 +58,7 @@ type ModelColumns<TModel extends Record<string, unknown>> = {
   [K in keyof TModel]?: true | ColumnSymbol;
 };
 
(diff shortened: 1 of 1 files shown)


jacoblee93:
Thanks! Will be out in the next release

PR #2848, opened 2023-10-09
chore: bump zod and prisma to fix example build errors

bracesproul:
Zod and prisma deps in the examples dir were out of sync with the langchain dir, causing build errors


jacoblee93:
Thanks!

PR #2853, opened 2023-10-09
Allow ParentDocumentRetriever to subclass MultiVectorRetriever

jacoblee93:
~~Makes `Docstore` subclass `BaseStore<string, Document>`~~ Makes ParentDocumentRetriever take a `BaseStore` instead of a `Docstore`.


langchain/src/schema/index.ts
```diff
@@ -723,8 +724,29 @@ export abstract class BaseEntityStore extends Serializable {
...
-export abstract class Docstore {
+export abstract class Docstore extends BaseStore<string, Document> {
+  lc_namespace = ["langchain", "schema"];
+
   abstract search(search: string): Promise<Document>;
 
   abstract add(texts: Record<string, Document>): Promise<void>;
+
+  async mget(keys: string[]): Promise<Document[]> {
+    return Promise.all(keys.map((key) => this.search(key)));
```
eyurtsev, line 735: Discussed offline, this probably won't work since `search` may be fuzzy search through content (like a vectorstore)
rather than get page by id

jacoblee93:
@eyurtsev this will ensure backcompat for the existing docstore people are using with this 👍

langchain/src/retrievers/parent_document.ts
```diff
@@ -73,9 +67,9 @@ export class ParentDocumentRetriever extends BaseRetriever {
     }
     const parentDocs: Document[] = [];
     for (const parentDocId of parentDocIds) {
-      const parentDoc = await this.docstore.search(parentDocId);
-      if (parentDoc !== undefined) {
-        parentDocs.push(parentDoc);
+      const storedParentDocs = await this.docstore.mget([parentDocId]);
```
eyurtsev, line 70: can we use `mget(parentDocIds)` since it's the optimized variant

jacoblee93: Ah right

PR #2860, opened 2023-10-10
Fix CustomListOutputParser format instructions

jeromeetienne:
This PR contains a little fix on ```CustomListOutputParser``` argument handling.

I think there may be a bug in the format instruction of ```CustomListOutputParser```
- ```length``` constructor arguments controls the number of lines the user want - [documentation](https://js.langchain.com/docs/modules/model_io/output_parsers/custom_list_parser)
- it may be number or undefined if the user dont want to control the number of lines - [documentation](https://js.langchain.com/docs/api/output_parsers/classes/CustomListOutputParser)
- nevertheless, if length is undefined, the format instructions include "undefined items". 

if length constructor arguments is provided and equals to 3, the format instruction will be:
```
Your response should be a list of 3 items separated by "|" (eg: `foo| bar| baz`)
```

**BEFORE**: note the ```undefined items```
if length constructor arguments is not provided, the format instruction will be:
```
Your response should be a list of undefined items separated by "|" (eg: `foo| bar| baz`)
```

**AFTER**: if length constructor arguments is not provided, the format instruction will be:
```
Your response should be a list of items separated by "|" (eg: `foo| bar| baz`)
```

my twitter handle is https://twitter.com/jerome_etienne

PS: thanks for the excelent library!

---

Little test program:

```
import { CustomListOutputParser } from "langchain/output_parsers";

// With a `CustomListOutputParser`, we can parse a list with a specific length and separator.
const parser = new CustomListOutputParser({ 
  // length: 3, 
  separator: "|||" 
});

const formatInstructions = parser.getFormatInstructions();

console.log(`Format Instruction:\n${formatInstructions}`);
```


jacoblee93:
Thanks for the catch!

PR #2874, opened 2023-10-11
Fix: validate input or throw when using MessagesPlaceholder

### LINKED ISSUE #2871: Better error message for invalid input on `MessagesPlaceholder`
When trying to pass content to `MessagesPlaceholder` and it's not of type `Message` we should throw a better error

bracesproul:
Closes #2871 

Throw when input passed to `MessagesPlaceholder.formatMessages` is not of correct type

### DIFF
--- a/langchain/src/prompts/chat.ts
+++ b/langchain/src/prompts/chat.ts
@@ -158,9 +158,7 @@ export class MessagesPlaceholder<
     return [this.variableName];
   }
 
-  validateInputOrThrow(
-    input: Array<unknown>
-  ): input is BaseMessage[] {
+  validateInputOrThrow(input: Array<unknown>): input is BaseMessage[] {
     let isInputBaseMessage = false;
 
     if (Array.isArray(input)) {
--- a/langchain/src/prompts/chat.ts
+++ b/langchain/src/prompts/chat.ts
@@ -175,7 +175,7 @@ export class MessagesPlaceholder<
     if (!isInputBaseMessage) {
       const readableInput =
         typeof input === "string" ? input : JSON.stringify(input, null, 2);
-  
+
       const error = new Error(
         `Error: Field "${variableName}" in prompt uses a MessagesPlaceholder, which expects an array of BaseMessages as an input value. Received: ${readableInput}`
       );
--- a/langchain/src/prompts/chat.ts
+++ b/langchain/src/prompts/chat.ts
@@ -158,9 +158,35 @@ export class MessagesPlaceholder<
     return [this.variableName];
   }
 
+  validateInputOrThrow(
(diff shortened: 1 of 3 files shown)


langchain/src/prompts/chat.ts
```diff
@@ -158,9 +158,33 @@ export class MessagesPlaceholder<
...
 
+  validateInputOrThrow(input: Array<unknown>): input is BaseMessage[] {
+    let isInputBaseMessage = false;
+
+    if (Array.isArray(input)) {
+      isInputBaseMessage = input.every((message) =>
+        isBaseMessage(message as BaseMessage)
+      );
+    } else {
+      isInputBaseMessage = isBaseMessage(input as BaseMessage);
```
jacoblee93, line 169: Can you change `isBaseMessage` to tolerate `null/undefined` values?

https://github.com/langchain-ai/langchainjs/blob/main/langchain/src/schema/index.ts#L461

langchain/src/prompts/chat.ts
```diff
@@ -158,9 +158,33 @@ export class MessagesPlaceholder<
...
+    }
+
+    if (!isInputBaseMessage) {
+      const readableInput =
+        typeof input === "string" ? input : JSON.stringify(input, null, 2);
+      throw new Error(
```
jacoblee93, line 175: I would also add something to the error message about which input the array of messages is expected for

bracesproul: Refactored to `Class 'MessagesPlaceholder' expects an array of BaseMessage instances but received:...`

jacoblee93: And actually maybe add that specifically `MessagesPlaceholder expects...`

bracesproul: Updated

jacoblee93: Can we do: `Error: Field "${variableName}" in prompt uses a MessagesPlaceholder, which expects an array of BaseMessages as an input value. Received: ${readableInput}`

jacoblee93: And then also pass `variableName` into the validator?

bracesproul: Updated
```typescript
throw new Error(
        `Error: Field "${variableName}" in prompt uses a MessagesPlaceholder, which expects an array of BaseMessages as an input value. Received: ${readableInput}`
      );
```
and `validateInputOrThrow` now takes in two inputs, first the unknown value, and second the variable name

jacoblee93:
Thank you!

jacoblee93 (approved):

PR #2888, opened 2023-10-11
Unstructured API - additional props

perstarkse:
This commit adds additional optional props for use with the UnstructuredLoader. Info about them retrieved from https://github.com/Unstructured-IO/unstructured-api and https://unstructured-io.github.io/unstructured/api.html


jacoblee93:
Thank you!

PR #2891, opened 2023-10-11
Feat/streaming token usage

### LINKED ISSUE #2623: Support token usage reporting for streaming calls
```javacript
export const getAnswer = async (input) => {
  const {  question } = input

  const controller = new AbortController();
  
  const streamingModel = new ChatOpenAI({
    modelName: 'gpt-3.5-turbo',
    openAIApiKey: decryptData(openAIApiKey),
    streaming: true,
    maxRetries: 10,
    maxConcurrency: 10,
    callbacks: [
      {
        handleLLMNewToken(token) {
          // send to client        
        },
        handleLLMEnd: async (output) => {
          console.log(output.llmOutput.tokenUsage); <-- Problem: get empty {}
        },
        handleLLMError: async (err) => {
          console.error(err);
        },
      },
    ],
  });

  const vectorStore = await vectorStoreByChatbot();

   const prompt = ChatPromptTemplate.fromPromptMessages([
     SystemMessagePromptTemplate.fromTemplate(
       'The following is a friendly conversation between a human and an AI. The AI is talkative and provides lots of specific details from its context. If the AI does not know the answer to a question, it truthfully says it does not know.'
     ),
     new MessagesPlaceholder('history'),
     HumanMessagePromptTemplate.fromTemplate('{input}'),
   ]);

  
  const memory = new ConversationSummaryBufferMemory({
    llm: streamingModel,
    maxTokenLimit: 1024,
    returnMessages: true,
    memoryKey: 'chat_history',
    outputKey: 'text',
    humanPrefix: 'Customer',
  });
 
  const chain = ConversationalRetrievalQAChain.fromLLM(
    streamingModel,
    vectorStore.asRetriever(),
    {
      memory: memory,
      returnSourceDocuments: true,
      prompt: prompt,
    }
  );

  try {
     await chain.call({
      question: question,
      signal: controller.signal,
    });
  } catch (error) {
    console.log('Ask failed: ' + error);
  }
};
```
above is my code. could u help me check i cant get tokenUsage inside the function handleLLMEnd. i checked log get empty {}

wlyh514:
Add token usage support for `ChatOpenAI` streaming calls, including function calls. 

Fixes #2623


jacoblee93:
Sorry for the delay - this would be awesome but we'll need to evaluate the new dep. Give us a bit longer please.

wlyh514:
> Sorry for the delay - this would be awesome but we'll need to evaluate the new dep. Give us a bit longer please.

Thank you for your response!

I'm mainly using the `openai-chat-tokens` dep for calculating function definition token usage, specifically [this piece of code](https://github.com/hmarr/openai-chat-tokens/blob/main/src/functions.ts). I could instead copy it into the langchainjs codebase, as the `openai-chat-tokens` library is under MIT license. Will that be a better approach?

jacoblee93:
Yeah I think that would make this a bit more palatable - you can add them as a source under `langchain/src/util/` like we had to do for a few others for build/slight modification reasons

wlyh514:
Done! Let me know if there's other improvements I can add.

langchain/src/chat_models/openai.ts
```diff
@@ -523,19 +563,79 @@ export class ChatOpenAI<
...
+  }
+
+  async getNumTokensFromMessages(messages: BaseMessage[]) {
     let totalCount = 0;
     let tokensPerMessage = 0;
     let tokensPerName = 0;
 
     // From: https://github.com/openai/openai-cookbook/blob/main/examples/How_to_format_inputs_to_ChatGPT_models.ipynb
-    if (getModelNameForTiktoken(this.modelName) === "gpt-3.5-turbo") {
+    if (this.modelName === "gpt-3.5-turbo-0301") {
```
jacoblee93, line 635: Do you have a source for this?

langchain/src/chat_models/openai.ts
```diff
@@ -548,7 +648,23 @@ export class ChatOpenAI<
           message.name !== undefined
             ? tokensPerName + (await this.getNumTokens(message.name))
             : 0;
-        const count = textCount + tokensPerMessage + roleCount + nameCount;
+        let count = textCount + tokensPerMessage + roleCount + nameCount;
+
+        const openAIMessage = messageToOpenAIMessage(message);
+        if (openAIMessage.role === "function") {
+          count -= 2;
```
jacoblee93, line 655: Can you add a comment for why this is?

langchain/src/chat_models/openai.ts
```diff
@@ -470,7 +492,25 @@ export class ChatOpenAI<
...
+      const promptTokenUsage = await this.getNumTokensFromPrompt(
+        messages,
+        functions,
+        function_call
+      );
+      const completionTokenUsage = await this.getNumTokensFromGenerations(
+        generations
+      );
+
+      tokenUsage.promptTokens = promptTokenUsage;
```
jacoblee93, line 510: Since it's estimated, could we change these properties to e.g. `estimatedPromptTokens`?

I suspect we may see some updates at developer day next week

langchain/src/chat_models/openai.ts
```diff
@@ -523,19 +563,79 @@ export class ChatOpenAI<
...
+
+    let tokens = (await this.getNumTokensFromMessages(messages)).totalCount;
+
+    // If there are functions, add the function definitions as they count towards token usage
+    if (functions && function_call !== "auto") {
+      const promptDefinitions = formatFunctionDefinitions(
+        functions as unknown as FunctionDef[]
+      );
+      tokens += await this.getNumTokens(promptDefinitions);
+      tokens += 9; // Add nine per completion
```
jacoblee93, line 588: Why 9?

jacoblee93 (changes requested):
This is great, see questions/comments

langchain/src/chat_models/openai.ts
wlyh514: It is taken from https://github.com/hmarr/openai-chat-tokens/blob/main/src/index.ts, I'll add it into a comment

jacoblee93 (approved):

jacoblee93:
Thank you for your patience!

Stadly:
Great work on this, @wlyh514! Do you think it would be possible to implement token usage estimation also when using `streamEvents`? https://github.com/langchain-ai/langchainjs/issues/6583

wlyh514:
> Great work on this, @wlyh514! Do you think it would be possible to implement token usage estimation also when using `streamEvents`? #6583

Sorry for the late reply, let me see what I can do.

PR #2892, opened 2023-10-12
Add AnthropicFunctionsBedrock

efriis:


langchain/src/chat_models/bedrock.ts
```diff
@@ -86,7 +87,12 @@ export function convertMessagesToPrompt(
  * configured with various parameters such as the model to use, the AWS
  * region, and the maximum number of tokens to generate.
  */
-export class ChatBedrock extends SimpleChatModel implements BaseBedrockInput {
+export class ChatBedrock<
```
efriis, line 90: Added the generic to support passing in functions, function_call, and tools

PR #2906, opened 2023-10-13
Add YandexGPT LLM model

tyumentsev4:
Introducing an ability to work with the [YandexGPT](https://cloud.yandex.com/en/services/yandexgpt) language model.


jacoblee93:
Looks good, thank you!

PR #2908, opened 2023-10-13
fix: batch embedding queries, fix get client in tests

### LINKED ISSUE #2905: Bedrock Embeddings ThrottlingException
When attempting to generate a vector store using the ```fromDocuments``` method  on approximately 10,000 documents (specifically on the ```HNSWLib``` class) , it reaches the rate limit quickly. Most likely due to the ```BedrockEmbeddings.embedDocuments``` method just being a simple ```Promise.all``` on every document at once.

bracesproul:
Closes #2905 
Batch requests for aws bedrock embeddings

### DIFF
--- a/langchain/src/embeddings/bedrock.ts
+++ b/langchain/src/embeddings/bedrock.ts
@@ -4,7 +4,6 @@ import {
 } from "@aws-sdk/client-bedrock-runtime";
 import { Embeddings, EmbeddingsParams } from "./base.js";
 import type { CredentialType } from "../util/bedrock.js";
-import { chunkArray } from "../util/chunk.js";
 
 /**
  * Interface that extends EmbeddingsParams and defines additional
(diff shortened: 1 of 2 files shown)


langchain/src/embeddings/bedrock.ts
```diff
@@ -53,6 +55,7 @@ export class BedrockEmbeddings
       });
   }
 
+  /** @deprecated */
```
bracesproul, line 58: Side question, could this method be deleted instead or would that not maintain backcompat?

langchain/src/embeddings/bedrock.ts
```diff
@@ -93,13 +115,23 @@ export class BedrockEmbeddings
...
+   * @param documents Array of texts for which to generate embeddings.
    * @returns Promise that resolves to a 2D array of embeddings for each input document.
    */
-  embedDocuments(documents: string[]): Promise<number[][]> {
-    return Promise.all(documents.map((document) => this._embedText(document)));
+  async embedDocuments(documents: string[]): Promise<number[][]> {
+    const batches = chunkArray(documents, this.batchSize);
+    const embeddings: number[][] = [];
+
+    for (const batch of batches) {
```
jacoblee93, line 128: You don't actually have to batch here - just `Promise.all` everything. The `caller` will take care of concurrency.

jacoblee93: https://github.com/langchain-ai/langchainjs/blob/main/langchain/src/embeddings/openai.ts#L159

jacoblee93:
Awesome, thank you!

PR #2914, opened 2023-10-13
Use prefixMessages in ChatOpenAI

### LINKED ISSUE #2913: ChatOpenAI does not use perfixMessages in constructor params
`ChatOpenAI` class that's supposed to replace `OpenAIChat` does not use the `prefixMessages` in constructor params

glenchao:
Fixes #2913

### DIFF
--- a/langchain/src/chat_models/openai.ts
+++ b/langchain/src/chat_models/openai.ts
@@ -214,6 +214,8 @@ export class ChatOpenAI<
 
   modelName = "gpt-3.5-turbo";
 
+  prefixMessages?: OpenAIClient.Chat.ChatCompletionMessageParam[] | undefined;
+
   modelKwargs?: OpenAIChatInput["modelKwargs"];
 
   stop?: string[];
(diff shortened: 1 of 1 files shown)


jacoblee93:
Why not just pass a `ChatPromptTemplate`?

glenchao:
> Thanks for the PR! Why not just pass a `ChatPromptTemplate`?

Sorry, still very new to langchainjs... could you elaborate a bit more? I'm just trying to add what OpenAIChat had and is missing here. `prefixMessages` param is current unused.

jacoblee93:
Check out: https://js.langchain.com/docs/modules/model_io/prompts/prompt_templates/#chat-prompt-template

glenchao:
Thanks for the link. I understand chat prompt template, but I'm not sure what you are suggesting that I do with it. 

I'm trying to use the plan and execute agent with the ChatOpenAI model. I found two ways to get a system message to the agent.
1. chat model instantiation with prefix messages (while ChatOpenAI accepts the param, but it doesn't use it like OpenAIChat)
2. create my own planner instead of using the default one so I can supply my own prompt to the llm chain

jacoblee93:
Ah gotcha - option 2 is a better choice for this actually.

We'll be looking into improving our agent docs in general (CC @bracesproul). Going to close this for now.

PR #2920, opened 2023-10-14
Update graph_db_custom_prompt.ts

tomasonjo:
To avoid missing parameter error as https://github.com/langchain-ai/langchain/issues/11706, we should use double curly brackets in the cypher examples


jacoblee93:
Thank you!

PR #2921, opened 2023-10-14
fix: update imports for "parent_document_retriever" example

r13v:
There is an incorrect storage module import used in the "parent_document_retriever" example.


jacoblee93:
Thanks for the PR!

Both will work in the latest version of LangChain and we're going with the BaseStore for the future. Should still be backwards compatible though.

PR #2927, opened 2023-10-15
feat: runnables powered agent

### LINKED ISSUE #2919: Using fallbacks with "initializeAgentExecutorWithOptions"
I'm trying to use fallback model into the initializeAgentExecutorWithOptions function but I got this error:

```
node_modules\langchain\dist\agents\openai\index.cjs:121
        if (llm._modelType() !== "base_chat_model" || llm._llmType() !== "openai") {
                ^

TypeError: llm._modelType is not a function
    at OpenAIAgent.fromLLMAndTools (D:\Clad Work\Projects\IA ChatBot\chatbot-server\node_modules\langchain\dist\agents\openai\index.cjs:121:17)
    at initializeAgentExecutorWithOptions (D:\Clad Work\Projects\IA ChatBot\chatbot-server\node_modules\langchain\dist\agents\initialize.cjs:117:47)
```

My goal is to use the 16k version of GPT when needed because some of the tools results can be large but I don't want to use the 16k version without reason. Also as it's the tools output which are large I can't switch manually by catching error as all the tools actions will be lost and i will need to remake all the request again

Here is my code:

```

	let modelBase = new ChatOpenAI({
		modelName: process.env.OPENAI_CHAT_MODEL,
		temperature: 0.6,
		streaming: true
	});


	const modelLarge = new ChatOpenAI({
		modelName: "gpt-3.5-turbo-16k",
		temperature: 0.6,
		streaming: true
	});


	const model = modelBase.withFallbacks({
		fallbacks: [modelLarge],
	});
	const executor = await initializeAgentExecutorWithOptions(tools, model, {
		agentType: "openai-functions",
		memory: userMemories.get(username),
		verbose: true,
		agentArgs: {
			prefix,
		},
	});
```

What is the solution to use fallbacks with agent executor ?

Thanks in advance

bracesproul:
Closes #2919

Allow runnables to be passed to `Agent` or `AgentExecutor` class
Reassign agent input to be instance of `RunnableAgent` if it's runnable (inside `AgentExecutor`)
Deprecate `llmChain` in favor of `runnable` input for `Agent` class

Notes:
- do we even need to update the `Agent` class?


langchain/src/agents/tests/agent.int.test.ts
```diff
@@ -33,7 +35,61 @@ test("Run agent from hub", async () => {
...
   });
-  console.log(res);
+  console.log(
+    {
+      res,
+    },
+    "Run agent from hub response"
+  );
+  expect(res.output).not.toEqual("");
+  expect(res.output).not.toEqual("Agent stopped due to max iterations.");
```
bracesproul, line 45: Added these to all tests since in earlier commits they were passing with this message as the response

langchain/src/schema/runnable/base.ts
```diff
@@ -547,7 +547,7 @@ export abstract class Runnable<
 
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   static isRunnable(thing: any): thing is Runnable {
-    return thing.lc_runnable;
+    return thing ? thing.lc_runnable : false;
```
bracesproul, line 550: Without this I was getting a cannot read properties of undefined error

langchain/src/agents/agent.ts
```diff
@@ -122,6 +124,51 @@ export abstract class BaseSingleActionAgent extends BaseAgent {
...
+> extends BaseSingleActionAgent {
+  protected lc_runnable = true;
+
+  lc_namespace = ["langchain", "agents", "runnable"];
+
+  runnable: Runnable<RunInput, RunOutput>;
+
+  stop?: string[];
+
+  get inputKeys(): string[] {
```
jacoblee93, line 147: Do we need to populate this?

langchain/src/agents/chat/index.ts
```diff
@@ -165,7 +165,7 @@ export class ChatAgent extends Agent {
       args?.outputParser ?? ChatAgent.getDefaultOutputParser();
 
     return new ChatAgent({
-      llmChain: chain,
+      runnable: chain,
```
jacoblee93, line 168: I don't think we need to change old agents, just update what `AgentExecutor` accepts?

https://python.langchain.com/docs/modules/agents/how_to/agent_structured#create-the-agent

langchain/src/agents/chat_convo/index.ts
```diff
@@ -207,7 +207,7 @@ export class ChatConversationalAgent extends Agent {
       callbacks: args?.callbacks ?? args?.callbackManager,
     });
     return new ChatConversationalAgent({
-      llmChain: chain,
+      runnable: chain,
```
jacoblee93, line 210: See above

langchain/src/agents/executor.ts
```diff
@@ -1,25 +1,48 @@
...
-  agent: BaseSingleActionAgent | BaseMultiActionAgent;
-  tools: this["agent"]["ToolType"][];
+export interface AgentExecutorInput<
+  RunInput extends ChainValues & {
+    agent_scratchpad?: string | BaseMessage[];
+    stop?: string[];
+  } = any,
+  RunOutput extends AgentAction | AgentFinish = any
+> extends ChainInputs {
+  agent:
```
jacoblee93, line 41: This should always be `BaseSingleAction` or `BaseMultiAction`, input params should change to also allow `Runnable` and cast should occur in constructor

jacoblee93 (commented):
Don't think we need to change existing `Agent` extended implementations, just allow `AgentExecutor` to accept a runnable, new `RunnableAgent` class, and docs

langchain/src/agents/agent.ts
```diff
@@ -122,6 +124,51 @@ export abstract class BaseSingleActionAgent extends BaseAgent {
...
+/**
+ * Class representing a single action agent which accepts runnables.
+ * Extends the BaseSingleActionAgent class and provides methods for
+ * planning agent actions with runnables.
+ */
+export class RunnableAgent<
+  RunInput extends ChainValues & {
+    agent_scratchpad?: string | BaseMessage[];
+    stop?: string[];
+  } = any,
```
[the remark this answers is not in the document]
bracesproul: Wasn't causing lint to fail, I'm guessing if I run it it'll log those warning though. Will remove

jacoblee93:
Also, some tests would be helpful here!

langchain/src/agents/executor.ts
bracesproul: Resolved offline

langchain/src/agents/agent.ts
bracesproul: Resolved offline

langchain/src/agents/tests/agent.int.test.ts
bracesproul, line 1: add tests with structured outputs

langchain-infra:
https://langchainjs-docs-lztienuc2-langchain.vercel.app

langchain/src/agents/agent.ts
```diff
@@ -354,11 +403,26 @@ export abstract class Agent extends BaseSingleActionAgent {
       newInputs.stop = this._stop();
     }
 
-    const output = await this.llmChain.predict(newInputs, callbackManager);
+    /**
+     * If a runnable is passed in, then the output is `AgentAction | AgentFinish`.
+     * If an LLMChain was passed, the output will be `{ text: string }`.
+     */
+    const output = (await this.llmChain.invoke(newInputs, callbackManager)) as
```
jacoblee93, line 410: This shouldn't change?

langchain/src/agents/openai/index.ts
```diff
@@ -232,13 +232,13 @@ export class OpenAIAgent extends Agent {
...
 
-    // Split inputs between prompt and llm
     const llm = this.llmChain.llm as ChatOpenAI;
+
     const valuesForPrompt = { ...newInputs };
     const valuesForLLM: (typeof llm)["CallOptions"] = {
       tools: this.tools,
     };
-    for (const key of this.llmChain.llm.callKeys) {
+    for (const key of llm.callKeys) {
```
jacoblee93, line 241: Revert this file

jacoblee93 (changes requested):
Looks solid! I think a few files changed that shouldn't have, but otherwise just missing docs and should be good to go.

langchain/src/agents/tests/structured_output_runnables.int.test.ts
```diff
@@ -0,0 +1,118 @@
...
+import { createRetrieverTool } from "../toolkits/index.js";
+import { RecursiveCharacterTextSplitter } from "../../text_splitter.js";
+import { HNSWLib } from "../../vectorstores/hnswlib.js";
+import { OpenAIEmbeddings } from "../../embeddings/openai.js";
+import { formatToOpenAIFunction } from "../../tools/convert_to_openai.js";
+import { AgentExecutor } from "../executor.js";
+import { formatForOpenAIFunctions } from "../format_scratchpad.js";
+
+/** Define a custom structured output parser. */
+const structuredOutputParser = (
```
jacoblee93, line 22: This is basically the equivalent of this right?

https://python.langchain.com/docs/modules/agents/how_to/agent_structured#create-the-custom-parsing-logic

langchain/src/agents/tests/structured_output_runnables.int.test.ts
```diff
@@ -0,0 +1,118 @@
...
+    name: "state-of-union-retriever",
+    description:
+      "Query a retriever to get information about state of the union address",
+  });
+  /** Bind both retriever and response functions to LLM */
+  const llmWithTools = llm.bind({
+    functions: [
+      formatToOpenAIFunction(retrieverTool),
+      {
+        name: "Response",
```
jacoblee93, line 89: Style nit: lower case

langchain/src/agents/tests/structured_output_runnables.int.test.ts
bracesproul: Yeah, just wrote that but in ts

langchain/src/agents/executor.ts
```diff
@@ -1,25 +1,50 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  } = any,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  RunOutput extends AgentAction | AgentFinish = any
+> extends ChainInputs {
+  agent:
+    | BaseSingleActionAgent
+    | BaseMultiActionAgent
+    | Runnable<RunInput, RunOutput>;
+  tools: ExtractToolType<this["agent"]>[];
```
jacoblee93, line 47: Do we actually need this? Can't we just declare this on `RunnableAgent`?

jacoblee93 (commented):
Looks good, one more question!

docs/docs/expression_language/cookbook/agents.md
```diff
@@ -0,0 +1,179 @@
...
+};
+```
+
+After this, we can bind our two functions to the LLM, and create a runnable sequence which will be used as the agent.
+
+```typescript
+const llmWithTools = llm.bind({
+  functions: [
+    formatToOpenAIFunction(retrieverTool),
+    {
```
jacoblee93, line 130: Go into a bit more detail here about this response function

langchain/src/agents/executor.ts
bracesproul: The issue is we're not taking in a `RunnableAgent`, instead we're taking in a runnable or base agent, and reassigning the runnable to `RunnableAgent`
If we want to remove `ExtractToolType` we'd need to add `ToolType` to the base `Runnable` class

docs/docs/expression_language/cookbook/agents.md
jacoblee93: I would actually declare it as a separate variable above

docs/docs/expression_language/cookbook/agents.md
```diff
@@ -0,0 +1,179 @@
...
+  res,
+});
+```
+
+The output will look something like this
+
+```typescript
+{
+  res: {
+    inputs: "{\n" +
```
jacoblee93, line 171: Why is this `inputs`? And could our output parser parse this into an object?

langchain/src/agents/executor.ts
jacoblee93: Oh right. Never mind

docs/docs/expression_language/cookbook/agents.md
bracesproul: No particular reason, I had just defined that as the output in `structuredOutputParser` (copied from py)
I could spread the response object instead so it would be 
```typescript
{
  res: {
    '  "answer": "The President nominated Circuit Court of Appeals Judge Ketanji Brown Jackson, who is described as one of our nation’s top legal minds and will continue Justice Breyer’s legacy of excellence.",\n' +
    '  "sources": [\n' +
    '    "State of the Union Address"\n' +
    "  ]\n"
  }
}
```

jacoblee93: Just a bit counterintuitive that the resulting output is `inputs`

bracesproul: Ya makes sense, pushed a commit which spreads the return values

PR #2929, opened 2023-10-16
fix: update llm_streaming_stream_method.ts doc

cjbt:
Removes unneeded comments that can cause confusion
- `.stream()` works without the need to add `streaming: true` to the LLM constructor
- Removes reference to `handleLLMNewToken` since `handleLLMNewToken` does not exist within [this example](https://js.langchain.com/docs/modules/model_io/models/llms/how_to/streaming_llm#using-stream)

https://twitter.com/cjtantay


jacoblee93:
Thank you!

PR #2930, opened 2023-10-16
Add YandexGPT Chat model

tyumentsev4:
Introducing an ability to work with the [YandexGPT](https://cloud.yandex.com/en/services/yandexgpt) chat model.


jacoblee93:
Hey @tyumentsev4, changed the name to `ChatYandexGPT` to match our convention but looks good otherwise! Will merge if that change is ok with you?

tyumentsev4:
@jacoblee93 thank you everything is fine

jacoblee93:
Great, thanks!

PR #2931, opened 2023-10-16
Make format_instructions an optional argument in chat agent

EKMeyerson:
This simple change allows users to fully overwrite the prompts when creating a chat agent, and optimize them for their use case. The change makes the agent code much more generally useful. Previously, they could overwrite everything except for FORMAT_INSTRUCTIONS, which is essential to customization.


jacoblee93:
Seems good to me, thanks!

PR #2932, opened 2023-10-16
Mark the docs scripts files as executable.

obi1kenobi:
Script files like `build.sh` and `dev.sh` in `docs` are meant to be executed, so let's mark them as executable.


jacoblee93:
Thanks!

PR #2933, opened 2023-10-16
Update repo path to point to `langchain-ai` GitHub org.

obi1kenobi:
The repo is now part of the `langchain-ai` GitHub org and not under the `hwchase17` account.


jacoblee93:
Thank you!

PR #2942, opened 2023-10-17
Weaviate - throw error if storing vectors fails

the-powerpointer:
When storing vectors at Weaviate, different errors may occur. The current implementation already processes some of them, like connection reset or similar.
However, still when we get an API response, this does not mean that the insertion into the DB was successful.
Examples of such failures:

- store a document without having the schema created before
- store a document with some metadata of a type which is different than the current schema

For such errors we need to check the response from the API call.

The former implementation resulted in such cases in the weird status that the method `addVectors(..)` returned an array of ids, suggesting that the objects are available in Weaviate, but they are not. Furthermore no error is logged at all, so one don't even notice that and what went wrong.

This change collects all error messages from the API responses, and if there are any, lets the method `addVectors(..)` throw an Error containing those messages.


langchain/src/vectorstores/weaviate.ts
```diff
@@ -164,12 +164,22 @@ export class WeaviateStore extends VectorStore {
     });
 
     try {
-      await this.client.batch
+      const responses = await this.client.batch
```
jacoblee93, line 167: So this was buggy before and `client.batch` doesn't throw at all?

langchain/src/vectorstores/weaviate.ts
```diff
@@ -164,12 +164,22 @@ export class WeaviateStore extends VectorStore {
...
         .do();
+      // if storing vectors fails, we need to know why
+      const errorMessages: string[] = [];
+      responses.forEach((response) => {
+        if (response?.result?.errors?.error) {
+          errorMessages.push(...response.result.errors.error.map((err) => err.message ?? ""));
+        }
+      });
+      if (errorMessages.length > 0) {
+        throw new Error(errorMessages.join("\n"));
```
jacoblee93, line 179: Might be good to have some base case here if no messages can be parsed - e.g. `Unspecified error`

jacoblee93 (changes requested):
Thanks for this! Can you also run `yarn format` and `yarn lint`?

langchain/src/vectorstores/weaviate.ts
the-powerpointer: Ahm, no. As written in the PR description, certain errors were already catched correctly. Just try and address a server, which does not exist, e.g. `http://localhost:8080` -> that results in an error `FetchError: request to http://localhost:8080/v1/batch/objects failed, reason: connect ECONNREFUSED ::1:8080` also without the change of this PR.

However, if the server processes the request and sends a response, it still may contain an error. In my cases these were always 4xx errors, mostly 422 - Unprocessable Entity. However, in principle I could also imagine that Weaviate could respond with 5xx errors this way.
These kind of errors are catched with the changes of this PR.

langchain/src/vectorstores/weaviate.ts
the-powerpointer: Yep, I didn't really think about a sensible fallback in line 175 and just took the empty string to fulfil typing restrictions.
Anyways, just preparing for this possibility I'd change the fallback in line 175 from empty string to `"!! Unfortunately no error message was presented in the API response !!"`. 
Commit follows.

the-powerpointer:
Ideally I'd like to add some tests to prove that 4xx and 5xx errors are handled correctly. However, this requires some deep mocking of the Weaviate classes, and I'm unfortunately not an expert in JS/TS at all 😟, not to name deep mocking in Jest...

Anyway, what is the issue with deploying the preview? Is that anything I can influence? I saw that also other PRs fail with this error...

jacoblee93:
Yeah we're having an issue with Vercel deploys. Don't worry about it, will merge tonight!

jacoblee93 (approved):

jacoblee93:
Thanks!

PR #2944, opened 2023-10-17
Run exports tests in parallel on GitHub CI.

obi1kenobi:
This is about 6min faster than before: [~13min](https://github.com/langchain-ai/langchainjs/actions/runs/6550835570/job/17790748230) -> ~7min now. The issue previously was that we were running 7 CPU- and network-intensive docker containers on the same tiny 2-core GitHub runner. This PR splits each container on its own runner, which means they all run in parallel.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,21 +83,6 @@ jobs:
         run: yarn run test:unit
 
   test-exports:
-    name: Environment Tests
-    runs-on: ubuntu-latest
-    env:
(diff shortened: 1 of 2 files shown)


jacoblee93:
Thank you!

obi1kenobi:
I think Vercel failed due to the docs timeout @jacoblee93 mentioned. I strongly doubt my PR caused it, given that I didn't change anything at all about the docs configuration.

PR #2945, opened 2023-10-17
Feat: violations of expectations chain

bracesproul:
Based on [this](https://arxiv.org/pdf/2310.06983.pdf) paper
Implemented because will be useful for the long term memory app


langchain/scripts/create-entrypoints.js
```diff
@@ -34,6 +34,7 @@ const entrypoints = {
   "chains/query_constructor/ir": "chains/query_constructor/ir",
   "chains/sql_db": "chains/sql_db/index",
   "chains/graph_qa/cypher": "chains/graph_qa/cypher",
+  "chains/violation_of_expectations": "chains/violation_of_expectations/violation_of_expectations_chain",
```
jacoblee93, line 37: Filename might need to be `violation_of_expectations/index.ts` or `violation_of_expectations.ts` - I seem to remember some build issues?

jacoblee93: Also feels like we should put this under `experimental`?

jacoblee93:
This is really neat, would say belongs in `experimental` at the moment though - perhaps a notebook showing off an example flow as well?

langchain/src/chains/violation_of_expectations/violation_of_expectations_chain.ts
```diff
@@ -0,0 +1,473 @@
...
+   *
+   * @description
+   * The method iterates over the chat history and pushes each message into a temporary array.
+   * When it encounters an AI message, it checks for a subsequent user message.
+   * If a user message is found, it is considered as the user response to the AI message.
+   * If no user message is found after the AI message, the user response is undefined.
+   * The method then pushes the chunk (sequence of messages and user response) into the result array.
+   * This process continues until all messages in the chat history have been processed.
+   */
+  getMessageChunks(chatHistory: BaseMessage[]): MessageChunkResult[] {
```
jacoblee93, line 120: Naming is a vague to me - basically we're returning the conversation to date and whatever the user responded with?

bracesproul:
> This is really neat, would say belongs in `experimental` at the moment though - perhaps a notebook showing off an example flow as well?
> 
> `langchain/experimental/chains/violation_of_expectation`
> 
> The idea is you'll call this on a new message in the chat history right?

Experimental is probably a good idea, will refactor.

I've set it up to take in an entire chat, and process them all at once. I didn't think it would make sense to set it up in a way where it runs once for each chat message, because that implies use while a chat is ongoing. Since it's so slow I don't think that's practical.

langchain/src/chains/violation_of_expectations/violation_of_expectations_chain.ts
bracesproul: Agree my naming on methods could be improved, will fix

bracesproul: Refactored to `chunkMessagesByAIResponse`

jacoblee93:
Thank you!

PR #2948, opened 2023-10-18
Feat: runnable assign

bracesproul:
Added `RunnableAssign` method & `.assign()` method on `RunnablePassthrough`
Updated LCEL docs for SQL to use `RunnablePassthrough.assign()` --will update more in another PR to keep this one small.
I had to add a `public getStepsKeys()` to `RunnableMap` since the steps property was protected


langchain/src/schema/runnable/passthrough.ts
```diff
@@ -26,4 +105,16 @@ export class RunnablePassthrough<RunInput> extends Runnable<
       options
     );
   }
+
+  static assign(
+    mapping: Record<
+      string,
+      | Runnable<Record<string, unknown>, unknown>
+      | ((input: Record<string, unknown>) => unknown)
```
jacoblee93, line 113: Use `RunnableLike` here:

https://github.com/langchain-ai/langchainjs/blob/main/langchain/src/schema/runnable/base.ts#L23

langchain/src/schema/runnable/passthrough.ts
```diff
@@ -1,6 +1,85 @@
...
+        }
+      }
+
+      if (Object.keys(filtered).length > 0) {
+        yield filtered as RunOutput;
+      }
+    }
+
+    const firstMapChunk = await mapOutput.next();
+    if (firstMapChunk.value) {
```
jacoblee93, line 58: Why a special case?

langchain/src/schema/runnable/passthrough.ts
```diff
@@ -1,6 +1,85 @@
...
+  async invoke(input: RunInput, options?: Partial<CallOptions>) {
+    const mapperResult = await this.mapper.invoke(input, options);
+
+    return {
+      ...input,
+      ...mapperResult,
+    } as RunOutput;
+  }
+
+  async *transform(
```
jacoblee93, line 31: Will want a test case for this, a lot of subtleties here

langchain/src/schema/runnable/passthrough.ts
```diff
@@ -1,6 +1,85 @@
...
+    this.mapper = mapper;
+  }
+
+  async invoke(input: RunInput, options?: Partial<CallOptions>) {
+    const mapperResult = await this.mapper.invoke(input, options);
+
+    return {
+      ...input,
+      ...mapperResult,
+    } as RunOutput;
```
jacoblee93, line 28: Cast is necessary? Can we just add it as the return type above?

langchain/src/schema/runnable/passthrough.ts
jacoblee93: Also, is this implemented generally for runnable maps yet in JS?

langchain/src/schema/runnable/passthrough.ts
```diff
@@ -1,6 +1,85 @@
...
+    options?: Partial<CallOptions>
+  ): AsyncGenerator<RunOutput> {
+    yield this.invoke(input, options);
+  }
+
+  async stream(
+    input: RunInput,
+    options?: Partial<CallOptions>
+  ): Promise<IterableReadableStream<RunOutput>> {
+    return IterableReadableStream.fromAsyncGenerator(
```
jacoblee93, line 77: Do you need the override here?

langchain/src/schema/runnable/passthrough.ts
jacoblee93: Yeah doesn't look like maps support stream/transform chunks yet - would be good to add at some point but separately.

jacoblee93 (commented):
Looks solid, def needs some tests too though

langchain/src/schema/runnable/passthrough.ts
bracesproul: We have to cast since the result of .invoke is `Record<string, any>` and not RunOutput

The runnable.invoke types are `Runnable<RunInput, any>` so it's being typed as `Record<string, any>` on the RunnableMap side

langchain/src/schema/runnable/passthrough.ts
bracesproul: Removed entirely, issue #2978 is tracking implementing it in runnable map & assign

langchain/src/schema/runnable/passthrough.ts
jacoblee93: Got it, we should try to fix the typing at some point

jacoblee93:
Thank you!

jacoblee93 (commented):
Thanks!

PR #2949, opened 2023-10-18
fix: ensure metadata is cast to jsonb for pgvector

devinburnette:
## Description

While trying to integrate pgvector as a vector store, I ran into the following error when making a query with langchain:

```
2023-10-16 18:27:54.326 EDT [87858] ERROR:  operator does not exist: json @> unknown at character 106
2023-10-16 18:27:54.326 EDT [87858] HINT:  No operator matches the given name and argument types. You might need to add explicit type casts.
2023-10-16 18:27:54.326 EDT [87858] STATEMENT:
              SELECT *, embedding <=> $1 as "_distance"
              FROM langchain_pg_embedding
              WHERE cmetadata @> $2
              ORDER BY "_distance" ASC
              LIMIT $3;
```

## The problem

The `@>` operator is only supported by JSONB I believe. I use the python version of langchain to run the document loaders and store my embeddings in the vector store. Over there, they are creating the embeddings table with the metadata column as JSON data type, instead of JSONB. I have a [PR](https://github.com/langchain-ai/langchain/pull/11897) over there to change that, but there may be more implications to doing so than benefits, I'm not really sure yet, waiting to hear back from the maintainers.

So I'm proposing this change to ensure that we always cast the metadata as jsonb to prevent this error in the case that someone wants to have this column as JSON, but still maintain query functionality parity between the 2 libraries.

## Testing / Reproducing it

You can reproduce this, by creating a pgvector store using the python langchain library and then try a similar search on it using this library and you should get the same error. Workarounds I've been using have been to just manually alter the metadata column in my database from JSON to JSONB after it's created by the python langchain, but I'd really rather not do it that way.

I tested this change manually on my database by doing a simple select on the table with the cast in place to verify it works with both data types:

```sql
alter table langchain_pg_embedding alter column cmetadata set data type jsonb;
select cmetadata from langchain_pg_embedding where cmetadata::jsonb @> '{"file_name": "PULL_REQUEST_TEMPLATE.md"}' limit 1;
alter table langchain_pg_embedding alter column cmetadata set data type json;
select cmetadata from langchain_pg_embedding where cmetadata::jsonb @> '{"file_name": "PULL_REQUEST_TEMPLATE.md"}' limit 1;
```


jacoblee93:
Awesome, thank you! I think this seems reasonable, will test more later.

jacoblee93:
Tested and seems to be good - thank you for this 🙏

PR #2955, opened 2023-10-18
zep: Add native MMR to Zep Retriever

danielchalef:
- Adds native MMR support to Zep Memory Retriever
- Adds filtering to Zep Memory Retriever
- Adds examples for both of the above to existing retriever file
- Bumps required version of `zep-js`


jacoblee93:
Thank you!

PR #2959, opened 2023-10-18
Feat/security notice openapi

bracesproul:


docs/docs/modules/agents/toolkits/openapi.md
```diff
@@ -45,3 +45,9 @@ export const run = async () => {
   );
 };
 ```
+
+# Disclaimer ⚠️
+
+This agent can make requests to external APIs. Use with caution, especially when granting access to users.
+
+Consider adding limitations to what actions can be preformed via the agent, what APIs it can access, what headers can be passed and more.
```
jacoblee93, line 53: The potential issue is really that this chain can theoretically send requests with provided credentials/other secrets to hallucinated URLs (thought it should never) - would rephrase around that

jacoblee93 (commented):
Small comment

docs/docs/modules/agents/toolkits/openapi.md
bracesproul: Resolved with overall improved language

jacoblee93:
Thanks! We should also add this here:

https://js.langchain.com/docs/modules/chains/popular/api

PR #2969, opened 2023-10-19
Add ERNIE-Bot-4 from Baidu Wenxin Qianfan

sunner:
ERNIE-Bot-4 is the latest ERNIE serie model trained by Baidu. CEO said it is as good as GPT-4.

This commit also refactors the tests so that it will be easier to add new models or test cases.

NOTE: I removed `.skip` in tests. If langchain needs baidu keys to deploy ci/cd, I can provide.


jacoblee93:
Thank you! And thanks for the offer but we want to keep the overall count of integration tests down - in the future we may split individual integrations like this into their own packages, and running the test would make sense there.

PR #2980, opened 2023-10-19
fix: wildcard suffix for yielding keys

bracesproul:
When yielding keys, the `.yieldKeys(...)` method on `vercel_kv` & `ioredis` automatically checks if the prefix ends with `*`, and if not it applies it to the end.

Refactored tests a little to automatically clear all keys/disconnect

### DIFF
--- a/langchain/src/storage/tests/ioredis.int.test.ts
+++ b/langchain/src/storage/tests/ioredis.int.test.ts
@@ -4,27 +4,57 @@ import { test } from "@jest/globals";
 import { Redis } from "ioredis";
 import { RedisByteStore } from "../ioredis.js";
 
-test("RedisByteStore", async () => {
-  const store = new RedisByteStore({
-    client: new Redis({}),
+describe("RedisByteStore", () => {
+  const client = new Redis({});
+
+  afterEach(async () => await client.flushall());
+
+  afterAll(() => client.disconnect());
+
+  test("RedisByteStore", async () => {
+    const store = new RedisByteStore({
+      client,
+    });
+    const encoder = new TextEncoder();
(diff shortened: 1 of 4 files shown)


langchain/src/storage/tests/ioredis.int.test.ts
```diff
@@ -4,27 +4,57 @@ import { test } from "@jest/globals";
...
-  const store = new RedisByteStore({
-    client: new Redis({}),
+describe("RedisByteStore", () => {
+  const client = new Redis({});
+
+  afterEach(async () => await client.flushall());
+
+  afterAll(() => client.disconnect());
+
+  test("RedisByteStore", async () => {
```
bracesproul, line 14: This old tests is unchanged, except it's now inside this describe block & the client isn't initialized inside the test

langchain/src/storage/tests/vercel_kv.int.test.ts
```diff
@@ -7,58 +7,100 @@ import { VercelKVStore } from "../vercel_kv.js";
...
-  }
-  await store.mdelete(["key1", "key2"]);
-  const retrievedValues2 = await store.mget(["key1", "key2"]);
-  expect(retrievedValues2).toEqual([undefined, undefined]);
+describe("VercelKVStore", () => {
+  const client = getClient();
+
+  afterEach(async () => await client.flushall());
+
+  test("VercelKVStore can preform all operations", async () => {
```
bracesproul, line 28: This old test is unchanged, except it's now inside this describe block & the client isn't initialized inside the test

jacoblee93:
Nice catch, thanks!

PR #2986, opened 2023-10-20
Fixed order of LCEL menu and added tables in interface page

tsensei:
Fixes # (issue)

Fixed the ordering of the LCEL sub menu which was not in sync with the python docs and not correctly ordered.
Also added input type and output type table in the interface page of better readability.


bracesproul (approved):
lgtm, cc @jacoblee93 for running gh actions/docs preview

jacoblee93:
Thank you!

PR #3002, opened 2023-10-21
feat: update mvi integration for more metadata types

malandis:
Updates the Momento Vector Index integration to use the latest client library, which enables more diverse metadata value types. Because of this, we do not need to serialize the metadata in the langchain plugin to json anymore.

With this change, we also update the integration test to exercise the more diverse metadata value types.


jacoblee93:
Cool, sounds good!

malandis:
@jacoblee93 thanks for the pointer! I had not updated the examples with the newer version of the Momento client. Having done that the build and format pass. Lint passes in the `langchain` subdirectory.

jacoblee93:
Thanks!

PR #3003, opened 2023-10-21
IterableReadableStream.fromReadableStream reader releaseLock

### LINKED ISSUE #3001: IterableReadableStream.fromReadableStream didn't releaseLock source stream's reader
### IterableReadableStream.fromReadableStream didn't releaseLock source stream's reader

const reader = stream.getReader(); 

this reader can't be releaseLock!!! Sometimes, The target IterableReadableStream is cancel, but the source ReadableStream is continue work, e.g the websocket stream. So we should release the reader.

```typescript
  static fromReadableStream<T>(stream: ReadableStream<T>) {
    // From https://developer.mozilla.org/en-US/docs/Web/API/Streams_API/Using_readable_streams#reading_the_stream
    const reader = stream.getReader(); // this reader can't be releaseLock!!! 
    return new IterableReadableStream<T>({
      start(controller) {
        return pump();
        async function pump(): Promise<T | undefined> {
          return reader.read().then(({ done, value }) => {
            if (done) {
              controller.close();
              return;
            }
            controller.enqueue(value);
            return pump();
          });
        }
      },

    });
  }
```

AIXGEEK:
Fixes #3001 IterableReadableStream.fromReadableStream didn't releaseLock source stream's reader

const reader = stream.getReader();

this reader can't be releaseLock!!! Sometimes, The target IterableReadableStream is cancel, but the source ReadableStream is continue work, e.g the websocket stream. So we should release the reader.

### DIFF
--- a/langchain/src/util/stream.ts
+++ b/langchain/src/util/stream.ts
@@ -54,6 +54,9 @@ export class IterableReadableStream<T> extends ReadableStream<T> {
           });
         }
       },
+      cancel() {
+        reader.releaseLock();
+      },
     });
   }
 


jacoblee93:
Thank you for this!

Issue #1296, opened 2023-05-17
Error: No "GlobalWorkerOptions.workerSrc" specified.


geekros:
LangChain Version：0.0.75
Development Environment：Vue3+Vite+Ts+Electron

My usage process is as follows:

```
yarn add pdf-parse && yarn add pdfjs-dist
```

```
import { PDFLoader } from "langchain/document_loaders/fs/pdf";
const loader = new PDFLoader("D:\\202305\\Chat_Desktop\\example_data\\Analysis-and-Comparison-between-Optimism-and-StarkNet.pdf", {
    pdfjs: () => import("pdfjs-dist/legacy/build/pdf.js"),
});
docs = await loader.load();
```

Error：
```
Uncaught (in promise) Error: No "GlobalWorkerOptions.workerSrc" specified.
at get workerSrc [as workerSrc] (api.js:2304:11)
at _PDFWorker._initialize (api.js:2118:13)
at new _PDFWorker (api.js:2067:10)
at getDocument (api.js:384:9)
at PDFLoader.parse (D:\202305\Chat_Desktop\release\dist\preload\index.cjs:2903:27)
at async onImport (dialog.vue:161:36)
```

geekros:
```
import { PDFLoader } from "langchain/document_loaders/fs/pdf";
const loader = new PDFLoader("D:\\202305\\Chat_Desktop\\example_data\\Analysis-and-Comparison-between-Optimism-and-StarkNet.pdf", {
    pdfjs: () => import("pdfjs-dist/legacy/build/pdf.worker.js"),
});
docs = await loader.load();
```
Error reporting after using pdf.worker.js：
```
Uncaught (in promise) TypeError: getDocument is not a function
at PDFLoader.parse (D:\202305\Chat_Desktop\release\dist\preload\index.cjs:2924:27)
at async onImport (dialog.vue:159:36)
parse @ D:\202305\Chat_Desktop\release\dist\preload\index.cjs:2924
Promise.catch (async)
callWithAsyncErrorHandling @ runtime-core.esm-bundler.js:184
emit @ runtime-core.esm-bundler.js:730
(anonymous) @ runtime-core.esm-bundler.js:7465
handleClick @ button.vue:116
callWithErrorHandling @ runtime-core.esm-bundler.js:173
callWithAsyncErrorHandling @ runtime-core.esm-bundler.js:182
invoker @ runtime-dom.esm-bundler.js:345
```

geekros:
[...]

PR #3010, opened 2023-10-22
WebPDFLoader: allow usage of custom pdfjs build

DennisKo:
The node version of the `PDFLoader` allows for passing in a custom `pdfjs` build. Let's add that for the `WebPDFLoader` as well.

This allows fixing of errors like `Error: No "GlobalWorkerOptions.workerSrc" specified.` as issued here for example https://github.com/langchain-ai/langchainjs/issues/1296. It's a pretty common error when working with `pdfjs` -> https://github.com/mozilla/pdf.js/issues/10478#issuecomment-518673665


jacoblee93:
Thank you!

PR #3011, opened 2023-10-23
Feat/parent doc score threshold

### LINKED ISSUE #3006: Pls. add similarity score threshold search type to Parent Document Retriever
I'd like to get back all the parent documents of ALL child documents above a given similarity score threshold.
AFAIK we can't do it, at least out of the box.

bracesproul:
Closes #3006

### DIFF
--- a/examples/src/retrievers/parent_document_retriever_score_threshold.ts
+++ b/examples/src/retrievers/parent_document_retriever_score_threshold.ts
@@ -8,7 +8,7 @@ import { ScoreThresholdRetriever } from "langchain/retrievers/score_threshold";
 
 const vectorstore = new MemoryVectorStore(new OpenAIEmbeddings());
 const docstore = new InMemoryStore();
-const scoreThresholdRetriever = ScoreThresholdRetriever.fromVectorStore(
+const childDocumentRetriever = ScoreThresholdRetriever.fromVectorStore(
   vectorstore,
   {
     minSimilarityScore: 0.01, // Essentially no threshold
@@ -18,7 +18,7 @@ const scoreThresholdRetriever = ScoreThresholdRetriever.fromVectorStore(
 const retriever = new ParentDocumentRetriever({
   vectorstore,
   docstore,
-  retriever: scoreThresholdRetriever,
+  childDocumentRetriever,
   // Optional, not required if you're already passing in split documents
   parentSplitter: new RecursiveCharacterTextSplitter({
     chunkOverlap: 0,
--- a/examples/src/retrievers/parent_document_retriever_score_threshold.ts
+++ b/examples/src/retrievers/parent_document_retriever_score_threshold.ts
@@ -4,12 +4,21 @@ import { InMemoryStore } from "langchain/storage/in_memory";
 import { ParentDocumentRetriever } from "langchain/retrievers/parent_document";
 import { RecursiveCharacterTextSplitter } from "langchain/text_splitter";
 import { TextLoader } from "langchain/document_loaders/fs/text";
+import { ScoreThresholdRetriever } from "langchain/retrievers/score_threshold";
 
 const vectorstore = new MemoryVectorStore(new OpenAIEmbeddings());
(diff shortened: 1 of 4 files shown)


langchain/src/retrievers/parent_document.ts
```diff
@@ -45,6 +55,14 @@ export class ParentDocumentRetriever extends MultiVectorRetriever {
 
   protected parentK?: number;
 
+  protected retriever: ScoreThresholdRetriever<VectorStore> | undefined;
```
jacoblee93, line 58: Should we instead allow an arbitrary retriever to be passed here that overrides the default? Could imagine MMR etc. being useful

jacoblee93 (commented):
See comment

langchain/src/retrievers/parent_document.ts
bracesproul: Yeah agreed

Should probably update how the options for the retriever are passed in too

Working on it now

bracesproul: Updated code & docs

examples/src/retrievers/parent_document_retriever_score_threshold.ts
```diff
@@ -0,0 +1,51 @@
...
+  vectorstore,
+  {
+    minSimilarityScore: 0.01, // Essentially no threshold
+    maxK: 1, // Only return the top result
+  }
+);
+const retriever = new ParentDocumentRetriever({
+  vectorstore,
+  docstore,
+  retriever: scoreThresholdRetriever,
```
jacoblee93, line 21: nit: how about `childDocumentRetriever` for clarity?

langchain/src/retrievers/parent_document.ts
```diff
@@ -54,10 +61,18 @@ export class ParentDocumentRetriever extends MultiVectorRetriever {
...
     this.parentK = fields.parentK;
+    this.retriever = fields.retriever;
   }
 
   async _getRelevantDocuments(query: string): Promise<Document[]> {
-    const subDocs = await this.vectorstore.similaritySearch(query, this.childK);
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    let subDocs: Document<Record<string, any>>[] = [];
+    if (this.retriever) {
+      subDocs = await this.retriever.getRelevantDocuments(query);
```
jacoblee93, line 71: Pass `childK`?

langchain/src/retrievers/parent_document.ts
```diff
@@ -129,7 +145,11 @@ export class ParentDocumentRetriever extends MultiVectorRetriever {
       embeddedDocs.push(...taggedSubDocs);
       fullDocs[parentDocId] = parentDoc;
     }
-    await this.vectorstore.addDocuments(embeddedDocs);
+    if (this.retriever) {
+      await this.retriever.addDocuments(embeddedDocs);
```
jacoblee93, line 149: I don't think this is a standard method

jacoblee93 (dismissed):
A few more comments

langchain/src/retrievers/parent_document.ts
bracesproul: No param for that in `getRelevantDocuments` instead you'd pass it when you initialize the class outside of `ParentDocumentRetriever`

ex:
```typescript
const scoreThresholdRetriever = ScoreThresholdRetriever.fromVectorStore(
  vectorstore,
  {
    ...,
    maxK: 1,
  }
);
```

langchain/src/retrievers/parent_document.ts
bracesproul: If it's a class extending from `VectorStoreRetriever` then it will have it.

langchain/src/retrievers/parent_document.ts
jacoblee93: Oh, right thanks

jacoblee93:
Looks good, thank you!

PR #3013, opened 2023-10-23
Feat: generic file system kv store

bracesproul:


langchain/src/storage/file_system.ts
```diff
@@ -0,0 +1,146 @@
...
+      }
+      values = JSON.parse(fileContent);
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } catch (e: any) {
+      // If the file does not exist, create it
+      if (("code" in e && e.code === "EISDIR") || e.code === "ENOENT") {
+        await fsPromises.writeFile(this.path, "", "utf-8");
+        return {} as Record<string, T>;
+      }
+      throw new Error(`Error reading and parsing file at path: ${this.path}`);
```
jacoblee93, line 44: Shouldn't we have something about the actual error in the error message?

langchain/src/storage/file_system.ts
```diff
@@ -0,0 +1,155 @@
...
+  private async getParsedFile(key: string): Promise<Uint8Array | undefined> {
+    try {
+      const fileContent = await fsPromises.readFile(this.getFullPath(key));
+      return fileContent;
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } catch (e: any) {
+      // If the file does not exist, create it
+      if (("code" in e && e.code === "EISDIR") || e.code === "ENOENT") {
+        await fsPromises.writeFile(this.getFullPath(key), "");
+        return undefined;
```
jacoblee93, line 33: This will return `undefined` the first time and then empty string the second time won't it?

bracesproul: Ah yeah that's right since I removed the return undefined check

bracesproul: Fixed

langchain/src/storage/tests/file_system.test.ts
```diff
@@ -0,0 +1,97 @@
+/* eslint-disable no-process-env */
+
+import { test } from "@jest/globals";
+import fsPromises from "fs/promises";
+import { LocalFileStore } from "../file_system.js";
+
+describe("LocalFileStore", () => {
+  const keys = ["key1", "key2"];
+  const rootPath = "./file_system_store_test";
```
jacoblee93, line 9: Use `os.tempdir` here

https://github.com/langchain-ai/langchainjs/blob/4e35156fdf709ae22a85a79c1ce882b4e1ef4bd5/langchain/src/document_loaders/web/azure_blob_storage_file.ts#L64

jacoblee93 (changes requested):

jacoblee93 (approved):

jacoblee93:
Thanks!

drewB:
May I ask what the purpose of this store is?  I tried using it with MultiVectorRetriever it is in incompatiable because it is declared as `BaseStore<string, Uint8Array>` and MultiVectorRetriever requires  `BaseStore<string, Document>`

jacoblee93:
There is a wrapper class called `EncoderBackedStore` to make this usable there - we need docs on it though CC @bracesproul

drewB:
Thanks.  So looking at that class, looks like I would want to do something like:

```typescript
const docstore = createDocumentStoreFromByteStore(new LocalFileStore({rootPath: localFileStoreDirectory}))

const retriever = new MultiVectorRetriever({
        vectorstore: vectorStore,
        docstore,
        idKey,
});
```

Correct?

jacoblee93:
Yes that looks correct!

PR #3014, opened 2023-10-23
Ensure shell expansions are quoted in docs workflows.

obi1kenobi:
Two improvements:
- Ensure that shell command arguments that use env var expansion are always quoted if they are intended to be one argument. Otherwise they may expand to multiple arguments, and become a vector for shell injection attacks.
- Avoid using GitHub Actions templating (`${{ ... }}`) in shell commands, and use env vars instead. Such templating cannot be escaped against injection attacks since it runs before the shell sees the command, so it's always better to stick the templated data in an env var and use shell expansion instead.

I don't believe the current workflows were exploitable as currently written, but security is hard and it's always better to be safe than sorry.

### DIFF
--- a/.github/workflows/docs-preview.yml
+++ b/.github/workflows/docs-preview.yml
@@ -29,10 +29,10 @@ jobs:
       - name: Install Vercel CLI
         run: npm install --global vercel@latest
       - name: Pull Vercel Environment Information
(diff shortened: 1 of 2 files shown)


jacoblee93:
Got it, thanks!

PR #3020, opened 2023-10-24
feat: added react output parser and tests

bracesproul:
Moved the `formatForOpenAIFunctions` function from
`/agents/format_scratchpad` --> `/agents/format_scratchpad/openai_functions` (match py)
but entry point is still `agents/format_scratchpad` to satisfy backwards compatibility

added format to xml, string, and base message utils


jacoblee93:
Thanks! And still good practice to keep backcompat for anything we have documented, so let's keep

langchain/scripts/create-entrypoints.js
```diff
@@ -16,7 +16,11 @@ const entrypoints = {
   "agents/toolkits": "agents/toolkits/index",
   "agents/toolkits/aws_sfn": "agents/toolkits/aws_sfn",
   "agents/toolkits/sql": "agents/toolkits/sql/index",
-  "agents/format_scratchpad": "agents/format_scratchpad",
+  "agents/format_scratchpad": "agents/format_scratchpad/openai_functions",
+  "agents/format_scratchpad/log": "agents/format_scratchpad/log",
+  "agents/format_scratchpad/xml": "agents/format_scratchpad/xml",
+  "agents/format_scratchpad/log_to_message": "agents/format_scratchpad/log_to_message",
+  "agents/react/outputParser": "agents/react/outputParser",
```
jacoblee93, line 23: We don't camel case entrypoints, use snake case

jacoblee93 (dismissed):
Update the entrypoint

jacoblee93:
Otherwise looks good, thanks!

PR #3024, opened 2023-10-24
Avoid hardcoding embedding size to OpenAI dimensions in Qdrant client

jacoblee93:
Also improves thrown error messages


bracesproul (approved):
lgtm, thanks for this!

PR #3026, opened 2023-10-24
feat: add chat_model iflytek xinghuo

### LINKED ISSUE #2926: feat: add chat model for iflytek xinghuo llm
### feat: add chat model for iflytek xinghuo llm
iflytek xinghuo llm is popular as same as baidu in China. Can I add a new chat model named ChatIflytekXinghuo. 
iflytek xinghuo official site: https://xinghuo.xfyun.cn

AIXGEEK:
Fixes #2926 (issue)

iflytek xinghuo llm is popular as same as baidu in China. Can I add a new chat model named ChatIflytekXinghuo.
iflytek xinghuo official site: https://xinghuo.xfyun.cn/.


langchain/scripts/create-entrypoints.js
bracesproul, line 1: you also need to add the entry-point to the `requiresOptionalDependency` object since it requires `ws`

xinghuo-assistant: yes, i will add it.

AIXGEEK: ok, i will add it. thanks!

langchain/src/util/websocket_stream.ts
```diff
@@ -0,0 +1,101 @@
+import WsWebSocket from "ws";
```
jacoblee93, line 1: I like this but can we leave it inside the `iflytek` class for now?

langchain/src/util/websocket_stream.ts
```diff
@@ -0,0 +1,101 @@
...
+    if (options.signal?.aborted) {
+      throw new DOMException("This operation was aborted", "AbortError");
+    }
+
+    this.url = url;
+
+    let ws: WebSocket;
+
+    if (getEnv() === "browser") {
+      ws = new WebSocket(url, options.protocols ?? []);
```
jacoblee93, line 49: I unfortunately think even the presence of the import statement will break most bundlers - might be nice to find some compatible package? Or just mark this as `Node` only

jacoblee93: We can also have separate entrypoints for web vs. Node and factor out common logic:

https://github.com/langchain-ai/langchainjs/tree/main/langchain/src/chat_models/googlevertexai

langchain/src/chat_models/iflytek_xinghuo.ts
```diff
@@ -0,0 +1,475 @@
...
+    const messagesMapped: XinghuoMessage[] = messages.map((message) => ({
+      role: messageToXinghuoRole(message),
+      content: message.content,
+    }));
+    const data = await new Promise<ChatCompletionResponse>(
+      (resolve, reject) => {
+        let response: ChatCompletionResponse;
+        let rejected = false;
+        let resolved = false;
+        void this.completion(
```
jacoblee93, line 374: This `void` needs a linter ignore line

jacoblee93: And actually, it would be better to refactor `this.completion` to either not return a promise or be an iterator

jacoblee93 (changes requested):
Thank you! Looks good overall, see comments

langchain/src/util/websocket_stream.ts
AIXGEEK: yes, I will change this name. WebsocketStream is a explainer to chrome, but is not launch. I think WebSocketStream is a good design, I do it base on these:
[websocketstream-explainer](https://github.com/ricea/websocketstream-explainer/blob/master/README.md)
[chrome developer current status](https://developer.chrome.com/articles/websocketstream/)

langchain/src/util/websocket_stream.ts
AIXGEEK: yes, i make two entry for xinghuo chat model in new commit. It's a good idea.

langchain/src/chat_models/iflytek_xinghuo.ts
AIXGEEK: I hava refactored completion method to either not return a promise or be an iterator.

AIXGEEK:
@jacoblee93 I have read these comment, and fix related code. Thanks!

langchain/src/chat_models/iflytek_xinghuo/common.ts
```diff
@@ -0,0 +1,479 @@
...
+        case "v2.1":
+          this.domain = "generalv2";
+          break;
+        case "v3.1":
+          this.domain = "generalv3";
+          break;
+        default:
+          this.domain = "generalv2";
+      }
+      this.apiUrl = `wss://spark-api.xf-yun.com/${this.version}/chat`;
```
jacoblee93, line 275: Do you have a source for this API URL?

jacoblee93:
Thanks for your patience and the refactor - one more question!

langchain/src/chat_models/iflytek_xinghuo/common.ts
AIXGEEK: yes, I have an official document site. But it's writed by chinese. The english document is not updated.
the site: https://www.xfyun.cn/doc/spark/Web.html

image

jacoblee93:
Ok thanks! We'll also want to add a docs page, but we can make that a followup!

jacoblee93 (approved):

jacoblee93:
Thank you!

PR #3028, opened 2023-10-24
Respect organization in OpenAIEmbeddings

jacoblee93:


bracesproul (approved):
lgtm, only comment is we should add a note somewhere in the docs about the default env var `OPENAI_ORGANIZATION`

jacoblee93:
> lgtm, only comment is we should add a note somewhere in the docs about the default env var `OPENAI_ORGANIZATION`

Rushing this out, will follow up

PR #3029, opened 2023-10-24
feat: added openai functions output parser class

bracesproul:
Moved the output parsing function to a new class `OpenAIFunctionsAgentOutputParser` to match py
Added an entry-point for this class so it can be accessed externally (for use in openai function runnables, etc)


bracesproul:
cc @jacoblee93, just a refactor so it can be exported in a class

langchain/src/agents/openai/output_parser.ts
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * Type that represents an agent action with an optional message log.
+ */
+export type FunctionsAgentAction = AgentAction & {
+  messageLog?: BaseMessage[];
+};
+
+export class OpenAIFunctionsAgentOutputParser extends AgentActionOutputParser {
+  lc_namespace = ["langchain", "agents", "openai"];
```
jacoblee93, line 21: Oh, for all of these we will want to add `static lc_name` to handle minification:

https://github.com/langchain-ai/langchainjs/blob/main/langchain/src/chat_models/anthropic.ts#L151

langchain/src/agents/openai/output_parser.ts
```diff
@@ -0,0 +1,83 @@
...
+      `OpenAIFunctionsAgentOutputParser can only parse messages.\nPassed input: ${text}`
+    );
+  }
+
+  async parseResult(generations: Generation[] | ChatGeneration[]) {
+    if ("message" in generations[0] && isBaseMessage(generations[0].message)) {
+      return this.parseAIMessage(generations[0].message);
+    }
+    throw new Error(
+      "parseResult on OpenAIFunctionsAgentOutputParser only works on ChatGeneration output"
```
jacoblee93, line 38: Why allow both as input here?

langchain/src/agents/openai/output_parser.ts
```diff
@@ -0,0 +1,83 @@
...
+    );
+  }
+
+  /**
+   * Parses the output message into a FunctionsAgentAction or AgentFinish
+   * object.
+   * @param message The BaseMessage to parse.
+   * @returns A FunctionsAgentAction or AgentFinish object.
+   */
+  parseAIMessage(message: BaseMessage): FunctionsAgentAction | AgentFinish {
```
jacoblee93, line 48: Fine to have separate but could we just put this in `parseResult`?

jacoblee93 (changes requested):
A few questions

langchain/src/agents/openai/output_parser.ts
bracesproul: Oversight on my part, I was just copypasta-ing the base class inputs.

langchain/src/agents/openai/output_parser.ts
bracesproul: Yeah, refactoring to do this, had two because that's what they have in py

bracesproul: Ah never mind, need this because the message response from the `_call` method in openai functions isn't of type `ChatGeneration`

langchain/src/agents/openai/index.ts
```diff
@@ -253,6 +218,6 @@ export class OpenAIAgent extends Agent {
       valuesForLLM,
       callbackManager
     );
-    return parseOutput(message);
+    return this.outputParser.parseAIMessage(message);
```
jacoblee93, line 221: Use the standard `parseResult` method here unless there's a reason not to?

langchain/src/agents/openai/output_parser.ts
```diff
@@ -0,0 +1,86 @@
...
+  static lc_name() {
+    return "OpenAIFunctionsAgentOutputParser";
+  }
+
+  constructor() {
+    super(...arguments);
+  }
+
+  async parse(text: string): Promise<AgentAction | AgentFinish> {
+    throw new Error(
```
jacoblee93, line 31: This is odd?

PR #3027, opened 2023-10-24
Feat: runnable examples in agent how to docs

bracesproul:
- [x] Conversational
- [x] OpenAI functions
- [ ] ~~Plan and Execute~~
- [x] ReAct
- [x] Structured tool chat
- [x] XML Agent
- [x] General LCEL example

PR #3036, opened 2023-10-24
Fix output parser regex

cwoolum:
When using `AnthropicFunctions`, Claude always returns the entire context. This means we get responses like

```
Question: What is my weight in pounds?

Thought: The user stated their weight as 11.4 stones. I need to convert this to pounds.

Action:
```json
{
 "action": "determine_weight", 
 "action_input": "159.6"
}```

Observation: I have now set the user's weight to 159.6 pounds based on their input of 11.4 stones.

Thought: I now know the user's weight in pounds.

Action:
```json 
{
 "action": "Final Answer",
 "action_input": "Your weight in pounds is 159.6."
}```
```

we get all of the content from the first 3 backticks to the very last instance. This causes parsing errors. 

What we really want is to always grab the last instance of a JSON response along with any additionally nested code snippets that are not JSON.

For a response like 

```
 'Here we have some boilerplate nonsense```json
{
 \t\r
"action": "blogpost",
\t\r  "action_input": "```sql
SELECT * FROM orders
JOIN users ON users.id = orders.user_id
WHERE users.email = \'bud\'```"
\t\r}


\t\r``` and at the end there is more nonsense'
```

we actually also want to include the center content with the sql snippet.

One of the assumptions I'm making here is that functions will always have the code to be executed inside of triple backticks with `json` right after. Another set of JSON backticks cannot be nested within that.

### DIFF
--- a/langchain/src/agents/structured_chat/outputParser.ts
+++ b/langchain/src/agents/structured_chat/outputParser.ts
@@ -36,23 +36,28 @@ export class StructuredChatOutputParser extends AgentActionOutputParser {
    */
   async parse(text: string): Promise<AgentAction | AgentFinish> {
     try {
-      const regex = /```(?:json)?(.*)(```)/gs;
+      const regex = /.*```json\s*(?<response>[\s\S]*?)```\s*(?![\s\S]*```)/s;
       const actionMatch = regex.exec(text);
-      if (actionMatch === null) {
+
+      // Let's shortcut early here. If we weren't able to extract `response`,
+      // there's no point in continuing.
+      if (actionMatch?.groups?.response === undefined) {
         throw new OutputParserException(
           `Could not parse an action. The agent action must be within a markdown code block, and "action" must be a provided tool or "Final Answer"`
         );
       }
-      const response = JSON.parse(actionMatch[1].trim());
+
+      const response = JSON.parse(actionMatch.groups.response.trim());
       const { action, action_input } = response;
 
       if (action === "Final Answer") {
         return { returnValues: { output: action_input }, log: text };
       }
+
       return { tool: action, toolInput: action_input || {}, log: text };
     } catch (e) {
       throw new OutputParserException(
-        `Failed to parse. Text: "${text}". Error: ${e}`
+        `Failed to parse. Text: \n\n${text}\n\n Error: ${e}`
       );
     }
   }
--- a/langchain/src/agents/tests/structured_chat_output_parser.test.ts
+++ b/langchain/src/agents/tests/structured_chat_output_parser.test.ts
@@ -42,3 +42,42 @@ test("Can parse JSON with text in front of it", async () => {
     }
   }
 });
+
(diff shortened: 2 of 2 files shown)


langchain/src/agents/structured_chat/outputParser.ts
```diff
@@ -36,23 +36,28 @@ export class StructuredChatOutputParser extends AgentActionOutputParser {
...
   async parse(text: string): Promise<AgentAction | AgentFinish> {
     try {
-      const regex = /```(?:json)?(.*)(```)/gs;
+      const regex = /.*```json\s*(?<response>[\s\S]*?)```\s*(?![\s\S]*```)/s;
       const actionMatch = regex.exec(text);
-      if (actionMatch === null) {
+
+      // Let's shortcut early here. If we weren't able to extract `response`,
+      // there's no point in continuing.
+      if (actionMatch?.groups?.response === undefined) {
```
cwoolum, line 44: Converted this to use Regex groups to be more explicit.

langchain/src/agents/structured_chat/outputParser.ts
```diff
@@ -36,23 +36,28 @@ export class StructuredChatOutputParser extends AgentActionOutputParser {
...
 
       if (action === "Final Answer") {
         return { returnValues: { output: action_input }, log: text };
       }
+
       return { tool: action, toolInput: action_input || {}, log: text };
     } catch (e) {
       throw new OutputParserException(
-        `Failed to parse. Text: "${text}". Error: ${e}`
+        `Failed to parse. Text: \n\n${text}\n\n Error: ${e}`
```
cwoolum, line 60: Using newlines to make the failed text readable.

langchain/src/agents/tests/structured_chat_output_parser.test.ts
```diff
@@ -42,3 +42,42 @@ test("Can parse JSON with text in front of it", async () => {
     }
   }
 });
+
+test("Should throw an output parser exception if no JSON section is found", async () => {
```
cwoolum, line 46: Added a test for the failure case

cwoolum:
This change could potentially have a high blast radius. Am I missing any happy path cases?

jacoblee93:
@cwoolum we have #3027 that should land today showing how to use expression language and custom output parsers for agents - would love to add this as an AnthropicFunctions example rather than overwriting the base class?

cwoolum:
If I'm understanding the PR correctly, it uses user defined stop works to mark the response rather than a predefined regex?

I'm not sure the custom parser would fix this. The problem here for Anthropic is that the response includes the outputs for the entire conversation. This confused the parser because it didn't know which output to select. This could be mitigated by having the user specify a regex to parse this correctly but these regexes can be very complex and may be difficult to implement.

Let me know if I'm misunderstanding the functionality.

jacoblee93:
All that you changed here is the output parser, so using a runnable with your own defined output parser would be the same thing no?

PR #3040, opened 2023-10-25
Make "IN-filters" possible for Vercel Postgres Metadata

rbouschery:
This change enables IN-filters when doing a similarity search on the Vercel Postgres Vectorstore. 

This makes retrieving documents based on metadata filters much more flexible. Currently, only a single value can be checked, this allows for multiple values to be checked against a metadata property.

As for example this Feature Request discussion on the Python package shows, this seems to be possible without being documented in the Python package: https://github.com/langchain-ai/langchain/issues/9726


jacoblee93:
Nice! Will test later and merge in the next release.

langchain/src/vectorstores/vercel_postgres.ts
```diff
@@ -247,19 +247,42 @@ export class VercelPostgres extends VectorStore {
...
   ): Promise<[Document, number][]> {
     const embeddingString = `[${query.join(",")}]`;
-    const _filter = filter ?? "{}";
+    const _filter = filter ? filter : {};
+    const whereClauses = [];
+    const values = [embeddingString, k];
+    let paramCount = 2;
+
+    for (const [key, value] of Object.entries(_filter)) {
+      if (value.hasOwnProperty("in")) {
```
jacoblee93, line 256: So this would look like 
```
{
  "key": {
    "in": ["item", "item2"]
}
```
right?

Can you add an integration test?

jacoblee93 (changes requested):
I can do the test + docs tomorrow as well if you don't have time!

langchain/src/vectorstores/vercel_postgres.ts
rbouschery: Yep - that's right! And happy to add tests and docs, will add this later today!

jacoblee93: Sure, will do now!

jacoblee93 (approved):

jacoblee93:
This is excellent, thank you!

PR #3043, opened 2023-10-25
Use prepared statements + add concurrency for Cassandra

mieslep:
Inserts now use prepared statements. Batch size is now tunable, as is the number of concurrent requests that can be sent to Cassandra.


langchain/package.json
```diff
@@ -935,7 +935,7 @@
     "apify-client": "^2.7.1",
```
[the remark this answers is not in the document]
mieslep: The 4.7 driver introduced support for the native vector type. https://github.com/datastax/nodejs-driver/blob/master/CHANGELOG.md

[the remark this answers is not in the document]
mieslep: The 4.7 driver introduced support for the native vector type. https://github.com/datastax/nodejs-driver/blob/master/CHANGELOG.md

langchain/src/vectorstores/cassandra.ts
```diff
@@ -21,7 +21,54 @@ export interface CassandraLibArgs extends DseClientOptions {
...
+  maxConcurrency?: number;
+  batchSize?: number;
+}
+
+/**
+ * Class for managing concurrent tasks, similar to PromisePool (which is
+ * not part of the project's dependencies, and would be preferable to
+ * having this code).
+ */
+class ConcurrencyManager {
```
jacoblee93, line 34: We can use the `AsyncCaller` class which I believe would be similar?

https://github.com/langchain-ai/langchainjs/blob/main/langchain/src/embeddings/openai.ts#L246

langchain/src/vectorstores/cassandra.ts
```diff
@@ -63,7 +114,19 @@ export class CassandraStore extends VectorStore {
     this.table = args.table;
     this.primaryKey = args.primaryKey;
     this.metadataColumns = args.metadataColumns;
-    this.indices = args.indices;
+    this.indices = args.indices || [];
+    this.maxConcurrency = Math.floor(args.maxConcurrency || 25);
```
jacoblee93, line 118: If we use `AsyncCaller` as suggested above, we can just have the input extend `AsyncCallerParams`: https://github.com/langchain-ai/langchainjs/blob/main/langchain/src/embeddings/base.ts#L7

jacoblee93 (changes requested):
Thank you for this! We have a built-in concurrency utility you can use for consistency.

langchain/src/vectorstores/cassandra.ts
mieslep: done.

mieslep: done.

mieslep:
@jacoblee93 I've switched over to AsyncCaller, thanks for the pointers to that code.

jacoblee93:
LGTM if it's good with you!

jacoblee93 (approved):

PR #3049, opened 2023-10-25
Fix: bug where pgvector cant add more than chunk size

### LINKED ISSUE #3045: PGVector's `addDocuments` fails if number of documents > chunkSize
PGVector's `addDocuments` fails if `documents.length` is bigger than [`chunkSize`](https://github.com/langchain-ai/langchainjs/blob/4f56cf7ceb7f1abfc3f903c9d8a9bdb0c3d5b176/langchain/src/vectorstores/pgvector.ts#L172C11-L172C20).

The PG error is: `error: could not determine data type of parameter $1`

To reproduce:

```js
import { OpenAIEmbeddings } from 'langchain/embeddings/openai'
import { PGVectorStore } from 'langchain/vectorstores/pgvector'

const connectionString = process.env.DATABASE_URL
const config = {
  postgresConnectionOptions: connectionString,
  tableName: 'testlangchain',
  columns: {
    idColumnName: 'id',
    vectorColumnName: 'vector',
    contentColumnName: 'content',
    metadataColumnName: 'metadata',
  },
}

const embeddings = new OpenAIEmbeddings()
const pgvectorStore = await PGVectorStore.initialize(embeddings, config)

try {
  const docsToGenerate = 501
  const documents = []
  for (let i = 1; i <= docsToGenerate; i++) {
    documents.push({ pageContent: 'Lorem Ipsum', metadata: { a: i } })
  }

  await pgvectorStore.addDocuments(documents)
  await pgvectorStore.end()
} catch (e) {
  console.log(e.message)
}
```

Note `docsToGenerate` is set to 501, which is bigger than the default [`chunkSize`](https://github.com/langchain-ai/langchainjs/blob/4f56cf7ceb7f1abfc3f903c9d8a9bdb0c3d5b176/langchain/src/vectorstores/pgvector.ts#L172C11-L172C20). If changed to a value smaller than 500, the operation is successful.

bracesproul:
Closes #3045

fix bug where when you try to add more docs than the default chunk size (also made configurable in this PR) it failed
added a test to confirm.

### DIFF
--- a/langchain/src/vectorstores/pgvector.ts
+++ b/langchain/src/vectorstores/pgvector.ts
@@ -22,6 +22,12 @@ export interface PGVectorStoreArgs {
   };
   filter?: Metadata;
   verbose?: boolean;
+  /**
+   * The amount of documents to chunk by when
+   * adding vectors.
+   * @default 500
+   */
+  chunkSize?: number;
 }
 
 /**
@@ -51,6 +57,8 @@ export class PGVectorStore extends VectorStore {
 
   client?: PoolClient;
 
+  chunkSize = 500;
+
   _vectorstoreType(): string {
     return "pgvector";
   }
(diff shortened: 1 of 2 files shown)


jacoblee93:
Thank you!

PR #3055, opened 2023-10-26
Feat/step back notebook

bracesproul:


cookbook/step_back.ipynb
```diff
@@ -0,0 +1,308 @@
...
+    "  {\n",
+    "    input: \"Jan Sindel's was born in what country?\",\n",
+    "    output: \"what is Jan Sindel's personal history?\",\n",
+    "  },\n",
+    "];\n",
+    "const examplePrompt = ChatPromptTemplate.fromMessages([\n",
+    "  [\"human\", \"{input}\"],\n",
+    "  [\"ai\", \"{output}\"],\n",
+    "]);\n",
+    "const fewShotPrompt = new FewShotChatMessagePromptTemplate({\n",
```
jacoblee93, line 59: Can we use the `.fromExamples` initializer here or does that not make sense?

PR #3057, opened 2023-10-26
fix: LANGCHAIN_VERBOSE=false does not turn off verbose logging.

xianminx:
Fixes # LANGCHAIN_VERBOSE=false does not turn off verbose logging.


jacoblee93:
Nice, thank you!

jacoblee93 (approved):

PR #3065, opened 2023-10-26
feat: openai functions structured output with memory docs

bracesproul:


docs/docs/modules/chains/popular/structured_output.mdx
```diff
@@ -34,6 +34,170 @@ import FormatExample from "@examples/chains/openai_functions_structured_format.t
...
+
+* `food_recorder` -> A function to record what food a user likes.
+* `final_response` -> The function that returns a response to the user
+
+After each LLM response, we'll take our structured data and save it to memory (with a little formatting too). By doing this, we're enabling our LLM to
+remember what the user has said and use that information to generate a response.
+
+The first step is to create our functions. For this we'll use `zod` to create the function schemas.
+
+We're going to create two, one which returns an array of foods, with their name, whether or not they're healthy and their color.
```
bracesproul, line 50:
```suggestion
We're going to create two, one which returns an array of foods with their name, whether or not they're healthy and their color.
```

docs/docs/modules/chains/popular/structured_output.mdx
```diff
@@ -34,6 +34,170 @@ import FormatExample from "@examples/chains/openai_functions_structured_format.t
...
+    )
+    .describe("An array of food items mentioned in the text"),
+});
+
+const finalResponseZodSchema = z.object({
+  finalResponse: z.string().describe("The final response"),
+});
+```
+
+Next we'll construct the prompt. This is simple, we have two "messages", one from the AI with a simple explanation on when to use the functions, and an input variable for the chat history. The second only contains the users question.
```
bracesproul, line 71:
```suggestion
Next we'll construct the prompt. We have two "messages", one from the AI with a simple explanation on when to use the functions, and an input variable for the chat history. The second only contains the users question.
```

docs/docs/modules/chains/popular/structured_output.mdx
```diff
@@ -34,6 +34,170 @@ import FormatExample from "@examples/chains/openai_functions_structured_format.t
...
+    "ai",
+    `You are a helpful assistant.
+You are provided with two functions: one to use to record what food the user likes, and one to use when replying to user questions.
+Chat history: {history}`,
+  ],
+  ["human", "Question: {input}"],
+]);
+```
+
+Then, define the LLM model and bind the function arguments to it. By using `.bind()` we're adding the functions to the model for every call.
```
bracesproul, line 85:
```suggestion
Then, define the LLM model and bind the function arguments to it. By using `.bind()` we're adding the functions to the model for every call.

We also must wrap our `zod` schemas with `zodToJsonSchema` to appropriately format them for OpenAI.
```

docs/docs/modules/chains/popular/structured_output.mdx
```diff
@@ -34,6 +34,169 @@ import FormatExample from "@examples/chains/openai_functions_structured_format.t
...
+    { name: 'oxygen', healthy: true, color: 'transparent' },
+    { name: 'french fries', healthy: false, color: 'golden' }
+  ]
+}
+ */
+
+const response2 = await runnableChain.invoke({
+  input: "What food do I like?",
+});
+
```
bracesproul, line 190:
```suggestion
console.log(response);
/**
{
  foods: [
    { name: 'apples', healthy: true, color: 'red' },
    { name: 'bananas', healthy: true, color: 'yellow' },
    { name: 'oxygen', healthy: true, color: 'transparent' },
    { name: 'french fries', healthy: false, color: 'golden' }
  ]
}
 */
const response2 = await runnableChain.invoke({
  input: "What food do I like?",
});
```

docs/docs/modules/chains/popular/structured_output.mdx
```diff
@@ -34,6 +34,167 @@ import FormatExample from "@examples/chains/openai_functions_structured_format.t
...
+    {
+      name: "final_response",
+      description: "The function that returns a response to the user",
+      parameters: zodToJsonSchema(finalResponseZodSchema),
+    },
+  ],
+});
+```
+
+Next we can define our memory using the `BufferMemory` class, and the output parser for structured data: `JsonOutputFunctionsParser`.
```
jacoblee93, line 108: We should definitely add docs for `JsonOutputFunctionsParser`

docs/docs/modules/chains/popular/structured_output.mdx
```diff
@@ -34,6 +34,167 @@ import FormatExample from "@examples/chains/openai_functions_structured_format.t
...
+  memoryKey: "history", // The object key to store the memory under
+  inputKey: "question", // The object key for the input
+  outputKey: "answer", // The object key for the output
+});
+```
+
+Now for the most important step, we'll create a custom function which handles calling the LLM, saving the memory and returning the response.
+
+```typescript
+const executeQuery = async (input: { input: string; history: string }) => {
```
jacoblee93, line 123: This will unfortunately not trace well :(

We would want to do it somewhat like this: https://js.langchain.com/docs/expression_language/cookbook/adding_memory

Also, OpenAI has a special `FunctionMessage` category that might be better?

jacoblee93 (commented):
This is fine-ish but not sure how much it makes sense to chat with an extraction chain like this/what the use-case would be vs. showing how to do this with the OAI functions agent?

Would rather spend time organizing/updating the memory docs in general

PR #3066, opened 2023-10-27
feat: map reduce chain w/ LCEL

bracesproul:


examples/src/chains/map_reduce_lcel.ts
```diff
@@ -0,0 +1,163 @@
...
+const combinePrompt = PromptTemplate.fromTemplate(
+  "Combine these summaries:\n\n{context}"
+);
+
+// Define a util function to allow function invocation with preset arguments
+function partial<F extends (...args: any[]) => any>(
+  fn: F,
+  ...partialArgs: any[]
+): (...args: any[]) => ReturnType<F> {
+  return (...args: any[]): ReturnType<F> => fn(...partialArgs, ...args);
```
jacoblee93, line 38: Can't we just use `.bind`?

examples/src/chains/map_reduce_lcel.ts
```diff
@@ -0,0 +1,163 @@
...
+  tokenMax = 4000
+) => {
+  const editableConfig = config;
+  let docs = documents;
+  let collapseCount = 1;
+  while ((await getNumTokens(docs)) > tokenMax) {
+    if (editableConfig) {
+      editableConfig.runName = `Collapse ${collapseCount}`;
+    }
+    const invoke = partial(collapseChain.invoke);
```
jacoblee93, line 94: Partialing with no args doesn't make sense?

examples/src/chains/map_reduce_lcel.ts
```diff
@@ -0,0 +1,163 @@
...
+// Define the map chain to format, summarize, and parse the document
+const mapChain = RunnableSequence.from([
+  { context: async (i: Document) => formatDocument(i, documentPrompt) },
+  summarizePrompt,
+  model,
+  outputParser,
+]);
+
+// Define a wrapper chain to keep the original metadata while summarizing the document
+const mapAsDocChain = RunnableSequence.from([
```
jacoblee93, line 65: Can we just work this into `mapChain` to make it a bit easier to componentize?

examples/src/chains/map_reduce_lcel.ts
```diff
@@ -0,0 +1,163 @@
...
+  model,
+  outputParser,
+]).withConfig({ runName: "Reduce" });
+
+// Define the final map-reduce chain
+const mapReduceChain = RunnableSequence.from([
+  mapAsDocChain.map(),
+  collapse,
+  reduceChain,
+]).withConfig({ runName: "Map reduce" });
```
jacoblee93, line 117: I would add something in the docs about the flow:

```
To show how this can be implemented in Expression Language, we break the map-reduce chain down into three steps:

1. Summarization of each document
2. Collapsing
3. Reducing
```

But also, don't we only summarize if the docs are smaller than the original context window?

jacoblee93 (dismissed):
See comments

examples/src/chains/map_reduce_lcel.ts
bracesproul: Was able to refactor and scrap this util

examples/src/chains/map_reduce_lcel.ts
bracesproul: Yes, however I think what I had before was easier to comprehend.
New:
```typescript
// Define the final map-reduce chain
const mapReduceChain = RunnableSequence.from([
  RunnableSequence.from([
    RunnableMap.from({ doc: new RunnablePassthrough(), content: mapChain }),
    RunnableLambda.from(
      (input) =>
        new Document({ pageContent: input.content, metadata: input.doc.metadata })
    ),
  ]).withConfig({ runName: "Summarize (return doc)" }).map(),
  collapse,
  reduceChain,
]).withConfig({ runName: "Map reduce" });
```

Old:
```typescript
// Define a wrapper chain to keep the original metadata while summarizing the document
const mapAsDocChain = RunnableSequence.from([
  RunnableMap.from({ doc: new RunnablePassthrough(), content: mapChain }),
  RunnableLambda.from(
    (input) =>
      new Document({ pageContent: input.content, metadata: input.doc.metadata })
  ),
]).withConfig({ runName: "Summarize (return doc)" });

// Define the final map-reduce chain
const mapReduceChain = RunnableSequence.from([
  mapAsDocChain.map(),
  collapse,
  reduceChain,
]).withConfig({ runName: "Map reduce" });
```

examples/src/chains/map_reduce_lcel.ts
bracesproul: Updated mdx file with comments

examples/src/chains/map_reduce_lcel.ts
jacoblee93: Ok yeah agreed

examples/src/chains/map_reduce_lcel.ts
```diff
@@ -0,0 +1,156 @@
...
+  RunnableMap,
+  RunnablePassthrough,
+  RunnableSequence,
+} from "langchain/schema/runnable";
+
+// Initialize the OpenAI model
+const model = new ChatOpenAI({});
+
+// Define prompt templates for document formatting, summarizing, collapsing, and combining
+const documentPrompt = PromptTemplate.fromTemplate("{pageContent}");
```
jacoblee93, line 22: This seems unnecessary? Can't we just use that previous util you added?

I don't think the metadata is relevant here

examples/src/chains/map_reduce_lcel.ts
jacoblee93: What about:
```typescript
const mapChain =  RunnableSequence.from([
    { doc: new RunnablePassthrough(), content: mapChain },
     (input) => new Document({ pageContent: input.content, metadata: input.doc.metadata }),
  ]).withConfig({ runName: "Summarize (return doc)" }).map(),
```

examples/src/chains/map_reduce_lcel.ts
bracesproul: This is using one of the utils I added `formatDocument`?  I could use `formatDocumentsAsString` but then we'd lose passing the "source" docs through

examples/src/chains/map_reduce_lcel.ts
bracesproul: Yeah removing `RunnableLambda` works, pushed update

jacoblee93:
Looks good, will merge tonight thanks!

PR #3072, opened 2023-10-27
Jtcorrin/plan execute adjustments

### LINKED ISSUE #3063: [Question] PlanAndExecuteAgentExecutor initial system messages omits tools, leads to poorly suited plan
Hello,

Pretty much the title. The initial system message is:

`"Let's first understand the problem and devise a plan to solve the problem. Please output the plan starting with the header \"Plan:\" and then followed by a numbered list of steps. Please make the plan the minimum number of steps required to answer the query or complete the task accurately and precisely. Your steps should be general, and should not require a specific method to solve a step. If the task is a question, the final step in the plan must be the following: \"Given the above steps taken, please respond to the original query.\" At the end of your plan, say \"<END_OF_PLAN>\""`

The subsequent set of steps are typically (in my testing at least) poorly suited for the tools that I am providing the model. 

Question: I'll be attempting to merge the 2nd system message:

`Answer the following questions as best you can. You have access to the following tools: ...`

with this initial one. Has anyone attempted / come across anything similar? Any advice?

JTCorrin:
@JTCorrinData

Fixes #3063


jacoblee93:
Thanks, this makes sense! 

No worries if you're just exploring, but you're best off using other non-experimental agents like this one at the moment:

https://js.langchain.com/docs/modules/agents/agent_types/openai_functions_agent

langchain/src/experimental/plan_and_execute/prompt.ts
```diff
@@ -31,3 +33,38 @@ Current objective: {current_step}
...
+    `followed by a numbered list of steps.`,
+    `Please make the plan the minimum number of steps required`,
+    `to answer the query or complete the task accurately and precisely.`,
+    `${toolInstruction}`,
+    `If the task is a question, the final step in the plan must be the following: "Given the above steps taken,`,
+    `please respond to the original query."`,
+    `At the end of your plan, say "<END_OF_PLAN>"`,
+  ].join(" ");
+
+  return /* #__PURE__ */ ChatPromptTemplate.fromMessages([
```
jacoblee93, line 64: Can we re-use/delete the above `PLANNER_CHAT_PROMPT`?

JTCorrin: I've removed this in the latest commit. Thanks for taking look!

langchain/src/experimental/plan_and_execute/prompt.ts
```diff
@@ -31,3 +15,40 @@ Current objective: {current_step}
...
+    ? `You have a set of tools at your disposal to help you with this task: ${toolStrings}. You must consider these tools when coming up with your plan.`
+    : `Your steps should be general, and should not require a specific method to solve a step.`;
+
+    PLANNER_SYSTEM_PROMPT_MESSAGE_TEMPLATE = [
+    `Let's first understand the problem and devise a plan to solve the problem.`,
+    `Please output the plan starting with the header "Plan:"`,
+    `followed by a numbered list of steps.`,
+    `Please make the plan the minimum number of steps required`,
+    `to answer the query or complete the task accurately and precisely.`,
+    `${toolInstruction}`,
```
jacoblee93, line 41: We should make this always require tools to be passed in IMO

jacoblee93: Breaking changes are more ok here since it's under experimental

langchain/src/experimental/plan_and_execute/prompt.ts
```diff
@@ -4,25 +4,9 @@ import {
...
-export const PLANNER_CHAT_PROMPT =
-  /* #__PURE__ */ ChatPromptTemplate.fromMessages([
-    /* #__PURE__ */ SystemMessagePromptTemplate.fromTemplate(
-      PLANNER_SYSTEM_PROMPT_MESSAGE_TEMPLATE
-    ),
-    /* #__PURE__ */ HumanMessagePromptTemplate.fromTemplate(`{input}`),
-  ]);
+import { Tool } from "../../tools/base.js";
+
+export let PLANNER_SYSTEM_PROMPT_MESSAGE_TEMPLATE: string
```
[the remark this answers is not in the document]
JTCorrin: no because I'm not initialising it with any value. The value is calculated below

langchain/src/experimental/plan_and_execute/prompt.ts
JTCorrin: Could you elaborate - I don't follow. thanks!

jacoblee93: Would we ever want to have a prompt where there are no tool instructions? Probably not right?

JTCorrin: Ah I see! Thank you. Yes you are absolutely right. I'll get the fixed pushed momentarily.

jacoblee93:
@JTCorrin updated to use a prompt template variable to avoid formatting issues but this makes the `fromLLMAndTools` method now `async`. 

Breaking change to do things the right way is the best move here since it's an experimental module but just FYI this might impact your other PR

jacoblee93:
Thank you!

PR #3076, opened 2023-10-27
Jtcorrin/plan execute structured tools

### LINKED ISSUE #3074: feat: Allow plan_and_execute to accept DynamicStructuredTool
The PlanAndExecuteAgentExecutor does not accommodate the use of tools with schemas. Passing through DynamicStructuredTools allows the models to understand the requirements but at no point (from what I can see) is the schema of the tool revealed to the model.

I think it would make sense to give the user the ability to pass through structured tools with schemas.

JTCorrin:
This builds on my previous PR #3072 - it allows for structured tools to be pass through to the plan and execute executor. There is still work to be done on the prompt however

Fixes #3074


JTCorrin:
This now solves the problem I was trying to solve. Not sure its a particularly pretty solution though so I'm happy to take feedback or make adjustments if those who know the codebase more intimately would like to make suggestions.

langchain/src/experimental/plan_and_execute/agent_executor.ts
```diff
@@ -108,10 +117,25 @@ export class PlanAndExecuteAgentExecutor extends BaseChain {
...
   }: {
     llm: BaseLanguageModel;
-    tools: Tool[];
+    tools: Tool[] | DynamicStructuredTool[];
     humanMessageTemplate?: string;
   }) {
-    const agent = ChatAgent.fromLLMAndTools(llm, tools, {
+    let agent;
+
+    if (tools[0] instanceof DynamicStructuredTool) {
```
jacoblee93, line 125: You'll need to add a linter ignore here

jacoblee93: Alternatively, a type guard could be better

jacoblee93:
Seems fine other than the linter warning!

JTCorrin:
I think that covers it - I'll play around with it more this weekend but, as ever, happy to take any feedback or direction.

jacoblee93:
Hey I'm so sorry this fell behind - it's been an insane week for us. Will review now

PR #3083, opened 2023-10-28
Update ollama.mdx - `numThreads` to `numThread`

paulwongx:
The example in this [the docs](https://js.langchain.com/docs/modules/data_connection/text_embedding/integrations/ollama) for `OllamaEmbeddings` shows `numThreads: 6` for `OllamaEmbeddings` but it should actually accept a `numThread` key, not `numThreads`.

```ts
import { OllamaEmbeddings } from "langchain/embeddings/ollama";

const embeddings = new OllamaEmbeddings({
  model: "llama2", // default value
  baseUrl: "http://localhost:11434", // default value
  requestOptions: {
    useMMap: true,
    numThreads: 6, // <---- Should be `numThread`
    numGpu: 1,
  },
});

const documents = ["Hello World!", "Bye Bye"];

const documentEmbeddings = await embeddings.embedDocuments(documents);
```

The link to the [OllamaEmbeddings API source is here](https://github.com/langchain-ai/langchainjs/blob/b80dc8f5d8aaad857c72aade9053f2b355e10970/langchain/src/embeddings/ollama.ts#L68C7-L68C7) to confirm.

Fixes # (issue)

### DIFF
--- a/docs/docs/modules/data_connection/text_embedding/integrations/ollama.mdx
+++ b/docs/docs/modules/data_connection/text_embedding/integrations/ollama.mdx
@@ -29,7 +29,7 @@ const embeddings = new OllamaEmbeddings({
   baseUrl: "http://localhost:11434", // default value
   requestOptions: {
     useMMap: true, // use_mmap 1
-    numThreads: 6, // num_thread 6
+    numThread: 6, // num_thread 6
     numGpu: 1, // num_gpu 1
   },
 });


jacoblee93:
Thanks, nice catch!

paulwongx:
@jacoblee93 There's another issue with that example. It doesn't work out of the box because the fetch request is targetting `http://localhost:3000/api/embeddings` and fetch by default resolves to the ipv6 address of `::1` as per [this link](https://github.com/node-fetch/node-fetch/issues/1624#issuecomment-1235826631). However, it does work when the baseUrl is `http://127.0.0.1:11434/api/embeddings`. Currently Ollama server seems to be listening on ipv4 address only. 

Do you want me to update both examples with `127.0.0.1` as the default `baseUrl` value so it works out of the box, or should I change the [OllamaEmbeddings file baseUrl](https://github.com/langchain-ai/langchainjs/blob/14f37548ec0606c1481e30921e06d6140f0c2297/langchain/src/embeddings/ollama.ts#L27C5-L27C5) from `localhost` to `127.0.0.1`?

Note: I forgot the change the other example in my origin PR from `numThreads` to `numThread` which I will do in the next PR.

jacoblee93:
Ah yes please! I think the LLM and chat models are configured correctly with that, was an oversight here.

PR #3087, opened 2023-10-29
Ollama embedding - Changed `localhost` to `127.0.0.1`

paulwongx:
[Link to discussion](https://github.com/langchain-ai/langchainjs/pull/3083#issuecomment-1783917599)

1. Updates `numThreads` to `numThread` in the docs to use the API correctly [ref](https://github.com/langchain-ai/langchainjs/blob/b80dc8f5d8aaad857c72aade9053f2b355e10970/langchain/src/embeddings/ollama.ts#L68C7-L68C7)
2. Adds a function in the `OllamaEmbeddings` class to fetch `127.0.0.1` if user puts `localhost` to work out of the box with node's fetch since it default resolves to ipv6 (`::1`) which Ollama server isn't listening to. Now it works out of the box. Issue is related to node's native fetch command not working when fetching `localhost` that isn't listening on the ipv6 localhost address but the ipv4 localhost address of `127.0.0.1` as per [this](https://github.com/node-fetch/node-fetch/issues/1624#issuecomment-1235826631). The function is copied over from the Ollama class implementation. No change in the docs otherwise. 

Fixes # (issue)

### DIFF
--- a/langchain/src/embeddings/ollama.ts
+++ b/langchain/src/embeddings/ollama.ts
@@ -24,7 +24,7 @@ interface OllamaEmbeddingsParams extends EmbeddingsParams {
 export class OllamaEmbeddings extends Embeddings {
   model = "llama2";
 
-  baseUrl = "http://127.0.0.1:11434";
+  baseUrl = "http://localhost:11434";
 
   requestOptions?: OllamaRequestParams["options"];
 
@@ -95,7 +95,17 @@ export class OllamaEmbeddings extends Embeddings {
   async _request(prompt: string): Promise<number[]> {
     const { model, baseUrl, requestOptions } = this;
 
-    const response = await fetch(`${baseUrl}/api/embeddings`, {
+    let formattedBaseUrl = baseUrl;
+    if (formattedBaseUrl.startsWith("http://localhost:")) {
+      // Node 18 has issues with resolving "localhost"
+      // See https://github.com/node-fetch/node-fetch/issues/1624
+      formattedBaseUrl = formattedBaseUrl.replace(
+        "http://localhost:",
+        "http://127.0.0.1:"
+      );
+    }
+
+    const response = await fetch(`${formattedBaseUrl}/api/embeddings`, {
       method: "POST",
       headers: { "Content-Type": "application/json" },
       body: JSON.stringify({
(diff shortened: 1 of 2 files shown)


langchain/src/embeddings/ollama.ts
```diff
@@ -95,7 +95,17 @@ export class OllamaEmbeddings extends Embeddings {
   async _request(prompt: string): Promise<number[]> {
     const { model, baseUrl, requestOptions } = this;
 
-    const response = await fetch(`${baseUrl}/api/embeddings`, {
+    let formattedBaseUrl = baseUrl;
+    if (formattedBaseUrl.startsWith("http://localhost:")) {
+      // Node 18 has issues with resolving "localhost"
+      // See https://github.com/node-fetch/node-fetch/issues/1624
+      formattedBaseUrl = formattedBaseUrl.replace(
```
jacoblee93, line 102: Just to reduce the amount of magic, can we do something like this and change the default in the constructor instead?

https://github.com/langchain-ai/langchainjs/blob/main/langchain/src/chat_models/ollama.ts#L35

jacoblee93:
Small comment

langchain/src/embeddings/ollama.ts
paulwongx: I was about to change the baseUrl to 127.0.0.1 but after looking through the rest of the Ollama code, I realized that all the other Ollama functions like `ChatOllama` and `Ollama` class call on this `createOllamaStream` which has this reformatting logic

```ts
  let formattedBaseUrl = baseUrl;
  if (formattedBaseUrl.startsWith("http://localhost:")) {
    // Node 18 has issues with resolving "localhost"
    // See https://github.com/node-fetch/node-fetch/issues/1624
    formattedBaseUrl = formattedBaseUrl.replace(
      "http://localhost:",
      "http://127.0.0.1:"
    );
  }
```
[Source](https://github.com/langchain-ai/langchainjs/blob/d048313543d8cff5acc91bcb5ee1c6defa830cf5/langchain/src/util/ollama.ts#L96)  

I thought this was more appropriate 
- So the default baseUrl across Ollama classes can be consitently `localhost` - so the docs are consistent too in terms of showing the default baseUrl
- The end user won't mess it up if they override the default baseUrl with `localhost` either

Of course, I can just update the baseUrl in the `OllamaEmbeddings` to be 127.0.0.1 as first discussed if you desire. The docs will just be updated as well to 127.0.0.1 as default and will be inconsistent with the rest of the Ollama classes but once you confirm either approach, I'll make the PR.

jacoblee93:
Thank you!

PR #3088, opened 2023-10-29
Implement streaming for GoogleVertexAI text and chat models

### LINKED ISSUE #2594: Add support for Vertex AI Generative AI streaming response

afirstenberg:
Fixes #2594 
Obsoletes #2625 (which can be abandoned and deleted).

The parser here is... probably not the most efficient, but should be efficient enough for the purposes. And it doesn't require any new outside libraries.

### DIFF
--- a/langchain/src/chat_models/googlevertexai/common.ts
+++ b/langchain/src/chat_models/googlevertexai/common.ts
@@ -1,19 +1,26 @@
 import { BaseChatModel } from "../base.js";
 import {
   AIMessage,
+  AIMessageChunk,
   BaseMessage,
   ChatGeneration,
+  ChatGenerationChunk,
   ChatMessage,
   ChatResult,
   LLMResult,
 } from "../../schema/index.js";
-import { GoogleVertexAILLMConnection } from "../../util/googlevertexai-connection.js";
+import {
+  GoogleVertexAILLMConnection,
+  GoogleVertexAIStream,
+} from "../../util/googlevertexai-connection.js";
 import {
   GoogleVertexAIBaseLLMInput,
   GoogleVertexAIBasePrediction,
+  GoogleVertexAILLMPredictions,
   GoogleVertexAIModelParams,
 } from "../../types/googlevertexai-types.js";
 import { BaseLanguageModelCallOptions } from "../../base_language/index.js";
+import { CallbackManagerForLLMRun } from "../../callbacks/index.js";
 
 /**
  * Represents a single "example" exchange that can be provided to
@@ -192,6 +199,13 @@ export class BaseChatGoogleVertexAI<AuthOptions>
     AuthOptions
   >;
 
+  streamedConnection: GoogleVertexAILLMConnection<
+    BaseLanguageModelCallOptions,
+    GoogleVertexAIChatInstance,
+    GoogleVertexAIChatPrediction,
+    AuthOptions
+  >;
+
   get lc_aliases(): Record<string, string> {
     return {
       model: "model_name",
@@ -214,19 +228,46 @@ export class BaseChatGoogleVertexAI<AuthOptions>
     return [];
   }
 
-  // TODO: Add streaming support
+  async *_streamResponseChunks(
+    _messages: BaseMessage[],
+    _options: this["ParsedCallOptions"],
+    _runManager?: CallbackManagerForLLMRun
+  ): AsyncGenerator<ChatGenerationChunk> {
+    // Make the call as a streaming request
+    const instance: GoogleVertexAIChatInstance = this.createInstance(_messages);
+    const parameters = this.formatParameters();
+    const result = await this.streamedConnection.request(
+      [instance],
+      parameters,
+      _options
+    );
+
+    // Get the streaming parser of the response
+    const stream = result.data as GoogleVertexAIStream;
+
+    // Loop until the end of the stream
+    // During the loop, yield each time we get a chunk from the streaming parser
+    // that is either available or added to the queue
+    while (!stream.streamDone) {
+      const output = await stream.nextChunk();
+      const chunk =
+        output !== null
+          ? BaseChatGoogleVertexAI.convertPredictionChunk(output)
+          : new ChatGenerationChunk({
+              text: "",
+              message: new AIMessageChunk(""),
+              generationInfo: { finishReason: "stop" },
+            });
+      yield chunk;
+    }
+  }
+
   async _generate(
     messages: BaseMessage[],
     options: this["ParsedCallOptions"]
   ): Promise<ChatResult> {
     const instance: GoogleVertexAIChatInstance = this.createInstance(messages);
-
-    const parameters: GoogleVertexAIModelParams = {
-      temperature: this.temperature,
-      topK: this.topK,
-      topP: this.topP,
-      maxOutputTokens: this.maxOutputTokens,
-    };
+    const parameters: GoogleVertexAIModelParams = this.formatParameters();
 
     const result = await this.connection.request(
       [instance],
@@ -235,7 +276,9 @@ export class BaseChatGoogleVertexAI<AuthOptions>
     );
 
     const generations =
-      result?.data?.predictions?.map((prediction) =>
+      (
+        result?.data as GoogleVertexAILLMPredictions<GoogleVertexAIChatPrediction>
+      )?.predictions?.map((prediction) =>
         BaseChatGoogleVertexAI.convertPrediction(prediction)
       ) ?? [];
     return {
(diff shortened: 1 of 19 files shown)


langchain/src/chat_models/googlevertexai/common.ts
```diff
@@ -235,7 +276,9 @@ export class BaseChatGoogleVertexAI<AuthOptions>
     );
 
     const generations =
-      result?.data?.predictions?.map((prediction) =>
+      (
+        result?.data as GoogleVertexAILLMPredictions<GoogleVertexAIChatPrediction>
```
jacoblee93, line 280: Is the cast necessary? Should be `any` anyway and if we wanted to be sure we should use a type guard?

langchain/src/embeddings/googlevertexai.ts
```diff
@@ -114,9 +115,9 @@ export class GoogleVertexAIEmbeddings
       responses
         ?.map(
           (response) =>
-            response.data?.predictions?.map(
-              (result) => result.embeddings.values
-            ) ?? []
+            (
+              response?.data as GoogleVertexAILLMPredictions<GoogleVertexEmbeddingsResults>
```
jacoblee93, line 119: See above

langchain/src/experimental/multimodal_embeddings/googlevertexai.ts
```diff
@@ -127,7 +128,9 @@ export class GoogleVertexAIMultimodalEmbeddings
   responseToEmbeddings(
     response: GoogleVertexAILLMResponse<GoogleVertexAIMultimodalEmbeddingsResults>
   ): MediaEmbeddings[] {
-    return response.data.predictions.map((r) => ({
+    return (
+      response?.data as GoogleVertexAILLMPredictions<GoogleVertexAIMultimodalEmbeddingsResults>
```
jacoblee93, line 132: See above

langchain/src/chat_models/googlevertexai/common.ts
```diff
@@ -214,19 +228,46 @@ export class BaseChatGoogleVertexAI<AuthOptions>
...
+    // that is either available or added to the queue
+    while (!stream.streamDone) {
+      const output = await stream.nextChunk();
+      const chunk =
+        output !== null
+          ? BaseChatGoogleVertexAI.convertPredictionChunk(output)
+          : new ChatGenerationChunk({
+              text: "",
+              message: new AIMessageChunk(""),
+              generationInfo: { finishReason: "stop" },
```
jacoblee93, line 259: We should break in this case?

langchain/src/chat_models/googlevertexai/common.ts
```diff
@@ -214,19 +228,46 @@ export class BaseChatGoogleVertexAI<AuthOptions>
...
+      _options
+    );
+
+    // Get the streaming parser of the response
+    const stream = result.data as GoogleVertexAIStream;
+
+    // Loop until the end of the stream
+    // During the loop, yield each time we get a chunk from the streaming parser
+    // that is either available or added to the queue
+    while (!stream.streamDone) {
```
jacoblee93, line 251: Feels like this should be an iterator?

langchain/src/util/googlevertexai-connection.ts
```diff
@@ -106,29 +191,229 @@ export class GoogleVertexAILLMConnection<
...
+  }
+
+  /**
+   * Is the stream done?
+   * A stream is only done if all of the following are true:
+   * - There is no more data to be added to the text buffer
+   * - There is no more data in the text buffer
+   * - There are no chunks that are waiting to be consumed
+   */
+  get streamDone(): boolean {
```
jacoblee93, line 411: Would be awesome to implement as an iterator - I can take a crack at it later this week

langchain/src/util/googlevertexai-webauth.ts
```diff
@@ -63,6 +96,7 @@ export class WebGoogleAuth implements GoogleAbstractedClient {
     const res = await fetch(opts.url, fetchOptions);
 
     if (!res.ok) {
+      console.error(res);
```
jacoblee93, line 99: Remove log?

langchain/src/util/googlevertexai-connection.ts
```diff
@@ -106,29 +191,229 @@ export class GoogleVertexAILLMConnection<
...
+  async nextChunk(): Promise<any> {
+    if (this._chunkQueue.length > 0) {
+      // If there is data in the queue, return the next queue chunk
+      return this._chunkQueue.shift() as GenerationChunk;
+    } else {
+      // Otherwise, set up a promise that handleChunk will cause to be resolved
+      this._chunkPending = new Promise((resolve) => {
+        this._chunkResolution = resolve;
+      });
+      return await this._chunkPending;
```
jacoblee93, line 400: Redundant `await`

jacoblee93 (commented):
Looks awesome overall! See comments

langchain/src/chat_models/googlevertexai/common.ts
afirstenberg: I did it this way to make sure we have a clear final chunk, but am not opposed to just breaking in this case and not sending it.

langchain/src/chat_models/googlevertexai/common.ts
afirstenberg: This seemed cleaner (and more obvious) than using an any in the type for data and marking where to ignore it. (And ditto for all the other places where it happens.)

langchain/src/chat_models/googlevertexai/common.ts
afirstenberg: I had enough trouble trying to wrap my head around implementing the async iterator correctly. I don't think I wanted to make multiple, nested, async iterators.

langchain/src/util/googlevertexai-webauth.ts
afirstenberg: Oops. Yeah. This was left in while I was debugging why my auth with web wasn't working.

langchain/src/util/googlevertexai-connection.ts
afirstenberg: Mmm! Good point and good catch.

afirstenberg:
I shy away from making it an iterator for the reason I outlined above - this is more straightforward, and trying to wrap my head around async iterators that worked with async iterators is making my head spin.
Is there a reason in favor of them in this case?

langchain/src/chat_models/googlevertexai/common.ts
jacoblee93: Oh, no I mean break after getting `null`. It's fine with me to have the final chunk like this but we wouldn't want to receive any more.

jacoblee93:
Yeah they are tricky to implement, benefits are the standardized `for await const` loop consumption flow and the methods we have to convert them to readable streams, but I think not necessary in this case since they're not user facing.

jacoblee93:
Hey @afirstenberg actually this seems to have broken the web integration test for non-streaming:

```typescript
MacBook-Pro-2:langchainjs jacoblee$ yarn test:single langchain/src/chat_models/tests/chatgooglevertexai_web.int.test.ts
(node:20994) ExperimentalWarning: VM Modules is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
  console.log
    {
      "generations": [
        []
      ],
      "llmOutput": []
    }

      at Object.log (src/chat_models/tests/chatgooglevertexai_web.int.test.ts:39:13)

  console.log
    [ [] ]

      at Object.log (src/chat_models/tests/chatgooglevertexai_web.int.test.ts:68:13)

  console.log
    [ [] ]

      at Object.log (src/chat_models/tests/chatgooglevertexai_web.int.test.ts:86:13)

  console.log
    [
     []
    ]
```

jacoblee93:
Fixed it, should be good to go

langchain/src/chat_models/googlevertexai/common.ts
afirstenberg: We need to yield with that chunk, and after that it should be the last thing in the queue or that gets resolved, so the while loop will terminate.

PR #3089, opened 2023-10-29
feat: add file-system based cache

vdeturckheim:
This PR adds a FS-based cache in the library. It is intended for local use and made my life pretty handy in the past weeks.


langchain/src/cache/fs.ts
```diff
@@ -0,0 +1,70 @@
+import { BaseCache, Generation } from "../schema/index.js";
+import { getCacheKey } from "./base.js";
+import path from "path";
+import fs from "node:fs/promises";
+
+const DEFAULT_CACHE_DIR = path.join(process.cwd(), ".langchain-cache");
```
jacoblee93, line 6: Makes more sense to use something like this if not specified?

https://github.com/langchain-ai/langchainjs/blob/80fea35b8d4dceb64ad4394271977e8d94777455/langchain/src/storage/tests/file_system.test.ts#L11

langchain/src/cache/tests/fs.test.ts
```diff
@@ -0,0 +1,40 @@
...
+import { ChatOpenAI } from "../../chat_models/openai.js";
+import fs from "node:fs/promises";
+import path from "node:path";
+import * as os from "node:os";
+import { FSCache } from "../fs.js";
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+let tmpDir: string;
+
+describe("Test RedisCache", () => {
```
jacoblee93, line 13: Change test name

langchain/src/cache/fs.ts
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * A cache that uses the local filesystem as the backing store.
+ * This is useful for local development and testing. But it is not recommended for production use.
+ */
+export class FSCache extends BaseCache {
```
jacoblee93, line 12: Small preference for `LocalFileCache` and rename the file to `cache/file_system.ts` to match:

https://github.com/langchain-ai/langchainjs/blob/80fea35b8d4dceb64ad4394271977e8d94777455/langchain/src/storage/file_system.ts#L9

jacoblee93 (commented):
Thanks for this! Can you add a small paragraph to the docs as well?

langchain/src/cache/fs.ts
vdeturckheim: I think this one is up to what you prefer as a default behaviour. In my case, I want the cache to survive the end of the process execution. But both situation are fully possible. So if you confirm you prefer a temp dir, let's do it.

vdeturckheim:
Let's double confirm the default behaviour we prefer and I'll update the doc accordingly!

langchain/src/cache/fs.ts
jacoblee93: I'd prefer tempdir as a default then 👍

jacoblee93:
Left a reply above!

vdeturckheim:
Should be good to go. PTAL :)

jacoblee93:
Hey @vdeturckheim, it looks like the test fails for chat models, can you fix?

vdeturckheim:
Will do - sorry for the long ping

vdeturckheim:
The test fail was due to an OpenAI key not being available. I just rebased to see if that fixes the issue atm

jacoblee93:
Thank you for your patience!

PR #3090, opened 2023-10-29
Added PyInterpreterTool

mishushakov:
I have added the Pyodide interpreter to LangChainJS

Example:

```ts
import { StringOutputParser } from "../../schema/output_parser.js";
import { OpenAI } from "../../llms/openai.js";
import { PromptTemplate } from "../../prompts/index.js";
import { PythonInterpreterTool } from "../pyinterpreter.js";

const prompt = PromptTemplate.fromTemplate(
  `Can you generate python code that: {input}?`
);

const model = new OpenAI({});

const interpreter = new PythonInterpreterTool({
  indexURL: "../node_modules/pyodide",
});

const chain = prompt
  .pipe(model)
  .pipe(new StringOutputParser())
  .pipe(interpreter);

const result = await chain.invoke({
  input: "print the first 10 numbers of the fibonacci sequence",
});

console.log(result)

// prints:
// {"stdout": "0112358132134", "stderr": ""}
```


langchain/scripts/create-entrypoints.js
```diff
@@ -326,6 +326,7 @@ const requiresOptionalDependency = [
   "tools/sql",
   "tools/webbrowser",
   "tools/google_calendar",
+  "tools/pyinterpreter",
```
jacoblee93, line 329: Let's move to `experimental` since it's potentially running arbitrary code

jacoblee93: You'll also need to add an entrypoint above

langchain/package.json
```diff
@@ -1011,6 +1012,7 @@
     "playwright": "^1.32.1",
     "portkey-ai": "^0.1.11",
     "puppeteer": "^19.7.2",
+    "pyodide": "0.24.1",
```
jacoblee93, line 1015: Add under `peerDependenciesMeta` below

langchain/src/tools/index.ts
```diff
@@ -44,3 +44,4 @@ export {
 export { SearxngSearch } from "./searxng_search.js";
 export { SearchApi, type SearchApiParameters } from "./searchapi.js";
 export { formatToOpenAIFunction } from "./convert_to_openai.js";
+export { PythonInterpreterTool } from "./pyinterpreter.js";
```
jacoblee93, line 47: We can't export from here if it requires an optional peer dep, we'll need to use a custom entrypoint

jacoblee93 (changes requested):
Very cool, needs docs as well. Can you also add a security warning to those as well?

mishushakov:
Of course. Where should I put them? In the comments or do you have a special place for tools anywhere? 😄

What should the warning say? The code is executed in secure WASM sandbox...

I also have a question regarding the tests. Should it only test the tool itself or the complete chain like I did there?

Thanks

jacoblee93:
> Of course. Where should I put them? In the comments or do you have a special place for tools anywhere? 😄
> 
> What should the warning say? The code is executed in secure WASM sandbox...
> 
> I also have a question regarding the tests. Should it only test the tool itself or the complete chain like I did there?
> 
> Thanks

Best place would be to add a new folder called `tools/experimental` and put it there.

mishushakov:
Ok done. Let me know where I should put the docs and the warning

jacoblee93:
Hi so sorry, it's been a crazy week for us - will look at this and write docs tomorrow!

jacoblee93:
Hey @mishushakov, I moved things around but the integration test seems rather flaky - do you think you could come up with a more consistent one?

mishushakov:
Hey, I have been busy this week, but happy to tackle this on the weekend!

mishushakov:
Is this better? Or are you looking for something else entirely?

```
import { StringOutputParser } from "../../../schema/output_parser.js";
import { OpenAI } from "../../../llms/openai.js";
import { PromptTemplate } from "../../../prompts/index.js";
import { PythonInterpreterTool } from "../pyinterpreter.js";

describe("Python Interpreter testsuite", () => {
  test("hello langchain", async () => {
    const prompt = PromptTemplate.fromTemplate(
      `Can you generate python code that: {input}?`
    );

    const model = new OpenAI({});

    const interpreter = await PythonInterpreterTool.initialize({
      indexURL: "../node_modules/pyodide",
    });
    const chain = prompt
      .pipe(model)
      .pipe(new StringOutputParser())
      .pipe(interpreter);

    const result = await chain.invoke({
      input: `prints "Hello LangChain"`,
    });

    expect(JSON.parse(result).stdout).toBe("Hello LangChain");
  });
});
```

mishushakov:
The test does require OpenAI API key though

mishushakov:
Otherwise, a very basic test, that only tests the tool itself would look like:

```ts
import { PythonInterpreterTool } from "../pyinterpreter.js";

describe("Python Interpreter testsuite", () => {
  test("hello langchain", async () => {
    const interpreter = await PythonInterpreterTool.initialize({
      indexURL: "../node_modules/pyodide",
    });

    const result = await interpreter._call(`print("Hello LangChain")`)
    expect(JSON.parse(result).stdout).toBe("Hello LangChain");
  });
});
```

jacoblee93 (approved):

jacoblee93:
Thank you! It's really neat

PR #3091, opened 2023-10-29
Added GraphQLClientTool

mishushakov:
I have added a GraphQL client tool to execute graphql requests.

Example:

```ts
import { z } from "zod";
import { RunnableSequence } from "../../schema/runnable/index.js";
import { PromptTemplate } from "../../prompts/index.js";
import { OpenAI } from "../../llms/openai.js";
import { StructuredOutputParser } from "../../output_parsers/index.js";
import { GraphQLClientTool } from "../graphql.js";

  z.object({
    query: z.string(),
    variables: z.object({}),
  })
);

  PromptTemplate.fromTemplate(
    "Make a GraphQL query to get the country code of Germany.\n{format_instructions}\n{schema}"
  ),
  new OpenAI({ temperature: 0 }),
  parser,
  new GraphQLClientTool({
    endpoint: "https://countries.trevorblades.com",
  }),
]);

const response = await chain.invoke({
  schema: `
      type Continent {
        code: ID!
        countries: [Country!]!
        name: String!
      }

      input ContinentFilterInput {
        code: StringQueryOperatorInput
      }

      type Country {
        awsRegion: String!
        capital: String
        code: ID!
        continent: Continent!
        currencies: [String!]!
        currency: String
        emoji: String!
        emojiU: String!
        languages: [Language!]!
        name(lang: String): String!
        native: String!
        phone: String!
        phones: [String!]!
        states: [State!]!
        subdivisions: [Subdivision!]!
      }

      input CountryFilterInput {
        code: StringQueryOperatorInput
        continent: StringQueryOperatorInput
        currency: StringQueryOperatorInput
        name: StringQueryOperatorInput
      }

      type Language {
        code: ID!
        name: String!
        native: String!
        rtl: Boolean!
      }

      input LanguageFilterInput {
        code: StringQueryOperatorInput
      }

      type Query {
        continent(code: ID!): Continent
        continents(filter: ContinentFilterInput = {}): [Continent!]!
        countries(filter: CountryFilterInput = {}): [Country!]!
        country(code: ID!): Country
        language(code: ID!): Language
        languages(filter: LanguageFilterInput = {}): [Language!]!
      }

      type State {
        code: String
        country: Country!
        name: String!
      }

      input StringQueryOperatorInput {
        eq: String
        in: [String!]
        ne: String
        nin: [String!]
        regex: String
      }

      type Subdivision {
        code: ID!
        emoji: String
        name: String!
      }
      `,
  format_instructions: parser.getFormatInstructions(),
});

console.log(response);
```


jacoblee93 (changes requested):
Thanks for this! Similar to your Python interpreter one we've started flagging to folks that the LLM could generate arbitrary mutations, and that this should live under `experimental` with a security warning. Would you mind moving?

mishushakov:
Sure! Will follow up on both some time this week

mishushakov:
I have moved the tool under tools/experimental. Where should I put the docs and the warning now?

qnm:
@mishushakov This works pretty well for my purposes, however when I use it as part of a `structured-chat-zero-shot-react-description` agent, only `query` is passed to the tool, never `variables`. I ended up setting `variables` to optional in the type definition to work around this.

Unfortunately, I only started using langchain a couple of days ago so I don't really know if that's expected for that type of agent, or if something else is required. Any help is appreciated!

mishushakov:
Glad to hear that! Variables aren't required, so I should probably make it optional as you suggested 😀

jacoblee93:
Looks good - sounds like you're going to address @qnm's feedback though?

mishushakov:
I have updated the tool, tests and moved it to experimental/tools now
Anything else needed?

langchain/src/experimental/tools/graphql.ts
```diff
@@ -0,0 +1,47 @@
...
+    const response = await fetch(this._endpoint, {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json",
+        ...this._headers,
+      },
+      body: JSON.stringify({ query, variables }),
+    });
+
+    return await response.text();
```
bracesproul, line 45:
```suggestion
    return response.text();
```
shouldn't need to await on a return statement

bracesproul (commented):
Thanks for adding this! Looks good to me once my comments have been implemented!

mishushakov:
Hello and Merry Christmas! I have done so in the latest commit

mishushakov:
Yo @bracesproul & @jacoblee93, would you be able to review? Thank you so much 😊

PR #3104, opened 2023-10-31
feat(huggingface_inference): Add support for huggingface custom inference endpoint URL

OlegIvaniv:
This PR adds support for custom [Hugging Face inference endpoint](https://huggingface.co/docs/inference-endpoints/main/en/index) for both LM and embeddings. To implement this, we need to update the [@huggingface/inference](https://www.npmjs.com/package/@huggingface/inference) library to [v2](https://github.com/huggingface/huggingface.js/releases/tag/inference-v2.0.0).

If an `endpointUrl` parameter is provided when instantiating `HuggingFaceInference` or `HuggingFaceInferenceEmbeddings`, the custom endpoint will be utilized for inference calls.


jacoblee93:
Thanks for this! Looks good to me

PR #3108, opened 2023-10-31
fix pinecone bug

bracesproul:


omikader:
@bracesproul curious what was the bug?

bracesproul:
> @bracesproul curious what was the bug?

It's not necessarily a bug, but the config was missing some options `AsyncCallerParams` and the `addVectors` method wasn't using the `caller` to upsert

PR #3119, opened 2023-11-01
Update Neo4j examples

tomasonjo:
The examples didn't work as the graph schema was undefined if you start with an empty database.


jacoblee93:
Thank you!

PR #3120, opened 2023-11-01
Update handling missing APOC procedures

tomasonjo:
Better exception messaging of missing APOC procedures

### DIFF
--- a/langchain/src/graphs/neo4j_graph.ts
+++ b/langchain/src/graphs/neo4j_graph.ts
@@ -1,4 +1,4 @@
-import neo4j from "neo4j-driver";
+import neo4j, { Neo4jError } from "neo4j-driver";
 
 interface Neo4jGraphConfig {
   url: string;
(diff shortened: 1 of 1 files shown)


langchain/src/graphs/neo4j_graph.ts
```diff
@@ -1,4 +1,4 @@
-import neo4j from "neo4j-driver";
+import neo4j, { Neo4jError } from "neo4j-driver";
```
jacoblee93, line 1: Use `type`?

langchain/src/graphs/neo4j_graph.ts
```diff
@@ -58,7 +58,9 @@ export class Neo4jGraph {
       await graph.refreshSchema();
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
     } catch (error: any) {
-      throw new Error(`Error: ${error.message}`);
+      throw new Error(`Could not use APOC procedures.
+      Please ensure the APOC plugin is installed in Neo4j and that
```
jacoblee93, line 62: Is the indent intentional?

jacoblee93:
Seems fine, thank you!

langchain/src/graphs/neo4j_graph.ts
tomasonjo: No, I'm just not experienced with JS :) Do I simply remove the indent?

tomasonjo: fixed

PR #3127, opened 2023-11-01
Convex: Fix defaults doc comments

xixixao:
Forgot to update these before merging. Now they will show up in the compiled .d.ts files (where jump to definition will lead people).


bracesproul (commented):
Thanks!

PR #3129, opened 2023-11-01
Fixed crypto could not be resolved warning on build

bracesproul:
Was giving the `crypto could not be resolved. treating it as an optional dependency` warning on build (lost the exact warning message)

### DIFF
--- a/langchain/src/chat_models/iflytek_xinghuo/index.ts
+++ b/langchain/src/chat_models/iflytek_xinghuo/index.ts
@@ -4,6 +4,7 @@ import {
   BaseWebSocketStream,
   WebSocketStreamOptions,
 } from "../../util/iflytek_websocket_stream.js";
+import { getEnv } from "../../util/env.js";
 
 class WebSocketStream extends BaseWebSocketStream {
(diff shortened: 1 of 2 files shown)


langchain/src/chat_models/iflytek_xinghuo/index.ts
```diff
@@ -17,10 +18,11 @@ export class ChatIflytekXinghuo extends BaseChatIflytekXinghuo {
   async openWebSocketStream<WebSocketStream>(
     options: WebSocketStreamOptions
   ): Promise<WebSocketStream> {
+    const { createHmac } = await ChatIflytekXinghuo.imports();
+
     const host = "spark-api.xf-yun.com";
     const date = new Date().toUTCString();
     const url = `GET /${this.version}/chat HTTP/1.1`;
-    const { createHmac } = await import("crypto");
```
jacoblee93, line 23: Changing this to `node:crypto` alone should fix it I think?

jacoblee93 (commented):
Nice catch - but I think there's an easier fix and we don't want to import `node:crypto` in the web entrypoint

PR #3136, opened 2023-11-02
Cassandra ChatMessageHistory and Cassandra VectorStore enhancements

mieslep:
- Adds: CassandraChatMessageHistory
- Adds: Vector types of dot_product and euclildean to the default cosine
- Adds: Filter searches support operators other than just = 
- Adds: Support for composite primary keys, with partition definition
- Adds: Support for WITH clause on CREATE TABLE
- Fixes: Similarity score is returned, instead of just ranking
- Fixes: Performance enhancements (prepared statements, removing unnecessary vector value return)


langchain/src/vectorstores/cassandra.ts
```diff
@@ -215,54 +280,172 @@ export class CassandraStore extends VectorStore {
...
+                          ${
+                            this.metadataColumns.length > 0
+                              ? ", " +
+                                this.metadataColumns
+                                  .map((col) => `${col.name}`)
+                                  .join(", ")
+                              : ""
+                          }`;
+
+    cql = `CREATE CUSTOM INDEX IF NOT EXISTS idx_vector_${this.table}
```
jacoblee93, line 313: Any injection risk?

langchain/src/stores/message/cassandra.ts
```diff
@@ -0,0 +1,129 @@
+import { Client, DseClientOptions } from "cassandra-driver";
```
jacoblee93, line 1: You'll need to add an entrypoint in `create-entrypoints.js` for this:

jacoblee93 (changes requested):
All seems reasonable, two small comments

langchain/src/stores/message/cassandra.ts
mieslep: done

langchain/src/vectorstores/cassandra.ts
mieslep: none added - this is part of the vector store "create" process. The PR does resolve injection risks though, as previously the values were being used as part of the literal string, and the update uses bind variables instead.

langchain/src/vectorstores/tests/cassandra.int.test.ts
```diff
@@ -1,34 +1,49 @@
 /* eslint-disable no-process-env */
```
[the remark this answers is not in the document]
mieslep: these are in .skip .int tests

mieslep:
@jacoblee93 questions addressed, still haven't fully worked out how to rebase on top of main, as whenever I try I end up pulling in hundreds of unrelated files to the PR :/

jacoblee93:
All good, thank you!

jacoblee93 (approved):

PR #3137, opened 2023-11-02
Bedrock: Add Cohere and edge support

dqbd:
- [x] Remove Buffer to support edge
- [x] add streaming support for Cohere
- [x] use stop sequences from .bind()
- [x] rename ChatBedrock to BedrockChat to match Python
- [x] add serialisation flags and secrets
- [x] Separate into web-friendly endpoints to avoid importing "@aws-sdk/credential-provider-node"


langchain/src/chat_models/bedrock/index.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export {
+  convertMessagesToPromptAnthropic,
+  convertMessagesToPrompt,
+} from "./web.js";
+
+/**
+ * @deprecated Use `BedrockChat` instead.
+ */
+export const ChatBedrock = BedrockChat;
```
jacoblee93, line 27: 😭

jacoblee93:
Looks good, thank you! I'll take care of docs later today

PR #3147, opened 2023-11-03
Voyage AI embeddings

### LINKED ISSUE #3146: Add Voyage AI embeddings [integration] [sync python integrations]
Sync Voyage AI embeddings python integration with the js library.

valdo99:
Adding VoyageEmbedding class to the embeddins.
Added integration tests.

the code is highly inspired by minmax and cohere :)

implemented the rest apis documented here: https://docs.voyageai.com/embeddings/

Fixes #3146


langchain/src/embeddings/voyage.ts
```diff
@@ -0,0 +1,157 @@
+import { chunkArray } from "../util/chunk.js";
+import { getEnvironmentVariable } from "../util/env.js";
+import { Embeddings, EmbeddingsParams } from "./base.js";
+
+enum VoyageModelNames {
```
jacoblee93, line 5: Ah we can't use `enum` unfortunately as it'll break tree shaking :(

Can you use a type here instead?

langchain/src/embeddings/voyage.ts
```diff
@@ -0,0 +1,157 @@
...
+/**
+ * A class for generating embeddings using the Voyage AI API.
+ */
+export class VoyageEmbeddings
+  extends Embeddings
+  implements VoyageEmbeddingsParams
+{
+  modelName = VoyageModelNames.voyage01;
+
+  batchSize = 8;
```
jacoblee93, line 51: This seems a bit small?

langchain/src/embeddings/voyage.ts
```diff
@@ -0,0 +1,157 @@
...
+   * Constructor for the VoyageEmbeddings class.
+   * @param fields - An optional object with properties to configure the instance.
+   */
+  constructor(
+    fields?: Partial<VoyageEmbeddingsParams> & {
+      verbose?: boolean;
+      apiKey?: string;
+    }
+  ) {
+    const fieldsWithDefaults = { maxConcurrency: 2, ...fields };
```
jacoblee93, line 71: The default here was just for legacy purposes, you can leave it unset/set it higher if you choose

jacoblee93 (changes requested):
Thanks for this! Mostly LGTM, only one important change.

langchain/src/embeddings/voyage.ts
valdo99: https://docs.voyageai.com/embeddings/#via-http-request 

> as you can see: input - A single string or a list of strings. The maximum number of strings is 8.

langchain/src/embeddings/voyage.ts
valdo99: i'll just leave it unset

langchain/src/embeddings/voyage.ts
valdo99: @jacoblee93 i removed even the type, cause i see on the Voyage docs, that they have more embeddings model rolling out, so i'll just leave the possibility tu change the model name :)

jacoblee93:
Thanks! You'll also want to add docs, and can you run `yarn build` from `langchain/` to update build artifacts?

valdo99:
Yes my bad, added base docs + artifacts

valdo99:
@jacoblee93 are we good now? :)

jacoblee93:
LGTM thank you!

langchain/src/embeddings/voyage.ts
oliviertassinari: The batch size logic used in Python: https://github.com/voyage-ai/langchain-voyageai/blob/225468c4e121745e14cd23e316d9633863206d0b/libs/voyageai/langchain_voyageai/embeddings.py#L61.

valdo99: i think it's better to create a new issue to improve based on the new logic :)
happy to help if this is needed fast :)

oliviertassinari: The current JavaScript logic seems like a good approximation of the Python one. No issues on my end.

PR #3152, opened 2023-11-03
Generate api refs with typedoc

bracesproul:
API refs now on standalone site (typedoc, served via the public folder of a next.js app lol)
Update code block generation to link to new site


.gitignore
```diff
@@ -36,4 +36,6 @@ firestore-debug.log
 tmp/
 
 **/convex/.gitignore
-**/convex/tsconfig.json
\ No newline at end of file
+**/convex/tsconfig.json
+
+langchain/docs_dist/**/*
```
jacoblee93, line 41: Why not just `langchain/docs/dist`

api-docs/.eslintrc.json
```diff
@@ -0,0 +1,3 @@
+{
```
jacoblee93, line 1: I would suggest putting this within `docs/` rather than adding a new workspace if possible

package.json
```diff
@@ -8,7 +8,8 @@
   "workspaces": [
     "langchain",
     "examples",
-    "docs"
+    "docs",
+    "api-docs"
```
jacoblee93, line 12: Would prefer not to have a new workspace if possible - can we move into `docs` and add new commands there?

api-docs/README.md
```diff
@@ -0,0 +1,11 @@
+# Autogeneratd API documentation for LangChainJS
```
jacoblee93, line 1: Typo

docs/package.json
```diff
@@ -5,7 +5,8 @@
   "scripts": {
     "docusaurus": "docusaurus",
     "start": "rimraf ./docs/api &&  NODE_OPTIONS=--max-old-space-size=7168 docusaurus start",
-    "build": "rimraf ./build && NODE_OPTIONS=--max-old-space-size=7168 DOCUSAURUS_SSR_CONCURRENCY=4 docusaurus build",
+    "build": "yarn gen:typedoc && rimraf ./build && NODE_OPTIONS=--max-old-space-size=7168 DOCUSAURUS_SSR_CONCURRENCY=4 docusaurus build",
+    "gen:typedoc": "cd ../api-docs && yarn gen:typedoc",
```
jacoblee93, line 9: Yeah let's keep in same workspace if possible

docs/docusaurus.config.js
```diff
@@ -29,12 +29,6 @@ const config = {
   onBrokenMarkdownLinks: "throw",
 
   plugins: [
-    [
-      "docusaurus-plugin-typedoc",
```
jacoblee93, line 33: We can remove this as a dep and `.yarn/patch` right?

langchain/package.json
```diff
@@ -806,7 +806,8 @@
     "build:esm": "NODE_OPTIONS=--max-old-space-size=4096 tsc --outDir dist/ && rimraf dist/tests dist/**/tests",
     "build:cjs": "NODE_OPTIONS=--max-old-space-size=4096 tsc --outDir dist-cjs/ -p tsconfig.cjs.json && node scripts/move-cjs-to-dist.js && rimraf dist-cjs",
     "build:watch": "node scripts/create-entrypoints.js && tsc --outDir dist/ --watch",
-    "build:scripts": "node scripts/create-entrypoints.js && node scripts/check-tree-shaking.js && node scripts/generate-docs-llm-compatibility-table",
+    "build:scripts": "node scripts/create-entrypoints.js && node scripts/check-tree-shaking.js && yarn conditional:api-docs && node scripts/generate-docs-llm-compatibility-table",
+    "conditional:api-docs": "bash scripts/build-api-refs.sh",
```
jacoblee93, line 810: What does `conditional:` imply here?

langchain/scripts/ignore-class-properties.js
```diff
@@ -0,0 +1,69 @@
+import { readFile, writeFile, readdir, stat } from "fs/promises";
```
jacoblee93, line 1: Can we call `transform-for-api-docs.js` or something?

jacoblee93 (changes requested):
Looks good overall, high level I would really love to avoid making another workspace if possible

docs/api-docs/generate_typedoc.sh
```diff
@@ -1,11 +1,11 @@
 #!/bin/bash
 
-BASE_DIR="../langchain"
+BASE_DIR="../../langchain"
 # Create a directory to temp hold the files in src
 # `docs_dist` is gitignored by default.
-mkdir -p "$BASE_DIR/docs_dist/src"
+mkdir -p "$BASE_DIR/docs/dist"
```
jacoblee93, line 6: How about `/docs/api-docs/dist` to keep everything localized?

docs/package.json
```diff
@@ -6,7 +6,7 @@
     "docusaurus": "docusaurus",
     "start": "rimraf ./docs/api &&  NODE_OPTIONS=--max-old-space-size=7168 docusaurus start",
     "build": "yarn gen:typedoc && rimraf ./build && NODE_OPTIONS=--max-old-space-size=7168 DOCUSAURUS_SSR_CONCURRENCY=4 docusaurus build",
-    "gen:typedoc": "cd ../api-docs && yarn gen:typedoc",
+    "gen:typedoc": "cd ./api-docs && yarn gen:typedoc",
```
jacoblee93, line 9: `build:typedoc` is a better name?

jacoblee93 (changes requested):
Small naming nits and then I think we're good to go!

docs/api-docs/generate_typedoc.sh
jacoblee93: Or actually sorry but how about `docs/api_refs/` (we use snake case for directories elsewhere and `docs/api_docs` feels redundant)

jacoblee93:
And fix CI

PR #3153, opened 2023-11-04
Fix typo in output types table

jonathanmv:
When describing the [common interface](https://js.langchain.com/docs/expression_language/interface), there is a typo in the table that explains the different output types. 

This PR fixes the typo in the header of the table.


jacoblee93:
Thank you!

PR #3168, opened 2023-11-06
OpenAI Multimodal schema changes

jacoblee93:


jacoblee93:
CC @bracesproul @dqbd @nfcampos

langchain/src/schema/index.ts
```diff
@@ -101,8 +101,19 @@ export interface StoredGeneration {
 
 export type MessageType = "human" | "ai" | "generic" | "system" | "function";
 
+export type MessageContent =
+  | string
+  | (
+      | {
+          type: "text" | "image_url";
+          text?: string;
+          image_url: string;
```
dqbd, line 110: It should be this I think

```suggestion
          image_url: string | { url: string };
```

langchain/src/chat_models/openai.ts
```diff
@@ -94,7 +94,7 @@ function messageToOpenAIRole(message: BaseMessage): OpenAIRoleEnum {
 
 function messageToOpenAIMessage(message: BaseMessage): OpenAICompletionParam {
   const msg = {
-    content: message.content || null,
+    content: (message.content as any) || null,
```
bracesproul, line 97: Is this any typed because we do not know exactly what the types are yet?

langchain/src/schema/index.ts
```diff
@@ -101,8 +101,19 @@ export interface StoredGeneration {
 
 export type MessageType = "human" | "ai" | "generic" | "system" | "function";
 
+export type MessageContent =
+  | string
+  | (
+      | {
```
eyurtsev, line 107: i haven't written typescript -- so may have messed up the syntax... should this be something like:

{
          type: "text";
          text?: string;
} | {
          type: "image_url"; 
         image_url: string | { url: string };
}

langchain/src/chat_models/openai.ts
jacoblee93: Yeah just optimizing for speed

langchain/src/schema/index.ts
jacoblee93: Fixed

langchain/src/chat_models/openai.ts
jacoblee93: Will go back and fix before merging into `main` - linter warnings are good in this case

PR #3176, opened 2023-11-07
fix: updated convex chat message history indexes to match documentation

WrongGoose:
Updated the convex index type for ConvexChatMessageHistory as the basic ChatMessage history docs use a normal non-vector index for the session id.  I've included a [link](https://js.langchain.com/docs/integrations/chat_memory/convex) to the docs where the example chat memory setup is done.


jacoblee93:
Thanks!

PR #3189, opened 2023-11-08
Fix invalid usage of defaultProvider in Bedrock web.

Manouchehri:
Should match: https://github.com/langchain-ai/langchainjs/blob/c2169e9facf5953cd3eca92f529107eb5cd392c8/langchain/src/llms/bedrock/web.ts#L99

Otherwise, this error is thrown:

```
✘ [ERROR] Could not resolve "@aws-sdk/credential-provider-node"
```


jacoblee93:
Oh dear must have slipped through CC @dqbd

jacoblee93:
Thank you!

PR #3193, opened 2023-11-08
feat(pinecone): add support for delete by metadata filter

omikader:
This PR adds pass-through ability to delete records from Pinecone using metadata filtering which was [officially supported](https://sdk.pinecone.io/typescript/#md:delete-many-by-metadata-filter) in the v1 release of Pinecone's JS client.

This gives users the ability to delete vectors using matching metadata filters which can be more flexible than deleting by ID.


jacoblee93:
This is great, thank you!

PR #3194, opened 2023-11-08
Cleanup gen llm comp table script

bracesproul:


langchain/scripts/generate-docs-llm-compatibility-table.js
```diff
@@ -179,7 +180,6 @@ const checkClassMethods = async (
     all = await import(fullFilePath);
   } catch (error) {
     if (error.code === "ERR_MODULE_NOT_FOUND") {
-      console.warn(`Skipping file due to missing module: ${fullFilePath}`);
```
jacoblee93, line 182: Why do we need to look through `./dist` at all? There's not even a guarantee it'll be there

PR #3200, opened 2023-11-08
Brace/runnable chain

bracesproul:


langchain/src/chains/tests/llm_chain.int.test.ts
```diff
@@ -147,3 +147,20 @@ test("Test deserialize", async () => {
...
+
+  const prompt = PromptTemplate.fromTemplate(
+    "You are a bee --I mean a spelling bee.\nQuestion:{input}"
+  );
+
+  const chain = new LLMChain({ llm: runnableModel, prompt });
+
+  const response = await chain.invoke({ input: "How do you spell today?" });
+  console.log({ response });
+  expect(response).not.toContain("T-O-D-A-Y");
```
jacoblee93, line 165: Not to contain?

jacoblee93: Also wouldn't put an assertion this specific in an int test

langchain/src/chains/llm_chain.ts
```diff
@@ -130,33 +154,56 @@ export class LLMChain<
...
       }
     }
     const promptValue = await this.prompt.formatPromptValue(valuesForPrompt);
-    const { generations } = await this.llm.generatePrompt(
-      [promptValue],
-      valuesForLLM,
-      runManager?.getChild()
-    );
+    let response;
+    if (Runnable.isRunnable(this.llm)) {
```
jacoblee93, line 188: `BaseLanguageModel` is itself Runnable so this will always run

jacoblee93 (commented):
x

langchain/src/chains/tests/llm_chain.int.test.ts
bracesproul: Stop tokens are `T-O`.
but tbh, this test might be a little flaky. In like 5 runs I did it would always return something like `... T-O-D-A-Y` but if it were to return in another format this will fail...
If you have another idea on how we can check some bind method works I'm all ears.

jacoblee93: How about just a bound function call?

langchain/src/chains/llm_chain.ts
```diff
@@ -14,26 +22,56 @@ import {
...
   /** Kwargs to pass to LLM */
-  llmKwargs?: this["llm"]["CallOptions"];
+  llmKwargs?: CallOptionsIfAvailable<Model>;
   /** OutputParser to use */
   outputParser?: BaseLLMOutputParser<T>;
   /** Key to use for output, defaults to `text` */
   outputKey?: string;
 }
 
+function _getLanguageModel(llmLike: Runnable): BaseLanguageModel {
```
jacoblee93, line 57: I wouldn't try to do this actually, too brittle if we add more runnables

langchain/src/chains/llm_chain.ts
```diff
@@ -14,26 +22,56 @@ import {
...
+  } else if (llmLike instanceof RunnableBinding) {
+    return _getLanguageModel(llmLike.bound);
+    // eslint-disable-next-line no-instanceof/no-instanceof
+  } else if (llmLike instanceof RunnableWithFallbacks) {
+    return _getLanguageModel(llmLike.getRunnable);
+    // eslint-disable-next-line no-instanceof/no-instanceof
+  } else if (llmLike instanceof RunnableBranch) {
+    return _getLanguageModel(llmLike.default);
+  } else {
+    throw new Error("Unable to extract BaseLanguageModel from llmLie object.");
```
jacoblee93, line 71: `llmLike`

langchain/src/chains/base.ts
```diff
@@ -207,6 +207,12 @@ export abstract class BaseChain<
       );
     }
     await runManager?.handleChainEnd(outputValues);
+
+    // If a runnable is being used `outputValues` is
+    // not guaranteed to be an object
```
jacoblee93, line 212: We should guarantee this by only allowing runnables that emit objects? That's why we have static typing

langchain/src/chains/llm_chain.ts
```diff
@@ -130,40 +174,59 @@ export class LLMChain<
...
-    values: ChainValues & this["llm"]["CallOptions"],
+    values: ChainValues & CallOptionsIfAvailable<Model>,
     runManager?: CallbackManagerForChainRun
   ): Promise<ChainValues> {
     const valuesForPrompt = { ...values };
-    const valuesForLLM: this["llm"]["CallOptions"] = {
-      ...this.llmKwargs,
-    };
-    for (const key of this.llm.callKeys) {
+    console.log("values for prompt", valuesForPrompt);
```
jacoblee93, line 189: Remove

langchain/src/schema/runnable/base.ts
```diff
@@ -1408,6 +1408,10 @@ export class RunnableWithFallbacks<RunInput, RunOutput> extends Runnable<
     this.fallbacks = fields.fallbacks;
   }
 
+  get getRunnable() {
```
jacoblee93, line 1411: Call just `runnable`

jacoblee93 (changes requested):
Looks good overall! See comments

langchain/src/schema/runnable/base.ts
bracesproul: Can't since `runnable` is already defined above:
```typescript
  protected runnable: Runnable<RunInput, RunOutput>;
```

langchain/src/chains/base.ts
bracesproul: I believe we have to. If a user passes in an output parser which returns a string, then the result of `_call` in `LLMChain` will always be a string

langchain/src/chains/llm_chain.ts
```diff
@@ -14,26 +22,56 @@ import {
...
+  llmKwargs?: CallOptionsIfAvailable<Model>;
   /** OutputParser to use */
   outputParser?: BaseLLMOutputParser<T>;
   /** Key to use for output, defaults to `text` */
   outputKey?: string;
 }
 
+function _getLanguageModel(llmLike: Runnable): BaseLanguageModel {
+  // eslint-disable-next-line no-instanceof/no-instanceof
+  if (llmLike instanceof BaseLanguageModel) {
```
jacoblee93, line 59: Hey sorry one more thing - worried about regression since `instanceof` can get messed up by bundlers. Can we use a type guard here? I'm fine with the other instanceof checks

langchain/src/chains/llm_chain.ts
```diff
@@ -14,26 +22,51 @@ import {
...
-  llmKwargs?: this["llm"]["CallOptions"];
+  llmKwargs?: CallOptionsIfAvailable<Model>;
   /** OutputParser to use */
   outputParser?: BaseLLMOutputParser<T>;
   /** Key to use for output, defaults to `text` */
   outputKey?: string;
 }
 
+function _getLanguageModel(llmLike: Runnable): BaseLanguageModel {
+  if ("caller" in llmLike) {
```
jacoblee93, line 53: Unlikely but non-zero chance a runnable adds this as a prop

jacoblee93 (approved):

jacoblee93:
Ah this unfortunatley breaks signal passing - can you fix:

src/chains/tests/llm_chain.int.test.ts

jacoblee93:
Thanks!

PR #3207, opened 2023-11-09
Fix OpenAI Assistant docs

igo:


jacoblee93:
Thanks for the catch!

PR #3212, opened 2023-11-09
format prompt messages with openai image content

bracesproul:
Allow for formatting prompt strings when using openai image_url's in messages

### DIFF
--- a/langchain/src/prompts/chat.ts
+++ b/langchain/src/prompts/chat.ts
@@ -559,17 +559,58 @@ export class ChatPromptTemplate<
     return "chat";
   }
 
+  private async _parseImagePrompts(
+    message: BaseMessage,
+    inputValues: InputValues<
+      PartialVariableName | Extract<keyof RunInput, string>
+    >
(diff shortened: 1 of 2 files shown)


langchain/src/prompts/chat.ts
```diff
@@ -559,17 +559,58 @@ export class ChatPromptTemplate<
...
     let resultMessages: BaseMessage[] = [];
 
     for (const promptMessage of this.promptMessages) {
       // eslint-disable-next-line no-instanceof/no-instanceof
       if (promptMessage instanceof BaseMessage) {
-        resultMessages.push(promptMessage);
+        if (
+          typeof promptMessage.content !== "string" &&
+          promptMessage.content.some((content) => content.type === "image_url")
+        ) {
```
jacoblee93, line 607: I don't think we need a branch here? Looks like we can handle it all in one method and just return if it's a string?

bracesproul: Yes good call

PR #3213, opened 2023-11-09
feat(apify): support async dataset mapping function

omikader:
## Summary

This PR adds the ability to have an asynchronous dataset mapping function for the `ApifyDatasetLoader`.

## Context

Normally, the [Website Content Crawler](https://apify.com/apify/website-content-crawler) dataset result items look like this:

```
{
  "text": "This is the property that stores the pageContent of the scraped page",
  "url": "https://scraped-url.com",
  ...,
}
```

However, it also supports a mode called [`saveFiles`](https://apify.com/apify/website-content-crawler/input-schema) where the crawler is able to save any files it encounters during the web crawl. A dataset item corresponding to this kind of web-page looks like this:

```
{
  "text": null,
  "url": "https://scraped-url.com/uploads/super-important-document.pdf"
  ...,
}
```

Notice that the page content is not readily available. Instead, one must `fetch` the document content first and then use the convenient [`PDFLoader`](https://js.langchain.com/docs/integrations/document_loaders/file_loaders/pdf) to extract the text. However, doing this requires several `async` operations:
1. `fetch` the `response`
2. `response.blob()`
3. `loader.load()` the `PDFLoader`

Unfortunately, the `datasetMappingFunction` only supports the synchronous creation of documents from an Apify Dataset item. This PR allows the function to also support `Promise<Document<Metadata>>` by leveraging the [`AsyncCaller`](https://github.com/langchain-ai/langchainjs/blob/main/langchain/src/util/async_caller.ts#L85) -- similar to how @bracesproul  added support to the `PineconeStore` in https://github.com/langchain-ai/langchainjs/pull/3108.

### DIFF
--- a/langchain/src/document_loaders/web/apify_dataset.ts
+++ b/langchain/src/document_loaders/web/apify_dataset.ts
@@ -7,6 +7,7 @@ import {
   TaskCallOptions,
 } from "apify-client";
 
+import { AsyncCaller, AsyncCallerParams } from "../../util/async_caller.js";
 import { BaseDocumentLoader, DocumentLoader } from "../base.js";
 import { Document } from "../../document.js";
 import { getEnvironmentVariable } from "../../util/env.js";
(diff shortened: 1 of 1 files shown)


jacoblee93:
Looks good to me! Thank you

PR #3260, opened 2023-11-14
Meta Llama2 support for BedrockChat

### LINKED ISSUE #3257: Llama2 support for BedrockChat
[Amazon recently announced support for Meta's Llama2 13B parameter model in the Amazon Bedrock service](https://aws.amazon.com/blogs/aws/amazon-bedrock-now-provides-access-to-llama-2-chat-13b-model/). This is not yet supported in langchain, and being a new model provider there is no payload mapper either.

shafkevi:
This change will add in meta support for models in BedrockChat. Currently this is only the 13B parameter Llama2 model. So long as the model payload does not change, this will also work if the 70B model gets added.

The BedrockChat test suite has also been updated with convenient methods to more simply write tests for various new models.

Fixes #3257

### DIFF
--- a/langchain/src/chat_models/tests/chatbedrock.int.test.ts
+++ b/langchain/src/chat_models/tests/chatbedrock.int.test.ts
@@ -2,85 +2,162 @@
 /* eslint-disable @typescript-eslint/no-non-null-assertion */
 
 import { test, expect } from "@jest/globals";
-import { ChatBedrock } from "../bedrock/web.js";
+import { BedrockChat } from "../bedrock/web.js";
 import { HumanMessage } from "../../schema/index.js";
 
-test("Test Bedrock chat model: Claude-v2", async () => {
-  const region = process.env.BEDROCK_AWS_REGION ?? "us-east-1";
-  const model = "anthropic.claude-v2";
+void testChatModel(
+  "Test Bedrock chat model: Llama2 13B v1",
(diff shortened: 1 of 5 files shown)


Manouchehri:
Thanks, looks good to me!

langchain/src/chat_models/tests/chatbedrock.int.test.ts
```diff
@@ -2,85 +2,162 @@
...
 
 import { test, expect } from "@jest/globals";
-import { ChatBedrock } from "../bedrock/web.js";
+import { BedrockChat } from "../bedrock/web.js";
 import { HumanMessage } from "../../schema/index.js";
 
-test("Test Bedrock chat model: Claude-v2", async () => {
-  const region = process.env.BEDROCK_AWS_REGION ?? "us-east-1";
-  const model = "anthropic.claude-v2";
+void testChatModel(
```
jacoblee93, line 8: Can we do something like `test.each` below?

jacoblee93:
This is a great start! We'll want to add this to the LLM class as well. I can take it over and get it merged today.

shafkevi:
@jacoblee93 I just pushed changes for the LLM class, but if you plan to update the testing code or made alternate changes feel free to undo mine!

jacoblee93:
Thank you for this!

PR #3262, opened 2023-11-14
feat(apify): support Document[] return type for mapping function

omikader:
## Summary

This PR is a follow-up to https://github.com/langchain-ai/langchainjs/pull/3213 which enabled an asynchronous dataset mapping function.

## Context

This functionality is helpful because sometimes Apify dataset crawl items require asynchronous fetching/processing of the results from another source, e.g.
- download a file
- load the blob
- use the `PDFLoader`, `DocxLoader`, `CSVLoader`, etc.

However, those loader functions can return an array of `Document` which is incompatible with the current signature of `ApifyDatasetMappingFunction`. We can change the return type to be `Document<Metadata> | Array<Document<Metadata>>` to accommodate.

## Breaking Changes (N/A -- we decided to make it backwards compatible for now)

~~To keep things clean/simple, I removed the synchronous return type union as an option. Otherwise, the definition for `ApifyDatasetMappingFunction` would end up being the following -- which is a little hard to reason about, e.g.~~

```
export type ApifyDatasetMappingFunction<Metadata extends Record<string, any>> =
  (
    item: Record<string | number, unknown>
  ) =>
    | Document<Metadata>
    | Array<Document<Metadata>>
    | Promise<Document<Metadata> | Array<Document<Metadata>>>;
```

~~By only allowing the asynchronous return types, we can also simplify the invocation in `load` by dropping the extra asynchronous arrow function wrapper:~~

### Previous
```
const documentList = await Promise.all(
  dataset.items.map((item) =>
    this.caller.call(async () => this.datasetMappingFunction(item))
  )
);
```

### Current
```
const documentList = await Promise.all(
  dataset.items.map((item) =>
    this.caller.call(this.datasetMappingFunction, item)
  )
);
```

~~In doing this, folks who upgrade will need to explicitly annotate their `datasetMappingFunctions` as `async` which is a simple enough change -- but breaking nonetheless. I've updated the documentation to account for this.~~

~~Let me know if you think this is ok @jacoblee93 or if you'd rather I keep it backwards compatible. IMO it's worth the small break now.~~

### DIFF
--- a/examples/src/document_loaders/apify_dataset_existing.ts
+++ b/examples/src/document_loaders/apify_dataset_existing.ts
@@ -1,9 +1,9 @@
 import { ApifyDatasetLoader } from "langchain/document_loaders/web/apify_dataset";
 import { Document } from "langchain/document";
 import { HNSWLib } from "langchain/vectorstores/hnswlib";
+import { OpenAI } from "langchain/llms/openai";
 import { OpenAIEmbeddings } from "langchain/embeddings/openai";
 import { RetrievalQAChain } from "langchain/chains";
-import { OpenAI } from "langchain/llms/openai";
 
 /*
  * datasetMappingFunction is a function that maps your Apify dataset format to LangChain documents.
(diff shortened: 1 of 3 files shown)


jacoblee93:
Thanks for this and your continued updates here!

I think I'd prefer the backcompat yeah all things being equal - you can leave a note saying to break it in 0.1.0.

omikader:
@jacoblee93 it is now backwards compatible w/ comment 🙂

jacoblee93:
Thank you!

PR #3274, opened 2023-11-15
Brace/post git install

bracesproul:
Automatically install deps on pull if necessary

cc @nfcampos @jacoblee93

### DIFF
--- a/.husky/post-checkout
+++ b/.husky/post-checkout
@@ -1,0 +1,14 @@
+#!/usr/bin/env sh
+
+# If yarn install has never run, run it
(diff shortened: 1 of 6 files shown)


jacoblee93:
Will this run if we fetch/pull from contributor branches? A little worried about auto-running/installing docs without proper verification.

bracesproul:
> Will this run if we fetch/pull from contributor branches? A little worried about auto-running/installing docs without proper verification.

Ah yeah, I believe it will. Didn't think about that case...

This will run anytime any pull (merge, checkout, fetch etc) is made on the repo so it should in that case

jacoblee93:
I think a little too dangerous potentially but love the idea

PR #3292, opened 2023-11-16
fix: remove blank space between words when use pdf-loader

ppxu:
Extends #3218


jacoblee93:
I had a PDF that seemed to need this - see the test case.

There may be some better way to parse the loader output?

jacoblee93:
Ww could make it a config option?

ppxu:
i see, i always use pdf-loader to load chinese pdfs, so this question is very typical
i add a skipBlank option in pdf-loader for needed

jacoblee93:
Thanks for being persistent, will merge #3306 instead to allow potentially for other types of separators.

PR #3304, opened 2023-11-16
Add PrismaVectorStore filter IN operator

### LINKED ISSUE #3301: [Prisma] 'IN' operator in filter not available
Hello,

I'm currently working with the `PrismaVectorStore` class and have noticed that the `IN` operator is not available in the filter.
This will be useful for filtering on a key included in a set of values.

```
PrismaVectorStore.withModel(this.databaseService).create(
  new OpenAIEmbeddings(),
  {
    prisma: Prisma,
    tableName: 'Document',
    vectorColumnName: 'vector',
    columns: {
      content: PrismaVectorStore.ContentColumn,
    },
    filter: {
      foreignKeyId: {
        in: ['<id1>', '<id2>', ...],
      },
    },
  },
);
```

I can do the PR if needed ;)

thanks in advance !

Njuelle:
## Summary
Add support for `IN` operator in `PrismaVectorStore` class to allow filter on multiple possible values

```
PrismaVectorStore.withModel(this.databaseService).create(
  new OpenAIEmbeddings(),
  {
    prisma: Prisma,
    tableName: 'Document',
    vectorColumnName: 'vector',
    columns: {
      content: PrismaVectorStore.ContentColumn,
    },
    filter: {
      foreignKeyId: {
        in: ['<id1>', '<id2>', ...],
      },
    },
  },
);
```

The introduced code uses a `switch` instruction to make it easy to add other operators in the future.

Fixes #3301

### DIFF
--- a/docs/core_docs/docs/integrations/vectorstores/prisma.mdx
+++ b/docs/core_docs/docs/integrations/vectorstores/prisma.mdx
@@ -82,7 +82,7 @@ import Schema from "@examples/indexes/vector_stores/prisma_vectorstore/prisma/sc
 
 <CodeBlock language="typescript">{Example}</CodeBlock>
 
-Different SQL operators are available to perform filter: `equals`, `in`, `lt`, `lte`, `gt`, `gte`, `not`.
+The following SQL operators are available as filters: `equals`, `in`, `lt`, `lte`, `gt`, `gte`, `not`.
 
 The samples above uses the following schema:
 
--- a/docs/core_docs/docs/integrations/vectorstores/prisma.mdx
+++ b/docs/core_docs/docs/integrations/vectorstores/prisma.mdx
@@ -82,6 +82,8 @@ import Schema from "@examples/indexes/vector_stores/prisma_vectorstore/prisma/sc
 
 <CodeBlock language="typescript">{Example}</CodeBlock>
 
+Differents SQL operators are available to perform filter: `equals`, `lte`, `lt`, `gt`, `gte`, `not`.
+
 The samples above uses the following schema:
 
 <CodeBlock language="prisma">{Schema}</CodeBlock>
--- a/docs/core_docs/docs/integrations/vectorstores/prisma.mdx
+++ b/docs/core_docs/docs/integrations/vectorstores/prisma.mdx
@@ -82,7 +82,7 @@ import Schema from "@examples/indexes/vector_stores/prisma_vectorstore/prisma/sc
 
 <CodeBlock language="typescript">{Example}</CodeBlock>
 
-Differents SQL operators are available to perform filter: `equals`, `lte`, `lt`, `gt`, `gte`, `not`.
+Differents SQL operators are available to perform filter: `equals`, ìn`, `lte`, `lt`, `gt`, `gte`, `not`.
 
 The samples above uses the following schema:
 
--- a/docs/core_docs/docs/integrations/vectorstores/prisma.mdx
+++ b/docs/core_docs/docs/integrations/vectorstores/prisma.mdx
@@ -82,7 +82,7 @@ import Schema from "@examples/indexes/vector_stores/prisma_vectorstore/prisma/sc
 
 <CodeBlock language="typescript">{Example}</CodeBlock>
 
-Differents SQL operators are available to perform filter: `equals`, ìn`, `lte`, `lt`, `gt`, `gte`, `not`.
+Differents SQL operators are available to perform filter: `equals`, `in`, `lt`, `lte`, `gt`, `gte`, `not`.
 
 The samples above uses the following schema:
 
--- a/docs/core_docs/docs/integrations/vectorstores/prisma.mdx
+++ b/docs/core_docs/docs/integrations/vectorstores/prisma.mdx
@@ -82,7 +82,7 @@ import Schema from "@examples/indexes/vector_stores/prisma_vectorstore/prisma/sc
 
 <CodeBlock language="typescript">{Example}</CodeBlock>
 
-Differents SQL operators are available to perform filter: `equals`, `in`, `lt`, `lte`, `gt`, `gte`, `not`.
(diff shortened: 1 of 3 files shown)


langchain/src/vectorstores/prisma.ts
```diff
@@ -410,9 +412,18 @@ export class PrismaVectorStore<
...
           const colRaw = this.Prisma.raw(`"${key}"`);
-          const opRaw = this.Prisma.raw(OpMap[opName as keyof typeof OpMap]);
-          return this.Prisma.sql`${colRaw} ${opRaw} ${value}`;
+          const opRaw = this.Prisma.raw(OpMap[opNameKey]);
+
+          switch (OpMap[opNameKey]) {
+            case OpMap.in:
+              return this.Prisma.sql`${colRaw} ${opRaw} (${(
+                value as string[]
+              ).join(",")})`;
```
bracesproul, line 423: Instead of casting could we instead use a typeguard? Something like this
```suggestion
          const opRaw = this.Prisma.raw(OpMap[opNameKey]);

          switch (OpMap[opNameKey]) {
            case OpMap.in: {
              if (!Array.isArray(values) || !values.every((v) => typeof v === "string")) {
                throw new Error("...")
              }
              return this.Prisma.sql`${colRaw} ${opRaw} (${(
                value as string[]
              ).join(",")})`;
            }
```

bracesproul (commented):
Thank you for this! Left one small comment

bracesproul (commented):
Also, it would be very helpful to have a section in the [Prisma docs](https://js.langchain.com/docs/integrations/vectorstores/prisma) showing this is possible

Njuelle:
@bracesproul 
I updated the code example in the documentation and added a tiny section about operators. Let me know if it's ok for you

langchain/src/vectorstores/prisma.ts
```diff
@@ -410,9 +412,25 @@ export class PrismaVectorStore<
...
+          const opRaw = this.Prisma.raw(OpMap[opNameKey]);
+
+          switch (OpMap[opNameKey]) {
+            case OpMap.in: {
+              if (
+                !Array.isArray(value) ||
+                !value.every((v) => typeof v === "string")
+              ) {
+                throw new Error(
+                  "Invalid filter: IN operator requires an array of strings"
```
bracesproul, line 426:
```suggestion
                  `Invalid filter: IN operator requires an array of strings. Received: ${JSON.stringify(value, null, 2)}`
```

examples/src/indexes/vector_stores/prisma_vectorstore/prisma.ts
```diff
@@ -58,11 +58,14 @@ export const run = async () => {
   const resultTwo = await vectorStore.similaritySearch("Hello world", 1);
   console.log(resultTwo);
 
-  // Override the local filter
+  // Override the local filter,
```
bracesproul, line 61:
```suggestion
  // Override the local filter
```

bracesproul (commented):
lgtm after these two comments thank you for this!, cc @jacoblee93 for a second look

langchain/src/vectorstores/prisma.ts
```diff
@@ -410,9 +412,29 @@ export class PrismaVectorStore<
         Object.entries(ops).map(([opName, value]) => {
           // column name, operators cannot be parametrised
           // these fields are thus not escaped by Prisma and can be dangerous if user input is used
+          const opNameKey = opName as keyof typeof OpMap;
           const colRaw = this.Prisma.raw(`"${key}"`);
-          const opRaw = this.Prisma.raw(OpMap[opName as keyof typeof OpMap]);
-          return this.Prisma.sql`${colRaw} ${opRaw} ${value}`;
+          const opRaw = this.Prisma.raw(OpMap[opNameKey]);
+
+          switch (OpMap[opNameKey]) {
```
jacoblee93, line 419: Why not just use `opRaw` here?

docs/core_docs/docs/integrations/vectorstores/prisma.mdx
```diff
@@ -82,6 +82,8 @@ import Schema from "@examples/indexes/vector_stores/prisma_vectorstore/prisma/sc
 
 <CodeBlock language="typescript">{Example}</CodeBlock>
 
+Differents SQL operators are available to perform filter: `equals`, `in`, `lt`, `lte`, `gt`, `gte`, `not`.
```
jacoblee93, line 85: Typo

jacoblee93:
LGTM!

langchain/src/vectorstores/prisma.ts
```diff
@@ -410,9 +412,29 @@ export class PrismaVectorStore<
...
+            case OpMap.in: {
+              if (
+                !Array.isArray(value) ||
+                !value.every((v) => typeof v === "string")
```
eduardolundgren, line 423: The proposed implementation only supports string. The fix can be generalized to support other types.

```js
buildSqlFilterStr(filter?: any) {
  if (filter == null) return null;
  // @ts-ignore
  const sql = this.Prisma.join(
    Object.entries(filter).flatMap(([key, ops]) =>
      // @ts-ignore
      Object.entries(ops).map(([opName, value]) => {
        // Patch 1 starts here (support enums)
        const cast = typeof value === 'string' ? '::text' : '';
        // Patch 1 ends here
        // @ts-ignore
        const colRaw = this.Prisma.raw(`"${key}"${cast}`);
        // @ts-ignore
        const opRaw = this.Prisma.raw(OpMap[opName as keyof typeof OpMap]);
        // Patch 2 starts here (supports all types)
        if (opName === 'in') {
          // @ts-ignore
          if (value.length === 0) {
            // @ts-ignore
            value.push(0);
          }
          // @ts-ignore
          return this.Prisma.sql`${colRaw} ${opRaw} (${this.Prisma.join(value)})`;
        }
        // Patch 2 ends here
        // @ts-ignore
        return this.Prisma.sql`${colRaw} ${opRaw} ${value}`;
      })
    ),
    ' AND ',
    ' WHERE '
  );
  return sql;
};
```

Issue #3297, opened 2023-11-16
OpenSearch vector store as retriever assumes the filters are always singular "term"


faileon:
Hi all,
I am trying to implement a filtering capability in my RAG application. I use OpenSearch to store embeddings. Each embedding has a metadata that contain `doc_id`. I can easily filter for embeddings that have specific `doc_id`, but I cannot filter for multiple ids.

The way to do filter for multiple values in OpenSearch is:
```json
{
  "query": {
    "bool": {
      "filter": [
        {
          "terms": {
            "metadata.doc_id": [
              "id1",
              "id2"
            ]
          }
        }
      ]
    }
  }
}
```

To filter singular values in OpenSearch:
```json
{
  "query": {
    "bool": {
      "filter": [
        {
          "term": {
            "metadata.doc_id": "id1"
          }
        }
      ]
    }
  }
}
```

Note the difference between `term` and `terms` and the usage of array.

Langchain's [implementation](https://github.com/langchain-ai/langchainjs/blob/main/langchain/src/vectorstores/opensearch.ts#L293) always uses `term`.

I took a stab in the dark and created my class overriding the method in following way:

```ts
override buildMetadataTerms(
    filter?: OpenSearchFilter
  ): { term: Record<string, unknown> }[] {
    if (filter == null) return [];
    const result = [];
    for (const [key, value] of Object.entries(filter)) {
            const isArray = Array.isArray(value);
            const aggregatorKey = isArray ? 'terms' : 'term';
            result.push({ [aggregatorKey]: { [`metadata.${key}`]: value } });
    }
    return result;
  }
```

I can submit a PR if the solution is OK with y'all.

PR #3312, opened 2023-11-17
feat: Added support of terms filter in OpenSearch vector store

faileon:
Hi,
solves #3297 

In the future it can be extended further to implement other [term-level queries](https://opensearch.org/docs/latest/query-dsl/term/index/), but for now `term` and `terms` is sufficient enough, as they are imho the most used.

Since I am primarily working with OpenSearch in my workplace, I'll gladly add those features when the need arises.


jacoblee93:
Thank you!

PR #3314, opened 2023-11-17
feat: add filters to `ChaindeskRetriever`

antoinewg:
reference: https://docs.chaindesk.ai/api-reference/endpoint/datastores/query

cc @gmpetrov

### DIFF
--- a/langchain/src/retrievers/chaindesk.ts
+++ b/langchain/src/retrievers/chaindesk.ts
@@ -7,6 +7,7 @@ export interface ChaindeskRetrieverArgs
     BaseRetrieverInput {
   datastoreId: string;
   topK?: number;
+  filters?: Record<string, unknown>;
   apiKey?: string;
 }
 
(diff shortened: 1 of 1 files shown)


langchain/src/retrievers/chaindesk.ts
```diff
@@ -7,6 +7,7 @@ export interface ChaindeskRetrieverArgs
     BaseRetrieverInput {
   datastoreId: string;
   topK?: number;
+  filters?: Record<string, unknown>;
```
jacoblee93, line 10: We generally call them just `filter` (no s)

jacoblee93:
Seems good and backwards compatible to me, see comment!

jacoblee93:
Can you run `yarn format`?

antoinewg:
should be good now @jacoblee93 :)

jacoblee93:
Thank you!

PR #3318, opened 2023-11-17
Improve MessageContent type

netzhuffle:
The fields are mutually exclusive. This represents this in its type.

This will help with both avoiding wrong combinations as well as reduces unnecessary undefined checks.

### DIFF
--- a/langchain/src/schema/index.ts
+++ b/langchain/src/schema/index.ts
@@ -118,7 +118,7 @@ export type MessageContent =
         }
       | {
           type: "image_url";
-          image_url: string | { url: string; detail?: "low" | "high" };
+          image_url: string | { url: string; detail?: "auto" | "low" | "high" };
         }
     )[];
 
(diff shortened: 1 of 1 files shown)


hinthornw (approved):

langchain/src/schema/index.ts
```diff
@@ -111,11 +111,16 @@ export type MessageType =
...
-      image_url?: string | { url: string; detail?: "low" | "high" };
-    }[];
+  | (
+      | {
+          type: "text";
+          text: string;
+        }
+      | {
+          type: "image_url";
+          image_url: string | { url: string; detail?: "low" | "high" };
```
bracesproul, line 121:
```suggestion
          image_url: string | { url: string; detail?: "auto" | "low" | "high" };
```
just realized we're also missing this if someone wants to declare it

jacoblee93:
Thank you!

PR #3322, opened 2023-11-17
Remove console.log from googlevertexai-connection.ts

raioalbano:
Remove console.log from googlevertexai-connection.ts

Problem:
- In case of invalid address, an error is logged exposing authentication token  in logs.
- this log should be handled by langchain callback mechanism.

Fixes # (issue)


bracesproul (commented):
thank you for this! Looks like the lint is failing, if you could run `yarn lint && yarn format` from `langchainjs/langchain` and push up the changes that would be great

jacoblee93:
Thank you!

PR #3332, opened 2023-11-18
Updates to Vectara Implementation

ofermend:
- Updated random doc_id generation (if not provided)
- Some tests were skipped, so made them active and retested
- Added deleteDocuments() and updated testing to remove documents after testing is done.

Twitter: @ofermend


langchain/src/vectorstores/tests/vectara.int.test.ts
```diff
@@ -94,7 +89,7 @@ describe("VectaraStore", () => {
       apiKey: process.env.VECTARA_API_KEY || "",
     };
 
-    test.skip("with fakeEmbeddings doesn't throw error", () => {
+    test("with fakeEmbeddings doesn't throw error", () => {
```
jacoblee93, line 92: Can you leave the .skip in? We don't run every integration test every time, they're for testing on an as-needed basis.

jacoblee93:
Seems reasonable! Remember to run `yarn format` and `yarn lint`. Also might want some docs for delete?

jacoblee93:
Thanks!

PR #3336, opened 2023-11-19
Update Ollama functions

jacoblee93:


sabatale:
Maybe not the best place to ask but are there any plans to support "auto" value for function_call?

```
      if (!functions.length) {
        throw new Error(
          `If "function_call" is specified, you must also pass a matching function in "functions".`
        );
      }
```

jacoblee93:
Would be a good add! Planning to do some Ollama updates next week after the big merge of #3610 so may have some time then.

Issue #2555, opened 2023-09-07
Add self-query support for Vectara


ofermend:
Add support for Self-query for the Vectara integration.
Vectara's [filtering](https://docs.vectara.com/docs/common-use-cases/filtering-by-metadata/filter-overview) is simply a string in pseudo-SQL format. Meta-data fields exists at the "doc" level and "part" level. Part refers to a piece of the document.
For self-querying we would like the filter to apply to the "doc" level metadata.

The python implementation PR is: https://github.com/langchain-ai/langchain/pull/10332 for reference.
@jacoblee93.

PR #3338, opened 2023-11-19
Adding self query for vectara

adeelehsan:
Add support for Self-query for the Vectara integration.
Vectara's [filtering](https://docs.vectara.com/docs/common-use-cases/filtering-by-metadata/filter-overview) is simply a string in pseudo-SQL format. Meta-data fields exists at the "doc" level and "part" level. Part refers to a piece of the document.
For self-querying we would like the filter to apply to the "doc" level metadata.

Fixes # ([2555](https://github.com/langchain-ai/langchainjs/issues/2555))


langchain/src/retrievers/self_query/vectara.ts
```diff
@@ -0,0 +1,181 @@
...
+import { isFilterEmpty } from "./utils.js";
+
+type AllowedOperator = Exclude<Operator, NOT>;
+
+export type VectaraVisitorResult =
+  | VectaraOperationResult
+  | VectaraComparisonResult
+  | VectaraVisitorStructuredQueryResult;
+// eslint-disable-next-line @typescript-eslint/ban-types
+export type VectaraOperationResult = String;
```
jacoblee93, line 23: This doesn't cause problems somewhere down the road?

jacoblee93:
Thanks @adeelehsan! `String` type hack is a bit suspect but if it works I am happy to merge.

adeelehsan:
> Thanks @adeelehsan! `String` type hack is a bit suspect but if it works I am happy to merge.

@jacoblee93 thanks for reviewing it. 
Yeah it works. `Vectara` is expecting filters as a string instead of object.
Default implementation of  methods `operation` and `comparison` methods are expecting return type object thats why I have defined the String as object.

  You can merge it.

jacoblee93:
Thanks!

PR #3339, opened 2023-11-19
Allow to stream files with `GithubRepoLoader`

### LINKED ISSUE #3331: Allow to load Github documents using stream
Hey,

Since I tried to load documents from a large Github repository using `GithubRepoLoader`. I ran on `JavaScript heap out of memory` error.
It could be nice to allow to load Github documents using stream

I started to work on a PR that allow this behaviour using stream with generators.

Proposed usage:

```js
const loader = new GithubRepoLoader(
  "https://github.com/langchain-ai/langchainjs",
  {
    branch: "main",
    recursive: false,
    unknown: "warn",
    maxConcurrency: 3, // Defaults to 2
  }
);

const docs = [];
for await (const doc of loader.loadAsStream()) {
  docs.push(doc);
}

console.log({ docs });
```

wdyt?

Njuelle:
## Summary
Allow to stream Github files with `GithubRepoLoader`, using generators. 
Ths is useful on larges repositories to process them in a memory efficient manner.

Usage:
```typescript
const loader = new GithubRepoLoader(
  "https://github.com/langchain-ai/langchainjs",
  {
    branch: "main",
    recursive: false,
    unknown: "warn",
    maxConcurrency: 3,
  }
);

const docs = [];
for await (const doc of loader.loadAsStream()) {
  docs.push(doc);
}

console.log({ docs });
```

Fixes #3331

### DIFF
--- a/langchain/src/document_loaders/web/github.ts
+++ b/langchain/src/document_loaders/web/github.ts
@@ -255,6 +255,27 @@ export class GithubRepoLoader
   }
 
   /**
+   * Asynchronously streams documents from the entire GitHub repository.
+   * It is suitable for situations where processing large repositories in a memory-efficient manner is required.
+   * @yields Yields a Promise that resolves to a Document object for each file or submodule content found in the repository.
+   */
+  public async *loadAsStream(): AsyncGenerator<Document, void, undefined> {
+    this.log(
+      `Loading documents from ${this.baseUrl}/${this.owner}/${this.repo}/${this.initialPath}...`
+    );
+    yield* await this.processRepoAsStream(this.initialPath);
+
+    if (!this.processSubmodules) {
+      return;
+    }
+
+    await this.getSubmoduleInfo();
+    for (const submoduleInfo of this.submoduleInfos) {
+      yield* await this.loadSubmoduleAsStream(submoduleInfo);
+    }
+  }
+
+  /**
    * Loads the information about Git submodules from the repository, if available.
    */
   private async getSubmoduleInfo(): Promise<void> {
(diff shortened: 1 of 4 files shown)


langchain/src/document_loaders/web/github.ts
```diff
@@ -503,6 +601,41 @@ export class GithubRepoLoader
...
+          yield new Document({
+            pageContent: fileResponse.contents,
+            metadata: fileResponse.metadata,
+          });
+        } catch {
+          this.handleError(`Failed to fetch file content: ${file.path}`);
+        }
+      }
+
+      if (this.recursive) {
```
jacoblee93, line 633: Should this be an `else` with the above? Basically if it's not `type === "file"`

langchain/src/document_loaders/web/github.ts
```diff
@@ -485,6 +547,42 @@ export class GithubRepoLoader
...
+          yield new Document({
+            pageContent: fileResponse.contents,
+            metadata: fileResponse.metadata,
+          });
+        } catch (error) {
+          this.handleError(
+            `Failed to fetch file content: ${file.path}, ${error}`
+          );
+        }
+      }
```
jacoblee93, line 578: See below

jacoblee93:
This is great! See comments.

Also, we should maybe look into moving the standard implementation to use this flow internally and gather streamed docs so as not to have dupes, but that can be later.

langchain/src/document_loaders/web/github.ts
```diff
@@ -254,6 +254,27 @@ export class GithubRepoLoader
     return documents;
   }
 
+  /**
+   * Asynchronously streams documents from the entire GitHub repository.
+   * It is suitable for situations where processing large repositories in a memory-efficient manner is required.
+   * @yields Yields a Promise that resolves to a Document object for each file or submodule content found in the repository.
+   */
+  public async *loadAsStream(): AsyncGenerator<Document, void, undefined> {
```
jacoblee93, line 262: You could have this return an `IterableReadableStream`:

https://github.com/langchain-ai/langchainjs/blob/05b0a4ff747baf481c4c5aa5d8e4e7730cc38a44/langchain/src/schema/runnable/base.ts#L226

langchain/src/document_loaders/web/github.ts
Njuelle: in general, it's easier to read code without `else if`, which is why I've chosen this method here.
But I can change it if you prefer

langchain/src/document_loaders/web/github.ts
Njuelle: It's not a `ReadableStream` here, but a stream with Async Generators.
like here:   https://github.com/langchain-ai/langchainjs/blob/main/langchain/src/chat_models/openai.ts#L450

langchain/src/document_loaders/web/github.ts
jacoblee93: But if `this.recursive` is true, for files it'll 

```
          yield new Document({
            pageContent: fileResponse.contents,
            metadata: fileResponse.metadata,
          });
```

And then also

```
yield* await this.processDirectoryAsStream(file.path);
```

Njuelle: fixed!

PR #3346, opened 2023-11-20
Update OpenAIAgent to support Runnable models

### LINKED ISSUE #3344: ChatOpenAI JSON mode errors in agent chain: llm.predictMessages is not a function
I'm binding a `ChatOpenAI` model with JSON mode as specified in the [docs](https://js.langchain.com/docs/integrations/chat/openai):
```ts
    modelName: "gpt-4-1106-preview",
  }).bind({
    response_format: {
      type: "json_object",
    },
  });
```

I'm then using this model in a chain as part of an agent:
```ts
const chain = new LLMChain({
  prompt: ChatPromptTemplate.fromTemplate(
    "Return the current weather. Use JSON format."
  ),
  llm: model,
});
const agent = new OpenAIAgent({
  llmChain: chain,
  allowedTools: tools.map((tool) => tool.name),
  tools,
});
const executor = new AgentExecutor({
  agent,
  tools,
  verbose: true,
  maxIterations: 10,
});
```

After calling the executor, however, I'm seeing this error: `TypeError: llm.predictMessages is not a function`.

I found this issue (#3226) and its fix (#3200) which patches a similar error for `llm.callKeys`, but it doesn't look like the [`predictMessages` call in the agent class](https://github.com/langchain-ai/langchainjs/blob/main/langchain/src/agents/openai/index.ts#L219-L223) was handled yet.

gramliu:
Handles the case when an `OpenAIAgent` is called with `.plan` on an `llmChain` instantiated with a `Runnable` instead of a regular chat model. Currently, this case is not handled and results in an error: `TypeError: llm.predictMessages is not a function`. This resolves the bug by checking if it's runnable and appropriately calling `runnable.invoke`

Fixes #3344

### DIFF
--- a/langchain/src/agents/openai/index.ts
+++ b/langchain/src/agents/openai/index.ts
@@ -232,24 +232,16 @@ export class OpenAIAgent extends Agent {
     const promptValue =
       await this.llmChain.prompt.formatPromptValue(valuesForPrompt);
 
-    let message: BaseMessage;
-    if (Runnable.isRunnable(llm)) {
-      message = await (
-        llm as Runnable<
-          BaseLanguageModelInput,
-          BaseMessageChunk,
-          ChatOpenAICallOptions
-        >
-      ).invoke(promptValue.toChatMessages(), {
-        callbacks: callbackManager,
-      });
-    } else {
-      message = await (llm as ChatOpenAI).predictMessages(
-        promptValue.toChatMessages(),
-        valuesForLLM,
-        callbackManager,
-      );
-    }
+    const message = await (llm as 
+      Runnable<
+        BaseLanguageModelInput,
+        BaseMessageChunk,
+        ChatOpenAICallOptions
+      >
+    ).invoke(promptValue.toChatMessages(), {
+      ...valuesForLLM,
+      callbacks: callbackManager,
+    });
     return this.outputParser.parseAIMessage(message);
   }
 }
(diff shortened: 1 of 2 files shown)


jacoblee93:
Great update, thank you!

I think we may be able to use `.invoke` in both cases, actually, might check that out.

langchain/src/agents/openai/index.ts
```diff
@@ -199,28 +207,49 @@ export class OpenAIAgent extends Agent {
...
-    const message = await llm.predictMessages(
-      promptValue.toChatMessages(),
-      valuesForLLM,
-      callbackManager
-    );
+    const promptValue =
+      await this.llmChain.prompt.formatPromptValue(valuesForPrompt);
+
+    let message: BaseMessage;
+    if (Runnable.isRunnable(llm)) {
```
bracesproul, line 236: All llms extend runnables so this if statement won't work in the way it's intended to.

Instead you should be checking for a class property which would identify the openai LLM, and else would be runnable

bracesproul (changes requested):

langchain/src/agents/openai/index.ts
jacoblee93: Good point, I think we don't need the branch at all then if it's still working. Will remove.

bracesproul (commented):
lgtm, cc @jacoblee93

jacoblee93:
We should refactor to not use an `llmChain` here but I think this is fine for now

Issue #2699, opened 2023-09-22
Enhancement Request: Upgrade Filter Functionality in VectorStore Wrapping Elasticsearch


shotaro427:
## Issue Description:
I would like to propose an upgrade to the filter functionality in the VectorStore that wraps Elasticsearch. Specifically, I'd like to make it as flexible as the Python package langchain in terms of filtering capabilities.

## Current State:
Currently, the filtering is limited to using only the "term" operator. This is handled in the function buildMetadataTerms located in src/vectorstores/elasticsearch.ts.

## Requested Feature:
I'd like to request an enhancement that would allow for more flexible filtering options, similar to what is available in the langchain Python package.

## Additional Context:
Is there a specific reason that we are limited to using only the "term" operator? Any context or historical reasons for this limitation would be appreciated as it could inform the development of this feature.

## Next Steps:
If specific implementation strategies are required, please feel free to propose them in the discussion.

shotaro427:
When will this feature be released?
If you are looking for my own commitments, can you give me the contributor rules?

dqbd:
Hey @shotaro427!

Contributions are more than welcome here! Feel free to check out the contributing guidelines here: https://github.com/langchain-ai/langchainjs/blob/main/CONTRIBUTING.md

PR #3349, opened 2023-11-20
langchain[patch]: Enhance filter functionality for Elasticsearch VectorStore

LaurensTsestigh:
Based on an existing issue of the langchain project, I created a PR with the requested enhancement as we need this as well in our product where we use the langchain package: https://github.com/langchain-ai/langchainjs/issues/2699

As stated by the dosubot:

> With this modification, we could use any operator supported by Elasticsearch in our filters, not just the "term" operator.
> As for the different types of operators supported by Elasticsearch, they include but are not limited to: "term", "terms", "range", "exists", "missing", "prefix", "wildcard", "regexp", "fuzzy", "type", "ids", "geo_distance", "geo_bounding_box", "geo_polygon", "geo_distance_range", "geo_shape", and "script".

Thanks in advance


langchain/src/vectorstores/elasticsearch.ts
```diff
@@ -38,7 +38,7 @@ export interface ElasticClientArgs {
 /**
  * Type representing a filter object in Elasticsearch.
  */
-type ElasticFilter = object;
+type ElasticFilter = { field: string, operator: string, value: any }[];
```
jacoblee93, line 41: This looks breaking? Shouldn't this allow `term` as a property?

LaurensTsestigh: Hi jacoblee, thanks for taking your time to check this PR out.

With the `operator` property we can be more flexible and allow for "term", "terms", "range", "exists", etc. instead.
As I typed it as an array of objects, we can have multiple different filters applied.
The breaking part seems to be in the docs, but it seems I can't change those?

jacoblee93: Ah got it, yeah looks like it might have just been a placeholder.

jacoblee93:
Thanks! Can you merge and run `yarn format` and `yarn lint`

langchain/src/vectorstores/elasticsearch.ts
```diff
@@ -299,11 +299,11 @@ export class ElasticVectorSearch extends VectorStore {
 
   private buildMetadataTerms(
     filter?: ElasticFilter
-  ): { term: Record<string, unknown> }[] {
+  ): { [operator: string]: { [field: string]: any } }[] {
     if (filter == null) return [];
     const result = [];
-    for (const [key, value] of Object.entries(filter)) {
-      result.push({ term: { [`metadata.${key}`]: value } });
+    for (const condition of filter) {
```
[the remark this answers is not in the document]
LaurensTsestigh: Yes you are right, we can't go breaking people's code. 
I've adapted it to be completely backwards-compatible

jacoblee93:
Perfect, thank you!

LaurensTsestigh:
> Perfect, thank you!

I've updated the PR with main and saw the dosubot removed the 'lgtm' label

jacoblee93:
Ah, yeah if you push another commit I have to retrigger CI so please don't do that!

LaurensTsestigh:
> Ah, yeah if you push another commit I have to retrigger CI so please don't do that unless there's a conflict!

I'll keep it in mind 😄 

Thanks a lot, have a nice day!

PR #3353, opened 2023-11-20
feat: add support for collection name in PG Vector

devinburnette:
## Description
This PR adds support for collection names in the PG Vector store integration that currently [already exists](https://github.com/langchain-ai/langchain/blob/master/libs/langchain/langchain/vectorstores/pgvector.py#L113) in the python version of langchain. Collections add a nice way to logically group document embeddings in the vector store. My specific use case is around privilege preservation of certain types of documents. I'm creating and loading my embeddings table from the python version, but I'd like to be able to query it using the javascript version.

The gist of it is there's another table added and a foreign key referencing the uuid from the collections table. My goal is to make this a seamless addition by not causing any breaking changes to anyone not already using collections. So I updated the creation and insert logic to conditionally support the collection feature also, not just the query logic. Let me know if there's a better way to introduce this change or an easier approach.

## Alternatives Considered
I thought about using a filter, however filters are currently implemented in a way that only applies to metadata. So then I thought about abandoning collections altogether and using some field of the metadata, but I found that the metadata fields are too inconsistent across different document loaders for that to be a valuable approach. Plus I think the goal of this library is to eventually reach feature parity with the python version.

## Testing
To test this, I ran a local build with my changes implemented into my client. I also ran a modified version of the pgvector example in `examples/src/indexes/vector_stores/pgvector_vectorstore/pgvector.ts` to make sure that each branch of logic worked the way that I would expect it to.

### DIFF
--- a/langchain/src/vectorstores/pgvector.ts
+++ b/langchain/src/vectorstores/pgvector.ts
@@ -149,7 +149,9 @@ export class PGVectorStore extends VectorStore {
       SELECT uuid from ${this.collectionTableName}
       WHERE name = $1;
     `;
-    const queryResult = await this.pool.query(queryString, [this.collectionName]);
+    const queryResult = await this.pool.query(queryString, [
+      this.collectionName,
+    ]);
     let collectionId = queryResult.rows[0]?.uuid;
 
     if (!collectionId) {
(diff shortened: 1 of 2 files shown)


jacoblee93:
Awesome, thank you!

PR #3354, opened 2023-11-21
Jacob/core

jacoblee93:


langchain-core/src/runnables/tests/base_message.test.ts
bracesproul, line 1: Looks like this is in the wrong dir?

Issue #3360, opened 2023-11-21
Add option to only allow specified file types in Github Loader


JamDevCore:
My idea is to add a new parameter to the github loader which will contain an array of allowed file types e.g ["js", "html", "ts"] - currently this process is arduous as you have to select which files/paths you *don't* want (which could be a lot) rather than what you *do* want.

E.g

```
protected shouldIgnore(path: string, fileType: string): boolean {
    if (fileType !== "dir" && isBinaryPath(path)) {
      return true;
    }
    if (this.ignore !== undefined) {
      return this.ignore.ignores(path);
    }
    if(this.onlyAllowFileTypes && this.onlyAllowFileTypes.length > 0) {
      return !this.onlyAllowFileTypes.includes(fileType);
    }
    return (
      fileType !== "dir" &&
      this.ignoreFiles.some((pattern) => {
        if (typeof pattern === "string") {
          return path === pattern;
        }
        try {
          return pattern.test(path);
        } catch {
          throw new Error(`Unknown ignore file pattern: ${pattern}`);
        }
      })
    );
  }
```

JamDevCore:
I have a PR ready to go https://github.com/langchain-ai/langchainjs/pull/3361

PR #3361, opened 2023-11-21
Adds option to only allow specified file types

JamDevCore:
Adds a new parameter to github loader which will contain an array of allowed file types e.g ["js", "html", "ts"] - currently this process is arduous as you have to select which files/paths you *don't* want (which could be a lot) rather than what you *do* want.

Fixes # (issue)

https://github.com/langchain-ai/langchainjs/issues/3360


jacoblee93:
Thanks for this - is there a way we could make this more generalizable? Like "only allow certain patterns"?

benjamincburns:
Cleaning up old/outdated PRs. Sorry we didn't get to this one in time!

PR #3368, opened 2023-11-22
feat: update version of closevector-web and its usage

jo32:
Hi, I'm currently updating the version of CloseVector-Web. This update is necessary because the previous version requires setting a private attribute _uuid.

### DIFF
--- a/langchain/package.json
+++ b/langchain/package.json
@@ -1025,7 +1025,7 @@
     "chromadb": "*",
     "closevector-common": "0.1.0-alpha.1",
     "closevector-node": "0.1.0-alpha.10",
-    "closevector-web": "0.1.0-alpha.16",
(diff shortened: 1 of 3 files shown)


langchain/package.json
```diff
@@ -1025,7 +1025,7 @@
     "chromadb": "*",
     "closevector-common": "0.1.0-alpha.1",
     "closevector-node": "0.1.0-alpha.10",
-    "closevector-web": "0.1.0-alpha.16",
+    "closevector-web": "0.1.0-alpha.18",
```
jacoblee93, line 1028: Can you update the devdependency too?

jo32: oh yes, of course.

jo32:
@jacoblee93 deps are updated, thank you for reminding that.

PR #3370, opened 2023-11-22
fix: add maxRetries parameter to openAI params (so that timed out requests can be retried)

### LINKED ISSUE #2706: Retry logic for OpenAI timeouts
I'm seeing the following error in prod:

```
Error [TimeoutError]: Request timed out.
    at wrapOpenAIClientError (file:///app/node_modules/langchain/dist/util/openai.js:6:17)
    at file:///app/node_modules/langchain/dist/chat_models/openai.js:518:31
    at async RetryOperation._fn (/app/node_modules/p-retry/index.js:50:12) {
  attemptNumber: 1,
  retriesLeft: 6
}
```

It's getting captured in my catch block, so I'm fairly sure the retries aren't happening, unless the first attempt is the one that gets re-thrown or something confusing like that. Is it possible this doesn't meet the criteria for retryable? Could this be addressed using the FailedAttemptHandler interface?

I've not set a timeout for the LLM. I'm having a hard time figuring out the default value.

```
  /**
   * error is not retryable.
   */
  onFailedAttempt?: FailedAttemptHandler;
  ```

### PR DESCRIPTION
**Pull request by codenameakshay**
Fixes #2706

### CODE REVIEW

File: langchain/src/chat_models/openai.ts [hunk 1]
```diff
@@ -778,7 +781,7 @@ export class ChatOpenAI<
         ...this.clientConfig,
         baseURL: endpoint,
         timeout: this.timeout,
-        maxRetries: 0,
```
jacoblee93, line 781: This should already be handled by the `AsyncCaller` class and tested - is something not working?

https://github.com/langchain-ai/langchainjs/blob/main/langchain-core/src/language_models/base.ts#L168
https://github.com/langchain-ai/langchainjs/blob/main/langchain-core/src/utils/async_caller.ts#L49

codenameakshay: I am getting a lot of `Request timed out` for the new `gpt-3.5-turbo-1106` model, and whenever that happens (I specify the maxRetries), it never retries. So I thought that maybe that maxRetries was not getting passed through, and as in #2706 changing AsyncCaller is not feasible, as it is being used at other places as well.

jacoblee93: Oh, I actually think it'd probably a better idea to retry timeouts there?

jacoblee93: Ok @codenameakshay, checked in. The reason we don't retry on timeouts is that you could set a low one and retrying would defeat the purpose.

Would it make sense to try setting a higher timeout instead? Or is it already taking a lot longer than expected?

jacoblee93: We may not be catching timeout errors properly - will look into it.

codenameakshay: I set a timeout value of 90000, that is 90 seconds.

PR #3377, opened 2023-11-22
Adds standalone @langchain/anthropic package

jacoblee93:


.github/workflows/test-exports.yml
```diff
@@ -134,7 +134,7 @@ jobs:
       - name: Install dependencies
         run: yarn install --immutable
       - name: Build
-        run: yarn workspace langchain-core build && yarn workspace langchain build
+        run: yarn workspace langchain-core build && yarn workspace @langchain/anthropic build && yarn workspace langchain build
```
nfcampos, line 137: should we add this command to package json?

jacoblee93: Yes probably, will do when I do OpenAI

PR #3389, opened 2023-11-23
fix: Add HuggingFaceInference includeCredentials param

Xelaan:
IncludeCredentials parameter added to HuggingFaceInference to address the issue "The 'credentials' field on 'RequestInitializerDict' is not implemented." when using HuggingFace's HfInference on Cloudflare Workers.


jacoblee93:
Thank you for this!

PR #3392, opened 2023-11-23
core[patch]: Add LLM/ChatModel callbacks to cached generation

### LINKED ISSUE #3391: Use callbacks when using LLM with cache
Integrations with observability platforms like LangSmith have been great for tracing calls to OpenAI. However, when a cache is specified, the callbacks (`handleLLMStart`, `handleLLMEnd`, and `handleLLMError`) [aren't called](https://github.com/langchain-ai/langchainjs/blob/main/langchain-core/src/language_models/chat_models.ts#L328-L368) unlike in their [uncached counterparts](https://github.com/langchain-ai/langchainjs/blob/main/langchain-core/src/language_models/chat_models.ts#L240-L279). 

I presume this is the case because the callbacks don't currently handle the case of cached responses, so there wouldn't be a way to tell if tokens were actually consumed or not. That being said, I think it would be pretty worthwhile to add this and even initially just specify that it's cached in the callback metadata. I would very much like to see the messages being generated on LangSmith and not just the calls to the tools.

## Uncached
image

## Cached
image

gramliu:
This creates a `_generateCached` helper function in the base and chat language models. This triggers the relevant callback events (`handleLLM/ChatModelStart`, `handleLLMEnd`, `handleLLMError`), even when pulling generations from the cache. To differentiate from regular LLM calls, this specifies an additional field, `cached: true`, in the `extraParams` passed to these callbacks.

Fixes #3391


langchain-core/src/language_models/chat_models.ts
```diff
@@ -293,6 +296,104 @@ export abstract class BaseChatModel<
...
+      { verbose: this.verbose }
+    );
+    const extra = {
+      options: parsedOptions,
+      invocation_params: this?.invocationParams(parsedOptions),
+      batch_size: 1,
+      cached: true,
+    };
+    const runManagers = await callbackManager_?.handleChatModelStart(
+      this.toJSON(),
```
jacoblee93, line 327: Sorry about this awful signature - prime candidate for a refactor 😬

langchain-core/src/language_models/chat_models.ts
```diff
@@ -293,6 +296,104 @@ export abstract class BaseChatModel<
     return output;
   }
 
+  /** @ignore */
```
jacoblee93, line 299: I don't think we need this?

langchain-core/src/language_models/chat_models.ts
```diff
@@ -293,6 +296,104 @@ export abstract class BaseChatModel<
...
+      undefined,
+      extra,
+      undefined,
+      undefined,
+      handledOptions.runName
+    );
+
+    // generate results
+    const llmStringKey =
+      this._getSerializedCacheKeyParametersForCall(parsedOptions);
```
jacoblee93, line 339: Let's pass this in as a parameter instead of recomputing

langchain-core/src/language_models/chat_models.ts
```diff
@@ -293,6 +296,104 @@ export abstract class BaseChatModel<
     return output;
   }
 
+  /** @ignore */
+  async _generateCached(
+    messages: BaseMessageLike[][],
+    cache: BaseCache<Generation[]>,
+    parsedOptions: this["ParsedCallOptions"],
+    handledOptions: RunnableConfig
```
jacoblee93, line 304: Wouldn't mind just making this an object parameter with instead of multiple args

langchain-core/src/language_models/llms.ts
```diff
@@ -282,6 +284,97 @@ export abstract class BaseLLM<
...
+        return result;
+      })
+    );
+
+    // Ignore results with a null value. These are just absent from the cache.
+    const cachedResults = results.filter(
+      (result) =>
+        (result.status === "fulfilled" && result.value != null) ||
+        result.status === "rejected"
+    ) as PromiseSettledResult<Generation[]>[];
```
jacoblee93, line 339: Do we need this cast?

Can you just do:

```
const cachedResults: PromiseSettledResult<Generation[]>[] = results.filter...
```

langchain-core/src/language_models/chat_models.ts
```diff
@@ -293,6 +296,104 @@ export abstract class BaseChatModel<
...
+        return result;
+      })
+    );
+
+    // Ignore results with a null value. These are just absent from the cache.
+    const cachedResults = results.filter(
+      (result) =>
+        (result.status === "fulfilled" && result.value != null) ||
+        result.status === "rejected"
+    ) as PromiseSettledResult<Generation[]>[];
```
jacoblee93, line 361: Do we need this? Can you just do:

```
const cachedResults: PromiseSettledResult<ChatGeneration[]>[] = results.filter(
```

jacoblee93: Actually looking at this more, if something gets filtered out, it would result in a call to the improper callback handler below? Since we're not filtering out `runManagers`

I think avoiding the filter and just returning `results` directly from this function after calling all the callback handlers is neater. Then we can handle missing prompt indices outside of this function.

jacoblee93 (changes requested):
Thanks for this, great catch that it's missing! See comments

langchain-core/src/language_models/chat_models.ts
```diff
@@ -293,6 +296,104 @@ export abstract class BaseChatModel<
...
+        }
+
+        return result;
+      })
+    );
+
+    // Ignore results with a null value. These are just absent from the cache.
+    const cachedResults = results.filter(
+      (result) =>
+        (result.status === "fulfilled" && result.value != null) ||
```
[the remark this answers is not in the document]
gramliu: `!= null` is intentional since this encompasses both `undefined` and `null` ([related SO post](https://stackoverflow.com/a/15992131)).

langchain-core/src/language_models/chat_models.ts
gramliu: Sorry for the super late response. Great catch! I hadn't noticed this bug. I'm still inclined to keeping the callback handler functionality within `_generateCached` since that's the bulk of what the function is really doing and abstracting away from the `generate` function that calls it. 

I did come up with a workaround though by mapping together the `runManagers` with the cached results before filtering. Then, rather than indexing into `runManagers` while calling the callback managers, we simply call the mapped `runManager`.

gramliu:
@jacoblee93 bumping this for your review when you have time!

jacoblee93:
Yes sorry! Working through the backlog post restructure. Will aim for tomorrow.

gramliu:
No worries. Just wanted to bump in case it got buried. Thanks!

jacoblee93:
I think good, small nit on naming!

jacoblee93:
Thank you!

jacoblee93:
Added a callback for `handleNewToken` as well.

PR #3394, opened 2023-11-24
integration[minor]: Llama Cpp streaming

nigel-daniels:
This PR adds streaming to the Llama CPP LLM and Chat modules.


langchain/src/chat_models/llama_cpp.ts
```diff
@@ -128,6 +133,43 @@ export class ChatLlamaCpp extends SimpleChatModel<LlamaCppCallOptions> {
...
   }
 
+  async *_streamResponseChunks(
+    input: BaseMessage[],
+    // @ts-expect-error - TS6133: 'options' is declared but its value is never read.
+    options: this["ParsedCallOptions"],
+    runManager?: CallbackManagerForLLMRun
+  ): AsyncGenerator<ChatGenerationChunk> {
+    if (input.length !== 1) {
+      throw new Error("Only one human message should be provided.");
```
jacoblee93, line 143: Is there a way we can fix this for multi-turn messages?

langchain/src/llms/llama_cpp.ts
```diff
@@ -87,4 +89,29 @@ export class LlamaCpp extends LLM<LlamaCppCallOptions> {
...
   }
+
+  async *_streamResponseChunks(
+    prompt: string,
+    // @ts-expect-error - TS6133: 'options' is declared but its value is never read.
+    options: this["ParsedCallOptions"],
+    runManager?: CallbackManagerForLLMRun
+  ): AsyncGenerator<GenerationChunk> {
+    const promptOptions = {
+      temperature: this?.temperature,
```
jacoblee93, line 100: maxTokens should be here too?

jacoblee93:
Some small questions but I think good to go

jacoblee93:
Thank you!

langchain/src/chat_models/llama_cpp.ts
nigel-daniels: I’ll explore this, ‘llama-node-cpp’ sets some of this up explicitly with the instantiation of a session. I should be able to compose a prompt annotated for ‘llama2’, although milage may vary if someone is using a none ‘llama2’ model.

langchain/src/llms/llama_cpp.ts
nigel-daniels: ‘maxTokens’ is set on the prompt method but to stream I’m using the evaluation method directly on the context. Unfortunately ‘maxTokens’ is not an option available on that method :(

PR #3395, opened 2023-11-24
integration[patch]: feat: implement max marginal relevance search for Weaviate vector store

the-powerpointer:
Following the examples of other vector stores.


jacoblee93:
Looks great! Will add some docs later.

jacoblee93:
Thank you!

PR #3399, opened 2023-11-24
langchain[minor]: feat(LLM Integration): WatsonX AI Integration

### LINKED ISSUE #2378: IBM WatsonX Integration
With IBM's recent release of [WatsonX.ai](https://www.ibm.com/products/watsonx-ai) you can now host LLMs through IBM Cloud and perform completion via their API.
I am working on an integration for LangChainJS, but have some questions with regards to best practices from the LangChain side.

As of this moment there is no JS package which wraps the endpoints, so the present integration would serve as said wrapper.

The existing integrations within LangChain all seem to use a package for managing endpoints, but for this use case would we be open to having the LangChain integration serve as said package?

chasemcdo:
Adds IBM WatsonX.ai Integration.
There does not presently exist a TypeScript library which wraps the interaction with WatsonX AI Production endpoints, so this PR creates a LangChain integration which does just that.

Therefore, this integration behaves a bit different than the majority of LangChain LLM integrations in that the `_call` function is a direct fetch to the WatsonX API endpoints.

Resolves #2378


langchain/src/llms/watsonx_ai.ts
```diff
@@ -0,0 +1,180 @@
...
+      throw new Error("Missing IBM Cloud API Key");
+    }
+
+    if (!this.projectId) {
+      throw new Error("Missing WatsonX AI Project ID");
+    }
+  }
+
+  _llmType() {
+    return "WatsonxAI";
```
jacoblee93, line 107: These are generally all lowercase

langchain/src/llms/watsonx_ai.ts
```diff
@@ -0,0 +1,180 @@
...
+        message: string;
+      }[];
+    }
+    const response = (await this.caller.call(async () =>
+      fetch(this.endpoint, {
+        method: "POST",
+        headers: {
+          "Content-Type": "application/json",
+          Accept: "application/json",
+          Authorization: `Bearer ${await this.generateToken()}`,
```
jacoblee93, line 137: Should we be caching this or is it expected to generate a new access token on each model call?

jacoblee93:
Looks good! See comments

langchain/src/llms/watsonx_ai.ts
chasemcdo: That is a great point. The token is reusable for up to an hour by default, so likely makes sense to cache.
I have updated `generateToken` to reuse the token if it is still valid.

jacoblee93:
Thank you!

faileon:
@chasemcdo 
I've started working on this today without realizing you are already on it.

I've implemented the streaming methods so far, which this PR lacks. Do you mind if I submit a PR when this gets merged or do you already have that in progress?

jacoblee93:
You can also base the PR off of this branch and it should work

jacoblee93:
Hey @chasemcdo no need to keep merging `main`!

PR #3409, opened 2023-11-25
langchain[patch]: Implements support for Personal Access Token Authentication in the ConfluenceLoader

### LINKED ISSUE #3408: Support personal access token (PAT) in ConfluenceLoader
See issue for langchain: [Support personal access token (PAT) in ConfluenceLoader #5191](https://github.com/langchain-ai/langchain/issues/5191)
Fixed by PR: [Implements support for Personal Access Token Authentication in the ConfluenceLoader #5385](https://github.com/langchain-ai/langchain/pull/5385)

>The [Atlassian API](https://atlassian-python-api.readthedocs.io/) (including Confluence) supports just passing a PAT (as token=) to authenticate as a user, unfortunately the LangChain abstraction doesn't.

>Suggestion:
>Add an optional "token" parameter to ConfluenceLoader and use it to authenticate within as an alternative to api_key/password/oauth.

This issue is the same for lanchainjs. 

When having a confluence server instance with version 7.19 > the only option is to use PAT. For legacy support I agree with the suggestion of adding an optional parameter but to instead name it `personalAccessToken` and to add logic to either use that or username & password.

yoHasse:
Fixes #3408

Implements a new optional parameter for the ConfluenceLoader: `personalAccessToken`.
This allows the use of personal access authentication when using the on-prem server version of Confluence.


examples/src/document_loaders/confluence.ts
```diff
@@ -2,19 +2,29 @@ import { ConfluencePagesLoader } from "langchain/document_loaders/web/confluence
...
+const personalAccessToken = process.env.CONFLUENCE_PAT;
 
 if (username && accessToken) {
   const loader = new ConfluencePagesLoader({
     baseUrl: "https://example.atlassian.net/wiki",
     spaceKey: "~EXAMPLE362906de5d343d49dcdbae5dEXAMPLE",
-    username,
-    accessToken,
+    username: username,
+    accessToken: accessToken,
```
bracesproul, line 12:
```suggestion
    username,
    accessToken,
```
nit

examples/src/document_loaders/confluence.ts
```diff
@@ -2,19 +2,29 @@ import { ConfluencePagesLoader } from "langchain/document_loaders/web/confluence
...
   });
 
   const documents = await loader.load();
   console.log(documents);
-} else {
+} else if (personalAccessToken) {
+  const loader = new ConfluencePagesLoader({
+    baseUrl: "https://example.atlassian.net/wiki",
+    spaceKey: "~EXAMPLE362906de5d343d49dcdbae5dEXAMPLE",
+    personalAccessToken: personalAccessToken,
```
bracesproul, line 21:
```suggestion
    personalAccessToken,
```
nit

langchain/src/document_loaders/web/confluence.ts
```diff
@@ -54,25 +55,40 @@ export class ConfluencePagesLoader extends BaseDocumentLoader {
...
 
   constructor({
     baseUrl,
     spaceKey,
-    username,
-    accessToken,
+    username = "",
+    accessToken = "",
     limit = 25,
+    personalAccessToken = "",
```
bracesproul, line 72: If we're going to make all optional we should throw if none are passed. Also, initializing with an empty string is not good practice. We should instead default undefined.

bracesproul (commented):
Thanks for adding this, left a couple comments

langchain/src/document_loaders/web/confluence.ts
yoHasse: > If we're going to make all optional we should throw if none are passed....

I would recommend allowing the option to not include any credentials, which would result in no Authorization header (after below change). This approach would allow fetching data from Confluence instances that have publicly accessible spaces and pages, a functionality that is available at least for Server/Data Center instances. 

I'd like to run this by you first to see if you're okay with this approach?

```Typescript
  private authorizationHeader(): string {
    if (this.personalAccessToken) {
      return `Bearer ${this.personalAccessToken}`;
    } else if (this.username && this.accessToken) {
      const authToken = Buffer.from(
        `${this.username}:${this.accessToken}`
      ).toString("base64");
      return `Basic ${authToken}`;
    }

    return undefined;
  }
  
...

  const initialHeaders: HeadersInit = {
    Accept: "application/json",
  };
  
  const authHeader = this.authorizationHeader();
  if (authHeader) {
    initialHeaders.Authorization = authHeader;
  }

  const response = await fetch(url, {
    headers: initialHeaders,
  });
  
  
```

langchain/src/document_loaders/web/confluence.ts
```diff
@@ -54,25 +55,40 @@ export class ConfluencePagesLoader extends BaseDocumentLoader {
...
     super();
     this.baseUrl = baseUrl;
     this.spaceKey = spaceKey;
     this.username = username;
     this.accessToken = accessToken;
     this.limit = limit;
+    this.personalAccessToken = personalAccessToken;
+  }
+
+  private authorizationHeader(): string {
```
jacoblee93, line 83: Could make this `private get authorizationHeader(): string` but not too opinionated

jacoblee93:
Thank you!

PR #3425, opened 2023-11-28
Fix null or undefined records causing error in Xata similarity search method

### LINKED ISSUE #3418: `similaritySearchVectorWithScore` method on XataVectorSearch gives `TypeError: records.map is not a function`
I am using `"langchain": "^0.0.150"`, and `"@xata.io/client": "^0.27.0"`. So when I import the xata client and create a vector store from it
```javascript
    // Fetch vector store
    this.vectorStore = new XataVectorSearch(new OpenAIEmbeddings(), {
      client: this.client(),
      table: 'articles',
    });
```
and I try to call `similaritySearchVectorWithScore` method like this ->
```javascript
const docs = await vectorStore.similaritySearchVectorWithScore(
          queryEmbeddings, // embeddings of the query
          10,
        );
```
I get this error ->
```
TypeError: records.map is not a function
    at XataVectorSearch.similaritySearchVectorWithScore (/home/codenameakshay/Developer/kalam-ai/node_modules/langchain/dist/vectorstores/xata.cjs:97:24)
    at /home/codenameakshay/Developer/kalam-ai/src/llm/agents/handleInternetSearch.ts:102:22
```
It says the error is in this line 97 in `xata.cjs`. I think the records are coming empty/null, and then the map function is not callable on that.
image

codenameakshay:
This pull request fixes an issue where null or undefined records were causing an error in the Xata similarity search method. The method now checks if the records exist before processing them, and returns an empty array if they are null or undefined. This prevents the error from occurring and ensures the method can handle all types of records properly.

Fixes #3418

### DIFF
--- a/langchain/src/vectorstores/xata.ts
+++ b/langchain/src/vectorstores/xata.ts
@@ -128,20 +128,22 @@ export class XataVectorSearch<
     );
 
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    return records?.map((record: any) => [
-      new Document({
-        pageContent: record.content,
-        metadata: Object.fromEntries(
-          Object.entries(record).filter(
-            ([key]) =>
-              key !== "content" &&
-              key !== "embedding" &&
-              key !== "xata" &&
-              key !== "id"
-          )
-        ),
-      }),
-      record.xata.score,
-    ]) ?? [];
+    return (
+      records?.map((record: any) => [
+        new Document({
+          pageContent: record.content,
+          metadata: Object.fromEntries(
+            Object.entries(record).filter(
+              ([key]) =>
+                key !== "content" &&
+                key !== "embedding" &&
+                key !== "xata" &&
+                key !== "id"
+            )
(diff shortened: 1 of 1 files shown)


langchain/src/vectorstores/xata.ts
```diff
@@ -128,7 +128,7 @@ export class XataVectorSearch<
     );
 
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    return records.map((record: any) => [
+    return records?.map((record) => [
```
bracesproul, line 131: You can delete this eslint rule if the `any` is removed.

bracesproul (commented):
looks like build is failing due to bad types, if you could fix and re-request a review that would be great!

codenameakshay:
Have fixed the code, `any` got removed accidentally. Please check now.

tsg:
Many thanks for the fix, I’ll test it quickly today from our side (Xata), but generally looks good 👍.

Issue #2756, opened 2023-09-29
Add Qdrant custom payload on documents to query them by filter


uptoo-owner:
Hi !

The Qdrant integration works well with the content, vector, and payload.
However, it would be great to be able to add our own payload in JSON. This would allow querying the database on more criteria, like geolocation for instance.

It would enable the creation of one's own DynamicStructuredTool and therefore search within the same Qdrant base as the vectors.

An approach like this, for instance:

```/**
     * Method to add vectors to the Qdrant database. Each vector is associated
     * with a document, which is stored as the payload for a point in the
     * database.
     * @param vectors Array of vectors to be added to the Qdrant database.
     * @param documents Array of `Document` instances associated with the vectors.
     * @returns Promise that resolves when the vectors have been added to the database.
     */
    async addVectors(vectors, documents) {
        if (vectors.length === 0) {
            return;
        }
        await this.ensureCollection();
        const points = vectors.map((embedding, idx) => ({
            id: uuid(),
            vector: embedding,
            payload: {
                // add custom payload here <========== HERE
                // ...documents[idx].payload
                content: documents[idx].pageContent,
                metadata: documents[idx].metadata,
            },
        }));
        await this.client.upsert(this.collectionName, {
            wait: true,
            points,
        });
    }```

Thanks !

youngjaeheo2002:
Hello, is anyone working on this ticket at the moment? My group and I would like to contribute to this repository for a group project.

youngjaeheo2002:
Hi I'm going to assume that no one else is working on this and my team will get started working on it.

JustDoItSascha:
I would like to reopen this ticket, because from my point of view it's not correct. The customPayload can be passed as general dbConfig, but will not have any effect when using `fromExistingCollection()`. But I think it's often the use case that I want to add a customPayload to ALL documents at once, like a tenantId. So I created another MR for this.

jacoblee93:
Thank you! Will look as soon as possible

JustDoItSascha:
still relevant

dosubot[bot] ***@***.***> schrieb am Fr., 24. Jan. 2025,
17:00:

christian-bromann:
Let's continue this convo in #7069 .. thanks @JustDoItSascha for taking a stab at a fix for this 🙏

PR #3431, opened 2023-11-28
langchain[patch]: Issue #2756 Add Qdrant custom payload on documents to query them by filter

youngjaeheo2002:
Description:
- Added custom JSON payload integration to Qdrant integration 
- This allows querying database with more criteria such as geological distance
- Implemented additional functionality to addDocuments and addVectors functions in langchain/src/vectorstores/qdrant.ts to allow storing vectors with documents that have custom payloads
- Testing:
    - Added unit tests for Qdrant Vector Store adding custom payloads of size 1, size greater than 1, and no custom payload 

- Issue: #2756


langchain/src/vectorstores/qdrant.ts
```diff
@@ -115,6 +125,8 @@ export class QdrantVectorStore extends VectorStore {
       payload: {
         content: documents[idx].pageContent,
         metadata: documents[idx].metadata,
+        ...(documents[idx].customPayload ? documents[idx].customPayload != customPayload ? 
+          documents[idx].customPayload : customPayload : customPayload),
```
bracesproul, line 129: Could you just do something like this?
```suggestion
          documents[idx].customPayload ?? customPayload),
```
Not sure if this is valid though

langchain/src/vectorstores/qdrant.ts
```diff
@@ -204,20 +216,23 @@ export class QdrantVectorStore extends VectorStore {
...
+   * @param customPayload Optional 'object' in JSON format used to query database on more criteria
    * @returns Promise that resolves with a new `QdrantVectorStore` instance.
    */
   static async fromTexts(
     texts: string[],
     metadatas: object[] | object,
     embeddings: Embeddings,
-    dbConfig: QdrantLibArgs
+    dbConfig: QdrantLibArgs,
+    customPayload? : object
```
bracesproul, line 227:
```suggestion
    customPayload?: object
```
nit

langchain/src/vectorstores/qdrant.ts
```diff
@@ -204,20 +216,23 @@ export class QdrantVectorStore extends VectorStore {
...
   ): Promise<QdrantVectorStore> {
     const docs = [];
     for (let i = 0; i < texts.length; i += 1) {
       const metadata = Array.isArray(metadatas) ? metadatas[i] : metadatas;
       const newDoc = new Document({
         pageContent: texts[i],
         metadata,
+        customPayload: customPayload,
```
bracesproul, line 235:
```suggestion
        customPayload,
```
nit

bracesproul (commented):
overall looks good to me once my comments have been resolved
cc @jacoblee93 for a second look

bracesproul (commented):
Also one merge conflict, you'll have to push up the update to the `document` file inside `langchain-core` since the contents of that file have been moved.

Dominik-luszcz:
Hi, I added the optional parameter customPayload to the Documents file in langchain-core, but the only test cases that fail are the:
```
expect(docs[0]).toMatchInlineSnapshot(`
    Document {
      "metadata": {
        "blobType": "application/jsonl+json",
        "line": 1,
        "source": "blob",
      },
      "pageContent": "This is a sentence.",
    }
  `);
```
cases in the blob.test.ts files. To pass these test cases, all we need to do is add `"customPayload": undefined` before the "metadata" in the snapshot above. Am I allowed to add this to was the test cases? Or should I try to prevent the undefined values in the loader.load() function?

jacoblee93:
We can't change the core document schema - it seems like it's some custom querying logic? What's the use-case for persisting it like this vs. some metadata field or just passing it in on ingestion?

Dominik-luszcz:
We saw that in the request, it was suggested to have the custom query parameters within the document indicating that it was needed. Under further consideration it seems like this was unnecessary. We can change it so that it does not have it inside the document.

Dominik-luszcz:
We have updated the code. Please give us any feedback when you have the chance.

langchain/src/vectorstores/qdrant.ts
```diff
@@ -84,13 +84,18 @@ export class QdrantVectorStore extends VectorStore {
    * from the documents using the `Embeddings` instance and then adds the
    * vectors to the database.
    * @param documents Array of `Document` instances to be added to the Qdrant database.
+   * @param customPayload Optional 'object' in JSON format used to query database on more criteria
    * @returns Promise that resolves when the documents have been added to the database.
    */
-  async addDocuments(documents: Document[]): Promise<void> {
+  async addDocuments(
+    documents: Document[],
+    customPayload?: object[]
```
jacoblee93, line 92: Can you change this to an object with a field called `customPayload`? That better matches the base interface:

https://github.com/langchain-ai/langchainjs/blob/main/langchain-core/src/vectorstores.ts#L126

langchain/src/vectorstores/qdrant.ts
```diff
@@ -204,13 +215,15 @@ export class QdrantVectorStore extends VectorStore {
...
+   * @param customPayload Optional 'object' in JSON format used to query database on more criteria
    * @returns Promise that resolves with a new `QdrantVectorStore` instance.
    */
   static async fromTexts(
     texts: string[],
     metadatas: object[] | object,
     embeddings: Embeddings,
-    dbConfig: QdrantLibArgs
+    dbConfig: QdrantLibArgs,
+    customPayload?: object[]
```
jacoblee93, line 226: Would prefer this field to be within `dbConfig`

langchain/src/vectorstores/qdrant.ts
```diff
@@ -230,15 +248,17 @@ export class QdrantVectorStore extends VectorStore {
...
    * @param dbConfig `QdrantLibArgs` instance specifying the configuration for the Qdrant database.
+   * @param customPayload Optional 'object' in JSON format used to query database on more criteria
    * @returns Promise that resolves with a new `QdrantVectorStore` instance.
    */
   static async fromDocuments(
     docs: Document[],
     embeddings: Embeddings,
-    dbConfig: QdrantLibArgs
+    dbConfig: QdrantLibArgs,
+    customPayload?: object[]
```
jacoblee93, line 258: See above

jacoblee93 (commented):
Thank you for the changes! See comments

Dominik-luszcz:
We have made the changes, please let me know if there is anything wrong with them.

langchain/src/vectorstores/qdrant.ts
```diff
@@ -2,6 +2,7 @@ import { QdrantClient } from "@qdrant/js-client-rest";
 import type { Schemas as QdrantSchemas } from "@qdrant/js-client-rest";
 import { v4 as uuid } from "uuid";
 
+import { AddDocumentOptions } from "closevector-common";
```
jacoblee93, line 5: This looks like a mistake?

langchain/src/vectorstores/tests/qdrant.test.ts
```diff
@@ -1,5 +1,6 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import { jest, test, expect } from "@jest/globals";
+import { AddDocumentOptions } from "closevector-common";
```
jacoblee93, line 3: This looks like a mistake?

Dominik-luszcz:
> Just specify your own add document options type

Something like this inside qdrant.ts or would something else be better?

`export type qdrantDocumentOptions = {
  customPayload: object[];
};`

jacoblee93:
@Dominik-luszcz and @youngjaeheo2002 LMK if this looks good!

Dominik-luszcz:
Yeah it looks good to us! Thanks @jacoblee93!

jacoblee93:
Thank you!

PR #3437, opened 2023-11-29
core[minor]: Runnable with message history

bracesproul:
Also added a fake chat message class

### DIFF
--- a/langchain-core/src/runnables/base.ts
+++ b/langchain-core/src/runnables/base.ts
@@ -655,7 +655,7 @@ export class RunnableBinding<
       ...(this.configFactories
         ? await Promise.all(
             this.configFactories.map(
-              async (factoryMethod) => await factoryMethod(config)
+              async (configFactory) => await configFactory(config)
             )
           )
         : [])
@@ -798,7 +798,6 @@ export class RunnableBinding<
     onEnd?: (run: Run, config?: RunnableConfig) => void | Promise<void>;
     onError?: (run: Run, config?: RunnableConfig) => void | Promise<void>;
   }): Runnable<RunInput, RunOutput, CallOptions> {
-    //
     return new RunnableBinding<RunInput, RunOutput, CallOptions>({
       bound: this.bound,
       kwargs: this.kwargs,
--- a/langchain-core/src/chat_history.ts
+++ b/langchain-core/src/chat_history.ts
@@ -42,27 +42,23 @@ export class FakeChatMessageHistory extends BaseChatMessageHistory {
     super();
   }
 
-  public getMessages(): Promise<BaseMessage[]> {
-    return Promise.resolve(this.messages);
+  async getMessages(): Promise<BaseMessage[]> {
+    return this.messages;
   }
 
-  public addMessage(message: BaseMessage): Promise<void> {
+  async addMessage(message: BaseMessage): Promise<void> {
     this.messages.push(message);
-    return Promise.resolve();
   }
 
-  public addUserMessage(message: string): Promise<void> {
+  async addUserMessage(message: string): Promise<void> {
     this.messages.push(new HumanMessage(message));
-    return Promise.resolve();
   }
 
-  public addAIChatMessage(message: string): Promise<void> {
+  async addAIChatMessage(message: string): Promise<void> {
     this.messages.push(new AIMessage(message));
-    return Promise.resolve();
   }
 
-  public clear(): Promise<void> {
+  async clear(): Promise<void> {
     this.messages = [];
-    return Promise.resolve();
   }
 }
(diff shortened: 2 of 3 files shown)


langchain-core/src/runnables/config.ts
```diff
@@ -14,3 +14,76 @@ export async function getCallbackMangerForConfig(config?: RunnableConfig) {
     config?.metadata
   );
 }
+
+export function mergeConfigs<CallOptions extends RunnableConfig>(
```
jacoblee93, line 18: We aren't using this type param anywhere are we? Should we be?

langchain-core/src/runnables/base.ts
```diff
@@ -12,11 +12,19 @@ import {
...
+} from "./config.js";
 import { AsyncCaller } from "../utils/async_caller.js";
+import { Run } from "../tracers/base.js";
+import { RootListenersTracer } from "../tracers/root_listener.js";
 
 export type RunnableFunc<RunInput, RunOutput> = (
-  input: RunInput
+  input: RunInput,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  options?: any
```
jacoblee93, line 27: We talked about making this

```
Record<string, any> & {
  config: RunnableConfig,
}
```

didn't we?

langchain-core/src/runnables/history.ts
```diff
@@ -0,0 +1,189 @@
...
+        `Output values from 'Run' undefined. Run: ${JSON.stringify(
+          run,
+          null,
+          2
+        )}`
+      );
+    }
+    const outputMessages = this._getOutputMessages(outputValue);
+
+    for await (const message of [...inputMessages, ...outputMessages]) {
```
jacoblee93, line 162: This won't double add messages right?

langchain-core/src/runnables/base.ts
bracesproul: Doing that caused type errors in other places where a function was being passed in and it had more than one arg. 
Those args were nullable so it didn't matter then, but now that we allow for a second arg if function's being passed don't match this type it errors.
Example: calling `runnable.pipe(formatDocumentsAsString)`
`formatDocumentsAsString` has two args: One of `Document` and a second optional arg of `string`.
If we type the second arg as something other than `any` it'll cause that to error.
Granted we could update those examples to this: `runnable.pipe((docs) => formatDocumentsAsString(docs))`
But anyone using this code would start to get type errors.

langchain-core/src/runnables/config.ts
bracesproul: Whoops, looks like a refactor I made accidentally removed it. Re-added.

langchain-core/src/runnables/config.ts
```diff
@@ -14,3 +14,76 @@ export async function getCallbackMangerForConfig(config?: RunnableConfig) {
...
+  const copy: Record<string, any> = { ...config };
+  if (options) {
+    for (const key of Object.keys(options)) {
+      if (key === "metadata") {
+        copy[key] = { ...copy[key], ...options[key] };
+      } else if (key === "tags") {
+        copy[key] = (copy[key] ?? []).concat(options[key] ?? []);
+      } else if (key === "callbacks") {
+        const baseCallbacks = copy.callbacks;
+        const theseCallbacks = options.callbacks ?? config.callbacks;
```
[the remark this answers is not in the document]
bracesproul: Was copying PY but I agree your suggestion is cleaner.

langchain-core/src/runnables/history.ts
```diff
@@ -0,0 +1,189 @@
...
+      historyChain = RunnablePassthrough.assign({
+        [messagesKey]: historyChain,
+      }).withConfig({ runName: "insertHistory" });
+    }
+
+    const bound = historyChain
+      .pipe(
+        fields.runnable.withListeners({
+          onEnd: (run: Run, config: RunnableConfig = {}) =>
+            this._exitHistory(run, config),
```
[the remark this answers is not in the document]
bracesproul: yes since it's async

langchain-core/src/runnables/base.ts
```diff
@@ -673,13 +723,16 @@ export class RunnableBinding<
...
     const mergedOptions = Array.isArray(options)
-      ? options.map((individualOption) =>
-          this._mergeConfig({
-            ...individualOption,
-            ...this.kwargs,
-          })
+      ? await Promise.all(
+          options.map(
+            async (individualOption) =>
+              await this._mergeConfig({
```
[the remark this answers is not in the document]
bracesproul: Don't believe so, no.

jacoblee93:
Thanks!

PR #3443, opened 2023-11-29
langchain[patch]: onToken event added in ChatLlamaCpp call function

shareefmorayur:
Fixes # (issue)


jacoblee93:
Should we also add to the LLM class?

shareefmorayur:
> Should we also add to the LLM class?

@jacoblee93 it is added LLM class also. Could please verify and merge it?

jacoblee93:
Thank you!

Issue #2951, opened 2023-10-18
Add Jira Toolkit


Jason8908:
## Feature Request
We would like to implement a toolkit to allow agents to interact with the Jira interface.

## Motivation
Currently, the Python version of langchain has this feature implemented, but we cannot find an equivalent in the JavaScript version. Adding this feature would help to reach parity with the Python version in terms of features.

## Your contribution
If all goes well, our team will be looking to submit a PR around the end of November and we would love any suggestions on how to start looking at this issue!

PR #3446, opened 2023-11-29
feat: Added Jira Toolkit

TheZhangedMan:
## Description
Added an agent toolkit for Jira, along with an example file for running tests. We have added the to give the Jira Toolkit a prompt, and having the LLM figure out what actions to run on a Jira project, and return the appropriate information.

This toolkit has access to three main operations (create issue, get projects and run JQL command), which allow it to perform a lot of basic common functionality. However, there is also a catch all prompt, which scrapes the Jira API to figure out what actions could be called. 

For Example:

* The Jira API wrapper is defined with the Jira API key to allow langchain access to the Jira project. The `JiraAPIWrapper` is then passed into the Jira Toolkit. The toolkit is passed into an agent executor using a `zero-shot-react-description` agent. The executor is given the prompt: “Make a new issue in project PW to remind me to make more fried rice.”
* Upon calling the executor, the LLM figures it needs to use the create issue tool, passes in the appropriate parameters according to the prompt, and creates the issue in the specified Jira project.

## Issue
* [Add Jira Toolkit #2951](https://github.com/langchain-ai/langchainjs/issues/2951)


langchain/src/agents/toolkits/index.ts
```diff
@@ -18,3 +18,4 @@ export {
   type ConversationalRetrievalAgentOptions,
 } from "./conversational_retrieval/openai_functions.js";
 export { OpenAIAgentTokenBufferMemory } from "./conversational_retrieval/token_buffer_memory.js";
+export { JiraToolkit } from "./jira/jira.js";
```
jacoblee93, line 21: Let's give this its own entrypoint under `create-entrypoints.js`.

Then you'll need to run `yarn build` to update artifacts.

langchain/src/tools/index.ts
```diff
@@ -47,3 +47,4 @@ export {
   formatToOpenAIFunction,
   formatToOpenAITool,
 } from "./convert_to_openai.js";
+export { JiraAPIWrapper, JiraAction } from "./jira.js";
```
jacoblee93, line 50: See the above comment on toolkits

You'll also need to prefix types with `type` when exporting to avoid build issues

examples/src/agents/jira_toolkit.ts
```diff
@@ -0,0 +1,36 @@
...
+    }
+  );
+
+  console.log("Loaded agent.");
+
+  const input = `Make a new issue in project PW to remind me to make more fried rice.`;
+
+  console.log(`Executing with input "${input}"...`);
+
+  const result = await executor.call({ input });
```
jacoblee93, line 33: Prefer `.invoke`

jacoblee93:
Thanks! See comments, don't forget to run `yarn format` and `yarn lint`

langchain/src/agents/toolkits/index.ts
TheZhangedMan: I have done the following:
* Removed `export { JiraToolkit } from "./jira/jira.js";` (Line 21) from `langchain/src/agents/toolkits/index.ts` and added the following line to `create-entrypoints.js` under the `agents` key: `"agents/toolkits/jira": "agents/toolkits/jira/index,"`.
  * Consequently, the additional file `index.ts`, which exports `JiraToolkit`, was created under `langchain/src/agents/toolkits/jira/`.
  * Users now need to include the following to use `JiraToolkit`: `import { JiraToolkit } from "langchain/agents/toolkits/jira";`, which is now reflected in the example file.
  
Let me know if I need to make additional changes!

langchain/src/tools/index.ts
TheZhangedMan: I have done the following:
* Added the following line to `create-entrypoints.js` under the `tools` key: `"tools/jira": "tools/jira,"`.
  * No additional prefixes for types needed to be added to the line `export { JiraAPIWrapper, JiraAction } from "./jira.js";` since only classes are exported.
  * Just as in the comment above, imports need to be adjusted accordingly.
  
Let me know if I need to make additional changes!

examples/src/agents/jira_toolkit.ts
TheZhangedMan: I have changed executor.call() to executor.invoke() in the example file and then made sure everything was behaving normally after my changes.

jacoblee93:
Thanks! Apologies for the delay - we've been going through some refactoring.

I've moved it into the `@langchain/community` package - would you be able to add a test case or two?

christian-bromann:
@TheZhangedMan thanks for the contribution 🙏 sorry for not getting back to you on this earlier. I would advise maybe publish this as your own package on NPM? I think this is still super valuable but the code doesn't necessarily need to live within the langchain repository. I will go ahead and close this but please let me know if you end up publishing this so I can highlight this within our community.

PR #3450, opened 2023-11-30
core[patch]: Move tests into core

bracesproul:


langchain-core/src/runnables/tests/lib.ts
bracesproul, line 1: Not sure if we want to expose these classes inside langchain core so I've just duplicated the file, can change

jacoblee93: I think exporting is better 👍

jacoblee93:
Thanks! Yeah would say exporting + removing is better - I can land

PR #3451, opened 2023-11-30
fix(npm script): lint fix doesn't fix fixable errors

chasemcdo:
The current setup for `yarn lint:fix` means it does the same thing as `yarn lint` this is because the `--fix` flag is being set for `dpdm` and not for `eslint`.

This PR seeks to remedy that by creating two separate scripts `lint:eslint` and `lint:dpdm` and properly setting the `--fix` flag when `lint:fix` is called.

This also has the side effect of allowing contributors to independently check `eslint` and `dpdm` without impacting the current functionality.

Fixes: #3452


bracesproul (approved):
Thanks for this! I also pushed up a couple commits updating `langchain-core` and `libs/langchain-`(`openai`, `anthropic`) scripts.

cc @jacoblee93 for a second look

PR #3462, opened 2023-11-30
Fix for Prisma vectorstore build query IN filter

### LINKED ISSUE #3460: Prisma vectorestore incorrectly builds sql query with in operator
The `Prisma` vectorestore does not create a correct query when providing a filter with the `in` operator.

Line https://github.com/langchain-ai/langchainjs/blob/993ce7a822bd11e125419bff84a9951bd58e66e5/langchain/src/vectorstores/prisma.ts#L439C58-L439C75         
```js
return this.Prisma.sql`${colRaw} ${opRaw} (${value.join(",")})`;
```
results in `"Column" IN ("value1,value2")` whereas this should be `"Column" IN ("value1","value2")`.

This can be fixed by using 
```js
return this.Prisma.sql`${colRaw} ${opRaw} (${this.Prisma.join(value)})`;
```

Langchain version 0.0.198.

I'm currently trying to create a PR for this.

Knordy:
Fixing an issue, where the `IN` query filter created an incorrect value syntax by joining the values as a single value `("value1,value2")`. By using the Prisma join, the syntax is corrected again `("value1","value2")`.

Fixes #3460

### DIFF
--- a/langchain/src/vectorstores/prisma.ts
+++ b/langchain/src/vectorstores/prisma.ts
@@ -436,7 +436,7 @@ export class PrismaVectorStore<
                   )}`
                 );
               }
-              return this.Prisma.sql`${colRaw} ${opRaw} (${value.join(",")})`;
+              return this.Prisma.sql`${colRaw} ${opRaw} (${this.Prisma.join(value)})`;
             }
             case OpMap.isNull:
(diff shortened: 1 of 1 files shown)


jacoblee93:
Great catch, thanks!

PR #3463, opened 2023-11-30
langchain[minor]: Added multi-message streaming to llama_cpp

nigel-daniels:
The chat module for `llama_cpp` now handles multiple messages or a single string, see update documents.


jacoblee93:
Awesome, thank you so much!

Will probably want to have some kind of logic for different models in the future but I think this is great for now

nigel-daniels:
I agree ‘node-llama-cpp’ does have a range of chat wrappers but as the streaming is based on a raw call I’ll need to implement something different.

nigel-daniels:
FYI just submitted PR #3505 to support the appropriate input formats for multiple models supported by `node-llama-cpp`.

PR #3468, opened 2023-11-30
examples[ci]: Build examples dir in CI

bracesproul:
updated prisma dep to be pinned, was causing issues when it had the carrot(^)


bracesproul:
Stale, closing for the foreseeable future

PR #3475, opened 2023-12-01
langchain[minor]: feat(embedding integration): Gradient AI

### LINKED ISSUE #3007: Adding gradient.ai LLM and Embeddings
### Feature request

I propose to add the gradient-ts/js client of gradient.ai as a `LLM` and `Embedding`, similar to the Python implementation in https://python.langchain.com/docs/integrations/providers/gradient .

### Motivation

Gradient.ai offers fine-tuning and inference of models without paying for the server costs. 

I fine-tuned a model with a specific RAG template based on llama2 on gradient.ai, now i want to use it from Langchain, also in ts/js.

chasemcdo:
Adds support for Gradient AI's embeddings.

Resolves #3007


docs/core_docs/docs/integrations/text_embedding/gradient_ai.mdx
```diff
@@ -0,0 +1,33 @@
...
+You'll need to install the official Gradient Node SDK as a peer dependency:
+
+```bash npm2yarn
+npm i @gradientai/nodejs-sdk
+```
+
+You will need to set the following environment variables for using the Gradient AI API.
+
+1. `GRADIENT_ACCESS_TOKEN`
+2. `GRADIENT_WORKSPACE_ID`
```
bracesproul, line 16: nit (replace backtick with ``` ` ```)
```
backtick backtick backtick shell
export GRADIENT_ACCESS_TOKEN=<YOUR_ACCESS_TOKEN>
export GRADIENT_WORKSPACE_ID=<YOUR_WORKSPACE_ID>
backtick backtick backtick
```

bracesproul (approved):
lgtm, cc @jacoblee93 for a second look

jacoblee93:
Thank you!

langchain/src/embeddings/gradient_ai.ts
```diff
@@ -0,0 +1,117 @@
...
+    const data = await this.embedDocuments([text]);
+    return data[0];
+  }
+
+  /**
+   * Method to set the model to use for generating embeddings.
+   * @sets the class' `model` value to that of the retrieved Embeddings Model.
+   */
+  async setModel() {
+    if (this.model) return;
```
jacoblee93, line 107: Why is this necessary vs the user setting it?

chasemcdo: To comment on the edit (misinterpreted as setting of model not slug):
There is presently only one model available with the TypeScript definition for that defining it as:
```
export declare const GenerateEmbeddingSlugEnum: {
    readonly BgeLarge: "bge-large";
};
```
So we'll get a type issue unless we ignore it or also restrict our slug input to `bge-large`.

To me it didn't much sense to add an extra input that had only one correct value if it would need to be changed to match new typecasting in a future update anyways.

So saw the path as being:
- no model input until more models are added with a more flexible type definition
- update to new input definition when flexibility is improved with `bge-large` being the default for backwards compatibility

I can update to have the input if we want to, but it won't actually be changeable.

chasemcdo:
@jacoblee93 I can also whip up a quick contribution doc for the [text embeddings](https://github.com/langchain-ai/langchainjs/blob/07e6c4840829137a1b23160fc16771fa75c1e9f5/.github/contributing/INTEGRATIONS.md?plain=1#L155), since we are presently missing a specific one for text embeddings

langchain/src/embeddings/gradient_ai.ts
jacoblee93: Ok, no worries then! We do use this `StringWithAutocomplete` class in a few places but fine for now.

jacoblee93:
> @jacoblee93 I can also whip up a quick contribution doc for the [text embeddings](https://github.com/langchain-ai/langchainjs/blob/07e6c4840829137a1b23160fc16771fa75c1e9f5/.github/contributing/INTEGRATIONS.md?plain=1#L155), since we are presently missing a specific one for text embeddings.
> 
> Something like [this](https://github.com/langchain-ai/langchainjs/pull/3493)

Saw you opened a PR, thank you!

PR #3502, opened 2023-12-02
fix: query parameters are not passed correctly to WolframAlpha API

### LINKED ISSUE #3501: Queries to WolframAlpha are not passed correctly
When an agent uses WelframAlpha tool the query is not sent correctly, because the query is not encoded properly.

Test below shows that the plus sign well be ignored once the request URL is parsed.

Screenshot 2023-12-02 at 21 40 40

amellouki:
Fixes #3501


jacoblee93:
Can fix and merge tomorrow, thank you!

PR #3505, opened 2023-12-03
Llama stream multiple models supported

nigel-daniels:
This adds the `streamingModel` option to the streaming methods properties, this can be set to: `llama2`, `chatML`, `falcon`, or `general`. This then turns messages passed to the streaming model into the appropriate format for that model. If no option is specified it defaults to be `llama2`.

### DIFF
--- a/langchain/src/chat_models/llama_cpp.ts
+++ b/langchain/src/chat_models/llama_cpp.ts
@@ -16,6 +16,7 @@ import {
   BaseMessage,
   ChatGenerationChunk,
   AIMessageChunk,
+  ChatMessage,
 } from "../schema/index.js";
 
 /**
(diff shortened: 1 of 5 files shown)


langchain/src/chat_models/llama_cpp.ts
```diff
@@ -29,8 +30,14 @@ export interface LlamaCppInputs
 export interface LlamaCppCallOptions extends BaseLanguageModelCallOptions {
   /** The maximum number of tokens the response should contain. */
   maxTokens?: number;
-  /** A function called when matching the provided token array */
-  onToken?: (tokens: number[]) => void;
```
bracesproul, line 33: This is not going to be backwards compatible. Instead we should mark as deprecated

langchain/src/chat_models/llama_cpp.ts
```diff
@@ -29,8 +30,14 @@ export interface LlamaCppInputs
...
+  /** This lets the streaming method know which model we are using,
+   *  valid options are:
+   *  'llama2'	- A Llama2 model, this is the default.
+   *  'chatML'  - A ChatML model
+   *  'falcon'	- A Falcon model
+   *  'genral'  - Any other model, uses "### Human\n", "### Assistant\n" format
+   */
+  streamingModel?: string;
```
bracesproul, line 40: Use a union type or a check in the constructor that verifies the input is one of these?

langchain/src/chat_models/tests/chatllama_cpp.int.test.ts
```diff
@@ -85,14 +85,57 @@ test.skip("Test chain with memory", async () => {
...
+  const chunks = [];
+  for await (const chunk of stream) {
+    chunks.push(chunk.content);
+    console.log(chunk.content);
+  }
+
+  expect(chunks.length).toBeGreaterThan(1);
+});
+
+test.skip("test multi-mesage streaming call", async () => {
```
bracesproul, line 101: fyi if you were unaware, we don't run `int` tests in CI so no need to add `skip`

bracesproul (changes requested):
Overall looks good! I have a few comments, please re-request my review once resolved!

langchain/src/chat_models/llama_cpp.ts
nigel-daniels: The `onToken` was included in error and was not actually applied as the eventual implementation used a raw call. I can mark it as deprecated and not used if preferred.

nigel-daniels:
Resubmitted based off main under #3588.

PR #3517, opened 2023-12-04
feat: add inference for RunnableMap RunOutput type

dilling:
This PR adds a new RunnableMapLike type that helps with type inference when using RunnableMaps. Currently, RunnableMaps only output `Record<string, any>`. This allows the output type to be narrowed, which helps with type safety when composing multiple Runnables together.

This works when building RunnableMaps using the `from` methods or constructor
image

And also when composing RunnableMaps in within a RunnableSequence
image


langchain-core/src/runnables/tests/runnable.test.ts
```diff
@@ -81,15 +81,15 @@ test("Create a runnable sequence with a runnable map", async () => {
...
-  const inputs = {
+  const inputs = RunnableMap.from({
     question: (input: string) => input,
     documents: RunnableSequence.from([
       new FakeRetriever(),
       (docs: Document[]) => JSON.stringify(docs),
     ]),
     extraField: new FakeLLM({}),
-  };
-  const runnable = new RunnableMap({ steps: inputs })
```
dilling, line 92: The one unintended consequence of this change is that the compiler has more trouble inferring the RunInput type when the steps are passed in as a variable, rather than passed in directly, which makes the above change necessary. I tried to fix this with no success.

langchain-core/src/runnables/base.ts
```diff
@@ -1368,11 +1372,11 @@ export class RunnableSequence<
...
  */
-export class RunnableMap<RunInput> extends Runnable<
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export class RunnableMap<
   RunInput,
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  Record<string, any>
-> {
+  RunOutput extends Record<string, any> = Record<string, any>
+> extends Runnable<RunInput, RunOutput> {
```
bracesproul, line 1379: Need to move this eslint disable comment down two lines

langchain-core/src/runnables/base.ts
```diff
@@ -1663,9 +1672,9 @@ export function _coerceToRunnable<RunInput, RunOutput>(
   } else if (!Array.isArray(coerceable) && typeof coerceable === "object") {
     const runnables: Record<string, Runnable<RunInput>> = {};
     for (const [key, value] of Object.entries(coerceable)) {
-      runnables[key] = _coerceToRunnable(value);
+      runnables[key] = _coerceToRunnable(value as RunnableLike);
     }
-    return new RunnableMap<RunInput>({
+    return new RunnableMap({
```
bracesproul, line 1677: Why did you drop the generic passing here?

dilling: So when passing as RunnableMap<RunInput, RunOutput>, the type of Run output does not line up until the line below
```
    }) as unknown as Runnable<RunInput, Exclude<RunOutput, Error>>;
```

Passing _just_ the RunInput defaults the RunOutput to Record<string, any>. So you end up having to pass both values (which is not possible to the reason above) or neither

dilling:
@bracesproul thanks so much for taking a look! Just addressed the comments, let me know if there is any additional followup

jacoblee93:
Apologies for the delayed review! 

Yes I think this is great, thank you. Will do a bit more testing and hopefully cut a new release tonight.

jacoblee93:
Hey @dilling I think this is slightly better since it falls back to `any` if it has trouble inferring and doesn't break the existing tests - any issues you see?

dilling:
> Hey @dilling I think this is slightly better since it falls back to `any` if it has trouble inferring and doesn't break the existing tests - any issues you see?

Nope that looks good to me, thanks for adding tests 🙏

PR #3524, opened 2023-12-04
Fix records?.map is not a function bug in XataVectorStores

codenameakshay:
This pull request fixes a bug in the XataVectorStores module where the records?.map function was throwing an error. The bug was caused by incorrect usage of the records variable. This PR updates the code to destructure the records variable correctly before calling the map function. This ensures that the map function is only called when records is not null or undefined. This fix resolves the bug and improves the stability of the XataVectorStores module.

This was previously attempted to be fixed by the PR #3425

This ultimately fixes issue #3418

### DIFF
--- a/langchain/src/vectorstores/xata.ts
+++ b/langchain/src/vectorstores/xata.ts
@@ -118,7 +118,7 @@ export class XataVectorSearch<
     k: number,
     filter?: XataFilter | undefined
   ): Promise<[Document, number][]> {
-    const records = await this.client.db[this.table].vectorSearch(
+    const { records } = await this.client.db[this.table].vectorSearch(
       "embedding",
       query,
       {


jacoblee93:
This looks like it breaks the method signature?
Screenshot 2023-12-04 at 11 55 48 AM

jacoblee93:
CC @tsg

codenameakshay:
I have made this change, as requested by Phillip from Xata team. This is happening as in v `0.27.0` they have changed how `vectorSearch` method returns records.

image

eemmiillyy:
In version `0.27.0` we made a [breaking change](https://github.com/xataio/client-ts/commit/d282d18f025094e0729ade6009b34fc0d34ebbba) to the search endpoints in @xata.io/client; the shape of the records returned is now nested in an object. 

Since we're at `0.x.x`, only @xata.io/client versions at`0.25.x` are valid with the specified peer dependency range of `^0.25.1`. Depending on the package manager, the user would see an error or a warning that the dependency is not met, so I'm not sure "Fix records?.map is not a function error" is a bug per se, because when the peer dependency is actually met at `0.25.x` the error does not occur.

That being said, we probably want to bump the  @xata.io/client version eventually. I'll let @tsg chime in here. Note that making this code change would also require bumping all @xata.io/client dependencies inside `langchain/package.json` to `^0.27.0` and modifying any other Xata search calls to handle the new shape.

jacoblee93:
Ah I see. Yes, we will also need to bump the Xata peer dependency as well then to avoid people on the old versions from experiencing breaking changes.

tsg:
Thanks everyone, especially @codenameakshay for raising this! I've opened https://github.com/langchain-ai/langchainjs/pull/3553 which includes also the xata client upgrade.

@codenameakshay if you prefer to bring the upgrade in this PR, you basically need to set the new version in `package.json`, and run `yarn install`.

jacoblee93:
Fixed in #3553!

PR #3527, opened 2023-12-04
core[docs]: Docs & example for runnable history

bracesproul:


docs/core_docs/docs/expression_language/cookbook/with_history.mdx
```diff
@@ -0,0 +1,6 @@
+import CodeBlock from "@theme/CodeBlock";
+import Example from "@examples/guides/expression_language/runnable_history.ts";
+
+# Runnables With History
+
+<CodeBlock language="typescript">{Example}</CodeBlock>
```
jacoblee93, line 6: Would call this `Runnables With Chat History` and remove the `Memory` section from cookbook sidebar (don't remove page)

examples/src/guides/expression_language/runnable_history.ts
```diff
@@ -0,0 +1,90 @@
...
+  model,
+  new StringOutputParser(),
+]);
+
+// Create your `RunnableWithMessageHistory` object, passing in the
+// runnable created above.
+const withHistory = new RunnableWithMessageHistory({
+  runnable,
+  config: {},
+  getMessageHistory: await getListSessionHistory(),
```
jacoblee93, line 58: Don't think this needs to be `async`

examples/src/guides/expression_language/runnable_history.ts
```diff
@@ -0,0 +1,90 @@
...
+> {
+  const chatHistoryStore: { [key: string]: BaseListChatMessageHistory } = {};
+
+  async function getSessionHistory(
+    sessionId: string
+  ): Promise<BaseListChatMessageHistory> {
+    if (!(sessionId in chatHistoryStore)) {
+      chatHistoryStore[sessionId] = new ChatMessageHistory();
+    }
+    return chatHistoryStore[sessionId];
```
jacoblee93, line 29: Why double wrap this?

bracesproul: Not sure, good catch

examples/src/guides/expression_language/runnable_history.ts
```diff
@@ -0,0 +1,90 @@
...
+const model = new ChatOpenAI({});
+const prompt = ChatPromptTemplate.fromMessages([
+  ["ai", "You are a helpful assistant."],
+  ["human", "{question}"],
+]);
+
+// Create a simple runnable which passes a question to the AI.
+const runnable = RunnableSequence.from([
+  {
+    question: (messages: Array<BaseMessage>) =>
```
jacoblee93, line 45: Why is the input an array of messages?

bracesproul: Current message + history

jacoblee93: This should not be necessary, the whole point of this is so that we don't have to manage history externally

PR #3537, opened 2023-12-05
core[tests]: Better tests for runnable history

bracesproul:
Fixes # (issue)

### DIFF
--- a/langchain-core/src/runnables/tests/runnable_history.test.ts
+++ b/langchain-core/src/runnables/tests/runnable_history.test.ts
@@ -8,8 +8,10 @@ import {
 } from "../../chat_history.js";
 import {
   FakeChatMessageHistory,
+  FakeLLM,
   FakeListChatMessageHistory,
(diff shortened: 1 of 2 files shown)


langchain-core/src/runnables/tests/runnable_history.test.ts
```diff
@@ -90,3 +92,31 @@ test("Runnable with message history work with chat list memory", async () => {
...
+    new MessagesPlaceholder("history"),
+    ["human", "{input}"],
+  ]);
+  const model = new FakeLLM({});
+  const chain = prompt.pipe(model);
+
+  const getListMessageHistory = await getListSessionHistory();
+  const withHistory = new RunnableWithMessageHistory({
+    runnable: chain,
+    config: {},
```
jacoblee93, line 108: Do we need to pass `config`?

bracesproul: no I can make this optional

PR #2206, opened 2023-08-08
Support for Vectara's Context Config, Generative Summarization, and more

Kae7in:
## Add support for contextConfig, rerankingConfig, summary, and start position.

This PR gives users of LangChainJS's Vectara integration more fine-grained control provided by [Vectara's Search API](https://docs.vectara.com/docs/api-reference/search-apis/search).
- Optionally, the query context (`contextConfig`) provides additional information that the system may use to refine the results.
- The `start` field controls the starting position within the list of results
- The `rerankingConfig` enables reranking of results, to further increase relevance in certain scenarios
- The `summary` field allows you to use Vectara's "Gounded Generation", a groundbreaking way of producing generative summaries on top of your own data.

---

About me:
- Twitter handle: [`@KaelinHooper`](https://twitter.com/KaelinHooper)
- Website: [kaelin.io](https://kaelin.io/)

PR #3541, opened 2023-12-05
langchain[minor]: Expanded Support for Vectara's Search API and integrated summarization

ofermend:
This is a refactor of Vectara, based on the suggestion from https://github.com/langchain-ai/langchainjs/pull/2206 and guidance from @jacoblee93 for how to integrate summarization.
Also updated the interface to support more functionality from Vectara's query API, including MMR.

Original credits go to @Kae7in; completed by @ofermend


langchain/src/vectorstores/vectara.ts
```diff
@@ -41,30 +49,177 @@ export interface VectaraFile {
...
+    this.tags = fields?.tags ?? [];
+    this.metadata = fields?.metadata ?? {};
+    this.verbose = fields?.verbose ?? false;
+    this.vectara = fields.vectara;
+    this.topK = fields.topK ?? 10;
+    this.summaryConfig = fields.summaryConfig ?? { enabled: false, maxSummarizedResults: 0, responseLang: "eng" };
+  }
+
+  async invoke(input: string, options?: RunnableConfig): Promise<Document[]> {
+    return this.getRelevantDocuments(input, options);
```
jacoblee93, line 171: This is already present in the base class, no need to override:

https://github.com/langchain-ai/langchainjs/blob/main/langchain-core/src/retrievers.ts#L56

langchain/src/vectorstores/base.ts
```diff
@@ -198,7 +198,7 @@ export abstract class VectorStore extends Serializable {
       filter
     );
   }
-
+
```
jacoblee93, line 201: Remember to run `yarn format`!

langchain/src/vectorstores/vectara.ts
```diff
@@ -41,30 +49,177 @@ export interface VectaraFile {
...
+  metadata?: Record<string, unknown>;
+  verbose?: boolean;
+}
+
+export class VectaraRetriever extends Runnable<string, Document[]> {
+  static lc_name() {
+    return "VectaraRetriever";
+  }
+
+  lc_namespace = ["langchain", "retrievers", "vectaraRetriever"];
```
jacoblee93, line 143: We should move this to `langchain/src/retrievers` and call this `VectaraHybridSearchRetriever`?

langchain/src/vectorstores/vectara.ts
```diff
@@ -41,30 +49,177 @@ export interface VectaraFile {
...
+    );
+    const runManager = await callbackManager_?.handleRetrieverStart(
+      this.toJSON(),
+      query,
+      undefined,
+      undefined,
+      undefined,
+      undefined,
+      parsedConfig.runName
+    );
```
jacoblee93, line 198: Handling run managers and this awful API are all unnecessary if you override `_getRelevantDocuments` instead of this class:

https://github.com/langchain-ai/langchainjs/blob/main/langchain-core/src/retrievers.ts#L70

langchain/src/vectorstores/vectara.ts
```diff
@@ -361,37 +516,38 @@ export class VectaraStore extends VectorStore {
...
-   * Performs a similarity search and returns documents along with their
-   * scores.
+   * Performs a Vectara API call based on the arguments provided.
    * @param query The query string for the similarity search.
    * @param k Optional. The number of results to return. Default is 10.
    * @param filter Optional. A VectaraFilter object to refine the search results.
    * @returns A Promise that resolves to an array of tuples, each containing a Document and its score.
    */
-  async similaritySearchWithScore(
+  async vectara_query(
```
jacoblee93, line 525: Prefer camel case

langchain/src/vectorstores/vectara.ts
```diff
@@ -361,37 +516,38 @@ export class VectaraStore extends VectorStore {
...
       query: [
         {
           query,
-          numResults: k,
-          contextConfig: {
-            sentencesAfter: filter?.contextConfig?.sentencesAfter ?? 2,
-            sentencesBefore: filter?.contextConfig?.sentencesBefore ?? 2,
-          },
+          start,
+          numResults: mmrConfig?.enabled ? mmrConfig.mmrTopK : k,
```
jacoblee93, line 546: I think a lot of this would ideally be handled by the retriever created from `vectorstore.asRetriever()`? Or is it that we want this retriever to be able to do everything?

jacoblee93 (commented):
Thank you! See comments

langchain/src/vectorstores/vectara.ts
ofermend: But notice VectaraRetriever is does not extend BaseRetriever. won't it be necessary in this case?

langchain/src/vectorstores/vectara.ts
ofermend: The issue is that since my getRelevantDocs has "summary" argument that does not fit with the signature of _getRelevantDocuments(). Is there an easy way around it?
This is why I did not inherit from BaseRetriever and had to copu/paste all this API stuff.

langchain/src/vectorstores/vectara.ts
ofermend: Yes VectaraSummaryRetriever sounds good. Will make that change.

langchain/src/vectorstores/vectara.ts
jacoblee93: Why not just pass `summary` as a constructor argument and extend `BaseRetriever`?

ofermend: yes that should work. Thanks!

langchain/src/vectorstores/vectara.ts
ofermend: See the new version. Separated VectaraStore from VectaraSummaryRetriever in a different file. For this comment, do you mean that the vectaraQuery() function belongs more in the retriever than in the store? (and store can call it in e.g. similarity_search as a retriever?)
If so, would the constructor generate the retriever inside the VectaraStore?

PR #3548, opened 2023-12-05
langchain[minor]: Experimental Masking Module

Jordan-Gilliam:
#### Overview:
Dependencies: none
Tag maintainer: @jacoblee93 
Co-Authors: @ddzmitry @HarishRN @dgaleally

This pull request introduces the MaskingParser and the RegexMaskingTransformer an extendable module for masking and rehydrating strings. One of the primary use cases for this module is to redact PII (Personal Identifiable Information) from a string before making a call to an llm.

### Key Features:
- Add transformers for custom masking logic
- Parse strings
- Rehydrate masked strings
- State Management (Maintains a state map of masked values to their orignial counterparts - essential for rehydration)
- Hooks to run callback functions at various stages. 
- Error handling
- Overridable masking hash function (eg.[email-123***])
- Transformation Logic
- Rehydration Logic

##### Langhchainjs core
Introduction of the experimental/masking module
MaskingParser class,
Abstract MaskingTransformer
RegexMaskingTransformer class extending the abstract MaskingTransformer

##### Documentation
Experimental Module Docs 
Masking Docs
Nextjs masking example
Kitchen Sink masking example

##### Impact:
Inverted controls for consumers to mask and rehydrate strings
Groundwork for more comprehensive masking integrations in the future (eg. AWS comprehend)


examples/src/experimental/masking/next.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    // Stream the AI response based on the masked chat history and current message
+    const stream = await chain.stream({
+      chat_history: guardedHistory,
+      input: guardedMessageContent,
+    });
+
+    // npm i ai
+    // import { StreamingTextResponse } from "ai";
+    // return new StreamingTextResponse(stream);
```
jacoblee93, line 67: This can just be:

```typescript
return new Response(stream, {
  headers: {
    "content-type": "text/plain; charset=utf-8"
  }
});
```

docs/core_docs/docs/modules/experimental/mask/mask.mdx
```diff
@@ -0,0 +1,64 @@
...
+const phoneMask = () => `[phone-${Math.random().toString(16).slice(2)}]`;
+
+// Configure pii transformer
+const piiMaskingTransformer = new RegexMaskingTransformer({
+  email: { regex: /\S+@\S+\.\S+/g, mask: emailMask },
+  phone: { regex: /\d{3}-\d{3}-\d{4}/g, mask: phoneMask },
+});
+
+maskingParser = new MaskingParser();
+maskingParser.addTransformer(piiMaskingTransformer);
```
jacoblee93, line 30: Would it make more sense for the constructor to just take an array of `MaskingTransformers`?

langchain/src/experimental/masking/parser.ts
```diff
@@ -0,0 +1,139 @@
...
+/**
+ * MaskingParser class for handling the masking and rehydrating of messages.
+ */
+export class MaskingParser {
+  private transformers: MaskingTransformer[];
+  private state: Map<string, string>;
+  private config: MaskingParserConfig;
+
+  constructor(config: MaskingParserConfig = {}) {
+    this.transformers = config.transformers || [];
```
jacoblee93, line 13: Ah ok I see you have it as an option here - also a bit more conventional for us to prefer `??` but doesn't really matter

langchain/src/experimental/masking/parser.ts
```diff
@@ -0,0 +1,139 @@
...
+    }
+    // Check if any transformers have been added to the parser.
+    // If no transformers are present, throw an error as rehydration requires at least one transformer.
+    if (this.transformers.length === 0) {
+      throw new Error(
+        "MaskingParser.rehydrate Error: No transformers have been added. Please add at least one transformer before rehydrating."
+      );
+    }
+
+    if (state && !(state instanceof Map)) {
```
jacoblee93, line 112: Our linter does not allow `instanceof` since bundlers/chunking can make it unreliable in certain cases - I think it's fine here but you'll need to disable it with a comment:

https://github.com/langchain-ai/langchainjs/blob/dfd32715c44e9bc63f98bf5b8fd93dbd62b3320c/langchain-core/src/prompts/chat.ts#L502

langchain/src/experimental/masking/parser.ts
```diff
@@ -0,0 +1,139 @@
...
+   * Masks the provided message using the added transformers.
+   * This method sequentially applies each transformer's masking logic to the message.
+   * It utilizes a state map to track original values corresponding to their masked versions.
+   *
+   * @param message - The message to be masked.
+   * @returns A masked version of the message.
+   * @throws {TypeError} If the message is not a string.
+   * @throws {Error} If no transformers are added.
+   */
+  async parse(message: string): Promise<string> {
```
jacoblee93, line 44: `parse` -> `mask`?

langchain/src/experimental/masking/parser.ts
```diff
@@ -0,0 +1,139 @@
...
+    // It starts as the original message and gets transformed by each transformer.
+    let processedMessage = message;
+
+    // Iterate through each transformer added to the parser.
+    this.transformers.forEach((transformer) => {
+      // Apply the transformer's transform method to the current state of the message.
+      // The transform method returns a tuple containing the updated message and state.
+      // The state is a map that tracks the original values of masked content.
+      // This state is essential for the rehydration process to restore the original message.
+      [processedMessage, this.state] = transformer.transform(
```
jacoblee93, line 71: I think it might be nice to allow `async` transform functions instead of just sync?

langchain/src/experimental/masking/parser.ts
```diff
@@ -0,0 +1,139 @@
...
+    // This variable will undergo rehydration by each transformer in reverse order.
+    let rehydratedMessage = message;
+    this.transformers
+      .slice()
+      .reverse()
+      .forEach((transformer) => {
+        // Apply the transformer's rehydrate method to the current state of the message.
+        // The rehydrate method uses the stored state (this.state) to map masked values
+        // back to their original values, effectively undoing the masking transformation.
+        rehydratedMessage = transformer.rehydrate(
```
jacoblee93, line 129: See above - might be nice to allow `async` `rehydrate` functions as well and then just use a `for` loop here

langchain/src/experimental/masking/parser.ts
jacoblee93: Only other comment here is that I don't love mutating `this.state` internally within `transformer.transform` - could see some nasty bugs where a transformer implementation e.g. mistakenly clears state from previous transformers.

My preference would be to manually combine state at the parser level based on each transformer's output, but fine if you want to stick with this since it's in experimental.

langchain/src/experimental/masking/parser.ts
```diff
@@ -0,0 +1,139 @@
...
+   * that has been transformed (masked) by the transformers. This process is the inverse of the masking process.
+   *
+   * @param message - The masked message to be rehydrated.
+   * @returns The original (rehydrated) version of the message.
+   */
+  async rehydrate(
+    message: string,
+    state?: Map<string, string>
+  ): Promise<string> {
+    this.config.onRehydratingStart?.(message);
```
jacoblee93, line 97: Should this block?

langchain/src/experimental/masking/parser.ts
```diff
@@ -0,0 +1,139 @@
...
+   * This method sequentially applies each transformer's masking logic to the message.
+   * It utilizes a state map to track original values corresponding to their masked versions.
+   *
+   * @param message - The message to be masked.
+   * @returns A masked version of the message.
+   * @throws {TypeError} If the message is not a string.
+   * @throws {Error} If no transformers are added.
+   */
+  async parse(message: string): Promise<string> {
+    this.config.onMaskingStart?.(message);
```
jacoblee93, line 45: Should this block?

langchain/src/experimental/masking/parser.ts
```diff
@@ -0,0 +1,139 @@
...
+  addTransformer(transformer: MaskingTransformer) {
+    this.transformers.push(transformer);
+  }
+
+  /**
+   * Getter method for retrieving the current state.
+   * @returns The current state map.
+   */
+  public getState(): Map<string, string> {
+    return this.state;
```
jacoblee93, line 31: Is this necessary?

docs/core_docs/docs/modules/index.mdx
```diff
@@ -29,3 +29,7 @@ Persist application state between runs of a chain
 #### [Callbacks](/docs/modules/callbacks/)
 
 Log and stream intermediate steps of any chain
+
+#### [Experimental](/docs/modules/experimental/)
+
+Experimental modules not yet ready for production consumption
```
jacoblee93, line 35: I would phrase this more as `Experimental modules whose abstractions have not fully settled`

examples/src/experimental/masking/kitchen_sink.ts
```diff
@@ -0,0 +1,80 @@
...
+  },
+  dob: { regex: /\b\d{4}-\d{2}-\d{2}\b/g, mask: dobMask },
+  bankAccount: { regex: /\b\d{8,17}\b/g, mask: bankAccountMask },
+};
+
+// Create a RegexMaskingTransformer with multiple patterns
+const piiMaskingTransformer = new RegexMaskingTransformer(patterns);
+
+// Hooks for different stages of masking and rehydrating
+const onMaskingStart = (message: string) =>
```
jacoblee93, line 49: Allow async functions here?

langchain/src/experimental/masking/parser.ts
```diff
@@ -0,0 +1,139 @@
...
+    this.transformers = config.transformers || [];
+    this.state = new Map<string, string>();
+    this.config = config;
+  }
+
+  /**
+   * Adds a transformer to the parser.
+   * @param transformer - An instance of a class extending MaskingTransformer.
+   */
+  addTransformer(transformer: MaskingTransformer) {
```
jacoblee93, line 22: nit: disallow mutation after initialization?

jacoblee93 (commented):
Broadly looks great! Some questions/nits

docs/core_docs/docs/modules/experimental/mask/mask.mdx
Jordan-Gilliam: I will update the basic example to default to an array of transformers.

```typescript
const piiMaskingTransformer = new RegexMaskingTransformer({
  email: { regex: /\S+@\S+\.\S+/g, mask: emailMask },
  phone: { regex: /\d{3}-\d{3}-\d{4}/g, mask: phoneMask },
});

const maskingParser = new MaskingParser({transformers: [piiMaskingTransformer]});

File: langchain/src/experimental/masking/parser.ts (hunk 10 above)
Comment (Jordan-Gilliam) [reply langchain/src/experimental/masking/parser.ts:31]: This is necessary to consume and rehydrate state asynchronously. Imagine lambda one masks a string -> mask state stored in kv -> lambda x needs the state to rehydrate the masked values.

File: langchain/src/experimental/masking/parser.ts (hunk 13 above)
Comment (Jordan-Gilliam) [reply langchain/src/experimental/masking/parser.ts:22]: Thoughts on an init flag? Allow the addition of transformers before the parser is used, but disallows it once parsing has begun.

```typescript 
export class MaskingParser {
  private transformers: MaskingTransformer[];
  private state: Map<string, string>;
  private config: MaskingParserConfig;
  private isParsingStarted: boolean;  // New flag to track if parsing has started

  constructor(config: MaskingParserConfig = {}) {
    this.transformers = config.transformers || [];
    this.state = new Map<string, string>();
    this.config = config;
    this.isParsingStarted = false;  // Initialize the flag
  }

  addTransformer(transformer: MaskingTransformer) {
    if (this.isParsingStarted) {
      throw new Error("Cannot add transformers after parsing has started.");
    }
    this.transformers.push(transformer);
  }

  private startParsing() {
    this.isParsingStarted = true;  // Set the flag when parsing starts
  }

  async parse(message: string): Promise<string> {
    this.startParsing();  // Call this method at the start of parsing
    // ... rest of the parse method
  }

  async rehydrate(message: string, state?: Map<string, string>): Promise<string> {
    this.startParsing();  // Call this method at the start of rehydrating
    // ... rest of the rehydrate method
  }
}

Comment (jacoblee93) [reply langchain/src/experimental/masking/parser.ts:22]: I think a bit overcomplicated - would rather just leave it the way it is if you prefer this pattern, though my preference is for no mutation after initialization unless there's a good use-case for it.

File: langchain/src/experimental/masking/parser.ts (hunk 10 above)
Comment (jacoblee93) [reply langchain/src/experimental/masking/parser.ts:31]: Right, but this method vs. just having a public property?

File: langchain/src/experimental/masking/parser.ts [hunk 14]
```diff
@@ -0,0 +1,172 @@
...
+   * @returns A masked version of the message.
+   * @throws {TypeError} If the message is not a string.
+   * @throws {Error} If no transformers are added.
+   */
+  async mask(message: string): Promise<string> {
+    // If onMaskingStart is a function, handle it accordingly
+    if (this.config.onMaskingStart) {
+      try {
+        const result = this.config.onMaskingStart(message) as any;
+        if (result && typeof result.then === "function") {
```
jacoblee93, line 49: I think you can just do `await this.config.onMaskingStart(message)`, will work for sync functions too?

langchain/src/experimental/masking/parser.ts
```diff
@@ -0,0 +1,172 @@
...
+
+    // Check if there are any transformers added to the parser. If not, throw an error
+    // as masking requires at least one transformer to apply its logic.
+    if (this.transformers.length === 0) {
+      throw new Error(
+        "MaskingParser.mask Error: No transformers have been added. Please add at least one transformer before parsing."
+      );
+    }
+
+    if (typeof message !== "string") {
```
jacoblee93, line 65: IMO you don't have to be so defensive, the typing will take care of things, but no problem to keep if you're worried about vanilla JS devs

jacoblee93:
Thank you!

PR #3553, opened 2023-12-05
Upgrade xata client to 0.28.0 and apply required change

tsg:
This is a follow-up/replacement for #3524 that also upgrades the Xata client.

In addition, it also updates the column type names in the dos, and documents a bit better how to run the integration tests.


jacoblee93:
Thank you!

jacoblee93:
Thank you!

PR #3555, opened 2023-12-05
langchain[minor]: remote runnable stream log, additional serialization updates

bracesproul:


jacoblee93:
CC @efriis

langchain/src/runnables/remote.ts
```diff
@@ -325,4 +335,73 @@ export class RemoteRunnable<
...
+        "exclude_types",
+        "exclude_tags",
+      ];
+      const extraOptions: Record<string, unknown> = {};
+      for (const key of optionKeys) {
+        const keyAsKeyof = key as keyof RunInput;
+        if (input[keyAsKeyof] !== undefined && input[keyAsKeyof] !== null) {
+          extraOptions[key] = input[keyAsKeyof];
+        }
+      }
```
efriis, line 376: in python we force `diff=True` I think for stream log

langchain/src/runnables/remote.ts
```diff
@@ -325,4 +335,73 @@ export class RemoteRunnable<
...
+      kwargs,
+      ...getExtraOptionsFromInput(),
+    });
+    const { body } = response;
+    if (!body) {
+      throw new Error(
+        "Could not begin remote stream log. Please check the given URL and try again."
+      );
+    }
+    const runnableStream = convertEventStreamToIterableReadableDataStream(body);
```
efriis, line 402: not super familiar here - will `body` be the ReadableStream that this function expects?

Should confirm the objects yielded here are the same as from the langserve python client. I think this might be missing some of the `jsonpatch` mechanics

efriis (commented):
Might be worth pasting what we expect to see from the integration tests based on the output from the python client

langchain/src/runnables/remote.ts
bracesproul: Yes it should. The `body` object is of type `ReadableStream<Uint8Array> | null`. Not 100% on the `jsonpatch` though, could you provide some context as to what you're looking for?

langchain/src/runnables/remote.ts
bracesproul: Added that as default, and it'll throw if you try and pass `false`.

langchain/src/runnables/remote.ts
efriis: https://github.com/langchain-ai/langserve/blob/main/langserve/client.py#L673-L677

It's possible this is implemented somewhere else. Can test tomorrow if helpful, and just want to confirm this yields equivalent objects as LangServe does, and not just the patch ops

https://jsonpatch.com/

bracesproul: @efriis just pushed a change which should satisfy this, lmk thoughts

langchain/src/runnables/tests/runnable_remote.test.ts
```diff
@@ -160,4 +161,19 @@ describe("RemoteRunnable", () => {
       '"object1, object2, object3, object4, object5"'
     );
   });
+
+  test("Stream log local langserve", async () => {
+    const remote = new RemoteRunnable({ url: `${BASE_URL}/a` });
+    const stream = remote.streamLog({
```
efriis, line 167: could you add a unit test confirming the jsonpatch logic is working

langchain/src/runnables/remote.ts
```diff
@@ -325,4 +335,104 @@ export class RemoteRunnable<
...
+    async function consumeRunnableStream() {
+      try {
+        for await (const chunk of runnableStream) {
+          const patch = new RunLogPatch({
```
efriis, line 414: why are we injecting this new patch?

efriis:
Python code for the test we were running
```
from langserve.client import RemoteRunnable

r = RemoteRunnable("https://chat-langchain-backend.langchain.dev/chat")

async for g in r.astream_log(
    {
        "question": "What is this?",
    }
):
    print(g)
```

docs/core_docs/docs/expression_language/how_to/stream_log.mdx
```diff
@@ -0,0 +1,14 @@
...
+
+# Stream Intermediate Steps
+
+All runnables also have a method `streamLog()` which is used to stream (as they happen) all or part of the intermediate steps of your chain/sequence.
+
+This is useful to show progress to the user, to use intermediate results, or to debug your chain.
+
+You can stream all steps (default) or include/exclude steps by name, tags or metadata.
+
+This method yields [JSONPatch](https://jsonpatch.com/) ops that when applied in the same order as received build up the RunState.
```
jacoblee93, line 12: Would add an example of how to actually reconstruct the JSON object too with a package

examples/src/guides/expression_language/stream_log.ts
```diff
@@ -0,0 +1,74 @@
...
+----------------------------------------
+[{ op: 'add', path: '/streamed_output/-', value: AIMessageChunk { content: ' worked', additional_kwargs: {} }}]
+----------------------------------------
+[{ op: 'add', path: '/streamed_output/-', value: AIMessageChunk { content: ' at', additional_kwargs: {} }}]
+----------------------------------------
+[{ op: 'add', path: '/streamed_output/-', value: AIMessageChunk { content: ' Kens', additional_kwargs: {} }}]
+----------------------------------------
+[{ op: 'add', path: '/streamed_output/-', value: AIMessageChunk { content: 'ho', additional_kwargs: {} }}]
+----------------------------------------
+[{ op: 'add', path: '/streamed_output/-', value: AIMessageChunk { content: '.', additional_kwargs: {} }}]
```
jacoblee93, line 69: I think `value` should be serialized here?

jacoblee93:
@eyurtsev @efriis I updated the reviver function - could one of you take a look?

It seems like the format may have changed a fair bit (at least from my testing):

https://github.com/langchain-ai/langchainjs/pull/3555/files#diff-38d490399a9e82aa492c4ab547da501edde65685b571dad5cc228f84d7e5016cR69

langchain/src/runnables/tests/runnable_remote.test.ts
```diff
@@ -25,58 +25,58 @@ data: ["a", "b", "c", "d"]
 event: end`;
 
 const bResp = `event: data
-data: {"content": "", "additional_kwargs": {}, "type": "ai", "example": false, "is_chunk": true}
+data: {"content": "", "additional_kwargs": {}, "type": "AIMessageChunk", "example": false}
```
eyurtsev, line 28: The new tests are correct here

langchain/src/runnables/remote.ts
```diff
@@ -47,83 +54,110 @@ function isSuperset(set: Set<string>, subset: Set<string>) {
 function revive(obj: any): any {
```
eyurtsev, line 54: Could you double check that this client contains the same coverage as the list of well known types in python?

https://github.com/langchain-ai/langserve/blob/main/langserve/serialization.py#L54

eyurtsev (approved):

langchain/src/runnables/remote.ts
jacoblee93: I think we are not handling agent types properly, will update.

jacoblee93: `ChatPromptValueConcrete` also does not exist in JS, will figure out what to do there

eyurtsev (approved):

PR #3564, opened 2023-12-06
core[chore]: widen semver range for langsmith

falkenhawk:
caret in ^0.0.x doesn't do anything, it matches only the exact specified version, whereas tilde ~0.0.x matches that and all 0.0.x versions up to 0.1.0

https://nodesource.com/blog/semver-tilde-and-caret/#majorandminorzero00z00z

Currently it results in two versions of `langsmith` to be installed in `node_modules` together with `langchain` (`0.0.48` and `0.0.49`), as `langchain` permits `~0.0.48` for `langsmith`, which installs the latest available `0.0.x` version, and `@langchain/core` allows only `0.0.48` with `^0.0.48`.

`yarn why langsmith`
288361329-9cbfd7e9-b153-4d54-b767-686f8949bd0d

### DIFF
--- a/langchain-core/package.json
+++ b/langchain-core/package.json
@@ -39,7 +39,7 @@
     "camelcase": "6",
     "decamelize": "1.2.0",
     "js-tiktoken": "^1.0.8",
-    "langsmith": "^0.0.48",
+    "langsmith": "~0.0.48",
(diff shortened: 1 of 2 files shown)


jacoblee93:
Thanks, nice catch!

PR #3568, opened 2023-12-06
langchain[patch]: Update default PDF split lines separator

jacoblee93:


jordanburke:
Hi @jacoblee93 just ran into this change. The langchain default document says otherwise: https://js.langchain.com/docs/modules/data_connection/document_loaders/pdf#usage-one-document-per-page either the code or the doc should probably change.

PR #3577, opened 2023-12-07
feat: added ibm watsonx chat model and streaming to base llm model

faileon:
Hi everyone,
as per the great PR #3399 based on the #2378  I added my implementation of ChatModel class and streaming capabilities to both base LLM model and the Chat Model.

I am not very familiar with turborepo, so I hope this structure is fine. Please let me know anything that should be changed.

### DIFF
--- a/docs/api_refs/typedoc.json
+++ b/docs/api_refs/typedoc.json
@@ -204,6 +204,7 @@
     "./langchain/src/chat_models/llama_cpp.ts",
     "./langchain/src/chat_models/yandex.ts",
     "./langchain/src/chat_models/fake.ts",
+    "./langchain/src/chat_models/watsonx_ai.ts",
     "./langchain/src/schema/index.ts",
     "./langchain/src/schema/document.ts",
     "./langchain/src/schema/output_parser.ts",
--- a/examples/src/llms/watsonx_ai-chat.ts
+++ b/examples/src/llms/watsonx_ai-chat.ts
@@ -1,0 +1,20 @@
+import { WatsonChatModel } from "langchain/chat_models/watsonx_ai";
+
+const model = new WatsonChatModel({
+  clientConfig: {
+    region: "eu-de",
+  },
+  modelParameters: {
+    max_new_tokens: 100,
+  },
+});
+
+const stream = await model.stream(
+  "What would be a good company name for a company that makes colorful socks?"
+);
+
+let text = "";
+for await (const chunk of stream) {
+  text += chunk.content;
+  console.log(text);
+}
--- a/langchain/.gitignore
+++ b/langchain/.gitignore
@@ -556,6 +556,9 @@ chat_models/yandex.d.ts
 chat_models/fake.cjs
 chat_models/fake.js
 chat_models/fake.d.ts
+chat_models/watsonx_ai.cjs
+chat_models/watsonx_ai.js
+chat_models/watsonx_ai.d.ts
 schema.cjs
 schema.js
 schema.d.ts
--- a/langchain/package.json
+++ b/langchain/package.json
@@ -568,6 +568,9 @@
     "chat_models/fake.cjs",
     "chat_models/fake.js",
     "chat_models/fake.d.ts",
+    "chat_models/watsonx_ai.cjs",
+    "chat_models/watsonx_ai.js",
+    "chat_models/watsonx_ai.d.ts",
     "schema.cjs",
     "schema.js",
     "schema.d.ts",
@@ -2397,6 +2400,11 @@
       "types": "./chat_models/fake.d.ts",
       "import": "./chat_models/fake.js",
       "require": "./chat_models/fake.cjs"
+    },
+    "./chat_models/watsonx_ai": {
+      "types": "./chat_models/watsonx_ai.d.ts",
+      "import": "./chat_models/watsonx_ai.js",
+      "require": "./chat_models/watsonx_ai.cjs"
     },
     "./schema": {
       "types": "./schema.d.ts",
--- a/langchain/scripts/create-entrypoints.js
+++ b/langchain/scripts/create-entrypoints.js
@@ -217,6 +217,7 @@ const entrypoints = {
   "chat_models/llama_cpp": "chat_models/llama_cpp",
   "chat_models/yandex": "chat_models/yandex",
   "chat_models/fake": "chat_models/fake",
+  "chat_models/watsonx_ai": "chat_models/watsonx_ai",
   // schema
   schema: "schema/index",
   "schema/document": "schema/document",
--- a/langchain/src/chat_models/watsonx_ai.ts
+++ b/langchain/src/chat_models/watsonx_ai.ts
@@ -1,0 +1,154 @@
+import { getEnvironmentVariable } from "@langchain/core/utils/env";
+import { CallbackManagerForLLMRun } from "../callbacks/index.js";
+import {
+  AIMessageChunk,
+  BaseMessage,
+  ChatGenerationChunk,
+  ChatMessage,
+} from "../schema/index.js";
+import {
+  WatsonModelParameters,
+  WatsonxAIParams,
+} from "../types/watsonx-types.js";
+import { WatsonApiClient } from "../util/watsonx-client.js";
+import { BaseChatModelParams, SimpleChatModel } from "./base.js";
+
+export class WatsonChatModel extends SimpleChatModel {
+  private readonly watsonApiClient!: WatsonApiClient;
+
+  readonly modelId!: string;
+
+  readonly modelParameters?: WatsonModelParameters;
+
+  readonly projectId!: string;
+
+  constructor(fields: WatsonxAIParams & BaseChatModelParams) {
+    super(fields);
+
+    const {
+      clientConfig = {},
+      modelId = "meta-llama/llama-2-70b-chat",
+      modelParameters,
+      projectId = getEnvironmentVariable("WATSONX_PROJECT_ID") ?? "",
+    } = fields;
+
+    this.modelId = modelId;
+    this.modelParameters = modelParameters;
+    this.projectId = projectId;
+
+    const {
+      apiKey = getEnvironmentVariable("IBM_CLOUD_API_KEY"),
+      apiVersion = "2023-05-29",
+      region = "us-south",
+    } = clientConfig;
+
+    if (!apiKey) {
+      throw new Error("Missing IBM Cloud API Key");
+    }
+
+    if (!this.projectId) {
+      throw new Error("Missing WatsonX AI Project ID");
+    }
+
+    this.watsonApiClient = new WatsonApiClient({
+      apiKey,
+      apiVersion,
+      region,
+    });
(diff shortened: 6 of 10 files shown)


jacoblee93:
Thanks for your patience! @faileon @chasemcdo do you think you could have a last look? Moved around to account for recent refactoring

libs/langchain-community/src/chat_models/watsonx_ai.ts
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    this.watsonApiClient = new WatsonApiClient({
+      apiKey,
+      apiVersion,
+      region,
+    });
+  }
+
+  protected _formatMessagesAsPrompt(messages: BaseMessage[]): string {
```
jacoblee93, line 63: We should make a `Llama` adapter or something...

libs/langchain-community/src/llms/watsonx_ai.ts
```diff
@@ -69,48 +30,43 @@ export class WatsonxAI extends LLM<BaseLLMCallOptions> {
...
 
-  projectId?: string;
-
-  modelParameters?: Record<string, unknown>;
+  private readonly watsonApiClient: WatsonApiClient;
 
   constructor(fields: WatsonxAIParams) {
     super(fields);
 
-    this.region = fields?.region ?? this.region;
```
jacoblee93, line 95: Isn't removing all of these a breaking change?

faileon: I moved a few attributes to the WatsonClient as it's something I also needed in the ChatModel. I did remove `lc_serializable = true;` by accident and will add it back in.

I tried running the provided example at `examples/src/llms/watsonx_ai.ts` and it works like before.

libs/langchain-community/src/chat_models/watsonx_ai.ts
faileon: Yes definitely, I was thinking about it too. Right now this is blatantly copy pasted from the Ollama model and assumes the user only uses llama based chat model. However Watson offers the ability to run many models, including spinning any model from HF...

libs/langchain-community/src/llms/watsonx_ai.ts
```diff
@@ -127,74 +83,50 @@ export class WatsonxAI extends LLM<BaseLLMCallOptions> {
...
-    interface TokenResponse {
-      access_token: string;
-      expiration: number;
-    }
-
-    const urlTokenParams = new URLSearchParams();
-    urlTokenParams.append(
-      "grant_type",
-      "urn:ibm:params:oauth:grant-type:apikey"
+  async *_streamResponseChunks(
```
chasemcdo, line 96: I may be missing something, but it seems like while we have the stream function for the LLM it isn't actually usable in the current state?

I was expecting something closer to the [sagemaker implementation](https://github.com/langchain-ai/langchainjs/blob/88713ea81f3d015e0c813c19cd63a9be0bf5cd68/libs/langchain-community/src/llms/sagemaker_endpoint.ts#L180-L188) which would allow easy toggling of streaming.

Though the present `_call` function appears to directly call `generate_text`.

libs/langchain-community/src/types/watsonx-types.ts
```diff
@@ -0,0 +1,58 @@
...
+export interface WatsonModelParameters {
+  decoding_method?: "sample" | "greedy";
+  max_new_tokens?: number;
+  min_new_tokens?: number;
+  stop_sequences?: string[];
+  temperature?: number;
+  top_k?: number;
+  top_p?: number;
+  repetition_penalty?: number;
+}
```
chasemcdo, line 12: I had originally left this undefined to avoid missing parameters, but looks like you've done a great job of adding them!

One thing to add might be the [Random seed](https://dataplatform.cloud.ibm.com/docs/content/wsj/analyze-data/fm-model-parameters.html?context=wx#random-seed) not sure what the expected argument is off the top of my head.

libs/langchain-community/src/llms/watsonx_ai.ts
chasemcdo: It is breaking in the sense that if someone is dependent on the old schema they'll need to update it, but most of the functionality is there, but in a now more reusable format.

One thing that I am noticing missing is there was originally an [endpoint](https://github.com/langchain-ai/langchainjs/blob/06a39117aae61fcbf26a537de4bad8a156bbb9ab/libs/langchain-community/src/llms/watsonx_ai.ts#L14-L17) parameter which would allow the end user to overwrite the entire url if they so desired.

I had included this since the production endpoint is still labeled as "beta", so this would allow someone to migrate to a different (potentially more stable) endpoint as soon as it is released without the need to submit a PR to LangChain.

Thoughts on reintroducing that @faileon ?

chasemcdo (commented):
Looks awesome! Thanks for reverse engineering that

libs/langchain-community/src/llms/watsonx_ai.ts
jacoblee93: Not sure what usage is, but would love a shim for principle's sake. I can try to have a look later if you don't have time.

libs/langchain-community/src/types/watsonx-types.ts
chasemcdo, line 12: Did a bit more digging and looks like the Random Seed could be added like this:
```suggestion
export interface WatsonModelParameters {
  decoding_method?: "sample" | "greedy";
  max_new_tokens?: number;
  min_new_tokens?: number;
  stop_sequences?: string[];
  temperature?: number;
  top_k?: number;
  top_p?: number;
  repetition_penalty?: number;
  random_seed?: number;
}
```

libs/langchain-community/src/llms/watsonx_ai.ts
faileon: Ah I get it now, yes this would be breaking to someone using the old schema. I can revert it and make it backwards compatible.

Regarding the endpoint parameter that's a good catch, I agree it should be configurable, as it will most likely change once IBM releases non beta version.

I have some free time during this weekend again to work on it.

libs/langchain-community/src/types/watsonx-types.ts
faileon: Good catch, I didn't see the Random seed. I will update the model.

faileon: Awesome, thanks! I will update it.

libs/langchain-community/src/llms/watsonx_ai.ts
faileon: I was originally inspired by the way the ollama connector handles it, but let me take another look at this. I only tested the chat model I implemented for streaming and invoking.

libs/langchain-community/src/llms/watsonx_ai.ts
jacoblee93: Thank you! This is very recent, but we really try to not have any breaking changes in patches unless they are completely unavoidable.

libs/langchain-community/src/chat_models/watsonx_ai.ts
```diff
@@ -0,0 +1,157 @@
...
+export class WatsonxAIChat extends SimpleChatModel {
+  private readonly watsonApiClient: WatsonApiClient;
+
+  readonly modelId!: string;
+
+  readonly modelParameters?: WatsonModelParameters;
+
+  readonly projectId!: string;
+
+  constructor(fields: WatsonxAIParams & BaseChatModelParams) {
```
bracesproul, line 28: Lets redefine this as an interface above the class:
```typescript
interface WatsonxAIChatParams extends WatsonxAIParams, BaseChatModelParams {};
```
This way it's easy to:
1. export the class params type for use outside this file
2. add extra properties in the future.

Also, I believe all the properties on those two interfaces are optional, so we should be able to do:
```suggestion
  constructor(fields?: WatsonxAIChatParams) {
```

libs/langchain-community/src/chat_models/watsonx_ai.ts
```diff
@@ -0,0 +1,157 @@
...
+  ): Promise<string> {
+    const chunks = [];
+    const stream = this._streamResponseChunks(messages, options, runManager);
+    for await (const chunk of stream) {
+      chunks.push(chunk.message.content);
+    }
+    return chunks.join("");
+  }
+
+  override async *_streamResponseChunks(
```
bracesproul, line 105: I don't think you need the `override` here

libs/langchain-community/src/chat_models/watsonx_ai.ts
```diff
@@ -0,0 +1,157 @@
...
+    _messages: BaseMessage[],
+    _options: this["ParsedCallOptions"],
+    _runManager?: CallbackManagerForLLMRun
```
bracesproul, line 108: Drop the underscore if they're being used. Typically, variables prefixed with an underscore are unused, and the underscore is used to bypass a lint rule for no-unused-variables

libs/langchain-community/src/llms/watsonx_ai.ts
```diff
@@ -1,53 +1,16 @@
...
-  type BaseLLMParams,
   LLM,
 } from "@langchain/core/language_models/llms";
 import { getEnvironmentVariable } from "@langchain/core/utils/env";
-
-/**
- * The WatsonxAIParams interface defines the input parameters for
- * the WatsonxAI class.
- */
-export interface WatsonxAIParams extends BaseLLMParams {
```
bracesproul, line 12: If you move this interface declaration you'll need to re-export it from this file to keep backwards compatibility

libs/langchain-community/src/utils/watsonx-client.ts
```diff
@@ -0,0 +1,165 @@
...
+  token_type: string;
+  expires_in: number;
+  expiration: number;
+  scope: string;
+};
+
+export class WatsonApiClient {
+  private iamToken?: IAMTokenResponse;
+
+  private readonly apiKey!: string;
```
bracesproul, line 39: Same as comment above, no definite assignment assertions

libs/langchain-community/src/utils/watsonx-client.ts
```diff
@@ -0,0 +1,165 @@
...
+      "Content-Type": "application/x-www-form-urlencoded",
+    };
+
+    const formBody = Object.entries(payload).reduce((acc, [key, value]) => {
+      const encodedKey = encodeURIComponent(key as string);
+      const encodedValue = encodeURIComponent(value as string);
+      acc.push(`${encodedKey}=${encodedValue}`);
+      return acc;
+    }, [] as string[]);
+    const body = formBody.join("&");
```
bracesproul, line 75: Could this be better handled with `new URLSearchParams` instead?

bracesproul (commented):
thanks for this, code is looking pretty good! I left a few comments with some requests, once you've implemented please re-request my review

hntrl:
Closing due to inactivity

PR #3579, opened 2023-12-07
community[patch]: Cassandra init improvements

mieslep:
Enhancements:
* Support for index options clause
* Support for multiple values on filter 

Fixes:
* Issue with Cassandra initialization where the table/indexes were not ready before attempting to insert

### DIFF
--- a/docs/core_docs/docs/integrations/vectorstores/cassandra.mdx
+++ b/docs/core_docs/docs/integrations/vectorstores/cassandra.mdx
@@ -96,7 +96,26 @@ vector store is first created, and specifed in the constructor parameter `vector
   ...
 ```
 
-## Non-Equality Filters
+## Indices
+
+With Version 5, Cassandra introduced Storage Attached Indexes, or SAIs. These allow `WHERE` filtering without specifying
+the partition key, and allow for additional operator types such as non-equalities. You can define these with the `indices`
+parameter, which accepts zero or more dictionaries each containing `name` and `value` entries.
+
+Indices are optional, though required if using filtered queries on non-partition columns.
+
+- The `name` entry is part of the object name; on a table named `test_table` an index with `name: "some_column"`
+  would be `idx_test_table_some_column`.
+- The `value` entry is the column on which the index is created, surrounded by `(` and `)`. With the above column
+  `some_column` it would be specified as `value: "(some_column)"`.
+- An optional `options` entry is a map passed to the `WITH OPTIONS =` clause of the `CREATE CUSTOM INDEX` statement.
+  The specific entries on this map are index type specific.
+
+```typescript
+  indices: [{ name: "some_column", value: "(some_column)" }],
+```
+
+## Advanced Filtering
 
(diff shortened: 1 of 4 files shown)


langchain/src/vectorstores/cassandra.ts
```diff
@@ -280,6 +286,29 @@ export class CassandraStore extends VectorStore {
    * @returns Promise that resolves when the database has been initialized.
    */
   private async initialize(): Promise<void> {
+    if (this.initializationState === 2) {
+      // Already initialized
+      return Promise.resolve();
+    }
+
+    if (this.initializationState === 1 && this.initializationPromise) {
```
jacoblee93, line 294: Do we need multiple states here? Can we just always return `initializationPromise` if it is set since it'll immediately resolve?

langchain/src/vectorstores/cassandra.ts
```diff
@@ -313,15 +342,30 @@ export class CassandraStore extends VectorStore {
     cql = `CREATE CUSTOM INDEX IF NOT EXISTS idx_vector_${this.table}
            ON ${this.keyspace}.${
       this.table
-    }(vector) USING 'StorageAttachedIndex' WITH OPTIONS = {'similarity_function': '${this.vectorType.toUpperCase()}'};`;
+    }(vector) USING 'StorageAttachedIndex' WITH OPTIONS = {'similarity_function': '${this.vectorType.toLowerCase()}'};`;
```
jacoblee93, line 345: Is this breaking?

jacoblee93:
Apologies for the delay! Seems broadly fine, LMK what you think of my comments.

I'll take care of merge conflicts, there have been a lot of structural changes recently.

langchain/src/vectorstores/cassandra.ts
mieslep: this is case insensitive in Cassandra, the docs indicate lowercase so just making consistent.

langchain/src/vectorstores/cassandra.ts
mieslep: I went back-and-forth trying to get this to play nicely. Not a fan of JS threading model at all, but in my manual testing, I was finding that I could get the server up and able to accept requests, and while the Cassandra tables/indexes were building it would accept a new read/write to those tables and fail. There were some other felt-like-race-conditions happening too.... there might be a better/cleaner way to address this. Can I suggest cleaning that up with the next tranche of changes (I want to get MMR working!)

mieslep:
Yeah found that and figured out how to refactor around them (the new structure looks an improvement! I like how the import works now.).  One thing on this refactoring was that I discovered I needed to be in the libs/langchain-community directory to run test:single , not sure if that was a me thing or a README needs updating?

jacoblee93:
Likely a needed README update but I've also found myself reaching for that command at root. We have a few rough edges around contributor process that we need to smooth out as a result of the reorg, so thanks for highlighting! Will try to get that fixed today.

jacoblee93:
Thank you!

PR #3588, opened 2023-12-08
Added multiple model prompt formats to llama_cpp chat streaming.

nigel-daniels:
This adds multiple prompt formats to the `llama_cpp` chat model to support models other than Llama2 running locally.

### DIFF
--- a/langchain/src/chat_models/llama_cpp.ts
+++ b/langchain/src/chat_models/llama_cpp.ts
@@ -30,8 +30,16 @@ export interface LlamaCppInputs
 export interface LlamaCppCallOptions extends BaseLanguageModelCallOptions {
   /** The maximum number of tokens the response should contain. */
   maxTokens?: number;
-  /** A function called when matching the provided token array */
-  onToken?: (tokens: number[]) => void;
+  /** deprecated, This function is not used. */
+ onToken?: (tokens: number[]) => void;
+  /** This lets the streaming method know which model we are using,
+   *  valid options are:
+   *  'llama2'	- A Llama2 model, this is the default.
+   *  'chatML'  - A ChatML model
+   *  'falcon'	- A Falcon model
+   *  'general'  - Any other model, uses "### Human\n", "### Assistant\n" format
+   */
+  streamingModel?: string;
 }
 
 /**
@@ -64,6 +72,8 @@ export class ChatLlamaCpp extends SimpleChatModel<LlamaCppCallOptions> {
 
   maxTokens?: number;
 
+  streamingModel?: string;
+
   temperature?: number;
 
   topK?: number;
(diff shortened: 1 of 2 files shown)


langchain/src/chat_models/llama_cpp.ts
```diff
@@ -298,11 +317,50 @@ export class ChatLlamaCpp extends SimpleChatModel<LlamaCppCallOptions> {
...
+              break;
+            case "falcon":
+              messageText = `User: ${message.content}`;
+              break;
+            case "general":
+              messageText = `### Human\n${message.content}`;
+              break;
+            default:
+              messageText = `[INST] ${message.content} [/INST]`;
+              break;
```
giladgd, line 332: @nigel-daniels Why aren't you using a [`LlamaChatSession`](https://withcatai.github.io/node-llama-cpp/api/classes/LlamaChatSession) with the relevant `ChatPromptWrapper` for it?
For example, you have a `ChatPromptWrapper` for ChatML under [`ChatMLChatPromptWrapper`](https://withcatai.github.io/node-llama-cpp/api/classes/ChatMLChatPromptWrapper)
It's better to only have one centralized place for the handling of chat formats, especially as more will get supported in `node-llama-cpp` over time.

Furthermore, in the next major version of `node-llama-cpp` it'll figure out the right chat wrapper by default, so it won't require a manual config most of the time.

langchain/src/chat_models/llama_cpp.ts
```diff
@@ -298,11 +317,50 @@ export class ChatLlamaCpp extends SimpleChatModel<LlamaCppCallOptions> {
...
+          }
         } else if (message._getType() === "ai") {
```
giladgd, line 334: The next version will support the generation of a response based on chat histories that can have "system", "user" and "model" roles, so I think most of this code could be replaced by the `node-llama-cpp` functionality for this.

I predict that the next beta that includes this functionality will be ready in about ~2 weeks if you're willing to wait for it, but in any case, I think this better be handled by `node-llama-cpp` as it its responsibility to handle the interface against a model.

giladgd (changes requested):

langchain/src/chat_models/llama_cpp.ts
nigel-daniels: This is because the call for streaming uses the `context.evaluate()` method:

```
const stream = await this.caller.call(async () =>
      this._context.evaluate(this._context.encode(prompt), promptOptions)
    );
```

This returned a generator function where I can use `yield` to pass back the tokens as they arrive.

jacoblee93:
Thanks for the review @giladgd - @nigel-daniels should we wait a bit then?

nigel-daniels:
@jacoblee93 as I understood it this would only work if I was using the `session.prompt()` method but in order to get hold of a generator function I was calling the `context.evaluate()` and bypass the `session`. It's the session which applies the different wrapper types.
@giladgd do you have any thoughts on a better way to do this using the `session`?
Afaik If I use the `context` directly then there is no point in waiting.

nigel-daniels:
@giladgd looks like a good workaround, it will also ensure the Wrappers get used so this module can remain better aligned with `node-llama-cpp`. Thanks!

nigel-daniels:
Given the previous comments about the upcoming update in a couple of weeks and the restructure happening in the `Langchain.JS` main branch I'll close this PR and rebase on main when the 'node-llama-cpp' updates are out.

PR #3590, opened 2023-12-08
Adds delete method to PGVectorStore

MJDeligan:
### Features
This PR adds an implementation of the delete method for the PGVectorStore integration.

On top of the option to delete by ids, as defined in the Python API, this implementation allows the deletion of documents whose metadata matches a filter object (as defined by the postgres jsonb containment operator), which can passed as an alternative to the ids. This is in line with other VS integrations such as Pinecone (and is also a desired feature for my team), but I'm, of course, willing the to adjust.

### Additional Feats

This PR includes two simple tests to show that the two deletion methods (ids and metadata) work correctly.

The implementation was also tested with the recently added collections in mind.


MJDeligan:
Not sure if the tests fit the definition of integration tests as outlined by the contributing guidelines, since there is no need for real any real external service, but they do require a running local pgvector db.

jacoblee93:
Looks great, apologies for the delay!

PR #3591, opened 2023-12-08
Added async progres checker function in refineDocuments chain

faysal515:
Improvement made in the Document Refine Chain. This will help show the progress it has made on the refinement process.

### DIFF
--- a/langchain/src/chains/combine_docs_chain.ts
+++ b/langchain/src/chains/combine_docs_chain.ts
@@ -300,6 +300,7 @@ export interface RefineDocumentsChainInput extends StuffDocumentsChainInput {
   initialResponseName?: string;
   documentVariableName?: string;
   outputKey?: string;
+  progressEmitter?: (progress: number) => Promise<void>; // Optional progress emitter function
 }
 
 /**
@@ -326,6 +327,9 @@ export class RefineDocumentsChain
   initialResponseName = "existing_answer";
 
   refineLLMChain: LLMChain;
+  
+  progressEmitter?: (progress: number) => Promise<void>
+
 
   get defaultDocumentPrompt(): BasePromptTemplate {
     return new PromptTemplate({
@@ -364,6 +368,7 @@ export class RefineDocumentsChain
     this.documentPrompt = fields.documentPrompt ?? this.documentPrompt;
     this.initialResponseName =
       fields.initialResponseName ?? this.initialResponseName;
+    this.progressEmitter = fields.progressEmitter;
   }
 
   /** @ignore */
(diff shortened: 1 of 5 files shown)


langchain/src/chains/combine_docs_chain.ts
```diff
@@ -326,6 +327,9 @@ export class RefineDocumentsChain
   initialResponseName = "existing_answer";
 
   refineLLMChain: LLMChain;
+  
+  progressEmitter?: (progress: number) => Promise<void>
```
jacoblee93, line 331: We've called something similar `onDocumentLoaded` here:

https://github.com/langchain-ai/langchainjs/blob/06a39117aae61fcbf26a537de4bad8a156bbb9ab/langchain/src/document_loaders/web/notionapi.ts#L121

I think it's a bit more generalizable - would it make sense to rename to `onDocumentProcessed` or something similar?

faysal515:
@jacoblee93 Sure. Updating it

langchain/src/chains/combine_docs_chain.ts
faysal515: Updated in the new commit

faysal515:
@jacoblee93 docs & api_refs build is successful on my my machine. can you please share a screenshot of why vercel deployment is failing.

langchain/src/chains/combine_docs_chain.ts
```diff
@@ -439,6 +443,10 @@ export class RefineDocumentsChain
         runManager?.getChild("refine")
       );
       refineSteps.push(res);
+      if (this.onDocumentProcessed) {
+        const progress = (i / currentDocs.length).toFixed(2); // Progress calculation
+        await this.onDocumentProcessed(Number(progress)); // Emitting the progress
```
jacoblee93, line 448: Ah, was also suggesting that this emits the actual document in addition to just a fraction. User could calculate a `progress` meter on their own.

Let's actually hold off on this though - we're starting to look more at these older chains to see how we can modernize them more with LCEL principles.

faysal515: Right, instead of calculating and returning the progress, would make more sense to return the document itself,
and maybe the current loop number, It'd allow more freedom in the code side.

I understand that this is currently on hold to adapt and implement in LECL way, any tentative plan for when is this feature might be picked up? (I kind of need this for my project)

Thanks

jacoblee93: So sorry for missing this reply - could this work with the existing callback system by adding a `handleLLMEnd` callback to the model?

https://js.langchain.com/docs/modules/callbacks/how_to/create_handler

hntrl:
Closing due to inactivity

PR #3598, opened 2023-12-08
pass in authoptions correctly from initialization down to connection sdk

pixelcatgg:
We are using GCP Vertex AI and noticed that there was an attempt to make passing in credentials explicitly work, but there was some missing code to actually get the initialization to recognize the missing credentials. 

For example, initializing the SDK like so seems like it is possible from the types, as it is extending `GoogleAuthOptions`, but the initializer was never actually getting these arguments:

```
    const engine = new MatchingEngine(embeddings, {
      index: "11111",
      indexEndpoint: "1111",
      apiVersion: "v1beta1",
      endpoint: "us-central1-aiplatform.googleapis.com",
      location: "us-central1",
      docstore: store,
      authOptions: {
        scopes: ["https://www.googleapis.com/auth/cloud-platform"],
        projectId: "2222",
        credentials: {
          client_email: "test@gcloud.com",
          private_key: "privatekey",
        },
      },
    });
```

While doing `gcloud auth application-default login` works of course, it does not fit our use case as we are running on Serverless and containerized environments, where we cannot have a .json credentials file or run CLI commands to log in. We also cannot have Workload Federation, which is the preferred and safest method. So we have to pass in these values, which is possible using the native SDK. 

I have tested that this change works with the following:

1. Machine has no gcp credentials set, explicitly pass in authOptions
2. Machine has gcp credentials set, explicitly pass in authOptions (authOptions takes precedence)
3. Machine has gcp credentials set, omit authOptions, sdk will pull from machine env as is currently default

Feedback is welcome! Excited to contribute to Langchain :)


jacoblee93:
Oof, nice catch. Will merge and get pushed!

pixelcatgg:
> Oof, nice catch. Will merge and get pushed!

Thank you! Do you need me to fix merge conflicts @jacoblee93 ? Just noticed it now has some

jacoblee93:
Thanks!

PR #3599, opened 2023-12-08
Update AssemblyAI SDK

Swimburger:
- Update to the latest version of the AssemblyAI SDK which is compatible with multiple runtimes (node/web/workerd/bun/deno/etc.)
- Use latest SDK APIs, but provide backwards compatible overloads to not introduce breaking changes to users of the AssemblyAI LangChain integrations.


jacoblee93:
Thank you! Apologies for the delay!

PR #3606, opened 2023-12-10
feat: fix RunnableLambda typing when returning Runnables

dilling:
Currently the output type of RunnableLambda in not right in the case that you have a Lambda for a function that returns a Runnable, specifically like in [this example in the docs](https://js.langchain.com/docs/expression_language/how_to/routing#using-a-custom-function). 

Since the returned runnable is also invoked, the type `RunnableLambda<string, Runnable<string, number>` should extend `Runnable<string, string>`. In the current implementation, `RunnableLambda<string, Runnable<string, number>` extends `Runnable<string, Runnable<string, number>` instead.


langchain-core/src/runnables/base.ts
```diff
@@ -21,6 +21,11 @@ import { AsyncCaller } from "../utils/async_caller.js";
 import { Run } from "../tracers/base.js";
 import { RootListenersTracer } from "../tracers/root_listener.js";
 
+export type RunnableOutput<RunInput, FuncOutput>  = 
+  FuncOutput extends Runnable<RunInput, infer RunOutput> 
+    ? RunOutput
+    : FuncOutput;
+
```
dilling, line 28: Don't love the name of this, very open to suggestions

PR #3612, opened 2023-12-11
Update OpenAI embeddings `stripNewLines` to be default `false`

Knordy:
Updating the default value for `stripNewLines` to `false` in the `openai` embeddings. As removing newlines was beneficial for `V1` models (or `-001`), but should not be mandatory for `V2` models (or `-002`). This is explained in https://github.com/openai/openai-python/issues/418#issuecomment-1525939500

Therefor updating this field to be in line with the default model which is set to `text-embedding-ada-002`.

Also the langchain python library only enables stripping newlines for `-001` models: https://github.com/langchain-ai/langchain/blob/c0f4b95aa9961724ab4569049b4c3bc12ebbacfc/libs/langchain/langchain/embeddings/openai.py#L466


jacoblee93:
Ah that's regrettable but unfortunately we can't change this on a patch bump as it may have unexpected behavior for folks using it in production - can you change it to a note that we should update on the next minor version?

Knordy:
@jacoblee93 that's fine, however I'm not really sure what your request is. Am I correct that you mean to revert the changes, and add comments to change this in a next minor version?

jacoblee93:
Yes, something like `// TODO: Update to true on next minor release (see https://openai.com)`

Knordy:
@jacoblee93 I have reverted the default value, back to `false` again. Kept the update in the comment. Also added the note to update it on a next minor release. Referenced this PR, as it contains the information as to why this should be updated.

jacoblee93:
Thank you!

PR #3614, opened 2023-12-11
Rename LLMonitor Callback Handler

vincelwt:
Hi,

We are renaming our product from LLMonitor to Lunary.ai. 

This PR changes all references to LLMonitor throughout the code to Lunary.

(Tested and linted)

Thank you and have a great week!
Vince


jacoblee93:
This will break anyone who's currently using it - are we ok with that? We could also leave a path as `langchain/callbacks/llmonitor` as well?

jacoblee93:
Closing in favor of #3667

PR #3621, opened 2023-12-11
community[major]: Added integration with new Gemini API

alx13:
**Note**: library `@google/generative-ai` yet to be released. Raising this PR for early review.

**Description**: added Gemini models on `generativelanguage.googleapis.com`
**Twitter handle:** @alexalx13


jacoblee93:
Thanks @alx13! We're doing a bit of housekeeping with packages as you may have noticed - will have a look but this might be best in it's own package. I'll take care of any refactoring that needs to happen

alx13:
Thanks @jacoblee93 ! I tried to follow the new structure with `libs/langchain-community`.

One thing regarding backwards comparability is I changed the name of the environment variable for API key to be more consistent with Python langchain.

hsubox76 (approved):
LGTM, at least the part that interfaces with the `@google/generative-ai` API

afirstenberg:
Just to be clear - This isn't replacing the PaLM API chat model?

alx13:
@afirstenberg In this PR I'm not touching chat models, will do in the followups.

examples/src/llms/googlegenerativeai.ts
```diff
@@ -0,0 +1,67 @@
...
+  safetySettings: [
+    {
+      category: HarmCategory.HARM_CATEGORY_HARASSMENT,
+      threshold: HarmBlockThreshold.BLOCK_LOW_AND_ABOVE,
+    },
+  ],
+});
+const image = fs.readFileSync("./image.png").toString("base64");
+const image2 = fs.readFileSync("./image2.png").toString("base64");
+const input2 = [
```
jacoblee93, line 44: Should this be a chat model then?

libs/langchain-community/scripts/create-entrypoints.js
```diff
@@ -56,6 +57,7 @@ const entrypoints = {
   "llms/cloudflare_workersai": "llms/cloudflare_workersai",
   "llms/cohere": "llms/cohere",
   "llms/fireworks": "llms/fireworks",
+  "llms/googlegenerativeai": "llms/googlegenerativeai",
```
jacoblee93, line 60: Add to `requiresOptionalDependency` below as well

libs/langchain-community/src/embeddings/googlegenerativeai.ts
```diff
@@ -0,0 +1,215 @@
...
+    if (!this.apiKey) {
+      throw new Error(
+        "Please set an API key for Google GenerativeAI " +
+          "in the environmentb variable GOOGLE_API_KEY " +
+          "or in the `apiKey` field of the " +
+          "GoogleGenerativeAIEmbeddings constructor"
+      );
+    }
+
+    this.client = isEmbedContentModel
```
jacoblee93, line 132: Is the difference here multimodal vs. text embeddings?

libs/langchain-community/src/embeddings/googlegenerativeai.ts
```diff
@@ -0,0 +1,215 @@
...
+  /**
+   * Method that takes a document as input and returns a promise that
+   * resolves to an embedding for the document. It calls the _embedText
+   * method with the document as the input.
+   * @param document Document for which to generate an embedding.
+   * @returns Promise that resolves to an embedding for the input document.
+   */
+  embedQuery(document: string): Promise<number[]> {
+    if (this._isEmbedContentModel) {
+      return this.caller.callWithOptions(
```
jacoblee93, line 183: I think can just be `this.caller.call`?

libs/langchain-community/src/embeddings/googlegenerativeai.ts
```diff
@@ -0,0 +1,215 @@
...
+ * console.log({ documentRes });
+ * ```
+ */
+export class GoogleGenerativeAIEmbeddings
+  extends Embeddings
+  implements GoogleGenerativeAIEmbeddingsParams
+{
+  apiKey?: string;
+
+  modelName = "embedding-001";
```
jacoblee93, line 80: We should ideally keep defaults the same for backwards-compatibility

libs/langchain-community/src/llms/googlegenerativeai.ts
```diff
@@ -0,0 +1,380 @@
...
+  extends LLM
+  implements GoogleGenerativeAITextInput
+{
+  get lc_secrets(): { [key: string]: string } | undefined {
+    return {
+      apiKey: "GOOGLE_API_KEY",
+    };
+  }
+
+  modelName = "gemini-pro";
```
jacoblee93, line 118: See above, the other alternative is we just keep the deprecated old code and remove all references to it in the docs

libs/langchain-community/src/llms/googlegenerativeai.ts
```diff
@@ -0,0 +1,380 @@
...
+ */
+export class GoogleGenerativeAI
+  extends LLM
+  implements GoogleGenerativeAITextInput
+{
+  get lc_secrets(): { [key: string]: string } | undefined {
+    return {
+      apiKey: "GOOGLE_API_KEY",
+    };
+  }
```
jacoblee93, line 116: Should set `static lc_name` too for minified environments (though I think this won't be supported anyway)

libs/langchain-community/src/llms/googlegenerativeai.ts
```diff
@@ -0,0 +1,380 @@
...
+
+  async invoke(
+    input: BaseLanguageModelInput,
+    options?: this["ParsedCallOptions"]
+  ) {
+    if (this._isGenerateContentModel) {
+      return super.invoke(input, options);
+    }
+    let prompt = this._convertInputToGenerateContent(input);
```
jacoblee93, line 282: Do we need to override `invoke` or can we just override `_call`? Or is this because we want to support message inputs?

It's a bit odd to specifically add support for messages into an LLM as a recommended flow (though it's of course possible) - I think my preference would be to keep LLMs text in/text out and then add chat model support for this.

jacoblee93 (commented):
Thanks! Some high-level feedback. Excited to try this out!

examples/src/llms/googlegenerativeai.ts
alx13: From the API perspective it's the same handler which serves single-turn and multi-turn conversations.

It's more of a question should LLM support multi-modal inputs?

I just wanted to be consistent with Python counterpart: https://github.com/langchain-ai/langchain/pull/14557

libs/langchain-community/src/embeddings/googlegenerativeai.ts
```diff
@@ -0,0 +1,215 @@
...
+
+  taskType?: TaskType;
+  title?: string;
+  stripNewLines: boolean = true;
```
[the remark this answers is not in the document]
alx13: Already did, no changes :)

libs/langchain-community/src/embeddings/googlegenerativeai.ts
alx13: There are different methods with different arguments for different models:
for `embedding-001` it's  `embedContent` and for `embedding-gecko-002` it's `embedText`

libs/langchain-community/scripts/create-entrypoints.js
alx13: Done

libs/langchain-community/src/llms/googlegenerativeai.ts
alx13: Reverted back old code

libs/langchain-community/src/embeddings/googlegenerativeai.ts
alx13: Fixed

examples/src/llms/googlegenerativeai.ts
alx13: @jacoblee93 Refactoring this as a chat model.

libs/langchain-community/src/llms/googlegenerativeai.ts
afirstenberg: (I warned y'all about this months ago, I just want to remind you. 🙃)

Despite the misnomer of L"L"M, I'd want the non-chat version to be multimodal if possible.

I think when I was looking at this, however, I was looking at overriding `predictMessages()` instead, anticipating future multimodal responses as well.

libs/langchain-community/src/chat_models/googlegenerativeai.ts
alx13, line 1: @jacoblee93 Could you please check this? I will fix llms to accept only text and add examples tomorrow.

jacoblee93:
There was some backchanneling in Slack - will open a new draft PR that uses this!

libs/langchain-community/src/chat_models/googlegenerativeai.ts
jacoblee93: Will open something akin to https://github.com/langchain-ai/langchain/pull/14621 and make sure it's ready for tomorrow 👍

libs/langchain-community/src/llms/googlegenerativeai.ts
```diff
@@ -0,0 +1,363 @@
...
+            candidateCount: 1,
+            stopSequences: this.stopSequences,
+            maxOutputTokens: this.maxOutputTokens,
+            temperature: this.temperature,
+            topP: this.topP,
+            topK: this.topK,
+          },
+        })
+      : new TextServiceClient({
+          authClient: new GoogleAuth().fromAPIKey(this.apiKey),
```
afirstenberg, line 225: Is the use of GoogleAuth only for backwards compatibility?

(If so - can we reconsider it? Removing that would make this module available in browser-based JavaScript.)

alx13: Yes, for now new SDK only support new models. Let me check if I can get rid of this module for Auth.

jacoblee93: I think two different entrypoints is actually better since people would otherwise need to install two peer dependencies instead of just one. At the very least, it's not backwards compatible with PaLM.

jacoblee93: I've opened #3641 for Google Generative AI - will Vertex require a different SDK? Will it be backwards compatible with the existing Vertex class?

If that's the case I think my preferred approach would be to update the existing vertex class.

alx13: Yeah, unfortunately Vertex will require a different SDK: https://github.com/googleapis/nodejs-vertexai/
And because currently in langchainjs we have bespoke implementation of VertexAI  it would be a little cumbersome to extend base class.

jacoblee93: Could we stick with the bespoke one or have things significantly changed? 

Concern would be that it'd be breaking for folks currently using it. What would your ideal import path be?

jacoblee93: IMO I think quickest path forward is to merge and ship #3641, then worry about Vertex afterwards. It sounds like it shouldn't be in the same package due to dependency web vs. Node compatibility issues anyway.

jacoblee93:
Thank you @alx13!

PR #3623, opened 2023-12-11
langchain-mistralai[major]: Add MistralAI chat and embed

bracesproul:
**WIP until someone gets mistral API access and can run integration tests**


docs/core_docs/docs/integrations/chat/mistral.mdx
```diff
@@ -0,0 +1,45 @@
+---
+sidebar_label: Mistral
+---
+
+import CodeBlock from "@theme/CodeBlock";
+
+# ChatMistralAI
+
+[OMistral AI](https://mistral.ai/) is a research and hosting platform for LLMs.
```
jacoblee93, line 9: Typo?

libs/langchain-mistralai/package.json
```diff
@@ -0,0 +1,78 @@
...
+    "node": ">=18"
+  },
+  "main": "./index.js",
+  "types": "./index.d.ts",
+  "repository": {
+    "type": "git",
+    "url": "git@github.com:langchain-ai/langchainjs.git"
+  },
+  "scripts": {
+    "build": "yarn clean && yarn build:esm && yarn build:cjs && yarn build:scripts",
```
jacoblee93, line 16: Should update with `deps` on core?

libs/langchain-mistralai/src/chat_models.ts
```diff
@@ -0,0 +1,410 @@
...
+    if (!apiKey) {
+      throw new Error(
+        "API key MISTRAL_API_KEY is missing for MistralAI, but it is required."
+      );
+    }
+    this.client = new MistralClient(apiKey, fields?.endpoint);
+  }
+
+  _llmType() {
+    return "chat_mistral_ai";
```
jacoblee93, line 198: Just `mistral_ai` is better

libs/langchain-mistralai/src/chat_models.ts
```diff
@@ -0,0 +1,410 @@
...
+        res = this.client.chatStream(input);
+      } else {
+        res = await this.client.chat(input);
+      }
+      return res;
+    });
+  }
+
+  /** @ignore */
+  async _generate(
```
jacoblee93, line 279: If you're extending `SimpleChatModel`, I think you only need to implement `_call`?

Otherwise, extend `BaseChatModel`

libs/langchain-mistralai/src/chat_models.ts
```diff
@@ -0,0 +1,410 @@
...
+
+  async completionWithRetry(
+    input: MistralAIChatCompletionOptions,
+    streaming: false
+  ): Promise<MistralAIChatCompletionResult>;
+
+  async completionWithRetry(
+    input: MistralAIChatCompletionOptions,
+    // Hack to satisfy TS return type
+    streaming: boolean = this.streaming
```
jacoblee93, line 260: I think we solve this in other ways elsewhere?

docs/core_docs/docs/integrations/chat/mistral.mdx
```diff
@@ -0,0 +1,45 @@
...
+They're most known for their family of 7B models ([`mistral7b` // `mistral-tiny`](https://mistral.ai/news/announcing-mistral-7b/), [`mixtral8x7b` // `mistral-small`](https://mistral.ai/news/mixtral-of-experts/)).
+
+The LangChain implementation of Mistral's models uses their hosted generation API, making it easier to access their models without needing to run them locally.
+
+## Models
+
+Mistral's API offers access to two of their models: `mistral7b` and `mixtral8x7b`. You can access these through the API by using their model names:
+
+- `mistral7b` -> `mistral-tiny`
+- `mixtral8x7b` -> `mistral-small`
```
jacoblee93, line 19: Hah

jacoblee93 (commented):
Looks good overall! See comments

libs/langchain-mistralai/package.json
```diff
@@ -0,0 +1,78 @@
...
+    "test:watch": "NODE_OPTIONS=--experimental-vm-modules jest --watch --testPathIgnorePatterns=\\.int\\.test.ts",
+    "test:single": "NODE_OPTIONS=--experimental-vm-modules yarn run jest --config jest.config.cjs --testTimeout 100000",
+    "test:int": "NODE_OPTIONS=--experimental-vm-modules jest --testPathPattern=\\.int\\.test.ts --testTimeout 100000 --maxWorkers=50%",
+    "format": "prettier --write \"src\"",
+    "format:check": "prettier --check \"src\""
+  },
+  "author": "LangChain",
+  "license": "MIT",
+  "dependencies": {
+    "@langchain/core": "~0.0.7",
```
jacoblee93, line 36: Bump to latest core

docs/core_docs/docs/integrations/chat/mistral.mdx
bracesproul: whoops, copypasta from Ollama docs, forgot to drop that O

libs/langchain-mistralai/src/chat_models.ts
bracesproul: The template defaults to extend `SimpleChatModel`, should I do Base instead? Not sure the tradeoffs

libs/langchain-mistralai/src/chat_models.ts
bracesproul: Outdated comment, removed. But I still need to pass in a `streaming` param because you can call `.stream()` without `this.streaming` being true.

libs/langchain-mistralai/src/chat_models.ts
bracesproul: Pushed commit to extend `BaseChatModel`. For some reason I got a type error saying `ChatMistralAI` was incompatible with runnables inside the `examples` file when it extended `SimpleChatModel`....

jacoblee93: `SimpleChatModel` is designed to be simpler in that you just have to return a string instead of instantiate a chat message etc.

libs/langchain-mistralai/src/chat_models.ts
jacoblee93: Elsewhere if you call `.stream()` it will override any `this.streaming` property set on the instance and return the most chunks possible (which is what users want)

examples/src/models/chat/chat_stream_mistralai.ts
```diff
@@ -0,0 +1,45 @@
...
+const model = new ChatMistralAI({
+  apiKey: process.env.MISTRAL_API_KEY,
+  modelName: "mistral-small",
+});
+const prompt = ChatPromptTemplate.fromMessages([
+  ["system", "You are a helpful assistant"],
+  ["human", "{input}"],
+]);
+const chain = prompt.pipe(model);
+const response = await chain.stream({
```
jacoblee93, line 13: Would change this to use an output parser

libs/langchain-mistralai/package.json
```diff
@@ -0,0 +1,79 @@
...
+    "test:single": "NODE_OPTIONS=--experimental-vm-modules yarn run jest --config jest.config.cjs --testTimeout 100000",
+    "test:int": "NODE_OPTIONS=--experimental-vm-modules jest --testPathPattern=\\.int\\.test.ts --testTimeout 100000 --maxWorkers=50%",
+    "format": "prettier --write \"src\"",
+    "format:check": "prettier --check \"src\""
+  },
+  "author": "LangChain",
+  "license": "MIT",
+  "dependencies": {
+    "@langchain/core": "~0.1.0",
+    "@mistralai/mistralai": "~0.0.7"
```
jacoblee93, line 38: May want to use a caret for them

libs/langchain-mistralai/src/chat_models.ts
```diff
@@ -0,0 +1,409 @@
...
+  >
+  extends BaseChatModel<CallOptions>
+  implements ChatMistralAIInput
+{
+  // Used for tracing, replace with the same name as your class
+  static lc_name() {
+    return "ChatMistralAI";
+  }
+
+  modelName = "mistral-tiny";
```
jacoblee93, line 168: Default to `small`?

libs/langchain-mistralai/src/chat_models.ts
```diff
@@ -0,0 +1,409 @@
...
+    messages: BaseMessage[],
+    options: this["ParsedCallOptions"],
+    runManager?: CallbackManagerForLLMRun
+  ): Promise<string> {
+    const chatResponse = await this._generate(messages, options, runManager);
+    const { content } = chatResponse.generations[0].message;
+    if (typeof content === "string") {
+      return content;
+    }
+    throw new Error(
```
jacoblee93, line 218: This looks unnecessary?

libs/langchain-mistralai/src/embeddings.ts
```diff
@@ -0,0 +1,143 @@
...
+  batchSize?: number;
+  /**
+   * Whether to strip new lines from the input text. This is recommended,
+   * but may not be suitable for all use cases.
+   * @default {true}
+   */
+  stripNewLines?: boolean;
+}
+
+export const chunkArray = <T>(arr: T[], chunkSize: number) =>
```
jacoblee93, line 44: Do we need this? I think it's in `utils/`

HenryHengZJ:
@bracesproul @jacoblee93 I think it's missing this line to pass the `streaming` from `fields`:
`this.streaming = fields?.streaming ?? false;`

bracesproul:
> HenryHengZJ

Just merged a fix, version `0.0.3`.
In the process of doing this I discovered a bug on Mistral's end, so streaming will likely not work until they patch that.
Ticket here: https://github.com/mistralai/client-js/issues/18

HenryHengZJ:
thanks for the update!

PR #3624, opened 2023-12-12
all[patch]: Ensure other subpackages are built before test/build

jacoblee93:


libs/langchain-anthropic/package.json
```diff
@@ -13,7 +13,8 @@
     "url": "git@github.com:langchain-ai/langchainjs.git"
   },
   "scripts": {
-    "build": "yarn clean && yarn build:esm && yarn build:cjs && yarn build:scripts",
+    "build": "yarn run build:deps && yarn clean && yarn build:esm && yarn build:cjs && yarn build:scripts",
```
bracesproul, line 16: Will this not cause a circular build?

PR #3625, opened 2023-12-12
Add AzureOpenAI and AzureChatOpenAI classes for Python interop

dqbd:


jacoblee93:
Going to wait until next week as discussed 👍

PR #3627, opened 2023-12-12
community[major]: Add Together AI LLM integration 

bracesproul:
Created entry point inside `@langchain/community`. It is _**not**_ exported from `langchain`.


jacoblee93:
Looks good to me! Thank you!

Issue #3157, opened 2023-11-05
Unable to insert vector embeddings to postgres using langchain and pgvector


Shbhom:
### Issue Title:
Uploading Vector Embeddings to PostgreSQL Results in "Error inserting: invalid byte sequence for encoding 'UTF8': 0x00"

### Issue Description:
I have implemented an endpoint that allows users to upload text files, and then I create documents from those text files. These documents can be uploaded to either Pinecone or PostgreSQL databases. While the Pinecone upload works fine, I encounter an issue when trying to upload vector embeddings to PostgreSQL.

### Steps to Reproduce:
1. Create an endpoint that accepts text files using multer middleware.
2. Create document objects from the text files.
3. Attempt to upload these documents to a PostgreSQL database using the following configuration:

```ts
export async function uploader(req: Request, res: Response, next: NextFunction) {
    try {
        const { to } = req.query
        if (!to || typeof to !== "string") {
            throw new ErrorHandler("invalid queryParam or type eror", 400)
        }
        const txt = req.files
        if (!txt || !Array.isArray(txt)) {
        //allowed uploading multiple files so multer will return an array
            throw new ErrorHandler("no pdf file found", 400)
        }
        let loader
        let docs: Document[] = []
        for (let i = 0; i < txt.length; i++) {
            const element = txt[i];
            const data = await fs.readFile(path.resolve(element.path), { encoding: "utf-8" })
            const doc = new Document({ pageContent: data, metadata: { source: element.filename } })
            docs.push(doc)
        }
        console.log(docs)
        if (to === "pinecone") {
            const index = pine.Index(process.env.PINE_INDEX!)
            await PineconeStore.fromDocuments(docs, embeddings, { pineconeIndex: index, }).then(() => {
                console.log(`total docs uploaded : ${docs.length}`)
            })
            await clearDir("./uploads")
            return res.status(200).json({
                message: "docs uploaded succesfully"
            })

        } else if (to === "postgres") {
            const config = {
                postgresConnectionOptions: {
                    type: "postgres",
                    host: "127.0.0.1",
                    port: 5432,
                    user: "root",
                    password: "root",
                    database: "vectorstore",
                } as PoolConfig,
                tableName: "Docs",
                columns: {
                    idColumnName: "id",
                    vectorColumnName: "vector",
                    contentColumnName: "content",
                    metadataColumnName: "metadata",
                },
            };

[...]

PR #3628, opened 2023-12-12
community[patch]: removing null chars

phof:
I've also faced the issue documented in #3157; this resolves it by simply replacing null chars in `addVectors`.


jacoblee93:
Thank you!

Issue #2036, opened 2023-07-21
Documents are not saved in Opensearch when using textsplitters


nirajmchauhan:
I am using `CheerioWebBaseLoader` to load the documents and save them in the AWS Opensearch. What I realised is that the documents are not getting saved and there is no error as well. But when I remove the `TextSplitter` then the document is getting saved. Here is my code:

```ts
const loader = new CheerioWebBaseLoader(link);

const rawDocs = (await loader.load()).map((d) => {
  return new Document({
    metadata: { originalSource: link },
    pageContent: d.pageContent,
  });
});

const textSplitter = new RecursiveCharacterTextSplitter({
  chunkSize: 100,
  chunkOverlap: 10,
});

const docs = await textSplitter.splitDocuments(rawDocs);

await OpenSearchVectorStore.fromDocuments(
  docs,
  new OpenAIEmbeddings(),
  {
    client,
    indexName: "documents",
  }
);
```

I cannot go ahead without using `TextSplitter` as if the document is huge, then OpenAI gives 400 errors. Is there a solution to this?

nirajmchauhan:
Hey 🤖
I changed the chunk & overlap size as you suggested, still the documents are not getting saved.
```ts
const textSplitter = new RecursiveCharacterTextSplitter({
  chunkSize: 1000,
  chunkOverlap: 200,
});
```

jacoblee93:
The issue seems to be that OpenSearch can't handle certain types of metadata? 

I think the bot's response sounds reasonable here - adding a filter step to the vector store itself would be nice too if you feel up for opening a PR?

nirajmchauhan:
I ended up fixing this by adding indexing for `{metadata:{loc:{lines:{start: Int, end: Int}}}}`

kiraraty:
> I ended up fixing this by adding indexing for `{metadata:{loc:{lines:{start: Int, end: Int}}}}`

What's Int? please

nirajmchauhan:
> 

Its Integer

zhjch05:
@nirajmchauhan Would you mind share how did you add indexing? Appreciate the help! Facing same issue. Is there a way to add a general dynamic mapping for metadata that automatically handles this?

rushikeshRL:
@zhjch05 
You can update the mapping from Opensearch dashboard:
![image](https://github.com/langchain-ai/langchainjs/assets/115528161/6291233d-83d0-4f10-98ce-176d7b60b4c1)

[...]

Issue #1589, opened 2023-06-08
Opensearch store does not support object in metadata and fail silently


yeouchien:
Opensearch bulk indexing is failing silently due to default index mapping

https://github.com/hwchase17/langchainjs/blob/61e89dbaec07b4b50a55efe0f74faec5b2e701fd/langchain/src/vectorstores/opensearch.ts#L177-L186

### Code to repro:
```
  const loader = new PDFLoader('/tmp/pdfloader.pdf');
  const docs = await loader.loadAndSplit(new RecursiveCharacterTextSplitter());
  await OpenSearchVectorStore.fromDocuments(docs, new OpenAIEmbeddings(), {
    client
  });
```
code above will create the index but unable to index the documents because of the metadata created from `RecursiveCharacterTextSplitter` 
`metadata: { source: '/tmp/pdfloader.pdf', pdf: [Object], loc: [Object] }` 

### Workaround for this issue:
delete the object in metadata
```
  docs.forEach((d) => {
    d.metadata.source = basename(d.metadata.source);
    delete d.metadata.pdf;
    delete d.metadata.loc;
  });
```

### Potential solution:
1. do not support object in metadata, like pinecone
https://github.com/hwchase17/langchainjs/blob/61e89dbaec07b4b50a55efe0f74faec5b2e701fd/langchain/src/vectorstores/pinecone.ts#L59
2. throw appropriate error so that client know what's wrong.
https://github.com/opensearch-project/opensearch-js/blob/main/guides/bulk.md#handling-errors

raymondfeng:
Yes, we ran into the same issue with github document loader.

PR #3629, opened 2023-12-12
community[patch]: Update ElasticSearch mappings to successfully add documents from TextSplitter

### LINKED ISSUE #2857: Elastic Search Vector Store - Unable to Add Documents
Unable to add documents to Elastic Search Vector Store.

```typescript
  const loader = new DirectoryLoader("./documents", {
    ".txt": (path) => new TextLoader(path),
  });
  const documents = await loader.load()
  const text_splitter = new CharacterTextSplitter({chunkSize:1000, chunkOverlap:0})
  const docs = await text_splitter.splitDocuments(documents)
  console.log("Docs found...."+docs.length);
  const vectorStore = new ElasticVectorSearch(embeddings, clientArgs);
  const ids = await vectorStore.addDocuments(docs);
 
 // the following statement returns IDs but the insertion fails.
  console.log("Docs loaded...."+ids.length);
``` 

There is no error logged and the `ids` array is non-null, but the underlying vector store has no documents.

[Full Code](https://gist.github.com/vinodkiran/d299f8f0137dc52f6bee5c5782953e84)

mattraibert:
Fixes #2857

Explicitly map `.metadata.loc` as an object in Elasticsearch. `.metadata.*` was mapped to only allow keywords. So Elasticsearch would reject the object inserted at `.metadata.loc` by a `TextSplitter`.

Also throw an error when failing to insert into Elasticsearch.

The same solution could be applied to #1589 and #2036

https://github.com/langchain-ai/langchainjs/blob/2527979ccb1626ab993687810a805147cc24c619/libs/langchain-community/src/vectorstores/opensearch.ts#L254-L279


libs/langchain-community/src/vectorstores/elasticsearch.ts
```diff
@@ -266,16 +270,23 @@ export class ElasticVectorSearch extends VectorStore {
       mappings: {
         dynamic_templates: [
           {
-            // map all metadata properties to be keyword
-            "metadata.*": {
+            // map all metadata properties to be keyword except loc
+            metadata_except_loc: {
```
jacoblee93, line 274: Is there we can generalize this a bit more to handle all object properties?

mattraibert: Thanks for reviewing!

One reason I did it this way is that I tried to be as focused as possible on fixing the bug and leaving other existing behavior alone.

We could possibly use object type for every metadata field and perhaps it's the right thing to do given that `metadata` is defined as `Record<string, any>`. I think it would be like this:

```js
            "metadata.*": {
              match_mapping_type: "*",
              match: "metadata.*",
              mapping: { type: "object" },
            },
```

But I think there are a couple of reasonable arguments for keeping `metadata` fields declared as `keyword` in Elasticsearch. Searching through the codebase, I see that `loc` is the only field that is ever anything but a flat number or string. And I think being able to filter by `keyword` type fields is much simpler and more performant in Elasticsearch. For my use case, for example, I want to augment an embedding based search with keyword filtering, so I'd prefer to keep them as keywords.

Another solution to #2857 could be to change the way `loc` is stored so that it can be stored in a keyword field. The `TextSplitter`s could store something like `line_from` and `line_to`. In that case, it might make sense to update the type of `metadata` as well. This might be a bigger lift since it would potentially affect anything downstream of those `TextSplitter`s.

jacoblee93: Yeah we can't change text splitter metadata yet, though I think flattening it would be nice when we're ready to make a breaking change.

jacoblee93: I'm ok with this but can we add a TODO in the code as well? Also remember to run `yarn lint`!

jacoblee93:
Thanks!

jacoblee93:
Thank you!

PR #3636, opened 2023-12-13
community[minor]: Jacob/vectara summarization

jacoblee93:
cc @ofermend


ofermend:
Tested locally and all tests passed!
Thanks @jacoblee93 - LGTM.

PR #3665, opened 2023-12-14
langchain[patch]: Fix invert runId and threadId

### LINKED ISSUE #3664: Issue with Argument Order in submitToolOutputs Invocation
### Description
There seems to be a discrepancy in the order of arguments passed to the `submitToolOutputs` function in the `invoke` method of `experimental/openai_assistant/index.js`. Specifically, the order of `threadId` and `runId` is reversed in the last case of the `invoke` method.

### Expected Behavior
The `submitToolOutputs` function should receive arguments in the following order:
1. `threadId`
2. `runId`
3. `body`

### Actual Behavior
In the last `else` block of the `invoke` method, the `submitToolOutputs` function is called with `runId` first, followed by `threadId`, which is inconsistent with the expected argument order.

### Relevant Code Snippet
```javascript
// ... [other code in the invoke method] ...

else {
    // Submitting tool outputs to an existing run, outside the AgentExecutor
    // framework.
    run = await this.client.beta.threads.runs.submitToolOutputs(input.runId, input.threadId, {
        tool_outputs: input.toolOutputs,
    });
}

// ... [rest of the invoke method] ...
```

### Suggested Fix
Swap the positions of `input.runId` and `input.threadId` in the problematic call to align with the expected argument order.

shadyshoha:
Fix invert runId and threadId to match openai.beta.threads.runs.submitToolOutputs arguments

Fixes #3664

### DIFF
--- a/langchain/src/experimental/openai_assistant/index.ts
+++ b/langchain/src/experimental/openai_assistant/index.ts
@@ -135,8 +135,8 @@ export class OpenAIAssistantRunnable<
       // Submitting tool outputs to an existing run, outside the AgentExecutor
       // framework.
       run = await this.client.beta.threads.runs.submitToolOutputs(
-        input.runId,
         input.threadId,
+        input.runId,
         {
           tool_outputs: input.toolOutputs,
         }


jacoblee93:
Oh yikes 😬 thanks for the fix!

PR #3682, opened 2023-12-15
[WIP]core[major]: Update multi modal prompt template support

bracesproul:


langchain-core/src/prompts/chat.ts
```diff
@@ -307,6 +310,274 @@ export class ChatMessagePromptTemplate<
...
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  text?: string | Record<string, any>;
+}
+
+interface _ImageTemplateParam {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  image_url?: string | Record<string, any>;
+}
+
+class _StringImageMessagePromptTemplate<
```
jacoblee93, line 323: Why are these prefixed with an underscore?

langchain-core/src/prompts/chat.ts
```diff
@@ -307,6 +310,274 @@ export class ChatMessagePromptTemplate<
...
+        typeof item === "string" ||
+        (typeof item === "object" && "text" in item)
+      ) {
+        let text = "";
+        if (typeof item === "string") {
+          text = item;
+        } else if (typeof item.text === "string") {
+          text = item.text ?? "";
+        } else {
+          // handle
```
jacoblee93, line 458: Should do something here?

langchain-core/src/prompts/chat.ts
bracesproul: Copying python, cc @hinthornw

langchain-core/src/prompts/chat.ts
```diff
@@ -307,6 +310,274 @@ export class ChatMessagePromptTemplate<
...
+            inputVariables
+          });
+        } else {
+          throw new Error("Invalid image template");
+        }
+        prompt.push(imgTemplateObject);
+      }
+    }
+    // @ts-expect-error todo fix
+    return new this({ prompt, additionalOptions });
```
jacoblee93, line 507: The logic for the typing here is very tricky, we haven't figured out anything good for `ChatPromptTemplate.fromMessages()` yet which this looks similar to.

I would just use `any` for now.

langchain-core/src/prompts/chat.ts
```diff
@@ -307,6 +310,274 @@ export class ChatMessagePromptTemplate<
...
+
+  _messageClass:
+    | typeof HumanMessage
+    | typeof AIMessage
+    | typeof SystemMessage
+    | undefined;
+
+  // ChatMessage contains role field, others don't.
+  // Because of this, we have a separate class property for ChatMessage.
+  _chatMessageClass: typeof ChatMessage | undefined;
```
jacoblee93, line 363: Where did this `_` pattern come from for instance vars?

Why don't we just make it `protected`?

langchain-core/src/prompts/chat.ts
```diff
@@ -307,6 +310,274 @@ export class ChatMessagePromptTemplate<
...
+        return new this._chatMessageClass({
+          content: text,
+          role: this.getRoleFromMessageClass(this._chatMessageClass.name)
+        });
+      } else if (this._messageClass) {
+        return new this._messageClass({
+          content: text
+        });
+      } else {
+        throw new Error("_messageClass and _chatMessageClass are undefined.");
```
jacoblee93, line 526: When would we ever get into this state?

langchain-core/src/prompts/image.ts
```diff
@@ -0,0 +1,186 @@
...
+
+    if (this.validateTemplate) {
+      let totalInputVariables: string[] = this.inputVariables;
+      if (this.partialVariables) {
+        totalInputVariables = totalInputVariables.concat(
+          Object.keys(this.partialVariables)
+        );
+      }
+      /** @TODO Fix this */
+      // checkValidTemplate(
```
jacoblee93, line 73: So fix this 😛

langchain-core/src/prompts/image.ts
```diff
@@ -0,0 +1,186 @@
...
+    }
+    const output: ImageURL = { url };
+    if (detail) {
+      output.detail = detail;
+    }
+    return output as FormatOutput;
+  }
+
+  /** @ignore */
+  static async imports(): Promise<{
```
jacoblee93, line 161: We don't do this anymore?

langchain-core/src/utils/image.ts
```diff
@@ -0,0 +1,12 @@
+import * as fs from "node:fs/promises";
```
jacoblee93, line 1: We definitely want this to work in web envs too

langchain-core/src/prompts/serde.ts
```diff
@@ -13,7 +14,7 @@ export type SerializedPromptTemplate = {
   _type?: "prompt";
   input_variables: string[];
   template_format?: TemplateFormat;
-  template?: string;
+  template?: MessageContent;
```
jacoblee93, line 17: Is this necessary?

If it's breaking, we may as well deprecate the whole thing

langchain-core/src/prompts/image.ts
```diff
@@ -0,0 +1,186 @@
...
+  }
+
+  /**
+   * Formats the prompt template with the provided values.
+   * @param values The values to be used to format the prompt template.
+   * @returns A promise that resolves to a string which is the formatted prompt.
+   */
+  async format<FormatOutput = ImageURL>(
+    values: TypedPromptInputValues<RunInput>
+  ): Promise<FormatOutput> {
```
jacoblee93, line 119: This can't be right? Or is `ImagePromptTemplate` just the image content part?

jacoblee93 (changes requested):

langchain-core/src/prompts/chat.ts
bracesproul: We need it to satisfy typescript. I pushed a refactor to make both these conditions a little cleaner

langchain-core/src/prompts/chat.ts
bracesproul: Can do, was copying python but `protected` works too

langchain-core/src/prompts/chat.ts
bracesproul: nope, forgot to delete

langchain-core/src/prompts/serde.ts
bracesproul: Yes to necessary, but I don't believe this would be breaking since `template` can be `string`

Or I'm missing context

langchain-core/src/utils/image.ts
bracesproul: I believe this still should work because we're conditionally importing this only if you pass in a file path.

Correct me if I'm wrong about how the imports work though

langchain-core/src/prompts/image.ts
bracesproul: What's the alternative to conditionally import?

langchain-core/src/prompts/image.ts
bracesproul: `ImagePromptTemplate` I don't believe is ever intended to be used by an end user, and format returning the `ImageURL` type is accurate since that's the type which matches the OpenAI sdk:

```typescript
export interface ImageURL {
  /** Specifies the detail level of the image. */
  detail?: "auto" | "low" | "high";

  /** Either a URL of the image or the base64 encoded image data. */
  url: string;
}
```

langchain-core/src/prompts/chat.ts
```diff
@@ -331,6 +334,233 @@ export class ChatMessagePromptTemplate<
...
+  }
+
+  createMessage(content: MessageContent) {
+    if (this.messageClass) {
+      return new this.messageClass({ content });
+    } else if (this.chatMessageClass) {
+      // Assuming ChatMessage constructor also takes a content argument
+      return new this.chatMessageClass({
+        content,
+        role: this.getRoleFromMessageClass(this.chatMessageClass.name),
```
jacoblee93, line 425: This will fail in minified environments :( you should be able to use `lc_name` instead?

jacoblee93 (approved):

PR #3684, opened 2023-12-16
core[patch]: Use an interface for runnables to allow more compatibility between core versions

jacoblee93:
CC @nfcampos @bracesproul 

~Is this actually a good idea?~ Yes, it is a good idea

### DIFF
--- a/langchain-core/src/runnables/tests/runnable_interface.test.ts
+++ b/langchain-core/src/runnables/tests/runnable_interface.test.ts
@@ -43,7 +43,7 @@ interface RunnableInterfaceV0<RunInput, RunOutput, CallOptions = any> {
   ): Promise<IterableReadableStreamV0<RunOutput>>;
 }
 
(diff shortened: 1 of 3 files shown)


langchain-core/src/runnables/tests/runnable_interface.test.ts
```diff
@@ -0,0 +1,206 @@
...
+    batchOptions?: RunnableBatchOptionsV0
+  ): Promise<(RunOutput | Error)[]>;
+
+  stream(
+    input: RunInput,
+    options?: Partial<CallOptions>
+  ): Promise<IterableReadableStreamV0<RunOutput>>;
+}
+
+export class IterableReadableStreamV0<T> extends ReadableStream<T> {
```
bracesproul, line 46: Why export?

jacoblee93: Yeah don't need to

PR #3687, opened 2023-12-16
Add missing max_tokens option to TogetherAI

skadefro:
Add support for setting max_tokens for TogetherAI


jacoblee93:
Nice catch, thank you!

PR #3690, opened 2023-12-17
docs[major]: Generate API refs for all packages

bracesproul:
Changes:
- `create-entrypoints.js` script inside `api_refs`. This file manages adding entry-points to the `typedoc.json` file for all packages.
- Update code block generation to account for all package name prefixes
- Updated name to be `LangChain.js`
- Dropped `src/` and `/libs` from names/paths when displayed in the site


docs/api_refs/package.json
```diff
@@ -5,10 +5,10 @@
   "scripts": {
     "dev": "next dev -p 3001",
     "typedoc": "npx typedoc --options typedoc.json",
-    "build:scripts": "yarn typedoc && node ./scripts/update-typedoc-css.js",
+    "build:scripts": "node ./scripts/create-entrypoints.js && yarn typedoc && node ./scripts/update-typedoc-css.js",
     "build": "yarn clean && yarn run build:deps && yarn build:scripts && next build",
     "build:vercel": "yarn clean && yarn run build:deps --force && yarn build:scripts && next build",
-    "build:deps": "yarn run turbo:command build --filter=@langchain/core --filter=@langchain/anthropic --filter=@langchain/openai --filter=@langchain/community --filter=langchain --concurrency=1",
+    "build:deps": "yarn run turbo:command build --filter=@langchain/core --filter=@langchain/anthropic --filter=@langchain/openai --filter=@langchain/community --filter=@langchain/mistralai --filter=@langchain/google-genai --filter=langchain --concurrency=1",
```
jacoblee93, line 11: Will we need to continue to add packages here?

docs/core_docs/package.json
```diff
@@ -7,7 +7,7 @@
     "start": "yarn build:typedoc && rimraf ./docs/api && NODE_OPTIONS=--max-old-space-size=7168 docusaurus start",
     "build": "yarn clean && yarn build:typedoc && rimraf ./build && NODE_OPTIONS=--max-old-space-size=7168 DOCUSAURUS_SSR_CONCURRENCY=4 docusaurus build",
     "build:vercel": "yarn clean && yarn build:typedoc:vercel && rimraf ./build && NODE_OPTIONS=--max-old-space-size=7168 DOCUSAURUS_SSR_CONCURRENCY=4 docusaurus build",
-    "build:typedoc": "yarn run turbo:command build --filter=api_refs",
+    "build:typedoc": "yarn workspace api_refs build",
```
jacoblee93, line 10: Not with `turbo`?

tsconfig.json
```diff
@@ -0,0 +1,42 @@
+{
```
jacoblee93, line 1: We want this at root?

jacoblee93:
Have a lot going in tonight already so let's wait a little bit longer for this?

docs/api_refs/package.json
bracesproul: Yes, it might make more sense to turn this into a script which builds all packages inside `./libs` (excluding the template repo), langchain, and langchain core.

docs/core_docs/package.json
bracesproul: correct, this was the root of our problems with building on vercel, but when I fixed that I only removed it from the `build:typedoc:vercel` script.

tsconfig.json
bracesproul: We need it for the typedoc, it also is a good practice to have one shared root tsconfig which the rest of our projects extend from, but that's a task for another PR

docs/api_refs/package.json
jacoblee93: Or do we need the filter at all if we want to build everything?

docs/core_docs/package.json
jacoblee93: Ah ok

jacoblee93:
Longer term let's add headers like on https://api.python.langchain.com/ but this is great for now

docs/api_refs/package.json
bracesproul: We don't want to waste time building things like examples, core_docs, and other pkgs which don't matter to api refs

Issue #3594, opened 2023-12-08
PGVectorStore: Allow insertion of rows with set ids


MJDeligan:
Currently the implementation of the PGVectorStore only allows setting the page content and metadata for documents when adding them (or vectors directly). The Python API additionally allows passing ids that are used when inserting the rows.

This feature is necessary for an implementation of the [Indexing API](https://python.langchain.com/docs/modules/data_connection/indexing) as the hashed ids would need to be set manually for the rows of the vector store. See also #3390 .

Issue #3390, opened 2023-11-23
Python Module's Indexing API Equivalent


MJDeligan:
Hi,

I was previously working with langchain's python module, where I used the [Indexing API](https://python.langchain.com/docs/modules/data_connection/indexing) to upsert vectors into a postgres database.

I'm now working on a Javascript project and noticed that this API does not exist or is at least named differently. Is there an equivalent in the js-library?

Cheers

MJDeligan:
The delete and add methods would enable the implementation of an indexing API à la python, but it's missing the storage of hashed documents in a separate DB. I'm pretty sure the bot is confusing the index of vector stores with the 'Indexing' API. Can anyone confirm that this functionality is not yet implemented?

Leo310:
@MJDeligan how are you dealing with this problem now? Did you create your own indexing implementation?

MJDeligan:
> @MJDeligan how are you dealing with this problem now? Did you create your own indexing implementation?

@Leo310  Yeah, since we use PgVector, I first implemented a wrapper around the vector store to implement the delete method and changed the add_* methods to accept ids. Then I translated the indexes api, including the Record Manager. I'd do a PR, but I can't get the repo to install locally, so I'll have to look into that. If I find the time, I can post the implementations that I currently have.

Leo310:
@MJDeligan Ah ok nice. And I would really appreciate it if you could post the implementation or link me to it :)

Leo310:
@MJDeligan do you think you will share your implementation? If not, no problem, I just need to know, because then I will have to create my own

MJDeligan:
Hey @Leo310 , sorry for the delay! I put together a [gist](https://gist.github.com/MJDeligan/f01df6bd952a707fc20b8ed678782ac9) for reference. Beware that I just ripped this from the code base (with permission) and that it has not been used extensively, so I would definitely be careful. The HashedDocument, in our case, depends on an environment variable to get the namespace UUID, but you can also set a static one. Also the record manager uses the pg-format library, since I didn't wanna bother with writing templating functions to avoid SQL-Injections. 

Hope this helps and sorry for the messy code.

Leo310:
@MJDeligan awesome, thanks a lot 👍

MJDeligan:
Hey @jacoblee93 , are there currently any plans for this? I'd love to contribute.

jacoblee93:
[...]

PR #3692, opened 2023-12-17
community[minor]: Adds optional IDs parameter to PGVectorStore add-* methods

### LINKED ISSUE #3595: Bug: PGVectorStore error inserting documents when column names have uppercase letters
### Description
When the vector store is initialized with one of the column names containing uppercase letters, such as contentColumnName = pageContent, then adding documents to the store causes an error 👍 

` Error inserting: column "pageContent" of relation "testlangchain" does not exist`

This can be reproduced by simply creating a pgvectorstore where one of the columns contains uppercase lettering and trying to insert. The pgvector integration can also be used to demonstrate this.

### Cause
This is because the creation of the table and the insertion query for rows are incompatible:

The creation query uses \"\" to maintain uppercase letters in the column names:

```js
 await this.pool.query(`
      CREATE TABLE IF NOT EXISTS ${this.tableName} (
        "${this.idColumnName}" uuid NOT NULL DEFAULT uuid_generate_v4() PRIMARY KEY,
        "${this.contentColumnName}" text,
        "${this.metadataColumnName}" jsonb,
        "${this.vectorColumnName}" vector
      );
    `);
```

~~The insertion query however uses a parameterized query to set the column names dynamically:~~

The insertion query however just inserts the column names as is:

```js
private async buildInsertQuery(rows: (string | Record<string, unknown>)[][]) {
    let collectionId;
    if (this.collectionTableName) {
      collectionId = await this.getOrCreateCollection();
    }

    const columns = [
      this.contentColumnName,
      this.vectorColumnName,
      this.metadataColumnName,
    ];

    if (collectionId) {
      columns.push("collection_id");
    }

    const valuesPlaceholders = rows
      .map((_, j) => this.generatePlaceholderForRowAt(j, columns.length))
      .join(", ");

    const text = `
      INSERT INTO ${this.tableName}(
        ${columns}
      )
      VALUES ${valuesPlaceholders}
    `;
    return text;
  }
```

This does not preserve the uppercasing. ~~**Note that node-pg explicitly states that [identifiers are not supported](https://node-postgres.com/features/queries#parameterized-query) for parametrization.**~~

Options
1. Create two separate queries: one with the collection\_id and one without, both with column names written out in \"\" and execute only one of them 
2. The column names are put in \"\" , dynamically joined and inserted into the string (not as parameter)
3. Consider the [pg-format](https://www.npmjs.com/package/pg-format) library, which can also handle identifiers. This library would also be good for many of the other queries, since it significantly reduces the effort required to dynamically create insert queries, because it can also handle arrays and objects, eliminating the need to dynamically generate placeholders.

MJDeligan:
Related Issue: #3594 

## Features

Adds an optional "ids" parameter to all "add-*" methods of the PGVectorStore, which allows users to provide custom ids for the inserted records. While this is not important for most ordinary uses of the VectorStore it is crucial for some APIs, such as the [Indexing API](https://python.langchain.com/docs/modules/data_connection/indexing). See also #3390 .

The implementation is consistent with the Python API, although I added a check that the number of documents and ids are equal. Ultimately, the behavior is consistent, since in both cases providing a different number of documents and ids leads to an error, but catching it early lets us throw a more informative error.

## Additional Features

A test for adding documents with given ids was added.

This also partially fixes #3595 , since the column names in the insert statement are now in double quotes, meaning column names can contain uppercase letters and not cause errors during insertion.

## Comments

The way we check for whether the id column should be part of the insert query is not ideal, since it can easily lead to programming errors during future extensions. Alternatively, we could always set the id column and either use the provided one or generate one for each document before insertion.


bracesproul (commented):
Thanks for adding this! I pushed up one commit resolving some merge conflicts, everything looks good now.

My one thought is would the `ids` parameter be better suited living in an optional `options` object? Might be overkill but we should consider before merging. cc @jacoblee93 

If we decide not, then it's good to merge.

jacoblee93:
Yes, would prefer an object param for consistency.

For Pinecone, we did it this way:

https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-community/src/vectorstores/pinecone.ts#L87

But would prefer `options?: { ids: string[] }`.

jacoblee93:
Updated - we can add shorthand later if you really feel strongly but would prefer to keep consistency as much as possible.

Thank you for the great PR!

MJDeligan:
> Updated - we can add shorthand later if you really feel strongly but would prefer to keep consistency as much as possible.

Thank you! I don't think a shorthand will be necessary.

PR #3700, opened 2023-12-17
RunnableWithMessageHistory now calls to getMessages

### LINKED ISSUE #3698: Defining a Custom History Leads To Unexpected Behaviour
This is a semantic bug in langchain core.

**Recreating the Issue:**
Say you create a custom history class to interact with your own db/etc, as follows, where the appropriate functions may be DB calls from another part of your application:
```typescript
export class ChatAgentHistory extends BaseChatMessageHistory{
    lc_namespace = ["ChatAgentHistory"];

    private sessionId: string;
    constructor(sessionId: string){
        super();
        this.sessionId = sessionId;
    }

    async getMessages(): Promise<BaseMessage[]>{
        return await getMessages(this.sessionId);
    }

    async addMessage(message: BaseMessage) {
        await createMessage(this.sessionId, message);
    }
    async addUserMessage(message: string): Promise<void> {
        await createUserMessage(this.sessionId, message);
    }
    async addAIChatMessage(message: string): Promise<void> {
        await createAssistantMessage(this.sessionId, message);
    }
    clear(): Promise<void> {
        await clearMessages(this.sessionId);
    }
}
```
Given the interface provided by BaseChatMessageHistory, this should be a complete implementation (assuming the other functions work as expected).

If you use this similar to [this](https://js.langchain.com/docs/expression_language/how_to/with_history) tutorial, as follows:
```typescript
const getMessageHistory = (sessionId: string) => {
    return new ChatAgentHistory(sessionId);
};
const historyPrompt = new RunnableWithMessageHistory({
    runnable: prompt,
    getMessageHistory,
    inputMessagesKey: "agent_scratchpad",
    config,
});
```

You will encounter an error that "historyMessages is not iterable".

This is not expected behaviour.

**Why does this occur?**
In the [_enterHistory function](https://github.com/langchain-ai/langchainjs/blob/1794285a31b1430b0ced51a79188bd14dbbf0c54/langchain-core/src/runnables/history.ts#L132), the code is as follows:
```typescript
    const history = kwargs?.config?.configurable?.messageHistory;

    if (this.historyMessagesKey) {
      return history.messages;
    }

    const inputVal =
      input ||
      (this.inputMessagesKey ? input[this.inputMessagesKey] : undefined);
    const historyMessages = history ? history.messages : [];
    const returnType = [
      ...historyMessages,
      ...this._getInputMessages(inputVal),
    ];
    return returnType;
```
The issue here is that the [BaseChatMessageHistory abstract class](https://github.com/langchain-ai/langchainjs/blob/1794285a31b1430b0ced51a79188bd14dbbf0c54/langchain-core/src/chat_history.ts#L8) defines a getMessages() function, however, the RunnableWithMessageHistory expects an object of 'messages' instead. I would argue that this is poor semantics, as there is nothing in the BaseChatMessageHistory guaranteeing that this property exists.

~~I will create a PR that calls into the getMessages() function if it exists, otherwise defaulting to the current behaviour.~~**Edit:** This requiresa bit more thought to handle async functions.

Speuce:
**Description of Fix**
This PR calls into getMessages() if it is defined (and .messages is not), fixing the semantic error of relying solely on object properties to retrieve messages rather than calling into the getMessages function, which is guaranteed for BaseChatMessageHistory

Fixes #3698

### DIFF
--- a/langchain-core/src/runnables/history.ts
+++ b/langchain-core/src/runnables/history.ts
@@ -145,9 +145,10 @@ export class RunnableWithMessageHistory<
       (this.inputMessagesKey ? input[this.inputMessagesKey] : undefined);
     let historyMessages = [];
     if (history) {
-      historyMessages = history.getMessages
-        ? history.getMessages()
-        : history.messages;
+      historyMessages =
+        history.getMessages && !history.messages
+          ? history.getMessages()
+          : history.messages;
     }
     const returnType = [
       ...historyMessages,
--- a/langchain-core/src/runnables/history.ts
+++ b/langchain-core/src/runnables/history.ts
@@ -143,7 +143,10 @@ export class RunnableWithMessageHistory<
     const inputVal =
       input ||
       (this.inputMessagesKey ? input[this.inputMessagesKey] : undefined);
-    const historyMessages = history ? history.messages : [];
+    let historyMessages = [];
+    if (history) {
+      historyMessages = (history.getMessages) ? history.getMessages() : history.messages;
+    }
     const returnType = [
       ...historyMessages,
       ...this._getInputMessages(inputVal),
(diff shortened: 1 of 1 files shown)


Speuce:
NVM this wont work with async

PR #3702, opened 2023-12-18
Add YandexGPT embeddings models

tyumentsev4:
Introducing an ability to work with the YandexGPT chat model.

### DIFF
--- a/libs/langchain-community/src/embeddings/yandex.ts
+++ b/libs/langchain-community/src/embeddings/yandex.ts
@@ -1,0 +1,151 @@
+import { getEnvironmentVariable } from "@langchain/core/utils/env";
+import { Embeddings, EmbeddingsParams } from "@langchain/core/embeddings";
+
+const apiUrl =
+  "https://llm.api.cloud.yandex.net/foundationModels/v1/textEmbedding";
+
+export interface YandexGPTEmbeddingsParams extends EmbeddingsParams {
+  /** Model name to use. */
+  model?: string;
+
+  /** Model version to use. */
+  modelVersion?: string;
+
+  /** Model version to use. */
+
+  /** Model URI to use. */
+  modelURI?: string;
+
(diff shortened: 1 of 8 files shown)


libs/langchain-community/src/embeddings/yandex.ts
```diff
@@ -0,0 +1,151 @@
...
+  /**
+   * Yandex Cloud IAM token for service or user account
+   * with the `ai.languageModels.user` role.
+   */
+  iamToken?: string;
+}
+
+/**
+ * Class for generating embeddings using the YandexGPT Foundation models API. Extends the
+ * Embeddings class and implements GradientEmbeddingsParams and
```
bracesproul, line 37:
```suggestion
 * Embeddings class and implements YandexGPTEmbeddingsParams and
```
nit

libs/langchain-community/src/embeddings/yandex.ts
```diff
@@ -0,0 +1,151 @@
...
+    }
+  }
+
+  /**
+   * Method to generate embeddings for an array of documents.
+   * @param texts Array of documents to generate embeddings for.
+   * @returns Promise that resolves to a 2D array of embeddings for each document.
+   */
+  async embedDocuments(texts: string[]): Promise<number[][]> {
+    return this.caller.call(async () => {
```
bracesproul, line 94: Small nit, in other embeddings classes we wrap the calls with an `embeddingWithRetry` method. Not exactly necessary in this case, but could prove helpful in the future.
[Example](https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-mistralai/src/embeddings.ts#L133C3-L133C36)

bracesproul (commented):
A few nits, also it would be helpful to have an integration test here (`embeddings/tests/yandex.int.test.ts`)

tyumentsev4:
Check #3791

PR #3703, opened 2023-12-18
Update YandexGPT API

tyumentsev4:
This pull request proposes to stop using the deprecated YandexGPT API v1alpha. Go to use  [Foundation Models API](https://cloud.yandex.com/en/docs/yandexgpt/api-ref/v1/).

### DIFF
--- a/docs/core_docs/docs/integrations/chat/yandex.mdx
+++ b/docs/core_docs/docs/integrations/chat/yandex.mdx
@@ -17,6 +17,10 @@ Next, you have two authentication options:
 - [API key](https://cloud.yandex.com/en/docs/iam/operations/api-key/create)
   You can specify the key in a constructor parameter as `api_key` or in an environment variable `YC_API_KEY`.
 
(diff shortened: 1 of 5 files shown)


bracesproul (changes requested):
Thanks for these improvements! Unfortunately, pushing these changes would be breaking to current users. Instead, if you create a new integration package for Yandex you can push these updates.

Here's what I'd recommend:
- [Create a new integration package](https://js.langchain.com/docs/contributing#want-to-add-a-specific-integration) `@langchain/yandex` via `npx create-langchain-integration`
- Copy/paste the LLM, chat model, embeddings and any other yandex integrations into this package.
- Leave the existing code inside `langchain-community` as is, without making any changes.
- Mark the `langchain-community` exported classes/functions as deprecated via a JSDoc comment (`/** @deprecated */`), update the docs to say something along the lines of "the integration in community is deprecated, please use the `@langchain/yandex` integration." (I can write docs/deprecation notices if you create the new package!)

This way we can have these changes you're proposing here, without breaking any existing users!

Let me know if you have any questions about this, I know it's a lot.

tyumentsev4:
Check #3791

PR #3709, opened 2023-12-19
all[patch]: Fix typing across different core versions by using interfaces instead of abstract classes

jacoblee93:
Following on #3684, adding more interfaces for:

- BaseLanguageModel
- BaseRetriever
- Embeddings
- Document
- VectorStore
- VectorStoreRetriever
- BasePromptValue
- StringPromptValue
- ChatPromptValue

And making modules in LangChain and vectorstores and a few other modules in community take them as parameters.

The goal is to make typing less strict if projects find themselves with multiple versions of core. Should be invisible.

CC @nfcampos @bracesproul

### DIFF
--- a/langchain-core/src/language_models/base.ts
+++ b/langchain-core/src/language_models/base.ts
@@ -2,7 +2,7 @@ import type { Tiktoken, TiktokenModel } from "js-tiktoken/lite";
 
 import { type BaseCache, InMemoryCache } from "../caches.js";
 import {
-  type BasePromptValue,
+  type BasePromptValueInterface,
   StringPromptValue,
   ChatPromptValue,
 } from "../prompt_values.js";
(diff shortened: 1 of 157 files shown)


langchain-core/src/language_models/base.ts
```diff
@@ -239,6 +239,48 @@ export type BaseLanguageModelInput =
   | string
   | BaseMessageLike[];
 
+export interface BaseLanguageModelInterface<
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  RunOutput = any,
+  CallOptions extends BaseLanguageModelCallOptions = BaseLanguageModelCallOptions
+> extends RunnableInterface<BaseLanguageModelInput, RunOutput, CallOptions> {
+  CallOptions: CallOptions;
```
jacoblee93, line 247: @nfcampos this seems to be the equivalent of `declare CallOptions` in a class?

PR #3711, opened 2023-12-19
core[minor]: Move chunk array to core

bracesproul:
Currently, it's defined inside `@langchain/community`, `@langchain/openai`, and `@langchain/mistralai`

Moving it to core will allow for deleting the duplicated code.

I'll put up a PR after core releases this, refactoring the other packages.


jacoblee93:
Thanks!

PR #3714, opened 2023-12-19
Add `stop` option to Together AI

sunghyunzz:
The [Together AI API Reference](https://docs.together.ai/reference/inference) states that the `stop` body parameter is supported, but the current implementation does not support this option. So I just added it.


bracesproul (approved):
thanks for adding this!

PR #3717, opened 2023-12-19
all[major]: Better release workflow

bracesproul:
TODO:
go through the flow locally now that 2fa is enabled on my npm account

# New release flow.
Invoke it by calling `yarn release`
There are three parameters which can be passed, one required and two optional.
- Required: `--workspace <workspace name>`. eg: `--workspace @langchain/core` (always appended as the first flag when running `yarn release`)
- Optional: `--version <version number>` eg: `--version 1.0.8`. Defaults to adding one to the patch version.
- Optional: `--bump-deps` eg `--bump-deps` Will find all packages in the repo which depend on this workspace and checkout a new branch, update the dep version, run yarn install, commit & push to new branch.
- Optional: `--tag <tag>` eg `--tag beta` Add a tag to the NPM release

Full example: `yarn release @langchain/core --version 1.0.9 --bump-deps --tag beta`. 

## Steps taken in script
- Checkout `release` branch & push
- Run `build`, `lint`, & `test` inside the input workspace
- Build `langchain` (required to run export tests)
- Run export tests (`test:exports:docker`)
- Prompts the user in the terminal for their NPM OTP, this is then passed as a command line arg to `release-it` via `--npm.otp=<tag>`.
- Runs `release-it` inside input workspace (publishes new npm version, updates version in pkg json, commits & pushes to current branch --`release`).
- If `--bump-deps` was passed, it find all workspaces which depend on the input workspace, checks out a new branch, updates the dep version, runs `yarn install`, commits & pushes to new branch.

### Other
- Added a non integration test to each package which lacked one before (`yarn test` would fail if no tests were found)
- Updated/added a jest script in packages, was trying to import from a location it couldn't access.
- Updated template to ship `.release-it.json`, and `release-it` dev dep


libs/create-langchain-integration/template/package.json
```diff
@@ -22,7 +22,7 @@
     "lint:fix": "yarn lint --fix",
     "clean": "rm -rf dist/ && NODE_OPTIONS=--max-old-space-size=4096 node scripts/create-entrypoints.js pre",
     "prepack": "yarn build",
-    "release": "release-it --only-version --config .release-it.json",
+    "release": "release-it --ci --config .release-it.json",
```
jacoblee93, line 25: What's the change here for?

libs/langchain-anthropic/src/tests/anthropic.test.ts
```diff
@@ -0,0 +1,8 @@
+import { test, expect } from "@jest/globals";
+
+// This is required while there are no (non integration) tests
+// inside this package. Otherwise, jest will complain about
+// "No tests found, exiting with code 1"
```
jacoblee93, line 5: Is there a more elegant way to avoid the error?

release_workspace.js
```diff
@@ -0,0 +1,294 @@
...
+
+/**
+ * Increments the last numeric character in a version string by 1.
+ * If the last character is not numeric, it searches backwards
+ * to find the last numeric character to increment.
+ * 
+ * @param {string} version 
+ * @returns {string} The new version
+ */
+function bumpVersion(version) {
```
jacoblee93, line 46: I think `release-it` does this by default if configured correctly?

libs/langchain-anthropic/src/tests/anthropic.test.ts
nfcampos: Adding an actual test haha

libs/create-langchain-integration/template/scripts/jest-setup-after-env.js
```diff
@@ -0,0 +1,3 @@
+import { awaitAllCallbacks } from "@langchain/core/callbacks/promises";
```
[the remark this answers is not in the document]
bracesproul: It was throwing earlier for me because it was trying to import a file that didn't exist. file was likely copypasta from langchain, there it had access to the file via `import ... from "../src/callbacks/promises.js"` but once inside these integrations that import would fail.

release_workspace.js
jacoblee93: Oh, this is for bumping deps

libs/create-langchain-integration/template/package.json
bracesproul: Actually, i need to delete these scripts since the code now runs `release-it` and passes in args all inside a `spawn` in the script.

bracesproul: Ah but I drop `--only-version` since it now explicitly passes the version number in. And `--ci` just says say yes to all the questions it asks, eg `Publish to npm?`. As if it was running in an async ci environment

release_workspace.js
```diff
@@ -0,0 +1,294 @@
...
+  );
+
+  // For console log, get all workspaces which depend and filter out duplicates.
+  const allWhichDependOn = new Set([
+    ...allWorkspacesWhichDependOn,
+    ...allWorkspacesWhichDevDependOn,
+    ...allWorkspacesWhichPeerDependOn,
+  ].map(({ packageJSON }) => packageJSON.name));
+
+  if (allWhichDependOn.size !== 0) {
```
jacoblee93, line 164: We don't always want to do this though? Only if we ask to bump?

libs/langchain-anthropic/src/tests/anthropic.test.ts
bracesproul: will investigate

release_workspace.js
```diff
@@ -0,0 +1,294 @@
...
+  // Find the workspace package.json's.
+  const allWorkspaces = getAllWorkspaces();
+  const matchingWorkspace = allWorkspaces.find(({ packageJSON }) => packageJSON.name === options.workspace);
+  
+  if (!matchingWorkspace) {
+    throw new Error(`Could not find workspace ${options.workspace}`);
+  }
+
+  // Bump version by 1 or use the version passed in.
+  const newVersion = options.version ?? bumpVersion(matchingWorkspace.packageJSON.version);
```
jacoblee93, line 261: Let's try to use the default `release-it` menu?

release_workspace.js
bracesproul: yes, unless we want to pass a custom version to change to, in that case you need to specify the version directly. Granted, not sure if this ever comes up I just assumed it might. If not, lmk and I can revert the changes which prevents `release-it` from bumping the version on its own.

release_workspace.js
bracesproul: This only runs if you pass the `--bump-deps` flag

libs/langchain-anthropic/src/tests/anthropic.test.ts
jacoblee93: Yeah we can just import + instantiate a model or something

libs/create-langchain-integration/template/scripts/jest-setup-after-env.js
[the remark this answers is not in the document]
jacoblee93: Ah I thought I saw it even with the `core` import but may have been mistaken

PR #3719, opened 2023-12-19
examples[patch]: Fixes type error in vectara example

bracesproul:
Don't have enough context to know why the `summary` element is no longer returned, removed for now. cc @jacoblee93


jacoblee93:
Thanks

PR #3721, opened 2023-12-20
Fix deserialisation of additional_kwargs and tool_call_id

dqbd:
Currently, when deserialising, `additional_kwargs` are passed to constructor as `additionalKwargs`


jacoblee93:
Thank yoU!

PR #3731, opened 2023-12-20
all[minor]: Remove duplicated chunk arr code, import from core

bracesproul:


libs/create-langchain-integration/template/package.json
```diff
@@ -33,7 +33,7 @@
   "author": "LangChain",
   "license": "MIT",
   "dependencies": {
-    "@langchain/core": "~0.1.2"
+    "@langchain/core": "~0.1.3"
```
bracesproul, line 36: Thoughts on changing this to `latest` @jacoblee93 ?

jacoblee93: How about `0.1`? Would that resolve to latest minor version?

jacoblee93:
Looks good but we will also need to be a little bit careful and bump the core dependencies of all the associated packages at the same time - maybe best to just add to core first and then do that asynchronously as needed?

bracesproul:
> Looks good but we will also need to be a little bit careful and bump the core dependencies of all the associated packages at the same time - maybe best to just add to core first and then start using/removing in separate PRs?

No need, I moved to core a couple days ago so it's already been released there. Only now getting around to refactoring the other packages.

Issue #3685, opened 2023-12-16
docx load error


iccyuan:
When I try to pass docs = await loader.load();Appear TypeError: extractRawText is not a function
yarn dev is normal;
yarn start appears TypeError: extractRawText is not a function
I tried to modify it and it worked fine, mammoth 1.6.0
``` typescription
async function DocxLoaderImports() {
  try {
    const { default: mod } = await import("mammoth");
    const { extractRawText } = mod;
    return { extractRawText };
  } catch (e) {
    console.error(e);
    throw new Error(
      "Failed to load mammoth. Please install it with eg. `npm install mammoth`."
    );
  }
}
```
```typescription

async function DocxLoaderImports() {
  try {
    const { extractRawText } = await import("mammoth");
    return { extractRawText };
  } catch (e) {
    console.error(e);
    throw new Error(
      "Failed to load mammoth. Please install it with eg. `npm install mammoth`."
    );
  }
}
```

iccyuan:
Reconfirmed that the problem exists

jacoblee93:
Thanks for opening this!

mantrakp04:
same issue has resurfaced, i tried downgrading mammoth as well, currently on the latest release of langchainjs.

PR #3737, opened 2023-12-21
langchain[patch]: fix:docx loader load error

iccyuan:
Fixes # [3685](https://github.com/langchain-ai/langchainjs/issues/3685)


jacoblee93:
Thank you! Seems like it was a version pinning issue.

PR #3760, opened 2023-12-22
langchain[patch]: Allow for passing bytestore into multi vector retriever

bracesproul:


examples/src/retrievers/multi_vector_hypothetical.ts
```diff
@@ -12,6 +12,7 @@ import { InMemoryStore } from "langchain/storage/in_memory";
 import { TextLoader } from "langchain/document_loaders/fs/text";
 import { Document } from "langchain/document";
 import { JsonKeyOutputFunctionsParser } from "langchain/output_parsers";
+import { createDocumentStoreFromByteStore } from "langchain/storage/encoder_backed";
```
jacoblee93, line 15: The whole idea here was to never see this function again?

examples/src/retrievers/multi_vector_hypothetical.ts
```diff
@@ -89,9 +90,11 @@ const keyValuePairs: [string, Document][] = docs.map((originalDoc, i) => [
...
 ]);
 
-// The docstore to use to store the original chunks
-const docstore = new InMemoryStore();
-await docstore.mset(keyValuePairs);
+// The byteStore to use to store the original chunks
+const byteStore = new InMemoryStore<Uint8Array>();
+// Convert the byteStore to a docStore to use for retrieval
+const docStore = createDocumentStoreFromByteStore(byteStore);
+await docStore.mset(keyValuePairs);
```
jacoblee93, line 97: Examples should use `retriever.docstore.mset` or we could maybe add something like `retriever.addDocuments`

examples/src/retrievers/multi_vector_hypothetical.ts
bracesproul: Whoops, was unaware. Will fix.

examples/src/retrievers/multi_vector_hypothetical.ts
bracesproul: I like the `addDocuments` method, will implement

langchain/src/retrievers/multi_vector.ts
```diff
@@ -3,15 +3,20 @@ import {
...
 import { Document } from "../document.js";
+import { createDocumentStoreFromByteStore } from "../storage/encoder_backed.js";
 
 /**
  * Arguments for the MultiVectorRetriever class.
  */
 export interface MultiVectorRetrieverInput extends BaseRetrieverInput {
   vectorstore: VectorStoreInterface;
-  docstore: BaseStoreInterface<string, Document>;
+  /** @deprecated Prefer `baseStore`. */
```
jacoblee93, line 17: `byteStore`

langchain/src/retrievers/multi_vector.ts
```diff
@@ -55,12 +60,44 @@ export class MultiVectorRetriever extends BaseRetriever {
...
+        "byteStore and docstore are undefined. Please provide at least one."
+      );
+    }
     this.idKey = args.idKey ?? "doc_id";
     this.childK = args.childK;
     this.parentK = args.parentK;
   }
 
+  /**
+   * Add documents to the byte store.
```
jacoblee93, line 78: This should also add index them in the vectorstore too

If it doesn't make sense to do so, we can leave it as `retriever.docstore.add()` above.

PR #3780, opened 2023-12-23
community[minor]: feat: turbopuffer vector store

mattzcarey:
Basic vector store for Turbopuffer vector store https://turbopuffer.com/


libs/langchain-community/src/vectorstores/turbopuffer.ts
```diff
@@ -0,0 +1,213 @@
...
+    return {
+      headers: {
+        Authorization: `Bearer ${this.apiKey}`,
+        "Content-Type": "application/json",
+      },
+    };
+  }
+
+  async addVectors(
+    vectors: number[][],
```
sirupsen, line 91: It's also valid to pass a `number[]`, but this may be too uncommon in JS (versus e.g. Python, where it's common)—not strictly required for my sake to include though :)

libs/langchain-community/src/vectorstores/turbopuffer.ts
```diff
@@ -0,0 +1,213 @@
...
+        headers: this.getJsonHeader().headers,
+        body: JSON.stringify(data),
+      });
+    } catch (error) {
+      console.error("Error storing vectors:", error);
+      throw error;
+    }
+  }
+
+  async addDocuments(
```
sirupsen, line 136: It is great that Langchain already seems to encourage batch-adding.

libs/langchain-community/src/vectorstores/turbopuffer.ts
```diff
@@ -0,0 +1,213 @@
...
+        docIds,
+        vectors,
+        attributes,
+      };
+
+      await fetch(`${this.apiUrl}/vectors/${this.namespace}`, {
+        method: "POST",
+        headers: this.getJsonHeader().headers,
+        body: JSON.stringify(data),
+      });
```
sirupsen, line 129: I am not completely intimate with the `fetch()`, API, but presumably it won't throw a proper error on e.g. a `4xx` error code which could contain an error like dimensionality not being consistent across all the `number[][]` vectors. I am not familiar enough with Langchain to know what's standard in other vector storage backends.

libs/langchain-community/src/vectorstores/turbopuffer.ts
```diff
@@ -0,0 +1,213 @@
...
+    };
+
+    const response = await fetch(
+      `${this.apiUrl}/vectors/${this.namespace}/query`,
+      {
+        method: "POST",
+        headers: this.getJsonHeader().headers,
+        body: JSON.stringify(data),
+      }
+    );
```
sirupsen, line 171: Similar note on proper error handling.

sirupsen (commented):
Thank you so much for this Matt. ❤️‍🔥 

(For context, I'm co-founder of turbopuffer, so wanted to take a look—mostly for the turbopuffer side of this)

Is there any documentation where this will need to be added as well?

jacoblee93:
Thanks Simon! Yes, I was planning to get to docs today or tomorrow after testing.

libs/langchain-community/src/vectorstores/turbopuffer.ts
jacoblee93: Yes, we'll want to parse + check the response for sure.

libs/langchain-community/src/vectorstores/turbopuffer.ts
jacoblee93: We can also add concurrency/more intelligent batching insert like this:

https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-openai/src/embeddings.ts#L175

 Are there any default values that makes sense to you?

sirupsen: The bigger the batch the better. Somewhere around 3,000 seems to work for most dimensionalities.

bracesproul (dismissed):
Thanks for this! I pushed up a commit adding docs and fixing some issues with the API requests. Unfortunately I didn't have time to get metadata filtering working. If you could get that implemented and re-request my review that would be great!

libs/langchain-community/src/vectorstores/turbopuffer.ts
```diff
@@ -0,0 +1,280 @@
...
+
+      if (documents.length === 0) {
+        throw new Error("No documents provided");
+      }
+
+      const batchedVectors = chunkArray(vectors, this.batchSize);
+      const batchedDocuments = chunkArray(documents, this.batchSize);
+      const batchedIds = options?.ids
+        ? chunkArray(options.ids, this.batchSize)
+        : batchedDocuments.map((docs, index) =>
```
jacoblee93, line 152: We definitely don't want to assign ids numerically like this because they'll be overwritten on subsequent insert. Use `uuid` or even just don't specify them if Turbopuffer handles it?

libs/langchain-community/src/vectorstores/turbopuffer.ts
```diff
@@ -0,0 +1,280 @@
...
+
+  protected apiKey: string;
+
+  protected namespace = "default";
+
+  protected apiUrl = "https://api.turbopuffer.com/v1";
+
+  caller: AsyncCaller;
+
+  batchSize = 500;
```
jacoblee93, line 70: Below, @sirupsen recommended a default batch size of 3000

libs/langchain-community/src/vectorstores/turbopuffer.ts
```diff
@@ -0,0 +1,280 @@
...
+    return this.addVectors(vectors, documents, options);
+  }
+
+  async queryVectors(
+    query: number[],
+    k: number,
+    includeAttributes?: string[],
+    includeVector?: boolean,
+    // See https://Turbopuffer.com/docs/reference/query for more info
+    filter?: TurbopufferFilterType
```
jacoblee93, line 216: Should be `this["FilterType"]`

libs/langchain-community/src/vectorstores/turbopuffer.ts
```diff
@@ -0,0 +1,280 @@
...
+  ): Promise<TurbopufferQueryResult[]> {
+    const data = {
+      vector: query,
+      top_k: k,
+      distance_metric: this.distanceMetric,
+      filters: filter,
```
jacoblee93, line 222: We store document contents in the attributes field above as:
```
        const attributes = {
          pageContent: batchDocs.map((doc) => doc.pageContent),
          metadata: batchDocs.map((doc) => JSON.stringify(doc.metadata ?? {})),
        };
```

Could be why passing `filter` through directly isn't working? We could also store raw document content as something like `__lc_page_content` and make sure we handle it while querying and inserting.

libs/langchain-community/src/vectorstores/turbopuffer.ts
mattzcarey: Turbopuffer currently only supports 64 bit unsigned integer values. Using `v4` from `uuid` will give 128-bit values. Do you have a preference for how they are truncated @jacoblee93 ?

@sirupsen do you have any idea on timeline for 128 bit values or preference here also?

jacoblee93: Good to know! No preference, but if we can just let turbopuffer auto-assign, that seems like the best approach.

CC @bracesproul who's been taking over landing this.

sirupsen: @jacoblee93 @mattzcarey let me try to see how far I can get on 128 bit / UUID support tomorrow... I'll circle back then. Randomizing a 64 bit integer isn't terrible at these low counts, but still not collision numbers I'm comfortable with given the behaviour on a collision (overriding a previous value).

jacoblee93: Ok, @bracesproul let's get the rest in good shape and wait another day then 👍

sirupsen: Hey... give me till next week, I ran into generic hell in Rust shoving in the string type for ids, before we introduce proper schema support. I'll get it working, just needed more than today :/

jacoblee93: No problem at all! Just ping me here or in Slack

bracesproul: hey @sirupsen have you gotten around to 128 bit yet? No worries if not, just checking back in.

jacoblee93: Chatted with them externally, still in progress!

sirupsen: shipped monday!

sirupsen (approved):
Overall LGTM, haven't reviewed in a ton of details, but things like batch size are the biggest things I look for 👍🏻 Thanks everyone!

jacoblee93:
Thank you!

PR #3783, opened 2023-12-25
feat: `PromptTemplate` format 'handlebars'

davidfant:
This PR adds a new `TemplateFormat` `handlebars`. The Python repo has f-string and jinja2, but the TS repo currently only has f-string. Previously I used Guidance (<1.0), which has a handlebar-like syntax, and it helps create complex prompts with if/elses, loops, etc.


davidfant:
Side question: when developing locally and wanting to use this Langchain build in my other project, what's the best way to do that? I'm trying `yarn link` in the langchain repo but I'm getting `Unknown Syntax Error: Not enough positional arguments.`

jacoblee93:
Thanks for the PR! We're keeping the core prompt to just use `f-string`, but we could add this in `langchain` as a new option. I can handle the refactor.

bracesproul:
> Side question: when developing locally and wanting to use this Langchain build in my other project, what's the best way to do that? I'm trying `yarn link` in the langchain repo but I'm getting `Unknown Syntax Error: Not enough positional arguments.`

you can add a file path instead of an npm version, this is how I do it on my mac:
```json
    "langchain": "file:/Users/langchain",
```

davidfant:
> Thanks for the PR! We're keeping the core prompt to just use `f-string`, but we could add this in `langchain` as a new option. I can handle the refactor.

Curious, what is the rationale for when to add things in langchain vs langchain core?

jacoblee93:
Two things I think about:

1. Fundamental pieces like base classes/schema that other classes rely on.
2. Parity and compatibility with Python.

We're generally trying to be pretty conservative with it, as it's easy to move things into core, but would be much harder to remove them. The split only occurred a few weeks ago though.

We'll be writing more about it in the near future - all this is extremely recent!

Will do a bit more research but saw a few CVEs with Handlebars around code execution, so that would be a good reason to put it in the main LangChain package to start as well.

davidfant:
@jacoblee93 @bracesproul I want to use this Langchain fork in my app on prod. What's the best way to use this fork from within my own Docker container? This works locally for development, but not for deployment:

> > Side question: when developing locally and wanting to use this Langchain build in my other project, what's the best way to do that? I'm trying `yarn link` in the langchain repo but I'm getting `Unknown Syntax Error: Not enough positional arguments.`
> 
> you can add a file path instead of an npm version, this is how I do it on my mac:
> 
> ```json
>     "langchain": "file:/Users/langchain",
> ```

jacoblee93:
Can you use it as a git dependency?

jacoblee93:
Hey @davidfant I opened #3885 - it's a bit rough but LMK what your thoughts are on it and if it'd fit your use-case. Please merge it into your branch if you approve.

davidfant:
> Can you use it as a git dependency?

How? I tried `yarn add https://github.com/langchain-ai/langchainjs#jacob/handlebars` but it doesn't work because the repo isn't built and is using workspaces



> Hey @davidfant spent some time on this this morning and opened #3885 - it's a bit rough but LMK what your thoughts are on it and if it'd fit your use-case. Please merge it into your branch if you approve.

Appreciate it! Will try once I figure out the above

davidfant:
Added comments to #3885 
It's working for me. FYI this is what I ended up doing locally and in my Docker container to use your branch:
```
git clone -b jacob/handlebars https://github.com/langchain-ai/langchainjs.git \
 && cd langchainjs \
 && yarn \
 && yarn workspace langchain build \
 && YARN_IGNORE_PATH=1 yarn workspace langchain link \
 && YARN_IGNORE_PATH=1 yarn workspace @langchain/core link \
 && YARN_IGNORE_PATH=1 yarn workspace @langchain/openai link
```

jacoblee93:
Going to close this in favor of #3885 - will aim to get that with today's releases!

PR #3792, opened 2023-12-26
core[minor]: Nc/dec26/runnable stream

nfcampos:
Fixes # (issue)

### DIFF
--- a/langchain-core/src/runnables/base.ts
+++ b/langchain-core/src/runnables/base.ts
@@ -12,7 +12,9 @@ import {
 import { Serializable } from "../load/serializable.js";
 import {
   IterableReadableStream,
+  concat,
   type IterableReadableStreamInterface,
+  atee,
 } from "../utils/stream.js";
 import {
   RunnableConfig,
@@ -428,13 +430,13 @@ export abstract class Runnable<
     const callbackManager_ = await getCallbackMangerForConfig(options);
     const runManager = await callbackManager_?.handleChainStart(
       this.toJSON(),
-            { input: "" },
-            undefined,
-            options?.runType,
-            undefined,
-            undefined,
-            options?.runName
-          );
+      { input: "" },
+      undefined,
+      options?.runType,
+      undefined,
+      undefined,
+      options?.runName
+    );
     async function* wrapInputForTracing() {
       for await (const chunk of inputGenerator) {
         if (finalInputSupported) {
@@ -1485,7 +1487,7 @@ export class RunnableMap<
         Object.entries(this.steps).map(async ([key, runnable]) => {
           output[key] = await runnable.invoke(
             input,
-            this._patchConfig(options, runManager?.getChild(key))
+            this._patchConfig(options, runManager?.getChild(`map:key:${key}`))
           );
         })
       );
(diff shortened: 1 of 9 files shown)


langchain-core/src/runnables/base.ts
```diff
@@ -1537,22 +1588,29 @@ export class RunnableLambda<RunInput, RunOutput> extends Runnable<
...
+      if (config?.recursionLimit === 0) {
+        throw new Error("Recursion limit reached.");
+      }
       output = await output.invoke(
         input,
-        this._patchConfig(config, runManager?.getChild())
+        this._patchConfig(
+          config,
+          runManager?.getChild(),
+          (config?.recursionLimit ?? 25) - 1
```
jacoblee93, line 1604: Factor out magic number 25 somewhere else?

langchain-core/src/runnables/passthrough.ts
```diff
@@ -31,6 +33,66 @@ export class RunnableAssign<
...
     } as RunOutput;
   }
+
+  async *_transform(
+    generator: AsyncGenerator<RunInput>,
+    runManager?: CallbackManagerForChainRun,
+    options?: Partial<RunnableConfig>
+  ): AsyncGenerator<RunOutput> {
+    // collect mapper keys
+    const mapper_keys = this.mapper.getStepsKeys();
```
jacoblee93, line 43: Prefer camel case

langchain-core/src/utils/stream.ts
```diff
@@ -98,3 +98,63 @@ export class IterableReadableStream<T>
     });
   }
 }
+
+export function atee<T>(
```
jacoblee93, line 102: Just `tee` is better?

langchain-core/src/runnables/passthrough.ts
```diff
@@ -31,6 +33,66 @@ export class RunnableAssign<
...
+    // create two input gens, one for the mapper, one for the input
+    const [forPassthrough, forMapper] = atee(generator, 2);
+    // create mapper output gen
+    const mapperOutput = this.mapper.transform(
```
jacoblee93, line 47: What's the purpose of the tee? Can we not just:

```typescript
      const filtered = Object.fromEntries(
        Object.entries(chunk).filter(([key]) => !mapper_keys.includes(key))
      );
      if (Object.keys(filtered).length > 0) {
        yield filtered as unknown as RunOutput;
      }
```
below along with the other chunks?

langchain-core/src/utils/stream.ts
nfcampos: Tee would be for non-async generators

langchain-core/src/runnables/base.ts
```diff
@@ -433,30 +428,24 @@ export abstract class Runnable<
...
+      { input: "" },
+      undefined,
+      options?.runType,
+      undefined,
+      undefined,
+      options?.runName
+    );
     async function* wrapInputForTracing() {
       for await (const chunk of inputGenerator) {
-        if (!runManager) {
```
jacoblee93, line 440: Ah nice that this is fixed

langchain-core/src/runnables/passthrough.ts
nfcampos: We can't consume the same generator twice

langchain-core/src/utils/stream.ts
```diff
@@ -98,3 +98,63 @@ export class IterableReadableStream<T>
...
+export function concat<
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  T extends Array<any> | string | number | Record<string, any> | any
+>(first: T, second: T): T {
+  if (Array.isArray(first) && Array.isArray(second)) {
+    return first.concat(second) as T;
+  } else if (typeof first === "string" && typeof second === "string") {
+    return (first + second) as T;
+  } else if (typeof first === "number" && typeof second === "number") {
+    return (first + second) as T;
```
jacoblee93, line 136: This seems like an odd thing to do

nfcampos: This is what we do in py

langchain-core/src/runnables/passthrough.ts
jacoblee93: Oh I get it, never mind.

jacoblee93:
Thanks so much for this!

langchain-core/src/utils/stream.ts
jacoblee93: Streaming number would be a bit weird in general I guess

PR #3794, opened 2023-12-26
[HOLD]docs[minor]: Allow for ipynb files in docs

bracesproul:
On hold until we figure out import maps


CONTRIBUTING.md
```diff
@@ -270,6 +270,19 @@ Similar to linting, we recognize documentation can be annoying. If you do not wa
...
 Documentation and the skeleton lives under the `docs/` folder. Example code is imported from under the `examples/` folder.
 
+### To contribute a new documentation notebook:
+
+#### Install dependencies
+
+1. [Quarto](https://quarto.org/) - package that converts Jupyter notebooks (.ipynb files) into mdx files for serving in Docusaurus (only necessary if building docs site locally).
+2. Navigate to the `docs/core_docs/docs/` directory.
+3. Use the provided yarn script to create a new Jupyter notebook:
+```shell
```
jacoblee93, line 280: Why not just create a new file and name it `x.ipynb`?

docs/core_docs/docs/deno.json
```diff
@@ -0,0 +1,25 @@
...
+    "@langchain/openai": "npm:/file:../../../libs/langchain-openai",
+    "@langchain/openai/": "npm:/file:../../../libs/langchain-openai/",
+    "@langchain/mistralai": "npm:/file:../../../libs/langchain-mistralai",
+    "@langchain/mistralai/": "npm:/file:../../../libs/langchain-mistralai/",
+    "@langchain/anthropic": "npm:/file:../../../libs/langchain-anthropic",
+    "@langchain/anthropic/": "npm:/file:../../../libs/langchain-anthropic/",
+    "@langchain/cohere": "npm:/file:../../../libs/langchain-cohere",
+    "@langchain/cohere/": "npm:/file:../../../libs/langchain-cohere/",
+    "@langchain/google-genai": "npm:/file:../../../libs/langchain-google-genai",
+    "@langchain/google-genai/": "npm:/file:../../../libs/langchain-google-genai/",
```
jacoblee93, line 16: This will need to be updated every time we add an integration package?

bracesproul: Yes, along with any import someone wants to add in a notebook. I could write a script to read the notebook imports and update this @ build time but that doesn't handle runtime

CONTRIBUTING.md
bracesproul: If you want to access anything like `process.` you'll need to import a util file, this script writes a notebook which automatically adds that as the first cell.

docs/core_docs/docs/deno.json
jacoblee93: Yeah no worries, can treat it like a dep. Can you add docs though?

jacoblee93:
Would be nice to not need an autogen script and I'm not sure why it's necessary?

Also, we're moving the assistants API notebook from agent types here: #3802

Could you pick a different notebook to try this on?

bracesproul:
> Would be nice to not need an autogen script and I'm not sure why it's necessary?
> 
> Also, we're moving the assistants API notebook from agent types here: #3802
> 
> Could you pick a different notebook to try this on?

Unfortunate that's the doc I picked haha, will refactor.

Generating the file isn't necessary if you don't plan on explicitly calling things like `process.cwd()` or `process.env.MY_ENV` as those are not available in Deno. The script just imports a util file which writes new global variables for `process`, amongst others (eg, in Deno `writeFile` is a globally available method... so I export that as well)

CONTRIBUTING.md
jacoblee93: Can we shim it or start enforcing `getEnvironmentVariable` usage?

jacoblee93:
I think fine to try to work around it for now - we have the `getEnvrionmentVariable` util which should work in Deno anyway

docs/core_docs/.gitignore
```diff
@@ -29,3 +29,9 @@ yarn-error.log*
 !.yarn/releases
 !.yarn/sdks
 !.yarn/versions
+
+.quarto/
+
+
+# notebooks
+docs/modules/agents/agent_types/openai_assistant_2.md
```
jacoblee93, line 37: Can we just move all existing `.md` files to `.mdx` so that we don't need to individually add these?

docs/core_docs/scripts/notebook_init_utils.js
```diff
@@ -0,0 +1,11 @@
+/* eslint-disable */
```
jacoblee93, line 1: Yeah let's cut these if we can't find an automatic way to import them. If we really need something Node specific we can just use `.mdx`

I can also reach out to the Deno team to see if they have suggestions, we have a channel with them.

docs/core_docs/docs/deno.json
```diff
@@ -1,21 +1,18 @@
...
-    "@langchain/mistralai": "npm:/file:../../../libs/langchain-mistralai",
-    "@langchain/mistralai/": "npm:/file:../../../libs/langchain-mistralai/",
-    "@langchain/anthropic": "npm:/file:../../../libs/langchain-anthropic",
-    "@langchain/anthropic/": "npm:/file:../../../libs/langchain-anthropic/",
-    "@langchain/cohere": "npm:/file:../../../libs/langchain-cohere",
-    "@langchain/cohere/": "npm:/file:../../../libs/langchain-cohere/",
-    "@langchain/google-genai": "npm:/file:../../../libs/langchain-google-genai",
-    "@langchain/google-genai/": "npm:/file:../../../libs/langchain-google-genai/",
+    "langchain/": "npm:langchain/",
+    "@langchain/core/": "npm:langchain/core/",
```
jacoblee93, line 4: This is wrong but let's also really try to figure out how to do this pulling in local code

bracesproul: whoops, fixed. and agreed.

PR #3795, opened 2023-12-26
Implement optional message placeholder in js

nfcampos:
Fixes # (issue)

### DIFF
--- a/langchain-core/src/prompts/chat.ts
+++ b/langchain-core/src/prompts/chat.ts
@@ -75,6 +75,7 @@ export abstract class BaseMessagePromptTemplate<
  */
 export interface MessagesPlaceholderFields<T extends string> {
   variableName: T;
+  optional?: boolean;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


langchain-core/src/prompts/chat.ts
```diff
@@ -108,16 +114,27 @@ export class MessagesPlaceholder<
...
   validateInputOrThrow(
-    input: Array<unknown>,
+    input: Array<unknown> | undefined,
     variableName: Extract<keyof RunInput, string>
   ): input is BaseMessage[] {
+    if (this.optional && !input) {
+      return false;
+    } else if (!input) {
+      const error = new Error(
+        `Error: Field "${variableName}" in prompt uses a MessagesPlaceholder, which expects an array of BaseMessages as an input value. Received: undefined`
```
jacoblee93, line 132: Suppose on the off chance if someone's using vanilla JS slightly better to `JSON.stringify(input)`

nfcampos: I just copied from below

PR #3797, opened 2023-12-26
Refactor the MessageContent type.

afirstenberg:
- Break it into individual subtypes to make it clear what the definition is and to be able to use these subtypes as parameter or return types.
- Add a new subtype blob_url which takes a required URL and a required content type (MIME type). This is for models that require the mime type (Gemini, for example) and makes it clear that this may be used for content besides images (ie - video, 3-d models, etc).

### DIFF
--- a/langchain-core/src/messages/index.ts
+++ b/langchain-core/src/messages/index.ts
@@ -34,18 +34,37 @@ export type MessageType =
   | "function"
   | "tool";
 
+export type MessageContentText =
+  {
+    type: "text";
+    text: string;
+  };
+
+export type MessageContentImageUrl =
+  {
+    type: "image_url";
+    image_url: string | { url: string; detail?: "auto" | "low" | "high" };
+  };
+
+export type MessageContentBlobUrl =
+  {
+    type: "blob_url";
+    blob_url: {
+      url: string;
+      content_type: string;
+    };
+  };
+
+export type MessageContentComplex =
+  (
+    | MessageContentText
+    | MessageContentImageUrl
+    | MessageContentBlobUrl
+  );
+
 export type MessageContent =
   | string
-  | (
-      | {
-          type: "text";
-          text: string;
-        }
-      | {
-          type: "image_url";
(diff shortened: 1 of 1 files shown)


langchain-core/src/messages/index.ts
```diff
@@ -135,11 +154,6 @@ export abstract class BaseMessage
 
   lc_serializable = true;
 
-  get lc_aliases(): Record<string, string> {
```
bracesproul, line 138: why did you drop this and the one below?

bracesproul (commented):
besides my one comment, lgtm. cc @jacoblee93

bracesproul:
Ty for doing this @afirstenberg !!

langchain-core/src/messages/index.ts
afirstenberg: Looks like an error on my part. Those were both added on 20 Dec and I may originally have been working off an older version and then just copied the file over and didn't notice.

jacoblee93:
Will definitely want to sync with @baskaryan and the Python folks before merging but I think seems reasonable

afirstenberg:
Any chance we can get this reviewed?
Would it make things faster if we removed the `blob_url` subtype for now and I opened that up as a separate pull request?

(@jacoblee93 @baskaryan)

jacoblee93:
Yes that would make it faster - would prefer to add blob stuff after google releases more multimodal formats if that's ok with you.

Also would like to cut down on the number of Google packages, but still thinking through the best way how...

afirstenberg:
We'll discuss the blob part separately. (And the Google packages.)
For now, I've removed that part, so this is just refactoring MessageContent so it is easier to read and so the individual types can be used directly.

duckdegen:
@jacoblee93 what is currently preventing this PR from being merged?

Is there any action necessary on the contributor side?
Asking because this is blocking #3960 from being merged.

Ty

jacoblee93:
So sorry for the delay @afirstenberg and @duckdegen, this looks great.

PR #3806, opened 2023-12-27
Add optional name for runnable sequence

nfcampos:
Fixes # (issue)

### DIFF
--- a/langchain-core/src/runnables/base.ts
+++ b/langchain-core/src/runnables/base.ts
@@ -124,6 +124,15 @@ export abstract class Runnable<
 {
   protected lc_runnable = true;
 
+  name?: string;
+
+  getName(suffix?: string): string {
+    const name =
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      this.name ?? (this.constructor as any).lc_name ?? this.constructor.name;
+    return suffix ? `${name}${suffix}` : name;
+  }
+
   abstract invoke(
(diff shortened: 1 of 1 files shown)


langchain-core/src/runnables/base.ts
```diff
@@ -124,6 +124,15 @@ export abstract class Runnable<
 {
   protected lc_runnable = true;
 
+  name?: string;
+
+  getName(suffix?: string): string {
```
jacoblee93, line 129: Do we need to add to the interface?

langchain-core/src/runnables/base.ts
```diff
@@ -124,6 +124,15 @@ export abstract class Runnable<
...
+    const name =
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      this.name ?? (this.constructor as any).lc_name ?? this.constructor.name;
```
jacoblee93, line 132: `lc_name` is a function

nfcampos: Oops

langchain-core/src/runnables/base.ts
nfcampos: Yes this will be used by the schemas too, and I think it makes sense to be public

nfcampos: Actually I won't add to interface

PR #3807, opened 2023-12-27
Add .pick and .assign methods to Runnable

nfcampos:
Fixes # (issue)


langchain-core/src/runnables/base.ts
```diff
@@ -540,6 +540,31 @@ export abstract class Runnable<
...
+  pick(keys: string | string[]): RunnableSequence {
+    // eslint-disable-next-line @typescript-eslint/no-use-before-define, @typescript-eslint/no-explicit-any
+    return this.pipe(new RunnablePick(keys) as any);
+  }
+
+  /**
+   * Assigns new fields to the dict output of this runnable. Returns a new runnable.
+   */
+  assign(
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
jacoblee93, line 555: Don't need this linter comment

langchain-core/src/runnables/passthrough.ts
```diff
@@ -1,206 +1,12 @@
...
-  ): Promise<IterableReadableStream<RunOutput>> {
-    async function* generator() {
-      yield input;
-    }
-    return IterableReadableStream.fromAsyncGenerator(
-      this.transform(generator(), options)
-    );
-  }
-}
+export { RunnableAssign };
```
jacoblee93, line 9: I think you can remove this export? Only entrypoint is through `index.ts`

langchain-core/src/runnables/base.ts
```diff
@@ -540,6 +540,31 @@ export abstract class Runnable<
     });
   }
 
+  /**
+   * Pick keys from the dict output of this runnable. Returns a new runnable.
+   */
+  pick(keys: string | string[]): RunnableSequence {
+    // eslint-disable-next-line @typescript-eslint/no-use-before-define, @typescript-eslint/no-explicit-any
+    return this.pipe(new RunnablePick(keys) as any);
```
jacoblee93, line 548: Do we need the cast? And if we do, `as Runnable` is better?

langchain-core/src/runnables/passthrough.ts
nfcampos: Done

langchain-core/src/runnables/base.ts
nfcampos: we need because of the dict input restriction

PR #3809, opened 2023-12-27
langchain[minor]: Add stuff docs chain

nfcampos:
Fixes # (issue)


langchain/src/chains/combine_documents/stuff.ts
```diff
@@ -0,0 +1,61 @@
...
+    "page_content" will automatically retrieve the `Document.page_content`, and all 
+    other inputs variables will be automatically retrieved from the `Document.metadata` dictionary. Default to a prompt that only contains `Document.page_content`.
+ * @param documentSeparator String separator to use between formatted document strings.
+ * @returns An LCEL `Runnable` chain.
+    Expects a dictionary as input with a list of `Document`s being passed under
+    the "context" key.
+    Return type depends on the `output_parser` used.
+ */
+export async function createStuffDocumentsChain(
+  llm: LanguageModelLike,
```
jacoblee93, line 38: Prefer "bag of props" syntax here

Would be difficult to remember argument order IMO

langchain/src/chains/combine_documents/refine.ts
```diff
@@ -0,0 +1,144 @@
...
+      summary. Respond with only the updated summary and no other text.
+      Here is your working summary:\n\n{output}.`],
+ *  ["user", "Here is the next page:\n\n{context}"]
+ * ])
+ * 
+ * const chain = await createRefineDocumentsChain(llm, initialPrompt, refinePrompt);
+ * const output = await chain.invoke({ context: documents, question: "..." });
+ */
+export async function createRefineDocumentsChain(
+  llm: LanguageModelLike,
```
jacoblee93, line 68: Prefer "bag of props" syntax here

Would be difficult to remember argument order IMO

langchain/src/chains/combine_documents/refine.ts
```diff
@@ -0,0 +1,144 @@
...
+      // remaining documents
+      (inputs: Record<string, Document[]>) => {
+        const docs = inputs[DOCUMENTS_KEY];
+        if (docs.length <= 1) {
+          return new RunnablePassthrough();
+        } else if (docs.length === 2) {
+          return refineStep;
+        } else {
+          return RunnableSequence.from(
+            docs.slice(1).map(() => refineStep) as unknown as [
```
jacoblee93, line 123: Such a hard cast is necessary? Could we type `inputs[DOCUMENTS_KEY]` to `Runnable[]` instead?

nfcampos: No the issue is the array length. There's no way to make typescript understand an array map has at least 2 elements

jacoblee93:
We'll definitely want tests, I can add tests later if you want to move on

Issue #3784, opened 2023-12-26
MomentoChatMessageHistory is not working on the Cloudflare Workers because of the SDK difference


hideokamoto:
When I tried to use Momento for memory the chat history in my Cloudflare Workers application, I got the following error:

```
✘ [ERROR] Could not resolve "@gomomento/sdk"

    node_modules/langchain/dist/util/momento.js:2:28:
      2 │ import { CreateCache } from "@gomomento/sdk";
        ╵                             ~~~~~~~~~~~~~~~~

  You can mark the path "@gomomento/sdk" as external to exclude it from the bundle, which will remove this error.


╭─────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ [b] open a browser, [d] open Devtools, [l] turn off local mode, [c] clear console, [x] to exit                  │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯

✘ [ERROR] Failed to build


If you think this is a bug then please create an issue at https://github.com/cloudflare/workers-sdk/issues/new/choose
```

### Create Cloudflare Workers application by using Hono

```
$ npm create hono@latest
```

And choose the `cloudflare-workers`.

### Install libraries

```
npm i langchain xhr4sw @gomomento/sdk-web
```

### Insert the example code

via: https://js.langchain.com/docs/integrations/chat_memory/momento

```typescript
import { Hono } from "hono";
import {
    CacheClient,
    Configurations,
    CredentialProvider,
  } from "@gomomento/sdk-web"; // `from "gomomento/sdk-web";` for browser/edge
  import { BufferMemory } from "langchain/memory";
  import { ChatOpenAI } from "langchain/chat_models/openai";
  import { ConversationChain } from "langchain/chains";
  import { MomentoChatMessageHistory } from "langchain/stores/message/momento";
  
import XMLHttpRequestPolyfill from "xhr4sw";
/**
 * This is needed to polyfill as otherwise we get HttpRequest not defined
 */
Object.defineProperty(self, 'XMLHttpRequest', {
	configurable: false,
	enumerable: true,
	writable: false,
	value: XMLHttpRequestPolyfill
});

 const app = new Hono<{
    Bindings: {
        OPENAI_API_KEY: string;
        MOMENT_API_KEY: string;
    }
}>()

app.get('test', async c => {
    const cacheClient = new CacheClient({
        configuration: Configurations.Laptop.v1(),
        credentialProvider: CredentialProvider.fromString({
[...]

PR #3813, opened 2023-12-28
fix: use @gomoment/sdk-core instead of @gomoment/sdk for edge server integrations (#3784)

hideokamoto:
Patch for the issue:
 MomentoChatMessageHistory is not working on the Cloudflare Workers because of the SDK difference #3784


malandis:
LGTM. Thank you @hideokamoto 👍

jacoblee93:
Thank you!

PR #3802, opened 2023-12-27
langchain[minor]: Adds new agent create methods and docs

jacoblee93:
CC @nfcampos @hwchase17

PR #3819, opened 2023-12-29
core[minor]: Streaming List Parsers

nfcampos:
- Implement streaming for CSV parser
- Implement two new streaming list parsers: Numbered list, Markdown List

Fixes # (issue)


jacoblee93:
Note to self: update the table in #3802 when this is merged

PR #3842, opened 2023-12-30
docs[patch]: relevant example for map reduce

anass-arrhioui-nw:
The old example doesn't showcase the utility of loadQAMapReduceChain, and the console.log output is not displayed in the docs. The new example is more relevant and includes a console.log output.


bracesproul (approved):
Nice! Thanks for improving this, will get it merged in as soon as CI passes!

PR #3843, opened 2023-12-30
fix(doc): Sidebar how_to (tools)

gkhngyk:
- Fix position and name failure by adding _category_.yml file.
- Do not hide the agents_with_vectorstores.mdx just because there is only one document in the how-to.

Screenshot 2023-12-31 at 01 44 47


bracesproul (approved):
Good catch! Thanks for opening this pr, will get it merged in as soon as CI passes!

PR #3844, opened 2023-12-31
core[patch]: Add generics to `RunnablePassthrough.assign`

### LINKED ISSUE #3827: RunnablePassthrough.assign doesn't infer input types
I assume this is a bug on the typescript types for RunnablePassthrough.assign(). My assumption is that RunnablePassthrough.assign() allows to add additional keys to a given input object via an inline function that can receive an input parameter and returns a value. 

```ts
type Input = { question: string; history: string };

const getDocuments = async (input: Input) => {
  return Promise.resolve([
    { id: 1, text: 'foo' },
    { id: 2, text: 'bar' },
  ]);
};

  {
    question: (input: Input) => input.question,
    history: (input: Input) => input.history,
  },
  RunnablePassthrough.assign({
    documents: async (input: Input) => getDocuments(input),
//  ^^^^^^^^
//  Type '(input: Input) => Promise<{ id: number; text: string; }[]>' is not assignable to type 'RunnableLike<Record<string, unknown>, any>'.
//    Type '(input: Input) => Promise<{ id: number; text: string; }[]>' is not assignable to type 'RunnableFunc<Record<string, unknown>, any>'.
//      Types of parameters 'input' and 'input' are incompatible.
//        Type 'Record<string, unknown>' is missing the following properties from type 'Input': question, history
  }),
]);

// result should contain the original input plus the documents
const result = chain.invoke({ question: 'Hello World', history: '' });
```

I get a type error on `documents` because its parameter is typed as `Record<string, unknown>` and incompatible with the `Input` type.

I would expect that the input type is either inferred by the parameter of the inline function `(input: Input) => {...}` or it should be possible to specify it as generic type on `assign` like RunnablePassthrough.assign<Input>(...)`. The same is possible with RunnableMap.

zirkelc:
Fixes #3827

This PR adds the generics `RunInput` and `RunOutput` to `RunnablePasstrough.assign()`. It infers run input type from the given RunnableMap and constructs the output type as merged type of run input and output.

Here's an example:

```ts
  type InferRunnableOutput<R extends Runnable<any, any>> = R extends Runnable<any, infer O>
    ? O
    : never;

  const inputMap = RunnableMap.from({
    context: (input: { context: string; question: string }) => input.context,
    question: (input: { context: string; question: string }) => input.question,
  });

  const chat = new FakeChatModel({});

  const assignChat = RunnablePassthrough.assign({
    chat: (input: InferRunnableOutput<typeof inputMap>) => chat,
  });

  const llm = new FakeStreamingLLM({ sleep: 0 });

  const assignLLM = RunnablePassthrough.assign({
    llm: (input: InferRunnableOutput<typeof assignChat>) => llm,
  });

  type Output = InferRunnableOutput<typeof assignLLM>;
  //   ^^^^^^
  // type Output = {
  //     context: string;
  //     question: string;
  // } & {
  //     chat: FakeChatModel;
  // } & {
  //     llm: FakeStreamingLLM;
  // }
```

The final output type is being inferred as the merged type from `inputMap`, `assignChat` and `assignLLM`.


langchain-core/src/runnables/passthrough.ts
```diff
@@ -94,12 +94,12 @@ export class RunnablePassthrough<RunInput> extends Runnable<
    * });
    * ```
    */
-  static assign(
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    mapping: RunnableMapLike<Record<string, unknown>, Record<string, unknown>>
-  ): RunnableAssign<Record<string, unknown>, Record<string, unknown>> {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  static assign<RunInput extends Record<string, any>, RunOutput extends Record<string, any>>(
```
bracesproul, line 98: Any reason you switched `Record<string, unknown>` to `Record<string, any>`? If possible I'd prefer to keep `unknown` over `any`, but if that was giving issues it's not too big a deal.

bracesproul (commented):
One small question, besides that, this looks like a good improvement!

bracesproul:
Also looks like lint is failing, if you could run `yarn format` from root and push up the changes that would be great

langchain-core/src/runnables/passthrough.ts
zirkelc: No particular reason, I just wanted to align it with the other runnables that use `Record<string, any>`. I switched it back to `Record<string, unknown>`, it shouldn't cause any issues.

zirkelc:
Hey @bracesproul I applied the formatting and switched back to `Record<string, unknown>`.  Let me know if there's something else to do! :-)

bracesproul (approved):
lgtm, cc @jacoblee93 for a second look.

Thank you for adding this!

jacoblee93:
This is great, thank you!

PR #3846, opened 2024-01-01
fix: textFieldMaxLength parameter missing

CoalYa:
Currently, the textFieldMaxLength in the dbconfig parameter is lost when using the Milvus.fromDocuments method
```
const vectorStore = await Milvus.fromDocuments(docs, new OpenAIEmbeddings(), {
  collectionName: "goldel_escher_bach",
  textFieldMaxLength: 1024
});
```

### DIFF
--- a/libs/langchain-community/src/vectorstores/milvus.ts
+++ b/libs/langchain-community/src/vectorstores/milvus.ts
@@ -490,18 +490,7 @@ export class Milvus extends VectorStore {
     embeddings: EmbeddingsInterface,
     dbConfig?: MilvusLibArgs
   ): Promise<Milvus> {
-    const args: MilvusLibArgs = {
-      collectionName: dbConfig?.collectionName || genCollectionName(),
-      url: dbConfig?.url,
-      ssl: dbConfig?.ssl,
-      username: dbConfig?.username,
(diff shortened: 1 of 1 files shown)


libs/langchain-community/src/vectorstores/milvus.ts
```diff
@@ -490,18 +490,7 @@ export class Milvus extends VectorStore {
...
-      ssl: dbConfig?.ssl,
-      username: dbConfig?.username,
-      password: dbConfig?.password,
-      textField: dbConfig?.textField,
-      primaryField: dbConfig?.primaryField,
-      vectorField: dbConfig?.vectorField,
-      clientConfig: dbConfig?.clientConfig,
-      autoId: dbConfig?.autoId,
-    };
+    const args: MilvusLibArgs = dbConfig ?? {};
```
bracesproul, line 493:
```suggestion
    const args: MilvusLibArgs = {
      ...(dbConfig ? dbConfig : {}),
      collectionName: dbConfig?.collectionName || genCollectionName()
    }
```

bracesproul (approved):
Good catch, thanks for this! I pushed one small commit, but as soon as CI passes I can get this merged in!

PR #3849, opened 2024-01-01
Neo4j propagate error in query method

tomasonjo:


libs/langchain-community/src/graphs/neo4j_graph.ts
```diff
@@ -118,8 +118,8 @@ export class Neo4jGraph {
       ) {
         throw new Error("Procedure not found in Neo4j.");
       }
+      throw error;
```
bracesproul, line 121: Although this does look like a better solution, it's not backwards compatible. Existing applications would be expecting this to return `undefined` but now it'll throw and it could cause major issues for anyone not expecting that.

bracesproul (commented):
I'm going to close this PR as it's not backwards compatible, however when we eventually get around to releasing a new major LangChain.js version which contains breaking changes, this should go in. I'd suggest opening a new PR solely containing an `@TODO` comment explaining what to change when we can make breaking changes.

PR #3854, opened 2024-01-02
fix a bug in Bedrock streaming

### LINKED ISSUE #3850: Bedrock streaming breaks
Streaming with bedrock breaks with an exception.
It does not always occure, it depends on the behaviour of the bedrock service that cannot be controlled by the client.

From my observations it typically happens after many successful chunks arrive successfuly.
I was able to reproduce it many times with the following setup:
Bedrock + Anthropic + Claude v2 + Long Hebrew completions.
This is not the only way to reproduce it, but it happens very ofen in the given setup.

Here is an example of a Hebrew prompt to reproduce:
`כתוב מאמר ארוך על משבר האקלים`

This is the exception:

`ERROR bedrock-chat-web Logger  AIQuestionStream: runtime error: Offset is outside the bounds of the DataView / stack: RangeError: Offset is outside the bounds of the DataView
    at DataView.getUint32 (<anonymous>)
    at getMessageLength (C:\Users\username\Desktop\bedrock-chat-web\server\node_modules\langchain\node_modules\@langchain\community\dist\chat_models\bedrock\web.cjs:335:25)
    at [Symbol.asyncIterator] (C:\Users\username\Desktop\bedrock-chat-web\server\node_modules\langchain\node_modules\@langchain\community\dist\chat_models\bedrock\web.cjs:344:41)
    at async BedrockChat._streamResponseChunks (C:\Users\username\Desktop\bedrock-chat-web\server\node_modules\langchain\node_modules\@langchain\community\dist\chat_models\bedrock\web.cjs:289:30)
    at async BedrockChat._call (C:\Users\username\Desktop\bedrock-chat-web\server\node_modules\langchain\node_modules\@langchain\community\dist\chat_models\bedrock\web.cjs:210:30)
    at async BedrockChat._generate (C:\Users\username\Desktop\bedrock-chat-web\server\node_modules\@langchain\core\dist\language_models\chat_models.cjs:344:22)
    at async Promise.allSettled (index 0)
    at async BedrockChat._generateUncached (C:\Users\username\Desktop\bedrock-chat-web\server\node_modules\@langchain\core\dist\language_models\chat_models.cjs:114:25)
    at async LLMChain._call (C:\Users\username\Desktop\bedrock-chat-web\server\node_modules\langchain\dist\chains\llm_chain.cjs:157:37)`

--
I've looked into the source code.
There is a structure to the event stream.
The code reads the message length from the first 4 byts of the stream and waits for the buffer to fill before it tries to encode it and pass it on.

The code below does not check for buffer.byteLength between 1 and 4 and failes to read an Uint32Array (4 bytes) from a buffer of 1 to 3 bytes length.

https://github.com/langchain-ai/langchainjs/blob/df490a49709c3c78b71cc1c29b3be37b3c8b8909/libs/langchain-community/src/chat_models/bedrock/web.ts#L401

It looks like a new bug introduced in #3364

Yofial:
Fixes #3850

### DIFF
--- a/libs/langchain-community/src/chat_models/bedrock/web.ts
+++ b/libs/langchain-community/src/chat_models/bedrock/web.ts
@@ -393,7 +393,7 @@ export class BedrockChat extends SimpleChatModel implements BaseBedrockInput {
     const PRELUDE_TOTAL_LENGTH_BYTES = 4;
 
     function getMessageLength(buffer: Uint8Array) {
-      if (buffer.byteLength < PRELUDE_TOTAL_LENGTH_BYTES ) return 0;
+      if (buffer.byteLength < PRELUDE_TOTAL_LENGTH_BYTES) return 0;
       const view = new DataView(
         buffer.buffer,
         buffer.byteOffset,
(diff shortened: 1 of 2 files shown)


bracesproul (approved):
looks reasonable, thanks for catching & fixing this bug!

PR #3855, opened 2024-01-02
Todo: propagate error in Neo4jGraph

tomasonjo:
When a new major version of LangChainJS will be released, please change the code to propagate the error in Neo4jGraph.
Based on https://github.com/langchain-ai/langchainjs/pull/3849

### DIFF
--- a/libs/langchain-community/src/graphs/neo4j_graph.ts
+++ b/libs/langchain-community/src/graphs/neo4j_graph.ts
@@ -119,7 +119,7 @@ export class Neo4jGraph {
         throw new Error("Procedure not found in Neo4j.");
       }
     }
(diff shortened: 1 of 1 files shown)


bracesproul (approved):
thanks for this!

libs/langchain-community/src/graphs/neo4j_graph.ts
```diff
@@ -119,6 +119,7 @@ export class Neo4jGraph {
         throw new Error("Procedure not found in Neo4j.");
       }
     }
+    //@TODO Propagate error instead of returning undefined
```
bracesproul, line 122:
```suggestion
    /** @TODO Propagate error instead of returning undefined */
```

PR #3857, opened 2024-01-02
Add generic output type parameter to JsonOutputFunctionsParser

zirkelc:
This PR adds a generic type parameter for the json output type. When used in combination with a Zod schema, it is possible to infer the output type from this schema via `z.infer<typeof schema>`.

Example: 
```ts
  const schema = z.object({
    setup: z.string().describe("The setup for the joke"),
    punchline: z.string().describe("The punchline to the joke"),
  });

  const parser = new JsonOutputFunctionsParser<z.infer<typeof schema>>({
    diff: true,
  });
  
  const chain = prompt.pipe(model).pipe(parser);
  //    ^^^^^
  // Inferred as:
  // RunnableSequence<ParamsFromFString<"tell me a long joke about {foo}">, {
  //  setup: string;
  //  punchline: string;
  // }>
```

### DIFF
--- a/langchain/src/output_parsers/openai_functions.ts
+++ b/langchain/src/output_parsers/openai_functions.ts
@@ -79,7 +79,9 @@ export class OutputFunctionsParser extends BaseLLMOutputParser<string> {
  * Class for parsing the output of an LLM into a JSON object. Uses an
  * instance of `OutputFunctionsParser` to parse the output.
  */
-export class JsonOutputFunctionsParser extends BaseCumulativeTransformOutputParser<object> {
+export class JsonOutputFunctionsParser<
+  T = object
+> extends BaseCumulativeTransformOutputParser<T> {
   static lc_name() {
     return "JsonOutputFunctionsParser";
   }
@@ -113,7 +115,7 @@ export class JsonOutputFunctionsParser extends BaseCumulativeTransformOutputPars
 
   async parsePartialResult(
     generations: ChatGeneration[]
-  ): Promise<object | undefined> {
+  ): Promise<T | undefined> {
     const generation = generations[0];
     if (!generation.message) {
       return undefined;
@@ -130,7 +132,7 @@ export class JsonOutputFunctionsParser extends BaseCumulativeTransformOutputPars
     return {
       ...functionCall,
       arguments: parsePartialJson(functionCall.arguments),
-    };
+    } as T;
   }
 
   /**
@@ -139,9 +141,7 @@ export class JsonOutputFunctionsParser extends BaseCumulativeTransformOutputPars
    * @param generations The output of the LLM to parse.
    * @returns A JSON object representation of the function call or its arguments.
    */
-  async parseResult(
-    generations: Generation[] | ChatGeneration[]
-  ): Promise<object> {
+  async parseResult(generations: Generation[] | ChatGeneration[]): Promise<T> {
     const result = await this.outputParser.parseResult(generations);
     if (!result) {
       throw new Error(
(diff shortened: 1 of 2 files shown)


langchain/src/output_parsers/openai_functions.ts
```diff
@@ -130,7 +132,7 @@ export class JsonOutputFunctionsParser extends BaseCumulativeTransformOutputPars
     return {
       ...functionCall,
       arguments: parsePartialJson(functionCall.arguments),
-    };
+    } as T;
```
jacoblee93, line 135: This should probably be `Partial<T>` right?

langchain/src/output_parsers/tests/openai_functions.int.test.ts
```diff
@@ -46,7 +48,10 @@ test("Streaming JSON patch", async () => {
   for await (const chunk of stream) {
     console.log(chunk);
     chunks.push(chunk);
-    aggregate = applyPatch(aggregate, chunk as Operation[]).newDocument;
+    aggregate = applyPatch(
+      aggregate,
+      chunk as unknown as Operation[]
```
jacoblee93, line 53: This seems like a regression?

jacoblee93:
This is great, thank you! Some small questions

langchain/src/output_parsers/openai_functions.ts
zirkelc: Yes, it should, but the parent class `BaseCumulativeTransformOutputParser` defines it as `parsePartialResult(generations: Generation[] | ChatGeneration[]): Promise<T | undefined>`:

```ts
export declare abstract class BaseCumulativeTransformOutputParser<T = unknown> extends BaseTransformOutputParser<T> {
    protected diff: boolean;
    constructor(fields?: BaseCumulativeTransformOutputParserInput);
    protected abstract _diff(prev: any | undefined, next: any): any;
    abstract parsePartialResult(generations: Generation[] | ChatGeneration[]): Promise<T | undefined>;
    _transform(inputGenerator: AsyncGenerator<string | BaseMessage>): AsyncGenerator<T>;
}
```

So changing the return type `Partial<T>` doesn't comply with parent definition. I assume changing the parent class would be a major change?

langchain/src/output_parsers/tests/openai_functions.int.test.ts
zirkelc: Without the cast to unknown, it gives the following error:

```ts
Conversion of type '{ setup: string; punchline: string; }' to type 'Operation[]' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Type '{ setup: string; punchline: string; }' is missing the following properties from type 'Operation[]': length, pop, push, concat, and 29 more.ts(2352)
```

There is no overlap to the `Operation` type:

```ts
export type Operation = AddOperation<any> | RemoveOperation | ReplaceOperation<any> | MoveOperation | CopyOperation | TestOperation<any> | GetOperation<any>;

export interface BaseOperation {
    path: string;
}
export interface AddOperation<T> extends BaseOperation {
    op: "add";
    value: T;
}
export interface RemoveOperation extends BaseOperation {
    op: "remove";
}
export interface ReplaceOperation<T> extends BaseOperation {
    op: "replace";
    value: T;
}
export interface MoveOperation extends BaseOperation {
    op: "move";
    from: string;
}
export interface CopyOperation extends BaseOperation {
    op: "copy";
    from: string;
}
export interface TestOperation<T> extends BaseOperation {
    op: "test";
    value: T;
}
export interface GetOperation<T> extends BaseOperation {
    op: "_get";
    value: T;
}
export interface PatchResult<T> extends Array<OperationResult<T>> {
    newDocument: T;
}
```

But this problem should have existed before, it just wasn't reported as an error because `chunk` was an `object` but now it is strongly typed to `{ setup: string; punchline: string; }`.

jacoblee93:
I think @bracesproul added this along with #3928 - will close but thank you for pushing on this 🙏

PR #3870, opened 2024-01-03
docs[patch]: Add storage docs

bracesproul:


docs/core_docs/docs/integrations/storage/in_memory.mdx
```diff
@@ -0,0 +1,13 @@
+# In Memory Store
```
jacoblee93, line 1: Looks like in Python this is under `integrations/stores`:

https://python.langchain.com/docs/integrations/stores/in_memory

docs/core_docs/docs/modules/storage/index.mdx
```diff
@@ -0,0 +1,42 @@
+---
```
jacoblee93, line 1: Let's combine into one section and put under `integrations` for now

PR #3871, opened 2024-01-03
core[patch]: Add token usage interface to core

bracesproul:
a ton of integrations now use this and are redefining every time. once this gets released in core I'll put up another pr importing it


langchain-core/src/token_usage.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Shared interface for token usage
+ * return type from LLM calls.
+ */
+export interface TokenUsage {
```
jacoblee93, line 5: Should export from `base_language` instead?

bracesproul: Ah yeah that's probably better. Wasn't 100% on where to put this. ty

PR #3886, opened 2024-01-03
docs[patch]: Replace lc proper OpenAI imports with @langchain/openai

bracesproul:


package.json
```diff
@@ -52,6 +52,7 @@
     "typescript": "~5.1.6"
   },
   "dependencies": {
+    "glob": "^10.3.10",
```
jacoblee93, line 55: Looks like an accident?

PR #3901, opened 2024-01-04
Update embedQuery to use inputType

### LINKED ISSUE #3900: Cohere API rejection when embedQuery using v3 models
`input_type` is a required parameter when using v3 models. It is ignored for v2 models.  Currently the Cohere API rejects the embedding request when a v3 model is used.

EricThompson-PeopleReign:
Input type is a required parameter when using v3 models. It is ignored for v2 models.

Adding it here satisfies both use cases and avoids a Cohere API rejection when using v3 models.

Fixes #3900


bracesproul (approved):
Thanks for this, good catch!

jacoblee93:
Thank you, pushing a new version now.

PR #3907, opened 2024-01-04
allow streaming responses for `ollama-functions` chat model

ljhskyso:
Fixes # (issue)


langchain/src/base_language/index.ts
```diff
@@ -8,6 +8,7 @@ export {
   type BaseLanguageModelInput,
   type BaseLanguageModelInterface,
   BaseLanguageModel,
+  type FunctionDefinition,
```
bracesproul, line 11: Instead of this, just import it from core inside `langchain/src/experimental/chat_models/ollama_functions.ts`
```suggestion
```

langchain/src/experimental/chat_models/ollama_functions.ts
```diff
@@ -1,8 +1,8 @@
 import { BaseChatModel, BaseChatModelParams } from "../../chat_models/base.js";
 import { CallbackManagerForLLMRun } from "../../callbacks/manager.js";
-import { AIMessage, BaseMessage, ChatResult } from "../../schema/index.js";
+import { AIMessage, BaseMessage, ChatGenerationChunk, ChatResult } from "../../schema/index.js";
 import { ChatOllama, type ChatOllamaInput } from "../../chat_models/ollama.js";
-import { BaseFunctionCallOptions } from "../../base_language/index.js";
+import { BaseFunctionCallOptions, FunctionDefinition } from "../../base_language/index.js";
```
bracesproul, line 5:
```suggestion
import { BaseFunctionCallOptions } from "../../base_language/index.js";
import { FunctionDefinition } from "@langchain/core/language_models/base";
```

bracesproul (commented):
Overall this looks good, do you think you could add some tests to go along with it too? Thank you!

bracesproul:
@ljhskyso it looks like format ci is failing, if you could run `yarn format` from root and push up the changes, it should fix it.

ljhskyso:
@bracesproul comments addressed (added an integration test and ran `yarn format`)

bracesproul (approved):
Thanks for this! I pushed up a couple commits just adding simple docs & a nit to the test.

This will be a great addition to have!!

jacoblee93:
Streaming function calls actually works a bit differently than implemented - the partial function call is returned in `additional_kwargs`, the same place as the invoke call:

https://js.langchain.com/docs/modules/model_io/output_parsers/types/json_functions#streaming

I can have a look at this a bit later - agreed that it would be a great addition.

jacoblee93:
I think for now it's not worth it, will revisit in the future

PR #3909, opened 2024-01-04
fix(doc): add streaming parameter to ChatOpenAI

gkhngyk:
- Add the streaming parameter to ChatOpenAI in the doc.
*For agentExecutor.stream() to work properly, the ChatOpenAI streaming parameter must be true.*


gkhngyk:
Hey @bracesproul 
I made a small change in the document and only the commit I changed should appear, but I also see my previously approved commits in the pull request. Will this cause problems in the repo?

jacoblee93:
This should only be necessary for the `streamLog` method unless I'm missing something?

PR #3910, opened 2024-01-04
community[minor]: Cassandra: add MMR; split native and Astra connections

mieslep:
Adds:
* MaximalMarginalRelevance (MMR) to Cassandra

Improves:
* Cleaner distinction between connection to Apache Cassandra versus Astra
* Initialization logic


libs/langchain-community/src/vectorstores/cassandra.ts
```diff
@@ -649,4 +730,113 @@ export class CassandraStore extends VectorStore {
     // Wait for all insert operations to complete.
```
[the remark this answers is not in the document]
mieslep: @jacoblee93 I was not entirely sure how to best handle this tbh. Rather than requiring a user to provide the secure bundle (a .zip file with TLS and other connection-related metadata), this can be downloaded via a service call. I noticed some other uses of axios but none in community. I'm caching this in a tmp subdirectory.

mieslep:
@jacoblee93 I still need to update the message store stuff here, but keen to get your feedback on the handling of the connection bundle above.

mieslep:
@jacoblee93 this is now updated and ready for your full consideration as a PR.

jacoblee93:
Thank you! Will have a look as soon as I can, hopefully today.

libs/langchain-community/src/stores/message/cassandra.ts
```diff
@@ -131,6 +134,8 @@ export class CassandraChatMessageHistory extends BaseListChatMessageHistory {
       return;
     }
 
+    this.client = await CassandraStore.getClient(this.options);
```
jacoblee93, line 137: I think this is fine for now, another option would be to put this under `utils/`

libs/langchain-community/src/vectorstores/cassandra.ts
```diff
@@ -158,20 +181,22 @@ export class CassandraStore extends VectorStore {
...
+   * @param filter Optional filter to be applied as a WHERE clause.
+   * @param includeEmbedding Whether to include the embedding vectors in the results.
    * @returns Promise that resolves with an array of tuples, each containing a Document and a score.
    */
   async similaritySearchVectorWithScore(
     query: number[],
     k: number,
-    filter?: WhereClause
+    filter?: WhereClause,
+    includeEmbedding?: boolean
```
jacoblee93, line 192: Instead of adding another parameter, factor out query logic into a helper function like this:

https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-community/src/vectorstores/supabase.ts#L151

Then we can call that from both this method and the MMR method below

libs/langchain-community/src/vectorstores/cassandra.ts
```diff
@@ -220,6 +251,49 @@ export class CassandraStore extends VectorStore {
...
+   */
+  async maxMarginalRelevanceSearch(
+    query: string,
+    options: MaxMarginalRelevanceSearchOptions<this["FilterType"]>
+  ): Promise<Document[]> {
+    const { k, fetchK = 4 * k, lambda = 0.5, filter } = options;
+
+    const queryEmbedding = await this.embeddings.embedQuery(query);
+
+    const queryResults = await this.similaritySearchVectorWithScore(
```
jacoblee93, line 272: See above comment

libs/langchain-community/src/vectorstores/cassandra.ts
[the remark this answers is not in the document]
jacoblee93: I'm not sure but we should use the native `fetch` object instead of Axios if possible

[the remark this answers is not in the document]
jacoblee93: Ah, sorry missed that this is referencing the below code

libs/langchain-community/src/vectorstores/cassandra.ts
```diff
@@ -649,4 +730,113 @@ export class CassandraStore extends VectorStore {
...
+
+    // Create a copy of args to avoid direct modification
+    const modifiedArgs = { ...args };
+
+    // Initialize cloud if it's not already defined
+    modifiedArgs.cloud = modifiedArgs.cloud || { secureConnectBundle: "" };
+
+    if (!modifiedArgs.cloud.secureConnectBundle) {
+      // Store the secureConnectBundle in a temporary file
+      const dir = path.join(os.tmpdir(), "cassandra-astra");
```
jacoblee93, line 767: This is fine but the issue is that this will make this integration Node only

libs/langchain-community/src/vectorstores/cassandra.ts
```diff
@@ -649,4 +730,113 @@ export class CassandraStore extends VectorStore {
...
+          const bundleURLTemplate =
+            "https://api.astra.datastax.com/v2/databases/{database_id}/secureBundleURL?all=true";
+          const url = bundleURLTemplate.replace(
+            "{database_id}",
+            astraConfig.datacenterID
+          );
+          const postResponse = await fetch(url, {
+            method: "POST",
+            headers: {
+              Authorization: `Bearer ${astraConfig.token}`,
```
jacoblee93, line 793: Meta question: if you can fetch the bundle using an auth token and the bundle doesn't expire, what is the purpose of authenticating with the bundle instead of just skipping a step and using a token directly?

libs/langchain-community/src/vectorstores/cassandra.ts
```diff
@@ -649,4 +730,113 @@ export class CassandraStore extends VectorStore {
     // Wait for all insert operations to complete.
     await Promise.all(insertPromises);
   }
+
+  /**
+   * Method to get the CassandraClient
+   * @param argsWithDefaults connection arguments
+   * @returns CassandraClient object
+   */
+  static async getClient(args: ClientArgs): Promise<CassandraClient> {
```
jacoblee93, line 739: See above, may be worth factoring this into a `utils/cassandra` class

jacoblee93:
Looks solid overall! See comments

libs/langchain-community/src/vectorstores/cassandra.ts
mieslep: The AstraDB mode works in that way - it's a RESTful service. In the Cassandra side (when we connect via the driver), the bundle includes a .json file with some metadata, and also artifacts to support the mTLS connection requirements.

Now, these are signed with a 10 year key...so they do expire. I'll add some logic here to keep the file current.

libs/langchain-community/src/vectorstores/cassandra.ts
mieslep: yes it is already node only.

libs/langchain-community/src/vectorstores/cassandra.ts
mieslep: Hmm. I see what you're saying here, but consider:

1. The query logic itself is in one place; 
2. The MMR function needs to do a similaritySearchVectorWithScore - it just needs the vectors to do its thing; 
3. A caller to similaritySearchVectorWithScore probably does not want the vectors back (why spend time/bandwidth processing things you don't need), but there may be a reason they do, so this gives the caller an option. 

As a new parameter it is optional (for backwards compatibility) and defaults to the prior behaviour... 

That said I can refactor to a separate helper function, but I'd like to keep the new parameter too :)

mieslep:
okay @jacoblee93 comments should be addressed in this latest update!

mieslep:
any idea when this will get merged @jacoblee93 ?

jacoblee93:
So sorry, will make sure it's in today!

libs/langchain-community/jest.config.cjs
```diff
@@ -5,6 +5,7 @@ module.exports = {
   modulePathIgnorePatterns: ["dist/", "docs/"],
   moduleNameMapper: {
     "^(\\.{1,2}/.*)\\.js$": "$1",
+    "^@utils/(.*)$": "<rootDir>/src/utils/$1",
```
jacoblee93, line 8: What is this for?

libs/langchain-community/src/vectorstores/cassandra.ts
jacoblee93: Thanks - messing with the base class signature (even adding more params) has burned us in the past.

libs/langchain-community/src/vectorstores/cassandra.ts
```diff
@@ -220,6 +240,66 @@ export class CassandraStore extends VectorStore {
...
+   * @param k The number of similar vectors to return.
+   * @param filter Optional filter to be applied as a WHERE clause.
+   * @param includeEmbedding Whether to include the embedding vectors in the results.
+   * @returns Promise that resolves with an array of tuples, each containing a Document and a score.
+   */
+  async similaritySearchVectorWithScore(
+    query: number[],
+    k: number,
+    filter?: WhereClause,
+    includeEmbedding?: boolean
```
jacoblee93, line 255: Remove extra param

jacoblee93:
Some questions before merging

libs/langchain-community/jest.config.cjs
mieslep: Coming back to this now. This was added by "yarn build" at the top level - when I remove it the build adds it back in. I'm really not sufficiently familiar with this whole build environment/config to tell you why it is being added?

libs/langchain-community/src/vectorstores/cassandra.ts
mieslep: addressed

mieslep:
Grr I get so annoyed at how rebasing on main creates a mess. lmk if I need to fix this @jacoblee93 ?

mieslep:
I'm done poking at this branch now @jacoblee93 :)

jacoblee93:
Thank you!

mieslep:
Not sure what those build failures are @jacoblee93 , what you did on changes there makes sense and fwiw I did a complete clear-out and rebuild and when I build that line no longer gets added. 

Not entirely sure what change I'd need to do to allow an application to
`import { CassandraClientFactory } from "@langchain/community/utils/cassandra";`

but I'm hoping that this is an artifact of my local setup / knowledge of JS/TS/Node and that once this is merged/released all will be well with the world :)

mieslep:
(actually could those failures be doc-related? not able to get mdx  preview working in the dev container with vs code)

jacoblee93:
Yeah something is funny with our Vercel cache, will rerun on my end

jacoblee93:
> Not sure what those build failures are @jacoblee93 , what you did on changes there makes sense and fwiw I did a complete clear-out and rebuild and when I build that line no longer gets added.
> 
> Not entirely sure what change I'd need to do to allow an application to `import { CassandraClientFactory } from "@langchain/community/utils/cassandra";`
> 
> but I'm hoping that this is an artifact of my local setup / knowledge of JS/TS/Node and that once this is merged/released all will be well with the world :)

You'd want to add a new entrypoint under `create-entrypoints.js` but let's do that in a followup as this has already been open for a while!

mieslep:
Thanks @jacoblee93 . Not finding how to get `create-langchain-integration` working as documented in CONTRIBUTING.md:

```
node ➜ /workspaces/langchainjs/libs/langchain-scripts (cassandraUtils) $ npx create-langchain-integration
sh: 1: create-langchain-integration: not found
node ➜ /workspaces/langchainjs/libs/langchain-scripts (cassandraUtils) $ cd ../
node ➜ /workspaces/langchainjs/libs (cassandraUtils) $ npx create-langchain-integration
sh: 1: create-langchain-integration: not found
node ➜ /workspaces/langchainjs/libs (cassandraUtils) $ cd langchain-community/
node ➜ /workspaces/langchainjs/libs/langchain-community (cassandraUtils) $ npx create-langchain-integration
sh: 1: create-langchain-integration: not found
node ➜ /workspaces/langchainjs/libs/langchain-community (cassandraUtils) $ cd ../../
node ➜ /workspaces/langchainjs (cassandraUtils) $ npx create-langchain-integration
sh: 1: create-langchain-integration: not found
node ➜ /workspaces/langchainjs (cassandraUtils) $ 
```

PR #3917, opened 2024-01-04
docs[patch]: Added tooltips to mdx files with `@langchain/core` in code blocks

bracesproul:


jacoblee93:
Going to close this for now, I don't think we want this.

PR #3919, opened 2024-01-05
azure-openai[major]: Integration of Azure OpenAI SDK with Langchain

sarangan12:
I am a part of [Azure JS SDK Team](https://github.com/Azure/azure-sdk-for-js). Our repository provides several SDKs which provide a standard way to communicate with the Azure Services. One such SDK is [Azure OpenAI SDK](https://github.com/Azure/azure-sdk-for-js/tree/main/sdk/openai/openai).

Currently, in Langchain SDK, users communicate with Azure OpenAI endpoints through the [`AzureOpenAI`](https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-openai/src/azure/llms.ts#L10C14-L10C26) class. But, in essence it uses the same OpenAI SDK as can be seen [here](https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-openai/src/llms.ts#L501C39-L501C50). It would be great if we could integrate the Azure OpenAI JS SDK with Langchain. This PR is created to introduce Azure OpenAI JS SDK within Langchain.

Please note, that this does not impact any existing functionality within Langchain and currently is a stand alone piece. We will keep updating and enhancing this. Please review and approve the same. Thanks


libs/langchain-openai/src/azure/chat_models_sdk.ts
```diff
@@ -0,0 +1,391 @@
...
+              name: "disableKeepAlive",
+              sendRequest: (request, next) => {
+                request.disableKeepAlive = true;
+                return next(request);
+              },
+            },
+            position: "perCall",
+          },
+        ],
+      }
```
deyaaeldeen, line 223: Why is this needed?

libs/langchain-openai/src/azure/llms_sdk.ts
```diff
@@ -0,0 +1,414 @@
...
+              name: "disableKeepAlive",
+              sendRequest: (request, next) => {
+                request.disableKeepAlive = true;
+                return next(request);
+              },
+            },
+            position: "perCall",
+          },
+        ],
+      }
```
deyaaeldeen, line 181: same question

libs/langchain-openai/src/azure/chat_models_sdk.ts
sarangan12: For this issue - https://github.com/Azure/azure-sdk-for-js/issues/27959

libs/langchain-openai/src/azure/llms_sdk.ts
sarangan12: Same as the previous one

jacoblee93:
This would probably be best as a separate standalone package? Like `@langchain/azureopenai`?

I'm happy to do the refactoring!

sarangan12:
> This would probably be best as a separate standalone package? Like `@langchain/azureopenai`?
> 
> I'm happy to do the refactoring!

Thanks @jacoblee93. I am working on splitting the package. I will update this PR once the splitting is done

sarangan12:
@jacoblee93 I have split the code into a separate package. Could you please review the code? Thanks

langchain-core/src/utils/stream.ts
```diff
@@ -102,6 +102,9 @@ export class IterableReadableStream<T>
         // Fix: `else if (value)` will hang the streaming when nullish value (e.g. empty string) is pulled
         controller.enqueue(value);
       },
+      async cancel() {
```
jacoblee93, line 105: Note to self: ship a new version of core when this is merged

langchain/scripts/create-entrypoints.js
```diff
@@ -76,6 +76,7 @@ const entrypoints = {
   "llms/load": "llms/load",
   "llms/base": "llms/base",
   "llms/openai": "llms/openai",
+  "llms/azure_openai": "llms/azure_openai",
```
jacoblee93, line 79: Let's not export this from `langchain` - we're deprecating integration exports from the main package in favor of the integrations packages. Can you revert all the LangChain changes?

langchain/src/llms/azure_openai-chat.ts
```diff
@@ -0,0 +1,92 @@
...
+import { CallbackManagerForLLMRun } from "../callbacks/manager.js";
+import type { Generation, LLMResult } from "../schema/index.js";
+import { promptLayerTrackRequest } from "../util/prompt-layer.js";
+
+export { AzureOpenAIChat };
+
+/**
+ * PromptLayer wrapper to OpenAIChat
+ */
+export class PromptLayerAzureOpenAIChat extends AzureOpenAIChat {
```
jacoblee93, line 12: We can remove this

libs/langchain-azure-openai/src/chat_models.ts
```diff
@@ -0,0 +1,687 @@
...
+      modelName: "model",
+      azureOpenAIApiKey: "azure_openai_api_key",
+      azureOpenAIEndpoint: "azure_openai_api_endpoint",
+      azureOpenAIApiDeploymentName: "azure_openai_api_deployment_name",
+    };
+  }
+
+  lc_serializable = true;
+
+  functions?: FunctionDefinition[] | undefined;
```
jacoblee93, line 207: Can you use the function calling types here?

https://github.com/langchain-ai/langchainjs/blob/main/langchain-core/src/language_models/base.ts#L198

They should be equivalent.

libs/langchain-azure-openai/src/chat_models.ts
```diff
@@ -0,0 +1,687 @@
...
+    this.temperature = fields?.temperature ?? this.temperature;
+    this.topP = fields?.topP ?? this.topP;
+    this.frequencyPenalty = fields?.frequencyPenalty ?? this.frequencyPenalty;
+    this.presencePenalty = fields?.presencePenalty ?? this.presencePenalty;
+    this.maxTokens = fields?.maxTokens;
+    this.n = fields?.n ?? this.n;
+    this.logitBias = fields?.logitBias;
+    this.stop = fields?.stop;
+    this.user = fields?.user;
+    this.functions = fields?.functions;
```
jacoblee93, line 292: We've been binding these at runtime like this:

https://js.langchain.com/docs/modules/model_io/chat/function_calling

Would prefer to keep it that way and standardize around the raw `function_call` to avoid confusion.

libs/langchain-azure-openai/src/embeddings.ts
```diff
@@ -0,0 +1,140 @@
...
+  isTokenCredential,
+} from "@azure/core-auth";
+import { chunkArray } from "@langchain/core/utils/chunk_array";
+import { AzureOpenAIInput, AzureOpenAIEmbeddingsParams } from "./types.js";
+
+export class AzureOpenAIEmbeddings
+  extends Embeddings
+  implements AzureOpenAIEmbeddingsParams, AzureOpenAIInput
+{
+  modelName = "text-embedding-ada-002";
```
jacoblee93, line 20: Should we add a `dimensions` property here too for v3?

libs/langchain-azure-openai/src/legacy.ts
```diff
@@ -0,0 +1,355 @@
+import {
```
jacoblee93, line 1: Let's leave this out! Unless you have a reason to keep it?

jacoblee93:
Looks solid overall - see comments!

langchain/scripts/create-entrypoints.js
sarangan12: Per the comments, **ALL** the changes from the langchain package have been reverted back in the latest commit.

langchain/src/llms/azure_openai-chat.ts
sarangan12: Per the previous comment, I have removed **ALL** the changes from the langchain package. So, this file has also been removed as part of it.

libs/langchain-azure-openai/src/embeddings.ts
sarangan12: We could add the `dimensions` property as an update later after the initial commit. I would like to do the required testing around it and ensure that no breaks in the code. Also, I will be contributing on a consistent basis to this repository. So, we could add this later

libs/langchain-azure-openai/README.md
```diff
@@ -0,0 +1,144 @@
...
+```typescript
+import { AzureChatOpenAI } from "@langchain/azure-openai";
+
+const model = new AzureChatOpenAI({
+  modelName: "gpt-4-1106-preview",
+  // Note that the following are optional, and will default to the values below
+  // if not provided.
+  azureOpenAIEndpoint: process.env.AZURE_OPENAI_API_ENDPOINT,
+  azureOpenAIApiKey: process.env.AZURE_OPENAI_API_KEY,
+  azureOpenAIApiDeploymentName: process.env.AZURE_OPENAI_API_DEPLOYMENT_NAME,
```
deyaaeldeen, line 64: I wonder why do we need to provide a model name? Azure exposes a specific model through a specific deployment, should this be renamed to deployment name?

libs/langchain-azure-openai/README.md
```diff
@@ -0,0 +1,144 @@
...
+    }
+  }
+}
+```
+
+The field you need depends on the package manager you're using, but we recommend adding a field for the common `yarn`, `npm`, and `pnpm` to maximize compatibility.
+
+## Chat Models
+
+This package contains the `AzureChatOpenAI` class, which is the recommended way to interface with the OpenAI series of models.
```
deyaaeldeen, line 43: I think we should highlight that this is for Azure:
```suggestion
This package contains the `AzureChatOpenAI` class, which is the recommended way to interface with deployed models on Azure OpenAI.
```

libs/langchain-azure-openai/README.md
```diff
@@ -0,0 +1,144 @@
...
+  azureOpenAIEndpoint: process.env.AZURE_OPENAI_API_ENDPOINT,
+  azureOpenAIApiKey: process.env.AZURE_OPENAI_API_KEY,
+  azureOpenAIApiDeploymentName: process.env.AZURE_OPENAI_API_DEPLOYMENT_NAME,
+});
+const response = await model.stream(new HumanMessage("Hello world!"));
+```
+
+## Embeddings
+
+This package also adds support for OpenAI's embeddings model.
```
deyaaeldeen, line 87:
```suggestion
This package also supports embeddings with Azure OpenAI.
```

libs/langchain-azure-openai/package.json
```diff
@@ -0,0 +1,98 @@
+{
+  "name": "@langchain/azure-openai",
+  "version": "0.0.1",
```
deyaaeldeen, line 3: Just double checking, is this the intended version?

libs/langchain-azure-openai/package.json
```diff
@@ -0,0 +1,98 @@
...
+    "test:single": "yarn run build:deps && NODE_OPTIONS=--experimental-vm-modules yarn run jest --config jest.config.cjs --testTimeout 100000",
+    "test:int": "NODE_OPTIONS=--experimental-vm-modules jest --testPathPattern=\\.int\\.test.ts --testTimeout 100000 --maxWorkers=50%",
+    "format": "prettier --write \"src\"",
+    "format:check": "prettier --check \"src\""
+  },
+  "author": "LangChain",
+  "license": "MIT",
+  "dependencies": {
+    "@azure/core-auth": "^1.5.0",
+    "@azure/openai": "^1.0.0-beta.11",
```
deyaaeldeen, line 39: The library is still in beta and subsequent beta versions may contain breaking changes, so I think we should pin for now:
```suggestion
    "@azure/openai": "1.0.0-beta.11",
```

libs/langchain-azure-openai/src/chat_models.ts
sarangan12: Per the comments, this has been removed and tested.

libs/langchain-azure-openai/src/chat_models.ts
sarangan12: Yes. They are interoperable. But this section has been removed based on the PR Comment: https://github.com/langchain-ai/langchainjs/pull/3919/#discussion_r1468116004

docs/core_docs/docs/integrations/text_embedding/azure_openai.mdx
```diff
@@ -4,6 +4,74 @@ keywords: [AzureOpenAIEmbeddings]
...
+You'll also need to have an Azure OpenAI instance deployed. You can deploy a version on Azure Portal following [this guide](https://learn.microsoft.com/azure/ai-services/openai/how-to/create-resource?pivots=web-portal).
+
+Once you have your instance running, make sure you have the endpoint and key. You can find them in the Azure Portal, under the "Keys and Endpoint" section of your instance.
+
+You can then define the following environment variables to use the service:
+
+```bash
+AZURE_OPENAI_API_ENDPOINT=<YOUR_ENDPOINT>
+AZURE_OPENAI_API_KEY=<YOUR_KEY>
+AZURE_OPENAI_API_EMBEDDING_DEPLOYMENT_NAME=<YOUR_EMBEDDING_DEPLOYMENT_NAME>
```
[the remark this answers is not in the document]
sarangan12: It is already supported. Only for the Embeddings, we have 2 variables for deployment name. Because, some models support embeddings and some do not. Also, this is to keep consistent with the langchain-openai package

libs/langchain-azure-openai/README.md
sarangan12: The model name is not used by @azure/openai package. But, this is needed for internal functionalities of langchain. Also, this is consistent with the langchain/openai package

libs/langchain-azure-openai/README.md
sarangan12: Done

libs/langchain-azure-openai/README.md
sarangan12: Done

libs/langchain-azure-openai/package.json
sarangan12: Yes. This is consistent with other lib integration packages

libs/langchain-azure-openai/package.json
sarangan12: Done

libs/langchain-azure-openai/src/legacy.ts
sarangan12: @jacoblee93  I would like this to continue for now, because I want to keep it consistent with @langchain/openai package. When we retire the legacy class in that package, we could remove it from here also. Would that be ok? But, if you really want to drop this out, I can remove it. Please suggest.

sarangan12:
@jacoblee93 Thanks for your reviews. I have addressed all the comments and updated the PR. It would be great if you could review and approve the PR?

libs/langchain-azure-openai/src/legacy.ts
jacoblee93: Let's cut it - we want to steer people away from it. It's already been dropped from the docs and marked as deprecated in the other package.

sarangan12:
@jacoblee93 A gentle reminder on this PR. Could you please review and merge this PR? TIA

jacoblee93:
Looking now!

jacoblee93:
Hey @sarangan12, it looks great! I made a few updates - all the code changes are isolated in this commit:

https://github.com/langchain-ai/langchainjs/pull/3919/commits/31da8c581aa00dc7c8095d103ae185e7b05c888e

It moves everything to use our built-in `AsyncCaller` class for retries. Can you LMK if integration tests pass for you? I seem to be having a bit of trouble with my Azure keys.

jacoblee93:
I actually think the error handling might be a little bit wonky as well - would you be able to test that easily as well?

sarangan12:
@jacoblee93 Thanks a lot. YES. I have tested the package with your commit. All the tests have passed and there are no issues.

sarangan12:
@jacoblee93 Thanks a lot. I have tested the package with your commit. All the tests have passed and there are no issues with success cases. Yes. As you have mentioned the error scenarios have become wonky. For example, if there is a missing deployment or invalid endpoint, instead of reporting that, it is always reporting as timeout error. I will try to debug and fix the issue. 

As you are the expert in the Langchain/AsyncCaller Code, it would be great if you also could take a look at it, Thanks

sarangan12:
@jacoblee93 With the latest commit, all test cases have passed and the error scenarios have also been verified. Please take a look.

jacoblee93:
@sarangan12 I'm fine to merge this and deploy as is - the issue is the errors thrown by the SDK don't seem to contain any additional information outside of the error message.

The way the built-in `AsyncCaller` class works is by retrying certain status codes on errors:

https://github.com/langchain-ai/langchainjs/blob/main/langchain-core/src/utils/async_caller.ts#L18

Would it be possible to pass the raw status code through? We could alternatively initialize the caller with a specialized error handler but it would be nice to retry things correctly by default.

sarangan12:
@jacoblee93 Thanks for the comment. I will work with my team to get more details on to the error object and easy way to access it. Once it has been released, I will update the langchain codebase too. 

In the mean time, we can merge and deploy this. I can add the error addition as the next feature release. Thanks

langchain-core/src/utils/async_caller.ts
```diff
@@ -115,7 +115,9 @@ export class AsyncCaller {
               if (error instanceof Error) {
                 throw error;
               } else {
-                throw new Error(error);
+                const err = new Error(error.message);
```
jacoblee93, line 118: We should not be handling this here

jacoblee93:
Thank you!

jacoblee93:
This is live now:

https://www.npmjs.com/package/@langchain/azure-openai

Thank you for your patience!

TheShoes:
How would i pass custom headers?

also when i give it a deployment it complains about not having one till i put in a embeddingsdeployment.

I have the whole process working in python but wanted to try it in js

PR #3928, opened 2024-01-05
langchain[minor]: Adds create runnable chain functions

bracesproul:


docs/core_docs/docs/modules/agents/how_to/openai_functions.mdx
```diff
@@ -0,0 +1,44 @@
+# Using OpenAI functions
```
jacoblee93, line 1: Why is this under `agents`?

jacoblee93: I think we can hide this page for now and add an entry under the `chains` table

examples/src/chains/openai_functions_json_schema.ts
```diff
@@ -0,0 +1,41 @@
...
+};
+
+const model = new ChatOpenAI();
+const prompt = ChatPromptTemplate.fromMessages([
+  ["human", "Human description: {description}"],
+]);
+const outputParser = new JsonOutputFunctionsParser();
+
+const runnable = createStructuredOutputRunnable(
+  JSONSchema,
```
jacoblee93, line 29: Can we call just `schema`?

PR #3929, opened 2024-01-06
community[minor]: feat: Support batching and deletes for AstraDB Vector Store

mfortman11:
Improves the AstraDBVectorStore by adding support for batching and deletes.

(I should have included this when adding the vector store by my Twitter/X handle is @MikeFortman11)


libs/langchain-community/src/vectorstores/astradb.ts
```diff
@@ -40,18 +49,35 @@ export class AstraDBVectorStore extends VectorStore {
...
+    const {
+      token,
+      endpoint,
+      collection,
+      collectionOptions,
+      idKey,
+      contentKey,
+      batchSize,
+      ...callerArgs
+    } = args;
```
bracesproul, line 72: This is missing `namespace` from `AstraLibArgs`, don't want that getting included in `callerArgs`.

Side note, it doesn't look like this was implemented anywhere previously. Is it needed?

libs/langchain-community/src/vectorstores/astradb.ts
```diff
@@ -40,18 +49,35 @@ export class AstraDBVectorStore extends VectorStore {
 
   private readonly contentKey: string; // if undefined the entirety of the content aside from the id and embedding will be stored as content
 
+  private readonly batchSize: number;
```
bracesproul, line 52: Should set a `batchSize` default since it's optional. Maybe `250`?
```suggestion
  private readonly batchSize = 250;
```

libs/langchain-community/src/vectorstores/astradb.ts
```diff
@@ -98,7 +124,12 @@ export class AstraDBVectorStore extends VectorStore {
       ...documents[idx].metadata,
     }));
 
-    await this.collection.insertMany(docs);
+    const chunkedDocs = chunkArray(docs, this.batchSize);
+    const batchCalls = chunkedDocs.map((chunk) =>
+      this.collection?.insertMany(chunk)
+    );
```
bracesproul, line 130: `this.caller` should wrap the `this.collection?.insertMany` call. See [this file](https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-openai/src/embeddings.ts#L258) for an example

bracesproul (dismissed):
Left a couple comments, mainly need the `this.caller` to wrap any API calls.
This is a very good improvement though, thanks for adding this!

libs/langchain-community/src/vectorstores/astradb.ts
mfortman11: I removed namespace for now since it is unused but will be adding it in the future when the new version of TS client is ready.

Batching wasn't implemented before but it is needed as the `insertMany` has a cap of 20 documents

libs/langchain-community/src/vectorstores/astradb.ts
mfortman11: default is set in the constructor. 20 is the max supported by `insertMax`
```javascript
this.batchSize = batchSize && batchSize <= 20 ? batchSize : 20;
```

libs/langchain-community/src/vectorstores/astradb.ts
mfortman11: My bad on this one thanks for the catch

mfortman11:
@bracesproul thanks for the review all comments have been addressed

jacoblee93:
Thank you!

PR #3930, opened 2024-01-06
langchain[patch][tests]: Added tests for new create agent functions

bracesproul:


langchain/src/agents/tests/create_agent_functions.test.ts
```diff
@@ -0,0 +1,98 @@
...
+  createReactAgent,
+  createStructuredChatAgent,
+  createXmlAgent,
+} from "../index.js";
+import { ChatAnthropic } from "../../chat_models/anthropic.js";
+import { Calculator } from "../../tools/calculator.js";
+
+const tools = [new Calculator()];
+
+test("createStructuredChatAgent works", async () => {
```
jacoblee93, line 19: This is technically an integration test as it uses outside resources (hub, OpenAI) - I think fine to just show that you can initialize it at all with an arbitrary prompt and a FakeLLM.

PR #3932, opened 2024-01-06
community[patch]: Add streaming for HuggingFace Inference (also add the support of stop-sequences)

ljhskyso:
Fixes # (issue)


libs/langchain-community/src/llms/hf.ts
```diff
@@ -20,6 +22,11 @@ export interface HFInput {
    */
   maxTokens?: number;
 
+  /**
+   * The model will stop generating text when one of the strings in the list is generated.
+   */
+  stopSequences?: string[];
```
bracesproul, line 28: The `stopSequences` input API has been deprecated in favor of the LCEL syntax: `model.bind({ stop: ... })`. Please verify that functionality doesn't already exist, and if not, could you implement it that way.

libs/langchain-community/src/llms/hf.ts
```diff
@@ -107,40 +117,72 @@ export class HuggingFaceInference extends LLM implements HFInput {
...
 
+  invocationParams(_options?: this["ParsedCallOptions"]) {
+    return {
+      model: this.model,
+      parameters: {
+        // make it behave similar to openai, returning only the generated text
+        return_full_text: false,
+        temperature: this.temperature,
+        max_new_tokens: this.maxTokens,
+        stop_sequences: this.stopSequences,
```
bracesproul, line 128: Since we're passing `options` through, we should prioritize the passed option, and fallback to the class property.
This should apply to the rest of the `parameters` too.
```typescript
options.stop ?? this.stop
```

libs/langchain-community/src/llms/tests/huggingface_hub.int.test.ts
```diff
@@ -6,3 +6,13 @@ test("Test HuggingFace", async () => {
   const res = await model.call("1 + 1 =");
   console.log(res);
 }, 50000);
+
+test("Test HuggingFace with streaming", async () => {
```
bracesproul, line 10: if you could also add a test for stop sequences that would be great!!

bracesproul (changes requested):
Largely looks good to me, just have one comment about stop sequences API, and this should have a section [in the docs](https://js.langchain.com/docs/integrations/llms/huggingface_inference). Doesn't need to be complex, but should show off that it's possible.

thank you for doing this!!

jacoblee93:
I think the param is actually called `stop`?

jacoblee93:
It's hard to pin down but I believe the parameter is indeed `stop`:

https://huggingface.co/docs/api-inference/detailed_parameters#question-answering-task

jacoblee93:
Thank you!

PR #3943, opened 2024-01-07
core[patch]: fix: update dereference.ts to match upstream

glorat:
dereference.ts is taken from an upstream repo https://raw.githubusercontent.com/cfworker/cfworker/main/packages/json-schema/src/dereference.ts

which contains a fix for when self.location.origin is null (which can happen if being used in a nodejs environment). This PR ports over the fix in line with upstream

### DIFF
--- a/langchain-core/src/utils/@cfworker/json-schema/src/dereference.ts
+++ b/langchain-core/src/utils/@cfworker/json-schema/src/dereference.ts
@@ -67,12 +67,12 @@ export const ignoredKeyword: Record<string, boolean> = {
  */
 export let initialBaseURI =
   // @ts-ignore
-  typeof self !== "undefined" && self.location
+  typeof self !== 'undefined' &&
+  self.location &&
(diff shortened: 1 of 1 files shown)


jacoblee93:
Thanks for this!

glorat:
Since I didn't post an original issue for this (although happy to do so if requested), the bug triggers if running unit tests in happy-dom or jsdom (I'm not sure which is default for vitest) such that the simulated browser environment doesn't fully populate self.location, causing this code to crash on initialisation.

In absence of this patch, the workaround is to set `// @vitest-environment node` on the test file, assuming no browser simulation is required for the test.

jacoblee93:
Odd that it literally resolves to the string `"null"` but looks good to me!

glorat:
> Odd that it literally resolves to the string `"null"` but looks good to me!

I thought that too but in my test in the error environment, that's literally what was returned!

PR #3944, opened 2024-01-07
Add Graph API

gkorland:


gkorland:
@jacoblee93

hntrl:
This is obviously a very old PR, but I don't think at this point we need to make a case for a common graph abstraction. Lmk if you think otherwise.

PR #3948, opened 2024-01-08
Match Python serialization for JSON output OAI parsers

dqbd:


bracesproul (commented):
I believe it was [decided in slack](https://langchain.slack.com/archives/C052N6J5MJ5/p1704730095585859?thread_ts=1704729244.100869&cid=C052N6J5MJ5) it's okay to leave as is, so we don't need to make a breaking change?

langchain/src/output_parsers/openai_tools.ts
```diff
@@ -2,9 +2,16 @@ import { BaseLLMOutputParser } from "@langchain/core/output_parsers";
...
 
 export type ParsedToolCall = {
-  name: string;
+  type: string;
+
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  arguments: Record<string, any>;
+  args: Record<string, any>;
+
+  /** @deprecated Use `type` instead */
```
bracesproul, line 10:
```suggestion
  /** @deprecated Use `type` instead. Will be removed in 0.2.0. */
```

langchain/src/output_parsers/openai_tools.ts
```diff
@@ -2,9 +2,16 @@ import { BaseLLMOutputParser } from "@langchain/core/output_parsers";
...
+  name: string;
+
+  /** @deprecated Use `args` instead */
```
bracesproul, line 13:
```suggestion
  /** @deprecated Use `args` instead. Will be removed in 0.2.0. */
```

jacoblee93:
Thank you!

PR #3950, opened 2024-01-08
Add the Milvus partition feature

Mista-Shelby:
Hi !

Milvus has a great partition feature but it can't be used with langchainjs : this is my shot at adding the milvus partition support !


libs/langchain-community/src/vectorstores/milvus.ts
```diff
@@ -344,6 +350,29 @@ export class Milvus extends VectorStore {
...
+  /**
+   * Ensures that a partition exists in the Milvus collection.
+   * @returns Promise resolving to void.
+   */
+  async ensurePartition() {
+    const hasPartResp = await this.client.hasPartition({
+      collection_name: this.collectionName,
+      partition_name: this.partitionName,
+    });
+    if (hasPartResp.status.error_code !== ErrorCode.SUCCESS) {
```
bracesproul, line 362: Do we need a check to only run this if `this.partitionName` is truthy? Couldn't exactly tell from [their docs](https://milvus.io/docs/create_partition.md) on this so I want to make sure this wont' start throwing for everyone who uses Milvus without passing a `partitionName` through.

bracesproul (commented):
One quick question, besides that everything lgtm!

bracesproul:
@Mista-Shelby looks like format is failing, if you could run `yarn format` from root and push up whatever changes it should resolve this.

libs/langchain-community/src/vectorstores/milvus.ts
Mista-Shelby: Hi @bracesproul !
Thanks for your review. `this.partition_name` has a default value `'Default partition'` (line 128) which is the name of the default partition generated by Milvus for every collection... as of today :) Do you think we need a more future-proof check ?
Thanks !

Mista-Shelby:
> @Mista-Shelby looks like format is failing, if you could run `yarn format` from root and push up whatever changes it should resolve this.

Will do, sorry for any inconvenience !

libs/langchain-community/src/vectorstores/milvus.ts
jacoblee93: It looks like the default value is `_default`: https://milvus.io/docs/create_partition.md

But yeah I am not sure I'd rely on that - could we make it opt-in only?

jacoblee93:
Let's try to preserve the non-partition behavior if a partitionName is not passed if possible?

Mista-Shelby:
> Let's try to preserve the non-partition behavior if a partitionName is not passed if possible?

Of course, I'll take a look at it.

jacoblee93:
Thank you for your patience! A test would be nice, but this shouldn't break anyone existing.

PR #3958, opened 2024-01-08
docs[patch]: Guide for debugging

bracesproul:


jacoblee93:
Thank you!

Would still love to cut down on `verbose: true` at some point - maybe have levels in the future.

PR #3960, opened 2024-01-08
Initial work implementing Gemini LLM and chat models for both Google AI Studio and Vertex AI

### LINKED ISSUE #3172: Feature Request: Version number available via function call
It would be useful if there was a consistent way to get the currently running version of the library from within library code.

There is already a way to do this in the Python version.

Use case:
I'd like to include library version when making API calls to Google endpoints. This will allow for metrics to better understand how much the library is used and how well people keep updated with the latest version of the library.

cc: @jacoblee93

### LINKED ISSUE #3884: Acces Gemini using Google Vertex
Is it possible to access Gemini using Google Vertex ? I saw it will be implemented (https://github.com/langchain-ai/langchainjs/issues/2562#issuecomment-1855013347) but not sure it has been done.

afirstenberg:
Initial work implementing Gemini LLM and chat models for both Google AI Studio and Vertex AI using API Keys, Web authentication, and Google authentication.

This establishes three new packages:
- @langchain/google-common : the bulk of the implementation, common types, base classes, etc
- @langchain/google-gauth : Subclasses that work with Google authentication
- @langchain/google-webauth : Subclasses that work with web authentication

This also lays the foundation for future work, including migrating PaLM backwards compatibility into these packages.

I think I got the packages setup correctly, but I could be wrong.

Depends on #3797 

Fixes #3884 
Fixes #3172


bracesproul (commented):
Looks like build is failing because it needs to run `yarn install`. If you could run that from root and push up the changes to the `yarn.lock` file that would be great

afirstenberg:
> Looks like build is failing because it needs to run `yarn install`. If you could run that from root and push up the changes to the `yarn.lock` file that would be great

Well... it is also failing because it assumes the changes that are in #3797. So the `yarn build` failed.

axe-me:
I wonder when this could be merged in. Waiting for this Gemini integration.

aaronzhongg:
Also interested, thanks for the work!

duckdegen:
+1 in requesting this, we have tons of workarounds in our codebase which make using the vertex AI in conjunction with langchain really hard <3

Is there anything we can do to unstick this? 

@afirstenberg feel free to ask for help

afirstenberg:
Thanks for the offer @duckdegen! (And for the interest from everyone else.)

Right now it is blocked on #3797 being merged, which is a refactor of one of the data types, and then we can make sure all the checks for this one pass. (And then it boils down to the LangChain team having time to review a massive code addition - but it should be easier since it will all be in three isolated packages.)

Once it is in place, we should be able to start addressing all the things in the TODO list, and hopefully each of those will go faster.

Is there anything in particular that you need from the module?

duckdegen:
Specifically what i'm dependent on, is that it uses whatever endpoints are exposed by this module/package:

`import {VertexAI} from '@google-cloud/vertexai';`

This is because the googlegenerativeAi API is geofenced and does not support the EU and some other big countries.

afirstenberg:
@duckdegen - Definitely.
As long as you _don't_ set an API key, it will default to the Vertex AI implementation for the "gemini" models.
If you _do_ have an API key set, it will use the Google AI Studio / Makersuite implementation for the "gemini" models.

(The rest... is a long story which I plan to post an article about. Because Google didn't make it easy.)

jacoblee93:
Thank you for your patience!

PR #3969, opened 2024-01-09
Fixed parsing of tool_input and final_answer tags

jmlon:
When parsing the inputs and outputs from a tool, it is not properly renoving the end xml tag. For example for the input, the toolInput includes the </tool_input> which causes problems when passing this argument to the tool.

{
  tool: 'getLengthOfString',
  toolInput: 'This is a Happy String!**</tool_input>**',
  log: '<tool>getLengthOfString</tool><tool_input>This is a Happy String!</tool_input>'
}

Similarly, for the final_answer tag, it returns the end tags as well, for example:

{
  returnValues: {
    output: 'The length of the string "Parangaricutirimícuaro" is 22.**</final_answer>**'
  },
  log: '<final_answer>The length of the string "Parangaricutirimícuaro" is 22.</final_answer>'
}

I have change the code to use a regular expression and properly extract the text between the XML elements.

### DIFF
--- a/langchain/src/agents/xml/output_parser.ts
+++ b/langchain/src/agents/xml/output_parser.ts
@@ -38,12 +38,11 @@ export class XMLAgentOutputParser extends AgentActionOutputParser {
    */
   async parse(text: string): Promise<AgentAction | AgentFinish> {
     if (text.includes("</tool>")) {
-      const [tool, toolInput] = text.split("</tool>");
-      const _tool = tool.split("<tool>")[1];
-      const _toolInput = toolInput.split("<tool_input>")[1];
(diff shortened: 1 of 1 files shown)


jacoblee93:
The stop sequence should take care of this, no?

Still worthwhile I think, thank you!

jmlon:
> The stop sequence should take care of this, no?
> 
> Still worthwhile I think, thank you!

Actually I ended up here, because even with the stop bindings, I still got the closing tags in the answers and they were causing a lot of problems later on the execution of the agent.

jacoblee93:
Closing in favor of #3975

PR #3981, opened 2024-01-10
Update openai to ^4.24.2 and include logprobs in `generationInfo`

functorism:
Update `openai` to `^4.24.2` and include logprobs in `generationInfo` in `ChatOpenAI` `_generate`

### DIFF
--- a/langchain-core/src/language_models/base.ts
+++ b/langchain-core/src/language_models/base.ts
@@ -216,7 +216,7 @@ export interface FunctionDefinition {
    * To describe a function that accepts no parameters, provide the value
    * `{"type": "object", "properties": {}}`.
    */
-  parameters?: Record<string, unknown>;
+  parameters: Record<string, unknown>;
 
   /**
    * A description of what the function does, used by the model to choose when and
--- a/langchain-core/src/language_models/base.ts
+++ b/langchain-core/src/language_models/base.ts
@@ -216,7 +216,7 @@ export interface FunctionDefinition {
    * To describe a function that accepts no parameters, provide the value
    * `{"type": "object", "properties": {}}`.
    */
-  parameters: Record<string, unknown>;
+  parameters?: Record<string, unknown>;
 
   /**
    * A description of what the function does, used by the model to choose when and
--- a/libs/langchain-openai/src/chat_models.ts
+++ b/libs/langchain-openai/src/chat_models.ts
@@ -596,9 +596,10 @@ export class ChatOpenAI<
             part.message ?? { role: "assistant" }
           ),
         };
-        if (part.finish_reason) {
-          generation.generationInfo = { finish_reason: part.finish_reason };
-        }
+        generation.generationInfo = {
+          ...(part.finish_reason ? { finishReason: part.finish_reason } : {}),
+          ...(part.logprobs ? { logprobs: part.logprobs } : {}),
+        };
         generations.push(generation);
       }
       return {
--- a/libs/langchain-openai/src/chat_models.ts
+++ b/libs/langchain-openai/src/chat_models.ts
@@ -597,7 +597,7 @@ export class ChatOpenAI<
           ),
         };
         generation.generationInfo = {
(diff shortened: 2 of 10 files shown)


libs/langchain-openai/src/utils/azure.ts
```diff
@@ -3,7 +3,7 @@ export interface OpenAIEndpointConfig {
   azureOpenAIApiInstanceName?: string;
   azureOpenAIApiKey?: string;
   azureOpenAIBasePath?: string;
-  baseURL?: string;
+  baseURL?: string | null;
```
jacoblee93, line 6: This doesn't look necessary?

libs/langchain-openai/src/chat_models.ts
```diff
@@ -596,9 +596,10 @@ export class ChatOpenAI<
             part.message ?? { role: "assistant" }
           ),
         };
-        if (part.finish_reason) {
-          generation.generationInfo = { finish_reason: part.finish_reason };
-        }
+        generation.generationInfo = {
+          ...(part.finish_reason ? { finishReason: part.finish_reason } : {}),
```
jacoblee93, line 600: Should be `finish_reason` in the `generationInfo` as well

libs/langchain-openai/src/utils/azure.ts
functorism: It's because of this change: https://github.com/openai/openai-node/commit/eb82824f89c31d752cdb4f198716a199e3b6eafd#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80R26

But maybe it should be dealt with differently.

langchain-core/src/language_models/base.ts
```diff
@@ -216,7 +216,7 @@ export interface FunctionDefinition {
    * To describe a function that accepts no parameters, provide the value
    * `{"type": "object", "properties": {}}`.
    */
-  parameters: Record<string, unknown>;
+  parameters?: Record<string, unknown>;
```
jacoblee93, line 219: Core changes should be done in a separate PR

jacoblee93 (commented):
Thanks so much for this!

langchain-core/src/language_models/base.ts
functorism, line 219: Likewise this change is because of: https://github.com/openai/openai-node/commit/c4fadb4513fc2e35cf0e4b9512ec39a6bd0e42dd#diff-30f3fcee562aad291d98f76d670f7cef5920b2ad946be42c3f6d02297767ec2cR772-R774

libs/langchain-openai/src/chat_models.ts
functorism: Fixed :+1:

langchain-core/src/language_models/base.ts
functorism: Should I open a new PR with that singular change that gets merged before this one?

functorism:
@jacoblee93 The api-refs build fine in the dev container now, a bit unclear to me why the deployment fails. I can't see any details.

langchain-core/src/language_models/base.ts
jacoblee93: Yes, if you wouldn't mind that would be really helpful!

jacoblee93:
> @jacoblee93 The api-refs build fine in the dev container now, a bit unclear to me why the deployment fails. I can't see any details.

CI failure appears to be a flake, only blocker would be splitting the core change out separately (we will need to release a new version of core ahead of this). I can handle later today if you are busy!

functorism:
@jacoblee93 If it's easy for you to slip such a thing in then that would be awesome. Much appreciated.

jacoblee93:
Let's not change the internal types for now - I updated the OpenAI package to use the function definition type from `@langchain/core`. This is great though, will merge and cut a new release of `@langchain/openai`.

jacoblee93:
Thank you! It's live in the latest version of `@langchain/openai`

PR #3982, opened 2024-01-10
cloudflare[major]: Add integration package, fix Vectorize for latest workers

kristianfreeman:
This PR fixes an issue where the Vectorize API has been [updated internally](https://developers.cloudflare.com/vectorize/changelog/#2023-11-08) and requires some changes to the query logic inside the Vectorize integration for Langchain.

For Workers projects created prior to this PR being merged, users will either need to [update compatibility dates/flags](https://developers.cloudflare.com/workers/configuration/compatibility-dates/#vectorize-query-with-metadata-optionally-returned) in their `wrangler.toml` or update the version of Langchain they're using. @jacoblee93 How would you recommend I document that? I can add something in this PR or in a follow up (e.g. to say "update to Langchain version x") if that makes more sense.


jacoblee93:
Thanks for this! I think we will also need to bump `@cloudflare/types`

I should have some time today to start splitting into a separately versioned package as discussed - will migrate over the example app as well once that's ready:

https://github.com/langchain-ai/langchain-cloudflare-nuxt-template/

PR #3984, opened 2024-01-10
community[patch]: allow upserts in MongoDB Atlas

RizeCrime:
Previously the MongoDB Vectorstore could only insert new Documents via `insertMany()`, but that lead to duplicates if you were actively working on said Documents.

Now there are two additional args you can pass when initializing the Vectorstore:
- `primaryKey: string` for setting the key which is used for filtering
- `overwrite: boolean` to update any documents that match the filter, with the new version of the document

Overwriting has upsert enabled by default (and there's currently no way to change it), so if a document doesn't already exist it will get inserted instead.

### DIFF
--- a/libs/langchain-community/src/vectorstores/mongodb_atlas.ts
+++ b/libs/langchain-community/src/vectorstores/mongodb_atlas.ts
@@ -10,13 +10,22 @@ import { maximalMarginalRelevance } from "@langchain/core/utils/math";
 /**
  * Type that defines the arguments required to initialize the
  * MongoDBAtlasVectorSearch class. It includes the MongoDB collection,
- * index name, text key, and embedding key.
+ * index name, text key, embedding key, primary key, and overwrite flag.
+ *
+ * @param collection MongoDB collection to store the vectors.
+ * @param indexName A Collections Index Name.
+ * @param textKey Corresponds to the plaintext of 'pageContent'.
+ * @param embeddingKey Key to store the embedding under.
+ * @param overwrite Whether to update/overwrite existing documents.
+ * @param primaryKey The Key to use for updating/overwriting documents.
  */
 export type MongoDBAtlasVectorSearchLibArgs = {
   readonly collection: Collection<MongoDBDocument>;
   readonly indexName?: string;
   readonly textKey?: string;
   readonly embeddingKey?: string;
+  readonly overwrite?: boolean;
+  readonly primaryKey?: string;
 };
 
 /**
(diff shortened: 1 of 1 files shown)


libs/langchain-community/src/vectorstores/mongodb_atlas.ts
```diff
@@ -76,7 +91,18 @@ export class MongoDBAtlasVectorSearch extends VectorStore {
...
+      await this.collection.insertMany(docs);
+    } else {
+      const promises = docs.map((doc) => {
+        return this.collection.updateOne(
+          { [this.primaryKey]: doc[this.primaryKey] },
+          { $set: doc },
+          { upsert: true }
+        );
+      });
+      await Promise.all(promises);
```
bracesproul, line 104: Instead of a `Promise.all` we should use the [`AsyncCaller`](https://github.com/langchain-ai/langchainjs/blob/main/langchain-core/src/utils/async_caller.ts) for some extra functionality. See[ this code example ](https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-community/src/vectorstores/typesense.ts#L134)for inspiration on how to

bracesproul (commented):
This upgrade looks really good! Left one comment on how we should execute the calls.

Thanks for this!

libs/langchain-community/src/vectorstores/mongodb_atlas.ts
```diff
@@ -61,6 +80,9 @@ export class MongoDBAtlasVectorSearch extends VectorStore {
     this.indexName = args.indexName ?? "default";
     this.textKey = args.textKey ?? "text";
     this.embeddingKey = args.embeddingKey ?? "embedding";
+    this.primaryKey = args.primaryKey ?? "_id";
+    this.overwrite = args.overwrite ?? false;
```
jacoblee93, line 84: Default for most vectorstores is actually upsert - would be nice to get this to match at some point

jacoblee93:
Thank you!

libs/langchain-community/src/vectorstores/mongodb_atlas.ts
RizeCrime: Can do, I'll try to push it some time tomorrow.

jacoblee93:
@RizeCrime I've opened #3999, which is a bit more in line with how other vectorstores currently handle upserts.

Basically, it adds a third `options.ids` property to the vectorstore. If ids are passed, we use upsert-y behavior. Otherwise, they're treated as inserts.

Let me know what you think of this!

RizeCrime:
much better than what I cobbled together

jacoblee93:
Thank you for pushing on this - will get the other PR out tonight!

PR #3991, opened 2024-01-11
allow empty arrays for Weaviate metadata

the-powerpointer:
Weaviate metadata may be of type array. This means, also empty array values are supported.
The current implementation of the `flattenObjectForWeaviate` function however more or less accidentally omits empty arrays completely, so they never arrive at Weaviate.
If one specifies an empty array as metadata value, then this value should also arrive at Weaviate.
This PR fixes this.


the-powerpointer:
@jacoblee93 Can I do anything about the failing Preview deployments?

jacoblee93:
Looks like you'd have to merge in `main` (we added a new partner package and I think it's a caching issue in Vercel) but I think this is fine, so will merge shortly thanks!

PR #3992, opened 2024-01-11
Add support for recurring events in GoogleCalendarViewTool

jordi-murgo:
The **GoogleCalendarViewTool** module from **@langchain/community/tools/google_calendar** currently does not export recurring events correctly. When displaying the start and end dates of a recurring series, it displays the start and end dates of the first element of the serie, rather than the specific date within the specified time period.

This pull request addresses this issue by updating the code to correctly display the date of occurrence within the requested time period for recurring events. It ensures that the exported data accurately reflects the individual occurrences of recurring events.

Changes Made:

Modified the code in GoogleCalendarViewTool to calculate and display the correct date of occurrence within the specified time period for recurring events.
Tested the changes thoroughly to ensure that they do not introduce any new issues and that they resolve the existing problem.

Please review and merge this pull request to improve the functionality of the GoogleCalendarViewTool module.

Thanx, @jordimurgo


libs/langchain-community/src/tools/google_calendar/commands/run-view-events.ts
```diff
@@ -8,12 +8,30 @@ import { StringOutputParser } from "@langchain/core/output_parsers";
 import { VIEW_EVENTS_PROMPT } from "../prompts/index.js";
 import { getTimezoneOffsetInHours } from "../utils/get-timezone-offset-in-hours.js";
 
+import { RRuleSet } from "rrule-rust";
+import { DateTime } from "luxon";
```
bracesproul, line 12: We want to avoid adding new packages unless absolutely necessary. Can you try to implement this without new packages? Maybe using already installed packages in the repo?

bracesproul (changes requested):
Thanks for catching this case! ideally we aren't adding new packages when possible, and even more so inside existing modules as we don't want new users to have to install extra packages if they want to upgrade. Could you look into a solution without using external dependencies?

libs/langchain-community/src/tools/google_calendar/commands/run-view-events.ts
jordi-murgo: Is really hard to implement an RRULE library. 
Luxon is convenient, because TimeZones in calendars, but can be changed.

christian-bromann:
@jordi-murgo thanks for the contribution 🙏 sorry for not getting back to you on this earlier. I would advise maybe publish your own `GoogleCalendarViewTool` on NPM that uses Luxon and RRule? I think this is still super valuable but the code doesn't necessarily need to live within the langchain repository. I will go ahead and close this but please let me know if you end up publishing this so I can highlight this within our community.

PR #3998, opened 2024-01-11
core[minor]: Add ability for runnable passthrough to call a func

bracesproul:


jacoblee93:
This needs tests

langchain-core/src/runnables/passthrough.ts
```diff
@@ -44,10 +64,29 @@ export class RunnablePassthrough<RunInput> extends Runnable<
 
   lc_serializable = true;
 
+  func?: RunnablePassthroughFunc<RunInput | AsyncGenerator<RunInput>>;
+
+  constructor(fields?: {
+    func?: RunnablePassthroughFunc<RunInput | AsyncGenerator<RunInput>>;
```
jacoblee93, line 70: Should `RunnablePassthrough.assign()` also take functions like this?

langchain-core/src/runnables/passthrough.ts
```diff
@@ -44,10 +64,29 @@ export class RunnablePassthrough<RunInput> extends Runnable<
...
+  }) {
+    super();
```
jacoblee93, line 72: Always pass through `fields`

PR #4006, opened 2024-01-12
langchain[minor]: update sql_utils: add schema for mssql

therajatg:
Fixes:

1. Previously, the typeorm configuration did not consider the schema in the case of MSSQL, resulting in tables not being recognized if they were not in the default schema (dbo). This issue prevented the generation of natural language responses or SQL queries for data in schemas other than the default. This has been fixed.

2. In cases where the schema is not explicitly defined in typeorm, Instead of automatically assigning the "dbo" schema, kept the schema undefined. This change ensures that when no schema is specified, "dbo" is not attached to the table name in the query and it'll work as before.

My twitter handle: therajatg


langchain/src/util/sql_utils.ts
```diff
@@ -177,17 +177,28 @@ export const getTableAndColumnsName = async (
...
-    sql =
+    const schema = appDataSource.options?.schema;
+    const sqlWithSchema =
       "SELECT " +
       "TABLE_NAME AS table_name, " +
       "COLUMN_NAME AS column_name, " +
       "DATA_TYPE AS data_type, " +
       "IS_NULLABLE AS is_nullable " +
       "FROM INFORMATION_SCHEMA.COLUMNS " +
+      `WHERE TABLE_SCHEMA = '${schema}'` +
```
bracesproul, line 188: If this can be undefined, should we do this instead?
```suggestion
      schema ? `WHERE TABLE_SCHEMA = '${schema}'` : "" +
```

bracesproul: This way we can just drop `sqlWithoutSchema`, see my below comment

langchain/src/util/sql_utils.ts
```diff
@@ -177,17 +177,28 @@ export const getTableAndColumnsName = async (
...
+      "SELECT " +
+      "TABLE_NAME AS table_name, " +
+      "COLUMN_NAME AS column_name, " +
+      "DATA_TYPE AS data_type, " +
+      "IS_NULLABLE AS is_nullable " +
+      "FROM INFORMATION_SCHEMA.COLUMNS " +
+      "ORDER BY TABLE_NAME, ORDINAL_POSITION;";
 
+    sql = schema ? sqlWithSchema : sqlWithoutSchema;
+    const rep = await appDataSource.query(sql);
```
bracesproul, line 201: Drop `sqlWithoutSchema` and rename `sqlWithSchema` to just `sql`
```suggestion
    const rep = await appDataSource.query(sql);
```

bracesproul (changes requested):
This is a good addition, thanks for pushing it up! I left one small comment

langchain/src/util/sql_utils.ts
therajatg: When there is no schema in the dataSource config, it is throwing syntax error ⇒ Incorrect syntax near the keyword. I think this is because when schema is not present in the configuration, WHERE TABLE_SCHEMA = "" denotes an empty string as schema name and “” is not a valid schema name.

therajatg: However, still removed the unnecessary code as you pointed out and made a new commit which is working as expected

jacoblee93:
Thank you!

PR #4008, opened 2024-01-12
Add Memgraph support

katarinasupe:
I added Memgraph support by extending `Neo4jGraph` class. In order to do that, I had to mark `schema` and `structuredSchema` as `protected` instead of `private`, so I was wondering if that's okay. Other than that, I mostly followed what we did in LangChain Python and I plan to fix a couple of things there to better align with this integration. 

@tomasonjo we spoke recently about Python integration so tagged you for awareness here.

I plan to add docs and example, but first wanted to see if this is okay to start with and if I have to first fix something in this contribution.


libs/langchain-community/src/graphs/memgraph_graph.ts
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  static async initialize(config: MemgraphGraphConfig): Promise<MemgraphGraph> {
+    const graph = new MemgraphGraph(config);
+
+    try {
+      await graph.verifyConnectivity();
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } catch (error: any) {
+      console.log("Failed to verify connection.");
```
bracesproul, line 47: nit
```suggestion
      console.error("Failed to verify connection.");
```

libs/langchain-community/src/graphs/memgraph_graph.ts
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    try {
+      await graph.refreshSchema();
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } catch (error: any) {
+      throw new Error(error.message);
+    } finally {
+      console.log("Schema refreshed successfully.");
```
bracesproul, line 56: Not sure we want this logging every time?
```suggestion
      console.debug("Schema refreshed successfully.");
```

libs/langchain-community/src/graphs/memgraph_graph.ts
```diff
@@ -0,0 +1,110 @@
...
+      const rawSchema = rawSchemaQueryResult?.[0]?.schema;
+
+      this.structuredSchema = {
+        nodeProps: rawSchema.node_props,
+        relProps: rawSchema.rel_props,
+        relationships: rawSchema.relationships,
+      };
+
+      // Format node properties
+      const formattedNodeProps = Object.entries(rawSchema.node_props)
```
bracesproul, line 74: Why are we using `rawSchema` here instead of `this.structuredSchema`?

bracesproul: Same for all occurrences below

libs/langchain-community/src/graphs/neo4j_graph.ts
```diff
@@ -42,9 +42,9 @@ export class Neo4jGraph {
 
   private database: string;
 
-  private schema = "";
+  protected schema = "";
 
-  private structuredSchema: StructuredSchema = {
+  protected structuredSchema: StructuredSchema = {
```
bracesproul, line 47: 👍

libs/langchain-community/src/graphs/tests/memgraph_graph.int.test.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    // expect(graph.getSchema()).toMatchInlineSnapshot(`
+    //   "Node properties are the following:
+    //   Node name: 'Actor', Node properties: [{"property":"name","type":"str"}]
+    //   Node name: 'Movie', Node properties: [{"property":"title","type":"str"}]
+    //   Relationship properties are the following:
+    //   Relationship name: 'ACTED_IN', Relationship properties: [{"property":"roles","type":"tuple"}]
+    //   The relationships are the following:
+    //   (:Actor)-[:ACTED_IN]->(:Movie)"
+    // `);
```
bracesproul, line 43: Should this be uncommented or deleted?

bracesproul (commented):
Left a couple comments, but overall this looks good! Also, it would be great to include docs in this PR, makes it easier to review if it's all in one.
Thanks for this!

katarinasupe:
Hi @bracesproul, thank you for the review. I will update the PR in a couple of days when I am back from the trip 😄

libs/langchain-community/src/graphs/memgraph_graph.ts
katarinasupe: It's the same if I use `rawSchema.node_props` or `this.structuredSchema.nodeProps` (similar for other occurrences). Do you prefer the latter, and if yes, why? I am not a ts/js expert, so any advice is more than welcome 😄

libs/langchain-community/src/graphs/tests/memgraph_graph.int.test.ts
katarinasupe: I will uncomment it :)

bracesproul (approved):
LGTM! Thank you for adding this & docs!

PR #4010, opened 2024-01-13
Add stacktrace to tracer errors

hinthornw:
@jacoblee93 are there env-specific gotchas with this? would be helpful to see the full stacktrace whenever there is an error


jacoblee93:
I think all good, thanks!

PR #4011, opened 2024-01-13
LangSmith Test Runner

hinthornw:
Implements the runOnDataset functionality for JS.

- [X] Implementation + integration tests for KV dataset
- [X] Support for custom run evaluators
- [X] Deserialization of chat messages, etc. 
- [X] Tests for LLM / Chat datasets
- [X] Support for string evaluators + auto-loading
- [X] Add example walkthrough

Known deficiencies:
- Doesn't (yet) print out the eval project name
- Currently doesn't apply run evaluators at the same time as the prediction step, which is slower
- Evaluator runs aren't traced to the evaluators project.


docs/core_docs/docs/guides/langsmith_evaluation.md
```diff
@@ -0,0 +1,233 @@
...
+```typescript
+import { Client } from "langsmith";
+
+const client = new Client();
+```
+
+Create a LangChain component and log runs to the platform. In this example, we will create a ReAct-style agent with access to a general search tool (DuckDuckGo). The agent's prompt can be viewed in the [Hub here](https://smith.langchain.com/hub/wfh/langsmith-agent-prompt).
+
+```typescript
+import { TavilySearchAPIRetriever } from "@langchain/community/retrievers/tavily_search_api";
```
jacoblee93, line 60: Add as an install above

docs/core_docs/docs/guides/langsmith_evaluation.md
```diff
@@ -0,0 +1,233 @@
...
+
+However, in this example, we will use environment variables.
+
+```bash
+npm install @langchain/core @langchain/openai langsmith uuid
+```
+
+```typescript
+import { v4 as uuidv4 } from "uuid";
+const unique_id = uuidv4().slice(0, 8);
```
jacoblee93, line 37: Prefer camel case

docs/core_docs/docs/guides/langsmith_evaluation.md
```diff
@@ -0,0 +1,233 @@
...
+import { DocumentInterface } from "@langchain/core/documents";
+import { StringOutputParser } from "@langchain/core/output_parsers";
+import { ChatPromptTemplate } from "@langchain/core/prompts";
+import { RunnableLambda, RunnableMap } from "@langchain/core/runnables";
+import { ChatOpenAI } from "@langchain/openai";
+
+const retriever = new TavilySearchAPIRetriever({
+  k: 10,
+});
+const formatDocs = (docs: DocumentInterface[]) =>
```
jacoblee93, line 70: Might be better to use just `Document` for clarity

docs/core_docs/docs/guides/langsmith_evaluation.md
```diff
@@ -0,0 +1,233 @@
...
+    "system",
+    "You are an AI assistant. Answer user questions as best as you are able." +
+      " Use the following search results if applicable.\n" +
+      " {docs}",
+  ],
+  ["human", "{input}"],
+]);
+
+const getString = (input: { input: string }) => input.input;
+const selectInput = new RunnableLambda({
```
jacoblee93, line 90: You can do `RunnableLambda.from()` but you also don't even need this - `.pipe` will coerce the function to a `RunnableLambda`

docs/core_docs/docs/guides/langsmith_evaluation.md
```diff
@@ -0,0 +1,233 @@
...
+```typescript
+const inputs = [
+  "What is LangChain?",
+  "What's LangSmith?",
+  "When was Llama-v2 released?",
+  "What is the langsmith cookbook?",
+  "When did langchain first announce the hub?",
+];
+
+const inputList = inputs.map((input) => ({ input }));
```
jacoblee93, line 114: Why not just make the array above contain items in the correct format?

docs/core_docs/docs/guides/langsmith_evaluation.md
```diff
@@ -0,0 +1,233 @@
...
+
+Create the langsmith client to interact with the API
+
+```typescript
+import { Client } from "langsmith";
+
+const client = new Client();
+```
+
+Create a LangChain component and log runs to the platform. In this example, we will create a ReAct-style agent with access to a general search tool (DuckDuckGo). The agent's prompt can be viewed in the [Hub here](https://smith.langchain.com/hub/wfh/langsmith-agent-prompt).
```
jacoblee93, line 57: Should we just use the default `createReactAgent()` method?

docs/core_docs/docs/guides/langsmith_evaluation.md
```diff
@@ -0,0 +1,233 @@
...
+
+### 2. Configure evaluation
+
+Manually comparing the results of chains in the UI is effective, but it can be time consuming.
+It can be helpful to use automated metrics and AI-assisted feedback to evaluate your component's performance.
+
+Below, we will create a custom run evaluator that logs a heuristic evaluation.
+
+```typescript
+import { RunEvalConfig } from "langchain/smith";
```
jacoblee93, line 173: We should also install `langchain` above, I've also not been installing `@langchain/core` explicitly since it'll always be part of `langchain`

docs/core_docs/docs/guides/langsmith_evaluation.md
```diff
@@ -0,0 +1,233 @@
...
+const notUnsure = (props: { run: Run; example?: Example }) => ({
+  key: "not_unsure",
+  score: !props.run?.outputs?.output.includes("not sure"),
+});
+
+const evaluation = new RunEvalConfig({
+  // The 'evaluators' are loaded from LangChain's evaluation
+  // library.
+  evaluators: [
+    new RunEvalConfig.LabeledCriteria({
```
jacoblee93, line 186: Could just import `LabeledCriteria`? But also, would it be better to call this a `LabeledCriteriaEvaluator`?

docs/core_docs/docs/guides/langsmith_evaluation.md
```diff
@@ -0,0 +1,233 @@
...
+import { RunEvalConfig } from "langchain/smith";
+import { Run, Example } from "langsmith";
+
+// An illustrative custom evaluator example
+const notUnsure = (props: { run: Run; example?: Example }) => ({
+  key: "not_unsure",
+  score: !props.run?.outputs?.output.includes("not sure"),
+});
+
+const evaluation = new RunEvalConfig({
```
jacoblee93, line 182: Too much OO? Could this config just be a type?

langchain/Untitled.ipynb
```diff
@@ -0,0 +1,29 @@
+{
```
jacoblee93, line 1: Extra file - notebook -> docs page isn't supported yet :(

langchain/scripts/create-entrypoints.js
```diff
@@ -336,6 +336,8 @@ const entrypoints = {
   "experimental/tools/pyinterpreter": "experimental/tools/pyinterpreter",
   // evaluation
   evaluation: "evaluation/index",
+  // smith (LangSmith Evaluation)
+  smith: "smith/index",
```
jacoblee93, line 340: This was the only change in this file right? The rest is just formatting?

langchain/src/smith/config.ts
```diff
@@ -0,0 +1,248 @@
...
+      evaluators?: (T | EvalConfig)[];
+      customEvaluators?: U[];
+      prepareData?: PrepareDataT;
+      eval_llm?: string;
+    } = {}
+  ) {
+    this.evaluators = config.evaluators;
+    this.customEvaluators = config.customEvaluators;
+    this.prepareData = config.prepareData;
+    this.eval_llm = config.eval_llm;
```
jacoblee93, line 246: Prefer camel case

langchain/src/smith/config.ts
```diff
@@ -0,0 +1,248 @@
...
+   * );
+   * ```
+   * @example
+   * ```ts
+   * const evalConfig = new RunEvalConfig(
+   * [new RunEvalConfig.Criteria(
+   *      { "mentionsAllFacts": "Does the include all facts provided in the reference?"
+   *  })],
+   */
+  static LabeledCriteria = class implements EvalConfig {
```
jacoblee93, line 201: My gut is these don't need to be classes - also, no need to nest them?

langchain/scripts/create-entrypoints.js
jacoblee93: More philosophically, does all this really belong in LangChain.js? Could it just be part of the `langsmith` SDK?

docs/core_docs/docs/guides/langsmith_evaluation.md
hinthornw: Hm Document wasn't working with the type system but I'll see

docs/core_docs/docs/guides/langsmith_evaluation.md
hinthornw: Even if first in the chain though?

docs/core_docs/docs/guides/langsmith_evaluation.md
hinthornw: Ya is that the one we're using now? Was mainly lazy / couldn't find the recommended one quickly

docs/core_docs/docs/guides/langsmith_evaluation.md
hinthornw: Could call it a `LabeledCriteriaConfig` if that's more clear? It gets initialized in the actual test runner

langchain/Untitled.ipynb
hinthornw: Oops TY@

langchain/scripts/create-entrypoints.js
hinthornw: Could be its own package. Reason for being here is langsmith can't have a dep on langchain

langchain/src/smith/config.ts
hinthornw: Ya - don't feel terribly strongly about nesting, but that makes it similar to python. I  was running into issues defining nested types so could also just have flat types instead of nested classes I believe

docs/core_docs/docs/guides/langsmith_evaluation.md
jacoblee93: Yes

docs/core_docs/docs/guides/langsmith_evaluation.md
jacoblee93: In that case see below, my preference would be to make it a type and just treat it like a plain object.

Issue #3983, opened 2024-01-10
Pinned zod-to-json-schema version


StefanTerdell:
Hello! I've just recently started using langchain a bit, and it turns out you use [one of my packages as well](https://github.com/StefanTerdell/zod-to-json-schema) :) I noticed the version has been pinned though. If you'd like I could verify that the current version works as intended and open a PR to unpin it as I believe the underlying issue has been resolved (the recursive stack overflow in babel transpiled environments). There have been some breaking changes to export paths of "non-api" components like types as well.

jacoblee93:
A patch there would be awesome! `zodToJsonSchema` has been really nice for a lot of the key structured output use cases. If things are patched, we can definitely unpin.

StefanTerdell:
Fixed in #4041

Issue #4009, opened 2024-01-12
Docs: Typescript example not working. Needed a workaround (typing)


ed3899:
At https://js.langchain.com/docs/modules/agents/how_to/custom_agent

The following example thrown a ts error.

```
import { formatToOpenAIFunction } from "langchain/tools";

const modelWithFunctions = model.bind({
  functions: tools.map((tool) => formatToOpenAIFunction(tool)),
});
```

The error:
```
Type 'Function' is not assignable to type 'FunctionDefinition'.
  Property 'parameters' is optional in type 'Function' but required in type 'FunctionDefinition'.ts(2322)
```

The likely cause: https://github.com/langchain-ai/langchainjs/blob/44b00c0d21b4faf15f78209339191a959077128f/langchain-core/src/language_models/base.ts#L202

Line 202

The params of FunctionDefinition:

```
 parameters: Record<string, unknown>;
```

Seem to cause the conflict with:
```
OpenAIClient.Chat.Completions.ChatCompletionCreateParams.Function
```

The workaround in the meantime:
```
formatToOpenAIFunction(this.customTool) as FunctionDefinition,
```

Just make the params in sync

jacoblee93:
Hey sorry you're encountering this - what version of LangChain are you using?

jacoblee93:
Never mind, I see the problem - you can fix this in the immediate term by switching the import to `@langchain/openai` but I will also patch this.

jacoblee93:
New core methods are live in `@langchain/core` 0.1.13:

If you upgrade to latest langchain or just bump core you can do:

```ts
import { convertToOpenAIFunction } from "@langchain/core/utils/function_calling";

const modelWithFunctions = model.bind({
  functions: tools.map((tool) => convertToOpenAIFunction(tool)),
});
```

https://js.langchain.com/docs/modules/agents/how_to/custom_agent#bind-tools-to-llm

PR #4023, opened 2024-01-15
core[minor]: Move function calling definitions and utils to core

jacoblee93:
We can look at unpinning/bumping `zod-to-json-schema` (as mentioned in #3983) in a separate PR.


jacoblee93:
This is the first step towards addressing #4009

PR #4032, opened 2024-01-15
core[patch]: Fix return type of .pipe()

nfcampos:
Fixes # (issue)

### DIFF
--- a/langchain-core/src/runnables/base.ts
+++ b/langchain-core/src/runnables/base.ts
@@ -539,7 +539,7 @@ export abstract class Runnable<
    */
   pipe<NewRunOutput>(
     coerceable: RunnableLike<RunOutput, NewRunOutput>
-  ): RunnableSequence<RunInput, Exclude<NewRunOutput, Error>> {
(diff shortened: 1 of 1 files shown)


langchain-core/src/runnables/base.ts
```diff
@@ -539,7 +539,7 @@ export abstract class Runnable<
    */
   pipe<NewRunOutput>(
     coerceable: RunnableLike<RunOutput, NewRunOutput>
-  ): RunnableSequence<RunInput, Exclude<NewRunOutput, Error>> {
+  ): Runnable<RunInput, Exclude<NewRunOutput, Error>> {
     // eslint-disable-next-line @typescript-eslint/no-use-before-define
```
bracesproul, line 543: I believe this should be `RunnableInterface<RunInput, Exclude<NewRunOutput, Error>>`. cc @jacoblee93 

Same with the below two

bracesproul: Never mind, this is not the case

PR #4042, opened 2024-01-16
Populate streamed_output for all runs handled by transformStreamWithConfig

jacoblee93:
Port of https://github.com/langchain-ai/langchain/pull/15599

CC @nfcampos


jacoblee93:
Standard `transform` does not seem to populate `streamed_output` but we can handle that in a separate PR.

PR #4048, opened 2024-01-17
core[patch]: Refactor patchConfig method and add ensureConfig

jacoblee93:
@nfcampos

### DIFF
--- a/langchain-core/src/runnables/config.ts
+++ b/langchain-core/src/runnables/config.ts
@@ -138,7 +138,7 @@ export function patchConfig<CallOptions extends RunnableConfig>(
     recursionLimit,
     runName,
     configurable,
-  }: RunnableConfig
+  }: RunnableConfig = {}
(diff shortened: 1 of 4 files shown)


langchain-core/src/runnables/config.ts
```diff
@@ -108,3 +111,55 @@ export function mergeConfigs<CallOptions extends RunnableConfig>(
...
+    newConfig.recursionLimit = recursionLimit;
+  }
+  if (maxConcurrency !== undefined) {
+    newConfig.maxConcurrency = maxConcurrency;
+  }
+  if (runName !== undefined) {
+    newConfig.runName = runName;
+  }
+  if (configurable !== undefined) {
+    newConfig.configurable = configurable;
```
nfcampos, line 162: this should merge both

jacoblee93: Ah thanks

PR #4057, opened 2024-01-17
community[patch]: Add neo4j timeout option

tomasonjo:


libs/langchain-community/src/graphs/neo4j_graph.ts
```diff
@@ -55,10 +58,12 @@ export class Neo4jGraph {
     username,
     password,
     database = "neo4j",
+    timeout,
   }: Neo4jGraphConfig) {
     try {
       this.driver = neo4j.driver(url, neo4j.auth.basic(username, password));
       this.database = database;
+      this.timeout = timeout ? timeout * 1000 : undefined;
```
jacoblee93, line 66: nit: We could just call this prop `timeoutMs` and avoid the math here?

tomasonjo: done

jacoblee93:
Thank you!

PR #4064, opened 2024-01-17
Indexing API Port

MJDeligan:
Hey guys,

sorry for the delay! As discussed in #3390 I've ported the [indexing API](https://python.langchain.com/docs/modules/data_connection/indexing) from the python module as well as the required RecordManager interfaces and implementations. Opening this as a draft since it certainly needs some adjustments.

### Features
- RecordManager interface
- RecordManager abstract class
- PostgresRecordManager
- InMemoryRecordManager
- index function including HashedDocument class
- tests for both RecordManagers and the index function
- An example how to use the index function based on [this](https://python.langchain.com/docs/modules/data_connection/indexing)

### Notes
In [this comment](https://github.com/langchain-ai/langchainjs/issues/3390#issuecomment-1867121807) I mentioned my struggle with organizing the code in the different packages. While I was able to move the RecordManager Interface out of `langchain/core` I couldn't move the index function out of `langchain/` since it depends on the BaseDocumentLoader. I didn't find an equivalent in the `langchain-community` package.

The index function currently has the problem of mishandling incremental mode when time measurement is not precise enough and/or there is little latency. This is usually not a problem, since there is likely to be latency and time measurements are precise, but in the case of the InMemoryRecordManager I had to change the filter condition for time-based filters. If I didn't include the timestamp passed, it would sometimes not return outdated keys, because the timestamp at which they were inserted and the timestamp at which the updated that would have deleted them started were the same ( using new Date() as a timestamp).

I also ran the index function tests with the PostgresRecordManager to ensure that this doesn't cause the tests to incorrectly pass.

The index function test requires a python database. I wanted to use something without an external dependency, but I couldn't find a suitable vectorstore :/

=====================

I'm a little swamped right now so I may not be able to immediately answer change requests. Feel free to change things at will or use this PR as a basis for your own implementation.

Sorry for the size,
Cheers


jacoblee93:
Thank you so much!

bracesproul:
Hey @MJDeligan thank you for pushing this up! I just opened up a PR [here](https://github.com/langchain-ai/langchainjs/pull/4081/files) containing the additions to the community package. We'll want to land that first, then the changes to the `langchain` package, along with docs. Let me know if you have any feedback on that PR! I didn't change much of the core code, mainly entrypoints and code locations.

MJDeligan:
Hey @bracesproul, PRs are looking great, thank you very much! Let me know if you need anything!

jacoblee93:
Thank you very much! Great job!

PR #4067, opened 2024-01-17
Update runnable config ensure and merge to match py

nfcampos:
- merge now supports merging many configs
- fix issues in edge cases of merge
- call ensure everywhere configs are used
- ensure assigns primitive configurable keys to tracer metadata

Fixes # (issue)

### DIFF
--- a/langchain-core/src/callbacks/manager.ts
+++ b/langchain-core/src/callbacks/manager.ts
@@ -495,7 +495,7 @@ export class CallbackManager
 
   name = "callback_manager";
 
-  private readonly _parentRunId?: string;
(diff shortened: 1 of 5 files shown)


jacoblee93:
Seems fine, but we don't need to do this in places like here to set defaults?

https://github.com/langchain-ai/langchainjs/blob/main/langchain-core/src/language_models/chat_models.ts#L139

jacoblee93:
Test might be a flake, can't repro locally

nfcampos:
@jacoblee93 yes but it is also called inside that _split method so technically we already are doing so

jacoblee93:
Thanks!

PR #4077, opened 2024-01-18
weaviate[major],langchain[patch],community[patch]: Adds weaviate vector store integration package

bracesproul:


docs/core_docs/docs/integrations/vectorstores/weaviate.mdx
```diff
@@ -6,18 +6,21 @@ import CodeBlock from "@theme/CodeBlock";
...
 
+Weaviate has their own standalone integration package with LangChain, accessible via [`@langchain/weaviate`](https://www.npmjs.com/package/@langchain/weaviate) on NPM!
+
 import IntegrationInstallTooltip from "@mdx_components/integration_install_tooltip.mdx";
 
 <IntegrationInstallTooltip></IntegrationInstallTooltip>
 
 ```bash npm2yarn
-npm install @langchain/openai weaviate-ts-client graphql @langchain/community
+npm install @langchain/weaviate @langchain/openai graphql @langchain/community
```
jacoblee93, line 23: We should not need all these deps anymore?

examples/package.json
```diff
@@ -71,6 +72,7 @@
     "uuid": "^9.0.0",
     "vectordb": "^0.1.4",
     "voy-search": "0.6.2",
+    "weaviate": "^0.0.14",
```
jacoblee93, line 75: What is this?

examples/package.json
```diff
@@ -71,6 +72,7 @@
...
     "weaviate-ts-client": "^1.0.0",
```
jacoblee93, line 76: Should bump this version too?

libs/langchain-weaviate/src/vectorstores.ts
```diff
@@ -0,0 +1,436 @@
...
+            pageContent: text,
+            metadata: rest,
+          }),
+          _additional.distance,
+          _additional.vector,
+        ]);
+      }
+      return documents;
+    } catch (e) {
+      throw Error(`'Error in similaritySearch' ${e}`);
```
jacoblee93, line 331: Should this error be in quotes?

jacoblee93 (commented):
Let's ping someone on their side and make sure they review as well!

docs/core_docs/docs/integrations/vectorstores/weaviate.mdx
bracesproul: ya just looked, correct on `graphql` but we still need openai, and community

examples/package.json
bracesproul: dropped

libs/langchain-weaviate/src/vectorstores.ts
bracesproul: dropped

hsm207:
@bracesproul thank you for your contribution! 
 
 It's impressive to see the inclusion of both unit and integration tests in your PR. 
 
Could you provide more details on the current test coverage for this integration? Understanding the extent of the coverage will help us ensure the robustness of the implementation and identify any potential areas that might need additional tests

bracesproul:
Hey @hsm207, these tests are just the same as in `@langchain/community` (I pretty much just copy-pasted the weaviate code into this new package).

As for the tests, the non integration test only checks the `flattenObjectForWeaviate` which flattens metadata for the `properties` field in the API requests.

The integration test tests:
- similaritySearch
- similaritySearch with metadata
- upserting via texts
- upserting via `Document`s
- deleting
- deleting with filters

Let me know if this helps, and if you need anything else!

hsm207:
Hey @jacoblee93, could you please let us know how do we check the test coverage in this project?

Given that this PR  is about the integration with Weaviate, it's important for us to ensure that the code meets our quality benchmarks. 

Insight into the test coverage would be very helpful for identifying any areas we might need to address.

jacoblee93:
Hi @hsm207, we don't currently measure code coverage but that would be a great thing to add. Let me get back to you once it's set up!

bracesproul:
Hey @hsm207 I'll be setting up test coverage this weekend, I'll tag you once it's in and we have a concrete number!

bracesproul:
@hsm207 The current weaviate integration has 20% test coverage.

hsm207:
@bracesproul thanks for the update.

20%  is unexpectedly low 🤣

Can you share the test coverage report? I'd like to see what we're missing. Ideally, we should be closer to 90% coverage before we proceed to review.

bracesproul:
> @bracesproul thanks for the update.
> 
> 20% is unexpectedly low 🤣
> 
> Can you share the test coverage report? I'd like to see what we're missing. Ideally, we should be closer to 90% coverage before we proceed to review.

Our code coverage tool doesn't have a share option, so instead I've included screenshots from the report. Let me know if you need anything else!
image
image
image
image
image
image
image
image
image

hsm207:
@bracesproul thanks for sharing the screenshots.

How do I interpret them? Does red mean the line is not covered by any tests? What about yellow?

bracesproul:
> @bracesproul thanks for sharing the screenshots.
> 
> How do I interpret them? Does red mean the line is not covered by any tests? What about yellow?

Sorry for the delay, yes the red lines are lines which `codecov` has determined are not tested, and yellow means partial test coverage. Let me know if you need anything else!

hsm207:
@bracesproul, thanks for the info.

Just had another look at the coverage report and, honestly, I'm puzzled. Even basic stuff like the constructor (line 109) is left untested. 🤔

More importantly, critical functions for the integration – like adding documents, searching, etc. – are also in the no-coverage zone. This doesn't just raise red flags; it's practically a stop sign 😝

Given this, I can't endorse merging this PR in its current state. It's not my call to make, though since I'm not a maintainer. @jacoblee93, over to you for the final decision.

jacoblee93:
@bracesproul can we fix?

bracesproul:
> @bracesproul can we fix?

Taking a look now!

bracesproul:
Hey @hsm207 so I pushed up some more tests, however in doing so I realized our codecov percentage was so low because it was only checking the tests we run in ci (non integration tests). Most of the tests in the langchain repo are integration tests, and we don't run those in CI for various reasons. We do however run them before releasing new versions, so they end up getting run a few times a week. Running jest coverage locally gives these results (includes integration tests):

image

One other thing to note, we're still unable to use the default export from `weaviate-ts-client`, even on the latest version without casting it as `any`.
The error it throws it:
```
Property 'client' does not exist on type 'typeof import("/Users/bracesproul/code/lang-chain-ai/wt/brace/weaviate-integration-pkg/node_modules/weaviate-ts-client/dist/index")'.ts(2339)
```
image

Is there anything we can do about this?

hsm207:
@bracesproul I'll defer to @jacoblee93 to decide if the coverage is sufficient.

> Is there anything we can do about this?

This is related to your typescript settings. I suggest to compare your tsconfig with this project and look for the difference:

https://github.com/weaviate/weaviate-breadboard-kit

this project also has a `weaviate.client` call but it does not throw any compile erorr. See [here](https://github.com/weaviate/weaviate-breadboard-kit/blob/571457b6bd861638642ac6e17bd3fef811b1bac2/src/lib/weaviate-client.ts#L37-L41)

PR #4078, opened 2024-01-18
langchain[patch]: Correcting token text-splitting behavior

matthewdparker:
- Currently,  the `while` clause gets evaluated _after_ the step-back by chunkOverlap, rather than before. As a result, if the final chunk's length exceeds (chunkSize - chunkOverlap) many tokens, an extra truncated chunk gets pushed.
- Corrected behavior will no longer push a truncated chunk if the final chunk's length exceeds (chunkSize - chunkOverlap).

Fixes # N/A


bracesproul (commented):
looks reasonable to me, thanks for pushing this fix! cc @jacoblee93 for a second look

jacoblee93:
Thanks for flagging - could we add a test for this?

I'm happy to dig in and do it if you don't have time

jacoblee93:
Thank you!

PR #4092, opened 2024-01-19
all[minor]: Replace all LC package scripts with `@langchain/scripts`

bracesproul:
### TODO
Fix `@langchain/community` adding index files to build artifacts.

Scripts will now be imported via `@langchain/scripts`.
Functions can be imported, and args passed the same as before.

`langchain` & `@langchain/community` will need to pass in a second arg (after the `pre` arg: `true`) so the extra functions for those two pkgs can be executed.

Since we have them in their own package now, we can write all the files as `.ts` (lets goooooooo!!)

cc @nfcampos I had to update `identify-secrets.ts` so it would make ts happy. Pretty much just added more type narrowing.

`./scripts` still exists, for this package.

I dropped the `tsconfig.cjs.json` file because it was causing the build to fail due to code not being avail in cjs:
```bash
src/move-cjs-to-dist.ts:6:40 - error TS1343: The 'import.meta' meta-property is only allowed when the '--module' option is 'es2020', 'es2022', 'esnext', 'system', 'node16', or 'nodenext'.

6   return resolve(dirname(fileURLToPath(import.meta.url)), relativePath);
                                         ~~~~~~~~~~~


Found 1 error in src/move-cjs-to-dist.ts:6
```


libs/langchain-scripts/package.json
```diff
@@ -0,0 +1,79 @@
...
+    "test:int": "NODE_OPTIONS=--experimental-vm-modules jest --testPathPattern=\\.int\\.test.ts --testTimeout 100000 --maxWorkers=50%",
+    "format": "prettier --write \"src\" \"scripts\"",
+    "format:check": "prettier --check \"src\" \"scripts\""
+  },
+  "author": "LangChain",
+  "license": "MIT",
+  "dependencies": {
+    "@langchain/core": "workspace:^",
+    "rollup": "^4.5.2",
+    "typescript": "<5.2.0"
```
bracesproul, line 39: was forced to move these to hard deps

jacoblee93:
I still see `tsconfig.cjs.json`?

langchain/scripts/move-cjs-to-dist.js
```diff
@@ -1,38 +1,17 @@
-import { resolve, dirname, parse, format } from "node:path";
```
jacoblee93, line 1: I was hoping to remove this and other scripts - can we not just build these via `package.json` commands?

jacoblee93:
Was hoping to replace most `scripts/` files with `package.json` commands and like a `langchain_build.config.json` field or something - would that be possible?

langchain/scripts/move-cjs-to-dist.js
bracesproul: Ah yeah I think that should be possible for them. Let me think on it and I'll get around to it tmro

langchain-core/index.cjs
```diff
@@ -0,0 +1 @@
+module.exports = require('./dist/index.cjs');
```
jacoblee93, line 1: Artifacts?

langchain-core/package.json
```diff
@@ -440,9 +444,6 @@
     "utils/types.d.ts",
     "vectorstores.cjs",
     "vectorstores.js",
-    "vectorstores.d.ts",
-    "index.cjs",
```
jacoblee93, line 444: Something weird happened?

langchain/scripts/check-tree-shaking.js
```diff
@@ -1,30 +1,7 @@
...
-
-  return entrypoints;
-}
-
-export function listExternals() {
-  return [
-    ...Object.keys(packageJson.dependencies),
-    ...Object.keys(packageJson.peerDependencies),
+checkTreeShaking({
+  extraInternals: [
```
jacoblee93, line 4: This should be in the `config.js` file no?

langchain/scripts/create-entrypoints.js
```diff
@@ -812,248 +811,14 @@ const extraImportMapEntries = [
...
-      .join(",\n") + ",\n];\n";
-  fs.writeFileSync(
-    `../${pkg}/${importConstantsPath}`,
-    "// Auto-generated by `scripts/create-entrypoints.js`. Do not edit manually.\n\nexport const optionalImportEntrypoints = [\n" +
-      contents
-  );
-};
-
-const command = process.argv[2];
+const absTsConfigPath = path.resolve(process.cwd(), "tsconfig.json");
```
jacoblee93, line 814: Same as above - should be a field in `config.js`?

langchain/scripts/move-cjs-to-dist.js
```diff
@@ -1,38 +1,17 @@
...
-
-      await writeFile(abs(`${dest}/${renamed}`), rewritten, "utf8");
-    }
-  }
-}
-
-moveAndRename("../dist-cjs", "../dist").catch((err) => {
-  console.error(err);
-  process.exit(1);
+moveAndRename({
```
jacoblee93, line 13: Should be possible via config fields

langchain/src/agents/react/prompt.ts
```diff
@@ -1,4 +1,4 @@
-export const FORMAT_INSTRUCTIONS /* #__PURE__ */ = `Use the following format:
+export const FORMAT_INSTRUCTIONS = `Use the following format:
```
jacoblee93, line 1: This is no longer necessary?

langchain/src/evaluation/comparison/pairwise.ts
```diff
@@ -20,7 +20,7 @@ import { PROMPT, PROMPT_WITH_REFERENCES } from "./prompt.js";
 import { ConstitutionalPrinciple } from "../../chains/index.js";
 import { Criteria, CriteriaLike } from "../criteria/criteria.js";
 
-const SUPPORTED_CRITERIA: Record<Criteria, string> = /* #__PURE__ */ {
+const SUPPORTED_CRITERIA: Record<Criteria, string> = {
```
jacoblee93, line 23: Why remove this? Was it redundant to begin with?

libs/create-langchain-integration/template/scripts/check-tree-shaking.js
```diff
@@ -1,80 +1,5 @@
...
-  if (failed) {
-    process.exit(1);
-  } else {
-    console.log("Tree shaking checks passed!");
-  }
-}
-
-checkTreeShaking();
+checkTreeShaking({
+  extraInternals: [/node\:/, /@langchain\/core\//],
```
jacoblee93, line 4: Make this a config field

jacoblee93:
Looking solid, see comments

PR #4103, opened 2024-01-21
langchain[minor]: Add imports migration script

bracesproul:


jacoblee93:
Could you resolve the merge conflicts?

langchain/package.json
```diff
@@ -1024,6 +1027,7 @@
     "d3-dsv": "^2.0.0",
     "epub2": "^3.0.1",
     "fast-xml-parser": "^4.2.7",
+    "glob": "^10.3.10",
```
jacoblee93, line 1030: Ah you will need to add these as `devDependencies` and mark them as optional below

PR #4111, opened 2024-01-22
scripts[patch]: Remove default index from generateFiles

bracesproul:


bracesproul:
Merged & released in another pr

PR #4124, opened 2024-01-22
langchain[minor],docs[minor]: Add langchain indexing API & docs

bracesproul:
For #4064


examples/src/indexes/indexing_api/indexing.ts
```diff
@@ -0,0 +1,494 @@
...
+    recordManager,
+    vectorStore,
+    options: {
+      cleanup: "incremental",
+      sourceIdKey: "source",
+    },
+  })
+);
+/*
+{
```
jacoblee93, line 409: Can we fix these indents?

examples/src/indexes/indexing_api/indexing.ts
```diff
@@ -0,0 +1,494 @@
...
+  metadata: { source: "kitty.txt" },
+};
+
+const doc2 = {
+  pageContent: "doggy",
+  metadata: { source: "doggy.txt" },
+};
+
+/**
+ * Hacky helper method to clear content. See the `full` mode section to to understand why it works.
```
jacoblee93, line 67: Can you just explain it in this comment?

I think would be better in smaller chunks/separate examples rather than a giant code block

PR #4147, opened 2024-01-24
community[minor]: fix Astra db peerDependency

mfortman11:
The dev and peer dependency previously did not match (my bad)
Also bumped the version


bracesproul (approved):
Ouch, sorry for not catching this in review, but thanks for catching yourself and pushing a fix! Shouldn't have affected anyone b/c you added the `^` but it's good to keep them in unison.

PR #4162, opened 2024-01-26
langchain[patch]: Update structured_output to allow parsing using zod

ashburnham:
Add zodSchema via constructor when both jsonSchema and zodSchema are passed in.

This happens at the end of createStructuredOutputChain and without this change the parseResult method will never use the zod schema for validation (which means we lose loads of useful zod functionality over just using static json schema for validation).

This follows on from work co-authored by @jacoblee93 in this commit 

https://github.com/langchain-ai/langchainjs/pull/3866


bracesproul (approved):
Good catch! Thanks for pushing up this fix, will get it merged asap!

jacoblee93:
Thank you! Could you please run `yarn format`? I can't push to your branch.

ashburnham:
I've just run `yarn format`. Back to you to merge :)

PR #4166, opened 2024-01-26
docs[patch]: Remove pinecone node-only warnings

erakor:
Removing the Node-only warnings since [the release of v1.0.1](https://github.com/pinecone-io/pinecone-ts-client/releases/tag/v1.0.1), Pinecone can run on Edge. The Pinecone integration [now uses v2.0.0.](https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-pinecone/package.json#L38)


jacoblee93:
Ah amazing thanks!

PR #4176, opened 2024-01-26
langchain[patch]: Add eval source links

hinthornw:
We want to auto-populate the traces from evaluators, where possible. This PR enables that


langchain/src/smith/runner_utils.ts
```diff
@@ -33,6 +35,30 @@ export type ChainOrFactory =
   | (() => (obj: unknown) => unknown)
   | (() => (obj: unknown) => Promise<unknown>);
 
+class RunIdResolver {
```
hinthornw, line 38: If this is a decent abstraction we could factor out and include in the langsmith docs too 🤷

jacoblee93: I like it! Would just change to `Extractor`

PR #4180, opened 2024-01-27
langchain[patch]: Trace in Evaluators Project

hinthornw:
Previously, you could get an un-traced run id if `LANGCHAIN_TRACING_V2` was unset. This PR
1. Makes it so the evaluators will predictably be traced within the `runOnDataset` function
2. Sends their races to the `evalators` project (as is done in python)

### DIFF
--- a/langchain/src/smith/runner_utils.ts
+++ b/langchain/src/smith/runner_utils.ts
@@ -49,13 +49,9 @@ class RunIdExtractor {
   handleChainStart = (
     _chain: Serialized,
     _inputs: ChainValues,
-    runId: string,
-    parentRunId?: string
+    runId: string
   ) => {
-    if (!parentRunId) {
-      // Only resolve root runs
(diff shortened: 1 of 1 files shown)


langchain/src/smith/runner_utils.ts
```diff
@@ -49,9 +49,13 @@ class RunIdExtractor {
...
     _chain: Serialized,
     _inputs: ChainValues,
-    runId: string
+    runId: string,
+    parentRunId?: string
   ) => {
-    this.runIdPromiseResolver(runId);
+    if (!parentRunId) {
+      // Only resolve root runs
+      this.runIdPromiseResolver(runId);
```
hinthornw, line 57: Also added this. Not strictly required, but I think it makes sense to only match on the root chain runs

jacoblee93:
Thank you!

PR #4188, opened 2024-01-28
cohere[patch]: Fix use of CohereChat with RunnableWithMessageHistory

### LINKED ISSUE #4186: ChatCohere memory doesn't work with RunnableWithMessageHistory
I'm using ChatCohere with RunnableWithMessageHistory and ChatMessageHistory

When I introduce myself and ask for my name back it doesn't remember.

trace shows that the messages were sent but response doesn't seem right.
https://smith.langchain.com/public/234db893-cec2-41f6-8964-13f0714061e3/r

I was able to fix and will submit a pr.

imjwang:
Hi, I've been using `ChatCohere` and wanted chat memory. Using Cohere's internal chat history works fine with `conversationId`. But I was unable to get it working with Langchain's memory system. I was using `RunnableWithMessageHistory`.

I realize it's in beta, and wanted to help.

So it seems like sending messages with Cohere client does not work properly. I was able to get it working by renaming chat_history and changing the behavior of the history array.

1. rename `chat_history` to `chatHistory`

It seems to only work with "chatHistory", their docstring is a little confusing.

https://github.com/cohere-ai/cohere-typescript/blob/c7d02986e6e4378d431ae19df1bf96b3388f7d1d/src/api/client/requests/ChatStreamRequest.ts#L32

2. Another issue seems to be with using `concat` here which returns a copied array but `chat_history` stays empty.

https://github.com/langchain-ai/langchainjs/blob/5df74e39a8b9888ec5a515d43594f856ffc1c2a7/libs/langchain-cohere/src/chat_models.ts#L277-L280

I used push for this pr:
```javascript
const chat_history = [];
if (cohereMessages.length > 1) {
    chat_history.push(...cohereMessages.slice(0, -1));
}
```

this works too:
```javascript
let chat_history = [];
if (cohereMessages.length > 1) {
  chat_history = chat_history.concat(cohereMessages.slice(0, -1));
}
```

Fixes #4186

### DIFF
--- a/libs/langchain-cohere/src/chat_models.ts
+++ b/libs/langchain-cohere/src/chat_models.ts
@@ -177,14 +177,14 @@ export class ChatCohere<
     // The last message in the array is the most recent, all other messages
     // are apart of the chat history.
     const { message } = cohereMessages[cohereMessages.length - 1];
-    const chat_history: Cohere.ChatMessage[] = [];
+    const chatHistory: Cohere.ChatMessage[] = [];
     if (cohereMessages.length > 1) {
-      chat_history.concat(cohereMessages.slice(0, -1));
+      chatHistory.push(...cohereMessages.slice(0, -1));
     }
     const input = {
       ...params,
       message,
-      chat_history,
+      chatHistory,
     };
 
     // Handle streaming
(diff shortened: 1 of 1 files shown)


jacoblee93:
Oh jeez, nice catch!

bracesproul (approved):
Whoops, thanks for catching this! We'll get it merged & released asap

PR #4191, opened 2024-01-29
langchain[patch]: adding start arg to confluence loader

fuleinist:
Adding optional `start` and `limit` arg to confluence load function so it can be used to offset the api call for fetching documents

e.g. 

```
const loader = new ConfluencePagesLoader({
   baseUrl: "https:
   spaceKey: "~EXAMPLE362906de5d343d49dcdbae5dEXAMPLE",
   username: "your-username",
   accessToken: "your-access-token",
 });
 const documents = await loader.load({ start: 5, limit: 50 });
 console.log(documents);
```


langchain/src/document_loaders/web/confluence.ts
```diff
@@ -100,11 +100,14 @@ export class ConfluencePagesLoader extends BaseDocumentLoader {
   /**
    * Fetches all the pages in the specified space and converts each page to
    * a Document instance.
+   * @param options the extra options of the load function
+   * @param options.limit The limit parameter to overwrite the size to fetch pages.
+   * @param options.start The start parameter to set inital offset to fetch pages.
    * @returns Promise resolving to an array of Document instances.
    */
-  public async load(): Promise<Document[]> {
+  public async load(options?: { start?: number, limit?: number }): Promise<Document[]> {
```
jacoblee93, line 108: A little worried that we're deviating from the base interface, but this seems reasonable.

jacoblee93:
Could you please run `yarn format`?

langchain/src/document_loaders/web/confluence.ts
fuleinist: the base interface doesn't support a few things to make this a really useful loader for indexing a full confluence space. 

Planning to add support to [expand](https://developer.atlassian.com/cloud/confluence/rest/v1/intro/#using:~:text=use%20the%20expand%20query%20parameter) parameter in my next patch as well

fuleinist:
> Could you please run `yarn format`?

not sure why but both `yarn format:check` and `yarn lint` have passed in local

image
image

jacoblee93:
Could you resolve the merge conflict?

jacoblee93:
Thank you!

PR #4192, opened 2024-01-29
Fix Deprecated Import from docs/agents/quick_start.mdx

leonvanzyl:
This import is deprecated.
import { createRetrieverTool } from "langchain/agents/toolkits";

Should be:
import { createRetrieverTool } from "langchain/tools/retriever";


jacoblee93:
You are correct, thank you!

PR #4193, opened 2024-01-29
community[patch]: adding cassandra utils

mieslep:
Adds
* Cassandra utils, allowing for native client

Fixes
* Typo in factory name


mieslep:
@jacoblee93 I think this is all the more we need, 

I got stuck on 

is not really up-to-date perhaps?
> Instead, you should add a separate entrypoint for your integration in [libs/langchain-community/scripts/create-entrypoints.js](https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-community/scripts/create-entrypoints.js) within the entrypoints object

I'm not sure where I can add a doc page for the utils as there is no docs/core_docs/docs/integrations/utils directory 

But this can be added later on - more useful to have utils/cassandra exposed sooner rather than later.

jacoblee93:
Haha whoops - thank you!

jacoblee93:
Thank you!

PR #4206, opened 2024-01-30
langchain[patch]: add expand param to ConfluencePagesLoaderParams

fuleinist:
add expand param to ConfluencePagesLoaderParams so the default value `expand=body.storage` can be overwritten.

The detailed description can be found at [here](https://developer.atlassian.com/server/confluence/expansions-in-the-rest-api/)


jacoblee93:
Thank you!

PR #4207, opened 2024-01-30
Update pgvector docker image

mkesper:
PGVector docker releases became more official:

https://github.com/pgvector/pgvector#docker
https://github.com/pgvector/pgvector?tab=readme-ov-file#docker-1

This updates the pgvector docker image to the new namespace. Old address is deprecated and will not be updated.


jacoblee93:
Thank you!

PR #4220, opened 2024-01-31
langchain[patch]: update metadata and default expand param for confluence loader

fuleinist:
- adding more metadata to confluence document_loader
- adjust default expansion to support metadata from version, see [here](https://developer.atlassian.com/server/confluence/expansions-in-the-rest-api/#expand-one-or-more-items) for more details


jacoblee93:
Thanks for this! It's complaining about formatting again though :(

jacoblee93:
Thank you for your patience!

PR #4225, opened 2024-01-31
community[minor]: Add support for TogetherAI chat json/tool calling

bracesproul:


libs/langchain-community/src/chat_models/togetherai.ts
```diff
@@ -15,11 +14,15 @@ type TogetherAIUnsupportedArgs =
...
 
-type TogetherAIUnsupportedCallOptions = "functions" | "function_call" | "tools";
+type TogetherAIUnsupportedCallOptions = "functions" | "function_call";
 
-export type ChatTogetherAICallOptions = Partial<
-  Omit<ChatOpenAICallOptions, TogetherAIUnsupportedCallOptions>
->;
+export interface ChatTogetherAICallOptions
+  extends Omit<ChatOpenAICallOptions, TogetherAIUnsupportedCallOptions> {
+  response_format: {
```
bracesproul, line 21: Switching this to an `interface` from `type` so my addition of `response_format` will override what `ChatOpenAICallOptions` has defined.

jacoblee93: This is fine but you can also do `& {response_format: Object}` for types

PR #4241, opened 2024-02-01
core[patch]: Emit stream log chunks without awaiting for the first stream chunk

### LINKED ISSUE #4230: Streaming tokens does not work
The documentation on how to do streaming is here https://js.langchain.com/docs/modules/agents/how_to/streaming

If I follow this, streaming still does not work. I've created an example repo https://github.com/rednebmas/langchain-streaming-bug

jacoblee93:
Fixes #4230

### DIFF
--- a/langchain/src/document.ts
+++ b/langchain/src/document.ts
@@ -1,9 +1,1 @@
-import { logVersion010MigrationWarning } from "./util/entrypoint_deprecation.js";
-
-/* #__PURE__ */ logVersion010MigrationWarning({
-  oldEntrypointName: "document",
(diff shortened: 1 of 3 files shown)


langchain/src/document.ts
```diff
@@ -1,9 +1 @@
-import { logVersion010MigrationWarning } from "./util/entrypoint_deprecation.js";
```
bracesproul, line 1: Why drop this?

jacoblee93: Was logging things when importing from hub due to the `load` call.

PR #4252, opened 2024-02-02
all[patch]: Add homepage to package.json for proper readmes

bracesproul:
This allows each NPM package displayed readme to match the readme of that specific package

Tested with weaviate [here](https://www.npmjs.com/package/@langchain/weaviate)

cc @jacoblee93 incase adding this has unwanted side effects?


jacoblee93:
I think fine, thanks!

PR #4258, opened 2024-02-02
core[minor]: Add XML output parser

bracesproul:


langchain-core/src/output_parsers/xml.ts
```diff
@@ -0,0 +1,78 @@
...
+2. If tags are not given, make them on your own.
+3. Remember to always open and close all the tags.
+
+As an example, for the tags ["foo", "bar", "baz"]:
+1. String "<foo>\n   <bar>\n      <baz></baz>\n   </bar>\n</foo>" is a well-formatted instance of the schema. 
+2. String "<foo>\n   <bar>\n   </foo>" is a badly-formatted instance.
+3. String "<foo>\n   <tag>\n   </tag>\n</foo>" is a badly-formatted instance.
+
+Here are the output tags:
+\`\`\
```
jacoblee93, line 15: Missing a backtick?

jacoblee93:
I think we can do streaming with something like this: https://github.com/matteodelabre/saxophone

PR #4268, opened 2024-02-02
core[patch]: Make documents serializable

jacoblee93:
`lc_namespace` taken from:

https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/load/mapping.py#L653


jacoblee93:
CC @dqbd @baskaryan

PR #4271, opened 2024-02-02
examples[patch],langchain[patch],community[patch]: Remove dup fake testing models

bracesproul:
Duplicate code, deleted and replaced imports with core imports
In the case of `langchain` I had to re-export for backcompat so I added a warning


libs/langchain-community/src/utils/testing.ts
bracesproul, line 1: no entrypoint, don't have to re-export

PR #4302, opened 2024-02-06
community[minor]: Add Dria retriever

erhant:
[Dria](https://dria.co/) is a hub of public RAG models for developers to both contribute and utilize a shared embedding lake. This PR adds a retriever that can retrieve documents from Dria.


jacoblee93:
Nice PR, thank you!

PR #4304, opened 2024-02-06
Fix typos and grammer errors in concepts.mdx

barisbll:
While reading the documentation I noticed some spelling errors, I fixed them in this commit

It is a small change but I would love a shoutout if possible https://twitter.com/barisbll_dev


jacoblee93:
Thank you! We definitely will!

PR #4312, opened 2024-02-07
ci[minor]: Add new test:unit:ci and build before tests in ci file

bracesproul:


turbo.json
```diff
@@ -76,6 +76,10 @@
       "outputs": [],
       "dependsOn": ["^build"]
     },
+    "test:ci": {
+      "outputs": [],
+      "dependsOn": ["test"]
```
bracesproul, line 81: No `^` is intentional here.

Adding the `^` will cause it to run at the top level, which depends on build. No `^` only runs it on workspaces, and none depend on build (unless they do... lol)

PR #4321, opened 2024-02-07
community[minor]: Added `SQLiteRecordManager`

### LINKED ISSUE #4281: [Feature Request] Upstash Vector VectorStore
Upstash has launched [Upstash Vector](https://upstash.com/docs/vector/overall/whatisvector). It would be great to have this added as a Vector Store option in the community package.

I'm happy to take a crack at this if there's appetite for it to be added. Just wanted to file the issue first.

### LINKED ISSUE #4320: [Feature Request] Done: Implement SQLite RecordManager
Currently there is a postgres and memory implementation for RecordManager. While I was working on the chat_langchain repository, I noticed that the python version of ingest.py supports sqlite for a RecordManager so I implemented one in the JS library as well. 

PR is ready, but wanted to open an issue first

jasonnathan:
Fixes #4320


bracesproul (approved):
Thanks for adding this! I just took a quick look & pushed up some small nits:
1. I added the dependency to `peerDeps` and made it optional, we always need to do this when adding another dependency for a given module.
2. I added an entrypoint inside `langchain.config.js`. When adding new modules, they must be added inside the `entrypoints` object in that file. Furthermore, if that module requires a dependency which is not already a hard dep, we need to add it to the `requiresOptionalDependency` array.

Besides those two things for the build, it looks pretty good to me! Thanks again for pushing this up

jasonnathan:
Oh gotcha! I'll keep that in mind, thanks

rxyshww:
how to use SQLiteRecordManager？thanks

PR #4332, opened 2024-02-07
langchain[patch]: Fix run on dataset method for built-in evaluators

jacoblee93:
CC @agola11 @hinthornw


langchain/src/smith/runner_utils.ts
```diff
@@ -59,6 +59,30 @@ class RunIdExtractor {
...
+
+  constructor() {
+    super();
+    this.runPromise = new Promise<Run>((extract) => {
+      this.runPromiseResolver = extract;
+    });
+  }
+
+  async persistRun(run: Run) {
+    this.runPromiseResolver(run);
```
hinthornw, line 78: does this always work for nested runs? Or do we need to check that it's the root trace?

langchain/src/smith/tests/run_on_dataset.int.test.ts
```diff
@@ -0,0 +1,256 @@
...
+      {
+        evaluatorType: "labeled_criteria",
+        criteria: "correctness",
+        feedbackKey: "correctness",
+        formatEvaluatorInputs: ({
+          rawInput,
+          rawPrediction,
+          rawReferenceOutput,
+        }: any) => {
+          console.log(rawInput);
```
hinthornw, line 227: Still want to keep these?

hinthornw (approved):

langchain/src/smith/tests/run_on_dataset.int.test.ts
jacoblee93: Sure could remove

langchain/src/smith/runner_utils.ts
jacoblee93: Not sure

Gr33nLight:
Commenting on this so that I know when this is merged.

Gr33nLight:
Noice, was gonna create an issue about this but it seems this is the fix 👍🏻 hope this is released soon.

jacoblee93:
Thanks @Gr33nLight! Am about to ship another version now, will post here when that's live.

jacoblee93:
This is now live in langchain@0.1.16 @Gr33nLight

Gr33nLight:
@jacoblee93 Great! Tested it and it works.

PR #4341, opened 2024-02-08
Add stream support for RunnableWithFallbacks

szuvi:
**What:**
- Implement stream iterator for RunnableWithFallbacks

***

**Why:**
- Calling `.stream()` on a `RunnableWithFallbacks` currently returns whole response (like `.invoke()`) not split into chunks.

### DIFF
--- a/langchain-core/src/runnables/base.ts
+++ b/langchain-core/src/runnables/base.ts
@@ -1915,6 +1915,61 @@ export class RunnableWithFallbacks<RunInput, RunOutput> extends Runnable<
     );
     throw firstError;
   }
+
+  async *_streamIterator(
+    input: RunInput,
+    options?: Partial<RunnableConfig>
+  ): AsyncGenerator<RunOutput> {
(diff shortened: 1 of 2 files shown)


jacoblee93:
Thank you so much for this!

We've been a bit hesitant to implement this because if the stream cuts off midway, downstream pieces may have already started processing chunks from the previous step, and if you pick up the stream with the next option in the fallbacks, the stream will start from the beginning.

I think the most we could do is catch errors on the initial stream creation, before the first chunk is emitted. @nfcampos, do you have any thoughts?

szuvi:
@jacoblee93 Wouldn't that be an issue with retry in general while streaming?

ianwoodfill:
> I think the most we could do is catch errors on the initial stream creation, before the first chunk is emitted. @nfcampos, do you have any thoughts?

This seems like a valuable capability IMO. Most of endpoint errors occur during the initial stream creation anyways, e.g. 429s, 500s.

hntrl:
Closing due to inactivity

PR #4349, opened 2024-02-09
core[minor]: Adds streamEvents method to runnables

jacoblee93:


jacoblee93:
Going to add a few more tests but should be good for a look @nfcampos

PR #4350, opened 2024-02-09
langchain[patch]: Fix: Prevent text-splitter `createDocments` metadata being overwritten.

iperzic:
This PR fixes a potential issue with text-splitter when `createDocuments` is called with an empty array for metadata.

If the `metadatas` argument is an empty array, it should be populated with an empty object. The previous logic was using `fill` to fill all array positions with the same object reference. If that object was to modify, it will be applied to every array element. Luckily, it was not noticeable because spread operator is being used.

This change guarantees that every element will have a unique object, thus preventing nasty object reference bugs.


jacoblee93:
Nice catch, thank you!

jacoblee93:
Can you run `yarn test`? It's breaking our text splitter test.

PR #4351, opened 2024-02-09
redis[patch]: Add RedisAddOptions to .fromDocuments method in order to perform upsert/update

hiitsmax:
By adding the optional parameter "docsOptions" to the RedisVectorStore.fromDocuments method we are now able to define the index schema directly from there. 

This should be useful when performing upserts or updating already existing keys in general.

### DIFF
--- a/libs/langchain-redis/src/vectorstores.ts
+++ b/libs/langchain-redis/src/vectorstores.ts
@@ -282,7 +282,7 @@ export class RedisVectorStore extends VectorStore {
     metadatas: object[] | object,
     embeddings: EmbeddingsInterface,
     dbConfig: RedisVectorStoreConfig,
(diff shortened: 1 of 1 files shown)


jacoblee93:
Thanks for this! Looks reasonable but the package in `langchain-community` is deprecated in favor of `langchain-redis`. Could you move this change there and also add it to `fromTexts`?

hiitsmax:
Thanks for the hint!

Now it should be all in the right place ( `fromTexts ` included) (:

bracesproul (approved):
Thanks for adding this! I pushed up a commit removing the updates to the `@langchain/community` module. Since it's deprecated, we don't want to add more features and instead want to encourage people to update to the `@langchain/redis` package. Thanks again!

PR #4361, opened 2024-02-10
docs[patch]: Fixes the build but not a long term solution

### LINKED ISSUE #4360: Syntax error in human_in_the_loop.mdx after quatro processing
This is a bug in quatro. human_in_the_loop.ipynb is causing an error with the docs build. Somehow the correct syntax in the ipynb file:
```mdx
import CodeBlock from "@theme/CodeBlock";
import HumanFeedback from "@examples/use_cases/human_in_the_loop/accept-feedback.ts";
```
is wrongly interpreted as:
```mdx
import CodeBlock from “(**theme/CodeBlock?**)”; import HumanFeedback
from “(**examples/use_cases/human_in_the_loop/accept-feedback.ts?**)”;
```

jasonnathan:
Fixes #4360


bracesproul (commented):
Hey, unfortunately we can't do this because `readline` is not compatiable with deno (the runtime which these notebooks use). However, I'll look into the docs issue this weekend and figure out a fix.

jasonnathan:
Maybe try without cache on your end?

PR #4368, opened 2024-02-11
langchain[patch]: Upgrade Google Cloud Storage dependency

### LINKED ISSUE #4305: Outdated Google Cloud Storage Dependency
I'm using the Langchain JS library for an application, and have noticed it's pinning a version of Google Cloud Storage that is quite outdated:

` WARN  Issues with peer dependencies found. 
`
`
└─┬ langchain 0.1.13
`
`
  └── ✕ unmet peer @google-cloud/storage@^6.10.1: found 7.7.0
`

Is there a reason an old version is pinned? Or can we easily update it to the current v7 release? I'm using the Cloud Storage library directly and was doing so before integrating Langchain, so would of course prefer not to downgrade and deal with any resulting API changes that may be involved.

JackEdgar:
Update the Google Cloud Storage dependency to the latest version. The parts of the API that are being utilised in gcs.ts seem to be unaffected, so this is just a small improvement that fixes #4305.


jacoblee93:
Thank you! If this does cause problems, we can split it out into a separate package.

jacoblee93:
Actually, if it doesn't matter - can you change this to:

```
^6.10.1 || ^7.7.0
```

And update the entry in `peerDependencies`?

jacoblee93:
Thank you!

PR #4369, opened 2024-02-11
langchain[patch]: Update cypher qa prompt

tomasonjo:
Unlike vector results, the LLM has to completely trust the context of a graph database result, even if it doesn't provide whole context. We tried with instructions, but it seems that adding a single example is the way to go to solve this issue.


jacoblee93:
Thank you!

PR #4372, opened 2024-02-12
community[patch]: allow to modify OpenSerach number of shards/replicas

ansnoussi:
- Add possibility to specify the number of shards/replicas when creating indices for OpenSerach vector store.


libs/langchain-community/src/vectorstores/opensearch.ts
```diff
@@ -18,6 +18,8 @@ interface VectorSearchOptions {
   readonly m?: number;
   readonly efConstruction?: number;
   readonly efSearch?: number;
+  readonly numberOfShards?: number;
```
jacoblee93, line 21: Slight preference for `shardCount` and `replicaCount` - what do you think?

jacoblee93:
Looks great, thank you!

PR #4374, opened 2024-02-12
Safety-blocked Gemini response responds with no `parts`

jarib:
This adds a failing test based on the response.


jacoblee93:
Thanks for this!

What should happen here? Thrown error with a better message?

jarib:
This is my first time working with models that have this kind of structured safety blocking – most would just reply with a refusal to respond, I think. But yes, throwing (`SafetyError`?) with a better message or `.reason` from within the Google/Gemini code makes sense to me. 

I guess the alternative would be some sort of empty response and a mechanism for the user to check the reason. Not sure what makes most sense given how other models handle this.

See also https://github.com/langchain-ai/langchainjs/issues/4375 for this PR.

afirstenberg:
I think it is tricky because there could be some or a partial response before the safety trigger kicks in. And each chunk in a streamed response has its own safety ratings.

Certainly there are some cases, however, where no text is returned due to safety (or other reasons).

I think the tests are good, or at least a good indication that we need to address this.

@jacoblee93 - Can you assign this and #4375 to me please?

afirstenberg:
While the fix in #4583 doesn't use this exact code, it certainly was inspired by the test and the issue.
Your call if you want this exact test in it, or if the tests I put in place are sufficient.

PR #4379, opened 2024-02-12
core[minor]: Added document compressors base classes to core

bracesproul:


langchain-core/src/retrievers/document_compressors/base.ts
```diff
@@ -0,0 +1,90 @@
+import { type DocumentInterface } from "../../documents/document.js";
+import { BaseDocumentTransformer } from "../../documents/transformers.js";
+import { Callbacks } from "../../callbacks/manager.js";
+
+/**
+ * Base Document Compression class. All compressors should extend this class.
```
bracesproul, line 6: Add conceptual doc here @bracesproul

PR #4382, opened 2024-02-12
langchain[patch]: Bump langchainhub dep to latest

### LINKED ISSUE #4289: specifying commit hash in `hub.pull` is not working
# Description

For example, these two should be different:

```js
import { pull } from "langchain/hub"

await pull("langchain-ai/chain-of-density")
await pull("langchain-ai/chain-of-density:ba34ae10")
```

But the result is the same.

## A simple reproduction

```js
> (await pull("langchain-ai/chain-of-density:ba34ae10")).promptMessages[0].prompt.template === (await pull("langchain-ai/chain-of-density")).promptMessages[0].prompt.template
// true
```

---

Note that the python SDK is working, but in the JavaScript SDK it seems specifying a commit hash is useless. I don't know why.

bracesproul:
closes #4289

### DIFF
--- a/langchain/package.json
+++ b/langchain/package.json
@@ -1512,7 +1512,7 @@
     "js-tiktoken": "^1.0.7",
     "js-yaml": "^4.1.0",
     "jsonpointer": "^5.0.1",
-    "langchainhub": "^0.0.8",
+    "langchainhub": "~0.0.8",
     "langsmith": "~0.0.59",
     "ml-distance": "^4.0.0",
     "openapi-types": "^12.1.3",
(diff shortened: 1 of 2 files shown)


langchain/package.json
```diff
@@ -1512,7 +1512,7 @@
     "js-tiktoken": "^1.0.7",
     "js-yaml": "^4.1.0",
     "jsonpointer": "^5.0.1",
-    "langchainhub": "~0.0.6",
+    "langchainhub": "^0.0.8",
```
jacoblee93, line 1515: Should be tilde to allow for patch updates

Also, if a new version has been published, new installs will get that one?

PR #4387, opened 2024-02-13
cloudflare[patch]: Add metadata filtering support to Cloudflare Vectorize

davkorss:
Support for metadata filtering using Cloudflare Vectorize


jacoblee93:
Thank you!

Issue #4396, opened 2024-02-13
Unsafe usage of string interpolation in SQL queries


MJDeligan:
APIs based on Postgres use the [node-postgres](https://node-postgres.com/) library. This library's parameterized queries work for values but not for identifiers ([source](https://node-postgres.com/features/queries#parameterized-query)). In many places table and column identifiers can be set by developers. In these cases simple string interpolation is used to build the dynamic query. In PGVectorStore for example:

```typescript
await this.pool.query(`
      CREATE TABLE IF NOT EXISTS ${this.tableName} (
        "${this.idColumnName}" uuid NOT NULL DEFAULT uuid_generate_v4() PRIMARY KEY,
        "${this.contentColumnName}" text,
        "${this.metadataColumnName}" jsonb,
        "${this.vectorColumnName}" vector
      );
`);
```

node-postgres also allows for multiple statements to be executed in one query. See [here](https://stackoverflow.com/a/68703509), also confirmed by myself.

This means that a table name such as 

```sql
yougotpwned (
    id SERIAL PRIMARY KEY
);
<malicious queries here>
CREATE TABLE IF NOT EXISTS legittable
```

could run arbitrary queries on the database through SQL injection. 

Developers creating multi-tenant applications could face this problem if they use any kind of user provided content to generate table or column names. Developers may trust that this library (langchainjs) securely sanitizes these inputs.

In 8.11.0, pg-node introduced an `escapeIdentifier` utility which may be used to sanitize table and column names, although someone with more expertise in security may be able to better determine its viability. Alternatively, the [pg-format](https://github.com/datalanche/node-pg-format) library can be used, however this library has been archived since last year with no official successor fork.

PR #4398, opened 2024-02-13
docs[patch]: Add warnings about SQL Injection for Postgres integrations

MJDeligan:
Concerning #4396 .

This PR escapes identifiers used in SQL queries made with node-postgres in `PGVectorStore` and `PostgresRecordManager`.  This implementation handles inputs that previously caused arbitrary queries to be executed. 

This can be a breaking change for some users, since the identifiers users pass to the API may be different from the identifiers that are used in the queries. Identifiers containing uppercase letters would have previously been converted to all lowercase but are now conserved. Users can still use the values saved in an object's field directly in their query for example: 

```typescript
const pgVectorStore: PGVectorStore;

await pool.query(`SELECT * FROM ${pgVectorStore.tableName};` // tableName is escaped
```

`AnalyticDB`  also uses node-postgres with unsafe string interpolation for identifiers, but I could not find a way to test it locally and do not have access to an instance.

### DIFF
--- a/docs/core_docs/docs/integrations/vectorstores/analyticdb.mdx
+++ b/docs/core_docs/docs/integrations/vectorstores/analyticdb.mdx
@@ -44,6 +44,11 @@ npm install @langchain/openai @langchain/community
 
 ## Usage
 
+::::danger Security
+User-generated data such as usernames should not be used as input for the collection name.  
+**This may lead to SQL Injection!**
+::::
+
 import UsageExample from "@examples/indexes/vector_stores/analyticdb.ts";
 
 <CodeBlock language="typescript">{UsageExample}</CodeBlock>
--- a/docs/core_docs/docs/integrations/vectorstores/pgvector.mdx
+++ b/docs/core_docs/docs/integrations/vectorstores/pgvector.mdx
@@ -36,6 +36,11 @@ You can find more information on how to setup `pgvector` in the [official reposi
 
 ## Usage
 
+::::danger Security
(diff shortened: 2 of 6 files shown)


jacoblee93:
Ah gotcha - I think our stance on was that is it's kind of XSS'ing yourself since you're initializing the constructor with constants you pass yourself vs documents from userland. I suppose there's a niche case where someone could initialize a table based on user input?

Since it's technically breaking - do you have a strong feeling towards just adding warnings in the documentation instead?

MJDeligan:
Yeah, I was thinking about the niche case where developers might want a table per user/organization. A warning in the docs would suffice though. Maybe it would make sense to introduce in a breaking patch, since there's also no way to recommend an alternative where developers might sanitize themselves, because the queries are inconsistent with the use of double quotes.

Would you like me to reset the branch and update the docs here or should I close the PR and open a new one?

jacoblee93:
Thanks - feel free to reuse this same branch. I'd also be down to add a configuration flag that enables this behavior if that sounds like a good compromise?

MJDeligan:
Updated the docs with warnings (there's no section for RecordManagers, maybe one should be added?). 

I looked into using a configuration flag. Because of the inconsistent use of double quotes and identifiers being built from other identifiers there's no straight-forward way to dynamically escape the identifiers. It would just turn the code into a mess for niche use cases. Warnings in the docs should be enough and we can still consider introducing this  in a later patch.

jacoblee93:
Thank you so much!

PR #4409, opened 2024-02-14
community[patch]: Allow using a pool for PostgresRecordManager

### LINKED ISSUE #4407: Allow using a pool for PostgresRecordManager
The constructor of PostgresRecordManager in libs/langchain-community/src/indexes/postgres.ts always creates a new pool.

Please allow using a preexisting pool. Else, if you want to manage multiple indices in one DB you'll run out of connections very quickly.

mkesper:
Enable to use a predefined pool for PostgresRecordManager, allowing to use multiple indices per database. Without, each index will create a new pool, leading to Postgres connection exhaustion.

Fixes #4407

### DIFF
--- a/libs/langchain-community/src/indexes/postgres.ts
+++ b/libs/langchain-community/src/indexes/postgres.ts
@@ -6,7 +6,8 @@ import {
 } from "./base.js";
 
 export type PostgresRecordManagerOptions = {
-  postgresConnectionOptions: PoolConfig;
+  postgresConnectionOptions?: PoolConfig;
+  pool?: Pool
   tableName?: string;
 };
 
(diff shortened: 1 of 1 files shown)


libs/langchain-community/src/indexes/postgres.ts
```diff
@@ -20,9 +21,19 @@ export class PostgresRecordManager implements RecordManagerInterface {
   namespace: string;
 
   constructor(namespace: string, config: PostgresRecordManagerOptions) {
-    const { postgresConnectionOptions, tableName } = config;
+    const { postgresConnectionOptions, pool, tableName } = config;
     this.namespace = namespace;
-    this.pool = new pg.Pool(postgresConnectionOptions);
+    if (pool instanceof Pool) {
```
MJDeligan, line 26: violates eslint rule. checking that pool is not undefined should be fine

libs/langchain-community/src/indexes/postgres.ts
```diff
@@ -20,9 +21,19 @@ export class PostgresRecordManager implements RecordManagerInterface {
...
+      this.pool = pool
+    } else {
+      try {
+        this.pool = new pg.Pool(postgresConnectionOptions);
+      } catch (err: unknown) {
```
MJDeligan, line 31: pool constructor does not throw when config is undefined/invalid as far as I can tell. checking that pool and config aren't both undefined should work

MJDeligan (commented):
Hey @mkesper , thanks for adding this, great thinking! Left a couple of comments (not a maintainer, wrote the original implementation)

mkesper:
@MJDeligan Thanks, you're right, that code could be improved. :) Since I posted I'm not anymore sure what's the best way to handle the problem that connections are held for a long time. Could also be a problem in the code calling the RecordManager not closing it. (A hint for that in the examples would be really nice).

MJDeligan:
@mkesper I think you're right that having multiple pools will lead to the creation of too many clients. Being able to pass your own pool is definitely good. 

Not ending the pool(s) will of course worsen this. Sorry about the documentation, the long example is a little unwieldy. Maybe a very minimal example at the beginning would be good to add.

jacoblee93:
Thank you for the review @MJDeligan!

I think technically the instanceof check is ok here (with a linter ignore) because it's Node only. Have seen instanceof checks be unreliable in environments that use bundlers like Next.js since you can wind up with multiple declarations in different chunks. But probably best to avoid yes.

PR #4410, opened 2024-02-14
Remove deprecated call of serializable.js

mkesper:
Use @langchain/core module instead.

./load/serializable.js seems deprecated?


jacoblee93:
Thank you!

PR #4418, opened 2024-02-15
community[minor]: add CassandraKVStore

mieslep:
**Adds:**
* stores/CassandraKVStore 
* utils/CassandraTable 
* Support for IN operator

**Deprecations:**
* Interfaces {Column, Index, Filter, WhereClause} now exported from utils/cassandra rather than vectorstores/cassandra


mieslep:
@jacoblee93 I'm not sure what is the docs failure - the doc preview in the Dev Container doesn't work for me (Windows fwiw, but should be isolated from that with the container right?). The process for building/previewing docs isn't terribly clear to me from CONTRIBUTING.md :/

mieslep:
@jacoblee93 figured out the docs problem (`yarn docs` worked great, and allowed me to preview the docs too which is nice! I was trying previously to use the MDX preview, but that's not worked for me)...

libs/langchain-community/src/vectorstores/cassandra.ts
```diff
@@ -11,46 +8,29 @@ import {
...
-
-export interface Filter {
-  name: string;
-  value: unknown | [unknown, ...unknown[]];
-  operator?: string;
-}
-
-export type WhereClause = Filter[] | Filter | Record<string, unknown>;
+// Above Column, Index, Filter, WhereClause are exported from "../utils/cassandra.js"
+// Code that previously imported from this file should now import from "../utils/cassandra.js"
```
jacoblee93, line 21: Why not re-export these?

`export { Filter, Index, Column };`

https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-openai/src/chat_models.ts#L44

libs/langchain-community/src/vectorstores/cassandra.ts
```diff
@@ -1,8 +1,5 @@
 /* eslint-disable prefer-template */
-import {
-  AsyncCaller,
-  AsyncCallerParams,
-} from "@langchain/core/utils/async_caller";
+import { v4 as uuidv4 } from "uuid";
```
jacoblee93, line 2: Big diff but I assume this is mostly just moving code around?

libs/langchain-community/src/utils/cassandra.ts
```diff
@@ -246,3 +350,881 @@ export class CassandraClientFactory {
...
+        // Wrap placeolders in a () if the operator is IN
+        if (normalizedOperator === "IN") {
+          return `${name} ${operator} (${placeholders})`;
+        } else {
+          return `${name} ${operator} ${placeholders}`;
+        }
+      }
+    });
+
+    return `WHERE ${whereConditions.join(" AND ")}`;
```
jacoblee93, line 974: Any injection risk here?

jacoblee93: Or is the assumption that nothing here is user-specified? If so, might be good to add a quick warning to docs.

jacoblee93:
Thank you! I think looks broadly reasonable, see comments.

libs/langchain-community/src/vectorstores/cassandra.ts
mieslep: The act of moving things created all sorts of build challenges, and at some point removing these exports allowed the build to work (whereas their inclusion as a re-export did not work).  Seems to be building now, I far prefer deprecation to outright breaks :) Change pushed.

libs/langchain-community/src/vectorstores/cassandra.ts
mieslep: Yes, the table-related logic moved to utils and consolidated into the `CassandraTable` class, so `vectorstore` was most impacted; `message` then got updated to make use of the `CassandraTable` class rather than accessing the client directly.

libs/langchain-community/src/utils/cassandra.ts
mieslep: All external inputs are in bind variables, so "end-user" inputs are protected from injection. LangChain users (developers) remain fully able to shoot themselves in the foot 18 different ways :)

jacoblee93:
Thank you!

PR #4430, opened 2024-02-16
langchain[minor]: Fix missing traces when traceable used in runOnDataset, API update

dqbd:
1. Getting rid of the list of eval types as last arg syntax, smaller API surface
2. Added more strict type constraints on which evaluators we do support
3. Add EmbeddingDistance() config factory
4. Introduce a subclass of RunTree, which uses LangChain Callback Manager to handle weird cases with LS batching and proper order
5. Fix stringification of errors for other non-chain callbacks


jacoblee93:
Thank you!

PR #4434, opened 2024-02-16
community[patch]: add neo4j-graph generics, don't swallow query execution errors

adam-cowley:
* Added routing mode to allow the app to specify whether the query is routed to the leader or cluster - (defaults to `WRITE` for backwards compatibility)
* Modified `Neo4jGraph.query()` to accept generics for cleaner code
* Stopped `.query() from swallowing errors - a source of much personal confusion & frustration
* Bumped `neo4j-driver` version

Although the routing method defaults to Write, this isn't best practice.  Especially, if you consider the security note in the comments.  If possible, this should default to `READ`.

### DIFF
--- a/libs/langchain-community/package.json
+++ b/libs/langchain-community/package.json
@@ -148,7 +148,7 @@
     "lunary": "^0.6.11",
     "mongodb": "^5.2.0",
     "mysql2": "^3.3.3",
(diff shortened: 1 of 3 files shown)


jacoblee93:
Thank you!

PR #4451, opened 2024-02-17
feat(langchain-community): Allow custom model endpoints to be used with GoogleVertexAI LLM class

x-rudolfs:
### Description

According to my research it currently is not possible to use the `GoogleVertexAI` class with a fine-tuned model, that was fine tuned on the Google Cloud VertexAI platform. 

For fine-tuned models, Google creates their own endpoints which have exactly the same APIs used by their own published models. 

To solve this issue in one of my projects, I had to extend from the `GoogleVertexAI` as a base-class and override the function used to build urls (GoogleVertexAILLMConnection.buildUrl).

I don't believe that approach is good in the long run, so I decided to create this PR. 

I have spent some time thinking how would I fix this if I was one of the LangChain maintainers, but couldn't come up with anything "smart", so I opted for the "make work" approach here. 

This might not be a perfect solution, but it might make it easier for engineers to build things while a better alternative appears.

### DIFF
--- a/libs/langchain-community/src/types/googlevertexai-types.ts
+++ b/libs/langchain-community/src/types/googlevertexai-types.ts
@@ -14,6 +14,22 @@ export interface GoogleVertexAIConnectionParams<AuthOptions>
 
   /** The version of the API functions. Part of the path. */
   apiVersion?: string;
+
+  /**
+   * Whether you are planning to connect to a public VertexAI model
+   *
+   * In case you are connecting to a default model, that was published by Google
+   * and lives under the /publishers/google/models/:model path in Google Cloud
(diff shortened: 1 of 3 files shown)


jacoblee93:
Thanks for this! Could this instead be just a fully custom URL that overrides `buildUrl`?

x-rudolfs:
> Thanks for this! Could this instead be just a fully custom URL that overrides `buildUrl`? Or does the convenience outweigh that?

Yes, that makes sense. I'll adjust the PR + fix the code formatting issue. Thank you for the suggestion!

jacoblee93:
Looks fine to me, thank you!

Issue #4282, opened 2024-02-04
Support `keep_alive` parameter for Ollama APIs


andrewnguonly:
The `keep_alive` parameter was added in Ollama `v0.1.23`: https://github.com/ollama/ollama/blob/v0.1.23/api/types.go#L45

At first, this seems like a straightforward change, but the Ollama API represents `keep_alive` internally as a Go `Duration`. Not sure how this type translates in the LangChainJS API. `Duration` supports string representations and has an upper bound.

PR #4454, opened 2024-02-18
community[patch]: Add `keep_alive` parameter to Ollama API.

andrewnguonly:
Fixes # https://github.com/langchain-ai/langchainjs/issues/4282

### DIFF
--- a/libs/langchain-community/src/chat_models/ollama.ts
+++ b/libs/langchain-community/src/chat_models/ollama.ts
@@ -66,6 +66,8 @@ export class ChatOllama
 
   baseUrl = "http://localhost:11434";
 
+  keepAlive = "5m";
+
   embeddingOnly?: boolean;
 
   f16KV?: boolean;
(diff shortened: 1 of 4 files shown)


jacoblee93:
Thanks for this!

Would it make more sense to queue up a `langchain-ollama` integration package that wraps their new SDK directly and try to get that in sometime in the next few days?

alvarosevilla95:
> Thanks for this!
> 
> Would it make more sense to queue up a `langchain-ollama` integration package that wraps their new SDK directly and try to get that in sometime in the next few days?

Just 2 cents from a random user, but I was looking forward to seeing this added. While a full integration with their SDK sounds ideal, if this could be merged now rather than wait on that I would greatly apprecieate it.

jacoblee93:
@andrewnguonly yes I think @alvarosevilla95 is right in that this should be added to the chat model and embedding class as well.

As for what makes a partner package - deps are a big part of it, yes. Community has a ton of peer deps that can cause incidental issues due to the way that npm handles peer dependency conflicts, and we can version breaking SDK changes more intelligently with an integration package.

I could timebox a new package to today and if I run into difficulty, I could merge this one and #4466 instead and release it if that sounds good?

jacoblee93:
Ok, sounds like it's Node only for a little while longer - let's merge these! Will get it in by the end of the day.

jacoblee93:
Thank you!

jacoblee93:
This and #4466 are live in the latest community release! Stay tuned for a separate Ollama package

PR #4455, opened 2024-02-19
docs[patch]: Docs update: update cloudflare_vectorize.mdx

mggger:
Incorrect file extension type, causing the example to be unexecutable.


jacoblee93:
Ah, thanks!

PR #4450, opened 2024-02-17
community[minor]: Postgres Integration for Chat Memory

MJDeligan:
Closes #4123 .

Chat message history backed by Postgres. Implementation based off of the [Python version](https://github.com/langchain-ai/langchain/blob/master/libs/community/langchain_community/chat_message_histories/postgres.py) and other memory store integration in langchainjs.

The constructor has an optional flag to escape the table name as suggested for other postgres integrations by @jacoblee93  [here](https://github.com/langchain-ai/langchainjs/pull/4398#issuecomment-1945070785).

Includes integration tests from oter

PR #4459, opened 2024-02-20
community[minor]: Add Astra DB chat message history store

mfortman11:
Adds `AstraDBChatMessageHistory` as a message store


examples/src/memory/astradb.ts
```diff
@@ -0,0 +1,48 @@
+import { BufferMemory } from "langchain/memory";
```
jacoblee93, line 1: We're trying to do all new examples with LCEL - could you update this to be more like: 

https://github.com/langchain-ai/langchainjs/pull/4450/files#diff-65d27e9b05b811faae72376bd775edc281f6b37233130c14f38613089a580554R1

libs/langchain-community/src/stores/message/astradb.ts
```diff
@@ -0,0 +1,116 @@
+import { BaseListChatMessageHistory } from "@langchain/core/chat_history";
```
jacoblee93, line 1: Can you also add a separate entrypoint + mark it as optional to `langchain-community/langchain.config.js` and run `yarn build` in community?

https://github.com/langchain-ai/langchainjs/blob/main/CONTRIBUTING.md#adding-an-entrypoint

jacoblee93:
Thanks! See comments.

mfortman11:
@jacoblee93 thanks for the feedback should be all set now

jacoblee93:
Looks great, thanks!

PR #4466, opened 2024-02-20
community[patch]: Allow custom headers on Ollama / ChatOllama / OllamaEmbeddings 

### LINKED ISSUE #4465: Custom headers for Ollama / ChatOllama
It seems there is no way to pass custom http headers to a (Chat)Ollama llm. This can be useful in certain cases, and it is possible on the python implementation.

It could be an additional argument passed in the llm instance construction containing the extra headers that should be added to the requests

alvarosevilla95:
Adds a constructor param to Ollama / ChatOllama / OllamaEmbeddings allowing custom headers to be included on the API requests

Closes #4465


jacoblee93:
Thank you!

PR #4480, opened 2024-02-21
community[patch]: Remove quotes around PostgresRecordManager table create statement

alan-bentley-dsg:
Wrapping the table name in quotes locks the creation of tables to the default schema in postgres. Removing the quotes enables the creation of tables in other schemes enabling flexibility

Fixes # (issue)


jacoblee93:
This will probably break people who supply a table with a capital letter right? Could we instead also support an additional param?

CC @MJDeligan

MJDeligan:
> This will probably break people who supply a table with a capital letter right? Could we instead also support an additional param?

Yes, it would be a breaking change, but I like the idea of being able to pass a schema param for flexibility!

alan-bentley-dsg:
@jacoblee93 @MJDeligan thanks for the feedback! I will make an update later this morning  to provide an optional parameter for schema, if not supplied we can fall back to the original implementation with quotes

alan-bentley-dsg:
Updated with schema param @MJDeligan @jacoblee93

MJDeligan:
I think the combined name of schema and table needs to be used in ever query right? The tests fail for me locally, because the ```CREATE INDEX``` queries still have the double quotes around the identifier even though the identifier itself already has it. If I fix that then I get errors because the table does not exist, probably because the schema is not added in other queries? I'm not too familiar with postgres' search path across schemas. 

We could add the schema param in the constructor to use it throughout the RecordManager.

A small sanity test for this would also be nice.

alan-bentley-dsg:
Good call out on the double quotes @MJDeligan , that was a typo on my end. I fixed this and also refactored the param out from the createSchema method to the constructor as an optional argument and applied this to the remaining queries in the Record Manager.  Also wrote a quick smoke test as requested to validate this change.

MJDeligan:
Thanks for the update! Functionality looks good, tests are passing. Couple of things: The schema parameter in the constructor can be part of the config object param, since it's related to the rest of the config. It would be nice to have the finalTableName as a property on the record manager, so that we don't repeat the logic for each query and we make it easier to check what the actual relation name is. For the test, it would be good if a new variable was used for the record manager with schema and we'd immediately end it within the test. That way we do not need to reinitialize after each test. Thoughts? @jacoblee93 

Good work!

jacoblee93:
I agree with everything @MJDeligan said! Not a fan of positional args personally unless they're very thoughtful as you can end up with horrible stuff like this:

https://github.com/langchain-ai/langchainjs/blob/main/langchain-core/src/language_models/chat_models.ts#L185

alan-bentley-dsg:
Sounds good! I agree the config would be a more appropriate location for the schema option - i will update this and the test when I’m back in front of my machine Monday. Thanks again for your feedback!

todo:
- [ ] refactor schema out of constructor and into config object 
- [ ] Set ‘finalTableName’ as property on ‘PostgresRecordManager’ 
- [ ] Update test with new ‘PostgresRecordManager’ variable to avoid reinitialization; also close connection within test

alan-bentley-dsg:
Updated with requested changes @jacoblee93 @MJDeligan

jacoblee93:
Looks great, thank you!

PR #4485, opened 2024-02-21
groq[minor]: Fix Readme example

gradenr:
Fixes the example in the Groq readme.  It refers to an undefined variable.


bracesproul (approved):
whoops, thanks for catching this!

PR #4502, opened 2024-02-23
community[patch]: fix astra initialize error

nicoloboschi:
Currently in the `AstraDBVectorStore#initialize` method if there is an error while creating the collection, the error is suppressed, assuming the collection already exists. This is wrong because:
- if the collection already exists with the same options, the Astra Data API won't return errors
- In case of auth, network or server errors, the user can't really see the errors and no error is raised

### DIFF
--- a/libs/langchain-community/src/vectorstores/astradb.ts
+++ b/libs/langchain-community/src/vectorstores/astradb.ts
@@ -88,16 +88,10 @@ export class AstraDBVectorStore extends VectorStore {
    * @returns Promise that resolves if connected to the collection.
    */
   async initialize(): Promise<void> {
-    try {
-      await this.astraDBClient.createCollection(
+    await this.astraDBClient.createCollection(
         this.collectionName,
         this.collectionOptions
-      );
-    } catch (error) {
-      console.debug(
-        `Collection already exists, connecting to ${this.collectionName}`
-      );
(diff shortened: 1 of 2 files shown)


jacoblee93:
Thank you!

Gr33nLight:
Hello regarding this, I noticed a pretty big impact of the call ` this.astraDBClient.createCollection ` in the ` initialize `  in terms of response time.  Even if the collection already exists, this call takes multiple seconds to complete (2.2 to 2.5s im my testing). This is not ideal if the collection is already present, may a suggest in a pull request the option to disable automatic collection creation?

@nicoloboschi  @jacoblee93

nicoloboschi:
@Gr33nLight I don't think it's related to the pull itself. the API calls are the same.
Checking if the collection already exists would take the same time.
Skipping the collection creation is very risky from a usability point of view 🤔

Gr33nLight:
Yes I agree in regards to the usability, It is more of a specific case, in our product speed is a priority and having 2s added to every request is not the best. We are using a modified version so its not that big of a issue for us, but I wanted to point this out anyways as I was not experiencing this slow down with the other adapters. 
Maybe I'll check with the guys at Datastax to see if something can be done at client level, to maybe speed up some calls.

Issue #4503, opened 2024-02-23
Add schema support in `ensureTableInDatabase`


n1md7:
https://github.com/langchain-ai/langchainjs/blob/b2151ec4d546ff71c7045883251e839042fb4485/libs/langchain-community/src/vectorstores/typeorm.ts#L224

When schema is used other than `public` it is not working since it assumes the default one.

Maybe adding an explicit name would be better?

Something like `CREATE TABLE IF NOT EXISTS ${this.schemaName}.${this.tableName}`

n1md7:
@dosu-bot okay, I can work on a PR

PR #4504, opened 2024-02-23
feat(community): Add schema support for `ensureTableInDatabase` in typeorm vectorstore.

n1md7:
Added schema name to override it when needed.

It's backward-compatible and should not break existing functionality. 

Fixes https://github.com/langchain-ai/langchainjs/issues/4503

### DIFF
--- a/libs/langchain-community/src/vectorstores/typeorm.ts
+++ b/libs/langchain-community/src/vectorstores/typeorm.ts
@@ -13,6 +13,7 @@ import { getEnvironmentVariable } from "@langchain/core/utils/env";
 export interface TypeORMVectorStoreArgs {
   postgresConnectionOptions: DataSourceOptions;
   tableName?: string;
+  schemaName?: string;
   filter?: Metadata;
   verbose?: boolean;
 }
@@ -38,6 +39,8 @@ export class TypeORMVectorStore extends VectorStore {
 
   tableName: string;
 
+  schemaName: string;
+
   documentEntity: EntitySchema;
 
   filter?: Metadata;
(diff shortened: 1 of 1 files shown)


jacoblee93:
Thank you!

jacoblee93:
Oh hey - should this be in quotes like this?

https://github.com/langchain-ai/langchainjs/pull/4480

MJDeligan:
Would setting the default schema like this not be breaking for users who've set a search path other than "public" ?

n1md7:
> Oh hey - should this be in quotes like this?
> 
> #4480

I'll change it

n1md7:
> Would setting the default schema like this not be breaking for users who've set a search path other than "public" ?

I'll verify that

jacoblee93:
We could use a different query based on the presence/absence of `schemaName`? I think that would be the way to go.

jacoblee93:
And thanks for flagging @MJDeligan!

n1md7:
> We could use a different query based on the presence/absence of `schemaName`? I think that would be the way to go.
> 
> Sort of like this: #4543

Oh right, so when the schema is present I prepend it with the table name otherwise it stays as it is.

MJDeligan:
Looks good! I'm not familiar with TypeORM. How does it know to use the schema for the other operations? Would that have to passed as part of the postgres connection options?

n1md7:
> Looks good! I'm not familiar with TypeORM. How does it know to use the schema for the other operations? Would that have to passed as part of the postgres connection options?

Yes, in TypeORM schema can be specified when driver is Postgres. When it's omitted it fallbacks to **public** at least in most cases but the logic can be seen here https://github.com/typeorm/typeorm/blob/master/src/driver/postgres/PostgresDriver.ts#L360

n1md7:
@jacoblee93 wdyt?

benjamincburns:
Thanks for the contribution, @n1md7. Sorry it took us so long to get this merged!

PR #4505, opened 2024-02-23
Remove unneeded Async from createOpenAIToolsAgent fn

Pckool:
This function does not need to be async. I want to call this in a class constructor and it's more complicated because this isn't a sync fn

### DIFF
--- a/langchain/src/agents/openai_tools/index.ts
+++ b/langchain/src/agents/openai_tools/index.ts
@@ -95,7 +95,7 @@ export type CreateOpenAIToolsAgentParams = {
  * });
  * ```
  */
-export async function createOpenAIToolsAgent({
+export function createOpenAIToolsAgent({
   llm,
   tools,
   prompt,


jacoblee93:
It's for consistency with some of the other `createXAgent` methods which do some prompt partialing - not ideal but I think we'd like to keep it that way.

You could use an initializer method instead like this?

https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-pinecone/src/vectorstores.ts#L352

```ts
  static async fromDocuments(
    docs: Document[],
    dbConfig: PineconeStoreParams
  ): Promise<PineconeStore> {
    const args = dbConfig;
    args.textKey = dbConfig.textKey ?? "text";

    const instance = new this(embeddings, args);
    await instance.addDocuments(docs);
    return instance;
  }
```

Or take the agent runnable as a parameter.

Pckool:
Ah okay that makes sense, thanks for the review!

PR #4515, opened 2024-02-26
Add links to free DataCamp Code-Alongs

filipsch:
This PR adds links to some more free resources for people to get acquainted with the topics of the cookbooks without having to configure their system.


jacoblee93:
Hey! Thanks so much for this.

Could we move this into the `details` tag as an option alongside Scrimba? These guides would also be great to add here:

https://js.langchain.com/docs/additional_resources/tutorials

filipsch:
Hey @jacoblee93! 
- Good call on including it in the details tag. I went ahead and restructured things.
- I also added the guides to the tutorials section as you suggested.
Thanks

jacoblee93:
Thank you!

jacoblee93:
Oh, hey sorry just dug in and it looks like it's a paid course... would prefer to not have that in the details pages since that's the case.

I'm ok with having it under `Tutorials` but could you add something about the course not being free?

filipsch:
Yes it's a Python course. It's free (but a registration is required). Is there a better location to include links to Python content?

jacoblee93:
Ah ok, yes then the LangChain Python repo:

https://github.com/langchain-ai/langchain

PR #4516, opened 2024-02-26
community[patch]: allow automatic metadata and pks on vector store

nicoloboschi:
In `CassandraVectorStore` currently seems to support automatic primary key (omitting `primaryKey`) but it fails to create the PK column because the name is not valid on cassandra (`_id`). 

Additionally, at the moment you're required to pass specific metadata columns. There's the support for using a generic column `metadata map<string, string>` but the index creation fails due to some syntax error.

This PR fixes both.
- We use `id` instead of `_id`
- Fixed the create index statement


libs/langchain-community/src/vectorstores/cassandra.ts
```diff
@@ -49,7 +49,7 @@ export class CassandraStore extends VectorStore {
 
   private readonly table: CassandraTable;
 
-  private readonly idColumnAutoName = "_id";
+  private readonly idColumnAutoName = "id";
```
jacoblee93, line 52: This seems not backwards compatible?

nicoloboschi: There's no harm since it's impossible that someone has a column named _id because it's not allowed in Cassandra. The feature wasn't working before

nicoloboschi:
@jacoblee93 any chance can you merge this one ?

jacoblee93:
Thank you!

PR #4517, opened 2024-02-26
community[patch]: add support for customizable bundle url

nicoloboschi:
Similar option available in LangChain python (via CassIO): https://github.com/CassioML/cassio/pull/128


jacoblee93:
Thank you!

PR #4522, opened 2024-02-26
mistral[minor]: Add support for tool calling, add withStructuredOutput

bracesproul:
Depends on https://github.com/langchain-ai/langchainjs/pull/4525

- [ ] add `withStructuredOutput` method


libs/langchain-mistralai/src/chat_models.ts
```diff
@@ -26,13 +31,51 @@ import {
...
+  topP?: number;
+  randomSeed?: number;
+  safeMode?: boolean;
+  safePrompt?: boolean;
+  toolChoice?: MistralAIToolChoice;
+  responseFormat?: ResponseFormat;
+};
+
+interface MistralAICallOptions extends BaseLanguageModelCallOptions {
+  responseFormat?: {
```
jacoblee93, line 72: Is this new? How does it interact with `tool_choice`?

libs/langchain-mistralai/src/chat_models.ts
```diff
@@ -121,41 +170,80 @@ function convertMessagesToMistralMessages(
...
+  if (delta.role) {
+    role = delta.role;
+  } else if (toolCallsWithIndex) {
+    role = "tool";
+  }
   const content = delta.content ?? "";
+  let additional_kwargs;
+  if (toolCallsWithIndex) {
+    additional_kwargs = {
+      tool_calls: toolCallsWithIndex,
```
jacoblee93, line 221: Shouldn't this just be the raw tool calls? Why are we mutating here?

libs/langchain-mistralai/src/chat_models.ts
bracesproul: whoops I think I need to make this `response_format`. Same as in openai (json mode)

PR #4533, opened 2024-02-27
langchain[minor]: Delete & re-export json tools parser code

bracesproul:


jacoblee93 (approved):

docs/core_docs/docs/modules/model_io/output_parsers/types/openai_tools.ipynb
```diff
@@ -160,7 +160,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "import { JsonOutputToolsParser } from \"langchain/output_parsers\";\n",
+    "import { JsonOutputToolsParser } from \"@langchain/core/output_parsers/openai_tools\";\n",
```
jacoblee93, line 163: Can we just import from `@langchain/core/output_parsers` instead for simplicity?

bracesproul: ah yeah this would work too, not sure if we need another release b/c I don't think that entrypoint exists yet. Will check

bracesproul:
Too many merge conflicts, tracking issue here https://github.com/langchain-ai/langchainjs/issues/5348

PR #4539, opened 2024-02-27
community[patch]: Neo4j will no longer return undefined

adam-cowley:
The `Neo4jGraph.query()` method would return `undefined` if an error was thrown at query time.  This is no longer the case, so the method will either throw or return an empty array if no records are returned.


jacoblee93:
Thank you!

PR #4543, opened 2024-02-27
community[patch]: Adds support for passing schemaName to pgvector

ZuesYousif:
Adds schemaName option to PGVectorStoreArgs
Adds extensionSchemaName option to PGVectorStoreArgs


libs/langchain-community/src/vectorstores/pgvector.ts
```diff
@@ -521,7 +561,23 @@ export class PGVectorStore extends VectorStore {
...
+
+        ALTER TABLE ${this.schemaName}.${this.tableName}
+          ADD COLUMN collection_id uuid;
+
+        ALTER TABLE ${this.schemaName}.${this.tableName}
+          ADD CONSTRAINT ${this.schemaName}.${this.tableName}_collection_id_fkey
+          FOREIGN KEY (collection_id)
+          REFERENCES ${this.collectionTableName}(uuid)
+          ON DELETE CASCADE;
+      `
```
jacoblee93, line 579: Don't forget to run `yarn format`!

libs/langchain-community/src/vectorstores/pgvector.ts
```diff
@@ -521,7 +561,23 @@ export class PGVectorStore extends VectorStore {
           FOREIGN KEY (collection_id)
           REFERENCES ${this.collectionTableName}(uuid)
           ON DELETE CASCADE;
-      `);
+      ` : `
+        CREATE TABLE IF NOT EXISTS ${this.schemaName}.${this.collectionTableName} (
```
jacoblee93, line 565: Should these be in quotes to account for capitalization?

jacoblee93:
Looks good to me overall - main question is should these new/old queries be in quotes though to account for capitalizations?

ZuesYousif:
> Looks good to me overall - main question is should these new/old queries be in quotes though to account for capitalizations?

Good catch there, I've added some string wrapping around a few calls where it seems appropriate

MJDeligan:
Love the push towards flexibility in the Postgres interfaces! I've left a comment. Maybe we could go with a similar approach as [here](https://github.com/langchain-ai/langchainjs/pull/4480), where the schema + table name is computed during construction and then inserted into the queries. That way we don't need to choose between two almost identical queries each time.

(Rest can be ignored, just my reasoning for the suggestion) 
Since the queries can still be subject to change (such as filters in similarity search), having two queries that do the same thing can be dangerous, especially since the tests only cover one of them. When extending the query you may forget to change the second one or make a typo in it, which is only discovered at run time.

ZuesYousif:
> Love the push towards flexibility in the Postgres interfaces! I've left a comment. Maybe we could go with a similar approach as [here](https://github.com/langchain-ai/langchainjs/pull/4480), where the schema + table name is computed during construction and then inserted into the queries. That way we don't need to choose between two almost identical queries each time.
> 
> (Rest can be ignored, just my reasoning for the suggestion) Since the queries can still be subject to change (such as filters in similarity search), having two queries that do the same thing can be dangerous, especially since the tests only cover one of them. When extending the query you may forget to change the second one or make a typo in it, which is only discovered at run time.

That's a good consideration I had overlooked. Just pushed a change that follows the suggested approach.

libs/langchain-community/src/vectorstores/pgvector.ts
```diff
@@ -506,22 +526,24 @@ export class PGVectorStore extends VectorStore {
...
+        ALTER TABLE ${tableName}
           ADD COLUMN collection_id uuid;
 
-        ALTER TABLE ${this.tableName}
-          ADD CONSTRAINT ${this.tableName}_collection_id_fkey
+        ALTER TABLE ${tableName}
+          ADD CONSTRAINT "${tableName.replaceAll('"', '')}_collection_id_fkey"
           FOREIGN KEY (collection_id)
-          REFERENCES ${this.collectionTableName}(uuid)
+          REFERENCES "${this.collectionTableName}(uuid)"
```
MJDeligan, line 543: This might be a syntax error?

ZuesYousif: Good catch, those quotes should be removed now

ZuesYousif:
> Looks good! Thanks for the quick fixes! My only concern is about existing users, since we're adding quotes in some places where there were none before. I remember there was some inconsistency before with this in the implementation. It seems like the table name was originally unquoted. It should probably stay that way to not break existing usage. In the table identifier with schema name it should be fine to quote since existing users are not affected by this, right?

Hows that look? I removed the wrapper in every instance where schemaName isn't present to avoid that case.

MJDeligan (commented):
Great work! Looked through it again and marked places where quotes were added + 1 part where I'm not sure. As a final touch, if you don't mind, could we have the computed table name as a property on the class and a quick `yarn format`? Loving the quick response time, thank you so much!

libs/langchain-community/src/vectorstores/pgvector.ts
```diff
@@ -506,22 +521,23 @@ export class PGVectorStore extends VectorStore {
...
         );
 
-        ALTER TABLE ${this.tableName}
+        ALTER TABLE ${this.computedTableName}
           ADD COLUMN collection_id uuid;
 
-        ALTER TABLE ${this.tableName}
-          ADD CONSTRAINT ${this.tableName}_collection_id_fkey
+        ALTER TABLE ${this.computedTableName}
+          ADD CONSTRAINT ${this.computedTableName}_collection_id_fkey
```
MJDeligan, line 535: Shoud be this.tableName in the constraint identifier. Think this would throw cause of quotes?

ZuesYousif: Yes that would throw due to the quotes. Just pushed the change using this.tableName instead

ZuesYousif:
> Could get to it tomorrow, I'm on an ancient laptop right now, which would take forever

No worries. I was able to get it working & just pushed up the formatted code

MJDeligan:
Great, good work! Looks good to me, @jacoblee93 ! As far as I can tell, it's backwards compatible.

ZuesYousif:
I found a bug where the operator `<=>` from pgvector was out of scope when pgvector is nested under a schema. To fix this I pushed a small change that sets the schema in a query prior to using the operator. I tried passing it under the same pool.query call however it said only one query is allowed & after testing the changes under a it's own call it seems to be working now.

MJDeligan:
Does setting the search_path only persist for that specific connection or does it change at the db level?

ZuesYousif:
> Does setting the search_path only persist for that specific connection or does it change at the db level?

It looks like it only lasts for the duration of that connection. When I manually query the db with `SELECT CURRENT_SCHEMA, CURRENT_SCHEMA();` after running this the db still returns `public`

MJDeligan:
Gotcha! Is that for the pool in general or is this tied to the client the pool creates to run the query? In the first case, we may run into issues if we ever allow passing a pool instance, instead of creating a new one (as proposed in #4409 and present in PostgresChatHistory) among other unforeseen consequences for users. In the latter case, we could get unlucky and the pool uses a different client for each query. Have you seen [this](https://dba.stackexchange.com/questions/191695/is-it-possible-to-schema-qualify-an-operator) ? Might be applicable here?.

ZuesYousif:
> Gotcha! Is that for the pool in general or is this tied to the client the pool creates to run the query? In the first case, we may run into issues if we ever allow passing a pool instance, instead of creating a new one (as proposed in #4409 and present in PostgresChatHistory) among other unforeseen consequences for users. In the latter case, we could get unlucky and the pool uses a different client for each query. Have you seen [this](https://dba.stackexchange.com/questions/191695/is-it-possible-to-schema-qualify-an-operator) ? Might be applicable here?.

I haven't seen that yet, thanks that helps a lot in this case. I pushed a refactor that opts for the `OPERATOR` function instead which seems to fix the initial scope bug.

libs/langchain-azure-openai/.eslintrc.cjs
jacoblee93, line 1: Could you revert this deletion?

libs/langchain-community/src/vectorstores/pgvector.ts
```diff
@@ -51,12 +53,18 @@ export class PGVectorStore extends VectorStore {
 
   collectionMetadata: Metadata | null;
 
+  schemaName: string | null;
+
   idColumnName: string;
 
   vectorColumnName: string;
 
+  computedTableName: string;
```
jacoblee93, line 62: nit: could be a getter method? Would fit `computed` better.

jacoblee93:
Thank you both!

PR #4546, opened 2024-02-27
core[minor]: Add JSON functions output parser to core

bracesproul:


langchain-core/src/output_parsers/openai_tools/json_output_functions_parsers.ts
```diff
@@ -0,0 +1,209 @@
...
+    }
+    if (this.argsOnly) {
+      return parsePartialJson(functionCall.arguments);
+    }
+
+    return {
+      ...functionCall,
+      arguments: parsePartialJson(functionCall.arguments),
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } as Record<string, any> as Output;
```
bracesproul, line 136: cc @jacoblee93 I had to add this cast because `functionCall` does not overlap enough with `Record<string, any>` for some reason...

PR #4560, opened 2024-02-27
community[patch]: Allow a raw filter option for Redis Vector Store

### LINKED ISSUE #4559: Redis Vector Store Should Support Custom Filter Queries
Currently the filter for Redis Vector Store is a simple "does the meta data contain any of these strings" check.  While a more robust system would be nice, even a simple "raw" option would work such that clients can use and combination of AND, OR, etc. for filtering.

EricThompson-PeopleReign:
- Update existing filter test to make sure the `join` is exercised.
- Add raw filter (pass string instead of array) option

Fixes #4559


jacoblee93:
Looks reasonable, thank you!

jacoblee93:
Seems unit test is failing?

EricThompson-PeopleReign:
> Seems unit test is failing?

Oops!  Typo in the adjustment of the `prepareFilter` code.  I have corrected it and pushed.

EricThompson-PeopleReign:
@jacoblee93 could you have another look?

Issue #4375, opened 2024-02-12
Safety-blocked Gemini Pro requests causes failure


jarib:
Not sure what's the correct fix for this, but [here's a](https://github.com/langchain-ai/langchainjs/pull/4374) is a draft PR with a  failing test.

```
  ● Mock ChatGoogle › 2. Blocked response

    TypeError: Cannot read properties of undefined (reading 'message')

      51 |         const chatGeneration = result.generations[0][0];
      52 |         // TODO: Remove cast after figuring out inheritance
    > 53 |         return chatGeneration.message;
         |                               ^
      54 |     }
      55 |     // eslint-disable-next-line require-yield
      56 |     async *_streamResponseChunks(_messages, _options, _runManager) {

      at ChatGoogle.invoke (../../langchain-core/dist/language_models/chat_models.js:53:31)
      at RunnableSequence.invoke (../../langchain-core/dist/runnables/base.js:1014:33)
      at Object.<anonymous> (src/tests/chat_models.test.ts:245:20)

 PASS  src/tests/llms.test.ts
 ```

jacoblee93:
@afirstenberg

afirstenberg:
@jacoblee93  - Please assign this to me. {:
(The other one was for the tests. This will remind me I need to make a fix.)

jacoblee93:
Thank you so much!

PR #4583, opened 2024-02-29
google-common/webauth/gauth [patch]: Fixes for error, safety, and image handling

afirstenberg:
Addresses several oddly related issues:
- New error handling tailored to Google's error codes and messages
- Prompt block errors and exceptional finish reasons now raise an Error
  - This is tailorable by writing a handler that treats this differently, including rewriting the response
- Provide custom .stream() and .invoke() methods for the GoogleLLM classes so they can accept image parts.
  - This is done by creating an calling a chat model.

This included extensive additional tests.

(There are also some tests that are currently skipped because Google is raising safety errors on them. Other tests have been added to replace these.)

Fixes #4374 #4375


afirstenberg:
@jacoblee93  - If you can look at `GoogleBaseLLM.stream()` to make sure the solution to create an `IterableReadableStream<string>` makes sense, I'd appreciate it.

libs/langchain-google-common/src/chat_models.ts
```diff
@@ -81,14 +85,18 @@ export abstract class ChatGoogleBase<AuthOptions>
 
   safetySettings: GoogleAISafetySetting[] = [];
 
+  safetyHandler;
```
jacoblee93, line 88: I think this should be typed?

libs/langchain-google-common/src/llms.ts
```diff
@@ -207,7 +244,44 @@ export abstract class GoogleBaseLLM<AuthOptions>
       {},
       options as BaseLanguageModelCallOptions
     );
-    const ret = responseToBaseMessage(result);
+    const ret = safeResponseToBaseMessage(result, this.safetyHandler);
     return ret;
   }
+
+  createProxyChat(): ChatGoogleBase<AuthOptions> {
+    console.log("this", this);
```
jacoblee93, line 252: Remove log

libs/langchain-google-common/src/llms.ts
```diff
@@ -207,7 +244,44 @@ export abstract class GoogleBaseLLM<AuthOptions>
...
+    // Based on code from Gemini Advanced
+    const transformStream = new TransformStream({
+      transform(chunk: BaseMessageChunk, controller) {
+        const text = chunkToString(chunk);
+        controller.enqueue(text);
+      },
+    });
+
+    const readableStream = chatStream.pipeThrough(transformStream);
+    return IterableReadableStream.fromReadableStream(readableStream);
```
jacoblee93, line 285: This looks fine if it passes tests!

jacoblee93, line 285: I think this will break the callback system - there's a few behind the scenes things going on. We'll want to override `_streamResponseChunk` instead.

jacoblee93:
Mostly looks great! Will dig in and override the proper method hopefully tomorrow.

libs/langchain-google-common/src/chat_models.ts
afirstenberg: It didn't _need_ to be, since it inherits from an interface that defines it. But agreed - clarity is good. I'll add it.

libs/langchain-google-common/src/llms.ts
afirstenberg: Doh! Got it.

libs/langchain-google-common/src/llms.ts
afirstenberg: The problem with overriding `_streamResponseChunk()` is that it **requires** a string `_prompt` while `stream()` takes a `BaseLanguageModelInput` which will allow sending a multimodal prompt to the LLM, which is part of why we're doing this.

I suppose we could change `_streamResponseChunk()` _in addition to_ adding this `stream()` method?

What are we breaking in callbacks by overriding `stream()`?

afirstenberg: The underlying problem is that `BaseLLM._streamIterator()` explicitly calls the static method `BaseLLM._convertInputToPromptValue()` which forces the prompt to be a string.

I guess we could override `_streamIterator()`, but that is a huge method to duplicate the code for. Unless it is safe to have our method call the chat model version of `_streamIterator()`?

afirstenberg: I've gone with overriding `_streamIterator()` see how that does?
(It passes all the tests I have for google-common, but I'm not sure what other tests there are that might be testing out stream().)

afirstenberg:
@jacoblee93 - The integration tests for google-gauth and google-webauth pass
Everything looks good to me.

Please update the version on all three packages when merged.

Thanks for your help!

jacoblee93:
Thank you!

PR #4589, opened 2024-02-29
community[patch]: Add retrieval params to neo4j vector

tomasonjo:
Sometimes, you want to use various parameters in the retrieval query of Neo4j Vector to personalize/customize results. Before, when there were only predefined chains, it didn't really make sense. Now that it's all about custom chains and LCEL, it is worth adding since users can inject any params they wish at query time. Isn't prone to SQL injection-type attacks since we use parameters and not concatenating strings.

### DIFF
--- a/libs/langchain-community/src/vectorstores/neo4j_vector.ts
+++ b/libs/langchain-community/src/vectorstores/neo4j_vector.ts
@@ -577,13 +577,18 @@ export class Neo4jVectorStore extends VectorStore {
     );
   }
 
-  async similaritySearch(query: string, k = 4): Promise<Document[]> {
+  async similaritySearch(
+    query: string,
+    k = 4,
+    params: Record<string, any> = {}
+  ): Promise<Document[]> {
     const embedding = await this.embeddings.embedQuery(query);
 
     const results = await this.similaritySearchVectorWithScore(
       embedding,
       k,
-      query
+      query,
+      params
     );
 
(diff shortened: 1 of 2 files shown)


libs/langchain-community/src/vectorstores/neo4j_vector.ts
```diff
@@ -577,13 +577,18 @@ export class Neo4jVectorStore extends VectorStore {
     );
   }
 
-  async similaritySearch(query: string, k = 4): Promise<Document[]> {
+  async similaritySearch(
+    query: string,
+    k = 4,
+    params: Record<string, any> = {}
```
jacoblee93, line 583: The convention has been to treat this final parameter as a search filter - would be nicer if possible to support adding `invocationParams` when initializing the client and spreading them when running the query.

jacoblee93: What are some params you'd want to pass here?

tomasonjo: you could personalize responses based on user ids, user role, locations, and probably more. This only works when you provide a corresponding retrieval_query with relevant parameters.

jacoblee93: Let's roll with it then

jacoblee93:
Thank you!

PR #4591, opened 2024-02-29
community[patch]: Add maxTokens to _streamResponseChunks options in llama_cpp.ts

adamkdean:
This PR introduces the missing maxTokens option to _streamResponseChunks.


bracesproul (approved):
thanks for this, good catch!

Issue #4540, opened 2024-02-27
Using a collection ID in PGVectorStore without a where clause seems to throw an issue


fiachra:
https://github.com/langchain-ai/langchainjs/blob/c2bc44010380d39d96e55965854cd02c6fcec485/libs/langchain-community/src/vectorstores/pgvector.ts#L461

I ran into a problem where I tried to use a collection and I got an error:
```
Uncaught DatabaseError error: syntax error at or near "AND"
```

It seems if the whereClauses array is empty then you get an `AND` clause without an initial `WHERE`

```
 const whereClause = whereClauses.length
      ? `WHERE ${whereClauses.join(" AND ")}`
      : "";

    const queryString = `
      SELECT *, ${this.vectorColumnName} <=> $1 as "_distance"
      FROM ${this.tableName}
      ${whereClause}
      ${collectionId ? "AND collection_id = $3" : ""}
      ORDER BY "_distance" ASC
      LIMIT $2;
      `;
  ```
  
Quick and dirty solution might be to  just set `WHERE` to true
  ```
 const whereClause = whereClauses.length
      ? `WHERE ${whereClauses.join(" AND ")}`
      : "WHERE true";
  ```

IlyaSavich:
I think it should be fixed there https://github.com/langchain-ai/langchainjs/pull/4597

jacoblee93:
Yes, thank you!

PR #4597, opened 2024-03-01
community[patch]: Fix using collections for pgvector

IlyaSavich:
Hi, I found there is an issue with pgvector using collections. This should fix it

Fixes # (issue) https://github.com/langchain-ai/langchainjs/issues/4540


jacoblee93:
Thank you, nice catch!

PR #4601, opened 2024-03-01
Corrected package installer for npm docs

FriedGil:


jacoblee93:
Whoops, thanks for the catch!

PR #4604, opened 2024-03-03
core[minor]: Added optional `tool_call_id` to BaseMessageFields interface

DrewsAbuse:
…

Fix TS issues

Screenshot 2024-03-03 at 11 16 55 PM


bracesproul (commented):
For this, you should use a `ToolMessage` like so:

```typescript
new ToolMessage({
  tool_call_id: toolCall.id,
  name: toolCall.function.name,
  content: toolCallResult,
});
```

PR #4607, opened 2024-03-04
docs[minor]: Simplify docs build

jacoblee93:
CC @bracesproul


package.json
```diff
@@ -34,7 +34,7 @@
     "test:exports:docker": "docker compose -f environment_tests/docker-compose.yml up --force-recreate",
     "example": "yarn workspace examples start",
     "precommit": "turbo run precommit",
-    "docs": "yarn workspace core_docs start",
+    "docs": "cd docs/core_docs && yarn workspace core_docs start",
```
bracesproul, line 37: don't believe you need this `cd` because you already have `yarn workspace`

bracesproul (approved):
lgtm w/ one comment

package.json
jacoblee93: Oh, yeah prob not

PR #4618, opened 2024-03-04
anthropic[minor]: Adds experimental ChatAnthropicTools model to integration package

jacoblee93:
@efriis @hwchase17


libs/langchain-anthropic/src/utils/function_calling.ts
efriis, line 1: would it be better to save these prompts and whatnot under experimental so it doesn't look like we're supporting after?

jacoblee93: I was going to axe the whole file, will move everything under experimental

PR #4619, opened 2024-03-04
core[minor]: Clean up withStructuredOutput typing

jacoblee93:
@bracesproul

### DIFF
--- a/langchain-core/src/language_models/base.ts
+++ b/langchain-core/src/language_models/base.ts
@@ -242,10 +242,10 @@ export type BaseLanguageModelInput =
   | BaseMessageLike[];
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-export type AnyObjectType = z.ZodObject<any, any, any, any>;
+export type StructuredOutputType = z.infer<z.ZodObject<any, any, any, any>>;
 
 export type StructuredOutputMethodParams<
-  RunOutput extends z.infer<AnyObjectType> = z.infer<AnyObjectType>,
+  RunOutput extends StructuredOutputType = StructuredOutputType,
   IncludeRaw extends boolean = false
 > = {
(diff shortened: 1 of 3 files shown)


langchain-core/src/language_models/base.ts
```diff
@@ -474,46 +488,34 @@ export abstract class BaseLanguageModel<
...
+    BaseLanguageModelInput,
+    { raw: BaseMessage; parsed: RunOutput }
+  >;
 
   /**
    * Model wrapper that returns outputs formatted to match the given schema.
    *
    * @template {BaseLanguageModelInput} RunInput The input type for the Runnable, expected to be the same input for the LLM.
-   * @template {z.ZodObject<any, any, any, any>} RunOutput The output type for the Runnable, expected to be a Zod schema object for structured output validation.
+   * @template {AnyObjectType} RunOutput The output type for the Runnable, expected to be a Zod schema object for structured output validation.
```
bracesproul, line 518: Can this be an object too now? Not just a zod object?

Also `@template` above should be dropped and one for `IncludeRaw` should be added

langchain-core/src/language_models/tests/chat_models.test.ts
```diff
@@ -36,3 +37,68 @@ test("Test ChatModel uses callbacks with a cache", async () => {
...
+    ],
+  });
+  expect(response.content).toEqual(response2.content);
+  expect(response2.content).toEqual(acc);
+});
+
+test("Test ChatModel withStructuredOutput", async () => {
+  const model = new FakeListChatModel({
+    responses: [`{ "test": true, "nested": { "somethingelse": "somevalue" } }`],
+  }).withStructuredOutput({
```
bracesproul, line 64: We should add generics in tests so if they ever break it'll fail build and we'll know

bracesproul (approved):
thanks for this jacob

PR #4633, opened 2024-03-05
core[patch]: Retry TimeoutErrors in OAI client code

### LINKED ISSUE #2706: Retry logic for OpenAI timeouts

wcummings:
Fixes #2706

### DIFF
--- a/langchain-core/src/utils/async_caller.ts
+++ b/langchain-core/src/utils/async_caller.ts
@@ -10,7 +10,6 @@ const STATUS_NO_RETRY = [
   405, // Method Not Allowed
   406, // Not Acceptable
   407, // Proxy Authentication Required
-  408, // Request Timeout
   409, // Conflict
 ];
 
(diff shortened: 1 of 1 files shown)


jacoblee93:
By popular demand, let's do it. Thank you!

PR #4634, opened 2024-03-05
docs[minor]: fix links on model io page

cdgugler:
Fixes broken links on Model I/O page.

### DIFF
--- a/docs/core_docs/docs/modules/model_io/index.mdx
+++ b/docs/core_docs/docs/modules/model_io/index.mdx
@@ -10,26 +10,26 @@ The core element of any language model application is... the model. LangChain gi
 
 ![model_io_diagram](/img/model_io.jpg)
 
-## [Conceptual Guide](./concepts)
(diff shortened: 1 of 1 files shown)


bracesproul (approved):
oh jeez thanks a ton for catching this. I guess docusourus' broken link checker won't throw if the link is prefixed with a `.`...

Question, did you do a search across all the docs for this or just this page? If not we need to do that and fix any other possible instances of this!!

Thanks again for pushing this up, will get it merged asap

cdgugler:
> oh jeez thanks a ton for catching this. I guess docusourus' broken link checker won't throw if the link is prefixed with a `.`...
> 
> Question, did you do a search across all the docs for this or just this page? If not we need to do that and fix any other possible instances of this!!
> 
> Thanks again for pushing this up, will get it merged asap

@bracesproul np! I didn't think to search across the docs, but that's a good idea. I just happened to be reading this page and noticed the broken links.

PR #4643, opened 2024-03-06
langchain[patch]: Allow return direct for runnable agents created via helper methods

niklas-lohmann:
Currently an error is thrown when using the `returnDirect` property of custom tools for multi-action agents. As described [here](https://github.com/langchain-ai/langchainjs/discussions/4631) this is probably not intended. This pull request changes the behavior. I tested my changes with OpenAI function calling agent - works fine.


jacoblee93:
Thanks for this!

Will aim to review tomorrow, sorry for the delay.

jacoblee93:
Looked into this a bit more - this gives me some pause:

https://github.com/langchain-ai/langchainjs/blob/bb001e888508653dedf3ab00479e5677ae058acb/langchain/src/agents/executor.ts#L562

If the agent returns multiple tools, which one do we return?

Maybe a sensible compromise is if there's only one tool called and `returnDirect: true` on that tool, return the output. What do you think?

jthack:
Also, could we get this for the python code as well @jacoblee93

jacoblee93:
Ok, so dug in a bit more and it seems like Python supports `RunnableAgent` and `RunnableMultiActionAgent`. We should do the same and avoid a somewhat hacky compromise.

You would then be able to use `returnDirect` with e.g. `createOpenAIFunctionsAgent` but not `createOpenAIToolsAgent`, which is the right behavior. I'll get a fix up as soon as I can.

niklas-lohmann:
Sounds great. Thanks in advance!

langchain/src/agents/agent.ts
```diff
@@ -159,14 +160,99 @@ function isAgentAction(input: unknown): input is AgentAction {
...
+      runName: this.defaultRunName,
+    });
+    if (this.streamRunnable) {
+      const stream = await this.runnable.stream(combinedInput, combinedConfig);
+      let finalOutput: AgentAction | AgentFinish | undefined;
+      for await (const chunk of stream) {
+        if (finalOutput === undefined) {
+          finalOutput = chunk;
+        } else {
+          throw new Error(
```
bracesproul, line 227: This is saying that if the runnable streams more than one chunk we'll throw? I'm assuming LLMs will stream back multiple chunks and not one.

langchain/src/agents/agent.ts
```diff
@@ -193,14 +285,39 @@ export class RunnableAgent extends BaseMultiActionAgent {
...
+    if (this.streamRunnable) {
+      const stream = await this.runnable.stream(combinedInput, combinedConfig);
+      let finalOutput: AgentAction | AgentFinish | AgentAction[] | undefined;
+      for await (const chunk of stream) {
+        if (finalOutput === undefined) {
+          finalOutput = chunk;
+        } else {
+          throw new Error(
+            [
+              `Multiple agent actions/finishes received in streamed agent output.`,
```
bracesproul, line 303: Same here

langchain/src/agents/executor.ts
```diff
@@ -393,10 +394,13 @@ export class AgentExecutor extends BaseChain<ChainValues, AgentExecutorOutput> {
     let agent: BaseSingleActionAgent | BaseMultiActionAgent;
     let returnOnlyOutputs = true;
     if (Runnable.isRunnable(input.agent)) {
-      agent = new RunnableAgent({ runnable: input.agent });
+      agent = new RunnableMultiActionAgent({ runnable: input.agent });
       // TODO: Update BaseChain implementation on breaking change
       returnOnlyOutputs = false;
     } else {
+      if (isRunnableAgent(input.agent)) {
```
bracesproul, line 401: More of a question for @jacoblee93 but why don't we just add `lc_runnable` to the runnable agent classes? That way we can use the `Runnable.isRunnable()` method

langchain/src/agents/agent.ts
jacoblee93: All the example `RunnableSequence`s powering the agents will gather the output of the LLM and parse it, returning it in one chunk. Streaming is done using intermediate steps.

langchain/src/agents/executor.ts
jacoblee93: Because they're not runnable

jacoblee93:
Thank you!

pieracc:
any news for Python? i'm having the same problem here

WebsheetPlugin:
I am not a fan of making this work only for one createOpenAIFunctionsAgent . 

More logical seems: If multiple tools, let the last selected tool decide if it is final.

or:

if all toos are final, then it's final in the end.

-----

Also, there is no way inside a tool to make it return as final?

-----

Because curently, I just want the agent to use tools. Also, it seems no agent will just use tools without answering, or am I wrong?

jacoblee93:
If you just want to use the LLM for routing, maybe this is a better approach?

https://js.langchain.com/docs/use_cases/tool_use/multiple_tools

WebsheetPlugin:
I am using it now, but not because I want to. I was not able to cancel the run. Overall there should be some way for the tool or tools to cancel the Agent and return message X. I think there are many use cases, and it's saving tokens.

There should be some meaningful logic to enable this in the case of multi-tools or not.

owais-globalnorthstar:
@jacoblee93 Can we do this implementation in langchain python too ?

PR #4649, opened 2024-03-06
community[patch]: PGVectorStore: Fixes collection table created on wrong schema

### LINKED ISSUE #4648: PGVectorStore: Schema not used for collection table
### Checked other resources

### Example Code

```typescript
import { PGVectorStore } from "@langchain/community/vectorstores/pgvector";

const config = {
      postgresConnectionOptions: {
        type: "postgres",
        host: "127.0.0.1",
        port: 5433,
        user: "myuser",
        password: "ChangeMe",
        database: "api",
      } as PoolConfig,
      tableName,
      collectionTableName: "langchain_pg_collection",
      collectionName: "langchain",
      columns: {
        idColumnName: "id",
        vectorColumnName: "vector",
        contentColumnName: "content",
        metadataColumnName: "metadata",
      },
      schemaName: schema,
    };

const pgvectorStore = await PGVectorStore.initialize(
  new OpenAIEmbeddings(),
  config
);
```

_No response_

### Description

When initializing the PGVectorStore, passing a collection table name and schema name, the schema is ignored for the collection table. This results in the Embeddings table being created on the passed schema while the collection table is created on the default schema. I believe this would not be expected by users.

### System Info

Behavior is independent from system versions.

MJDeligan:
Fixes #4648

With the introduction of the schema param in #4480 , users are now able to specify a schema for the tables. This schema is not used for the collection table, which I believe is unexpected behavior. This PR fixes the issue by using the same pattern as for the embeddings table, creating a computed collection table name from the schema and collection table name. This is not a breaking change as far as I can tell. 

This PR also adds a `pool` param to the config and makes the `postgresConnectionOptions` param optional. The vector store will only create a new pool if no pool is passed during initialization, giving users more freedom in managing their connections. 

I've added two test suites, which test the implementation when using a collection as well as when using a schema and collection. The test suites are skipped, but have been run to ensure they pass.


libs/langchain-community/src/vectorstores/pgvector.ts
```diff
@@ -113,6 +119,12 @@ export class PGVectorStore extends VectorStore {
       : `"${this.schemaName}"."${this.tableName}"`;
   }
 
+  get computedCollectionTableName() {
```
MJDeligan, line 122: This will return a string regardless of whether a collection table name was defined or not. This should not be an issue, because there is no reason to get this property unless one was passed, but we could change to return undefined.

jacoblee93:
Hey, thanks so much for this! Will review tomorrow.

jacoblee93:
Thank you!

jacoblee93:
CI failure is a flake. Thank you!

PR #4651, opened 2024-03-06
langchain[patch]: Add Possibility to use Contextual chunk headers in Parent Document Retriever

karol-f:
Add possibility to use Contextual chunk headers in Parent Document Retriever.

This is particularly important if you have several fine-grained child chunks that need to be correctly retrieved from the vector store.

### DIFF
--- a/langchain/src/retrievers/parent_document.ts
+++ b/langchain/src/retrievers/parent_document.ts
@@ -126,15 +126,18 @@ export class ParentDocumentRetriever extends MultiVectorRetriever {
    * This can be false if and only if `ids` are provided. You may want
    *   to set this to False if the documents are already in the docstore
    *   and you don't want to re-add them.
+   * @param config.chunkHeader Boolean of whether to add chunk header to docs to ease
+   * searching in vector store
    */
   async addDocuments(
     docs: Document[],
     config?: {
       ids?: string[];
       addToDocstore?: boolean;
+      chunkHeader?: boolean;
     }
   ): Promise<void> {
-    const { ids, addToDocstore = true } = config ?? {};
+    const { ids, addToDocstore = true, chunkHeader = false } = config ?? {};
     const parentDocs = this.parentSplitter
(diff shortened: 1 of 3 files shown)


jacoblee93:
Nice one, thank you!

jacoblee93:
Looks like you might need to run `yarn lint`?

karol-f:
> Looks like you might need to run `yarn lint`?

You are right, sorry. Fixed.

langchain/src/retrievers/parent_document.ts
```diff
@@ -159,7 +162,11 @@ export class ParentDocumentRetriever extends MultiVectorRetriever {
     for (let i = 0; i < parentDocs.length; i += 1) {
       const parentDoc = parentDocs[i];
       const parentDocId = parentDocIds[i];
-      const subDocs = await this.childSplitter.splitDocuments([parentDoc]);
+      const chunkHeaderOptions = chunkHeader && parentDoc?.metadata?.chunkHeader ? {
+        chunkHeader: parentDoc.metadata.chunkHeader,
```
jacoblee93, line 166: Depending on this to be in the `metadata` field of the document is a bit weird - could we make this a property on the class instead? Like `this.childSplitterOptions`?

jacoblee93:
See comment, also please run `yarn format` from root!

langchain/src/retrievers/parent_document.ts
jacoblee93: Ignore the above, I see the problem. Let me think about it a little longer - would be nice to not have a dependency on an untyped metadata key like this.

karol-f: I think we just have to the same way as when passing chunkHeader to parent chunk. It will be better solution. Thanks for spotting

karol-f:
Thank you again for the suggestions. I've pushed improved version that looks cleaner and don't use `metadata`

jacoblee93:
Thank you!

PR #4653, opened 2024-03-06
core[patch]: Update withStructuredOutput overload signature

jacoblee93:
CC @bracesproul @nfcampos


jacoblee93:
This new signature will unfortunately ruin typing for legacy `includeRaw: true`, but opens the door for some nicer experiences and I think better to bite the bullet now before more people start using it.

PR #4655, opened 2024-03-07
openai[patch],mistral[patch],anthropic[patch]: Update client library signatures for withStructuredOutput

jacoblee93:


jacoblee93:
CC @bracesproul

PR #4664, opened 2024-03-07
community[patch]: Allow must_not filter in Elasticsearch VectorStore

LaurensTsestigh:
For our project we needed to exclude some ids of certain entities when searching the VectorStore.
As the current implementation doesn't support the must_not filter yet, I've added it as an "exclude" operator to be flexible and populate the must_not filter accordingly.

### DIFF
--- a/libs/langchain-community/src/vectorstores/elasticsearch.ts
+++ b/libs/langchain-community/src/vectorstores/elasticsearch.ts
@@ -165,7 +165,7 @@ export class ElasticVectorSearch extends VectorStore {
       knn: {
         field: "embedding",
         query_vector: query,
-        filter: this.buildMetadataTerms(filter),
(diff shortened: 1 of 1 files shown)


libs/langchain-community/src/vectorstores/elasticsearch.ts
```diff
@@ -313,32 +313,44 @@ export class ElasticVectorSearch extends VectorStore {
   private buildMetadataTerms(
     filter?: ElasticFilter
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
jacoblee93, line 315: Move this lint ignore down and apply to the instances of `any`

jacoblee93:
Hey @LaurensTsestigh, I can't fix lint - can you update?

jacoblee93:
Thank you both!

PR #4668, opened 2024-03-08
azure-openai[patch]: feat: add azureOpenAIApiDeploymentName for AzureChatOpenAI constructor

SpringMT:
Fixes # (issue)

AzureOpenAIEmbeddings and AzureOpenAI can use `azureOpenAIApiDeploymentName` in constructer however AzureChatOpenAI not.

So, add `azureOpenAIApiDeploymentName` to AzureChatOpenAI.


libs/langchain-azure-openai/src/chat_models.ts
```diff
@@ -256,7 +256,8 @@ export class AzureChatOpenAI
       getEnvironmentVariable("AZURE_OPENAI_API_ENDPOINT");
 
     this.azureOpenAIApiCompletionsDeploymentName =
-      fields?.azureOpenAIEmbeddingsApiDeploymentName ??
+      (fields?.azureOpenAIEmbeddingsApiDeploymentName ||
```
jacoblee93, line 259: Should this even take into account `azureOpenAIEmbeddingsApiDeploymentName`? I guess we have to leave it there for backwards compatibility...

jacoblee93 (approved):
Thanks for this! Nice catch.

CC @sinedied @sarangan12

libs/langchain-azure-openai/src/chat_models.ts
SpringMT: While backward compatibility is important, I'm leaning towards removing azureOpenAIEmbeddingsApiDeploymentName since it seems to confuse users.
What do you think?

SpringMT: I created a PR for deprecating `azureOpenAIEmbeddingsApiDeploymentName` .
https://github.com/langchain-ai/langchainjs/pull/4714

jacoblee93:
Thank you! Sorry for the delay.

Issue #4684, opened 2024-03-09
Use Zod parse for validation


ashburnham:
### Privileged issue

- [X] I am a LangChain maintainer, or was asked directly by a LangChain maintainer to create an issue here.

### Issue Content

Great to see with_structured_output here 😎

One very important thing I think that has gone a bi backwards is the use of Zod in langchainjs

In the old code it uses Zod to parse the output, rather than using the zod schema to generate json. This is VERY important as you can use z.preprocess to handle simple (very common) LLM output errors BEFORE parsing to avoid loads of avoidable errors. e.g. when it's supposed to return an array, but it returns a long single string that that is a comm separated list of the elements, or when numbers come back as strings like "123".

https://github.com/colinhacks/zod?tab=readme-ov-file#preprocess

You can see how it used to be handled here

https://github.com/langchain-ai/langchainjs/blob/main/langchain/src/chains/openai_functions/structured_output.ts

In the new code, as far as I can tell, it just uses the zod schema to generate a json schema so so loses this rich functionality.

https://github.com/langchain-ai/langchainjs/blob/main/langchain/src/output_parsers/structured.ts

We can't (and I wouldn't recommend anyone else to) use structured outputs in production without this as the types of errors I describe are so common and so avoidable. They are especially common in Claude which doesn't do function calling so natively, and I know a lot of people are going to be using Claud with the recent releases.

I've submitted PRs for to make this work in the original version, but don't have time this month to do so. When 0.2.0 is released and the old code is dropped we won't be able to upgrade unless this functionality is maintained.

As with_structured_output is a priority, please can this feedback be taken on for 0.2.0?

Issue raised at the request of @jacoblee93

PR #4685, opened 2024-03-09
core[patch]: Allow tools output parser to take and use a Zod schema for validation

jacoblee93:
Step 1 for #4684


jacoblee93:
Needs a test or two

PR #4704, opened 2024-03-10
core[patch]: (fix): vector store callback

perezpefaur:
Description:
This PR addresses the optional callbacks parameter in the maxMarginalRelevanceSearch method within the VectorStore class. Previously, the parameter was not properly implemented, which could lead to issues when attempting to pass callback functions. This update ensures that callbacks are handled correctly, allowing users to pass optional success or error handling functions to the method.

Changes:

Corrected the parameter name from _callbacks to callbacks to match the method signature.
Implemented functionality to handle the callbacks parameter within the method, ensuring that provided callback functions are executed appropriately.

Additional Context:
These changes improve the method's flexibility and usability, allowing developers to integrate more robust error handling and success callbacks.

### DIFF
--- a/langchain-core/src/vectorstores.ts
+++ b/langchain-core/src/vectorstores.ts
@@ -206,7 +206,7 @@ export interface VectorStoreInterface extends Serializable {
   maxMarginalRelevanceSearch?(
     query: string,
     options: MaxMarginalRelevanceSearchOptions<this["FilterType"]>,
-    callbacks: Callbacks | undefined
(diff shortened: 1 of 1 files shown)


jacoblee93:
Thanks for the PR!

I don't think the variable name affects positional args?

And `Callbacks | undefined` should be equivalent to the optional `?`

What was the issue this fixes?

jacoblee93:
Feel free to reopen if I'm closing incorrectly

PR #4714, opened 2024-03-12
azure-openai[patch]: feat: use azureOpenAIApiDeploymentName for all azure-openai

SpringMT:
Fixes # (issue)
Ref #4668 

In `azure-openai`, `azureOpenAiembeddingsApiDeploymentName` is deprecated in favor of `azureOpenAIApiDeploymentName`.
This change introduces consistency but at the cost of losing backward compatibility for using `azureOpenAIEmbeddingsApiDeploymentName` in `AzureChatOpenAI`.


libs/langchain-azure-openai/src/embeddings.ts
```diff
@@ -48,11 +48,9 @@ export class AzureOpenAIEmbeddings
...
     super(fieldsWithDefaults);
 
-    this.azureOpenAIEmbeddingsApiDeploymentName =
-      (fieldsWithDefaults?.azureOpenAIEmbeddingsApiDeploymentName ||
-        fieldsWithDefaults?.azureOpenAIApiDeploymentName) ??
-      (getEnvironmentVariable("AZURE_OPENAI_API_EMBEDDINGS_DEPLOYMENT_NAME") ||
-        getEnvironmentVariable("AZURE_OPENAI_API_DEPLOYMENT_NAME"));
+    this.azureOpenAIApiDeploymentName =
+      fieldsWithDefaults?.azureOpenAIApiDeploymentName ??
+      getEnvironmentVariable("AZURE_OPENAI_API_DEPLOYMENT_NAME");
```
jacoblee93, line 53: I actually think we want to keep this env var for embeddings and have it use the current behavior - it's apparently common for completions vs. embeddings endpoints.

Yes let's make the change on the input field name though.

jacoblee93:
CC @sarangan12 @sinedied this is technically breaking but given the naming is very confusing, I think it'd be good to do now.

However, looks like ~7k downloads a week, which is more than 0: https://www.npmjs.com/package/@langchain/azure-openai

What do you think?

jacoblee93:
Discussed with some MS folks, I think the preference would be to avoid a breaking change, even though it's early.

I can patch this up later if you'd like.

SpringMT:
Thanks for your offer.
The PR does not neet to merge immediately, so you feel free to address it later when it's more convenient.
Thx!

jacoblee93:
Updated to restore backwards compatibility 👍

jacoblee93:
Thank you!

PR #4726, opened 2024-03-12
multiple[patch]: Switch deprecated `model.call` to `model.invoke` in tests and examples

janvi-kalra:
No-op. 
Noticed that `model.call` was deprecated so updating the library and documentation to `mode.invoke` 
Starter PR just to get the lay of the land 
Ran `yarn test` and all tests continued to pass.


docs/core_docs/.gitignore
```diff
@@ -43,6 +43,8 @@ docs/use_cases/question_answering/sources.md
 docs/use_cases/question_answering/sources.mdx
```
janvi-kalra, line 43: seems to have happened automatically - didn't manually change this

jacoblee93: Yeah, all good there

jacoblee93:
Fantastic, thank you!

jacoblee93:
Thank you! And congrats on your first PR!

PR #4727, opened 2024-03-12
anthropic[patch]: `ChatAnthropicTools`: `fixArrayXMLParameters` improvements

davidfant:
It seems like `ChatAnthropicTools` is just temporary until Anthropic officially releases tool calling: https://github.com/langchain-ai/langchainjs/pull/4618#discussion_r1511983435

This PR fixes some issues I've had when using `ChatAnthropicTools` with Opus:
1. Validating nested properties in the JSON schema failed
2. Opus returns number arrays like `[1,2,3]` or `1,2,3` consistently for me, but the current code parsed that as `["[1,2,3]"]` and `["1,2,3"]` respectively


libs/langchain-anthropic/src/experimental/utils/tool_calling.ts
```diff
@@ -74,9 +74,18 @@ export function fixArrayXMLParameters(
...
     // Crawl for lists indistinguishable from single items
     if (schema.properties && schema.properties[key] && schemaType === "array") {
-      fixedParameters[key] = Array.isArray(xmlParameters[key])
-        ? xmlParameters[key]
-        : [xmlParameters[key]];
+      const value = xmlParameters[key];
+      if (Array.isArray(value)) {
+          fixedParameters[key] = value;
+      } else if (typeof value === 'string') {
+        if (value.startsWith("[") && value.endsWith("]")) {
```
davidfant, line 81: Some of these heuristics might not be good, but it resolved the problems I encountered

jacoblee93: I think looks reasonable - basically it's sometimes returning an array or object as JSON?

jacoblee93: I guess there's the very off chance that someone asked for a JSON string - how commonly were you seeing this and with what params?

davidfant: I specifically saw this for number arrays and there it always responded with `<key>[1,2,3]</key>` or <key>1,2,3</key>, never the proper XML way.

I was using Opus with temperature 0

jacoblee93: Oh right, and we expect it to be an array based on the schema anyway. Then this makes sense, thank you!

jacoblee93 (approved):

libs/langchain-anthropic/src/experimental/utils/tool_calling.ts
```diff
@@ -91,7 +100,13 @@ export function fixArrayXMLParameters(
       typeof xmlParameters[key] === "object" &&
       xmlParameters[key] !== null
     ) {
-      fixedParameters[key] = fixArrayXMLParameters(schema, xmlParameters[key]);
+      fixedParameters[key] = fixArrayXMLParameters(
+        {
+          ...schema.properties[key],
+          definitions: schema.definitions,
```
jacoblee93, line 106: What's the purpose of the change here? TypeScript complains:

Screenshot 2024-03-13 at 10 25 09 AM

jacoblee93:
@davidfant was able to verify - some typing issue though. Why mutate the input into `fixArrayParams`?

PR #4731, opened 2024-03-13
core[patch]: Fix stream events and stream log with global backgrounded callbacks, add warning

jacoblee93:
Makes certain callback handlers always awaited.

CC @nfcampos @hinthornw

### DIFF
--- a/langchain-core/src/callbacks/manager.ts
+++ b/langchain-core/src/callbacks/manager.ts
@@ -20,6 +20,19 @@ import { consumeCallback } from "./promises.js";
 import { Serialized } from "../load/serializable.js";
 import type { DocumentInterface } from "../documents/document.js";
 
+if (
+  getEnvironmentVariable("LANGCHAIN_TRACING_V2") === "true" &&
+  getEnvironmentVariable("LANGCHAIN_CALLBACKS_BACKGROUND") !== "true"
+) {
+  console.warn(
+    [
+      "[WARN]: You have enabled LangSmith tracing without backgrounding callbacks.",
+      "[WARN]: If you are not using a serverless environment where you must wait for tracing calls to finish,",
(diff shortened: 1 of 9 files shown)


langchain-core/src/callbacks/manager.ts
```diff
@@ -20,6 +20,19 @@ import { consumeCallback } from "./promises.js";
...
 
+if (
+  getEnvironmentVariable("LANGCHAIN_TRACING_V2") === "true" &&
+  getEnvironmentVariable("LANGCHAIN_CALLBACKS_BACKGROUND") !== "true"
+) {
+  console.warn(
+    [
+      "[WARN]: You have enabled LangSmith tracing without backgrounding callbacks.",
+      "[WARN]: If you are not using a serverless environment where you must wait for tracing calls to finish,",
+      `[WARN]: we suggest setting "process.env.LANGCHAIN_CALLBACKS_BACKGROUND=true" to avoid additional latency.`,
```
hinthornw, line 31: Makes sense. We may want to do this in the `traceable` HOF in the SDK as well

langchain-core/src/tracers/log_stream.ts
```diff
@@ -320,7 +320,7 @@ export class LogStreamCallbackHandler extends BaseTracer {
   name = "log_stream_tracer";
 
   constructor(fields?: LogStreamCallbackHandlerInput) {
-    super(fields);
+    super({ _awaitHandler: true, ...fields });
```
hinthornw, line 323: oh cool

hinthornw (approved):

langchain-core/src/callbacks/manager.ts
jacoblee93: Yes let's do it

jacoblee93:
With message history and run collector also did not work with this setting, fixes those too

PR #4738, opened 2024-03-13
langchain[patch]: Add possibility to rerank retrieved docs in ParentDocumentRetriever and MultiQueryRetriever

karol-f:
With many documents from the vector store that are passed to LLM, final answers sometimes consist of information from
irrelevant documents, making it less precise and sometimes incorrect. Also, passing multiple irrelevant documents makes it
more expensive.
So there are two reasons to use rerank - precision and costs.

_Side note: there is currently Cohere SDK error on some environments with new Node/Next - https://github.com/cohere-ai/cohere-typescript/issues/136. It should be solved in the upcoming release of Cohere SDK._


langchain/src/retrievers/parent_document.ts
```diff
@@ -81,6 +84,10 @@ export class ParentDocumentRetriever extends MultiVectorRetriever {
     | VectorStoreRetrieverInterface<VectorStoreInterface>
     | undefined;
 
+  documentCompressor: BaseDocumentCompressor | undefined;
+
+  protected documentCompressorrMinRelevanceScore?: number = 0.4;
```
jacoblee93, line 89: One `r` here

karol-f: You are right, fixed!

examples/src/retrievers/parent_document_retriever_rerank.ts
```diff
@@ -0,0 +1,66 @@
+import { OpenAIEmbeddings } from "@langchain/openai";
+import { CohereRerank } from "@langchain/cohere";
```
jacoblee93, line 2: So this doesn't yet implement `BaseDocumentCompressor` - I will fix that before merging this.

karol-f: Thank you

jacoblee93:
Really sorry for the delay - I don't think I'll have time to implement this for Cohere soon.

If you'd like, we can remove the docs/example and get this merged sooner?

karol-f:
@jacoblee93 Sounds perfect. I've removed examples. Regards!

jacoblee93:
Hi really sorry for the delay - @bracesproul or I will address this on Monday now that DocumentCompressor is in core

karol-f:
> Hi really sorry for the delay - @bracesproul or I will address this on Monday now that DocumentCompressor is in core

Fantastic!

examples/src/retrievers/parent_document_retriever_rerank.ts
karol-f: Hi, I've changed a code so you can feed any document compressor to both Parent Retreiver and Multi Query Retreiver so it's much cleaner and it's up to developer what to use.

karol-f:
@jacoblee93 I've refactored the above code, and now it's document compressor agnostic so you can use Cohere reranking (as shown in the example) or any other solution that fit's you.

Tested on the real app using NPM library Patch Package.

jacoblee93:
Sorry about the delay!

jacoblee93:
Thank you for your patience!

karol-f:
Thank you for the good work. You are awesome maintainer!

PR #4739, opened 2024-03-13
community[patch]: Make PGVector constructors public to allow for pool reuse

jacoblee93:


examples/src/indexes/vector_stores/pgvector_vectorstore/pgvector_pool.ts
bracesproul, line 1: this would be better in a notebook.

imo even if you have to push up the docs in a separate PR after cutting a release w/ the update, it would be worth it since this is a large clunky file

jacoblee93: I think not worth the effort

PR #4743, opened 2024-03-14
core[patch]: Merge response metadata fields

jacoblee93:
CC @baskaryan


langchain-core/src/messages/index.ts
```diff
@@ -248,6 +248,70 @@ function isOpenAIToolCallArray(value?: unknown): value is OpenAIToolCall[] {
   );
 }
 
+function _mergeDicts(
```
jacoblee93, line 251: Flagging as high risk

PR #4753, opened 2024-03-14
core[patch], langchain[patch]: `LangSmith` tracing improvements for `OutputFixingParser` 

davidfant:
Problem: when using `OutputFixingParser`, the tracing with LangSmith doesn't work properly. This is because:
1. `BaseLLMOutputParser` doesn't pass on `callbacks` to `parseResult`
2. `OutputFixingParser` uses LLMChain, whose `.call` function for some reason doesn't pass along callbacks. When I create the following runnable myself and pass it in as `retryChain` in my PR, it works
```
retryChain: NAIVE_FIX_PROMPT
        .pipe(fields.extractLLM ?? fields.llm)
        .pipe(new StringOutputParser()) as any
```

### Before (LangSmith shows multiple traces)
image

### After (LangSmith trace shows the parsing traces as part of the parent trace)
image


langchain/src/output_parsers/fix.ts
```diff
@@ -72,16 +79,20 @@ export class OutputFixingParser<T> extends BaseOutputParser<T> {
...
-        );
-        const newCompletion: string = result[this.retryChain.outputKey];
-        return this.parser.parse(newCompletion);
+        const retryInput = {
+          instructions: this.parser.getFormatInstructions(),
+          completion,
+          error: e,
+        };
+
+        if (this.retryChain instanceof LLMChain) {
```
jacoblee93, line 88: `instanceof` is a bit dangerous here and will fail lint - use a typeguard instead

jacoblee93:
This is awesome, thank you!

PR #4758, opened 2024-03-14
community[patch]: bedrock: add model name to identifying params (cache keys)

jarib:
I noticed that when caching is enabled with `BedrockChat`, I get the same answers from completely different models. So it seems the model name is missing from the cache key. 

This PR adds a test and attempts to solve the problem, but I'm not sure if it's the correct solution. Any help appreciated.


libs/langchain-community/src/chat_models/tests/chatbedrock.test.ts
jacoblee93, line 1: Mark as an integration test instead (since it requires outside resources) by renaming it to `chatbedrock.int.test.ts`

jacoblee93: I can do it

jacoblee93:
Thank you!

PR #4761, opened 2024-03-14
community[minor]: Implement addGraphDocuments method for neo4j graph

easwee:
Porting over add_graph_documents method from python.


libs/langchain-community/src/graphs/graph_document.ts
```diff
@@ -0,0 +1,53 @@
+import { Serializable } from "@langchain/core/load/serializable";
+import { Document } from "@langchain/core/documents";
+
+export class Node implements Serializable {
+    id: string | number;
+    type: string;
```
adam-cowley, line 6: `label: string` or `labels: string[]` ?

easwee: Based on https://github.com/langchain-ai/langchain/blob/514fe807784f520449914a64ffc983538fa6743d/libs/community/langchain_community/graphs/graph_document.py#L20 it's just string.

libs/langchain-community/src/graphs/neo4j_graph.ts
```diff
@@ -1,4 +1,6 @@
 import neo4j, { RoutingControl } from "neo4j-driver";
+import { createHash } from "crypto";
```
jacoblee93, line 2: I am not sure if this is intended to work in web environments (e.g. Next.js) but I would suggest using this:

https://github.com/langchain-ai/langchainjs/blob/09db56a3521fd96db33a2deaf8378b2ab29cbb57/langchain/src/embeddings/cache_backed.ts#L1

libs/langchain-community/src/graphs/graph_document.ts
```diff
@@ -0,0 +1,69 @@
...
+  ) {
+    super();
+    this.source = source;
+    this.target = target;
+    this.type = type;
+    this.properties = properties;
+  }
+}
+
+export class GraphDocument extends Serializable {
```
jacoblee93, line 54: Could this extend `Document` instead? It will be a bit more interchangeable with other RAG pipeline things in the framework if we do it that way.

easwee: Updated - please review

tomasonjo:
LGTM

libs/langchain-community/src/graphs/graph_document.ts
```diff
@@ -0,0 +1,86 @@
...
+    id,
+    type = "Node",
+    properties = {},
+  }: {
+    id: string | number;
+    type: string;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    properties?: Record<string, any>;
+  }) {
+    super();
```
jacoblee93, line 24: Should technically pass in all inputs `fields` here to truly make it serializable but no worries

libs/langchain-community/src/graphs/graph_document.ts
```diff
@@ -0,0 +1,86 @@
...
+    this.properties = properties;
+  }
+}
+
+export class GraphDocument extends Document {
+  nodes: Node[];
+
+  relationships: Relationship[];
+
+  source: Document;
```
jacoblee93, line 68: Unnest this since we're now extending `Document` directly? Otherwise there's no point

jacoblee93: Basically implement `pageContent` and `metadata`

jacoblee93:
Really sorry for the delay - see above.

libs/langchain-community/src/graphs/graph_document.ts
tomasonjo: Would consisteny between python and JS be a higher priority? IMO, there is not much value in extending the doc for rag, as we never use the graph doc in retrieval setting, it's only used for ingestion atm. Even if we wanted to use it in retrieval, it would need custom components since none use nodes or rels atm

jacoblee93: Ok, fine by me.

jacoblee93:
Thank you for your patience!

libs/langchain-community/src/graphs/graph_document.ts
easwee: Will improve this in next PR that extends and uses this class.

PR #4778, opened 2024-03-15
anthropic[minor]: fix nested object fixArrayXMLParameters

alvarosevilla95:
When parsing a nested object in fixArrayXMLParameters, the base schema is passed on the recursive call. This causes a mismatch between the xml object (the nested type) and the schema (still the base type)

I'm not very happy with type assertion for the schema, not sure what's the correct check to make here though.


jacoblee93:
@davidfant I think you had this in your PR initially too?

jacoblee93:
Thank you!

libs/langchain-anthropic/src/experimental/utils/tool_calling.ts
```diff
@@ -100,7 +100,10 @@ export function fixArrayXMLParameters(
       typeof xmlParameters[key] === "object" &&
       xmlParameters[key] !== null
     ) {
-      fixedParameters[key] = fixArrayXMLParameters(schema, xmlParameters[key]);
+      fixedParameters[key] = fixArrayXMLParameters(
+        schema.properties[key] as JsonSchema7ObjectType,
```
jacoblee93, line 104: @davidfant had:

```
{
        ...schema.properties[key],
        definitions: schema.definitions,
}
```

for the first arg - does that make sense here?

davidfant: This will break JSON schema definitions. TBH I'm not sure how well this XML situation would even work with JSON schema definitions though, so it might not practically matter, but I'd still prefer adding in definitions

jacoblee93: Will change it to yours

jacoblee93:
Thank you both!

PR #4795, opened 2024-03-15
api_refs[minor]: Inherit class properties from parent classes

bracesproul:


docs/api_refs/package.json
```diff
@@ -7,7 +7,6 @@
     "typedoc:build": "npx typedoc --options typedoc.json",
     "build:scripts": "node ./scripts/create-entrypoints.js && yarn typedoc:build && node ./scripts/update-typedoc-css.js",
     "build": "yarn clean && yarn build:scripts && next build",
-    "build:vercel": "yarn clean && yarn build:scripts && next build",
```
bracesproul, line 10: Leftover from when we used a separate script. not needed anymore

PR #4799, opened 2024-03-16
openai[patch]: Bugfix: try catch around JSON stringify/parse when counting OpenAI function calling tokens

davidfant:
If OpenAI doesn't return valid JSON in arguments (which is rare, but happens) the OpenAI invocation just throws because token counting fails. Counting the tokens without removing newlines/spaces (which I assume is needed bc it was already there, but I don't know for sure), should be a good enough approximation of token count


libs/langchain-openai/src/chat_models.ts
```diff
@@ -730,14 +730,21 @@ export class ChatOpenAI<
...
+              JSON.stringify(
+                JSON.parse(
+                  openAIMessage.additional_kwargs.function_call?.arguments
+                )
               )
-            )
-          );
+            );
+          } catch (error) {
+            console.error("Error parsing function arguments", error, JSON.stringify(openAIMessage.additional_kwargs.function_call));
```
davidfant, line 743: Your call how you want to approach this kind of logging. I've had this as a patch-package patch for a while. Logging this is useful bc then I can easily see when arguments is invalid JSON

jacoblee93: Yes, this seems reasonable, thank you!

jacoblee93:
Thank you for this!

PR #4803, opened 2024-03-17
Add mistral as supported provider for AWS Bedrock

jarib:
Mistral is now supported in AWS Bedrock. 

```
$ aws bedrock list-foundation-models --region us-east-1 --query 'modelSummaries[*].modelId' | grep mistral
    "mistral.mistral-7b-instruct-v0:2",
    "mistral.mixtral-8x7b-instruct-v0:1"
$
```

Langchain has a duplicated list of valid models. I'd argue that they both can be removed. Langchain doesn't need to validate the models before passing them on to Bedrock.

### DIFF
--- a/libs/langchain-community/src/chat_models/tests/chatbedrock.int.test.ts
+++ b/libs/langchain-community/src/chat_models/tests/chatbedrock.int.test.ts
@@ -21,21 +21,21 @@ import { BedrockChat } from "../bedrock/web.js";
 void testChatModel(
   "Test Bedrock chat model: Claude-v2",
   "us-east-1",
-  "anthropic.claude-v2",
+  "mistral.mistral-7b-instruct-v0:2",
   "What is your name?"
 );
 
 void testChatStreamingModel(
   "Test Bedrock chat model streaming: Claude-v2",
   "us-east-1",
-  "anthropic.claude-v2",
+  "mistral.mistral-7b-instruct-v0:2",
   "What is your name and something about yourself?"
 );
 
 void testChatHandleLLMNewToken(
   "Test Bedrock chat model HandleLLMNewToken: Claude-v2",
   "us-east-1",
-  "anthropic.claude-v2",
+  "mistral.mistral-7b-instruct-v0:2",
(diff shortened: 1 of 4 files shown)


jacoblee93:
The issue is in the input/output parsing - I think there might be some other changes needed in the adapter?

esoler-sage:
I think provider needs to be set up in `prepareOutput` (it can use [InvokeModelWithResponseStream]) function and `prepareInput` function as well (max_tokens and stop are the properties that changes for the inputBody)

jarib:
Ah, I missed that. I've merged @esoler-sage's fix, thanks!

jacoblee93:
Thank you! FYI @efriis

libs/langchain-community/src/chat_models/tests/chatbedrock.int.test.ts
```diff
@@ -21,21 +21,21 @@ import { BedrockChat } from "../bedrock/web.js";
 void testChatModel(
   "Test Bedrock chat model: Claude-v2",
   "us-east-1",
-  "anthropic.claude-v2",
+  "mistral.mistral-7b-instruct-v0:2",
   "What is your name?"
```
jarib, line 25: Shouldn't this be added as separate tests? Claude support is pretty important too :D

efriis:
Good PR!

A lot of these if statements / double validation will be removed in an upcoming refactor of bedrock on the python side, and @jacoblee93 will probably tap in to do something similar in js. I believe the reason for validation is that Bedrock models don't have a standard api for their inputs and arguments, so checking and throwing an error fails quickly instead of bedrock introducing a new model that doesn't have a supported input/args format, and producing an output that doesn't make sense.

esoler-sage:
@efriis I was about to create 2 PR... token count on Bedrock Models and add different chat prompts to improve this line: https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-community/src/chat_models/bedrock/web.ts#L295.

For this second, probably it's better to wait until this refactoring is done?

PR #4814, opened 2024-03-18
redis[patch]: Escape additional metadata characters for RedisSearch metadata filtering usage

leejona16:
The `:` and `"` characters found in the stringified JSON are special characters in redisSearch queries, so we need to escape them in the stringified metadata if we want to be able to search the entire JSON object.

This enables key/value sub-string searches of the stringified metadata object for filtering purposes.


jacoblee93:
Thank you!

jacoblee93:
Could you fix the failing test?

PR #4820, opened 2024-03-19
community[patch]: faat: Add ERNIE-Speed-8K, ERNIE-Speed-128K from Baidu Wenxin

wizardAEI:
ERNIE-Speed is Baidu's latest high-performance large language model released in 2024, with excellent general capabilities, better at handling specific scenario questions, and also has excellent inference performance.

It has been tested, and I can provide the baiduApiKey and baiduSecretKey if needed.


jacoblee93:
Awesome, thank you!

PR #4848, opened 2024-03-21
community[patch]: Neo4j similaritySearchWithScore crashes on undefined

### LINKED ISSUE #4752: Neo4j similaritySearchWithScore crashes on undefined
### Checked other resources

### Description

Trying to get similarity results with score like this:
```ts
    const result = await vectorStore.similaritySearchWithScore('What is foo?', 6);
```
Produces the following error in the `removeLuceneChars` function:
> modifiedText = modifiedText.split(char).join(' ');
> TypeError: Cannot read properties of undefined (reading 'split')

The problem is that is trying to process the query, that is not passed from the base class. To my understanding it is not needed. Calling `similaritySearchWithScore` just preprocesses the query into vector and passing it to `similaritySearchVectorWithScore`.

The workaround I implemented currently is creating the vector myself, calling `similaritySearchVectorWithScore` directly and passing empty query:
```ts
const queryVector = await embeddings.embedQuery('What is foo?');
const result = await vectorStore.similaritySearchVectorWithScore(queryVector, 6, '');
```

https://github.com/langchain-ai/langchainjs/blob/37d68a5b1a2ab1039e1459cd6a35c43ae8e3ddd6/libs/langchain-community/src/vectorstores/neo4j_vector.ts#L618

### System Info

pnpm  8.6.2
node 20.10.0
"@langchain/community": "^0.0.37",
 "@langchain/core": "^0.1.45",
 "@langchain/openai": "^0.0.19",
"langchain": "^0.1.27",

easwee:
Fixes #4752

### DIFF
--- a/libs/langchain-community/src/vectorstores/neo4j_vector.ts
+++ b/libs/langchain-community/src/vectorstores/neo4j_vector.ts
@@ -303,8 +303,8 @@ export class Neo4jVectorStore extends VectorStore {
         RETURN reduce(str='', k IN ${JSON.stringify(textNodeProperties)} |
         str + '\\n' + k + ': ' + coalesce(node[k], '')) AS text,
         node {.*, \`${embeddingNodeProperty}\`: Null, id: Null, ${textNodeProperties
-          .map((prop) => `\`${prop}\`: Null`)
-          .join(", ")} } AS metadata, score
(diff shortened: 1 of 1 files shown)


jacoblee93:
Thank you!

PR #4849, opened 2024-03-21
langchain[patch]: fix: Include separator length when checking chunk size

Dschoordsch:
When checking splits for chunk size, only 1 separator was considered, this could lead to chunks exceeding the maximum size.


langchain/src/text_splitter.ts
```diff
@@ -188,8 +188,7 @@ export abstract class TextSplitter
     for (const d of splits) {
       const _len = await this.lengthFunction(d);
       if (
-        total + _len + (currentDoc.length > 0 ? separator.length : 0) >
-        this.chunkSize
+        total + _len + (currentDoc.length * separator.length) > this.chunkSize
```
Dschoordsch, line 191: The [python version](https://github.com/langchain-ai/langchain/blob/5d220975fc563a92f41aeb0907e8c3819da073f5/libs/text-splitters/langchain_text_splitters/base.py#L138) solves this slightly differently by adding and substracting the separator.length to total which adds a bit more clutter
* https://github.com/langchain-ai/langchain/blob/5d220975fc563a92f41aeb0907e8c3819da073f5/libs/text-splitters/langchain_text_splitters/base.py#L133
* https://github.com/langchain-ai/langchain/blob/5d220975fc563a92f41aeb0907e8c3819da073f5/libs/text-splitters/langchain_text_splitters/base.py#L138
* https://github.com/langchain-ai/langchain/blob/5d220975fc563a92f41aeb0907e8c3819da073f5/libs/text-splitters/langchain_text_splitters/base.py#L142

jacoblee93:
Thank you for this, and sorry for the delayed review!

langchain/src/text_splitter.ts
jacoblee93: Shouldn't this be additive rather than multiplicative though?

jacoblee93: Let me think through it a bit more! It's been a while since I've looked at this code.

jacoblee93: Oh, never mind, the naming is bad. I understand now, makes sense!

PR #4857, opened 2024-03-22
community[patch]: ElasticVectorSearch - fix error when no bulk operations 

telekosmos:
Fixes bug in discussion # ([discussion](https://github.com/langchain-ai/langchain/discussions/19396))

Description is in the attached issue/discussion. Basically, it was observed on `incremental` cleanup when calling the `index(...)` function a crash when trying to delete an empty list of elements. The deletion in Elastic is done by a `POST _bulk` request, which expects a non-emtpy list of operations. When there is no elements to delete, the list of operations is empty, hence the error. Easy to replicate straight in the Elastic Console (i.e. http://localhost:5601/app/dev_tools#/console) by writing just `POST _bulk`.

### DIFF
--- a/libs/langchain-community/src/vectorstores/elasticsearch.ts
+++ b/libs/langchain-community/src/vectorstores/elasticsearch.ts
@@ -193,7 +193,8 @@ export class ElasticVectorSearch extends VectorStore {
         _index: this.indexName,
       },
     }));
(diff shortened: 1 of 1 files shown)


jacoblee93:
Makes sense. Thank you!

PR #4864, opened 2024-03-24
google-common [major] / google-webauth [minor] / google-gauth [minor]: Gemini functions

### LINKED ISSUE #4253: [Feature Request] Google Gemini Vertex AI Function Calling
The new [@google-cloud/vertexai](https://github.com/googleapis/nodejs-vertexai) library makes it very easy to use function calling with Google Gemini Models. It would be awesome to see more models other than OpenAI that supports Function Calling in Langchain js 👍

afirstenberg:
Adds support for function calling through langchain tools.
Adds support for bind().
Adjust testing, since AI Studio has gotten more strict.

Tested with AI Studio and Vertex AI. (Note this does *not* change the google-genai package.)

Missing:
- Support for withStructuredOutput() (@jacoblee93 says he will handle)
- Documentation

Fixes #4253 

### New exports:
`@langchain/google-*/{types,utils}`
- I opted to export from separate entrypoints because if I exported all types/utils from the root entrypoint of `common` then we would be stuck with two bad options:
1. individually import & re-export modules inside packages like `gauth`/`vertexai`, etc so we aren't also exporting all the base classes from common. This means we will also need to remember to update the exports & cut releases for all these packages when a new type/util is added that we want to export.
2. re-export everything, including base classes. This would confuse users as they'd see `ChatGoogle`, `ChatVertexAI`, etc, etc in IDE autocompletes when looking at exports from, for example `@langchain/google-vertexai`

- Added a new util `zodToGeminiParameters` since the google function definition does not cleanly allow for passing the result of `zodToJsonSchema` into the `parameters` field of a function.


libs/langchain-google-common/src/connection.ts
```diff
@@ -277,6 +281,57 @@ export abstract class AbstractGoogleLLMConnection<
...
+        Array.isArray((tool as StructuredToolInterface).lc_namespace)
+      )
+    );
+  }
+
+  structuredToolToFunctionDeclaration(
+    tool: StructuredToolInterface
+  ): GeminiFunctionDeclaration {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const jsonSchema = zodToJsonSchema(tool.schema) as any;
```
bracesproul, line 298: Why is this being casted as `any`? The built in types should be enough?

libs/langchain-google-common/src/utils/gemini.ts
```diff
@@ -78,27 +87,70 @@ export function messageContentToParts(content: MessageContent): GeminiPart[] {
...
-  const parts: GeminiPart[] = messageContent.map((content) => {
-    // eslint-disable-next-line default-case
-    switch (content.type) {
-      case "text":
-        return messageContentText(content);
-      case "image_url":
-        return messageContentImageUrl(content);
+  const parts: GeminiPart[] = messageContent
+    .map((content) => {
+      // eslint-disable-next-line default-case
```
bracesproul, line 92: I think we want to throw an error if an unsupported type comes through here, instead of silently failing.

bracesproul:
```typescript
default:
  throw new Error(`Unsupported type received while converting message to message parts: ${content.type}`)
```

libs/langchain-google-common/package.json
```diff
@@ -39,7 +39,9 @@
   "author": "LangChain",
   "license": "MIT",
   "dependencies": {
-    "@langchain/core": "~0.1.1"
+    "@langchain/core": "~0.1.1",
+    "zod": "^3.22.4",
```
bracesproul, line 43: Zod is only used in a `test.ts` file. Should be inside `devDependencies` instead

bracesproul (changes requested):
Couple comments/requests. At a high level it looks very very good!

Going to merge/release the new vertex packages in a couple hours, and if you haven't resolved my requests here I'll just push them up and get this released too.

Thanks for your hard work on these Allen! Will be great to get them into the library

libs/langchain-google-common/src/connection.ts
afirstenberg: The issue is that I needed to remove two attributes from `jsonSchema` (which happens on line 301). Leaving it in as the built-in type was giving me errors in typescript (or in my IDE's linting of it).

libs/langchain-google-common/src/utils/gemini.ts
afirstenberg: At the moment, this can only have these two values.

I'm uncomfortable with the notion that a future update that does introduce a new content type could break this code and prevent it from working until we get a fix in.

But I see that's how it's handled elsewhere... so I guess I have no strong objection.

libs/langchain-google-common/package.json
afirstenberg: Huh! So it is.

docs/core_docs/docs/integrations/chat/google_palm.mdx
```diff
@@ -28,3 +29,97 @@ the model.
 import GooglePaLMExample from "@examples/models/chat/integration_googlepalm.ts";
 
 <CodeBlock language="typescript">{GooglePaLMExample}</CodeBlock>
+
+# ChatGoogleVertexAI
```
jacoblee93, line 33: This isn't the right header is it?

jacoblee93: Why are we adding this under `PaLM`?

docs/core_docs/docs/integrations/llms/google_palm.mdx
```diff
@@ -28,3 +29,100 @@ the model.
 import GooglePaLMExample from "@examples/models/llm/googlepalm.ts";
 
 <CodeBlock language="typescript">{GooglePaLMExample}</CodeBlock>
+
+# Google Vertex AI
```
jacoblee93, line 33: See above - why are we modifying legacy pages?

docs/core_docs/docs/integrations/platforms/google.mdx
```diff
@@ -69,9 +74,86 @@ const input2 = [
...
+:::tip
+Click [here](/docs/integrations/chat/google_vertex_ai) for the `@langchain/google-vertexai` specific integration docs
+:::
+
+</TabItem>
+</Tabs>
+
+The value of `image_url` must be a base64 encoded image (e.g., `data:image/png;base64,abcd124`).
+
+### Vertex AI (Legacy)
```
jacoblee93, line 156: Would just remove this from the platform page

examples/src/models/chat/integration_googlevertexai-tools.ts
```diff
@@ -0,0 +1,62 @@
...
+    {
+      "id": "calculator",
+      "type": "function",
+      "function": {
+        "name": "calculator",
+        "arguments": "{\"number2\":81623836,\"number1\":1628253239,\"operation\":\"multiply\"}"
+      }
+    }
+  ],
+  "tool_calls_parsed": [
```
jacoblee93, line 47: What's this?

libs/langchain-google-common/README.md
```diff
@@ -43,5 +43,10 @@ Tasks and services still to be implemented:
 * Multimodal embeddings
 * Vertex AI Search
 * Vertex AI Model Garden
+  * Online prediction endpoints
```
jacoblee93, line 46: Should also update table here:

https://js.langchain.com/docs/integrations/chat/

libs/langchain-google-common/src/chat_models.ts
```diff
@@ -74,7 +91,7 @@ export abstract class ChatGoogleBase<AuthOptions>
 
   lc_serializable = true;
 
-  model = "gemini-pro";
+  modelName = "gemini-pro";
```
jacoblee93, line 94: Is this not breaking?

jacoblee93: No worries if so, but why change

libs/langchain-google-common/src/chat_models.ts
```diff
@@ -210,4 +227,172 @@ export abstract class ChatGoogleBase<AuthOptions>
...
+    | Runnable<
+        BaseLanguageModelInput,
+        { raw: BaseMessage; parsed: RunOutput }
+      > {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    let schema: z.ZodType<RunOutput> | Record<string, any>;
+    let name;
+    let method;
+    let includeRaw;
+    if (isStructuredOutputMethodParams(outputSchema)) {
```
jacoblee93, line 276: Let's not support this in new implementations - it's just for legacy support

libs/langchain-google-common/src/chat_models.ts
```diff
@@ -210,4 +227,172 @@ export abstract class ChatGoogleBase<AuthOptions>
...
+        raw: llm,
+      },
+      parsedWithFallback,
+    ]).withConfig({
+      runName: "StructuredOutputRunnable",
+    });
+  }
+}
+
+function isZodSchema<
```
jacoblee93, line 377: Let's make sure we factor these into core, can you make an issue?

libs/langchain-google-common/src/connection.ts
```diff
@@ -159,7 +163,7 @@ export abstract class GoogleAIConnection<
   extends GoogleHostConnection<CallOptions, GoogleLLMResponse, AuthOptions>
   implements GoogleAIBaseLLMInput<AuthOptions>
 {
-  model: string;
+  modelName: string;
```
jacoblee93, line 166: It's fine if there's a reason since this is all undocumented but why the change?

libs/langchain-google-common/src/utils/common.ts
```diff
@@ -1,26 +1,40 @@
...
-  return copyAIModelParamsInto(params, {});
+  params: GoogleAIModelParams | undefined,
+  options: GoogleAIBaseLanguageModelCallOptions | undefined
+): GoogleAIModelRequestParams {
+  return copyAIModelParamsInto(params, options, {});
 }
 
 export function copyAIModelParamsInto(
   params: GoogleAIModelParams | undefined,
+  options: GoogleAIBaseLanguageModelCallOptions | undefined,
```
jacoblee93, line 18: Why not make this a third param instead of adding it into the middle?

jacoblee93:
Thanks both of you! Mostly nits, and looks like build is also failing.

libs/langchain-google-common/src/chat_models.ts
bracesproul: Technically yes, but these were never documented before, and only have ~250 downloads/week.

Removed to keep consistency

libs/langchain-google-common/src/utils/common.ts
afirstenberg: The function is only used internally with the last parameter being the target object.
I could have restructured it to take a variable number of parameters and use them to determine the order that the object should be copied in... But I'm not sure that was worth it.
One could make the argument that the first two parameters should be swapped I suppose... But I think we're over-thinking this now.

libs/langchain-google-common/src/utils/zod_to_gemini_parameters.ts
```diff
@@ -0,0 +1,14 @@
+import type { z } from "zod";
+import { zodToJsonSchema } from "zod-to-json-schema";
+import { GeminiFunctionSchema } from "../types.js";
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function zodToGeminiParameters(
```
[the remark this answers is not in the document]
afirstenberg: I'm not sure why this was broken out into a separate file. I don't _think_ I did so.

docs/core_docs/docs/integrations/llms/google_palm.mdx
bracesproul: They're not actually getting modified. The previous vertex page had docs on what is now legacy plam code, so I'm moving the content around so it's all on the proper pages now. (The headers were wrong though)

docs/core_docs/docs/integrations/platforms/google.mdx
bracesproul: Removed everything except the tip cards linking to legacy docs

examples/src/models/chat/integration_googlevertexai-tools.ts
afirstenberg: tool_calls_parsed is tool_calls, but with the JSON string parsed into an object.

libs/langchain-google-common/src/chat_models.ts
bracesproul: https://github.com/langchain-ai/langchainjs/issues/4893

libs/langchain-google-common/src/utils/zod_to_gemini_parameters.ts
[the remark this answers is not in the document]
bracesproul: @afirstenberg I did this because it's a useful util to have if you're trying to pass a zod schema to the model via the `.bind` method

examples/src/models/chat/integration_googlevertexai-tools.ts
jacoblee93: I get that it's convenient but let's remove for now - don't want people to get the wrong idea that it's a universal field.

libs/langchain-google-common/src/utils/zod_to_gemini_parameters.ts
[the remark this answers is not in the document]
afirstenberg: Two different thoughts:

1. Is there a reason it couldn't be in one of the existing utility files then? (Just seems like a weird thing to have a file all to itself.)

2a. I thought `.bind` and `.withStructuredOutput` should be able to take a zod schema already, so we handle that. If it doesn't - should it?
2b. If developers are passing in the schema, then we should sanitize it in our code and not rely on them to do so. (I mostly want this to be as compatible as possible with OpenAI so someone can easily switch between the two.)

PR #4869, opened 2024-03-25
Removes console.log from a log loop in stream events tests

rahilvora:
Fixes # (issue)


jacoblee93:
Thanks @rahilvora! I actually do want to skip this test since it's using a private endpoint I set up. We could also just fully remove the test, but I think it's nice to mark that we've thought of this case in the future.

PR #4877, opened 2024-03-25
anthropic[patch]: Apply the new format for transferring the "usage" attribute in the "m…

rechik:
…essage_delta" event of the "anthropic" provider.

Apparently, anthropic has moved "usage" out of "delta", placing them now on the same level. Here is the current view of the event.

event: message_delta
data: {"type":"message_delta","delta":{"stop_reason":"end_turn","stop_sequence":null},"usage":{"output_tokens":12}         }


jacoblee93:
Nice catch, thank you!

PR #4892, opened 2024-03-26
fix: match structured prompt serialization ID with Python, allow piping bound LLMs into prompt

dqbd:


langchain-core/src/prompts/structured.ts
```diff
@@ -8,6 +12,30 @@ import {
...
+  );
+}
+
+function isRunnableBinding(x: unknown): x is RunnableBinding<unknown, unknown> {
+  return (
+    typeof x === "object" &&
+    x != null &&
+    "lc_id" in x &&
+    Array.isArray(x.lc_id) &&
+    x.lc_id.join("/") === "langchain_core/runnables/RunnableBinding"
```
jacoblee93, line 35: We could use `.getName()`, would be simpler?

jacoblee93:
Seems fine, one small comment! Thank you!

langchain-core/src/prompts/structured.ts
dqbd: (synced offline: ideally we want to match the exact class)

PR #4897, opened 2024-03-27
openai[patch]: fix(openai): consume title from jsonschema as function name

dqbd:


jacoblee93:
CC @bracesproul we should do this for all the implementations, let's prioritize refactoring these parsing pieces into core soon.

PR #4898, opened 2024-03-27
azure-openai[minor]: update docs and fix openai api usage

sinedied:
Going through the docs I noticed the Azure OpenAI docs were not up to date with the latest version I had locally, a rebase may have gone wrong with the initial PR that created the package. So here's the updated docs.

I also noticed using an OpenAI API key was not working, so I fixed that too and added some tests for it.


sinedied:
I'm not sure why the docs deployment are failing, is there a command I can run locally to see what's wrong?

jacoblee93:
Thank you! Build failure looks like a flake.

PR #4914, opened 2024-03-28
community[patch]: Add experimental Neo4j graph transformer

easwee:
cc @tomasonjo


antonioleo89:
Awesome work, when will it be merged in main?

langchain/src/experimental/graph_transformers/llm.ts
```diff
@@ -0,0 +1,268 @@
+import { z } from "zod";
+import { BaseLanguageModel } from "@langchain/core/language_models/base";
+import { ChatPromptTemplate } from "@langchain/core/prompts";
+import { Document } from "@langchain/core/documents";
+import {
+  Node,
+  Relationship,
+  GraphDocument,
+} from "@langchain/community/graphs/graph_document";
```
easwee, line 9: @jacoblee93 A question here: do i need to add this file in some config in order to have it importable, since the IDE is complaining about this import even though the integration tests run fine?

jacoblee93: I think you might have to build `@langchain/community` locally - but we are actually planning on removing the dependency on `community` within LangChain so it may be best to move this file to an entrypoint like `experimental/llm_graph_transformers` within `community`, not `langchain`

langchain/src/experimental/graph_transformers/llm.ts
```diff
@@ -0,0 +1,268 @@
...
+
+    // Define chain
+    const schema = createSchema(allowedNodes, allowedRelationships);
+    const structuredLLM = llm.withStructuredOutput(zodToJsonSchema(schema));
+    this.chain = prompt.pipe(structuredLLM);
+  }
+
+  /**
+   * Method that processes a single document, transforming it into a graph
+   * document using an LLM based on the model's schema and constraints.
```
jacoblee93, line 187: Neat! Is there a better name for this? Maybe `GraphConstructor`?

jacoblee93:
This is pretty neat! Only thing that sticks out is that this should be in `@langchain/community`

We should really get you all a separate package at some point...

tomasonjo:
If it fits better within community, let's do that? @jacoblee93

langchain/src/experimental/graph_transformers/llm.ts
easwee: Moved to community/experimental

easwee:
> If it fits better within community, let's do that? @jacoblee93

Done.

jacoblee93:
Really sorry about the delay, will get this in tonight!

jacoblee93:
Thank you! Could use some docs but can be a followup - don't want to keep you waiting any longer

PR #4923, opened 2024-03-29
Fix: Handle Escaped Curly Braces in parseFString Function

kikoncuo:
Summary:

I've experienced and fixed an issue in the parseFString function where escaped curly braces (\}) caused an erroneous "Single '}' in template." error, disrupting the parsing of templates containing literal curly braces.
I was experiencing this specifically when I was including code as part of templates, even when I manually escaped the curly braces

There is likely a way to have code in a template to avoid this problem (Please help if someone knows), but I imagine this fix could still help some people

Fix Details:

The update modifies the nextBracket function to distinguish between escaped and unescaped closing braces, allowing for accurate parsing of templates. This ensures that escaped braces are treated as literals, preventing unnecessary errors and improving the function's handling of complex string scenarios.
Impact:

This change enhances the robustness of the parseFString function, making it more adaptable to various templates and resolving a key bug affecting users working with escaped characters in their strings.

### DIFF
--- a/langchain-core/src/prompts/template.ts
+++ b/langchain-core/src/prompts/template.ts
@@ -21,13 +21,25 @@ export const parseFString = (template: string): ParsedFStringNode[] => {
   const chars = template.split("");
   const nodes: ParsedFStringNode[] = [];
 
-  const nextBracket = (bracket: "}" | "{" | "{}", start: number) => {
+  const nextBracket = (bracket, start) => {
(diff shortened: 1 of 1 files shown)


jacoblee93:
Hey @kikoncuo! Thanks for the PR.

The proper way to escape brackets in this syntax is by doubling them (`{{` and `}}`). We should document this better.

kikoncuo:
Thanks for the help!

PR #4931, opened 2024-03-30
core[patch]: Add isZodSchema to core utils

bracesproul:


langchain-core/langchain.config.js
```diff
@@ -61,6 +61,7 @@ export const config = {
     "utils/tiktoken": "utils/tiktoken",
     "utils/types": "utils/types",
     "utils/beta_warning": "utils/beta_warning",
+    "utils/is_zod_schema": "utils/is_zod_schema",
```
jacoblee93, line 64: Feels wrong to give this its own entrypoint - I actually think adding it to `types` would be reasonable here.

langchain-core/src/utils/types/is_zod_schema.ts
```diff
@@ -0,0 +1,18 @@
+import { type z } from "zod";
```
jacoblee93, line 1: Going to rename this file to `type_guards.ts`

langchain-core/langchain.config.js
```diff
@@ -59,7 +59,7 @@ export const config = {
     "utils/stream": "utils/stream",
     "utils/testing": "utils/testing/index",
     "utils/tiktoken": "utils/tiktoken",
-    "utils/types": "utils/types",
+    "utils/types": "utils/types/index",
```
jacoblee93, line 62: We should combine some of these together at some point I think

PR #4932, opened 2024-03-30
core[minor]: Add document compressor class to core

bracesproul:


langchain-core/langchain.config.js
```diff
@@ -36,7 +36,8 @@ export const config = {
     prompt_values: "prompt_values",
     runnables: "runnables/index",
     "runnables/remote": "runnables/remote",
-    retrievers: "retrievers",
+    retrievers: "retrievers/index",
+    "retrievers/document_compressors/base": "retrievers/document_compressors/base",
```
jacoblee93, line 40: None of the other entrypoints end with `/base`

jacoblee93:
Small comment, and are we sure we're ready to commit to this?

bracesproul:
> Small comment, and are we sure we're ready to commit to this?

It's already in core in the py repo, only reason to wait would be if we wanted to make breaking changes?

jacoblee93:
Thanks!

PR #4942, opened 2024-04-01
community[patch]: llama cpp stream generation and abort by invoke method

wizardAEI:
Hi ! Recently, I have been using LangChain to develop my own application. 

When using Llama CPP, I noticed that the streaming generation solution mentioned in the documentation cannot be paused. 

However, after consulting the official documentation of node-llama-cpp, I found that the streaming and pausing functions can also be achieved using `invoke` : https://withcatai.github.io/node-llama-cpp/guide/chat-session#response-streaming. 

It only requires a slight modification of the onToken method to resolve this issue. Therefore, I made some changes to the LLM CPP part.

and there is test results:
image


jacoblee93:
Thank you! Nice catch

PR #4943, opened 2024-04-01
community[minor]: Add DuckDuckGoSearch tool

my8bit:
# Context:

Here is another try to add the DuckDuckGo as a community search tool. 
Duplicate of - [one](https://github.com/langchain-ai/langchainjs/pull/1184)  that unfortunately still stays in Open state that was reviewed by @dqbd 

I am open to any comments and suggestions for this merge request. I value your time and am willing to make the necessary changes to expedite its acceptance into the main branch as I consider these tools a valuable addition to the TS version of the community toolset. 

Thank you for the opportunity to contribute to the project!

# Change description:

DuckDuckGo tool that uses the existing mature under the hood.
Tools take the same options as the library it uses, plus the possibility to limit the max number of return results, as I experience that the initial amount might overload the context buffer with less relevant results from the end of the list

The DuckDuckGo tool inherits all options from the [underlying](https://www.npmjs.com/package/duck-duck-scrape) library, adding a feature to limit the maximum number of results returned. This addresses the possible issue of context overload by filtering out less relevant search results typically found at the end of the list.

### DIFF
--- a/libs/langchain-community/langchain.config.js
+++ b/libs/langchain-community/langchain.config.js
@@ -18,6 +18,7 @@ export const config = {
     "convex/values",
     "@rockset/client/dist/codegen/api.js",
     "discord.js",
+    "duck-duck-scrape",
     "mysql2/promise",
     "web-auth-library/google",
     "firebase-admin/app",
@@ -33,6 +34,7 @@ export const config = {
     "tools/aws_sfn": "tools/aws_sfn",
     "tools/bingserpapi": "tools/bingserpapi",
     "tools/brave_search": "tools/brave_search",
+    "tools/duckduckgo_search": "tools/duckduckgo_search",
     "tools/calculator": "tools/calculator",
     "tools/connery": "tools/connery",
     "tools/dadjokeapi": "tools/dadjokeapi",
@@ -231,6 +233,7 @@ export const config = {
   requiresOptionalDependency: [
     "tools/aws_sfn",
     "tools/aws_lambda",
+    "tools/duckduckgo_search",
     "tools/discord",
     "tools/gmail",
     "tools/google_calendar",
--- a/libs/langchain-community/package.json
+++ b/libs/langchain-community/package.json
@@ -131,6 +131,7 @@
     "dotenv": "^16.0.3",
     "dpdm": "^3.12.0",
     "dria": "^0.0.3",
+    "duck-duck-scrape": "^2.2.5",
     "eslint": "^8.33.0",
     "eslint-config-airbnb-base": "^15.0.0",
     "eslint-config-prettier": "^8.6.0",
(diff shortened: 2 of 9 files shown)


libs/langchain-community/package.json
```diff
@@ -445,6 +447,9 @@
     "dria": {
```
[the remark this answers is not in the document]
my8bit: By the way, I can put the light version of the implementation of the search results inside the source code of the tools itself, if you prefer to not have or stick to the fewer third-party dependencies. I was considering that as an option, but I've opted to integrate DuckDuckGo via a well-established third-party library to leverage its maturity and likelihood of timely updates in response to any changes by DuckDuckGo

I'm open to discussing which approach aligns best with our project's priorities and am ready to adjust based on our collective feedback.

libs/langchain-community/src/tools/duckduckgo_search.ts
```diff
@@ -0,0 +1,73 @@
+import { Tool } from "@langchain/core/tools";
+import { search, SearchOptions } from "duck-duck-scrape";
+
+/**
+ * Class for interacting with the DuckDuckGoSearch engine
+ * It extends the base Tool class and implements the _call method to
+ * perform the retrieve operation.
+ *
+ */
```
bracesproul, line 9: This should be directly above `export class DuckDuckGoSearch` so it shows up as the docstring for the class. Also, nit on the content
```suggestion
/**
 * Class for interacting with the DuckDuckGo search engine
 * It extends the base Tool class to perform retrieval.
 */
```

libs/langchain-community/src/tools/duckduckgo_search.ts
```diff
@@ -0,0 +1,73 @@
...
+   * The search options for the search using the SearchOptions interface
+   * from the duck-duck-scrape package.
+   * */
+
+  searchOptions?: SearchOptions;
+  /**
+   * @default 10
+   * The maximum number of results to return from the search. 
+   * Limiting to 10 to avoid context overload.
+   * */
```
bracesproul, line 28: Any JS/TSDoc should be directly above the code it's for
```suggestion
   */
  searchOptions?: SearchOptions;
  /**
   * @default 10
   * The maximum number of results to return from the search. 
   * Limiting to 10 to avoid context overload.
   */
```

bracesproul (changes requested):
Overall looks good, I left a couple comments. This also needs a page in the docs.

I pushed up some changes with the nits I mentioned & docs to a branch [here](https://github.com/langchain-ai/langchainjs/pull/4953) which you can merge into yours!

Thanks for adding this!

my8bit:
@bracesproul Many thanks for the review and your suggestions in the branch.

bracesproul (approved):
lgtm, thanks for adding!!

libs/langchain-community/langchain.config.js
```diff
@@ -18,6 +18,7 @@ export const config = {
     "convex/values",
     "@rockset/client/dist/codegen/api.js",
     "discord.js",
+    "duck-duck-scrape",
```
jacoblee93, line 21: I don't think this is strictly necessary?

jacoblee93:
Thank you!

PR #4954, opened 2024-04-02
core[minor]: Add runId to RunnableConfig

bracesproul:
Not implemented for LLMs yet. Py is very different from JS with base implementation, so it's not as simple as translate syntax


langchain-core/src/language_models/chat_models.ts
```diff
@@ -354,14 +355,17 @@ export abstract class BaseChatModel<
...
+      handledOptions.runId,
       undefined,
       extra,
       undefined,
       undefined,
       handledOptions.runName
     );
 
+    // eslint-disable-next-line no-param-reassign
+    delete handledOptions.runId;
```
jacoblee93, line 367: Why `delete` here?

langchain-core/src/language_models/chat_models.ts
```diff
@@ -185,13 +185,14 @@ export abstract class BaseChatModel<
...
         [messages],
-        undefined,
+        runnableConfig.runId,
         undefined,
         extra,
         undefined,
         undefined,
         runnableConfig.runName
       );
+      delete runnableConfig.runId;
```
jacoblee93, line 195: Instead of mutating the object, would it make more sense to do like:

```ts
const { runId, ...childConfig } = runnableConfig;
```

?

langchain-core/src/language_models/llms.ts
```diff
@@ -399,6 +399,27 @@ export abstract class BaseLLM<
     return output;
   }
 
+  /**
+   * @TODO implement inside generate methods.
+   */
+  static _getRunIdsArray(
```
jacoblee93, line 405: Where is this used?

jacoblee93:
See comments - also, why not implement for LLMs too?

langchain-core/src/language_models/chat_models.ts
bracesproul: there is no above?

langchain-core/src/language_models/chat_models.ts
bracesproul: Realized I could remove the deletion here, `runnableConfig` isn't used after this.

bracesproul:
@jacoblee93 added to LLMs, added some tests and cleaned up a couple other places. Should be good to go now

PR #4957, opened 2024-04-02
docs[patch]: Add wrapping library metadata to MongoDB docs example

baileympearson:
This PR adds logic similar to the implementation in Python added in https://github.com/langchain-ai/langchain/pull/13084. 

In JS, users instantiate `MongoClient`s and provide them to langchain instead of langchain instantiating the `MongoClient` for them.  As such, there's no way for langchain to add client metadata like the Python implementation does.  Additionally, there's no equivalent of `importlib.metadata` in Javascript that can return the version of a package.  So we'd like to just add this to the documentation example, so that when users use the code provided in the docs, they include the metadata as well.


jacoblee93:
Thank you!

PR #4960, opened 2024-04-02
langchain[patch]: SelfQueryRetriever - fallback to similarity search

guidev:
If there are no metadata filters (NO_FILTER), instead of explicitly returning [], it should fallback to a similarity search

This also matches the python implementation


jacoblee93:
Thank you! Sorry for the delay

PR #4983, opened 2024-04-05
Create webpack.yml

Laiiev:
Fixes # (issue)

### DIFF
--- a/.github/workflows/webpack.yml
+++ b/.github/workflows/webpack.yml
@@ -1,0 +1,28 @@
+name: NodeJS with Webpack
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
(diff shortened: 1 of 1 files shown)


.github/workflows/webpack.yml
```diff
@@ -0,0 +1,28 @@
...
+  pull_request:
+    branches: [ "main" ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [14.x, 16.x, 18.x]
```
jacoblee93, line 15: We only support Node 18 and up

jacoblee93:
Thanks for this!

If I understand correctly, this action will bundle LangChain.js using Webpack when run? To test that it works?

PR #4984, opened 2024-04-05
community[minor]: VectorStore integration for SAP HANA Cloud Vector Engine

cinqisap:
- **Description:**
This PR adds a VectorStore integration for SAP HANA Cloud Vector Engine, which is an upcoming feature in the SAP HANA Cloud database (https://blogs.sap.com/2023/11/02/sap-hana-clouds-vector-engine-announcement/). 
This typescript version is as close as Python version integration (https://github.com/langchain-ai/langchain/pull/16514)

- **Issue:** N/A
- **Dependencies:** [SAP HANA Node.js Client](https://www.npmjs.com/package/@sap/hana-client)
- **Twitter handle:** @sapopensource

Implementation of the integration:
`libs/langchain-community/src/vectorstores/hanavector.ts`

Unit tests: 
`libs/langchain-community/src/vectorstores/tests/hanavector.test.ts`

Integration tests: 
`libs/langchain-community/src/vectorstores/tests/hanavector.int.test.ts`

Examples: 
`examples/src/indexes/vector_stores/hana_vector/`

Documentation:
`docs/core_docs/docs/integrations/vectorstores/hanavector.mdx`

Integration tests and examples were run successfully.
Access credentials for the execution of the integration tests can be provided to the maintainers.


jacoblee93:
Hi @cinqisap, thanks so much for this!

The SAP client dep is quite large (186MB!) and I'm worried it'll slow down our CI. I see that it's only being used as a type in the connector itself - is there any lighter variant?

jacoblee93:
Yeah, it adds a full minute to our `yarn install`.

If there is no lighter version, what we can do is split it out into a partner package that is versioned separately, and skip building it in CI.

The other option I see is to inline the `examples/` in the docs, comment out the test cases, and define some other interface for `hanaClient.Connection` or even type it as `any` so that the client isn't actually a direct dep.

What do you prefer?

cinqisap:
hi, @jacoblee93, thanks for the suggestion!

I pushed a new commit to address this issue. My solution is to add support for another HANA Node.js driver [`hdb`](https://www.npmjs.com/package/hdb) which is only about 1MB and thus remove the large `@sap/hana-client` as a direct dev package.  Since the `hdb` did not originally support typescript, I added another declaration file to avoid errors during building time.

Now the client type is defined as any and can represent both drivers.

Looking forward to hearing more if the current solution is not as desired.

jacoblee93:
Yes that's a great solution! Thank you, I'll look shortly.

examples/src/indexes/vector_stores/hana_vector/hdb.d.ts
```diff
@@ -0,0 +1 @@
+declare module "hdb";
```
jacoblee93, line 1: Fine to add this as a dep to `examples/` - will do that shortly.

jacoblee93:
Ah, sorry missed the part about the declaration. Will fix.

jacoblee93:
Thank you!

PR #4989, opened 2024-04-05
core[patch]: Fix handling custom run id in sequence streams

bracesproul:
When streaming a runnable sequence before, custom runId's were not properly handled and tracing would lose some items inside the sequence due to duplicate runIds

(also includes some missing formatting)


langchain-core/src/runnables/base.ts
```diff
@@ -1587,14 +1587,15 @@ export class RunnableSequence<
...
     const runManager = await callbackManager_?.handleChainStart(
       this.toJSON(),
       _coerceToDict(input, "input"),
+      runId,
       undefined,
       undefined,
       undefined,
-      undefined,
-      options?.runName
+      otherOptions?.runName
```
bracesproul, line 1598: cc @jacoblee93 for handling without `no-param-reassign`

PR #4996, opened 2024-04-07
docs: ✏️ should pass string into a AIMessage / BaseMessage class

c6z3h:
Was trying to set up this QuickStart chatbot but faced some lint error. The return type of chain.invoke is a Promise<string>, but the input of demoEphemeralChatMessageHistory.addMessage() is of type BaseMessage. Thus we need to pass it into the constructor / class of AIMessage, appropriate in this context.

Also I'm getting a weird response from the bot. If i console.log() the response from the chat.invoke() statement, I get a back-and-forth conversation instead of a single AIMessage(). Is this expected behaviour?

Fixes # (issue)

### DIFF
--- a/docs/core_docs/docs/use_cases/chatbots/quickstart.mdx
+++ b/docs/core_docs/docs/use_cases/chatbots/quickstart.mdx
@@ -183,7 +183,7 @@ AIMessage {
 ```
 
 ```ts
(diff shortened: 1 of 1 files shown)


jacoblee93:
Thanks for the PR!

That chain looks like it's awaited and returns an `AIMessage`, so I believe it's correct?

And no, there should be only one message returned.

PR #5011, opened 2024-04-08
fix: correct langchain dependency for criteria evaluation

rgomezcasas:


jacoblee93:
Hey! Thanks for the PR.

The `@langchain/anthropic` package is used below, and is correct here. Many of the docs currently implicitly assume you've installed `langchain`, do you think that to not be the case?

https://js.langchain.com/docs/guides/evaluation/string/criteria#configuring-the-llm

rgomezcasas:
Ok, perfect then! Thanks!

PR #5014, opened 2024-04-08
feat(community): pgvector store: remove dependency with uuid-ossp extension

jouir:
Since PostgreSQL 13, the gen_random_uuid() function has reached the core. The uuid-ossp extension, providing the uuid_generate_v4() function, can be removed.

This is a **breaking change for PostgreSQL 12** which will be maintained until November 14, 2024. An improvement would be to issue `show server_version_num;` query and use gen_random_uuid() if the result is higher or equal than 130000 or uuid_generate_v4(). Unfortunately, my Javascript skills are pretty limited so help would be appreciated.

References:
* https://www.postgresql.org/docs/current/functions-uuid.html
* https://www.postgresql.org/support/versioning/


jouir:
Hello @jacoblee93,

There is no performance impact if the extension is used instead of the core function. The PR intent is to remove dependency with an extension and only rely on the core features of PostgreSQL. But using the extension until PostgreSQL 12 is not supported is fine.

jacoblee93:
Yes, let's leave it for now then. Will revisit when EOL

benjamincburns:
Thanks for the contribution @jouir!

PR #5016, opened 2024-04-08
openai[patch]: `ChatOpenAI.batch` function

davidfant:
This PR groups together API calls for prompts that are the same so that:
1. Less tokens are used
2. LangSmith shows the batch nicely as one run rather than `n` runs


jacoblee93:
Thanks for looking into this!

I think we would need to add this to the `OpenAICallOptions` type as well - I think I'm broadly ok with that, but want to CC @baskaryan for standardization.

Also - is it necessary? You are calling `.generate()` with this I assume? Would using `.batch()` fit your use-case instead?

davidfant:
@jacoblee93 I updated the PR to add `ChatOpenAI.batch` instead. Lmk what you think

libs/langchain-openai/src/chat_models.ts
```diff
@@ -627,6 +629,29 @@ export class ChatOpenAI<
...
+  ): Promise<BaseMessageChunk[]> {
+    const promptValues = inputs.map((i) =>
+      BaseChatModel._convertInputToPromptValue(i)
+    );
+
+    const promptValueStrings = promptValues.map((p) => p.toString());
+    if (promptValueStrings.every((p) => p === promptValueStrings[0])) {
+      const result = await this.generatePrompt(
+        [promptValues[0]],
+        { ...options, n: inputs.length } as CallOptions,
```
jacoblee93, line 644: We should probably upper bound this - I can handle it!

jacoblee93: Does this have the same output as just sending `n` requests? Or will it pick the top `n` candidates?

davidfant: > Does this have the same output as just sending `n` requests? Or will it pick the top `n` candidates?

Yes, this makes OpenAI create `n` independent results for the same prompt
https://platform.openai.com/docs/api-reference/chat/create

> Hey so chatted with the Python folks - this would change the tracing behavior for folks, and they have some concerns about overall behavior changing since it's a black box on OpenAI's end.
> 
> Could we table it for now? Sorry for the thrash - you can always wrap a `.generate()` call in a lambda.

Ok. IMO the tracing behavior is changed for the better, at least in terms of how this is visualized in LS. The goal with this was to avoid having lots of different logic for how to do requests depending on what model provider is used, but will just apply a patch locally for now.

functorism: OpenAI supporting `n` completions is a very high value feature, because of the fact that input tokens are priced only once. If you make `n` separate requests you eat the input token costs `n` times. This is an amazing aspect of the OpenAI pricing model, which many other providers don't support (for example Anthropic). I believe making it easy for users to benefit from this, even if they don't know about it is a great value add LangChain can provide.

OpenAI supports the `best_of` option, which has interplay with `n`.

> Generates best_of completions server-side and returns the "best" (the one with the highest log probability per token). Results cannot be streamed.

Users can also do this themselves now that chat completions return `logprobs`. It's a common pattern in my workflows to increase temperature for higher generation variance and utilizing the `logprobs` or simply doing self-consistency voting (https://arxiv.org/abs/2203.11171). The OpenAI pricing model has great synergy with these techniques, since you only pay extra for your generations.

I would almost argue that this feature of the API enables quality improving techniques where they would otherwise be cost prohibitive, and think leaning in and making these as easy to use as possible is of immense value.

jacoblee93: I'll figure it out on our end and get this merged. Thanks for weighing in!

bracesproul:
Going to close this PR due to merge conflicts. Please re-open off of `main` if you'd like to get this merged in! Thank you.

PR #5025, opened 2024-04-09
community[patch]: update  libs/langchain-community/src/vectorstores/analyticdb.ts 

lz82:
Fixes # (issue)
When the pageContent is empty in certain scenarios (such as when using unstructuredIo), an error occurs during embedding.


jacoblee93:
Thanks for this!

Is the issue empty string or `undefined`/`null`? Fine with defensive programming here but would it make sense to apply a fix to Unstructured too?

lz82:
> Thanks for this!
> 
> Is the issue empty string or `undefined`/`null`? Fine with defensive programming here but would it make sense to apply a fix to Unstructured too?

I've encountered a scenario where loading PDF files using https://unstructured.io/ results in some content being parsed as undefined. While it's possible to implement a fix before writing to the ADB, I believe that the purpose of langchain is to facilitate easier integration between upstream and downstream processes.

Wouldn't it be more efficient for the langchain to handle these cases of undefined values, ensuring a smoother data flow and reducing the need for additional checks and fixes on the client side?

jacoblee93:
Yeah we can, I'm thinking we should add this to the unstructured loader too since this won't be the only integration that will deal with it.

Fine for now!

jacoblee93:
Thank you!

PR #5028, opened 2024-04-09
Allow multiple metadata keys on RedisVectorStoreFilterType #5015

### LINKED ISSUE #5015: Allow multiple metadata keys on RedisVectorStoreFilterType
### Checked other resources

### Example Code

```typescript

// Add a sample document with some metadata fields:
vectorStore.addDocument(new Document({ pageContent: "Page Content!", metadata: { x: "y", z: 1, w: ["a", "b", "c"] } }))

// Narrow results with one metadata field:
vectorStore.similaritySearchWithScore("test string", 5, { x:  "y" }))

// Narrow results with two metadata fields (AND):
vectorStore.similaritySearchWithScore("test string", 5, { x:  "y", z: 1 }))

// Narrow results which contains an element on a metadata field list:
vectorStore.similaritySearchWithScore("test string", 5, { w: "a" }))

```

_No response_

### Description

This is an issue created from the following discussion:

[How to use RedisVectorStoreFilterType?](https://github.com/langchain-ai/langchainjs/discussions/5010)

Current implementation of ```RedisVectorStore``` saves the ```Document```'s metadata object as a JSON string in the metadata-key field.

This makes it impossible to search for multiple fields.

In order to make the ```RedisVectorStore``` more compatible to the ```Document``` metadata object, we need to change how metadata fields are stored and filtered.

A use case for this is a vector store of books, which metadata with fields such as author (string), year (number) and tags (list of strings) would narrow the vector search and improve its results.

We can achieve this with ```MemoryVectorStore``` filter function, and I am pretty sure we can do the same with Redis.

I can work on a PR for this if this interests the community.

Thank you.

### System Info

langchain@0.1.31 | MIT | deps: 17 | versions: 262

mauriciocirelli:
Fixes #5015

### DIFF
--- a/libs/langchain-redis/src/tests/vectorstores.int.test.ts
+++ b/libs/langchain-redis/src/tests/vectorstores.int.test.ts
@@ -2,6 +2,7 @@
 /* eslint-disable no-promise-executor-return */
 
 import { RedisClientType, createClient } from "redis";
+import { SchemaFieldTypes } from "redis";
 import { v4 as uuidv4 } from "uuid";
 import { test, expect } from "@jest/globals";
 import { faker } from "@faker-js/faker";
@@ -21,6 +22,9 @@ describe("RedisVectorStore", () => {
       redisClient: client as RedisClientType,
       indexName: "test-index",
       keyPrefix: "test:",
+      metadataSchema: {
+        ["foo"]: SchemaFieldTypes.TEXT,
+      }
     });
   });
 
@@ -66,9 +70,7 @@ describe("RedisVectorStore", () => {
     ]);
 
     // If the filter wasn't working, we'd get all 3 documents back
-    const results = await vectorStore.similaritySearch(pageContent, 3, [
-      `${uuid}`,
-    ]);
+    const results = await vectorStore.similaritySearch(pageContent, 3, `@foo:(${uuid})`);
 
     expect(results).toEqual([
       new Document({ metadata: { foo: uuid }, pageContent }),
--- a/libs/langchain-redis/src/tests/vectorstores.test.ts
+++ b/libs/langchain-redis/src/tests/vectorstores.test.ts
@@ -3,6 +3,7 @@ import { jest, test, expect, describe } from "@jest/globals";
 import { FakeEmbeddings } from "@langchain/core/utils/testing";
 
 import { RedisVectorStore } from "../vectorstores.js";
+import { SchemaFieldTypes } from "redis";
 
 const createRedisClientMockup = () => {
   const hSetMock = jest.fn();
@@ -34,6 +35,9 @@ test("RedisVectorStore with external keys", async () => {
   const store = new RedisVectorStore(embeddings, {
     redisClient: client as any,
     indexName: "documents",
+    metadataSchema: {
+      ["a"]: SchemaFieldTypes.NUMERIC,
+    }
   });
 
   expect(store).toBeDefined();
@@ -44,7 +48,6 @@ test("RedisVectorStore with external keys", async () => {
         pageContent: "hello",
         metadata: {
           a: 1,
-          b: { nested: [1, { a: 4 }] },
         },
       },
     ],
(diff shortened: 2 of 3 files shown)


jacoblee93:
Thanks for the PR!

I think this could be nice but can you please look into making it backwards compatible?

mauriciocirelli:
Hi @jacoblee93,

Thank you for your time looking into this.
I certainly can make it more backwards compatible. I just would like to clarify.

There are two 'incompatibilities' in my original PR:

1. Dropping of metadataKey parameter
I can have it back and it would add a JSON object as string into document, keeping its original functionality. Not sure if that is desirable, because we would have duplicated metadata stuff.. a 'metadataKey' field containing a JSON string and many metadata fields, according to the schema.

Originally:
```typescript
metadata = { a:"A", b: "B", c:"C" };

redis: JSON.stringify(metadata);
```
Now:
```typescript
metadata = { a:"A", b: "B", c:"C" };

redis: { metadata: JSON.stringify(metadata), a:"A", b: "B", c:"C" };
```

2. Filter is not accepting an array of strings:
We can accept an array of filters and OR them as originally coded. The behavior would be different, though, because we would be ORing Redis filter strings in this array instead of possible values for the 'metadataKey' field.

Originally:
```typescript
filter = [ "A", "B", "C"];
filterString = '@metadata{A|B|C}'
```
Now:
```typescript
filter = [ "@a:{A}", "b:{B}", "@c:{C}"];
filterString = '@a:{A} | b:{B} | @c:{C}'
```
Not that the old filter array would result in a filter string that does not make sense anymore:
```typescript
filter = [ "a", "b", "c"];
filterString = 'a|b|c'
```

This is why I made it with breaking changes... At least the developer would know that things are different now and would have to change the code to get the correct behavior. But we can keep things backwards compatible in the interfaces, but with a very different behavior at runtime.

Please, let me know your thoughs and how can we make it work.

Thank you.

mauriciocirelli:
Also, I have looked into other VectorStores implementations, such as [PgVector](https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-community/src/vectorstores/pgvector.ts)

It is also storing the metadata as a JSON object in a single column (metadataColumnName) in the database.

It would produce similar limitations, as we would not be able to filter by metadata fields in a reasonable way...

bsbodden:
@mauriciocirelli this is a nice PR, with the addition of the RediSearch schema definition you get the flexibility of the full power of the search engine, but that means (as you pointed out above https://github.com/langchain-ai/langchainjs/pull/5028#issuecomment-2051531154 ) that `filterString` assembly gets a little more complex to handle all the possible types you can declare as part of the search index schema (for the metadata). At a minimum you would have to handle the difference in syntax between `TAG` fields and `TEXT` fields. See this project for an idea of how to assemble the expressions https://github.com/redis/redis-om-node/tree/main/lib/search and handle ANDs and ORs.

For the backwards compatibility, off the top of my head, if the 'metadataKey' is present use that as the backwards compatibility trigger and do as it was done before. If the `metadataSchema` is set and they pass string just assume is a RediSearch expression and pass it along (this might be the easiest approach) and puts the complexity on the query /expression building and not in the framework having to spring a query builder. My 2 cents!

mauriciocirelli:
Hi @bsbodden 

Regarding the backwards compatibility, I think that's easy and reasonable.
Regarding the filter expression, I think it should be just a string in case a ```metadataSchema``` is provided.
Building the expression from an object and its schema is not trivial e and I think it is way beyond the scope of langchain. The developer should know how to build a Redis query (or use an external lib such as redis-on) as much as they have to know how to build Postgres queries (or use external ORM libs) or any other persistent storage queries they want to integrate to langchain.

That's just my opinion. If you are all okay with that, I can update the PR with these changes.

christian-bromann:
Hey @mauriciocirelli , sorry for the (very) delayed response. Unfortunately it seems like our team was able to pick this up. I apologize. If you could re-raise the PR I am happy to review it with you and get it landed. Thank you!.

PR #5036, opened 2024-04-09
core[minor],google[minor]: Add support for images

bracesproul:
wip
todo:
- [ ] tests
- [ ] docs


libs/langchain-google-gauth/src/tests/data/video.mp4
bracesproul, line 1: todo for brace: drop before merging

afirstenberg:
Oh please, I'm begging us... can we NOT add `audio_url` but just support a general URL?

PR #5043, opened 2024-04-10
core[minor],google-common[minor]: Add support for generic objects in prompts, gemini audio/video docs

bracesproul:


libs/langchain-google-common/src/utils/gemini.ts
```diff
@@ -101,6 +136,11 @@ export function messageContentToParts(content: MessageContent): GeminiPart[] {
             return messageContentImageUrl(content as MessageContentImageUrl);
           }
           break;
+        case "audio":
```
afirstenberg, line 139: Is the plan to change this to a generic "media" or "object" or "blob"? (And have the other methods named similarly)
This way we can support audio, video, and images at once.

libs/langchain-google-common/src/utils/gemini.ts
```diff
@@ -74,6 +83,29 @@ function messageContentImageUrl(
...
+  } else if ("mimeType" in content && "fileUri" in content) {
+    return {
+      fileData: {
+        mimeType: content.mimeType,
+        fileUri: content.fileUri,
+      },
+    };
+  }
+
+  throw new Error("Invalid audio content");
```
afirstenberg, line 106: nit: "Invalid media content"

libs/langchain-google-common/src/utils/gemini.ts
```diff
@@ -74,6 +83,29 @@ function messageContentImageUrl(
...
+  content: Record<string, any>
+): GeminiPartInlineData | GeminiPartFileData {
+  if ("mimeType" in content && "data" in content) {
+    return {
+      inlineData: {
+        mimeType: content.mimeType,
+        data: content.data,
+      },
+    };
+  } else if ("mimeType" in content && "fileUri" in content) {
```
afirstenberg, line 97: Pondering out loud:
- perhaps instead of "fileUri" we just make it "uri" or "url"
- then we just use `extractMimeType` method as above to see if it's a data: url (and "mimeType" is optional or ignored) and use `inlineData`
- If it isn't a data url, and we have the mime type, then we pass the url and mime types given to us using `fileData`

This would then turn the `messageContentImageUrl` function into calling this function with the attributes of `content` set.

(And it also means that if we add more sophisticated file handling later, we only have to change it in one place.)

docs/core_docs/docs/use_cases/media.mdx
```diff
@@ -0,0 +1,56 @@
+# Audio/Video Structured Extraction
+
+Google's Gemini API offers support for audio and video input, along with function calling.
+Together, we can pair these API features to extract structured data given audio or video input.
+
+In the following examples, we'll demonstrate how to read and send MP3 and MP4 files to the Gemini API, and receive structured output as a response.
```
afirstenberg, line 6: Love these examples! Perhaps comment that you don't *need* to use structured output with audio and video, but it always helps to understand what the results can be.

libs/langchain-google-common/src/utils/gemini.ts
```diff
@@ -101,6 +133,8 @@ export function messageContentToParts(content: MessageContent): GeminiPart[] {
             return messageContentImageUrl(content as MessageContentImageUrl);
           }
           break;
+        case "media":
```
afirstenberg, line 136: I wish I had seen https://github.com/langchain-ai/langchainjs/blame/fc2f9de2910a6728cf9c24f9146b55ba48d3790f/langchain-core/src/messages/index.ts#L56C69-L56C69 when it went in! 

I'm honest, I'm a little anxious about defining MessageContent types with magic string values and not real types. Even if they're fundamentally Record types. Makes it a lot more difficult for other implementations to use consistent naming.

libs/langchain-google-common/src/utils/gemini.ts
```diff
@@ -74,6 +83,29 @@ function messageContentImageUrl(
   }
 }
 
+function messageContentToMedia(
```
afirstenberg, line 86: Minor nit: Shouldn't this be `messageContentMedia` without the "to"? It is a "media" MessageContent type - we're not converting it to media.

afirstenberg:
Awesome getting this done so fast!

I don't see any fundamental issues. A couple of nits. A couple of other small suggestions.

My only major suggestion would be the changes that make the uri/url a little more flexible and have `messageContentImageUrl` call `messageContentMedia`. This lets us handle these media types the same way in the future.

libs/langchain-google-common/src/utils/gemini.ts
bracesproul, line 106:
```suggestion
  throw new Error("Invalid media content");
```

libs/langchain-google-common/src/utils/gemini.ts
bracesproul, line 86:
```suggestion
function messageContentMedia(
```

libs/langchain-google-common/src/utils/gemini.ts
```diff
@@ -101,6 +133,8 @@ export function messageContentToParts(content: MessageContent): GeminiPart[] {
             return messageContentImageUrl(content as MessageContentImageUrl);
           }
           break;
+        case "media":
+          return messageContentToMedia(content);
```
bracesproul, line 137:
```suggestion
          return messageContentMedia(content);
```

examples/lance_ls_eval_video.mp4
jacoblee93, line 1: Can we remove? Slows down git

examples/Mozart_Requiem_D_minor.mp3
jacoblee93, line 1: Remove

PR #5045, opened 2024-04-11
langchain[patch]: fix: make GithubRepoLoader robust against .gitmodules files w/o newline

the-powerpointer:
Occasionally .gitmodules files are not ending with a newline character. Still the file is valid.
This PR fixes the issue with the given regex by adding the newline character for parsing, if not present.


jacoblee93:
Thank you!

PR #5063, opened 2024-04-11
experimental[patch]: Fixed : order to descending to ensure matching of run ID with OpenAI'…

### LINKED ISSUE #5062: Incorrect order of message retrieval causing mismatch with OpenAI run ID
### Checked other resources

### Example Code

 Repeat the below code until there are 20 messages in a  single thread , after 20 messages . It will return as null value , since the run id doesnt match
 
  ```javascript
const {OpenAIAssistantRunnable} = require("langchain/experimental/openai_assistant");

const agent = new OpenAIAssistantRunnable({
      clientOptions: { apiKey: OPENAI_API_KEY },
      assistantId: assistant_id,
      asAgent: true,
    });

    let response;
    const thread_id: "Your thread_id here"

    if (thread_id) {
      response = await agent.invoke({
        content: "Your content here",
        threadId: thread_id,
      });
    } else {
      response = await agent.invoke({ content: "Your content here" });
    }
  ```
  

_No response_

### Description

I m trying to retrieve response from openAI assistant runnable , however when i exceed 20 messages in my thread. It returns in null .

### System Info

platform : windows
node : v20.10.0

Ghat0tkach:
changed the order to descending to ensure it retrieves the most recent 20 messages of the thread, ensuring that the run ID matches the OpenAI run's ID. If not , after 20 messages in a single thread. It will return null .

It returns empy / null value after 20 messages in a thread . its due to the fact thats its sorting by ascending order and as a result it checks for first 20 messages . Not the latest one .

Hence , the run id doesnt match with openAI assistant's run id and it returns null

Fixes #5062 (issue)

### DIFF
--- a/langchain/src/experimental/openai_assistant/index.ts
+++ b/langchain/src/experimental/openai_assistant/index.ts
@@ -291,7 +291,7 @@ export class OpenAIAssistantRunnable<
     const run = await this._waitForRun(runId, threadId);
     if (run.status === "completed") {
       const messages = await this.client.beta.threads.messages.list(threadId, {
-        order: "asc",
+        order: "desc",
       });
       const newMessages = messages.data.filter((msg) => msg.run_id === runId);
       if (!this.asAgent) {


jacoblee93:
Thank you!

PR #5080, opened 2024-04-12
community[patch]: deps: bump astra-client to ^1.0.0

nicoloboschi:


jacoblee93:
Seems good, looks like no external API changes?

jacoblee93:
Ah never mind, looks like this is breaking. I think probably worth breaking out into a partner package like in Python instead?

PR #5084, opened 2024-04-12
docs[minor],core[minor]: Add support for mustache templates

bracesproul:
Todo:
- [x] docs


langchain-core/src/prompts/prompt.ts
```diff
@@ -116,9 +116,19 @@ export class PromptTemplate<
 
   constructor(input: PromptTemplateInput<RunInput, PartialVariableName>) {
     super(input);
+    // If input is mustache and validateTemplate is not defined, set it to false
+    if (
+      input.templateFormat === "mustache" &&
+      input.validateTemplate === undefined
+    ) {
+      this.validateTemplate = false;
+    }
```
bracesproul, line 125: cc @nfcampos thoughts on this?

bracesproul: If I don't do this, you'll always need to explicitly `validateTemplate: false` in prompt options when using `mustache`

langchain-core/src/prompts/tests/structured.test.ts
```diff
@@ -90,7 +90,7 @@ test("Test format", async () => {
...
+    "{\"lc\":1,\"type\":\"constructor\",\"id\":[\"langchain_core\",\"prompts\",\"structured\",\"StructuredPrompt\"],\"kwargs\":{\"schema_\":{\"name\":\"yo\",\"description\":\"a structured output\",\"parameters\":{\"name\":{\"type\":\"string\"},\"value\":{\"type\":\"integer\"}}},\"input_variables\":[],\"messages\":[{\"lc\":1,\"type\":\"constructor\",\"id\":[\"langchain_core\",\"prompts\",\"chat\",\"HumanMessagePromptTemplate\"],\"kwargs\":{\"prompt\":{\"lc\":1,\"type\":\"constructor\",\"id\":[\"langchain_core\",\"prompts\",\"prompt\",\"PromptTemplate\"],\"kwargs\":{\"input_variables\":[],\"template_format\":\"f-string\",\"template\":\"I'm very structured, how about you?\",\"schema\":{\"name\":\"yo\",\"description\":\"a structured output\",\"parameters\":{\"name\":{\"type\":\"string\"},\"value\":{\"type\":\"integer\"}}}}}}}]}}"
```
bracesproul, line 93: cc @nfcampos new string also includes the `schema` variable.

langchain-core/src/prompts/prompt.ts
andrewnguonly: Seems reasonable. Maybe add log a warning/info message? It's helpful to know when values are changing internally.

andrewnguonly (approved):
Nice job with the docs!

I can definitely see this how the double bracket `{{` (mustache) vs single bracket `{` (f-string) difference might cause some confusion if someone is migrating or just forgets to add an extra bracket. Like a missed semi-colon, they might be debugging for a while until they realize they just missed the extra bracket. Maybe emphasize this in the docs? Over the top: Log warning if no templated values are found?

langchain-core/src/prompts/chat.ts
```diff
@@ -793,8 +825,17 @@ export class ChatPromptTemplate<
 
   validateTemplate = true;
 
+  templateFormat: TemplateFormat = "f-string";
+
   constructor(input: ChatPromptTemplateInput<RunInput, PartialVariableName>) {
     super(input);
+    // If input is mustache and validateTemplate is not defined, set it to false
```
jacoblee93, line 832: Why?

jacoblee93:
Seems fine overall, but how does this affect automatic typing? We use the `type ParamsFromFString` in `PromptTemplate.fromTemplate` args

langchain-core/src/prompts/tests/structured.test.ts
```diff
@@ -90,7 +90,7 @@ test("Test format", async () => {
...
+    '{"lc":1,"type":"constructor","id":["langchain_core","prompts","structured","StructuredPrompt"],"kwargs":{"schema_":{"name":"yo","description":"a structured output","parameters":{"name":{"type":"string"},"value":{"type":"integer"}}},"input_variables":[],"messages":[{"lc":1,"type":"constructor","id":["langchain_core","prompts","chat","HumanMessagePromptTemplate"],"kwargs":{"prompt":{"lc":1,"type":"constructor","id":["langchain_core","prompts","prompt","PromptTemplate"],"kwargs":{"input_variables":[],"template_format":"f-string","template":"I\'m very structured, how about you?","schema":{"name":"yo","description":"a structured output","parameters":{"name":{"type":"string"},"value":{"type":"integer"}}}}}}}]}}'
```
jacoblee93, line 93: Why did this change?

langchain-core/src/prompts/chat.ts
bracesproul: Since `this.validateTemplate` defaults to `true`, if `undefined` is passed it'll be assigned `true`.

langchain-core/src/prompts/tests/structured.test.ts
bracesproul: I believe it's because I started passing optional options through all the prompts, so that `templateFormat` could be passed through. And since I'm now always passing these options through where they weren't always being passed through, schema gets caught in this and is why it's logged. Not entirely sure though.

Is this related to serialization/is breaking?

jacoblee93: Probably fine if it's additive, but worth checking in with @dqbd or @nfcampos

nfcampos: I think it's more that serialization was broken before and this change fixed it!

PR #5087, opened 2024-04-13
Feat/openai n

jacoblee93:
@davidfant @functorism this is roughly what it would take to have the same featureset as `.batch` (separate runs per input, with error handling).

I think the maintenance overhead would be tough - would you all feel comfortable calling `.generate()` directly instead?


davidfant:
@jacoblee93 hmm, agree that this doesn't look clean on the `ChatOpenAI`-side. is there any other good way to accomplish this kind of batching without hacking the batch fn and without using generate? the problem with generate is that I want to use the same interface for across my app when doing LLM calls with an arbitrary `BaseChatModel` (mostly Claude and GPT-4). i want to avoid `if OpenAI, generate in a special way`

jacoblee93:
Yeah I gotcha - unfortunately I think nothing comes to mind CC @baskaryan @eyurtsev  :(

I think for now you could wrap it in a custom function? I hear you on wanting a unified interface for sure though.

PR #5088, opened 2024-04-13
Adds default maxRetries to 6 as per documentation

### LINKED ISSUE #5061: `maxRetries` not properly passed to `ChatOpenAI` in tracing
### Checked other resources

### Example Code

I have this code for creating a chain with ChatOpenAI client:
```typescript
const chain  = RunnableSequence.from([
            chatPrompt, // prompt defined above
            new ChatOpenAI({
              modelName: 'gpt-4-0125-preview',
              temperature: 0.1,
              maxRetries: 3,
              modelKwargs: { response_format: { type: 'json_object' } },
            }),
            suggestionsParser, // some parser defined above
          ]);
```
I invoke the chain like this:
```typescript
const response = await chain.invoke({ document_to_review: docContent }, { tags: ['suggestions'] });
```

_No response_

### Description

The openai call returned an error but no retries were made.
OpenAI error:
```
400 You requested a model that is not compatible with this engine. Please contact us through our help center at help.openai.com for further questions.

Error: 400 You requested a model that is not compatible with this engine. Please contact us through our help center at help.openai.com for further questions.
    at APIError.generate (/app/node_modules/.pnpm/openai@4.32.2/node_modules/openai/error.js:44:20)
    at OpenAI.makeStatusError (/app/node_modules/.pnpm/openai@4.32.2/node_modules/openai/core.js:263:33)
    at OpenAI.makeRequest (/app/node_modules/.pnpm/openai@4.32.2/node_modules/openai/core.js:306:30)
    at async /app/node_modules/.pnpm/@langchain+openai@0.0.26/node_modules/@langchain/openai/dist/chat_models.cjs:684:29
    at async RetryOperation._fn (/app/node_modules/.pnpm/p-retry@4.6.2/node_modules/p-retry/index.js:50:12)
```

The langsmith logs show that retries parameter is present in the metadata passed
```
{
  "id": [
    "langchain",
    "chat_models",
    "openai",
    "ChatOpenAI"
  ],
  "kwargs": {
    "max_retries": 3,
    "model": "gpt-4-0125-preview",
    "model_kwargs": {
      "response_format": {
        "type": "json_object"
      }
    },
    "openai_api_key": {
      "id": [
        "OPENAI_API_KEY"
      ],
      "lc": 1,
      "type": "secret"
    },
    "temperature": 0.1
  },
  "lc": 1,
  "type": "constructor"
}
```
But when looking at invocation metadata in LangSmith, the retry parameter isn't present:
image

How to properly pass `maxRetries` parameter to ChatOpenAI?
I want to specify a specific number, but does not specifying it at all **really** default to a 6 time retry with exponential backoff?

### System Info

Running on an ec2 instance with linux

Langchain versions:
- "@langchain/core": "0.1.53",
- "@langchain/openai": "^0.0.26",
- "langchain": "^0.1.31",

rahilvora:
Fixes #5061

This PR adds default value for `maxRetries` to `6` as per this [documentation](https://js.langchain.com/docs/modules/model_io/chat/dealing_with_api_errors).

### DIFF
--- a/libs/langchain-openai/src/chat_models.ts
+++ b/libs/langchain-openai/src/chat_models.ts
@@ -437,14 +437,9 @@ export class ChatOpenAI<
       fields?.configuration?.organization ??
       getEnvironmentVariable("OPENAI_ORGANIZATION");
 
-<<<<<<< Updated upstream
+    this.maxRetries = fields?.configuration?.maxRetries ?? 6;
     this.modelName = fields?.model ?? fields?.modelName ?? this.model;
     this.model = this.modelName;
-=======
-    // OpenAi ClientOptions interface defaults to 2
-    this.maxRetries = fields?.configuration?.maxRetries ?? 6;
-    this.modelName = fields?.modelName ?? this.modelName;
->>>>>>> Stashed changes
     this.modelKwargs = fields?.modelKwargs ?? {};
     this.timeout = fields?.timeout;
 
@@ -921,7 +916,8 @@ export class ChatOpenAI<
       const params = {
         ...this.clientConfig,
         baseURL: endpoint,
-        timeout: this.timeout
+        timeout: this.timeout,
+        maxRetries: 0,// 0 to avoid using the default SDK retry policy. We retry using the more generalized AsyncCaller
       };
       if (!params.baseURL) {
         delete params.baseURL;
--- a/libs/langchain-openai/src/chat_models.ts
+++ b/libs/langchain-openai/src/chat_models.ts
@@ -366,6 +366,8 @@ export class ChatOpenAI<
 
   maxTokens?: number;
 
+  maxRetries?: number;
+
   logprobs?: boolean;
 
   topLogprobs?: number;
@@ -435,8 +437,14 @@ export class ChatOpenAI<
       fields?.configuration?.organization ??
       getEnvironmentVariable("OPENAI_ORGANIZATION");
 
+<<<<<<< Updated upstream
     this.modelName = fields?.model ?? fields?.modelName ?? this.model;
     this.model = this.modelName;
+=======
+    // OpenAi ClientOptions interface defaults to 2
+    this.maxRetries = fields?.configuration?.maxRetries ?? 6;
+    this.modelName = fields?.modelName ?? this.modelName;
+>>>>>>> Stashed changes
     this.modelKwargs = fields?.modelKwargs ?? {};
     this.timeout = fields?.timeout;
 
@@ -480,6 +488,7 @@ export class ChatOpenAI<
         configuration?.baseOptions?.params ??
         fields?.configuration?.baseOptions?.params,
       ...configuration,
+      maxRetries: this.maxRetries,
       ...fields?.configuration,
     };
   }
@@ -912,8 +921,7 @@ export class ChatOpenAI<
       const params = {
         ...this.clientConfig,
         baseURL: endpoint,
-        timeout: this.timeout,
-        maxRetries: 0,
+        timeout: this.timeout
       };
       if (!params.baseURL) {
         delete params.baseURL;
--- a/libs/langchain-openai/src/tests/chat_models.int.test.ts
+++ b/libs/langchain-openai/src/tests/chat_models.int.test.ts
@@ -231,6 +231,15 @@ test("Test OpenAI with timeout in call options and node adapter", async () => {
   ).rejects.toThrow();
 }, 5000);
 
+test("Test OpenAI with maxRetries in call options and node adapter", async () => {
+  const model = new ChatOpenAI({ maxTokens: 5, maxRetries: 2 });
+  await expect(() =>
+    model.invoke([new HumanMessage("Print hello world")], {
+      options: { timeout: 10 },
+    })
+  ).rejects.toThrow();
+}, 5000);
+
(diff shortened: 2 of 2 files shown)


libs/langchain-openai/src/chat_models.ts
```diff
@@ -435,8 +437,14 @@ export class ChatOpenAI<
       fields?.configuration?.organization ??
       getEnvironmentVariable("OPENAI_ORGANIZATION");
 
+<<<<<<< Updated upstream
```
lukasugar, line 440: You should remove these git artifacts before merging: `<<<<<<<`, `=======`, `>>>>>>> Stashed changes`

libs/langchain-openai/src/chat_models.ts
```diff
@@ -912,8 +921,7 @@ export class ChatOpenAI<
       const params = {
         ...this.clientConfig,
         baseURL: endpoint,
-        timeout: this.timeout,
-        maxRetries: 0,
```
jacoblee93, line 916: We still need this to be 0 to avoid using the default SDK retry policy. We retry using the more generalized `AsyncCaller`

jacoblee93: Would be good to add a note here

libs/langchain-openai/src/chat_models.ts
```diff
@@ -480,6 +488,7 @@ export class ChatOpenAI<
         configuration?.baseOptions?.params ??
         fields?.configuration?.baseOptions?.params,
       ...configuration,
+      maxRetries: this.maxRetries,
```
jacoblee93, line 491: See below

libs/langchain-openai/src/chat_models.ts
```diff
@@ -435,8 +437,14 @@ export class ChatOpenAI<
       fields?.configuration?.organization ??
       getEnvironmentVariable("OPENAI_ORGANIZATION");
 
+<<<<<<< Updated upstream
     this.modelName = fields?.model ?? fields?.modelName ?? this.model;
     this.model = this.modelName;
+=======
+    // OpenAi ClientOptions interface defaults to 2
+    this.maxRetries = fields?.configuration?.maxRetries ?? 6;
```
jacoblee93, line 445: This should be 6?

https://github.com/langchain-ai/langchainjs/blob/main/langchain-core/src/utils/async_caller.ts#L93

libs/langchain-openai/src/chat_models.ts
rahilvora: I added this here incase `maxRetries` is not provided and we want to make sure we use the default `6` which we assigned above in `constructor`

libs/langchain-openai/src/chat_models.ts
rahilvora: I believe so, doc says `6` https://js.langchain.com/docs/modules/model_io/chat/dealing_with_api_errors

libs/langchain-openai/src/chat_models.ts
rahilvora: Done

libs/langchain-openai/src/chat_models.ts
rahilvora: My bad, forgot to check this after rebase. Fixed

libs/langchain-openai/src/chat_models.ts
jacoblee93: `this.clientConfig` gets passed to the `OpenAI` client, which we don't want.

jacoblee93:
Can you add a test confirming that this is:

1. Actually happening for retryable errors
2. Actually fixing the issue?

libs/langchain-openai/src/chat_models.ts
rahilvora: Quick Question,

`_getClientOptions` function is called from  `completionWithRetry` method. 

`_getClientOptions` checks for `this.client` and if not present (which I think won't be present because `this.client` is not instantiated anywhere in the file). In that case,

```
const params = {
  baseURL: endpoint,
  timeout: this.timeout,
  maxRetries: 0,// 0 to avoid using the default SDK retry policy. We retry using the more generalized AsyncCaller
};
      
if (!params.baseURL) {
  delete params.baseURL;
}

this.client = new OpenAIClient(params);
```

In this case, `maxRetries` is set to `0` instead of value passed by the caller. Do we want it to be `0` or it should be the value/option passed by the caller?

rahilvora:
Sure @jacoblee93 

Can you point me to the file in which I can add tests for making sure `maxRetries` are part of Langsmit tracing?

I am not able to see any tracing tests in `libs/langchain-openai/src/tests/chat_models.int.test.ts` file.

lukasugar:
@jacoblee93 why close this 😢

jacoblee93:
Because it's working as intended. 400s are not retryable

lukasugar:
@jacoblee93 for non 4XX errors, is the number of retries 6 or 0? Should be 6

jacoblee93:
Yes should be 6 times. You can customize retry behavior by passing a custom `onFailedAttempt` handler:

https://github.com/langchain-ai/langchainjs/blob/8d59ca0897b5d009f926f9a81fe3eee8e6aefce8/langchain-core/src/utils/async_caller.ts#L62

PR #5089, opened 2024-04-13
google-common[minor]: Support Gemini system messages

### LINKED ISSUE #5068: google-common [feature]: Support Gemini system messages
### Privileged issue

### Issue Content

See https://github.com/langchain-ai/langchain-google/issues/129

https://ai.google.dev/docs/system_instructions
https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/gemini#request_body

This isn't implemented for all models, so also add `convertSystemMessageToHumanContent` to be compatible with Python. Also see the warning message generated at https://github.com/langchain-ai/langchain-google/blob/62ba6ca5e39174d467e47c83119c8bb4275128f6/libs/vertexai/langchain_google_vertexai/chat_models.py#L175

afirstenberg:
Fixes #5068 

Please update version to *same* version number for
- google-common
- google-gauth
- google-webauth
- google-vertex
- google-vertex-webauth


libs/langchain-google-common/src/chat_models.ts
```diff
@@ -60,14 +61,68 @@ class ChatConnection<AuthOptions> extends AbstractGoogleLLMConnection<
...
+    return typeof this.convertSystemMessageToHumanContent === "boolean"
+      ? !this.convertSystemMessageToHumanContent
+      : this.computeUseSystemInstruction;
+  }
+
+  get computeUseSystemInstruction(): boolean {
+    // This works on models from April 2024 and later
+    //   Vertex AI: gemini-1.5-pro and gemini-1.0-002 and later
+    //   AI Studio: gemini-1.5-pro-latest
+    console.log('computeUseSystemInstruction', this.modelFamily, this.modelName, this.platform);
```
jacoblee93, line 82: Remove log

afirstenberg: Doh! Removed

libs/langchain-google-common/src/chat_models.ts
```diff
@@ -60,14 +61,67 @@ class ChatConnection<AuthOptions> extends AbstractGoogleLLMConnection<
...
+      }, []);
   }
+
+  formatSystemInstruction(
+    input: BaseMessage[],
+    _parameters: GoogleAIModelParams
+  ): GeminiContent {
+    const firstInput = input[0];
+    if (firstInput._getType() === "system" && this.useSystemInstruction) {
+      return baseMessageToContent(firstInput, undefined, true)[0];
```
jacoblee93, line 119: What if we had a system message in another part of the prompt? Google would throw an error?

libs/langchain-google-common/src/chat_models.ts
```diff
@@ -60,14 +61,67 @@ class ChatConnection<AuthOptions> extends AbstractGoogleLLMConnection<
...
     return input
-      .map((msg, i) => baseMessageToContent(msg, input[i - 1]))
-      .reduce((acc, cur) => [...acc, ...cur]);
+      .map((msg, i) =>
+        baseMessageToContent(msg, input[i - 1], this.useSystemInstruction)
+      )
+      .reduce((acc, cur) => {
+        // Filter out the system content, since those don't belong
+        // in the actual content.
+        const hasNoSystem = cur.every(content => content.role !== "system");
```
jacoblee93, line 108: Would we ever have system content mixed with other types of content?

jacoblee93:
There's a fair bit going on, but from what I gather:

- Not all Gemini models support system instructions
- The ones that do allow it don't allow it in `contents` but as a separate parameter on the request
- If a model supports system messages and receives a LangChain `SystemMessage` as the first message in its input, this PR removes that message from the input and maps that to the appropriate Google parameter.

Is that accurate ^?

libs/langchain-google-common/src/chat_models.ts
afirstenberg: Gemini puts the system prompt in another section away from the content (but in a content-like data structure).

So `formatContents()` and `formatSystemInstruction()` both take the results of `baseMessageToContent()` and have to do different things with it.

In this function (`formatContents`) it removes all the system messages, since Gemini will reject the format if its there.
In the other function referenced above (`formatSystemInstruction`) it only needs to care if the system instruction is first.

jacoblee93: Technically we're dropping data silently in this case where a system message is not first in the input? I think logging a warning would be nice or ideally throwing an error.

libs/langchain-google-common/src/chat_models.ts
afirstenberg: Google only accepts one system message (see the other notes about this).

In this code, it is just looking for it as the first message.
But I guess one could argue that any of these others are better approaches:
- Throws an error if one is found _not_ as the first message.
- Combine any of them found, in any position, into the system prompt

Thoughts?

afirstenberg:
> I am also getting 500s when I try to run the integration tests in `langchain-google-vertexai-web` with this PR

_glare_  Ok, let me double-check what is going on there.

afirstenberg:
> There's a fair bit going on, but from what I gather:

All completely correct.

The `computeUseSystemInstruction` accessor attempts to maintain backwards compatibility with the previous behavior and older versions.

libs/langchain-google-common/src/chat_models.ts
afirstenberg: Latest update handles this by throwing an error. I've also added tests for this.

afirstenberg:
> > I am also getting 500s when I try to run the integration tests in `langchain-google-vertexai-web` with this PR
> 
> _glare_ Ok, let me double-check what is going on there.

I couldn't duplicate this in my testing.
My guess is this is a transient 500 error.

jacoblee93:
Still getting some weirdness in the integration tests (it seems it's no longer returning multipart messages, but strings) but will merge and ship tonight! Thank you!

PR #5093, opened 2024-04-13
langchain[patch]: Adds metadata to vectorstore retriever memory

MJDeligan:
As per #3107 .

Adds a metadata field to VectorStoreRetreiverMemory, which automatically adds metadata to the created documents.

Includes:
- Metadata field as either a metadata object or a function returning a metadata object.
- Tests
- Extended Example with metadata usage


jacoblee93:
Looks reasonable, thank you!

PR #5097, opened 2024-04-14
langchain[patch],docs[patch]: Remove unnecessary `await` from `SelfQueryRetriever.fromLLM` calls

mlegy:
Fixes https://github.com/langchain-ai/langchainjs/discussions/5096


jacoblee93:
Thank you!

PR #5099, opened 2024-04-14
core[patch]: Update ai.ts

dwcarr:
In langchain-core/src/messages/ai.ts

The code currently only checks for undefined on rawToolCalls.  However, the provider (deep-infra in this case)  can return toolCalls=null, which passed the conditional but will then produce an error when rawToolCalls.length is accessed.

Changed the conditional to !(rawToolCalls == null) to address the issue, and verified that this fixed the issue.

Note that `rawToolCalls == null` is true if rawToolCalls is null OR undefined, thus handling both cases.

### DIFF
--- a/langchain-core/src/messages/ai.ts
+++ b/langchain-core/src/messages/ai.ts
@@ -71,7 +71,7 @@ export class AIMessage extends BaseMessage {
         );
       }
       try {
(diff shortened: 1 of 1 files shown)


jacoblee93:
Thank you!

PR #5103, opened 2024-04-15
google-common[patch]: adding client-info

tanyaasharma:
Fixes # (issue)


libs/langchain-google-common/src/connection.ts
```diff
@@ -53,11 +53,28 @@ export abstract class GoogleConnection<
 
   async _clientInfoHeaders(): Promise<Record<string, string>> {
     const clientLibraryVersion = await this._clientLibraryVersion();
+    const clientInfo= await this._getclientInfo();
     return {
       "User-Agent": clientLibraryVersion,
+      "Client-Info": clientInfo.toString(),
```
afirstenberg, line 59: I don't think we want `toString()` here. If this is a Record, then the string would be "[object Object]"

libs/langchain-google-common/src/connection.ts
```diff
@@ -53,11 +53,28 @@ export abstract class GoogleConnection<
...
     };
   }
 
+  async _getclientInfo(): Promise<Record<string, string>> {
```
afirstenberg, line 63: There is a lot of code duplication between this and `_clientLibraryVersion()`. Perhaps combine the two?

libs/langchain-google-common/src/connection.ts
```diff
@@ -53,11 +53,28 @@ export abstract class GoogleConnection<
...
+   const userAgent = await this._clientLibraryVersion();
+   const env = await getRuntimeEnvironment();
+      const langchainVersion = env?.libraryVersion ?? "0";
```
afirstenberg, line 66: Line formatting?
Did you run `yarn format`?

afirstenberg:
I don't know what the format of Client-Info is supposed to be (is it a standard HTTP header? I can't find a reference if so), but it almost certainly isn't the output `toString()` provides.

See other inline notes.

Might also be good to have a test to make sure these are in the format expect. See the test in llms.test.ts called "user agent header"

libs/langchain-google-common/src/connection.ts
```diff
@@ -53,11 +53,28 @@ export abstract class GoogleConnection<
...
+   const moduleName = await this._moduleName();
+   let clientLibraryVersion = `${langchainVersion}`;
```
jacoblee93, line 68: I unfortunately don't think this does what you think it does :( we don't currently have a way to extract version at runtime.

afirstenberg:
I think it looks better after these updates.

Remaining thoughts:
* As @jacoblee93 and I observed - `env?.libraryVersion` is undefined at the moment. So this new header may not be giving you what you think. What are you expecting it to be?
* Run the tests in "chat_models.test.ts", particularly the "user agent header" test to make sure it still passes.
* Add another test to "chat_models.test.ts" to test the expected outcomes from the Client-Info header.

jacoblee93:
Agree with @afirstenberg's comments above - the `libraryVersion` check looks like an oversight in first place. Could you verify this is doing what you expect?

libs/langchain-google-common/src/connection.ts
```diff
@@ -64,8 +64,9 @@ export abstract class GoogleConnection<
     clientLibraryVersion: string;
   }> {
     const env = await getRuntimeEnvironment();
+    const packageJson = require('../package.json');
```
afirstenberg, line 67: I've discussed this with @jacoblee93 before (who can correct me if I mis-state something).
Unfortunately, while this works in Node, it doesn't work in all environments. Otherwise it would have been done in `getRuntimeEnvironment()` (which would be the best place to fix it, so it is available to all packages).

`require` itself is node-specific, but it isn't the only problem here.

tanyaasharma:
Hi @afirstenberg @jacoblee93 ,
I was researching through langchain documentations in order to understand how to get langchain version in the runtime. Unfortunately i am not able to find the same in the docs and in API reference portal here https://api.js.langchain.com/. 

For now, i think we can make this hardcoded as we have mentioned in the package.json. Any thoughts?

afirstenberg:
> For now, i think we can make this hardcoded as we have mentioned in the package.json. Any thoughts?

Hardcoded to what value?

If you mean hardcoded to the same value in "package.json", there would need to be some way to update that value every time the value in package.json gets updated.

tanyaasharma:
Yes,lets just keep in mind to update this value, until we figure out some solution?

afirstenberg:
> Yes,lets just keep in mind to update this value, until we figure out some solution?

That doesn't scale, and is unlikely to happen. The release process itself is highly automated precisely so updating the versions manually aren't required.

Is there a  reason we can't leave this set at 0?  Or 0.0.0 if you insist on that formatting?

What analytics are you trying to get from this version info, specifically?

tanyaasharma:
Understood, I have informed the team, lets see if zero is fine. Will get back to you on this.

jacoblee93:
We currently don't support getting the current version at runtime :(

The solution would be to export some file with a string we update on release, but would need to build that out.

jacoblee93:
The `package.json` file isn't accessible at runtime to my knowledge

holtskinner:
Hi, I did the original implementation for this tracking in the Python LangChain library. If it's not possible to get the version number for the LangChain JS library, then I think it's fine to just put zeroes in place. This was arbitrarily added by me just so we could track it if it's useful.

tanyaasharma:
Sure, @holtskinner , I will put it as zero.

libs/langchain-google-common/src/tests/chat_models.test.ts
```diff
@@ -80,9 +80,13 @@ describe("Mock ChatGoogle", () => {
     await model.invoke(messages);
 
     expect(record?.opts?.headers).toHaveProperty("User-Agent");
+    expect(record?.opts?.headers).toHaveProperty("Client-Info");
     expect(record.opts.headers["User-Agent"]).toMatch(
       /langchain-js\/[0-9.]+-ChatConnection/
     );
+    expect(record.opts.headers["Client-Info"]).toMatch(
+      /langchain-[a-zA-Z0-9.-]+\/[0-9.]+\/[a-zA-Z0-9.-]+/
```
afirstenberg, line 88: This part of the test fails.

libs/langchain-google-common/src/connection.ts
```diff
@@ -52,22 +52,29 @@ export abstract class GoogleConnection<
...
-  async _clientLibraryVersion(): Promise<string> {
-    const env = await getRuntimeEnvironment();
+  async _getClientInfo(): Promise<{
+    userAgent: string;
+    clientLibraryVersion: string;
+  }> {
+    const env = await getRuntimeEnvironment();   
     const langchain = env?.library ?? "langchain-js";
-    const langchainVersion = env?.libraryVersion ?? "0";
+    const langchainVersion = env?.libraryVersion ?? "0";;
```
jacoblee93, line 68: Please run `yarn format`

jacoblee93:
I'm just going to hardcode this to `0` rather than rely on a non-existent parameter.

afirstenberg:
> I'm just going to hardcode this to `0` rather than rely on a non-existent parameter.

I would have said to keep it as `env?.libraryVersion ?? "0"`. 
`RuntimeEnvironment.libraryVersion` is a valid parameter of that type as returned by `getRuntimeEnvironment()`, so it might be set some day. {:
And the test will catch it if so.

But I'm at the point where it clearly doesn't matter.
LGTM

tanyaasharma:
@jacoblee93 , I have added formatting too, is there anything else I should add or its good to go?

bracesproul (approved):
lgtm, thanks for being patient with us. Will get this merged and included in the next `@langchain/google-common` release. Thank you!

Issue #5070, opened 2024-04-11
google-genai [feature]: Support apiVersion


afirstenberg:
### Privileged issue

- [X] I am a LangChain maintainer, or was asked directly by a LangChain maintainer to create an issue here.

### Issue Content

Gemini 1.5 requires the v1beta API Version to be specified.

ggallotti:
+1

PR #5115, opened 2024-04-16
google-genai[patch]: update google/generative-ai version

HenryHengZJ:
Fixes [#5070](https://github.com/langchain-ai/langchainjs/issues/5070)
Allow `gemini-1.5-pro-latest` to be usable using latest google/generative-ai version


libs/langchain-google-genai/src/chat_models.ts
```diff
@@ -186,7 +186,7 @@ export class ChatGoogleGenerativeAI
   private client: GenerativeModel;
 
   get _isMultimodalModel() {
-    return this.model.includes("vision");
+    return this.model.includes("vision") || this.model === 'gemini-1.5-pro-latest';
```
afirstenberg, line 189: I think locking it to gemini-1.5-pro-latest will cause frustration later.
Perhaps a leading string check on "gemini-1.5"?

afirstenberg:
See also #5070 
Although the package makes the apiVersion "v1beta", it would be good to take this opportunity to add it for the model attribute.

afirstenberg:
I think the changes here are pretty modest, and certainly good to get the library that supports Gemini 1.5 in is useful.
This is a prerequisite for all of the other new features (at least in the google-genai package).

libs/langchain-google-genai/src/chat_models.ts
HenryHengZJ: thats a good suggestion, we can do that

HenryHengZJ:
> See also #5070 Although the package makes the apiVersion "v1beta", it would be good to take this opportunity to add it for the model attribute.

what do you mean by the model attribute?

afirstenberg:
> > See also #5070 Although the package makes the apiVersion "v1beta", it would be good to take this opportunity to add it for the model attribute.
> 
> what do you mean by the model attribute?

Yeah, that was a bit confusing on my part. Sorry.

What I meant to say is that it would be a good idea to add it as a attribute anywhere that "modelName" is also an attribute, with a way to set it anywhere that "modelName" can be set.

afirstenberg:
I would have still liked to see support for an `apiVersion` parameter, but at this point I think it should just be released ASAP. (@jacoblee93, @bracesproul)

jacoblee93:
Thank you!

PR #5127, opened 2024-04-17
langchain[patch]: Parallelize applyEvaluators used by `runOnDataset`

davidfant:
This makes running evals that have slow evaluators (e.g. when each evaluator is an LLM call), run much faster


jacoblee93:
Thank you for the PR!

CC @hinthornw

langchain/src/smith/runner_utils.ts
```diff
@@ -534,23 +534,22 @@ const applyEvaluators = async ({
...
   });
-  const results: Record<
-    string,
-    { run_id: string; execution_time?: number; feedback: Feedback[] }
-  > = {};
-  for (let i = 0; i < runs.length; i += 1) {
-    const run = runs[i];
-    const example = examples[i];
+
+  const results = await Promise.all(runs.map(async (run, i): Promise<{
```
hinthornw, line 538: Wouldn't this run all evaluators in parallel over the entire dataset, with no semaphor (max_concurrency)?

That would likely lead to flakiness. We should respect the concurrency setttings passed in to the `runOnDataset` function

bracesproul: We should be able to instead wrap this inside an instance of [`AsyncCaller`](https://api.js.langchain.com/classes/langchain_core_utils_async_caller.AsyncCaller.html) which we can pass `maxConcurrency` to

jacoblee93:
@davidfant can you let me know if this works for you?

By the way, we are also working on porting the `evaluate` method as a new way to run evals.

CC @hinthornw

jacoblee93:
Thank you!

davidfant:
@jacoblee93 yea, lgtm!

PR #5132, opened 2024-04-17
community[patch]: Add Metadata Filtering for Upstash Vector Store

fahreddinozcan:
This PR adds a metadata filtering options for the query vectors.


jacoblee93:
Thank you!

PR #5135, opened 2024-04-17
core[patch]: Fixed "Bad control character in string literal"

Huaichen:
The error "Bad control character in string literal" occurs when calling JSON.parse(json). This typically happens because the JSON string contains control characters that are not allowed, such as newline characters. Any control characters in the JSON string should be escaped before parsing it using JSON.parse().

Fixes # (issue)

### DIFF
--- a/langchain-core/src/output_parsers/structured.ts
+++ b/langchain-core/src/output_parsers/structured.ts
@@ -104,7 +104,13 @@ ${JSON.stringify(zodToJsonSchema(this.schema))}
       const json = text.includes("```")
         ? text.trim().split(/```(?:json)?/)[1]
         : text.trim();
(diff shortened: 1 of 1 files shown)


jacoblee93:
I'm ok with this but could you add a test?

Huaichen:
> Thanks for the PR!
> 
> I'm ok with this but could you add a test?

Hi Jacob,

I added a test and submitted the changes in PR #5160. The test contains some new lines ('\n') and can trigger the 'Bad control character' error. However, it failed the automated testing because it did trigger the error.

Appreciate your attention to this matter!

Huaichen

jacoblee93:
Thank you! Will have a look at that PR!

Issue #5133, opened 2024-04-17
Astra DB - collection bug


jinchi2013:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain.js documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain.js rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```
import {
  AstraDBVectorStore,
  AstraLibArgs,
} from '@langchain/community/vectorstores/astradb'
import { formatDocumentsAsString } from 'langchain/util/document'
import { GoogleVertexAIEmbeddings } from '@langchain/community/embeddings/googlevertexai'

const VertexAIEmbeddings = new GoogleVertexAIEmbeddings()

const getAstraDBRetriever = async () => {
  const astraConfig: AstraLibArgs = {
    token: '{{token}}',
    endpoint: '{{endpoint}}',
    collection: '{{collection}}',
  }

  const store = await AstraDBVectorStore.fromExistingIndex(VertexAIEmbeddings, astraConfig)

  return store.asRetriever({
    k: topK,
    searchType: 'similarity',
    metadata,
  }).pipe(formatDocumentsAsString)
}
```

### Error Message and Stack Trace (if applicable)

"Collection 'default_keyspace.my_testing_collection' already exists CollectionAlreadyExistsError: Collection 'default_keyspace.my_testing_collection' already exists\n    at Db.createCollection (/Users/dev/github/web/common/temp/node_modules/.pnpm/@datastax+astra-db-ts@1.0.1/node_modules/@datastax/astra-db-ts/dist/data-api/db.js:343:23)\n    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)\n    at async AstraDBVectorStore.initialize (/Users/dev/github/web/common/temp/node_modules/.pnpm/@langchain+community@0.0.49_@datastax+astra-db-ts@1.0.1_redis@4.6.13/node_modules/@langchain/community/dist/vectorstores/astradb.cjs:105:9)\n    at async AstraDBVectorStore.fromExistingIndex (/Users/dev/github/web/common/temp/node_modules/.pnpm/@langchain+community@0.0.49_@datastax+astra-db-ts@1.0.1_redis@4.6.13/node_modules/@langchain/community/dist/vectorstores/astradb.cjs:271:9)"

### Description

[...]

PR #5142, opened 2024-04-18
community[patch]: astradb: set collection checkExists=false by default

nicoloboschi:
After upgrading the astra client in https://github.com/langchain-ai/langchainjs/pull/5080, there existing code is not 100% compatible anymore. 
The client now checks if the collection already exists by default and in that case, throws an error. 

In this PR we disable that flag by default to restore the same behaviour as before the upgrade.
Related issue: https://github.com/langchain-ai/langchainjs/issues/5133

(also bump min version to 1.0.1 which contains some bugfixes)

### DIFF
--- a/libs/langchain-community/package.json
+++ b/libs/langchain-community/package.json
@@ -63,7 +63,7 @@
     "@clickhouse/client": "^0.2.5",
     "@cloudflare/ai": "^1.0.12",
     "@cloudflare/workers-types": "^4.20230922.0",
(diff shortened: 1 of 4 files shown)


bracesproul (approved):
lgtm, thanks for pushing this one up!

nicoloboschi:
@bracesproul Hey, I fixed the format. Would you mind to merge the PR?

jacoblee93:
Thank you!

PR #5145, opened 2024-04-18
community[patch]: Invoke toTitleCase only when string is present to avoid errors

bilalshareef:
@jacoblee93


bracesproul (approved):
thanks for pushing this! lgtm

bilalshareef:
> thanks for pushing this! lgtm

Thanks for approving @bracesproul 
I just fixed a formatting(prettier) error. Now, it should be good to merge.

jacoblee93:
Thank you!

PR #5147, opened 2024-04-18
fix: incorrect text_type type declaration in Alibaba Tongyi Embeddings

xyspg:
corrects the type declaration for the `text_type` parameter in Alibaba Tongyi embeddings. The previous declaration incorrectly listed `text` as an option, which is not supported. The corrected declaration includes the intended options: `query` and `document` as shown in [official docs](https://help.aliyun.com/zh/dashscope/developer-reference/text-embedding-api-details) and comment above it.

Translation for the original comment: 
```js
/**
 * Value: query or document, default value is document
 * Description: After text is converted into vectors, it can be applied to downstream tasks such as retrieval, clustering, and classification.
 * For asymmetric tasks like retrieval, to achieve better retrieval effects, it is recommended to distinguish between query text (query) and
 * document text (document). For symmetric tasks like clustering and classification, there is no need to specify,
 * and the system's default value "document" can be used.
 */
```


bracesproul (approved):
good catch, thanks for fixing!

PR #5150, opened 2024-04-19
community[minor]: Add Upstash Embeddings Support

fahreddinozcan:
This PR adds Upstash Embeddings support, which allows users to send their data to Upstash Vector Database without creating the embeddings. Upstash Embeddings will handle the embedding operation for them.

More context [on Upstash announcement](https://twitter.com/upstash/status/1779908103200538776)


libs/langchain-community/src/vectorstores/upstash.ts
```diff
@@ -203,7 +272,7 @@ export class UpstashVectorStore extends VectorStore {
   static async fromTexts(
     texts: string[],
     metadatas: UpstashMetadata | UpstashMetadata[],
-    embeddings: EmbeddingsInterface,
+    embeddings: EmbeddingsInterface | "UpstashEmbeddings",
```
jacoblee93, line 275: We've been using `FakeEmbeddings` like this - could we do the same here?

https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-community/src/vectorstores/vectara.ts#L176

jacoblee93:
Sorry for the delay!

libs/langchain-community/src/vectorstores/upstash.ts
jacoblee93: Oh, I mean instead of passing a string here, pass `FakeEmbeddings`

jacoblee93 (changes requested):

libs/langchain-community/src/vectorstores/upstash.ts
fahreddinozcan: The thing is, passing something explicit like "UpstashEmbeddings" makes it easier to understand you have the option to either get the embeddings from Upstash, or passing a embedding client to create the embeddings your self. In the Vectara case, it seems that there's no option to pass an embedding, so all the embeddings will be created directly by it self. With Upstash, this is optional. Sorry for the delay here.

buggyhunter:
Hey @jacoblee93 any chance for review again?

jacoblee93:
> Hey @jacoblee93 any chance for review again?

Hey, sorry for the delay!

I left a note about using `FakeEmbeddings` and documenting it rather than the string. While I do agree that a string is more descriptive, we must adhere to the base class interface for now otherwise we'll have problems with types etc. We have some refactoring of vectorstores planned in the future and we can apply a better fix then.

fahreddinozcan:
> > Hey @jacoblee93 any chance for review again?
> 
> Hey, sorry for the delay!
> 
> I left a note about using `FakeEmbeddings` and documenting it rather than the string. While I do agree that a string is more descriptive, we must adhere to the base class interface for now otherwise we'll have problems with types etc. We have some refactoring of vectorstores planned in the future and we can apply a better fix then.

Hey, it's ready as you asked.

libs/langchain-community/src/vectorstores/upstash.ts
```diff
@@ -78,10 +96,14 @@ export class UpstashVectorStore extends VectorStore {
    */
   async addDocuments(
     documents: DocumentInterface[],
-    options?: { ids?: string[] }
+    options?: { ids?: string[]; UpstashEmbeddings?: boolean }
```
jacoblee93, line 99: Small style nit - we generally use camel case for properties. I can push a rename

libs/langchain-community/src/vectorstores/upstash.ts
```diff
@@ -45,22 +50,35 @@ const CONCURRENT_UPSERT_LIMIT = 1000;
 export class UpstashVectorStore extends VectorStore {
   declare FilterType: string;
 
+  declare embeddings: EmbeddingsInterface;
```
jacoblee93, line 53: Should make this an optional prop

libs/langchain-community/src/vectorstores/upstash.ts
```diff
@@ -45,22 +50,35 @@ const CONCURRENT_UPSERT_LIMIT = 1000;
...
+
   index: UpstashIndex;
 
   caller: AsyncCaller;
 
-  embeddings: EmbeddingsInterface;
+  upstashEmbeddingsConfig?: boolean;
```
jacoblee93, line 59: Renaming to `useUpstashEmbeddings` instead

libs/langchain-community/src/vectorstores/upstash.ts
```diff
@@ -128,6 +150,45 @@ export class UpstashVectorStore extends VectorStore {
     return documentIds;
   }
 
+  /**
+   * This method adds the provided documents to Upstash database. The pageContent of the documents will be embedded by Upstash Embeddings.
+   * @param documents Array of Document objects to be added to the Upstash database.
+   * @param options Optional object containing the array of ids for the documents.
+   * @returns Promise that resolves with the ids of the provided documents when the upsert operation is done.
+   */
+  async addData(documents: DocumentInterface[], options?: { ids?: string[] }) {
```
jacoblee93, line 159: Going to make this `protected`

jacoblee93:
Thank you! I renamed a couple of things but re-ran the integration tests and they passed. Will get this released today.

jacoblee93 (approved):

PR #5160, opened 2024-04-20
core[patch]: Fix "Bad control character in string literal"

Huaichen:
The error "Bad control character in string literal" occurs when calling JSON.parse(json). This typically happens because the JSON string contains control characters that are not allowed, such as newline characters. Any control characters in the JSON string should be escaped before parsing it using JSON.parse().

A test is added to "langchain-core/src/output_parsers/structured.ts" with adding several newlines into the json. The test failed due to "Error: SyntaxError: Bad control character in string literal in JSON at position 36 (line 2 column 36)".  

The code in parse() was updated by replacing the newline literal inside a string element with '\n', and removing the newline literal outside of the string elements.


jacoblee93:
Thank you! Will have a look at that PR.

langchain-core/src/output_parsers/structured.ts
```diff
@@ -104,7 +104,15 @@ ${JSON.stringify(zodToJsonSchema(this.schema))}
...
         ? text.trim().split(/```(?:json)?/)[1]
         : text.trim();
-      return await this.schema.parseAsync(JSON.parse(json));
+
+      const escapedJson = json
+        .replace(/"([^"\\]*(\\.[^"\\]*)*)"/g, (_match, capturedGroup) => {
+          const escapedInsideQuotes = capturedGroup.replace(/\n/g, "\\n");
+          return `"${escapedInsideQuotes}"`;
+        })
+        .replace(/\n/g, "");
```
jacoblee93, line 113: What is the purpose of this second replace?

jacoblee93:
Thank you! One small question

jacoblee93:
Thank you!

PR #5166, opened 2024-04-22
Update the google/generative-ai dependency

tsukasaoishi:
I updated the dependency @google/generative-ai because I wanted to use the API version v1beta.

Refs: https://github.com/google-gemini/generative-ai-js/pull/94


jacoblee93:
Yes merged the earlier PR - will ship a new version shortly!

Thank you!

Issue #3315, opened 2023-11-17
Support New Bedrock Embeddings (Cohere)


metaskills:
New embedding are. Seems these use an `input_type` arg which I would like to use too.

* cohere.embed-english-v3
* cohere.embed-multilingual-v3

BrianErikson:
As supplemental information that I found during my investigation, it appears that [embeddings/bedrock.ts#L88C15-L88C33](https://github.com/langchain-ai/langchainjs/blob/02ee2a0fa360b4ededaa51e956eda759f87ceb4c/langchain/src/embeddings/bedrock.ts#L88C15-L88C33) should be revised to accommodate the new cohere embedding schema on AWS Bedrock --
```
{
  "modelId": "cohere.embed-english-v3",
  "contentType": "application/json",
  "accept": "*/*",
  "body": {
    "texts": ["Hello world", "This is a test"],
    "input_type": "search_document"
 } 
}
```
[AWS Documentation on Cohere Embed input_type](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-cohere.html#model-parameters-embed)
Note that the existing schema still needs to be supported for existing Titan Embeddings users. 

Here's a sample of the current error:
`Error: An error occurred while embedding documents with Bedrock: Malformed input request: 3 schema violations found, please reformat your input and try again.`

vishal-oogway:
I'd like to test cohere embeddings available on Bedrock and think this helper class would be really beneficial. Is it possible to get @BrianErikson 's contribution merged into community extensions ?

KwangGan:
Is there any update on this ?

PR #5167, opened 2024-04-22
Adds BedrockCohereEmbeddings class + tests

vishalseshagiri:
- Adds BedrockCohereEmbeddings class and associated tests, adds async keyword to BedrockEmbeddings.embedQuery function - Code credits to @ BrianErikson (Most of the application code comes from their thread [here](https://github.com/langchain-ai/langchainjs/issues/3315#issuecomment-1821857715)). Some notes about the chosen design.
  - Consciously chose not to add conditions (to support Cohere embeddings) within the BedrockEmbeddings class to avoid making it brittle and unreadable.
  - Avoided extending the `BedrockEmbeddings` class in the `BedrockCohereEmbeddings` class since it'll have access to the protected `_embedText` function, which has no meaning in the context of the `BedrockCohereEmbeddings` class
- Adds missing async keyword BedrockEmbeddings.embedQuery function

Fixes # (issue)

### DIFF
--- a/libs/langchain-community/src/embeddings/bedrock.ts
+++ b/libs/langchain-community/src/embeddings/bedrock.ts
@@ -122,7 +122,7 @@ export class BedrockEmbeddings
    * @param document Document for which to generate an embedding.
    * @returns Promise that resolves to an embedding for the input document.
    */
-  embedQuery(document: string): Promise<number[]> {
+  async embedQuery(document: string): Promise<number[]> {
     return this.caller.callWithOptions(
       {},
       this._embedText.bind(this),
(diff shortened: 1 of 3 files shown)


jacoblee93:
@efriis this is the direction we're heading in Python right?

vishal-oogway:
> @efriis this is the direction we're heading in Python right?

Ooh looks like that's not the direction in Python https://github.com/langchain-ai/langchain/blob/master/libs/community/langchain_community/embeddings/bedrock.py#L144-L148. I can refactor it to mimic it, if that's the expectation @jacoblee93

tomi-bigpi:
FYI: This will easily run into the 96 max document limit in the AWS API. We had to add batching logic to the calls to avoid errors.

hntrl:
Hey @vishalseshagiri, sorry to revive an old PR!

The community package has become rather unwieldy, so generally we recommend repackaging changes like this into a standalone NPM module. Since this is aws related, and given there is still an interest for this, I'd recommend that we involve this within the @langchain/aws package instead of within the community package

PR #5170, opened 2024-04-22
community[patch]: AstraDB: add option to skip create collection call

nicoloboschi:
While using AstraDB there are cases whereas the collection already exists and you don't want to store to try to initialize it. For example, you might have read only permission and trying to create the collection will give you 403

Add boolean flag `skipCollectionProvisioning` to just assume the collection is already there with the correct settings.


jacoblee93:
Thank you!

Worth maybe splitting the `initialize` method long-term?

PR #5171, opened 2024-04-22
docs[patch]: Update Memgraph docs

katarinasupe:
I updated Memgraph docs so they reflect a new way of Memgraph Platform installation.


jacoblee93:
Thank you!

PR #5189, opened 2024-04-23
core[patch],langchain[patch]: Move BaseChain and LLMChain to legacy entrypoint in core

jacoblee93:
This will allow us to fully split community out of LangChain for 0.2

CC @eyurtsev @baskaryan


langchain/src/chains/llm_chain.ts
```diff
@@ -1,282 +1,3 @@
...
-  CallbackManager,
-  BaseCallbackConfig,
-  CallbackManagerForChainRun,
-  Callbacks,
-} from "@langchain/core/callbacks/manager";
-import { Runnable, type RunnableInterface } from "@langchain/core/runnables";
-import { BaseChain, ChainInputs } from "./base.js";
-import { SerializedLLMChain } from "./serde.js";
-import { NoOpOutputParser } from "../output_parsers/noop.js";
+import { type LLMChainInput, LLMChain } from "@langchain/core/legacy/chains";
```
eyurtsev, line 1: This is under the namespace of langchain/chains/llm in python rather than llm_chain

langchain-core/src/legacy/chains/llm.ts
```diff
@@ -0,0 +1,284 @@
+import {
```
eyurtsev, line 1: good matched

langchain-core/langchain.config.js
```diff
@@ -24,6 +24,7 @@ export const config = {
     "language_models/base": "language_models/base",
     "language_models/chat_models": "language_models/chat_models",
     "language_models/llms": "language_models/llms",
+    "legacy/chains": "legacy/chains",
```
eyurtsev, line 27: What is this config for?

jacoblee93: Stuff

langchain/src/chains/llm_chain.ts
jacoblee93: I think mismatch is ok, nobody should be importing from core directly anyway and the extra entrypoint is slightly burdensome

No problem to add if you feel strongly

PR #5190, opened 2024-04-23
google-common [feature]: Support JSON Mode for Gemini #5071

### LINKED ISSUE #5071: google-common [feature]: Support JSON Mode for Gemini
### Privileged issue

### Issue Content

See documentation for `responseMimeType` at https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/gemini#request_body

See https://github.com/langchain-ai/langchain-google/issues/144

JackFener:
Fixes #5071
Added the opportunity to choose responseMimeType when calling google ai models.
Officially available since Gemini 1.5 pro (but tested finely also on Gemini 1.0 pro), you can choose to have the LLM output formatted into JSON object.
The model.stream() method still returns a string containing the JSON. I would like to add the automatic parsing of the string, but it's not always true that Gemini returns a correctly formatted json.
I also don't like to have the .call() method returning 2 different types.
So by now the user still has to JSON.parse(response) to have an object response.
I would be happy to work also on the other google related tasks.
Not very active on X but you can find me at @Giacomo_fava or on LinkedIn at @giacomofavaai. Not here to self promote, just linking my handle as the pr comment suggests


libs/langchain-google-common/src/types.ts
```diff
@@ -91,6 +91,15 @@ export interface GoogleAIModelParams {
...
   convertSystemMessageToHumanContent?: boolean;
+
+  /**
+   * Available for gemini-1.5-pro.
+   * The output format of the generated candidate text.
+   * Supported MIME types:
+   *     text/plain: (default) Text output.
+   *     application/json: JSON response in the candidates.
+   */
+  responseMimeType?: "text/plain" | "application/json"
```
afirstenberg, line 102: Good to see the comment!
I would suggest making the two strings a separate type and using that type here and down in `GeminiGenerationConfig`.

libs/langchain-google-common/src/types.ts
```diff
@@ -218,6 +227,7 @@ export interface GeminiGenerationConfig {
   temperature?: number;
   topP?: number;
   topK?: number;
+  responseMimeType?: "text/plain" | "application/json"
```
afirstenberg, line 230: Other location to use the type I suggestd above.

libs/langchain-google-common/src/utils/common.ts
```diff
@@ -41,6 +41,11 @@ export function copyAIModelParamsInto(
     options?.convertSystemMessageToHumanContent ??
     params?.convertSystemMessageToHumanContent ??
     target?.convertSystemMessageToHumanContent;
+  ret.responseMimeType =
+    options?.responseMimeType ??
+    params?.responseMimeType ??
+    target?.responseMimeType ??
+    "text/plain";
```
afirstenberg, line 48: Rather than put the default here, we tend to have the defaults specified in the class that implements it (ie - `GoogleBaseLLM` and `ChatGoogleBase`.

libs/langchain-google-common/src/tests/llms.test.ts
```diff
@@ -521,4 +521,30 @@ describe("Mock Google LLM", () => {
...
   });
+
+  test("8: streamGenerateContent - streaming - json responseMimeType", async () => {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const record: Record<string, any> = {};
+    const projectId = mockId();
+    const authOptions: MockClientAuthInfo = {
+      record,
+      projectId,
+      resultFile: "llm-8-mock.json",
```
afirstenberg, line 532: I don't see the llm-8-mock.json file checked in.

libs/langchain-google-common/src/tests/llms.test.ts
```diff
@@ -521,4 +521,30 @@ describe("Mock Google LLM", () => {
     expect(responseArray).toHaveLength(3);
     console.log("record", JSON.stringify(record, null, 2));
   });
+
+  test("8: streamGenerateContent - streaming - json responseMimeType", async () => {
```
afirstenberg, line 525: Can we also get a test to make sure the request contains the `responseMimeType` setting in the generation config?

afirstenberg (commented):
Overall looks great! Thank you for the update!

Some thoughts below.

Two additional questions:
- Do we know if this also works on the AI Studio API?
- What happens if we include the setting on a model that doesn't support it?

libs/langchain-google-common/src/types.ts
JackFener: definitely I will

libs/langchain-google-common/src/tests/llms.test.ts
JackFener: my bad, will push it asap

JackFener:
just committed all the suggested changes :)

libs/langchain-google-common/src/llms.ts
```diff
@@ -14,7 +14,7 @@ import {
   GoogleAIModelParams,
   GoogleAISafetySetting,
   GooglePlatformType,
-  GeminiContent,
+  GeminiContent, GoogleAIResponseMimeType,
```
afirstenberg, line 17: Make sure you run `yarn format`

siddiqaa:
@JackFener does your fix / commit mean that the table in Model Overview here should say that JSON modes IS supported? https://python.langchain.com/v0.2/docs/integrations/chat/google_generative_ai/

rudder-sanket-gaikwad:
> @JackFener does your fix / commit mean that the table in Model Overview here should say that JSON modes IS supported? https://python.langchain.com/v0.2/docs/integrations/chat/google_generative_ai/

No it still shows unsupported

PR #5195, opened 2024-04-24
community[minor]: Improve Azure AI Search vector store support

sinedied:
This PR improves the existing Azure Cosmos DB vector store in the following ways:
- Add support for custom credentials
- Set default search query type


jacoblee93:
Seems reasonable, thank you!

PR #5197, opened 2024-04-24
community[minor]: Improve Azure Cosmos DB vector store support

sinedied:
This PR improves the existing Azure Cosmos DB vector store in the following ways:
- Add automatic index creation
- Add automatic embeddings length detection when not provided
- Allow to delete documents using mongoDB filters

It does not include any breaking changes.


libs/langchain-community/src/vectorstores/azure_cosmosdb.ts
```diff
@@ -179,10 +200,17 @@ export class AzureCosmosDBVectorStore extends VectorStore {
    */
   async createIndex(
     numLists = 100,
-    dimensions = 1536,
+    dimensions: number | undefined = undefined,
     similarity: AzureCosmosDBSimilarityType = AzureCosmosDBSimilarityType.COS
   ): Promise<void> {
-    await this.initPromise;
+    await this.connectPromise;
```
jacoblee93, line 206: If this is hypothetically called before `connectPromise` is initialized, could we throw instead?

libs/langchain-community/src/vectorstores/azure_cosmosdb.ts
```diff
@@ -205,16 +233,24 @@ export class AzureCosmosDBVectorStore extends VectorStore {
...
-   * @param ids IDs of the documents to be removed. If no IDs are specified,
-   *     all documents will be removed.
+   * @param filterOrIds A MongoDB filter object or list of IDs for the
+   *     documents to be removed. If no IDs or filter are specified, all
+   *     documents will be removed.
    * @returns A promise that resolves when the documents have been removed.
    */
-  async delete(ids?: string[]): Promise<void> {
+  async delete(
+    filterOrIds?: string[] | Filter<MongoDBDocument>
```
jacoblee93, line 242: Would prefer if this was either an array of `ids` or an `options: { filter: Filter<MongoDBDocument> }` param

libs/langchain-community/src/vectorstores/azure_cosmosdb.ts
jacoblee93: Ah never mind, I see below. Could we always just rely on `this.initPromise` instead?

sinedied: Since `init` calls `createIndex` we need 2 different promise there otherwise there's an interlocking:
- connectPromise => create DB + collections clients
- initPromise => connectPromise + createIndex

sinedied: For the throw question, the issue is that the promise is created in the constructor and there's no way for the user to wait for it or know when the connect task is done.

libs/langchain-community/src/vectorstores/azure_cosmosdb.ts
sinedied: Done!

libs/langchain-community/src/vectorstores/azure_cosmosdb.ts
```diff
@@ -205,19 +233,27 @@ export class AzureCosmosDBVectorStore extends VectorStore {
...
-   * @param ids IDs of the documents to be removed. If no IDs are specified,
-   *     all documents will be removed.
+   * If no IDs or filter are specified, all documents will be removed.
+   * @param ids A list of IDs for the documents to be removed.
+   * @param filter A MongoDB filter object or list of IDs for the documents to be removed.
    * @returns A promise that resolves when the documents have been removed.
    */
-  async delete(ids?: string[]): Promise<void> {
+  async delete(
+    ids?: string | string[],
```
jacoblee93, line 242: Hey sorry, no I mean a single param like this so as not to break the interface:

https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-community/src/vectorstores/pinecone.ts#L174

sinedied: Done, I updated and rebased the PR

libs/langchain-community/src/vectorstores/azure_cosmosdb.ts
```diff
@@ -205,19 +241,24 @@ export class AzureCosmosDBVectorStore extends VectorStore {
...
   /**
    * Removes specified documents from the AzureCosmosDBVectorStore.
-   * @param ids IDs of the documents to be removed. If no IDs are specified,
-   *     all documents will be removed.
+   * If no IDs or filter are specified, all documents will be removed.
+   * @param params Parameters for the delete operation.
    * @returns A promise that resolves when the documents have been removed.
    */
-  async delete(ids?: string[]): Promise<void> {
+  async delete(params: AzureCosmosDBDeleteParams = {}): Promise<void> {
```
bracesproul, line 248: This will be a breaking change. Instead we should overload this func so both input types are allowed, and we can narrow the type inside of the method.

libs/langchain-community/src/vectorstores/azure_cosmosdb.ts
```diff
@@ -236,27 +277,31 @@ export class AzureCosmosDBVectorStore extends VectorStore {
    * Method for adding vectors to the AzureCosmosDBVectorStore.
    * @param vectors Vectors to be added.
    * @param documents Corresponding documents to be added.
-   * @returns A promise that resolves when the vectors and documents have been added.
+   * @returns A promise that resolves to the added documents IDs.
    */
-  async addVectors(vectors: number[][], documents: Document[]): Promise<void> {
+  async addVectors(
+    vectors: number[][],
+    documents: Document[]
```
bracesproul, line 284: While we're at it, this should be `DocumentInterface`

libs/langchain-community/src/vectorstores/azure_cosmosdb.ts
```diff
@@ -236,27 +277,31 @@ export class AzureCosmosDBVectorStore extends VectorStore {
...
 
   /**
    * Method for adding documents to the AzureCosmosDBVectorStore. It first converts
    * the documents to texts and then adds them as vectors.
    * @param documents The documents to add.
-   * @returns A promise that resolves when the documents have been added.
+   * @returns A promise that resolves to the added documents IDs.
    */
-  async addDocuments(documents: Document[]): Promise<void> {
+  async addDocuments(documents: Document[]): Promise<string[]> {
```
bracesproul, line 302: `DocumentInterface` here too

libs/langchain-community/src/vectorstores/azure_cosmosdb.ts
```diff
@@ -355,9 +400,21 @@ export class AzureCosmosDBVectorStore extends VectorStore {
...
-    this.collection = this.database.collection(collectionName);
+    this.connectPromise = (async () => {
+      await client.connect();
+      this.database = client.db(databaseName);
+      this.collection = this.database.collection(collectionName);
+    })();
+
+    // Unless skipCreate is set, create the index
+    // This operation is no-op if the index already exists
+    if (!this.indexOptions.skipCreate) {
```
bracesproul, line 411: This could be breaking to someone who already has a flow setup which creates an index or fails. Instead, let's make this opt-in, document it well and add a note saying in version `0.3.0` (planned for a few months from now, breaking changes will ship w/ this release) it will change to opt-out.

bracesproul (commented):
Overall looks solid, couple comments. Please request my review once implemented, these will be great to get released!

libs/langchain-community/src/vectorstores/azure_cosmosdb.ts
jacoblee93: I think worth it

jacoblee93:
Thank you!

PR #5204, opened 2024-04-24
langchain[major],community[major],core[minor]: LangChain 0.2

jacoblee93:


langchain/src/schema/query_constructor.ts
bracesproul, line 1: No deprecation warning here, used throughout examples

deleting this file is breaking

jacoblee93: Feel free to add back in

PR #5205, opened 2024-04-24
core[patch]: Bugfix: cache missed because callback `_parentRunId` passed in to cache key function (`_getSerializedCacheKeyParametersForCall`)

davidfant:
This PR is a patch I've applied locally to fix the problem below. Whoever understands this code should prob change this fix to be more appropriate. Also, the TS types don't reflect a "config" being in callOptions, so idk where that comes from.

### Problem
Here's an example output from `_getSerializedCacheKeyParametersForCall`, which is used to generate BaseLanguageModel cache keys:
```
_model:"base_chat_model",_type:"openai",apiKey:"...",config:{"tags":[],"metadata":{},"callbacks":{"handlers":[{"lc":1,"type":"not_implemented","id":["langchain_core","callbacks","langchain_tracer","LangChainTracer"]}],"inheritableHandlers":[{"lc":1,"type":"not_implemented","id":["langchain_core","callbacks","langchain_tracer","LangChainTracer"]}],"tags":[],"inheritableTags":[],"metadata":{},"inheritableMetadata":{},"name":"callback_manager","_parentRunId":"1f2f3a62-f04c-4e82-9bd0-11c8747c734b"},"recursionLimit":99,"outputKeys":"__end__","configurable":{}},dangerouslyAllowBrowser:true,frequency_penalty:0,max_tokens:1000,model:"gpt-4-1106-preview",model_name:"gpt-4-1106-preview",n:1,outputKeys:"__end__",presence_penalty:0,stream:false,temperature:0,tool_choice:{"type":"function","function":{"name":"respond"}},tools:[{"type":"function","function":{"name":"respond","parameters":{"type":"object","properties":{"reasoning":{"type":"string"},"response":{"anyOf":[{"type":"object","properties":{"type":{"type":"string","const":"sendMessage"},"message":{"type":"string"}},"required":["type","message"],"additionalProperties":false},{"type":"object","properties":{"type":{"type":"string","const":"submitWorkflowGoal"},"name":{"type":"string"},"goal":{"type":"string"}},"required":["type","name","goal"],"additionalProperties":false}]}},"required":["reasoning","response"],"additionalProperties":false,"$schema":"http://json-schema.org/draft-07/schema#"}}}],top_p:1
```

What's problematic about this is `"_parentRunId":"1f2f3a62-f04c-4e82-9bd0-11c8747c734b"`, which causes the cache key to be different for each run. My fix makes sure that `config` isn't being passed in when generating this key.


jacoblee93:
Hmm interesting! Will investigate, thank you!

jacoblee93:
Oh, I think you might be using a `RunnableLambda` in LangSmith?

jacoblee93:
This has to do with a backwards compatibility shim around `RunnableLambda` typing - would be breaking to fix it properly, so this is fine for now.

jacoblee93:
Thank you!

PR #5215, opened 2024-04-25
community[patch]: Neo4j metadata filtering

easwee:
Porting metadata filtering from python.


libs/langchain-community/src/vectorstores/neo4j_vector.ts
```diff
@@ -4,10 +4,20 @@ import type { EmbeddingsInterface } from "@langchain/core/embeddings";
 import { VectorStore } from "@langchain/core/vectorstores";
 import { Document } from "@langchain/core/documents";
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type Any = any;
```
jacoblee93, line 8: Ha, you can just do `/* eslint-disable @typescript-eslint/no-explicit-any */` at the top if you prefer

libs/langchain-community/src/vectorstores/neo4j_vector.ts
```diff
@@ -4,10 +4,20 @@ import type { EmbeddingsInterface } from "@langchain/core/embeddings";
...
+
 export type SearchType = "vector" | "hybrid";
 
+export enum IndexType {
```
jacoblee93, line 12: We generally try to avoid `enum` due to bundler/environment issues.

libs/langchain-community/src/vectorstores/neo4j_vector.ts
```diff
@@ -576,16 +625,17 @@ export class Neo4jVectorStore extends VectorStore {
   async similaritySearch(
     query: string,
     k = 4,
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    params: Record<string, any> = {}
+    params: Record<string, Any> = {},
+    filter: Record<string, Any> | undefined = undefined
```
jacoblee93, line 629: Could we massage this back to the signature in the base class?

Stuff `filter` into `params`?

https://github.com/langchain-ai/langchainjs/blob/main/langchain-core/src/vectorstores.ts#L171

jacoblee93 (changes requested):

libs/langchain-community/src/vectorstores/neo4j_vector.ts
easwee: Refactored to plain type `export type IndexType = "NODE" | "RELATIONSHIP";`

libs/langchain-community/src/vectorstores/neo4j_vector.ts
jacoblee93: It causes broader package maintenance issues - we could look at updating the base class in the future but we generally try to supply injected params in the constructor, then reserve this arg for filtering.

How about:

```js
params: {
  filter: Record<string, Any>,
} & Record<string, any>
```

tomasonjo: Ok, this can work for me

easwee: Updated - filter is a params object prop.
Example usage:
image

jacoblee93 (approved):

PR #5218, opened 2024-04-25
community[patch]: Astra updates

mfortman11:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=etsTvLY)

Utilizes new astra client features

Add Docs now updates docs w/ the same ids


jacoblee93:
Looks good! Can you mark `batchSize` as deprecated to let them know that it no longer has an effect?

mfortman11:
All set

bracesproul (approved):
tyvm!

bracesproul (commented):
looks like CI is failing, could you fix and re-request my review so I can merge?

mfortman11:
@bracesproul sorry for the delay the the issue appears to be around `batchSize` with this change it is no longer need so i added the deprecated note to it per Jacobs request but that is the cause of the build error

```  /** @deprecated astra-db-ts > 1.0.0 handles this */
  private readonly batchSize: number; ```
  
  Lmk how i should handle this

bracesproul (approved):
I just pushed a commit doing two things:
1. dropped the `batchSize` class property. Since it was already marked as `private readonly` deleting it won't be breaking for anyone.
2. I added a `console.warn` notifying users that passing in the `batchSize` arg has no effect anymore.

Thanks for doing this! Going to get it merged asap

PR #5221, opened 2024-04-26
community[patch]: Fix Replicate API

smontgomerie:
Fixes [#20757](https://github.com/langchain-ai/langchain/issues/20757) but for LangChainJS


libs/langchain-community/src/llms/replicate.ts
```diff
@@ -147,7 +168,8 @@ export class Replicate extends LLM implements ReplicateInput {
     Replicate: typeof import("replicate").default;
   }> {
     try {
-      const { default: Replicate } = await import("replicate");
+      // eslint-disable-next-line @typescript-eslint/no-var-requires,global-require
+      const Replicate = require("replicate");
```
jacoblee93, line 172: This will cause issue in some environments I think?

We actually don't need dynamic imports anymore, since we've added different entrypoints.

jacoblee93:
Thanks for this! Ran format and updated imports, but something still seems to be up - think you could have another look?

libs/langchain-community/src/llms/replicate.ts
smontgomerie: `import` crashed for me in linux and osx with Jest. Is there a better way to do this?

smontgomerie:
> Thanks for this! Ran format and updated imports, but something still seems to be up - the tests all return just `""`. Think you could have another look?

It needs to be run with a REPLICATE_API_TOKEN env var. I assume CI/CD has this?

And also it took two times where it timed out before it started working. I assume this is because Replicate needs to spin up a model on demand? It does work very quickly for me after it spins up.

bracesproul:
Closing due to merge conflicts. Please re-open off of `main` if you'd still like to get this merged. Thank you!

PR #5227, opened 2024-04-26
core[patch]: Export `mergeConfigs()`

andrewnguonly:
This matches the Python implementation and is needed to implement `RunnableCallable` in LangGraphJS.

Python implementation of `RunnableCallable`: https://github.com/langchain-ai/langgraph/blob/main/langgraph/utils.py#L5


jacoblee93:
Thanks! Docs build is a flake.

PR #5246, opened 2024-04-29
langchain[patch]: Make `DynamicRunEvaluatorParams` generic

davidfant:
Currently I'm doing this kind of dirty type casting in my code, which I want to avoid
```ts
const input = params.input as unknown as MyInput;
const prediction = (params.prediction as MyPred | undefined)?.someKnownField
```


bracesproul (approved):
Nice, this will be good. Ty for doing this!

cc @hinthornw

PR #5247, opened 2024-04-29
google-common[patch]: remove all additionalProperties for gemini

### LINKED ISSUE #5240: google-common [bug]: zodToGeminiParameters leaves in some additionalProperties
### Checked other resources

### Example Code

A zod schema definition something like this:

```typescript
z.object({
  infoMap: z.object({
  });
})
```

produces JSON something like:

```json
                 "parameters": {
                    "type": "object",
                    "properties": {
                      "infoMap": {
                        "type": "object",
                        "properties": {},
                        "additionalProperties": false
                      }
                    }
                   }
```

Gemini API returns an HTTP 400 error code that includes the following:

```
"errors": [
      {
        "message": "Invalid JSON payload received. Unknown name \"additionalProperties\" at 'tools[0].function_declarations[3].parameters.properties[0].value': Cannot find field.",
        "reason": "invalid"
      }
    ]
```

### Description

The "additionalProperites" field is correctly stripped from the "parameters", but is left in  the child object.

From a quick look at `zodToGeminiParameters()` it just does this for the top-level object, and doesn't descend into child objects.

### System Info

gemini-common: 0.0.11
gemini-vertex: 0.0.11

Vertex AI Gemini API platform
model: gemini-1.0-pro-001

jeasonnow:
Fixes #5240

### DIFF
--- a/libs/langchain-google-common/src/types.ts
+++ b/libs/langchain-google-common/src/types.ts
@@ -274,3 +274,13 @@ export interface GoogleAISafetyHandler {
 export interface GoogleAISafetyParams {
   safetyHandler?: GoogleAISafetyHandler;
 }
+
+export type GeminiJsonSchema = Record<string, unknown> & {
+  properties?: Record<string, GeminiJsonSchema>;
+  type: GeminiFunctionSchemaType;
+};
+
+export interface GeminiJsonSchemaDirty extends GeminiJsonSchema {
+  properties?: Record<string, GeminiJsonSchemaDirty>;
+  additionalProperties?: boolean;
+}
--- a/libs/langchain-google-common/src/utils/zod_to_gemini_parameters.ts
+++ b/libs/langchain-google-common/src/utils/zod_to_gemini_parameters.ts
@@ -2,7 +2,41 @@
 
 import type { z } from "zod";
 import { zodToJsonSchema } from "zod-to-json-schema";
-import { GeminiFunctionSchema } from "../types.js";
+import {
+  GeminiFunctionSchema,
+  GeminiJsonSchema,
+  GeminiJsonSchemaDirty,
+} from "../types.js";
+
+function removeAdditionalProperties(
+  schema: GeminiJsonSchemaDirty
+): GeminiJsonSchema {
+  const updatedSchema: GeminiJsonSchemaDirty = { ...schema };
+  if (Object.hasOwn(updatedSchema, "additionalProperties")) {
+    delete updatedSchema.additionalProperties;
+  }
+  if (updatedSchema.properties) {
+    const keys = Object.keys(updatedSchema.properties);
+    removeProperties(updatedSchema.properties, keys, 0);
+  }
+
+  return updatedSchema;
+}
+
+function removeProperties(
+  properties: Record<string, GeminiJsonSchemaDirty>,
+  keys: string[],
+  index: number
+): void {
+  if (index >= keys.length) {
+    return;
+  }
+
+  const key = keys[index];
+  // eslint-disable-next-line no-param-reassign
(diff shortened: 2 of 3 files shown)


bracesproul (commented):
Besides my comment, it looks good. Please re-request my review once you've responded.

cc @afirstenberg

libs/langchain-google-common/src/utils/zod_to_gemini_parameters.ts
```diff
@@ -2,7 +2,37 @@
...
 import type { z } from "zod";
 import { zodToJsonSchema } from "zod-to-json-schema";
-import { GeminiFunctionSchema } from "../types.js";
+import { GeminiFunctionSchema, GeminiJsonSchema } from "../types.js";
+
+function removeAdditionalProperties(
+  schema: GeminiJsonSchema
+): GeminiJsonSchema {
+  const updatedSchema: GeminiJsonSchema = { ...schema };
+  if (typeof updatedSchema.additionalProperties === "boolean") {
```
afirstenberg, line 11: Why specifically test if its a boolean instead of just testing for its existence?

libs/langchain-google-common/src/types.ts
```diff
@@ -274,3 +274,9 @@ export interface GoogleAISafetyHandler {
 export interface GoogleAISafetyParams {
   safetyHandler?: GoogleAISafetyHandler;
 }
+
+export type GeminiJsonSchema = Record<string, unknown> & {
+  additionalProperties?: boolean;
```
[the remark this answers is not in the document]
afirstenberg: That's actually an interesting issue.
I think the issue is that if they don't cast it as some type that _has_ the field, TypeScript will complain.

I might say to create a special type (say, `GeminiJsonSchemaDirty`) for this that extends `GeminiJsonSchema` and adds the property... so we can check for it. I feel like there may be syntactic sugar that tells TypeScript to ignore the type checking for one operation, but I can't find that syntax quickly, and I think a subtype might make it clearer what is going on anyway.

afirstenberg:
Oh wow, I didn't expect someone to tackle this so quickly. I mostly opened the issue to remind myself to get to it eventually. Thank you, @jeasonnow.

libs/langchain-google-common/src/utils/zod_to_gemini_parameters.ts
afirstenberg: Perhaps use `Object.hasOwn()`?
(I don't have a strong opinion on this. I just want to make sure we understand why we're testing it this way, since it seems less obvious.)

jeasonnow: Oh, I think that’s a better approach. Thanks for your help.

libs/langchain-google-common/src/utils/zod_to_gemini_parameters.ts
```diff
@@ -11,8 +45,11 @@ export function zodToGeminiParameters(
   // Gemini doesn't accept either the $schema or additionalProperties
   // attributes, so we need to explicitly remove them.
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  const jsonSchema = zodToJsonSchema(zodObj) as any;
-  const { $schema, additionalProperties, ...rest } = jsonSchema;
+  // const jsonSchema = zodToJsonSchema(zodObj) as any;
+  const jsonSchema = removeAdditionalProperties(
+    zodToJsonSchema(zodObj) as GeminiJsonSchema
```
afirstenberg, line 50: I think this needs to be GeminiJsonSchemaDirty now

bracesproul (approved):
nice, thanks for handling this one!

PR #5251, opened 2024-04-29
azure-openai[patch]: fix only first embedding generated when batching enabled

sinedied:
When batching is enabled, only the first embedding for the batch was generated. This PR fixes that.


bracesproul (approved):
lgtm, I committed the suggestions I made fyi.

Will get this released soon!

Edit: Pushed up a commit running `yarn format`

PR #5268, opened 2024-05-02
scripts[minor]: Fix executable

bracesproul:


underwater:
Hi, I am getting a `TypeError [ERR_UNKNOWN_FILE_EXTENSION]" error ` which is breaking the build after this P/R.  
This happens in both windows and linux with or without the .dev container environment....
Is this a bug ?

underwater:
Just updated my node version from 20.9.0 to 20.13.1 and the build seems alright....
So 
`yarn build` breaks in langchain-core using node 20.9.0
`yarn build` is fine in langchain-core using node 20.13.1

TeCHiScy:
same here, Node.js v20.9.0 failed with error ERR_UNKNOWN_FILE_EXTENSION.

PR #5271, opened 2024-05-02
support alibaba_tongyi stream output

sinajia:
Fixes # (issue)


bracesproul (commented):
lgtm, could you add a test too? Thanks for doing this!

sinajia:
fix type error, add a test

libs/langchain-community/src/chat_models/alibaba_tongyi.ts
```diff
@@ -6,10 +7,12 @@ import {
   AIMessage,
   type BaseMessage,
   ChatMessage,
+  AIMessageChunk,
 } from "@langchain/core/messages";
 import { type ChatResult } from "@langchain/core/outputs";
-import { type CallbackManagerForLLMRun } from "@langchain/core/callbacks/manager";
```
jacoblee93, line 11: I think you still need this type?

jacoblee93:
Thank you! This will be released in about a week along with 0.2 - let me know if you need it cherry picked sooner.

PR #5274, opened 2024-05-02
Send document.metadata to AzureAISearch in addVectors()

keiohtani:
Fixes the issue with `document.metadata` not being sent to Azure AI Search when using `AzureAISearchVectorStore.fromDocuments()`. 
`AzureAISearchVectorStore.fromDocuments()` accepts `Document[]` as argument which has a parameter `document.metadata`, but internally this value is not passed to `AzureAISearchVectorStore.addVectors()` and it could be confusing to the users.

### DIFF
--- a/libs/langchain-community/src/vectorstores/azure_aisearch.ts
+++ b/libs/langchain-community/src/vectorstores/azure_aisearch.ts
@@ -314,6 +314,7 @@ export class AzureAISearchVectorStore extends VectorStore {
       metadata: {
         source: doc.metadata?.source,
         attributes: doc.metadata?.attributes ?? [],
+        ...doc.metadata,
       },
     }));
 


libs/langchain-community/src/vectorstores/azure_aisearch.ts
```diff
@@ -314,6 +314,7 @@ export class AzureAISearchVectorStore extends VectorStore {
       metadata: {
         source: doc.metadata?.source,
         attributes: doc.metadata?.attributes ?? [],
+        ...doc.metadata,
```
sinedied, line 317: This won't work and most probably triggers an error.
AFAIK Azure AI Search only supports a fixed schema (see index creation), hence the "workaround" of setting metadata as K/V pair in the generic `attributes` property.

This is something I plan to fix in a later PR by doing the K/V par transform automatically, and allowing to customize the index schema.

sinedied: If you don't provide `source` or `attributes` it should work anyways as there's a fallback, do you remember what kind of error you had? Metadata should not be mandatory, if there's an issue with empty metadata I'll be happy to fix it

keiohtani: Sure, 

This is the error message when a metadata field is defined in the index, 

```
}
Error: Azure AI Search uploadDocuments batch failed: RestError: The request is invalid. Details: Cannot find nested property 'metadata' on the resource type 'search.documentFields'.
    at EventEmitter.<anonymous> (/home/keiohtani/projects/chatifc-backend/node_modules/@langchain/community/dist/vectorstores/azure_aisearch.cjs:225:19)
    at EventEmitter.emit (node:events:514:28)
    at SearchIndexingBufferedSender.submitDocuments (/home/keiohtani/projects/chatifc-backend/node_modules/@azure/search-documents/src/searchIndexingBufferedSender.ts:463:22)
    at SearchIndexingBufferedSender.internalFlush (/home/keiohtani/projects/chatifc-backend/node_modules/@azure/search-documents/src/searchIndexingBufferedSender.ts:400:9)
    at AzureAISearchVectorStore.addVectors (/home/keiohtani/projects/chatifc-backend/node_modules/@langchain/community/dist/vectorstores/azure_aisearch.cjs:228:9)
    at AzureAISearchVectorStore.addDocuments (/home/keiohtani/projects/chatifc-backend/node_modules/@langchain/community/dist/vectorstores/azure_aisearch.cjs:201:25)
    at Function.fromDocuments (/home/keiohtani/projects/chatifc-backend/node_modules/@langchain/community/dist/vectorstores/azure_aisearch.cjs:576:9)
    at async Promise.all (index 0)
    at SummarizationChainService.createEmbeddings (/home/keiohtani/projects/chatifc-backend/src/services/LangChain/SummarizationChainService.ts:94:23)
```

keiohtani: And Azure AI Search does not allow an empty metadata field when creating from LangChain (in this case, we are using Python) 
image

keiohtani: Also, looking at the other method, `fromTexts`, The way `metadatas` is handled is confusing. 

```js
static async fromTexts(
    texts: string[],
    metadatas: object[] | object,
    config: AzureAISearchConfig
  ): Promise<AzureAISearchVectorStore> {
    const docs: Document<AzureAISearchDocumentMetadata>[] = [];
    for (let i = 0; i < texts.length; i += 1) {
      const metadata = Array.isArray(metadatas) ? metadatas[i] : metadatas;
      const newDoc = new Document({
        pageContent: texts[i],
        metadata,
      });
      docs.push(newDoc);
    }
    return AzureAISearchVectorStore.fromDocuments(docs, embeddings, config);
  }
```

At first glance, it seems this `metadatas` parameter will be used in the document creation, but because it's overwritten in the `addVectors()`, that's not the case and it's really hard to understand the behavior by just looking at the method itself.

keiohtani: @sinedied Following up on this

sinedied: For now, if you need metadata to be stored you need to use these two properties: https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-community/src/vectorstores/azure_aisearch.ts#L64-L65
- `source` and `attributes, the latter being able to store a collection of objects that have both a `key` and `value` attributes, see the schema here: https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-community/src/vectorstores/azure_aisearch.ts#L712-L723

It's not ideal I know, that's why I would like to propose a better implementation for this (the current one was based on the Python implementation). I'll probably work on it starting June, with the current plan:
1. Automatically take any metadata properties passed in and adapt them to fit in the current schema. That would solve your current issue, but it still would allow only string values.
2. Propose a way to customize the schema and the attributes used for storage.

keiohtani: This worked for me. Thank you.

keiohtani: Actually, it turns out my setup was incorrect and the metadata field was not correctly uploaded. Now adding the metadata field in the specified schema runs into this. 
```
Error: Azure AI Search uploadDocuments batch failed: RestError: The request is invalid. Details: An unexpected 'StartObject' node was found for property named '' when reading from the 
JSON reader. A 'PrimitiveValue' node was expected.
    at EventEmitter.<anonymous> (/home/keiohtani/projects/chatifc-backend/node_modules/@langchain/community/dist/vectorstores/azure_aisearch.cjs:225:19)
    at EventEmitter.emit (node:events:514:28)
    at SearchIndexingBufferedSender.submitDocuments (/home/keiohtani/projects/chatifc-backend/node_modules/@azure/search-documents/src/searchIndexingBufferedSender.ts:463:22)
    at SearchIndexingBufferedSender.internalFlush (/home/keiohtani/projects/chatifc-backend/node_modules/@azure/search-documents/src/searchIndexingBufferedSender.ts:400:9)
    at AzureAISearchVectorStore.addVectors (/home/keiohtani/projects/chatifc-backend/node_modules/@langchain/community/dist/vectorstores/azure_aisearch.cjs:228:9)
    at AzureAISearchVectorStore.addDocuments (/home/keiohtani/projects/chatifc-backend/node_modules/@langchain/community/dist/vectorstores/azure_aisearch.cjs:201:25)
    at Function.fromDocuments (/home/keiohtani/projects/chatifc-backend/node_modules/@langchain/community/dist/vectorstores/azure_aisearch.cjs:576:9)
[ERROR] 11:27:17 Error: Azure AI Search uploadDocuments batch failed: RestError: The request is invalid. Details: An unexpected 'StartObject' node was found for property named '' when 
reading from the JSON reader. A 'PrimitiveValue' node was expected.
```

This is what my mapped documents look like. 

```js
  createEmbeddings = async (
    documents: Document[],
    upn: string,
    conversationId: string
  ): Promise<void> => {
    const remappedDocuments: Document[] = documents.map((doc, index) => ({
      ...doc,
      metadata: {
        attributes: [
          {
            key: "conversationId",
            value: conversationId.toString(),
          },
          {
            key: "ts",
            value: doc.metadata.ts,
          },
          {
            key: "upn",
            value: upn,
          },
          {
            key: "order",
            value: index.toString(),
          },
        ],
      },
    }));
    AzureAISearchVectorStore.fromDocuments(remappedDocuments, this.embeddings, {
      indexName: this.#indexName,
      search: { type: AzureAISearchQueryType.Similarity },
      client: this.#searchClient,
    });
  };
```

Is this the right way to pass metadata? If so, `AzureAISearchVectorStore.fromDocuments` may not be creating metadata correctly.

bracesproul:
Closing this as stale. @keiohtani / @sinedied if you'd still like to get this merged please resolve the issues in the thread above & tag me when ready! Thank you.

kentmor:
I know I'm late to the party, and correct me if I'm wrong, but does this mean there's no way to get facets on metadata uploaded via these functions? You could do `facets: ["metadata/attributes/key", "metadata/attributes/value"],` but that's not really useful since you can't "pair" them. 

I've found several stackoverflow posts that mention [this](https://learn.microsoft.com/en-us/archive/blogs/onsearch/multi-level-taxonomy-facets-in-azure-search) as a work-around but I don't think Document[] object allow anything but pageContent and metadata.

PR #5287, opened 2024-05-05
cloudflare[minor]: use native Cloudflare Workers AI binding

Cherry:
As of 2024-04-11, there's no need to use the `@cloudflare/ai` package and it's deprecated:

https://developers.cloudflare.com/workers-ai/changelog/

This swaps the code over to use the native binding directly, and updates `@cloudflare/workers-types` which now ships the `Ai` type. No other code changes are needed.


libs/langchain-cloudflare/package.json
```diff
@@ -38,12 +38,11 @@
   "author": "LangChain",
   "license": "MIT",
   "dependencies": {
-    "@cloudflare/ai": "^1.0.47",
     "@langchain/core": "~0.1",
     "uuid": "^9.0.1"
   },
   "devDependencies": {
-    "@cloudflare/workers-types": "^4.20231218.0",
+    "@cloudflare/workers-types": "^4.20240502.0",
```
bracesproul, line 45: If we're replacing `@cloudflare/ai` with this, it needs to be moved out of dev deps and into standard deps

libs/langchain-cloudflare/src/embeddings.ts
```diff
@@ -14,7 +13,7 @@ type AiTextEmbeddingsOutput = {
 
 export interface CloudflareWorkersAIEmbeddingsParams extends EmbeddingsParams {
   /** Binding */
-  binding: Fetcher;
+  binding: Ai;
```
bracesproul, line 16: This will be a breaking change. Instead of this we should deprecate the `binding` input, and make it optional. With that we can add an additional (also optional) arg which is typed with the new `Ai` type.

bracesproul (changes requested):
Thanks for taking this on, I have a couple comments to prevent breaking changes.

libs/langchain-cloudflare/package.json
Cherry: `workers-types` only provides the types for the `Ai` - there's nothing runtime that would need this to be a dependency and not a devDependency`. The runtime for the AI binding is handled entirely in workerd via wrangler.

libs/langchain-cloudflare/src/embeddings.ts
Cherry: Hmm, good thoughts. The confusion here might come from the fact that the new AI binding is more of a binding than the old one (the old one is often referred to as a "wrapped binding" in the cf dev ecosystem), so I don't have a good name for this.

Do you have a suggestion for what this could be called? Or would it make sense to just cut a breaking change given how this is in `0.0.x` versions right now?

jacoblee93: Yes, let's make this a minor version.

Cherry: Makes sense to me!

jacoblee93: Would this also break people using this outside Cloudflare Workers?

We'll also want to update the docs but I can take care of that.

jacoblee93:
Hey so just got wind that the Rest API methods will be available in the main SDK:

https://www.npmjs.com/package/cloudflare

We should wait to land this until that is generally ready

libs/langchain-cloudflare/src/embeddings.ts
Cherry: This package is used outside of Cloudflare Workers? `Fetcher` is a purely Cloudflare Workers thing, in my experience - I had no idea this package was used externally?

Cherry:
The general `cloudflare` SDK makes sense outside of Workers, but bindings inside of Workers are faster and more secure - no additional authentication required, so using it inside a Worker for anything that has native bindings doesn't make a lot of sense in my experience.

jacoblee93:
Yeah for sure but ideally we don't lose functionality for people using it outside workers - I've done that myself in demos.

It should accept both a binding and a config object/client IMO

Cherry:
Agreed that would be ideal, but it doesn't look like it does today: https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-cloudflare/src/embeddings.ts. 

No functionality here will be lost, so perhaps that improvement would be better served in another PR?

Cherry:
As far as I can tell, this is ready to merge. Is there anything left to resolve before merging?

jacoblee93:
Apologies will have a look as soon as I can!

wanoo21:
Hey team, I have an issue that might be related to this!

image

I'm looking forward to this PR!

jacoblee93:
Hey, extremely sorry about the delay, this looks good!

libs/langchain-cloudflare/src/embeddings.ts
jacoblee93: You're right, fault was mine for looking quickly.

jacoblee93 (approved):

jacoblee93:
We are about to push minor versions of all packages related to the 0.3 release, will push this out alongside that.

PR #5302, opened 2024-05-07
qdrant[major]: Add `@langchain/qdrant` Package

Anush008:
## Description

This PR adds a `@langchain/qdrant` package based on the current community implementation.


bracesproul (approved):
Thanks for doing this! I pushed up some small commits with nits, but besides that lgtm!

PR #5307, opened 2024-05-07
scripts[patch]: Fix check broken links script bug

JuHyung-Son:
current check broken links script actually not working.
below code always executed and returned. so even if it has result.failedUrls, it never pushed

```js
if (typeof result.linksChecked === "number") {
        return acc + result.linksChecked;
      }
```


bracesproul (approved):
Very good catch! Thanks for pushing up a fix!

I also pushed up a commit fixing the eslint error

PR #5323, opened 2024-05-09
openai[minor]: Update OpenAI with Azure Specific Code

sarangan12:
Azure SDK Team has made changes to Open AI Node SDK in PR https://github.com/openai/openai-node/pull/718. In that PR, a new AzureOpenAI Client has been introduced. This enables the OpenAI SDK Users to access Azure Hosted Services with their credentials object instead of OpenAI Key.

This PR is to make changes to @langchain/openai package to make the corresponding changes so Langchain users also could access their Azure Hosted OpenAI service with the credentials object. Required test cases have already been added to the PR. 

@achandmsft @deyaaeldeen @xirzec FYI.....


bracesproul (commented):
Why wouldn't you just use the `@langchain/azure-openai` package?

sarangan12:
The `@langchain/azure-openai` package uses the `@azure/openai` SDK to access azure. Now, `openai` package itself has added features to support the same (independent of the `@azure/openai` package). This PR is to ensure that `@langhchain/openai` package takes advantage of this new feature provided by openai. 

It is possible that similar changes could happen in `@langchain/azure-openai` package (i.e. make sure it uses `openai` package instead of `@azure/openai` package). But, then there will be 2 packages (`@langchain/openai` and `@langchain/azure-openai` packages) and both depend/use the same `openai` package which will be redundant.

bracesproul:
> The `@langchain/azure-openai` package uses the `@azure/openai` SDK to access azure. Now, `openai` package itself has added features to support the same (independent of the `@azure/openai` package). This PR is to ensure that `@langhchain/openai` package takes advantage of this new feature provided by openai.
> 
> It is possible that similar changes could happen in `@langchain/azure-openai` package (i.e. make sure it uses `openai` package instead of `@azure/openai` package). But, then there will be 2 packages (`@langchain/openai` and `@langchain/azure-openai` packages) and both depend/use the same `openai` package which will be redundant.

But what are the features that would compel someone to use `@langchain/openai` to connect to Azure, instead of just using `@langchain/azure-openai`?

If it's just the ability to have one package installed, and use it to hit the OAI & Azure APIs, I don't think we want that since the packages are small enough already.

sarangan12:
@bracesproul Actually Yes. It is the reasoning that the users could just install one package (`@langchain/openai`) and get access to both OpenAI and Azure OpenAI and while doing this they could use the credentials (created in Azure) to access it. They do not have to use `@langchain/azure-openai` package at all. 

Now, if you prefer I could make the same change in `@langchain/azure-openai` package. But, then you will have 2 packages - both using the `openai` package. What would be your recommendation to such a proposal?

achandmsft:
@bracesproul, 
We (Microsoft) worked closely with OpenAI (and Stainless) and I also touched upon this strategy briefly with @jacoblee93 from LangChainJS a few weeks ago. Reiterating here: 

The goal here is to provide LangChainJS developers using OpenAI with the same quality of experience as LangChain (Python) developers. JavaScript developers want to seamlessly switch between OpenAI and Azure OpenAI endpoints using a single SDK without needing to change packages when talking to different OpenAI endpoints. All LangChain (Python) developers already enjoy this capability today without needing any Azure specific packages across the board: https://python.langchain.com/v0.1/docs/integrations/llms/azure_openai/#deployments. 

In order to enable this experience, we first developed an Azure OpenAI client natively in the OpenAI SDK for JavaScript in order to be on par with Python. We just shipped this support a few days ago (https://www.npmjs.com/package/openai). 

The problem with the additional azure specific package is not just the size, it's the fact that if there's a new feature in OpenAI that shows up in Azure OpenAI endpoint; they need to wait first for Azure OpenAI SDK for JavaScript to release a new version and then wait for the @langchain/azure-openai package to be updated to use this version. So they need to perform at least an additional package install/upgrade and if breaking, an additional code update. This is cumbersome (and error prone if there are versioning mismatches across these libraries as OpenAI APIs evolve rapidly). 

Our solution provides a simple streamlined developer experience around OpenAI for LangChainJS customers. Hope this helps clarify.

libs/langchain-openai/src/azure/embeddings.ts
```diff
@@ -0,0 +1,103 @@
...
+export class AzureOpenAIEmbeddings extends OpenAIEmbeddings {
+  constructor(
+    fields?: Partial<OpenAIEmbeddingsParams> &
+      Partial<AzureOpenAIInput> & {
+        verbose?: boolean;
+        /**
+         * The OpenAI API key to use.
+         * Alias for `apiKey`.
+         */
+        openAIApiKey?: string;
```
bracesproul, line 24: This param is only in other integrations for legacy purposes. We shouldn't include in new integrations
```suggestion
```

libs/langchain-openai/src/azure/embeddings.ts
```diff
@@ -0,0 +1,103 @@
...
+        /** The OpenAI API key to use. */
+        apiKey?: string;
+        configuration?: ClientOptions;
+        deploymentName?: string;
+        openAIApiVersion?: string;
+      },
+    configuration?: ClientOptions & LegacyOpenAIInput
+  ) {
+    const newFields = fields ? { ...fields } : fields;
+    if (newFields) {
```
bracesproul, line 34: nit
```suggestion
    const newFields = { ...fields };
    if (Object.entries(newFields).length) {
```

libs/langchain-openai/src/tests/azure/chat_models.int.test.ts
```diff
@@ -0,0 +1,829 @@
...
+import { AzureChatOpenAI } from "../../azure/chat_models.js";
+
+test("Test ChatOpenAI", async () => {
+  const chat = new AzureChatOpenAI({
+    modelName: "gpt-3.5-turbo",
+    maxTokens: 10,
+  });
+  const message = new HumanMessage("Hello!");
+  const res = await chat.call([message]);
+  console.log({ res });
```
bracesproul, line 33: Ideally these also `expect` something, so the test doesn't pass by default unless an error is thrown. E.g expect tokens > 5 or something

libs/langchain-openai/src/tests/azure/chat_models.int.test.ts
```diff
@@ -0,0 +1,829 @@
...
+import { NewTokenIndices } from "@langchain/core/callbacks/base";
+import { InMemoryCache } from "@langchain/core/caches";
+import { getEnvironmentVariable } from "@langchain/core/utils/env";
+import {
+  ClientSecretCredential,
+  getBearerTokenProvider,
+} from "@azure/identity";
+import { AzureChatOpenAI } from "../../azure/chat_models.js";
+
+test("Test ChatOpenAI", async () => {
```
bracesproul, line 26: All the test names need updating to be representative of what is being tested.

libs/langchain-openai/src/tests/azure/llms.int.test.ts
```diff
@@ -0,0 +1,333 @@
...
+import { AzureOpenAI } from "../../azure/llms.js";
+
+test("Test OpenAI", async () => {
+  const model = new AzureOpenAI({
+    maxTokens: 5,
+    modelName: "gpt-3.5-turbo-instruct",
+  });
+  const res = await model.invoke("Print hello world");
+  console.log({ res });
+});
```
bracesproul, line 20: Same here about test names and expecting something

libs/langchain-openai/src/types.ts
```diff
@@ -197,4 +197,6 @@ export declare interface AzureOpenAIInput {
    * will be result in the endpoint URL: https://westeurope.api.cognitive.microsoft.com/openai/deployments/{DeploymentName}/
    */
   azureOpenAIBasePath?: string;
+
+  azureADTokenProvider?: () => Promise<string>;
```
bracesproul, line 201: Could we have JSDoc here explaining what is it? I see the OpenAI SDK has some. We should at least copy that over.

bracesproul (changes requested):
Left a few comments. Good start, need to cleanup a couple nits. Thanks for working on this large PR!

Also, CI is failing, if you could resolve those errors that would be great!

libs/langchain-openai/src/tests/azure/chat_models.int.test.ts
sarangan12: Thanks for the feedback. I have updated all the test names representing what is being tested.

libs/langchain-openai/src/tests/azure/chat_models.int.test.ts
sarangan12: These tests are direct Azure equivalents of the ones in 1. [chat_models.int.test.ts](https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-openai/src/tests/chat_models.int.test.ts) 2. [embeddings.int.test.ts](https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-openai/src/tests/embeddings.int.test.ts) 3. [llms.int.test.ts](https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-openai/src/tests/llms.int.test.ts). Also, in other libs such as cloudflare, google-genai, etc. 

So, this follows the same pattern as these test cases.

libs/langchain-openai/src/tests/azure/llms.int.test.ts
sarangan12: I have updated the test names. Regarding expecting, I am using the same pattern as other libs. Kindly refer the explanation above. Thanks

sarangan12:
@bracesproul I have addressed the feedback. Could you please take a look at the PR? Thanks

jacoblee93:
I think consolidation is good - but what will become of the `@azure/openai` package then?

libs/langchain-openai/src/azure/llms.ts
```diff
@@ -27,15 +29,8 @@ export class AzureOpenAI extends OpenAI {
         configuration?: ClientOptions & LegacyOpenAIInput;
       }
   ) {
-    // assume the base URL does not contain "openai" nor "deployments" prefix
-    let basePath = fields?.openAIBasePath ?? "";
```
jacoblee93, line 31: Actually, this is breaking, isn't it?

I don't think this is heavily used yet and there are no examples in our docs, so I'm ok with a change now.

jacoblee93:
Thanks! As noted above, it's technically breaking but this code path doesn't seem to be in our docs yet. Let's try to shim things in the future though once it becomes more prominent.

PR #5339, opened 2024-05-13
feat[langsmith]: transparent handoff between @traceable HOF and LangChain

dqbd:
- Upgrade to newest LangSmith 0.1.30
- Transparent handoff between `@traceable` and LangChain


hinthornw (approved):

langchain-core/src/runnables/base.ts
```diff
@@ -2081,14 +2171,42 @@ export class RunnableLambda<RunInput, RunOutput> extends Runnable<
...
-    func: RunnableFunc<RunInput, RunOutput | Runnable<RunInput, RunOutput>>;
+    func:
+      | RunnableFunc<RunInput, RunOutput | Runnable<RunInput, RunOutput>>
+      | TraceableFunction<
+          RunnableFunc<RunInput, RunOutput | Runnable<RunInput, RunOutput>>
+        >;
   }) {
+    if (isTraceableFunction(fields.func)) {
+      // eslint-disable-next-line no-constructor-return
+      return RunnableTraceable.from(fields.func) as unknown as RunnableLambda<
```
jacoblee93, line 2182: Could/should `RunnableTraceable` extend `RunnableLambda`?

langchain/src/smith/runner_utils.ts
```diff
@@ -25,7 +25,7 @@ import {
 } from "langsmith";
 import { EvaluationResult, RunEvaluator } from "langsmith/evaluation";
 import { DataType } from "langsmith/schemas";
-import type { TraceableFunction } from "langsmith/traceable";
+import type { TraceableFunction } from "langsmith/singletons/traceable";
```
jacoblee93, line 28: No worries if it's already done, but seems like an unnecessary breaking change - why move this?

langchain-core/src/runnables/base.ts
dqbd: Resolved from Slack: probably, but wanted to get rid of weird warnings / errors when extending from a class not yet declared

langchain/src/smith/runner_utils.ts
```diff
@@ -165,7 +165,7 @@ class CallbackManagerRunTree extends RunTree {
     this.callbackManager = callbackManager;
   }
 
-  async createChild(config: RunTreeConfig): Promise<CallbackManagerRunTree> {
+  createChild(config: RunTreeConfig): CallbackManagerRunTree {
```
agola11, line 168: is this a breaking change?

jacoblee93: In theory if you were doing some really strange typing things or using `.then()`, but `await` (should by far be the most common way people use this) on a non async function still works so in practice people are going to be fine.

Going the other way (sync to async) would definitely be a breaking change though - we haven't cut a release yet so @dqbd are we sure we'd never want to do async operations here?

PR #5342, opened 2024-05-13
core[minor]: RunnableLambda should consume (async) iterator if the wrapped function returns one

nfcampos:
Fixes # (issue)

### DIFF
--- a/langchain-core/src/runnables/base.ts
+++ b/langchain-core/src/runnables/base.ts
@@ -2111,7 +2111,8 @@ function isIterator(thing: unknown): thing is Generator {
   return (
     typeof thing === "object" &&
     thing !== null &&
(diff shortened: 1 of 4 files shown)


langchain-core/src/runnables/base.ts
```diff
@@ -2068,6 +2068,10 @@ export class RunnableLambda<RunInput, RunOutput> extends Runnable<
       for await (const chunk of stream) {
         yield chunk;
       }
+    } else if (isAsyncIterator(output)) {
```
dqbd, line 2071: Do we want to consume generators as well?

nfcampos: yep good point, added

dqbd (approved):

jacoblee93:
Thank you!

PR #5347, opened 2024-05-13
langchain[patch]: Remove outdated test

jacoblee93:
@bracesproul


bracesproul (approved):
lgtm, only other thing is I think we can also remove the code inside gh actions which passes `COHERE_API_KEY` through the envs?

PR #5365, opened 2024-05-14
community[patch]: #3369 Streaming support for Replicate models

jeasonnow:
For #3369

[replicate.stream](https://github.com/replicate/replicate-javascript?tab=readme-ov-file#replicatestream)

The latest version of `replicate` now supports a `stream` method for streaming output. Based on this new version, we will add this feature to `langchainjs`.


jacoblee93:
Thank you! Looks great, I added an integration test

PR #5381, opened 2024-05-15
community[patch]: add ?| (arrayContains) filter on metadata to PGVector search

georgeherby:
I have metadata that is an array, for example, I have tags against a file, `tags: [tag1, tag2]` and I need to be able to filter on them. However, I am going to be providing an array as a filter and then find records where the tag item contains any of the tags in the filter.

filter: {
 tags: ['tag1', 'tag2']
}

This would get any records that have tag1 or tag2 in their array, even if it also has tag3 or tag4, so long as it matches one return it.

Note: I have not been able to run the integration tests as I don't have OpenAI keys (not sure if the pipelines will validate these). I have tested manually.

Twitter -> georgeherby_dev


jacoblee93:
Nice - looks reasonable to me, will try it myself! Thanks for your patience!

libs/langchain-community/src/vectorstores/pgvector.ts
```diff
@@ -501,6 +501,15 @@ export class PGVectorStore extends VectorStore {
           parameters.push(..._value.in);
           paramCount += _value.in.length;
         }
+        if (_value.arrayContains) {
+          whereClauses.push(
+            `${
+              this.metadataColumnName
+            }->'${key}' ?| array[${_value.arrayContains
```
jacoblee93, line 508: Ah wait, actually this has some SQL injection risk - can you parameterize?

georgeherby: I have addressed this by aligning the code to match how it's done in "IN" filters.

jacoblee93:
Thank you!

PR #5412, opened 2024-05-16
community[minor]: updated Browserbase loader

mishushakov:
Updated browserbase loader and docs


jacoblee93:
Thanks for waiting! We've moved all document loaders to `@langchain/community`, but this one somehow missed the cut. I've moved it there as part of this PR.

PR #5413, opened 2024-05-16
community[minor]: add Neo4j Graph enhancedSchema option

easwee:
Porting enhanced_schema option from Python lib.


libs/langchain-community/src/graphs/neo4j_graph.ts
```diff
@@ -211,36 +361,201 @@ export class Neo4jGraph {
...
+        outputDict[propName] = `{${returnClauses.pop()}}`;
+      }
+    }
+
+    const withClause = `WITH ${withClauses.join(", ")}`;
+    const returnClause = `RETURN {${Object.entries(outputDict)
+      .map(([k, v]) => `\`${k}\`: ${v}`)
+      .join(", ")}} AS output`;
+
+    const cypherQuery = [matchClause, withClause, returnClause].join("\n");
```
jacoblee93, line 556: Triple checking that there's no injection risk here - has been a while since I've looked at Neo4j

jacoblee93:
One question, and can you resolve merge conflicts please?

libs/langchain-community/src/graphs/neo4j_graph.ts
tomasonjo: Yeah... actually user doesn't have any parameters they could inject... we just sample the database values with enhanced schema. But we try and escape most values anyway as a general guideline.

easwee:
> One question, and can you resolve merge conflicts please?

Sure thing - done.

jacoblee93:
Thank you!

PR #5417, opened 2024-05-16
v0.1-anthropic[minor]: Add tool choice code & docs

bracesproul:


HenryHengZJ:
@bracesproul now that anthropic support streaming on tools as well, should we disable the check at line 619 `if (options.tools !== undefined && options.tools.length > 0)` ?

PR #5423, opened 2024-05-16
docs[minor]: Log actual prompt when using prompt hub

bracesproul:
everywhere `await pull...` is used I added extra cells logging the actual value of the prompt.

Also some link/bug fixes


docs/core_docs/docs/how_to/qa_chat_history_how_to.ipynb
```diff
@@ -108,6 +109,33 @@
         "});"
       ]
     },
+    {
+      "cell_type": "markdown",
+      "metadata": {},
+      "source": [
+        "Lets see what this prompt actually looks like"
```
jacoblee93, line 116: Missing apostrophe, should be `Let's`

PR #5426, opened 2024-05-16
@langchain/google-genai [feature]: support Google genai API version and base URL

sugarforever:
Description

Some Gemini preview models for example `gemini-1.5-flash-latest` is only supported by API version `v1beta`. The current @langchain/google-genai package doesn't expose options for API version.

This PR is to allow developers to specify API version and base URL by `apiVersion` and `baseUrl` when using Google Generative AI models.


bracesproul (approved):
nice, i like it!

will merge once CI passes

PR #5427, opened 2024-05-16
core[minor]: Unified model params for LS

bracesproul:
openai[minor]
anthropic[minor]
community[minor]
cohere[minor]
mistral[minor]

### DIFF
--- a/langchain-core/src/language_models/chat_models.ts
+++ b/langchain-core/src/language_models/chat_models.ts
@@ -196,7 +196,7 @@ export abstract class BaseChatModel<
         this._separateRunnableConfigFromCallOptions(options);
       const inheritableMetadata = {
         ...this.metadata,
-        ...this._getLsParams(callOptions),
+        ...this.getLsParams(callOptions),
       };
       const callbackManager_ = await CallbackManager.configure(
         runnableConfig.callbacks,
@@ -259,7 +259,7 @@ export abstract class BaseChatModel<
     }
   }
 
-  protected _getLsParams(options: this["ParsedCallOptions"]): LangSmithParams {
+  protected getLsParams(options: this["ParsedCallOptions"]): LangSmithParams {
     return {
       ls_model_type: "chat",
       ls_stop: options.stop,
@@ -278,7 +278,7 @@ export abstract class BaseChatModel<
 
     const inheritableMetadata = {
       ...this.metadata,
-      ...this._getLsParams(parsedOptions),
+      ...this.getLsParams(parsedOptions),
     };
 
     // create callback manager and start run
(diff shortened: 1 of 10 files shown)


langchain-core/src/language_models/chat_models.ts
```diff
@@ -359,7 +388,7 @@ export abstract class BaseChatModel<
       handledOptions.tags,
       this.tags,
       handledOptions.metadata,
-      this.metadata,
+      inheritableMetadata,
```
nfcampos, line 391: this is the non inheritable metadata I think?

nfcampos: and in py I think we put it as inheritable

bracesproul: You are correct. Fixed in recent commit, thanks

libs/langchain-cohere/src/chat_models.ts
```diff
@@ -145,6 +146,15 @@ export class ChatCohere<
     this.streaming = fields?.streaming ?? this.streaming;
   }
 
+  protected getLsParams(_: this["ParsedCallOptions"]): LangSmithParams {
+    return {
+      ls_provider: "cohere",
+      ls_model_name: this.model,
+      ls_model_type: "chat",
+      ls_temperature: this.temperature ?? undefined,
```
nfcampos, line 154: this seems to be missing a bunch of params like max tokens, stop, etc see https://docs.cohere.com/reference/chat

libs/langchain-community/src/chat_models/ollama.ts
```diff
@@ -176,6 +177,16 @@ export class ChatOllama
     this.format = fields.format;
   }
 
+  protected getLsParams(_: this["ParsedCallOptions"]): LangSmithParams {
+    return {
+      ls_provider: "ollama",
+      ls_model_name: this.model,
+      ls_model_type: "chat",
+      ls_temperature: this.temperature ?? undefined,
```
nfcampos, line 185: no max tokens?

libs/langchain-mistralai/src/chat_models.ts
```diff
@@ -417,6 +418,17 @@ export class ChatMistralAI<
...
   }
 
+  protected getLsParams(options: this["ParsedCallOptions"]): LangSmithParams {
+    const params = this.invocationParams(options);
+    return {
+      ls_provider: "mistral",
+      ls_model_name: this.model,
+      ls_model_type: "chat",
+      ls_temperature: params.temperature ?? undefined,
+      ls_max_tokens: params.maxTokens ?? undefined,
```
nfcampos, line 428: no stop seq?

bracesproul: correct. no support

PR #5441, opened 2024-05-17
core[minor]: Add ID

hinthornw:
To BaseMessage. Similar to python https://github.com/langchain-ai/langchain/blob/e3f30b4cdebf750175e1764c969e6d56921c865d/libs/core/langchain_core/messages/base.py#L36


langchain-core/src/messages/base.ts
```diff
@@ -195,7 +208,12 @@ export abstract class BaseMessage
       // eslint-disable-next-line no-param-reassign
       fields.response_metadata = {};
     }
+    if (!fields.id) {
+      // eslint-disable-next-line no-param-reassign
+      fields.id = undefined;
```
jacoblee93, line 213: This seems unnecessary?

hinthornw: This is necessary.  Otherwise ids added  via mutation won' tbe serialized

bracesproul:
Implemented in https://github.com/langchain-ai/langchainjs/pull/5817

PR #5443, opened 2024-05-17
core[patch]: Release 0.2.0 (next)

bracesproul:


langchain-core/package.json
firminovctr, line 1: I am getting this error now. Any ideas why?

PR #5456, opened 2024-05-18
community[patch]: Update voyage.ts to pass input_type  (default null, or query or document)

nicolas-geysse:
Fixes # (issue)
Improve embedding quality


libs/langchain-community/src/embeddings/voyage.ts
```diff
@@ -53,6 +58,7 @@ export class VoyageEmbeddings
 
   headers?: Record<string, string>;
 
+  inputType: string;
```
jacoblee93, line 61: I think this needs a `?` to mark as optional?

jacoblee93:
Thank you!

nicolas-geysse:
You're very welcome !
I just realized that I may have forgotten to commit here
libs/langchain-community/src/embeddings/tests/voyage.int.test.ts
and here
docs/core_docs/docs/integrations/text_embedding/voyageai.mdx
I just commit for this new files, but as I'm a total noob with Github, I don't know where has been pushed these new modifications. (can you see them ???)

jacoblee93:
> You're very welcome ! I just realized that I may have forgotten to commit here libs/langchain-community/src/embeddings/tests/voyage.int.test.ts and here docs/core_docs/docs/integrations/text_embedding/voyageai.mdx I just commit for this new files, but as I'm a total noob with Github, I don't know where has been pushed these new modifications. (can you see them ???)

Ah, no that doesn't seem to be on this PR. It seems reasonable but feel free to push if you'd like!

libs/langchain-community/src/embeddings/voyage.ts
nicolas-geysse: yes you absolutely right, I should have secured with   
inputType?: string; // Make inputType optional

In the VoyageEmbeddings Class

PR #5476, opened 2024-05-20
community[minor]: Add support for Tencent Hunyuan Chat Model and Embeddings

TeCHiScy:
* Tencent provides its Hunyuan Chat model and Embeddings through Tencent cloud.
* Langchain-python has implemented it, as described in [doc](https://python.langchain.com/v0.2/docs/integrations/chat/tencent_hunyuan/).
* This PR adds support in **langchain-js** for the Hunyuan Chat Model ([doc](https://cloud.tencent.com/document/product/1729/105701)), and Embedding ([doc](https://cloud.tencent.com/document/product/1729/102832)).


examples/src/models/chat/integration_tencent_hunyuan.ts
```diff
@@ -0,0 +1,68 @@
...
+  model: "hunyuan-lite",
+  streaming: true,
+  temperature: 1,
+});
+
+res = await hunyuanLL.invoke(messages);
+console.log(res);
+/*
+AIMessage {
+  content: '您好！我是一个AI语言模型HunYuan,由腾讯开发和提供支持.您可以通过输入您的问题或请求来获取帮助和建议.',
```
jacoblee93, line 59: Out of curiosity - does the LLM respond in Chinese when prompted in English?

libs/langchain-community/src/chat_models/tencent_hunyuan.ts
```diff
@@ -0,0 +1,495 @@
...
+          totalTokens: TotalTokens,
+          promptTokens: PromptTokens,
+          completionTokens: CompletionTokens,
+        },
+      },
+    };
+  }
+
+  /** @ignore */
+  async completionWithRetry(
```
jacoblee93, line 407: nit: we generally prefer these streaming implementations to be generators, but if it works, that's fine

libs/langchain-community/src/utils/tencent_hunyuan.ts
```diff
@@ -0,0 +1,56 @@
+import { createHash, createHmac, BinaryLike } from "crypto";
+
+const host = "hunyuan.tencentcloudapi.com";
```
jacoblee93, line 3: Should we make this configurable?

jacoblee93:
Looks solid overall, see comments about `crypto` usage.

examples/src/models/chat/integration_tencent_hunyuan.ts
TeCHiScy: It depends on model.
hunyuan-lite is a lightweight model which can understand english prompt, but responses almostly in Chinese. hunyuan-standard and hunyuan-pro can response both in Chinese/English based on the prompt.

libs/langchain-community/src/chat_models/tencent_hunyuan.ts
TeCHiScy: I have refactor the stream part of the chat model, with `_streamResponseChunks` being implemented, which returns `AsyncGenerator<ChatGenerationChunk>`. In `_generate`, it calls `_streamResponseChunks` when streaming is true, otherwise it calls `completionWithRetry`.

libs/langchain-community/src/utils/tencent_hunyuan.ts
TeCHiScy: > Should we make this configurable?

I made `host` configurable in the newest commit. Generally, users do not need to care or change the host, since hunyuan only has one `host` in Tencent Cloud, and is stable. However, consider the needs for debug in development/testing environment, making host configurable seems reasonable.

TeCHiScy: > > This will not work in all environments - you'll need to import from `node:crypto` or [SubtleCrypto](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto) depending on what environments you care about, then mark these entrypoints as `requiresOptionalDependencies` in `langchain.config.js`
> 
> I'd like to make this model usable either in node/browser environment. Is there some example that detectes the environment and import the appropriate package to use?

Problem (may) solved. 

The `crypto` package is used to sign the request for Tencent Cloud. In the newest commit, I provided two version of sign in `utils/tencent_hunyuan/index.ts` and `utils/tencent_hunyuan/web.ts`, with the former one using nodejs's `crypto` package and the latter one using `crypto-js` package (which supports browser environment). Users in different environment may import different version of `sign` method, and pass it to the model class. I updated documents and examples to illustrate this.

Do you think this is an acceptable way to suuport both nodejs and browser.

libs/langchain-community/langchain.config.js
```diff
@@ -288,6 +290,8 @@ export const config = {
     "utils/convex": "utils/convex",
     "utils/event_source_parse": "utils/event_source_parse",
     "utils/cassandra": "utils/cassandra",
+    "utils/tencent_hunyuan/web": "utils/tencent_hunyuan/web",
+    "utils/tencent_hunyuan/index": "utils/tencent_hunyuan/index",
```
jacoblee93, line 294: Remove the first `/index`:

`"utils/tencent_hunyuan": "utils/tencent_hunyuan/index",`

examples/src/models/chat/integration_tencent_hunyuan.ts
```diff
@@ -0,0 +1,115 @@
+import { ChatTencentHunyuan } from "@langchain/community/chat_models/tencent_hunyuan";
+import { HumanMessage } from "@langchain/core/messages";
+import type { LLMResult } from "@langchain/core/outputs";
+
+// in nodejs environment
+import { sign } from "@langchain/community/utils/tencent_hunyuan/index";
+
+// in browser environment
+// import { sign } from "@langchain/community/utils/tencent_hunyuan/web";
```
jacoblee93, line 9: Up to you, but could also just make separate chat model and embedding entrypoints for web vs. Node

libs/langchain-community/src/utils/tencent_hunyuan.ts
jacoblee93: Yes this is ok, it may be more intuitive to make `chat_models/tencent/web` and `chat_models/tencent` instead of this `utils` package. See:

https://github.com/langchain-ai/langchainjs/tree/main/libs/langchain-community/src/chat_models/bedrock

examples/src/models/chat/integration_tencent_hunyuan.ts
jacoblee93: For example: https://github.com/langchain-ai/langchainjs/tree/main/libs/langchain-community/src/chat_models/bedrock

TeCHiScy: Hi @jacoblee93 , I have adjusted according to the bedrock example, making separate entrypoints for web & node. In the newest commit, `sign` method is not exported, users only need to use the right entrypoint. Meanwhile, I have adjusted the entrypoint name for node from `/chat_models/tencent_hunyuan/index` to `/chat_models/tencent_hunyuan` to make it more intuitive, the entrypoint for web changes to `/chat_models/tencent_hunyuan/web` accordingly.

jacoblee93:
Thank you!

PR #5480, opened 2024-05-20
google-genai [patch]: set apiVersion

stuntnmore:
Fixes # (issue)
add setting api version on google gen ai, so it can support v1 or v1beta which has more models available

### DIFF
--- a/libs/langchain-google-genai/src/chat_models.ts
+++ b/libs/langchain-google-genai/src/chat_models.ts
@@ -191,7 +191,7 @@ export class ChatGoogleGenerativeAI
 
   apiKey?: string;
 
-  apiVersion?: string = "v1";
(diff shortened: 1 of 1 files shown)


afirstenberg:
@stuntnmore - I can't do this, but I suggest changing the name to something like:
google-genai [patch]: set apiVersion

afirstenberg:
Yeah, it looks like https://github.com/langchain-ai/langchainjs/pull/5426 added the `apiVesion` field, and then set it to "v1" by default which blocked the new models. Which undid what #5115 did (which was to use the latest version of the library, which used "v1beta" by default.)

LGTM

stuntnmore:
> @stuntnmore - I can't do this, but I suggest changing the name to something like: google-genai [patch]: set apiVersion

name changes :)

PR #5483, opened 2024-05-20
fix(google-genai): use correct base url and set correct constructor props

### LINKED ISSUE #5437: Endpoint issue with Gemini Pro 1.5 and Gemini Flash Support
### Checked other resources

### Example Code

the version should be able to be set from configuration, right now google apis have 2 versions, one called v1, and another called v1beta, the difference is on v1 some of the newest gemini models dont exist on it but on v1beta it does, so being able to set this would allow for more model support of newer generation gemini releases

_No response_

### Description

add the ability to set the google generative api version through config and if not set then default back to v1

### System Info

langchain + google community libs

yharaskrik:
Fixes #5437

### DIFF
--- a/libs/langchain-google-genai/src/chat_models.ts
+++ b/libs/langchain-google-genai/src/chat_models.ts
@@ -193,7 +193,7 @@ export class ChatGoogleGenerativeAI
 
   apiVersion?: string = "v1";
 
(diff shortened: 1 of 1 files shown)


bracesproul:
This was fixed by #5490

yharaskrik:
> This was fixed by #5490

Yup, yours wasn't open when I opened mine.

PR #5498, opened 2024-05-21
scripts[minor]: Fix extra newlines issue, switch all packages to use workspace dep

bracesproul:


libs/langchain-scripts/bin/build
```diff
@@ -601,7 +601,8 @@ const generateImportMap = ({
     "\n"
   )}\n${extraDeclarations.join("\n")}\n`;
 
-  if (!extraContent.trim()) {
+  extraContent.trim();
+  if (!/[a-zA-Z0-9]/.test(extraContent)) {
```
bracesproul, line 605: the `!extraContent` check wasn't working as intended since the variable still had two newlines in it

PR #5510, opened 2024-05-22
community[patch]: Default embeddingNodeProperty value

adam-cowley:
If the `embeddingNodeProperty` is not set in he config when calling `Neo4jVectorStore.fromExistingGraph()`, the property will be named `undefined`.


jacoblee93:
Thank you!

jacoblee93:
Thank you!

PR #5535, opened 2024-05-24
community[patch]: include metadata returned by Amazon bedrock knowledge base

madmed88:
Bedrock knowledge bases now support adding metadata to source documents.
https://docs.aws.amazon.com/bedrock/latest/userguide/knowledge-base-ds.html#kb-ds-metadata
The current retriever does not return those values.
This pull request fixes that.
fixes #4928


jacoblee93:
Thank you!

jacoblee93:
I can't see any docs on how this field is returned, so will assume this is correct - it shouldn't cause any issues if it is wrong anyway.

jacoblee93:
Based on the types in the updated SDK, looks good. Thank you!

PR #5538, opened 2024-05-24
scripts[major]: New build

bracesproul:
Improvements:
- `yarn lint`: 128.8s  -> 45.8s
- `yarn build --filter=examples`: 361.94s -> 25.04s (aka build all integration/core pkgs)
- Turbo is always used when running `yarn build` in a package. Dependencies are also always built & cached first when running `yarn build`
  - you can still run without turbo via `yarn build:internal`
- `yarn clean` now only cleans the `.turbo` and `dist/` folders, and does not do any generation like before
- Entire build process (with `tsc`, generating & writing files, etc etc) now happens inside the `@langchain/build` `lc-build:v2` script, instead of 5+ individual scripts


libs/langchain-scripts/tsconfig.build.json
bracesproul, line 1: This is used to generate the build script ONLY, then the built build script is used for the remainder of the build in this package

PR #5366, opened 2024-05-14
community[minor]: feat: Add Zep Cloud community components, update docs and examples

paul-paliychuk:
We have recently released our new `zep-cloud` sdks that are compatible with Zep Cloud (not Zep Open Source). 

We have also maintained our Cloud version of langchain components (ChatMessageHistory, VectorStore) as part of our sdks. 

This PRs goal is to port these components to langchain community repo, and close the gap with the existing Zep Open Source components already present in community repo (added `ZepCloudMemory`,`ZepCloudVectorStore`,`ZepCloudRetriever`). 
Also added a `ZepCloudChatMessageHistory` components together with some expression language examples ported from our repo. 

Also includes the documentation update - added pages for Zep Cloud components.

We have left the original open source components intact on purpose as to not introduce any breaking changes.

PR #5542, opened 2024-05-25
community[minor]: Integrate zep cloud components

jacoblee93:
Extends #5366 with updated examples


paul-paliychuk:
@jacoblee93 thank you!

PR #5549, opened 2024-05-26
fix(community): Throw more descriptive errors for Ali toyi chat models

sinajia:
Throw informative exception


libs/langchain-community/src/chat_models/alibaba_tongyi.ts
```diff
@@ -593,7 +603,7 @@ export class ChatAlibabaTongyi
           continue;
         }
         try {
-          yield JSON.parse(line.slice("data:".length).trim());
+          yield JSON.parse(line.slice("data:".length));
```
jacoblee93, line 606: Why remove the `trim` call?

libs/langchain-community/src/chat_models/alibaba_tongyi.ts
```diff
@@ -526,6 +526,16 @@ export class ChatAlibabaTongyi
...
 
     for await (const chunk of stream) {
+      /* if some error occurs:
+         {
+          "code": "DataInspectionFailed",
+          "message": "Output data may contain inappropriate content.",
+          "request_id": "43d18007-5aa5-9d18-b3b3-a55aba9ce8cb"
+        }
+      */
+      if (!chunk.output) {
```
jacoblee93, line 536: Is it safer to check for something like `chunk.code`?

benjamincburns:
```suggestion
      if (!chunk.output && chunk.code) {
```

benjamincburns (approved):

benjamincburns:
Thanks for the contribution @sinajia!

PR #5551, opened 2024-05-27
community[patch]: fix data.usage error because Moonshot sometimes does not return

crazyurus:
Moonshot does not return `data.usage` when `stream` is `true`. So we need to add a default value to avoid errors.


jacoblee93:
Thank you!

PR #5557, opened 2024-05-27
community[patch]: Upstash Vector Store Namespace Feature

fahreddinozcan:
This PR adds namespace partition feature to Upstash Vector Store, which allows users to partition their database without the need for seperate databases or metadata filtering.


jacoblee93:
Looks reasonable - can you run `yarn format` from root?

jacoblee93:
Thank you!

PR #5559, opened 2024-05-27
mongodb[minor]: add, implement delete method

telekosmos:
Method `delete` was missing in `MongoDBAtlasVectorSearch` class and it's needed if `incremental` _indexing_ strategy is used.

Be aware of the `// eslint-disable-line @typescript-eslint/no-explicit-any` as the `$in` operator needs explicitly an `ObjectID` type (as discussed [here](https://www.mongodb.com/community/forums/t/cant-use-string-ids-with-typescript/262793)) instead allowing an _union type_ `string | ObjectId`. The ids coming in are `uuid v4` ids, which don't match `ObjectID` type but it works just fine as a mongodb `_id`.

### DIFF
--- a/libs/langchain-mongodb/src/vectorstores.ts
+++ b/libs/langchain-mongodb/src/vectorstores.ts
@@ -4,6 +4,7 @@ import {
   VectorStore,
 } from "@langchain/core/vectorstores";
 import type { EmbeddingsInterface } from "@langchain/core/embeddings";
+import { chunkArray } from "@langchain/core/utils/chunk_array";
 import { Document } from "@langchain/core/documents";
 import { maximalMarginalRelevance } from "@langchain/core/utils/math";
 import {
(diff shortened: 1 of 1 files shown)


libs/langchain-mongodb/src/vectorstores.ts
```diff
@@ -256,6 +256,22 @@ export class MongoDBAtlasVectorSearch extends VectorStore {
...
   }
 
+  /**
+   * Delete documents from the collection
+   * @param ids - An array of document IDs to be deleted from the collection.
+   *
+   * @returns - A promise that resolves when all documents deleted
+   */
+  async delete(params: { ids: any[] }): Promise<void> { // eslint-disable-line @typescript-eslint/no-explicit-any
+    const splice = <T>(l: T[], c: number): T[][] =>
```
jacoblee93, line 266: Can we use the built-in `chunkArray` method instead?

https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-openai/src/embeddings.ts#L211

jacoblee93:
Looks good! See comment, also please run `yarn format`.

jacoblee93:
Thank you!

PR #5564, opened 2024-05-28
community[patch]: Add pgvector index using HNSW

jl4nz:
Add support to optionally create a HNSW index on for pgvector


jl4nz:
Adding this for reference https://python.langchain.com/v0.1/docs/integrations/vectorstores/pgembedding/#create-hnsw-index

libs/langchain-community/src/vectorstores/pgvector.ts
```diff
@@ -677,4 +677,55 @@ export class PGVectorStore extends VectorStore {
...
+      case "euclidean":
+        idxDistanceFunction = "vector_l2_ops";
+        break;
+      default:
+        throw new Error(`Unknown distance strategy: ${this.distanceStrategy}`);
+    }
+
+    const createIndexQuery = `CREATE INDEX IF NOT EXISTS ${
+      this.computedTableName
+    }_embedding_idx
```
jacoblee93, line 714: Should this name be differentiated based on the number of dimensions and/or column?

Could maybe see a case where someone has multiple embedding fields in a given table?

jacoblee93:
Very cool! CC @eyurtsev for a look as well.

libs/langchain-community/src/vectorstores/pgvector.ts
jl4nz: > Should this name be differentiated based on the number of dimensions and/or column?
As far as I can see, there's only support to create one embedding field per table. I'll change the index name to use the `vectorColumnName` that's computed in the class for consistency.

> Could maybe see a case where someone has multiple embedding fields in a given table?
Haven't seen this pattern yet... I guess is possible, but IMO its easier to handle one vector per table + metadata.

jacoblee93:
Hey apologies for the delay - @eyurtsev is on vacation but I'd really like him to take a look. Might be a few more days.

libs/langchain-community/src/vectorstores/pgvector.ts
```diff
@@ -677,4 +677,55 @@ export class PGVectorStore extends VectorStore {
...
+   * Method to create the HNSW index on the vector column.
+   *
+   * @param dims - Defines the number of dimensions in your vector data, max: 2000. For example, use 1536 for OpenAI's text-embedding-ada-002 model and 1024 for amazon.titan-embed-text-v2:0
+   * @param m - The max number of connections per layer (16 by default)
+   * @param efConstruction -  The size of the dynamic candidate list for constructing the graph (64 by default)
+   * @param distanceFunction -  The distance function name you want to use, is automatically selected based on the distanceStrategy.
+   * @returns Promise that resolves with the query response of creating the index.
+   */
+  async createHnswIndex(config?: {
+    dims?: number;
```
eyurtsev, line 691: Should we require the dimension instead of assigning a default? Users likely use lots of different embedding models?

libs/langchain-community/src/vectorstores/pgvector.ts
```diff
@@ -677,4 +677,55 @@ export class PGVectorStore extends VectorStore {
     this.client?.release();
     return this.pool.end();
   }
+
+  /**
+   * Method to create the HNSW index on the vector column.
+   *
+   * @param dims - Defines the number of dimensions in your vector data, max: 2000. For example, use 1536 for OpenAI's text-embedding-ada-002 model and 1024 for amazon.titan-embed-text-v2:0
+   * @param m - The max number of connections per layer (16 by default)
```
eyurtsev, line 685: Any reference material we could link to that would help someone figure out how to choose values for m?

libs/langchain-community/src/vectorstores/pgvector.ts
```diff
@@ -677,4 +677,55 @@ export class PGVectorStore extends VectorStore {
...
+      default:
+        throw new Error(`Unknown distance strategy: ${this.distanceStrategy}`);
+    }
+
+    const createIndexQuery = `CREATE INDEX IF NOT EXISTS ${
+      this.vectorColumnName
+    }_embedding_idx
+        ON ${this.computedTableName} USING hnsw ((${
+      this.vectorColumnName
+    }::vector(${config?.dims || 1536})) ${idxDistanceFunction})
```
eyurtsev, line 717: Do we have any utilities to sanitize interpolated values? This code should be only exposed to developers, but if we have existing sanitization code won't hurt to add some more defensive code

libs/langchain-community/src/vectorstores/pgvector.ts
eyurtsev: I think this works for now

libs/langchain-community/src/vectorstores/pgvector.ts
jl4nz: I added a reference to the paper in the docs (pgvector.mdx) and added a bit more context in the comments for m and efConstruction parameters

Looks like this:

More info at the [`Pgvector GitHub project`](https://github.com/pgvector/pgvector?tab=readme-ov-file#hnsw) and the HNSW paper from Malkov Yu A. and Yashunin D. A.. 2020. [`Efficient and robust approximate nearest neighbor search using hierarchical navigable small world graphs`](https://arxiv.org/pdf/1603.09320)

libs/langchain-community/src/vectorstores/pgvector.ts
jl4nz: I refactored to be mandatory... I also think that assigning the value is less prone error.

libs/langchain-community/src/vectorstores/pgvector.ts
jl4nz: I've had the `pg-format` lib https://www.npmjs.com/package/pg-format to format the sql literals.
It should provide some guards for this.

jacoblee93: Can we actually remove this (for now)? It'll make existing users install a new package.

In the near-ish term, we should factor out into a `@langchain/pgvector` package that can have this as a hard dep.

jacoblee93:
Hey @jl4nz, can we revert that new `pg-format` package addition? Will break people who are using the current version. We can add a TODO to add it later.

jl4nz:
> Hey @jl4nz, can we revert that new `pg-format` package addition? Will break people who are using the current version. We can add a TODO to add it later.

Got it, done.

jl4nz:
Hey @jacoblee93, just pinging to check if this would be good to go. 
Thanks!

jacoblee93:
Yes, thank you!

eyurtsev (approved):

PR #5568, opened 2024-05-28
community[patch]: add support for or in elastic vector query

### LINKED ISSUE #5558: DOC: How to use or and operator with filtering metadata in a elastic vector store.
### Checklist

- [X] I included a link to the documentation page I am referring to (if applicable).

### Issue with current documentation:

The idea is to filter in as retriever function to get only subset of documents using operator or and. The self query in langchain does not seem supported with elastic vector. 

### Idea or request for content:

The as retriever should allow filter to take a function which just gets called over all documents. 

The documentation is like a treasure hunt right now to find. Please consider improving.

djaffer:
Fixes #5558.

const filter = [
  { field: "genre", operator: "or", value: "Drama" },
  { field: "genre", operator: "or", value: "Action" },
];

const reteriver = vectorStore.asRetriever({
filter: filter
)
or 
ScoreThresholdRetriever.fromVectorStore(vectorStore, {
        minSimilarityScore: 0.5, // Finds results with at least this similarity score
        maxK: 1, // The maximum K value to use. Use it based to your chunk size to make sure you don't run out of tokens
        kIncrement: 1, // How much to increase K by each time. It'll fetch N results, then N + kIncrement, then N + kIncrement * 2, etc.
        filter: filter,
      });

After this change the filter will be passed to the elastic vector similaritySearchVectorWithScore function.

const relevantDocuments = await elasticVectorSearch.similaritySearchVectorWithScore(
  [/* your query vector */],
  10,
  filter
);

Fixes # (issue)


libs/langchain-community/src/vectorstores/elasticsearch.ts
```diff
@@ -342,15 +343,27 @@ export class ElasticVectorSearch extends VectorStore {
...
           [condition.operator]: {
             [metadataField]: condition.value,
           },
         });
       }
     }
-    return { must, must_not };
+    const result = { must, must_not };
+    
+    if(should.length > 0){
```
joemcelroy, line 362: can you also extend the fn return type to include should please?

joemcelroy (commented):
Ideally this should also include tests but im aware that the tests dont currently cover metadata filtering either. If possible, please cover this in the elasticsearch.int test file.

libs/langchain-community/src/vectorstores/elasticsearch.ts
djaffer: Done.

joemcelroy (approved):

PR #5577, opened 2024-05-28
infra[patch]: Yarn dedupe

bracesproul:
[`yarn dedup`](https://yarnpkg.com/cli/dedupe)

Hopefully this speeds up yarn install times


bracesproul:
Benefits (~30 sec) do not appear to outweigh the cost/risk of changes

PR #5584, opened 2024-05-28
langchain[minor]: Multi-file loader

theogravity:
Introduces the `MultiFileLoader`, which behaves like the `DirectoryLoader`, except the input is a list of file paths instead of a single directory path.


langchain/src/document_loaders/fs/multi_file.ts
```diff
@@ -0,0 +1,113 @@
+import type { extname as ExtnameT, resolve as ResolveT } from "node:path";
+import type { stat as StatT } from "node:fs/promises";
+import { Document } from "@langchain/core/documents";
+import { getEnv } from "@langchain/core/utils/env";
+import { BaseDocumentLoader } from "../base.js";
+import { type LoadersMapping, UnknownHandling } from "./directory.js";
```
theogravity, line 6: I'm not sure if these imports should be refactored / moved to some common file instead of being imported from `directory.js`.

Open to suggestions.

langchain/src/document_loaders/fs/multi_file.ts
```diff
@@ -0,0 +1,113 @@
...
+ *     ".pdf": (path: string) => new PDFLoader(path),
+ *   },
+ * );
+ *
+ * const docs = await multiFileLoader.load();
+ * console.log({ docs });
+ *
+ * ```
+ */
+export class MultiFileLoader extends BaseDocumentLoader {
```
theogravity, line 26: This implementation shares a lot in common with `DirectoryLoader` that I wonder if this is something that can be moved to a new base class so there is reuse.

langchain/src/document_loaders/fs/multi_file.ts
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  /**
+   * Imports the necessary functions from the `node:path` and
+   * `node:fs/promises` modules. It is used to dynamically import the
+   * functions when needed. If the import fails, it throws an error
+   * indicating that the modules failed to load.
+   * @returns A promise that resolves to an object containing the imported functions.
+   */
+  static async imports(): Promise<{
```
jacoblee93, line 97: We don't have to do this anymore, you can just import them at the top as normal:

jacoblee93:
Thanks for the PR!

Is this materially different from calling the `DirectoryLoader` in a loop? If it doesn't support individual files, would it make more sense to add that functionality?

theogravity:
It's a completely different use-case. I have files from different directories that I may pick and choose from vs entire directories, so looping with the DirectoryLoader wouldn't work as that is on a directory-basis.

langchain/src/document_loaders/fs/multi_file.ts
theogravity: This was pretty much copied from the directory loader. Should I also remove it from there as well?

langchain/src/document_loaders/fs/multi_file.ts
jacoblee93: This is fine

jacoblee93:
Thanks! Sorry for delay

PR #5586, opened 2024-05-28
core[minor],openai[patch]: Add usage metadata to `AIMessage`/`Chunk`

bracesproul:


libs/langchain-openai/src/chat_models.ts
```diff
@@ -553,6 +555,14 @@ export class ChatOpenAI<
       tool_choice: options?.tool_choice,
       response_format: options?.response_format,
       seed: options?.seed,
+      // Only set stream_options if the model is OpenAI
```
ccurme, line 558: Although the python implementation initially defaulted to include usage during streaming, I [updated this](https://github.com/langchain-ai/langchain/pull/22242) to no longer override OpenAI's default due to concerns about no longer having `"finish_reason"` in the final chunk of a stream.

examples/src/models/chat/integration_openai_stream_tokens.ts
```diff
@@ -0,0 +1,30 @@
+import { AIMessageChunk } from "@langchain/core/messages";
+import { ChatOpenAI } from "@langchain/openai";
+
+// Instantiate the model
+const model = new ChatOpenAI();
+
+const response = await model.stream("Hello, how are you?", {
+  // Pass the stream options
+  stream_options: {
```
jacoblee93, line 9: We've been using camel case for this kind of thing to match JS conventions (also `includeUsage` below)

examples/src/models/chat/integration_openai_stream_tokens.ts
```diff
@@ -0,0 +1,30 @@
...
+let finalResult: AIMessageChunk | undefined;
+for await (const chunk of response) {
+  if (finalResult) {
+    finalResult = finalResult.concat(chunk);
+  } else {
+    finalResult = chunk;
+  }
+}
+
+console.log(finalResult?.usage_metadata);
```
jacoblee93, line 24: This is fine to leave as camel case

libs/langchain-openai/src/chat_models.ts
```diff
@@ -553,6 +554,7 @@ export class ChatOpenAI<
       tool_choice: options?.tool_choice,
       response_format: options?.response_format,
       seed: options?.seed,
+      stream_options: options?.stream_options,
```
jacoblee93, line 557: Should we be passing this to OpenAI?

jacoblee93:
Thank you!

PR #5587, opened 2024-05-28
[Tracing] Support LANGSMITH_TRACING env var

hinthornw:


jacoblee93:
Thank you!

langchain-core/src/callbacks/manager.ts
```diff
@@ -984,7 +984,8 @@ export class CallbackManager
       getEnvironmentVariable("LANGCHAIN_VERBOSE") === "true" ||
       options?.verbose;
     const tracingV2Enabled =
-      getEnvironmentVariable("LANGCHAIN_TRACING_V2") === "true";
+      (getEnvironmentVariable("LANGCHAIN_TRACING_V2") ??
```
jacoblee93, line 987: I think this should actually be:

```js
getEnvironmentVariable("LANGCHAIN_TRACING_V2") === "true" || getEnvironmentVariable("LANGSMITH_TRACING") === "true"
```

jacoblee93:
Small comment - not sure desired behavior?

hinthornw (commented):
lgmt

PR #5603, opened 2024-05-29
mistralai[minor]: Add llms entrypoint, update chat model integration

bracesproul:


libs/langchain-mistralai/src/chat_models.ts
```diff
@@ -350,8 +338,18 @@ function _convertDeltaToMessageChunk(delta: {
 
 function _convertStructuredToolToMistralTool(
   tools: StructuredToolInterface[]
-): MistralAIToolInput[] {
-  return tools.map((tool) => convertToOpenAITool(tool) as MistralAIToolInput);
+): MistralAITool[] {
+  return tools.map((tool) => {
```
bracesproul, line 342: New `tool` type from mistral is not exactly compatible with OAI

libs/langchain-mistralai/src/chat_models.ts
```diff
@@ -70,40 +70,15 @@ interface TokenUsage {
...
-  responseFormat?: ResponseFormat;
-};
-
 interface MistralAICallOptions
   extends Omit<BaseLanguageModelCallOptions, "stop"> {
   response_format?: {
     type: "text" | "json_object";
   };
-  tools: StructuredToolInterface[] | MistralAIToolInput[];
+  tools: StructuredToolInterface[] | MistralAIToolInput[] | MistralAITool[];
```
bracesproul, line 81: Need to include the old type for back-compat

libs/langchain-mistralai/src/chat_models.ts
```diff
@@ -235,7 +210,7 @@ function convertMessagesToMistralMessages(
       content,
       tool_calls: toolCalls,
     };
-  });
+  }) as MistralAIMessage[];
```
bracesproul, line 213: Their sdk types don't play well with their API... casting allows us to use this type which doesn't throw errors from their API

PR #5604, opened 2024-05-29
core[minor]: Make LLMs and chat models always stream when invoked within streamEvents

jacoblee93:
Also fixed erroneous streaming callback behavior for a test model, which seems to have altered output for a few tests.

@nfcampos


langchain-core/src/language_models/chat_models.ts
```diff
@@ -306,48 +309,88 @@ export abstract class BaseChatModel<
...
+    // Even if stream is not explicitly called, check if model is implicitly
+    // called from streamEvents() or streamLog() to get all streamed events.
+    // Bail out if _streamResponseChunks not overridden
+    const hasStreamingHandler = !!runManagers?.[0].handlers.find((handler) => {
+      return isStreamEventsHandler(handler) || isLogStreamHandler(handler);
+    });
+    if (
+      hasStreamingHandler &&
+      baseMessages.length === 1 &&
+      this._streamResponseChunks !==
```
nfcampos, line 323: we don't do anything dumb like bind the method in constructor right?

jacoblee93: I don't think so, no. We're already relying on this check in another place

PR #5622, opened 2024-05-31
community[patch]: Fix bug in Neo4j new enhancedSchema flag

easwee:
Fixes broken formatSchema function when not using enhancedSchema option.

### DIFF
--- a/libs/langchain-community/src/graphs/neo4j_graph.ts
+++ b/libs/langchain-community/src/graphs/neo4j_graph.ts
@@ -188,20 +188,24 @@ function formatSchema(
     }
   } else {
     // Format node properties
-    formattedNodeProps = schema.nodeProps.map((el: Record<string, Any>) => {
-      const propsStr = el.properties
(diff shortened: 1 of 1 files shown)


libs/langchain-community/src/graphs/neo4j_graph.ts
```diff
@@ -188,20 +188,24 @@ function formatSchema(
...
-        .map((prop: Record<string, Any>) => `${prop.property}: ${prop.type} `)
-        .join(", ");
-      return `${el.type} {${propsStr} } `;
-    });
+    formattedRelProps = Object.entries(schema.relProps).map(
+      ([key, value]: [string, Any]) => {
+        const propsStr = value
+          .map((prop: Record<string, Any>) => `${prop.property}: ${prop.type} `)
+          .join(", ");
+        return `${key} {${propsStr} } `;
```
jacoblee93, line 206: Is the extra space here a mistake?

jacoblee93:
Thank you!

PR #5637, opened 2024-06-02
community[minor]: upgraded @mlc/web-llm dependency and updated it's ChatModel

tnfssc:
Using `@mlc/web-llm@^0.2.40` package with the chat model provided in `@langchain/community@^0.2.5` gives an error and doesn't run. This PR upgrades the peerDependency and devDependency `@mlc/web-llm` from `^0.2.35` to `^0.2.40` and fixes `webllm.ts` file appropriately to work with this version of dependency.


jacoblee93:
Thank you! Maybe we should consider locking the peer dependency... Seems like a lot of renaming going on.

PR #5674, opened 2024-06-05
docs[patch]: Remove duplicated heading section on rag.ipynb

froggy1014:
I suggest removing the duplicated heading section `## Preview` in `rag.ipynb`.

- https://js.langchain.com/v0.2/docs/tutorials/rag#preview

The duplicated heading section contains the same heading and explanation. 

I removed the first one because it mentions a limit of 20 lines of code, which doesn't make sense in the context below:

```
We can create a simple indexing pipeline and RAG chain to do this in ~20 lines of code:
```

P.S - There was a typo `isthe` in heading section that I removed


jacoblee93:
Thanks for this!

I agree it's redundant - could we leave the link to the source material in somewhere though?

bracesproul (approved):
Thanks for catching this!

jacoblee93:
Oh, sorry yes this is good!

PR #5677, opened 2024-06-05
community[patch]: allow to use postgres pool instance on PostgresRecordManager

josemussa:
This is a proposal to support passing an instance of Pool when instantiating PostgresRecordManager. This would enable the reuse of an existing database connection pool.

```ts
import pg from "pg";

const pool = new pg.Pool(postgresConnectionOptions);

const providedPoolRecordManager = new PostgresRecordManager("test", {
  ...config,
  pool,
});
```

Fixes # (issue)
https://github.com/langchain-ai/langchainjs/discussions/5676

### DIFF
--- a/libs/langchain-community/src/indexes/postgres.ts
+++ b/libs/langchain-community/src/indexes/postgres.ts
@@ -7,6 +7,7 @@ import {
 
 export type PostgresRecordManagerOptions = {
   postgresConnectionOptions: PoolConfig;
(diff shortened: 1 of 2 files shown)


jacoblee93:
Thank you!

biswarup-banerjee:
@josemussa  This is a really helpful fix! Thanks for this! Really cool!

PR #5714, opened 2024-06-10
openai[patch],azure-openai[patch]: Update legacy classes to respect the apiKey field

scottforte:


jacoblee93:
Thank you for this - CC @bracesproul let's make sure we add this to the new standardized tests.

scottforte:
> `ChatOpenAI`

Hey @jacoblee93 the page here does not show deprecated:
https://js.langchain.com/v0.2/docs/integrations/llms/openai

Does that UI page need to be updated as well?

jacoblee93:
OpenAI LLMs are not deprecated, however the only non-chat model they currently support is `gpt-3.5-turbo-instruct`:

https://platform.openai.com/docs/models/gpt-3-5-turbo

I agree a bit more funneling there would be helpful. Can have a look soon.

scottforte:
Ok - FYI The PR here LLM (non-chat) works with gpt-3.5-turbo-0125

jacoblee93:
Yes, however it's a very old legacy flow (we coerce it to use the chat endpoint behind the scenes) and shouldn't be used. I'll flag this with the rest of the team!

scottforte:
Super helpful insight - thank you

jacoblee93:
Merged #5718 to help folks in the future.

jacoblee93:
Thank you!

PR #5717, opened 2024-06-10
openai[minor],docs[minor]: Add parallel tool calls call option and docs

bracesproul:
Also bumped openai to latest version


jacoblee93:
Thanks, will merge later!

PR #5731, opened 2024-06-11
Add puppeteer agent tool

stevenmichaelthomas:
Adds puppeteer agent tool. When testing out projects using the in-built `WebBrowser`, we found the lack of JS execution with Cheerio to be challenging. 

This instead leverages Puppeteer to handle full page execution, returning a lightly cleaned version of HTML, plus the ability to generate embeddings and send valid chunks to the specified model. 

This is my first contribution here, so please let me know if I've missed anything, or if there is any feedback. Thank you!

### DIFF
--- a/libs/langchain-community/src/tools/puppeteer.ts
+++ b/libs/langchain-community/src/tools/puppeteer.ts
@@ -1,0 +1,177 @@
+import {launch} from "puppeteer";
+
+import type { BaseLanguageModelInterface } from "@langchain/core/language_models/base";
+import { Tool, ToolParams } from "@langchain/core/tools";
+import { RunnableSequence } from "@langchain/core/runnables";
+import { StringOutputParser } from "@langchain/core/output_parsers";
+import type { EmbeddingsInterface } from "@langchain/core/embeddings";
+
+import {
+  RecursiveCharacterTextSplitter,
+  TextSplitter,
+} from "@langchain/textsplitters";
+
+import { MemoryVectorStore } from "langchain/vectorstores/memory";
+import { formatDocumentsAsString } from "langchain/util/document";
+import { Document } from "langchain/document";
+import {load} from "cheerio";
+
+export const parseInputs = (inputs: string): [string, string] => {
+  const [baseUrl, task] = inputs.split(",").map((input) => {
+    let t = input.trim();
+    t = t.startsWith('"') ? t.slice(1) : t;
+    t = t.endsWith('"') ? t.slice(0, -1) : t;
+    t = t.endsWith("/") ? t.slice(0, -1) : t;
+    return t.trim();
+  });
+
+  return [baseUrl, task];
+};
+
+export const getRelevantHtml = async (
+  html: string,
+): Promise<string> => {
+  const $ = load(html);
+
+  const tagsToRemove = ['script', 'svg', 'style']
+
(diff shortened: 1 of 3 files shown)


jacoblee93:
Hey @stevenmichaelthomas! Thanks for the PR. Looks really cool!

I ran `yarn format` and registered the entrypoint artifacts to `langchain.config.js`. Could you add some docs and possibly an integration test as well?

libs/langchain-community/src/tools/puppeteer.ts
```diff
@@ -0,0 +1,181 @@
...
+    } else {
+      const docs = texts.map(
+        (pageContent: string) =>
+          new Document({
+            pageContent,
+            metadata: [],
+          })
+      );
+
+      const vectorStore = await MemoryVectorStore.fromDocuments(
```
jacoblee93, line 166: Consider using contextual compression - likely no need for an explicit vectorstore:

https://js.langchain.com/v0.2/docs/how_to/contextual_compression

https://github.com/langchain-ai/weblangchain/blob/main/nextjs/app/api/chat/stream_log/route.ts#L101

libs/langchain-community/src/tools/puppeteer.ts
```diff
@@ -0,0 +1,181 @@
...
+  await page.setExtraHTTPHeaders(headers);
+  await page.goto(baseUrl, { waitUntil: "networkidle0" });
+
+  const bodyHtml = await page.content();
+  await browser.close();
+
+  return bodyHtml;
+};
+
+const DEFAULT_HEADERS = {
```
jacoblee93, line 63: Move above usage to avoid linter issue

libs/langchain-community/src/tools/puppeteer.ts
```diff
@@ -0,0 +1,181 @@
...
+  "Sec-Fetch-Dest": "document",
+  "Sec-Fetch-Mode": "navigate",
+  "Sec-Fetch-Site": "cross-site",
+  "Upgrade-Insecure-Requests": "1",
+  "User-Agent":
+    "Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/111.0",
+};
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type Headers = Record<string, any>;
```
jacoblee93, line 79: can be `Record<string, string>`

libs/langchain-community/src/tools/puppeteer.ts
```diff
@@ -0,0 +1,181 @@
...
+
+    this.textSplitter =
+      textSplitter ??
+      new RecursiveCharacterTextSplitter({
+        chunkSize: 2000,
+        chunkOverlap: 200,
+      });
+  }
+
+  async _call(inputs: string) {
```
jacoblee93, line 137: You can make this extend `StructuredTool` and give it a schema instead, might be a bit nicer/more reliable than comma separated values

stevenmichaelthomas:
Thanks so much for the feedback! I'll incorporate changes and ping you when everything is added in and tests are passing.

TimSusa:
I like this idea

hntrl:
Hey @stevenmichaelthomas, sorry to revive an old PR!

PR #5756, opened 2024-06-13
openai[patch]: Fix tool content list

bracesproul:


libs/langchain-openai/src/chat_models.ts
```diff
@@ -228,11 +228,13 @@ function convertMessagesToOpenAIParams(messages: BaseMessage[]) {
     }
     if (message.additional_kwargs.function_call != null) {
       completionParam.function_call = message.additional_kwargs.function_call;
+      completionParam.content = null;
```
jacoblee93, line 231: Why not just do `delete completionParam.content`?

Prefer avoiding `null` when possible

PR #5760, opened 2024-06-14
community[minor]: Add Indexing API support for Milvus

miloradvojnovic:
Summary:
 - Added Indexing API support for Milvus
 - Implemented document addition by id (addDocuments method with ids argument)
 - Implemented delete by id (delete method with ids argument)
 - Bumped Milvus SDK version because Indexing API needs the upsert method
 - Added tests for Milvus Indexing


jacoblee93:
Looks great, thank you!

Issue #965, opened 2023-04-23
How to get token count, callbacks work but just for ChatOpenAI but not for RetrievalQAChain


theTechGoose:
I am trying to get a token count for a process, I am passing callbacks to the class initialization like this

```typescript
let finalTokens = 0
const initPayload = {
  openAIApiKey: process.env['OPEN_AI_KEY'],
  temperature: 1.5,
  callbacks: [
    {
      handleLLMEnd: (val) => {
        try {
          const tokens = val.llmOutput.tokenUsage.totalTokens
          finalTokens += tokens
          console.log({tokens, finalTokens})
        } catch {
          console.log(val.generations[0])
        }
      },
    },
  ],
};
```

However all of the calls from a RetrievalQAChain end up in the catch portion of that try-catch block as 'tokenUsage' does not exist for those calls. Can someone point me in the right direction?

miekassu:
I'm having a similar issue.
When I define `gpt-3.5-turbo` as the model for `OpenAI` construct, `llmOutput` is missing `tokenUsage` object.

Using the same construct, but not defining the model returns `token usage` as part of `llmOutput`

Not working:
```
  const model = new OpenAI({
    openAIApiKey: openAISecret,
    modelName: 'gpt-3.5-turbo',
    callbacks: [
      {
        handleLLMEnd: async (output: LLMResult) => {
          logger.info('output', { output })
          logger.info('tokenUsage', { tokenUsage: output.llmOutput })
          // tokenUsage: UNDEFINED
        },
      },
    ],
  })
```

Working:
```
  const model = new OpenAI({
    openAIApiKey: openAISecret,
    callbacks: [
      {
        handleLLMEnd: async (output: LLMResult) => {
          logger.info('output', { output })
          logger.info('tokenUsage', { tokenUsage: output.llmOutput })
          // tokenUsage: found
        },
      },
    ],
  })
```

ibrahimyaacob92:
still waiting for the solution on this

pond918:
tested with azure API:
```shell
[...]

PR #5761, opened 2024-06-14
openai[minor]: OpenAI - Ability to set stream_options and streamUsage when instantiating ChatOpenAI class.

deranga:
Fixes [965](https://github.com/langchain-ai/langchainjs/issues/965) - Where its not possible to get the usage back from OpenAI when streaming and using chains or agents.


bracesproul (changes requested):
hey, thanks for taking initiative on this! We're actually implementing a new `streamUsage` constructor arg & call option, but haven't gotten around to implementing it for OpenAI yet. If you wanted to implement it, [here's a PR](https://github.com/langchain-ai/langchainjs/pull/5730/files) you can copy from. If not, feel free to close this PR.

Please ping me if you have any questions on it!

bracesproul:
Actually, lets include both `streamOptions: { includeUsage: boolean }` (prefer camel case for our API) and `streamUsage: boolean`. OAI might add more fields to their `stream_options` API in the future, in which case we'd like a way to support those too!

deranga:
@bracesproul, I think I've made the changes you are after. 

I have a separate, but related question around embeddings token usage, happy to ask this in a separate issue.... The core langchainjs package has types that prevent returning metadata around the embedding response, and ideally from pov it would allow exposing the token usage if the integration supported it. Do you know if there are any plans to make it possible to access token usage from embeddings requests?

deranga:
@bracesproul, I've reviewed the PR for anthropic with streamUsage,  and updated this PR to include it for OpenAI.

I think the interface at the moment is a little strange as currently one of either stream_options or streamUsage would have to take precedence? For now I've made it that stream_options takes precedence if set, but this might lead to odd behaviour for users of the library?

Any input would be appreciated on how you want this to proceed.

deranga:
@jacoblee93, this is the PR you suggested in #965 alongside the streamUsage addition suggested above. Do let me know if there are any changes needed?

jacoblee93:
Hey! So sorry will look tomorrow!

deranga:
Want me to update the tests?

jacoblee93:
Getting a few 400s when running integration tests:

```
400 The 'stream_options' parameter is only allowed when 'stream' is enabled.
```

Seems like we may need to be a bit more selective on how we populate this?

deranga:
Is it possible to run the integration tests outside of being kicked off by a maintainer?

To address the 400 for stream: false when requesting usage data I've updated so that either stream_options: {include_usage:true}/ streamUsage true will force the stream flag to be set to true, and if not fallback to this.streaming. To make it BC I've altered streamUsage to default to false

jacoblee93:
Changed to default to true, avoid passing as an invocation param if `this.streaming` is false.

jacoblee93:
Thank you!

PR #5762, opened 2024-06-14
docs: Graph tutorial updates

iPagar:
- fix for a ts error: Property 'schema' is protected and only accessible within class 'Neo4jGraph' and its subclasses.
- add console.log to show query result


bracesproul (approved):
thanks you!

Issue #5771, opened 2024-06-15
WebLLM Example Code does not Work in Google Chrome


radiantone:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain.js documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain.js rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

Tried the example code from langchain and it throws exception in Google chrome.
web-llm 0.2.46

```
const model = new ChatWebLLM({
  model: 'Phi-3-mini-4k-instruct-q4f16_1-MLC',
  chatOptions: {
    temperature: 0.5,
  },
});
const response = await model.invoke([
  new HumanMessage({ content: "What is 1 + 1?" }),
]);
```

### Error Message and Stack Trace (if applicable)

```
Error: Model not loaded before calling chatCompletion(). Please ensure you have called `MLCEngine.reload(model)` to load the model before initiating chat operations, or initialize your engine using `CreateMLCEngine()` with a valid model configuration.
    at MLCEngine.<anonymous> (chunk-ZOKLVKAW.js?v=427e5d1f:13082:15)
    at Generator.next (<anonymous>)
    at chunk-ZOKLVKAW.js?v=427e5d1f:978:67
    at new Promise (<anonymous>)
    at __awaiter (chunk-ZOKLVKAW.js?v=427e5d1f:960:10)
    at MLCEngine.chatCompletion (chunk-ZOKLVKAW.js?v=427e5d1f:13080:12)
    at Completions.create (chunk-ZOKLVKAW.js?v=427e5d1f:11815:24)
    at ChatWebLLM._streamResponseChunks (@langchain_community_chat_models_webllm.js?v=427e5d1f:102:55)
    at _streamResponseChunks.next (<anonymous>)
    at ChatWebLLM._call (@langchain_community_chat_models_webllm.js?v=427e5d1f:125:22)

```

### Description

WebLLM Example Code does not Work in Google Chrome

### System Info

langchain@0.2.5 | MIT | deps: 16 | versions: 277
Typescript bindings for langchain
https://github.com/langchain-ai/langchainjs/tree/main/langchain/

keywords: llm, ai, gpt3, chain, prompt, prompt engineering, chatgpt, machine learning, ml, openai, embeddings, vectorstores

dist
[...]

PR #5776, opened 2024-06-16
community[minor]: Update ChatWebLLM to match dependency version and update example

Neet-Nestor:
Fixes https://github.com/mlc-ai/web-llm/issues/477

The current code sample in the documentation is incorrect as it misses the required `initailize()` call and thus the example code won't work. This PR fixes the example code. 

Meanwhile, this PR also updates WebLLM package from `v0.2.40` to `v0.2.46`, locked the version, and updated the function signature of `engine.reload()` to match by removing the `appConfig` parameter.


jacoblee93:
Thanks for this - let's also lock to that specific minor version (remove the carat). There have been a few breaking changes in patch releases now CC @CharlieFRuan

jacoblee93:
I think will also fix #5771

jacoblee93:
Worked for me! Thank you

PR #5777, opened 2024-06-16
update model name in these for making them eligible for multimodality

tanyaasharma:
… models

Fixes # (issue)

### DIFF
--- a/examples/src/models/chat/integration_googlevertexai-streaming.ts
+++ b/examples/src/models/chat/integration_googlevertexai-streaming.ts
@@ -4,7 +4,7 @@ import { ChatVertexAI } from "@langchain/google-vertexai";
 
 const model = new ChatVertexAI({
   temperature: 0.7,
-  model: "gemini-1.5-pro-001",
+  model: "gemini-1.5-flash-001",
 });
 const stream = await model.stream([
   ["system", "You are a funny assistant that answers in pirate language."],
--- a/examples/src/models/chat/integration_googlevertexai-streaming.ts
+++ b/examples/src/models/chat/integration_googlevertexai-streaming.ts
@@ -4,6 +4,7 @@ import { ChatVertexAI } from "@langchain/google-vertexai";
 
 const model = new ChatVertexAI({
   temperature: 0.7,
+  model: "gemini-1.5-pro-001",
 });
 const stream = await model.stream([
   ["system", "You are a funny assistant that answers in pirate language."],
--- a/examples/src/models/chat/integration_googlevertexai-tools.ts
+++ b/examples/src/models/chat/integration_googlevertexai-tools.ts
@@ -25,7 +25,7 @@ const geminiCalculatorTool: GeminiTool = {
 
 const model = new ChatVertexAI({
   temperature: 0.7,
-  model: "gemini-pro-vision",
+  model: "gemini-1.5-flash-001",
 }).bind({
   tools: [geminiCalculatorTool],
 });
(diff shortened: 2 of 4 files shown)


examples/src/models/chat/integration_googlevertexai-tools.ts
```diff
@@ -25,7 +25,7 @@ const geminiCalculatorTool: GeminiTool = {
 
 const model = new ChatVertexAI({
   temperature: 0.7,
-  model: "gemini-1.0-pro",
+  model: "gemini-pro-vision",
```
afirstenberg, line 28: Suggest "gemini-1.5-flash" or "gemini-1.5-flash-001" instead. (Pro Vision has been deprecated on the AI Studio side and will be supported for less time than 1.5 flash on the Vertex AI side.)

examples/src/models/chat/integration_googlevertexai-streaming.ts
```diff
@@ -4,6 +4,7 @@ import { ChatVertexAI } from "@langchain/google-vertexai";
 
 const model = new ChatVertexAI({
   temperature: 0.7,
+  model: "gemini-1.5-pro-001",
```
afirstenberg, line 7: While 1.5 Pro is fine, 1.5 Flash might be a better general purpose suggestion for the examples (since it is significantly cheaper).

examples/src/models/chat/integration_googlevertexai-wsa.ts
```diff
@@ -13,7 +13,7 @@ const calculatorSchema = z.object({
 
 const model = new ChatVertexAI({
   temperature: 0.7,
-  model: "gemini-1.0-pro",
+  model: "gemini-pro-vision",
```
afirstenberg, line 16: Suggest "gemini-1.5-flash" or "gemini-1.5-flash-001" instead. (Pro Vision has been deprecated on the AI Studio side and will be supported for less time than 1.5 flash on the Vertex AI side.)

examples/src/models/chat/integration_googlevertexai.ts
```diff
@@ -4,7 +4,7 @@ import { ChatVertexAI } from "@langchain/google-vertexai";
 
 const model = new ChatVertexAI({
   temperature: 0.7,
-  model: "gemini-1.0-pro",
+  model: "gemini-pro-vision",
```
afirstenberg, line 7: Suggest "gemini-1.5-flash" or "gemini-1.5-flash-001" instead. (Pro Vision has been deprecated on the AI Studio side and will be supported for less time than 1.5 flash on the Vertex AI side.)

afirstenberg (changes requested):
None of the examples actually use multimodal inputs, so I'm not sure why the need to switch, but if we're going to, let's switch to the most recent releases that have stable versions and that also support general-purpose needs at reasonable prices.

examples/src/models/chat/integration_googlevertexai.ts
tanyaasharma: Yes, Allen, i think you are right. Thank you, will update it

examples/src/models/chat/integration_googlevertexai-streaming.ts
LindaLawton: You realize if you start adding models to this that your going to have to update it every time one gets deprecated.

jacoblee93: Fair point - however we can't update the class default without changing behavior so we generally put it explicitly. New models are infrequent enough where bumping all the strings isn't a huge issue.

jacoblee93:
Thank you!

examples/src/models/chat/integration_googlevertexai-streaming.ts
LindaLawton: I think we have a different definition of **infrequent** I would be willing to bet your looking at ever few months.

PR #5796, opened 2024-06-18
Fix PostgresRecordManager table being created with name 'undefined'

### LINKED ISSUE #5795: PostgresRecordManager creates table named `undefined`
### Checked other resources

### Example Code

```typescript
const recordManagerConfig = {
  postgresConnectionOptions: {
    type: "postgres",
    host: "127.0.0.1",
    port: 5432,
    user: "myuser",
    password: "ChangeMe",
    database: "api",
  } as PoolConfig,
};

const recordManager = new PostgresRecordManager(
  "test_namespace",
  recordManagerConfig
);

// Creates a table named `undefined`
await recordManager.createSchema();
```

_No response_

### Description

When no explicit table name is passed in the constructor, the computed table name becomes `undefined`. The table name is optional and should have the fallback value `upsertion_records`. 

### System Info

Not relevant

MJDeligan:
The final table name is computed from the table name that is passed in the config which is an optional field. When no table name is explicitly passed, the final table name becomes `undefined`. This PR fixes the bug by using the correct `tableName` property, which is initialized with a fallback value.

Note: This is technically breaking for anyone who started using the api after the schema support was added.

Fixes #5795

### DIFF
--- a/libs/langchain-community/src/indexes/postgres.ts
+++ b/libs/langchain-community/src/indexes/postgres.ts
@@ -29,8 +29,8 @@ export class PostgresRecordManager implements RecordManagerInterface {
     this.pool = pool || new pg.Pool(postgresConnectionOptions);
     this.tableName = tableName || "upsertion_records";
     this.finalTableName = config.schema
-      ? `"${config.schema}"."${tableName}"`
-      : `"${tableName}"`;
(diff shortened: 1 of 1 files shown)


jacoblee93:
Agh, thanks. I think worth it

PR #5801, opened 2024-06-18
docs[patch]: Adds docs feedback text

jacoblee93:


docs/core_docs/src/theme/Feedback.js
```diff
@@ -161,6 +164,33 @@ export default function Feedback() {
...
+      siteConfig.customFields.supabaseUrl,
+      siteConfig.customFields.supabasePublicKey
+    );
+    const { error } = await supabase.from("feedback_details").insert({
+      feedback_id: feedbackId,
+      details,
+    });
+    if (error) {
+      console.error("Failed to add feedback details", error);
+      throw error;
```
bracesproul, line 189: Do we want to actually throw this error in production? I think it would be better to just silently fail.

bracesproul: Adding on that, I'd wrap this function in a try/catch so if something else goes wrong it doesn't throw, and instead just silently logs the error and returns.

docs/core_docs/src/theme/Feedback.js
```diff
@@ -161,6 +164,33 @@ export default function Feedback() {
     setFeedbackSent(true);
   };
 
+  const handleFeedbackDetails = async (e) => {
+    e.preventDefault();
+    if (!feedbackId) {
+      setFeedbackDetailsSent(true);
```
bracesproul, line 170: if `feedbackDetailsSent` is true do we want to return or allow the user to submit multiple times?

Also I don't understand why we're returning here & setting true if there is no ID?

docs/core_docs/src/theme/Feedback.js
```diff
@@ -110,7 +111,9 @@ const getIpAddress = async () => {
 
 export default function Feedback() {
   const { setCookie, checkCookie } = useCookie();
+  const [feedbackId, setFeedbackId] = useState(null);
```
bracesproul, line 114: Might make more sense to store the ID in the cookie which is tracking feedback, this way they can't spam it, and their ID will persist between refreshes (for 24hrs)

bracesproul (commented):
Couple questions/nits. otherwise lgtm

docs/core_docs/package.json
```diff
@@ -45,6 +45,7 @@
     "process": "^0.11.10",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
+    "uuid": "^10.0.0",
```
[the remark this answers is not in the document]
jacoblee93: Sure

docs/core_docs/src/theme/Feedback.js
jacoblee93: It'll just show up in the console which I think is fine, probably don't need both though

docs/core_docs/src/theme/Feedback.js
jacoblee93: If there's no id there, you're in a bad state and best thing to do is mark as done and no-op

docs/core_docs/src/theme/Feedback.js
jacoblee93: They can't visually get back into that state anyway

PR #5803, opened 2024-06-18
community[patch]: feat: Make Zep Memory message merging optional

paul-paliychuk:
We are working on the Zep Cloud n8n integration, and got a request from their team to expose an option to return zep memory as separate messages (as opposing to a condensed HumanMessage that contains summary, conversation facts and most recent message history). While we still want to maintain the condensed option as a default way to retrieve zep memory, we have added a `separateMessages` optional boolean input parameter to indicate if the user wants to return an array of messages instead.

Additionally, we updated zep-cloud sdk peer dependency min version to 1.0.6


paul-paliychuk:
@jacoblee93 Hey, not sure why the docs and api-refs deploys are failing here, can confirm that docs:api_refs and docs start locally.

paul-paliychuk:
Actually looks like they worked this time 👍

jacoblee93:
Thank you!

PR #5810, opened 2024-06-19
cohere[minor]: Add support for tool calling cohere

Anirudh31415926535:
This PR supports tool calling in js for cohere models.

High level feature introductions:
1. Parse chat history parameter in request to now include tool messages
2. Handle parsing logic separately as required for multi hop tool calling and single step tool calling
3. Add tests to ensure tool calling works as expected
4. Bump default model version to 'command-r-plus'
5. Relevant doc changes to remove hardcoded 'model=command' and to showcase tool calling support


langchain-core/tsconfig.json
```diff
@@ -9,7 +9,7 @@
       "ES2022.Object",
       "DOM"
     ],
-    "module": "ES2020",
+    "module": "NodeNext",
```
jacoblee93, line 12: Can you revert these? Should not be in this PR.

libs/langchain-cohere/package.json
```diff
@@ -25,7 +25,7 @@
     "test": "NODE_OPTIONS=--experimental-vm-modules jest --testPathIgnorePatterns=\\.int\\.test.ts --testTimeout 30000 --maxWorkers=50%",
     "test:watch": "NODE_OPTIONS=--experimental-vm-modules jest --watch --testPathIgnorePatterns=\\.int\\.test.ts",
     "test:single": "NODE_OPTIONS=--experimental-vm-modules yarn run jest --config jest.config.cjs --testTimeout 100000",
-    "test:int": "NODE_OPTIONS=--experimental-vm-modules jest --testPathPattern=\\.int\\.test.ts --testTimeout 100000 --maxWorkers=50%",
+    "test:int": "NODE_OPTIONS=--experimental-vm-modules jest --testPathPattern=\\chat_models.int\\.test.ts --testTimeout 100000 --maxWorkers=50%",
```
jacoblee93, line 28: Revert

package.json
```diff
@@ -69,5 +69,9 @@
       "eslint --cache --fix"
     ],
     "*.md": "prettier --config .prettierrc --write"
+  },
+  "dependencies": {
+    "jest-cli": "^29.7.0",
```
jacoblee93, line 74: Revert

libs/langchain-cohere/src/chat_models.ts
```diff
@@ -135,7 +196,7 @@ export class ChatCohere<
 
   client: CohereClient;
 
-  model = "command";
+  model = "command-r-plus";
```
jacoblee93, line 199: This is technically breaking

jacoblee93:
Thanks for this!

Could you try scoping the changes down a bit? Appears to be a lot of unrelated diffs.

Also, don't forget to run `yarn format` and `yarn lint`!

libs/langchain-cohere/src/chat_models.ts
```diff
@@ -445,3 +762,5 @@ export class ChatCohere<
 interface CohereLLMOutput {
   estimatedTokenUsage?: TokenUsage;
 }
+
+/* eslint-enable  @typescript-eslint/no-explicit-any */
```
jacoblee93, line 766: Not necessary

libs/langchain-cohere/src/chat_models.ts
```diff
@@ -189,40 +273,256 @@ export class ChatCohere<
...
+      forceSingleStep: options.forceSingleStep,
+      tools: options.tools,
     };
     // Filter undefined entries
     return Object.fromEntries(
       Object.entries(params).filter(([, value]) => value !== undefined)
     );
   }
 
+  /** @ignore */
```
jacoblee93, line 285: Make the method `private` or `protected` and prefix with an underscore (`_`) instead

libs/langchain-cohere/src/chat_models.ts
```diff
@@ -189,40 +273,256 @@ export class ChatCohere<
...
+      message: messageStr,
+      chatHistory,
+      toolResults: toolResults.length > 0 ? toolResults : undefined,
+      ...params,
+    };
+
+    return req;
+  }
+
+  getCurrChatTurnMessages(messages: BaseMessage[]): BaseMessage[] {
```
jacoblee93, line 386: See above

libs/langchain-cohere/src/chat_models.ts
jacoblee93, line 1: You should also add a `override bindTools()` method, it'll make this support other methods for free:

https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-anthropic/src/chat_models.ts#L589

libs/langchain-cohere/src/chat_models.ts
```diff
@@ -102,10 +151,45 @@ function convertMessagesToCohereMessages(
...
+    return {
+      role: getRole(message._getType()),
+      message: getContent(message.content),
+      toolResults,
+    };
+  } else if (
+    // eslint-disable-next-line no-instanceof/no-instanceof
+    message instanceof HumanMessage ||
+    // eslint-disable-next-line no-instanceof/no-instanceof
+    message instanceof SystemMessage
```
jacoblee93, line 182: Use e.g. `message._getType() === "human"` instead.

langchain-core/src/utils/function_calling.ts
```diff
@@ -18,6 +18,53 @@ export function convertToOpenAIFunction(
   };
 }
 
+/**
+ * Formats a `StructuredTool` instance into a format that is compatible
+ * with OpenAI function calling. It uses the `zodToJsonSchema`
+ * function to convert the schema of the `StructuredTool` into a JSON
+ * schema, which is then used as the parameters for the Cohere function.
+ */
+export function convertToCohereTool(tool: StructuredToolInterface): {
```
jacoblee93, line 27: Let's not add this to core - can we keep it in the Cohere package

libs/langchain-cohere/src/chat_models.ts
billytrend-cohere: if we're going to break here, can we make model a required param

libs/langchain-cohere/src/tests/chat_models.int.test.ts
```diff
@@ -140,3 +147,58 @@ test("Invoke token count usage_metadata", async () => {
...
+  expect(res.usage_metadata.total_tokens).toBe(
+    res.usage_metadata.input_tokens + res.usage_metadata.output_tokens
+  );
+  expect(res.tool_calls).toBeDefined();
+  expect(res.tool_calls?.length).toBe(1);
+  const tool_id = res.response_metadata.toolCalls[0].id;
+  messages.push(res);
+  messages.push(
+    new ToolMessage(
+      "Aidan Gomez is the president of Singapore",
```
billytrend-cohere, line 192: TIL

billytrend-cohere (approved):

libs/langchain-cohere/src/chat_models.ts
Anirudh31415926535: Hey @billytrend-cohere We're making the model a required param for Embedding, and Reranker. Imo it is better to leave model as a non-required param for chat, since the implication of an unforeseen model bump isn't huge compared to embedding/rerank.

Anirudh31415926535:
@jacoblee93 Hey, thanks so much for the review and comments! I've addressed them! Could you please help to take a look again if possible and then we can try to see if we can get this PR and the other one released by this week? Thanks so much! 💯

jacoblee93:
Yes will look today, thanks for the hard work!

libs/langchain-cohere/src/chat_models.ts
jacoblee93: Would still prefer this to be required (look at what happened with `gpt-4o`) but if you feel strongly I'm fine with keeping it.

jacoblee93:
Pushed some refactoring/lint/format stuff, will merge alongside the embedding PR and minor bump the package.

jacoblee93:
Thank you!

Issue #5076, opened 2024-04-11
google-common / google-vertex [Feature]: Support GCS as media/metadata store


afirstenberg:
### Privileged issue

- [X] I am a LangChain maintainer, or was asked directly by a LangChain maintainer to create an issue here.

### Issue Content

See #5075, but this would be the equivalent version to be used by Vertex AI

afirstenberg:
Being done as part of #5835

Issue #5075, opened 2024-04-11
google-common [Feature]: Support File API as media/metadata store


afirstenberg:
### Privileged issue

- [X] I am a LangChain maintainer, or was asked directly by a LangChain maintainer to create an issue here.

### Issue Content

Works on AI Studio only.

See https://ai.google.dev/api/rest/v1beta/files and https://ai.google.dev/api/rest/v1beta/media/upload
See https://github.com/google-gemini/cookbook/blob/main/quickstarts/Video.ipynb

See https://github.com/langchain-ai/langchain-google/issues/147

This is an odd kind of store 
- You store the media file, and get back metadata including the URL. 
- Then you can use this URL for `fileData` messages to include the media. 
- It manages the file (it will delete it after 2 days).
- **There is no way to download the media, however, just the metadata.**

I'd like to have this used in a couple of ways:
- As a stand-alone component so devs can save the media and manage the URL themselves (initial priority)
- As something used by the Chat module so if a `file_url` or `media` type is provided, it will save it to the File store and use the URL instead (future implementation)

What is the best interface to represent this with? In the past, @jacoblee93 has suggested subclassing `BaseStore` where the key is... a filename? And the value when set is just the file (in what form) and when retrieved is a Record<string,string>? I'm honestly not sure.

(I would also like a companion version, with the same interface, that words with Google Cloud Storage for use with Vertex. See #5076 )

afirstenberg:
Being done as part of #5835

Issue #5196, opened 2024-04-24
Unable to use non-Gemini models through ChatVertexAI


jarib:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain.js documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain.js rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```typescript

import { ChatVertexAI } from '@langchain/google-vertexai'

const chat = new ChatVertexAI({
    model: 'claude-3-opus@20240229',
    temperature: 0.0,
    maxOutputTokens: 1200,
})

const result = await chat.invoke('Tell me a story')

console.log(result)
```

### Error Message and Stack Trace (if applicable)

```
Error: Unable to verify model params: {"lc":1,"type":"constructor","id":["langchain","chat_models","chat_integration","ChatVertexAI"],"kwargs":{"model":"claude-3-opus","temperature":0,"max_output_tokens":1200,"platform_type":"gcp"}}
    at validateModelParams (file:///Users/redacted/src/langchain-repro/node_modules/@langchain/google-common/dist/utils/common.js:100:19)
    at copyAndValidateModelParamsInto (file:///Users/redacted/src/langchain-repro/node_modules/@langchain/google-common/dist/utils/common.js:105:5)
    at new ChatGoogleBase (file:///Users/redacted/src/langchain-repro/node_modules/@langchain/google-common/dist/chat_models.js:191:9)
    at new ChatGoogle (file:///Users/redacted/src/langchain-repro/node_modules/@langchain/google-gauth/dist/chat_models.js:12:9)
    at new ChatVertexAI (file:///Users/redacted/src/langchain-repro/node_modules/@langchain/google-vertexai/dist/chat_models.js:10:9)
    at main (file:///Users/redacted/src/langchain-repro/test.js:5:22)
    at file:///Users/redacted/src/langchain-repro/test.js:22:1
    at ModuleJob.run (node:internal/modules/esm/module_job:218:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:329:24)
    at async loadESM (node:internal/process/esm_loader:28:7)
```

### Description

I am trying to use ChatVertexAI with Anthropic Claude 3, but it seems this class only supports Gemini models and returns the above error message.

This appears to be a deliberate choice in the code:

[...]

PR #5835, opened 2024-06-21
google-common[major]: Media manager

afirstenberg:
WORK IN PROGRESS - FEEDBACK REQUESTED

There are still other tests and classes to implement, but this should give a first idea of where I see `MediaManager` as going.

The goal is to provide infrastructure that supports the `fileData` part in the Gemini API.
- For AI Studio, this must be a URL generated by the File API, while on Vertex AI, this must be a Google Cloud Storage gs: URI.
- The idea is that developers will be able to provide an arbitrary URI (ie - an https URL) or data and that this will get mapped to an appropriate supported URI for use in the Gemini API
- This is useful for image support, and necessary for video and audio support
- `MediaManager` is platform (and LLM) neutral, but some of the `BlobStore` implementations obviously aren't.
- The Gemini portions of the library will be passed a `MediaManager` - MediaMager will, itself, have no knowledge of the LLMs

TODO:
- A `BlobStore` for the AI Studio File API that backs the data somewhere
- examples, and documentation

Open questions:
- How do we get `MediaBlob` serializable

Will address #5075 #5076

cc: @jacoblee93 @bracesproul @eyurtsev


afirstenberg:
I think the merge with the usage results update should be straightforward, but needs to be carefully looked at to make sure we don't lose anything.

At this point - the tests in google-gauth show how the MediaManager can be used.

afirstenberg:
Although missing documentation, I think this is ready for review with intent to merge **but not immediately release**.  I do want to get it merged as quickly as reasonably possible.

Reasons to merge:
* There are changes that will be useful in addressing #5196, and I'd like to get this merged and make sure things are somewhat stable before I tackle that one
* No existing features change. This is an addition of a new feature, so people can continue to use while we work on documentation
* All changes are isolated to google-common and derived packages.
  *  Although there are a number of classes that would be reasonable to merge to core, these are isolated in one file and can be evaluated for that separately.

TODO:
* Documentation
* Examples

cc: @jacoblee93 @bracesproul

jacoblee93:
Hey @afirstenberg, apologies for the delay - looking now!

libs/langchain-google-common/src/connection.ts
```diff
@@ -177,6 +210,8 @@ export abstract class GoogleAIConnection<
 
   client: GoogleAbstractedClient;
 
+  api;  // FIXME: Make this a real type
```
jacoblee93, line 213: Explicitly type as `any`/`unknown`

libs/langchain-google-common/src/chat_models.ts
```diff
@@ -120,29 +119,30 @@ class ChatConnection<AuthOptions> extends AbstractGoogleLLMConnection<
       }, []);
   }
 
-  formatSystemInstruction(
+  async formatSystemInstruction(
```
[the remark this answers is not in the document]
afirstenberg: This should just be internal. I couldn't find any other place that calls it, and it is meant as part of `formatData()` which is called by `request()` which was async.

libs/langchain-google-common/src/auth.ts
```diff
@@ -58,15 +73,18 @@ export abstract class GoogleAbstractedFetchClient
         `Google request failed with status code ${res.status}: ${resText}`
       );
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      (error as any).response = res;
+      (error as any).details = {
```
jacoblee93, line 76: Good to leave `response` directly on the error object - `AsyncCaller` uses fields on it to decide retries.

libs/langchain-google-common/src/media.ts
```diff
@@ -0,0 +1,775 @@
...
+  }
+
+  async _body(
+    separator: string,
+    data: MediaBlob,
+    metadata: Record<string, unknown>
+  ): Promise<string> {
+    const contentType = data.mimetype;
+    const { encoded, encoding } = await data.encode();
+    const body = [
```
jacoblee93, line 43: Are we sure this is robust? Might be nice to use a library?

libs/langchain-google-common/src/chat_models.ts
```diff
@@ -120,29 +119,30 @@ class ChatConnection<AuthOptions> extends AbstractGoogleLLMConnection<
...
     _parameters: GoogleAIModelParams
-  ): GeminiContent {
+  ): Promise<GeminiContent> {
     if (!this.useSystemInstruction) {
       return {} as GeminiContent;
     }
 
     let ret = {} as GeminiContent;
-    input.forEach((message, index) => {
+    for (let index=0; index<input.length; index+=1) {
```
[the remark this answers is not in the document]
afirstenberg: Doh! Formatted now checked in

libs/langchain-google-common/src/connection.ts
afirstenberg: Typed as any.
(I'm expecting that when I do the claude api stuff that I'll give this a real type.)

afirstenberg:
@jacoblee93 - Issues addressed. I also found a bug when dealing with video files that I've also addressed.

libs/langchain-google-common/src/media.ts
jacoblee93, line 1: Would prefer if this were moved/exported from `experimental` for now

libs/langchain-google-common/src/media.ts
```diff
@@ -0,0 +1,775 @@
...
+    } else {
+      // TODO: Test that you can use OAuth to access
+      return this.buildAbstractedClient(fields);
+    }
+  }
+
+  async _regetMetadata(key: string): Promise<AIStudioFileObject> {
+    // Sleep for some time period
+    // eslint-disable-next-line no-promise-executor-return
+    await new Promise(resolve => setTimeout(resolve, this.retryTime));
```
jacoblee93: Could we just use `AsyncCaller` logic here?

libs/langchain-google-common/src/media.ts
```diff
@@ -0,0 +1,775 @@
...
+
+    // Fetch the latest metadata
+    return this._getMetadata(key);
+  }
+
+  async _set([key, blob]: [string, MediaBlob]): Promise<AIStudioFileSaveResponse> {
+    const response = await super._set([key, blob]) as AIStudioFileSaveResponse;
+
+    // console.log('response.data', response.data);
```
jacoblee93, line 704: Remove log

libs/langchain-google-common/src/media.ts
```diff
@@ -0,0 +1,775 @@
...
+  }
+}
+
+export interface BlobStoreGoogleParams<AuthOptions>
+  extends GoogleConnectionParams<AuthOptions>,
+    AsyncCallerParams,
+    BlobStoreOptions
+  {}
+
+export abstract class BlobStoreGoogle<
```
jacoblee93, line 103: Do we absolutely need a second layer of inheritance?

jacoblee93: Would prefer making this the base level for now until something along the lines of `BlobStore`/`MediaManager` makes it into core CC @eyurtsev

libs/langchain-google-common/src/tests/chat_models.test.ts
```diff
@@ -563,6 +561,196 @@ describe("Mock ChatGoogle", () => {
...
+    expect(parts[1]).toHaveProperty("fileData");
+    expect(parts[1].fileData).toHaveProperty("mimeType");
+    expect(parts[1].fileData).toHaveProperty("fileUri");
+
+    expect(result.content).toBe("A blue square.");
+  });
+
+  test("3. invoke - media - manager", async () => {
+
+    class MemStore extends InMemoryStore<MediaBlob> {
```
jacoblee93, line 655: CC @eyurtsev

libs/langchain-google-webauth/src/tests/chat_models.int.test.ts
```diff
@@ -0,0 +1,247 @@
...
+    const resolver = new SimpleWebBlobStore();
+    const mediaManager = new MediaManager({
+      aliasStore,
+      canonicalStore,
+      resolver,
+    })
+    const model = new ChatGoogle({
+      modelName: "gemini-1.5-flash",
+      apiVersion: "v1beta",
+      mediaManager,
```
jacoblee93, line 212: I think we have to add this to the chat model class above?

Also - thoughts on calling it `googleMediaManager` for now?

libs/langchain-google-common/src/media.ts
```diff
@@ -0,0 +1,775 @@
...
+}
+
+export interface AIStudioFileDownloadConnectionParams<AuthOptions>
+  extends AIStudioFileConnectionParams,
+    GoogleConnectionParams<AuthOptions> {
+  method: GoogleAbstractedClientOpsMethod;
+  name: string;
+}
+
+export class AIStudioFileDownloadConnection<
```
jacoblee93, line 600: Could we rely on an official SDK for some of this or are there the same problems?

libs/langchain-google-common/src/utils/media_core.ts
```diff
@@ -0,0 +1,465 @@
...
+
+  path?: string;
+}
+
+/**
+ * Represents a chunk of data that can be identified by the path where the
+ * data is (or will be) located, along with optional metadata about the data.
+ */
+export class MediaBlob
+  extends Serializable   // FIXME - I'm not sure this serializes or deserializes correctly
```
jacoblee93: This is fine, it'll just serialize as `not_implemented`

libs/langchain-google-common/src/utils/media_core.ts
```diff
@@ -0,0 +1,465 @@
...
+  implements MediaBlobParameters
+{
+  lc_serializable = true;
+
+  lc_namespace = ["langchain", "google-common"]; // FIXME - What should this be? And why?
```
jacoblee93, line 23: Only applicable if we want it to show up in traces (which we maybe do?) and just needs to match a Python equivalent

libs/langchain-google-common/src/utils/media_core.ts
```diff
@@ -0,0 +1,465 @@
...
+  metadata?: Record<string, unknown>;
+
+  path?: string;
+}
+
+/**
+ * Represents a chunk of data that can be identified by the path where the
+ * data is (or will be) located, along with optional metadata about the data.
+ */
+export class MediaBlob
```
jacoblee93, line 17: CC @eyurtsev

libs/langchain-google-common/src/utils/media_core.ts
```diff
@@ -0,0 +1,465 @@
...
+
+  defaultFetchOptions: BlobStoreFetchOptions;
+
+  constructor (opts?: BlobStoreOptions) {
+    super(opts);
+    this.defaultStoreOptions = opts?.defaultStoreOptions ?? {};
+    this.defaultFetchOptions = opts?.defaultFetchOptions ?? {};
+  }
+
+  async _realKey(key: string | MediaBlob): Promise<string> {
```
jacoblee93, line 175: Mark these as `protected` or `private`

libs/langchain-google-common/src/utils/media_core.ts
```diff
@@ -0,0 +1,465 @@
...
+    return ret?.[0] ?? await this._missingFetchBlob(realKey, allOpts);
+  }
+}
+
+export interface BackedBlobStoreOptions extends BlobStoreOptions {
+  backingStore: BaseStore<string, MediaBlob>;
+}
+
+export class BackedBlobStore extends BlobStore {
+  backingStore: BaseStore<string, MediaBlob>;
```
jacoblee93, line 346: Gut feeling is this should be part of the base class

libs/langchain-google-common/src/utils/media_core.ts
```diff
@@ -0,0 +1,465 @@
...
+    return this.backingStore.mset(keyValuePairs);
+  }
+
+  yieldKeys(prefix: string | undefined): AsyncGenerator<string> {
+    return this.backingStore.yieldKeys(prefix);
+  }
+
+}
+
+export class SimpleWebBlobStore extends BlobStore {
```
jacoblee93, line 370: Is this necessary? Why not make this abstract if so?

libs/langchain-google-common/src/utils/media_core.ts
```diff
@@ -0,0 +1,465 @@
...
+ * through the Base64 of the media or through a canonical URI that the LLM
+ * supports.
+ */
+export class MediaManager {
+
+  aliasStore: BlobStore;
+
+  canonicalStore: BlobStore;
+
+  resolver: BlobStore;
```
jacoblee93, line 463: A bit more docs on why three stores are necessary would be good

jacoblee93 (changes requested):
Thanks for this! I think broadly seems reasonable - let's export as experimental and mark as Google specific for now until we're ready to commit to a settled abstraction in core.

CC @eyurtsev for a second review as well

libs/langchain-google-common/src/media.ts
afirstenberg: Perhaps, but this seems like it serves a different purpose than `AsyncCaller` does and has different needs. Tho perhaps I misunderstand exactly how we can use `AsyncCaller` in this case.
* `AsyncCaller` reacts to network or HTTP errors specifically. This needs to react to the value of a returned field. (However, I guess we could change `AsyncCaller.onFailedAttempt` for this.)
* `AsyncCaller` retries immediately. The purpose here is to wait a fixed length of time before trying again. (Changing this would require changes to how we call `pRetry()`.
* `AsyncCaller` has a set number of retries. The best practice for using AI Studio File API is to wait indefinitely until the upload is either ready or has failed. (There are cases where a developer may want to upload and not care immediately if it is ready, and a negative `retryTime` takes this situation into account.

I'm open minded about using either `AsyncCaller` or `pRetry()` if you think it gains us something. I'm just not sure what we'd get in this case.

libs/langchain-google-common/src/media.ts
afirstenberg: I really wish we had a good way to be able to turn these kinds of logs on, however.

libs/langchain-google-common/src/utils/media_core.ts
afirstenberg: I do want to get this to serialize correctly, however. It should be possible to store a `MediaBlob` in an arbitrary thing, but this means telling... something... somehow... how to serialize and deserialize a `Blob`, which is a standard JS thing.

Currently, it calls `Blob.toString()` which is a debugging-quality string.

libs/langchain-google-common/src/utils/media_core.ts
afirstenberg: All of the stuff that I currently have in `media_core.js` are things that I think would make sense to be in `langchain/core` at some point. (ie - they aren't Google specific)

So perhaps wait till @eyurtsev weighs in on how to set the namespace for them.

libs/langchain-google-common/src/utils/media_core.ts
afirstenberg: I want to make clear for @eyurtsev's benefit that the name chosen is `MediaBlob` for two reasons:
1. To distinguish it from a `Blob`, which is already a major class in JS/TS
2. To easily and clearly use a JS/TS `Blob` as a property in this object.

But the goal was to have the same interface structure of a LangChain Python `Blob`.

libs/langchain-google-common/src/utils/media_core.ts
afirstenberg: The `backingStore` attribute?
Not all of the implementations need a separate way to store the data, since some can do it on their own. (For example, a BlobStore for Google Cloud Storage doesn't - it can store both data and metadata for a URL for the store.)
The `BackedBlobStore` class is mostly a way to have a BlobStore that enforces the BlobStore semantics of store/fetch but that is backed by other generic store systems that can store a serialized `MediaBlob` in... whatever permanent means is appropriate without having to write custom versions of all of them.

afirstenberg: (I guess I wouldn't object if it's part of the `BlobStore`. But there are classes where setting it makes no sense.)

libs/langchain-google-common/src/utils/media_core.ts
afirstenberg: Each one is there to address different use cases of the `MediaManager`.
* The `resolver` is so a developer can include a generic URI that doesn't depend on the platform this runs on and get the data and metadata for this.
  * The most common case is providing an HTTP or HTTPS URL.
  * Having a resolver that deals with data: URIs also makes sense.
 * The `canonicalStore` is the store that takes the data (and possibly metadata) and saves it in a platform-dependent way.
  * So folks using AI Studio API would store it in the AI Studio File API. Those who are using Vertex AI would store it in Google Cloud Storage. There is a different BlobStore for each of these, and this is what would be set here.
  * The `MediaBlob` that this returns is the URI that should be sent to the LLM.
* The `aliasStore` should provide a long-term mapping between the two. 
  * So given a generic URI (ie - the HTTPS URL) for an image, get the `MediaBlob` that was saved for this URL already
  * This is important if you're using the URL for a video - you don't want to re-download and then re-upload the entire video every time.

The main use cases:
* The first time through in most interactions, you'll provide the public URL for the image: https://example.com/butterfly.jpg
  * The `resolver` is used to get the image itself
  * The image is handed to the `canonicalStore` that stores it and returns `MediaBlob` that has the path for how to reference it in the platform-specific store
  * This `MediaBlob` is stored in the `aliasStore` using the key of the public URL.
  * The LLM API uses the path from the `MediaBlob` that came from the `canonicalStore`
* Future times through
  * The `aliasStore` is used to get the `MediaBlob` for the public URL
  * If this isn't valid for some reason, it treats it like the first time through
  * If this is valid, the LLM API uses the path from the `MediaBlob` that came from the `aliasStore`

The advantage is that for both cases, all the developer has to think about is what the appropriate store classes should be for their project and then sending a public URL. The `MediaManager` class "does the right thing". 
* For images, the Python class always downloads the contents at the public URL to include it inline.
* For media, the current implementation (in both Python and JS) requires the developer to make sure the media has been uploaded correctly already and to provide the equivalent of the URL from the canonical store, rather than the public URL.

We can also evaluate some less common use cases (where the URL format from the canonical store is used, for example) to see how these might be supported, but currently those are handled using the legacy support that the LLM already supports.

Going forward, it might make sense to think about how to take a `MediaBlob` directly (for locally generated images, for example, or images downloaded from an image generator), but I'm not sure we have a good way to do that in general now.

Hopefully that is clearer, but I'm happy to discuss further to make sure we've identified all the critical cases.

libs/langchain-google-common/src/media.ts
afirstenberg: Currently there are two different BlobStores from Google (and it is arguable there might be a third) that take advantage of the existing `AuthOptions` mechanism that we have in place.

But there are also BlobStores that are unrelated to the Google auth mechanism itself. They provide the `MediaBlob` store/fetch semantics, but don't need the Google auth frameworks.

I'm not opposed to collapsing a level, but I think at the moment that we gain some advantages from the abstraction. Happy to discuss further.

libs/langchain-google-common/src/media.ts
afirstenberg: Relying on the official SDK means importing another package, and I'm not sure what it gets us.

libs/langchain-google-common/src/utils/media_core.ts
afirstenberg: `SimpleWebBlobStore` is meant to be a simple BlobStore used as a retriever for the `MediaManager` for basic http/https requests. It deliberately doesn't try to do everything (hence the "simple" part), but does need to be functional.

eyurtsev:
@afirstenberg I'm reviewing this PR, should be done by Monday -- need a bit more time to think through the abstractions in media_core

afirstenberg:
> @afirstenberg I'm reviewing this PR, should be done by Monday -- need a bit more time to think through the abstractions in media_core

@eyurtsev - Understood, and I appreciate the review! Let me know if there is anything I can clarify or that you want to discuss.

libs/langchain-google-common/src/experimental/utils/media_core.ts
```diff
@@ -0,0 +1,527 @@
+import { v1, v4 } from "uuid"; // FIXME - it is importing the wrong uuid, so v6 and v7 aren't implemented
```
afirstenberg, line 1: The version of uuid it is loading is from langchainjs/node_modules/@types, which still seems to be an older version, despite moving to the latest version in the main branch. So I've commented out the v6 and v7 support for now.

libs/langchain-google-common/src/experimental/utils/media_core.ts
```diff
@@ -0,0 +1,527 @@
...
+  async *yieldKeys(_prefix: string | undefined): AsyncGenerator<string> {
+    this._notImplementedException();
+    yield "";
+  }
+}
+
+export interface MediaManagerConfiguration {
+  /**
+   * A map from the alias name to the canonical MediaBlob for that name
+   */
```
eyurtsev, line 462: is aliasStore identical to a cache or different?

eyurtsev:
Hi @afirstenberg, 

Awesome! Love the PR! There are a number of useful ideas in the core abstraction, that we'll carry over to the python side.

I left some smaller questions directly on the PR, but would love to discuss the high level abstraction:

1. Do developers need an "admin" API; e.g., to clean up all content older than a month?
2. Is a website allowed as an input? i.e., to fetch the HTML content of the page as context for the LLM? If so how would this work for a webpage that changes frequently (e.g., `news.google.com`)
3. Would a readhtrough abstraction help simplify the media manager while meeting all requirements?

This involves the following:

* Introduce a readthrough cache that has a hasher attached to it
* we'd need to change the the `mset` interface to allow it to return values -- this would allow the underlying store to return a blob with an updated URI (e.g., pointing to an google bucket)
* (note) The underlying blob store itself could be a readthrough store, which would allow maintaining an LRU in-memory readthrough cache if ever needed

```typescript

type MediaBlob = {
};

type BaseStore<K, V> = {
  get(key: K): V | undefined;
  set(key: K, value: V): V;
};

class ReadthroughStore implements BaseStore<string, Blob> {
  /**
   * Media manager for the blob store.
   */

  hasher: (uri: string) => string;
  /**
   * Function to hash URI into a hash.
   *
   * This function may need to make a network request in some cases to fetch 
   * the e-tag.
   * 
   * This function can be written generically to check the format of the URI
   * and implement different behavior whether it's a URL or a data URI.
   */

  fetcher: (uris: string[]) => (Blob | undefined)[];
  /**
   * A function that resolves URIs to blobs.
   */

  underlyingBlobstore: BaseStore<string, Blob>;
  /**
   * Underlying blob store where blobs can be stored.
   */

  constructor(
    hasher: (uri: string) => string,
    fetcher: (uris: string[]) => (Blob | undefined)[],
    underlyingBlobstore: BaseStore<string, Blob>
  ) {
    this.hasher = hasher;
    this.fetcher = fetcher;
    this.underlyingBlobstore = underlyingBlobstore;
  }

  get(key: string): Blob | undefined {
    /**
     * Get a blob by key.
     */
    const hash = this.hasher(key);
    let blob = this.underlyingBlobstore.get(hash);
    if (blob === undefined) {
      const blobs = this.fetcher([key]);
      blob = blobs[0];
      if (blob !== undefined) {
        this.underlyingBlobstore.set(hash, blob);
      }
    }
    return blob;
  }

  set(key: string, value: Blob): Blob {
    /**
     * Set a blob by key.
     */
    const hash = this.hasher(key);
    return this.underlyingBlobstore.set(hash, value);
  }
}
```

libs/langchain-google-common/src/experimental/utils/media_core.ts
```diff
@@ -0,0 +1,504 @@
...
+
+/**
+ * Represents a chunk of data that can be identified by the path where the
+ * data is (or will be) located, along with optional metadata about the data.
+ */
+export class MediaBlob
+  extends Serializable // FIXME - I'm not sure this serializes or deserializes correctly
+  implements MediaBlobParameters
+{
+  lc_serializable = true;
```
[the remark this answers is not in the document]
afirstenberg: It won't.

The problem is that I feel like it needs to.

In `BackedBlobStore` it can take a `BaseStore` with the intent that it can store arbitrary string -> `MediaBlob` mappings. While an in-memory one is easy, much more difficult would be one that stores a disk or db cache since they would need the `MediaBlob` to be serializeable.

So the question is how to make something that implements `Serializable` that will, somehow, turn the binary Blob contents into something that can be serialized and be read into a Blob later on the JS side or... something... on the python side.

libs/langchain-google-common/src/experimental/utils/media_core.ts
```diff
@@ -0,0 +1,504 @@
...
+  metadata?: Record<string, unknown>;
+
+  path?: string;
+}
+
+/**
+ * Represents a chunk of data that can be identified by the path where the
+ * data is (or will be) located, along with optional metadata about the data.
+ */
+export class MediaBlob
```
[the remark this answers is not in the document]
afirstenberg: I did try to model it after the `Blob` in Python. Not sure why I left the ID out of it. No objection to restoring it on this side in an attempt to keep the two compatible.

libs/langchain-google-common/src/experimental/utils/media_core.ts
```diff
@@ -0,0 +1,504 @@
...
+  lc_serializable = true;
+
+  lc_namespace = ["langchain", "google-common"]; // FIXME - What should this be? And why?
+
+  data?: Blob;
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  metadata?: Record<string, any>;
+
+  path?: string;
```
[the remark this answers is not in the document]
afirstenberg: It is left to be an arbitrary string, tho most of the way I had been using it is as a URI of some sort. I think it would make sense to expand the notion to allow it to be a file system path as well.

libs/langchain-google-common/src/experimental/utils/media_core.ts
```diff
@@ -0,0 +1,504 @@
...
+    opts: BlobStoreStoreOptions = {}
+  ): Promise<MediaBlob | undefined> {
+    const allOpts: BlobStoreStoreOptions = {
+      ...this.defaultStoreOptions,
+      ...opts,
+    };
+    const validBlob = await this._validStoreBlob(blob, allOpts);
+    if (typeof validBlob !== "undefined") {
+      const validKey = await validBlob.asUri();
+      await this.mset([[validKey, validBlob]]);
```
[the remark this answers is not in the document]
afirstenberg: VERY!

I was trying to stay out of making any changes to the core package itself.
(Tho everything in media_core.ts is intended to eventually go into core.)

But if we're ok with changing the definition of `mset()` to return what was set (which may not be the same as what was passed as a parameter), then this code would be a lot cleaner.

libs/langchain-google-common/src/experimental/utils/media_core.ts
```diff
@@ -0,0 +1,504 @@
...
+    return new MediaBlob({ path });
+  }
+
+  protected async _missingFetchBlob(
+    path: string,
+    opts?: BlobStoreFetchOptions
+  ): Promise<MediaBlob | undefined> {
+    switch (opts?.actionIfBlobMissing) {
+      case "emptyBlob":
+        return this._missingFetchBlobEmpty(path, opts);
```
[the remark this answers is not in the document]
afirstenberg: I don't remember exactly my reasoning, but I wanted to make it easy for subclasses or developers to do either based on what they needed. 

I think my thought process was that there were cases where the path needed to be preserved in a `MediaBlob`, even if there was no data that went along with it.

libs/langchain-google-common/src/experimental/utils/media_core.ts
```diff
@@ -0,0 +1,504 @@
...
+    const ret = await this.mget([realKey]);
+    return ret?.[0] ?? (await this._missingFetchBlob(realKey, allOpts));
+  }
+}
+
+export interface BackedBlobStoreOptions extends BlobStoreOptions {
+  backingStore: BaseStore<string, MediaBlob>;
+}
+
+export class BackedBlobStore extends BlobStore {
```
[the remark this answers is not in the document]
afirstenberg: Yes.

I use it in tests for `MediaManager` for an `aliasStore` that is backed by a memory store.
I can also see it used in general for either the `aliasStore` or a retriever that will be using some other defined store to store things. I don't want people to need to re-invent the storage methodology to support the `BlobStore` semantics, but I do want to be able to support them.

libs/langchain-google-common/src/experimental/utils/media_core.ts
```diff
@@ -0,0 +1,504 @@
...
+  mset(keyValuePairs: [string, MediaBlob][]): Promise<void> {
+    return this.backingStore.mset(keyValuePairs);
+  }
+
+  yieldKeys(prefix: string | undefined): AsyncGenerator<string> {
+    return this.backingStore.yieldKeys(prefix);
+  }
+}
+
+export class SimpleWebBlobStore extends BlobStore {
```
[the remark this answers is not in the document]
afirstenberg: A fair point.

However, we have code that is already doing this in the Gemini implementations. (See https://github.com/langchain-ai/langchain-google/blob/7c678dcc59bb4e4d7a1efdbb9572d74b75e87665/libs/genai/langchain_google_genai/_image_utils.py#L137 for where it is done in Python, for example.)

Making this a separate class puts controls of the URL retrieval in the hands of the developer. the `SimpleWebBlobStore` is just what it says - simple. But it can serve as a base for developers to make subclasses (that do things such as authentication if necessary) that still work against a "normal" http/https URL.

And we do need *something* that does the fetch for the `MediaManager` so that Gemini can work the same as OpenAI models (where you're expected to pass a public URL for images). Making it have the same signature as other `BlobStore`s makes sense for flexibility.

libs/langchain-google-common/src/experimental/utils/media_core.ts
afirstenberg: I want to say that yes, it is a cache.
Or at least I'm having a difficult time trying to articulate how it isn't.

I think the big difference is that the cache invalidation (or determining if a `MediaBlob` is valid or not) is handled as part of the `MediaManager` rather than the cache system itself.

libs/langchain-google-common/src/experimental/utils/media_core.ts
```diff
@@ -0,0 +1,504 @@
...
+    switch (opts?.actionIfBlobMissing) {
+      case "emptyBlob":
+        return this._missingFetchBlobEmpty(path, opts);
+      default:
+        return undefined;
+    }
+  }
+
+  async fetch(
+    key: string | MediaBlob,
```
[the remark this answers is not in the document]
afirstenberg: I think my thinking here was that, eventually, it would make sense for developers to be able to pass MediaBlobs around directly (ie - data that comes from image generation) without having to save it to an intermediary step.

I am not strongly committed to this.

libs/langchain-google-common/src/experimental/utils/media_core.ts
```diff
@@ -0,0 +1,504 @@
...
+
+export class SimpleWebBlobStore extends BlobStore {
+  _notImplementedException() {
+    throw new Error("Not implemented for SimpleWebBlobStore");
+  }
+
+  _hasValidPath(
+    blob: MediaBlob,
+    _opts?: BlobStoreStoreOptions
+  ): Promise<boolean> {
```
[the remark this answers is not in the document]
afirstenberg: Good question. It shouldn't. Especially with the comment about how it shouldn't be making network calls.

libs/langchain-google-common/src/experimental/utils/media_core.ts
```diff
@@ -0,0 +1,504 @@
...
+    const allOpts: BlobStoreStoreOptions = {
+      ...this.defaultStoreOptions,
+      ...opts,
+    };
+    const validBlob = await this._validStoreBlob(blob, allOpts);
+    if (typeof validBlob !== "undefined") {
+      const validKey = await validBlob.asUri();
+      await this.mset([[validKey, validBlob]]);
+      const savedKey = await validBlob.asUri();
+      return await this.fetch(savedKey);
```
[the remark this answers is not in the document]
afirstenberg: Having `mset()` return a value (as we discuss below) would almost certainly solve this.

This works around a need with the AI Studio File API implementation where the object returned from the API call contains the path, which is the crucial bit of the whole thing.

eyurtsev:
@afirstenberg My proposal would be to remove alias store from the media manager, and only keep one store.

Reasoning:

1. A media manager with a single persistent store (e.g., s3 bucket or google cloud), and an associated fetcher (something that scrapes data) is useful abstraction.
2. The store can be treated as a readthrough cache that implements the store interface. So it'll make it easy to layer on another cache on top (LRU in memory cache)

The required modifications:

1. mset needs to be able to return the Blob
2. All mget usage in the stores is done by the hash associated with the content
3. A store needs to be able to return the Blob by either reference or value. Returning by reference means populating the `path` variable properly (e.g., that's what the FileAPI will do)
3. We introduce a ReadthroughCache 

```python
class ReadthroughStore(BaseStore[str, MediaBlob]):
    def __init__(self, store: BaseStore[str, MediaBlob], backing_store: BaseStore[str, MediaBlob]) -> None:
        self.store = store
        self.backing_store = backing_store

    def mget(self, keys: Sequence[str]) -> List[Optional[MediaBlob]]:
        assert len(keys) == 1
        key = keys[0]
        result = self.store.mget([key])
        if result[0] is not None:
            return result
        else:
            value = self.backing_store.mget([key])
            if value[0] is not None:
                new_value = self.store.mset([(key, value[0])])
                return [new_value]
        return [None]
```

Then the file store and cloud store look like this:

```python
class FileAPI(BaseStore[str, MediaBlob]):
    def __init__(self):
        self.client = client

    def mget(self, keys: Sequence[str]) -> List[Optional[MediaBlob]]:
        assert len(keys) == 1
        key = keys[0]
        url = "url_from_file_api/{}".format(key)
        if self.client.exists(key):
            return [
                MediaBlob(
                    path=url,
                )
            ]
        else:
            return [None]

    def mset(self, items: Sequence[Tuple[str, MediaBlob]]) -> MediaBlob:
        assert len(items) == 1
        key, value = items[0]
        url = "url_from_file_api/{}".format(key)
        self.client.upload(url, value)
        return MediaBlob(
            path=url,
        )


class CloudStorage(BaseStore[str, MediaBlob]):
    def __init__(self):
        self.cloud_client = ...

    def mget(self, keys: Sequence[str]) -> List[Optional[MediaBlob]]:
        assert len(keys) == 1
        url = "some_other_url_url_from_file_api/{}".format(key)
        content = self.client.get(url)
        if content:
            return [
                MediaBlob(
                    data=content,  # <-- Can return data and path?
                    path=url,
                )
            ]
        else:
            return [None]

    def mset(self, items: Sequence[Tuple[str, MediaBlob]]) -> MediaBlob:
        url = "url_from_file_api/{}".format(key)
        self.client.upload(url, value)

```

afirstenberg:
> @afirstenberg My proposal would be to remove alias store from the media manager, and only keep one store.
> 
> Reasoning:
> 
> 1. A media manager with a single persistent store (e.g., s3 bucket or google cloud), and an associated fetcher (something that scrapes data) is useful abstraction.
> 2. The store can be treated as a readthrough cache that implements the store interface. So it'll make it easy to layer on another cache on top (LRU in memory cache)

@eyurtsev - I *think* I follow. Working things out to make sure we have all the use cases.

I *think* there is a problem with how you describe the `ReadThroughStore` above - it assumes that the `mget()` can somehow get a value from the backing store if it isn't present in the store. That isn't true with the AI Studio File API (which won't have the data). I *think* it is solveable by having `mset()` send things to the backing store first and caching its result in the store. See the flows I outline below.

So under this scheme, the`MediaManager` object would take two `BlobStore` classes:
* A resolver (such as a `SimpleWebBlobStore`?) that takes a public URL and returns a `MediaBlob` that has the URL as the path and the data as the contents from that path.
* A cache/store, which would be a `ReadThroughStore`
  * The store would be something like an `InMemoryStore` or a `LocalFileStore` or something that stores it on a db or whatever. It takes a (possibly public) URL and returns a `MediaBlob` from the cache or whatever that has the LLM-usable  URI as the path.
  * The backing store would be something like `BlobStoreAIStudioFile` or `BlobStoreGoogleCloudStorage` that takes a `MediaBlob` and returns a (new? modified?) `MediaBlob` that has the path set to a URI that the LLM can use to reference it.

The data flows for `MediaManager` would be:
* First time through asking for https://example.com/example.png
  * Getting from the cache returns nothing.
  * Sends the URL to the resolver, getting back resolved_blob
  * Saves the resolved_blob to the cache
    * This is where I think the code you outlined needs to change. So it would call `mset` on the store
    * This calls `mset` on the backing store which saves it, gets the new URI (say... example://001.png) and returns a new `MediaBlob`
    * It then saves this in the store against the URI from the resolved_blob
  * Returns the stored_blob which has the URI that the LLM API can use in a media part
* Next time through asking for https://example.com/example.png
  * Getting from the cache returns the stored_blob with URI example://001.png
  * The `MediaManager` verifies stored_blob and confirms that it is valid (the metadata says it hasn't expired, for example)
  * Returns the stored_blob which has the URI that the LLM API can use in a media part
* Sometime later, asking for https://example.com/example.png
  * Getting from the cache returns the stored_blob with URI example://001.png
  * The `MediaManager` verifies stored_blob and determines it has expired
  * Sends the URL to the resolver, getting back resolved_blob
  * Saves the resolved_blob to the cache, replacing the old blob
    * I would call `mset` on the store
    * This calls `mset` on the backing store which saves it, gets the new URI (say... example://002.png) and returns a new `MediaBlob`
    * It then saves this in the store against the URI from the resolved_blob, removing the old stored_blob
  * Returns the new_stored_blob which has the URI that the LLM API can use in a media part

Does all of this sound right to you?
Are there any use cases we've missed?

eyurtsev:
@afirstenberg yeah that's exactly correct -- thanks for catching the mset error. 

I was also thinking about providing a single hasher in the media manager (so the url gets mapped to a UUID once), but we can handle that logic inside of the stores as well. The only implication there is that yield_keys() will produce an iterable over the hashed keys not the keys that were passed as part of mset (which is fine).

eyurtsev:
I also don't know if the equivalent exists on the JS side but we have an encoder backed store: https://github.com/langchain-ai/langchain/blob/88588466071a9dde3a4c26c813e30b9db7842eca/libs/langchain/langchain/storage/encoder_backed.py#L20-L20

This allows wrapping existing stores that do `BaseStore[str, bytes]` with a serializer for the values (e.g., a Document) and a hasher for the keys

bracesproul:
> I also don't know if the equivalent exists on the JS side but we have an encoder backed store: https://github.com/langchain-ai/langchain/blob/88588466071a9dde3a4c26c813e30b9db7842eca/libs/langchain/langchain/storage/encoder_backed.py#L20-L20
> 
> This allows wrapping existing stores that do `BaseStore[str, bytes]` with a serializer for the values (e.g., a Document) and a hasher for the keys

[`BaseStore<K, V>`](https://v02.api.js.langchain.com/classes/langchain_core_stores.BaseStore.html)

afirstenberg:
@eyurtsev @jacoblee93 @bracesproul - I think this is ready (again) for review to be merged.

Some notes based on our earlier plans and discussions:
* `MediaBlob` has been changed to not use `Blob` internally. There was just too much work to get it to handle `Blob` correctly and still serialize. Interfaces were all kept. (@jacoblee93 and @bracesproul - Can you look at this particularly to make sure `Serializable` was implemented correctly. It passes my tests... but...)
* `BaseStore` doesn't implement an `mset()` that returns the value set because this would be a breaking change in core.
* `BlobStore` returns the value from `store()` and the `ReadThroughBlobStore` takes advantage of this using the algorithm that we discussed above.
* Everything remains in experimental.

I'd like to get this merged if we think it's reasonable, even if not perfect, and **not** cut a new version. I'd like to move on to some of the Vertex Model Garden support, and the changes in here in Connection are necessary for that.

libs/langchain-google-common/src/experimental/media.ts
```diff
@@ -0,0 +1,809 @@
...
+    AsyncCallerCallOptions,
+    ResponseType,
+    AuthOptions
+  >;
+
+  async _set(keyValuePair: [string, MediaBlob]): Promise<ResponseType> {
+    const [_key, blob] = keyValuePair;
+    const setMetadata = this.buildSetMetadata(keyValuePair);
+    /*
+    const metadata = {
```
jacoblee93, line 144: nit: remove

libs/langchain-google-common/src/experimental/utils/media_core.ts
jacoblee93: Weird - we can also update globally

libs/langchain-google-common/src/experimental/utils/media_core.ts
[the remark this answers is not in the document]
jacoblee93: This is all it does:

https://github.com/langchain-ai/langchainjs/blob/main/langchain-core/src/load/serializable.ts#L147

What happens if you call `JSON.stringify` on a blob?

libs/langchain-google-common/src/experimental/utils/media_core.ts
```diff
@@ -0,0 +1,504 @@
...
+ * Represents a chunk of data that can be identified by the path where the
+ * data is (or will be) located, along with optional metadata about the data.
+ */
+export class MediaBlob
+  extends Serializable // FIXME - I'm not sure this serializes or deserializes correctly
+  implements MediaBlobParameters
+{
+  lc_serializable = true;
+
+  lc_namespace = ["langchain", "google-common"]; // FIXME - What should this be? And why?
```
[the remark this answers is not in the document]
jacoblee93: Agreed, let's just leave this unset for now. It'll default to something based on path IIRC

It should be something in core eventually

libs/langchain-google-common/src/experimental/utils/media_core.ts
```diff
@@ -0,0 +1,674 @@
...
+  data?: MediaBlobData;
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  metadata?: Record<string, any>;
+
+  path?: string;
+
+  constructor(params?: MediaBlobParameters | Serialized) {
+    const sparams = params as SerializedConstructor;
+    super(sparams?.kwargs);
```
jacoblee93, line 52: This is a bit odd - you can just pass `params`

If it extends `Serializable`, any object params you pass will automatically be serialized

libs/langchain-google-common/src/experimental/utils/media_core.ts
```diff
@@ -0,0 +1,674 @@
...
+
+    const mparams = params as MediaBlobParameters;
+    this.data = mparams?.data || this.lc_kwargs?.data;
```
jacoblee93, line 55: `lc_kwargs` doesn't have any innate value, it's just set based on what you pass to `super`:

https://github.com/langchain-ai/langchainjs/blob/main/langchain-core/src/load/serializable.ts#L144

libs/langchain-google-common/src/experimental/utils/media_core.ts
```diff
@@ -0,0 +1,674 @@
...
+    this.metadata = mparams?.metadata || this.lc_kwargs?.metadata;
+    this.path = mparams?.path || this.lc_kwargs?.path;
+  }
+
+  get lc_attributes() {
```
jacoblee93, line 60: Don't need to set this if you pass it into `super` (which you're doing)

libs/langchain-google-common/src/experimental/utils/media_core.ts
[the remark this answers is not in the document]
jacoblee93: I don't really understand this one - why not just interact with `this.backingStore` directly?

If just used for tests, put in a `utils/testing` file

jacoblee93:
Left a few comments, happy to merge in experimental if you resolve the merge conflict!

libs/langchain-google-common/src/experimental/utils/media_core.ts
afirstenberg: That is what I thought. And yet, it wasn't serializing until I explicitly set these.

This is why I was so confused about serializing MediaBlob (and why I wanted attention to this part specifically). It didn't seem to be serializing or deserializing as I expected it to.

libs/langchain-google-common/src/experimental/utils/media_core.ts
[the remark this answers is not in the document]
afirstenberg: Because the `BlobStore` versions have different semantics for storing an object (they return values from a set while the `BaseStore` versions don't). So rather than reimplementing all of the `BaseStore` implementations (now and in the future), it guarantees those semantics for other `BaseStore` types.  A memory store being the biggest example, but anything that may save to a database or whatever also makes sense.

libs/langchain-google-common/src/experimental/utils/media_core.ts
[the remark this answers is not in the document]
afirstenberg: Unsetting it gives an error that `lc_namespace` is abstract and must be implemented.

libs/langchain-google-common/src/experimental/utils/media_core.ts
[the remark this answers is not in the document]
afirstenberg: You get an empty object. `{}`
I've gotten rid of `Blob` at this point since trying to serialize it to a base64 string is an async operation, and that would break all sorts of things.
Instead - there is a utility function to create a `MediaBlob` from a `Blob`, but the implementation of `MediaBlob` is all stuff that is routinely serializable.

libs/langchain-google-common/src/experimental/utils/media_core.ts
afirstenberg: The global version was updated, I thought. But it was still pulling in the wrong one for reasons I didn't understand

afirstenberg:
Code was merged and the conflict resolved.
I think things look reasonable for a merge and NOT push a new release.

libs/langchain-google-common/src/experimental/utils/media_core.ts
jacoblee93: So should remove this?

libs/langchain-google-common/src/experimental/utils/media_core.ts
jacoblee93: You can remove

libs/langchain-google-common/src/experimental/utils/media_core.ts
jacoblee93: Why the cast and `.kwargs`? Should probably be a type guard if anything?

But also why `| Serialized`?

jacoblee93 (changes requested):

jacoblee93:
Would prefer to remove some of the unnecessary (I believe?) serialization pieces before merging so we don't forget about them

jacoblee93 (approved):

PR #5842, opened 2024-06-21
community[minor]: JinaAI embeddings integration

AvaterClasher:
Description

Implemented a new embedding integration that makes JinaAI's embedding models available on langchainjs. A new class named JinaEmbeddings is created, which wraps the inference API.

This PR includes an example and integration tests related to the newly implemented class. Apart from the new integration's configuration, the original open-source components are intentionally left unchanged to avoid introducing any breaking changes.


jacoblee93:
Thank you!

PR #5852, opened 2024-06-21
community[minor]: Update bedrock to accept openai formatted tools

bracesproul:


libs/langchain-community/src/chat_models/tests/chatbedrock.int.test.ts
```diff
@@ -383,3 +385,103 @@ test.skip.each([
 
   expect(res.content.length).toBeGreaterThan(1);
 });
+
+test.skip("withStructuredOutput", async () => {
```
jarib, line 389: @bracesproul Do you plan to add `withStructuredOutput` as well, since you added this skipped tests? I was just poking around trying to understand why it's not currently supported, given that tool calling in general appears to work well with Claude on Bedrock.

PR #5881, opened 2024-06-24
langchain[patch]: fix:pass runManager to DocumentCompressor in MultiQueryRetriever

jm-szlendak:
The MultiQueryRetriever can accept an instance of DocumentCompressor, but when executing compression - it will not pass the run manager and effectively, any chains executed inside DocumentCompressor will not get callbacks executed. 

This PR fixes this issue.


jacoblee93:
Thank you!

PR #5885, opened 2024-06-25
core[patch]: fix: Document should handle empty pageContent (fixes #5884)

### LINKED ISSUE #5884: Document.pageContent should not allow undefined
### Checked other resources

### Example Code

```typescript
import { describe, it, expect } from 'vitest'
import { Document } from '@langchain/core/documents'

describe('langchain Type Compatibility Test', () => {
  it('Document should handle empty pageContent', () => {
    const doc = new Document({ pageContent: '' })
    expect(doc.pageContent).toStrictEqual('')
  })
})
```

AssertionError: expected undefined to strictly equal ''
Expected :
Actual   :undefined

### Description

pageContent should be set to '', as mandated by its type contract `string` (and not `string|undefined`)

The bug is visible in the code
 https://github.com/langchain-ai/langchainjs/blob/main/langchain-core/src/documents/document.ts
```typescript
export class Document<
  Metadata extends Record<string, any> = Record<string, any>
> implements DocumentInput, DocumentInterface
{
  pageContent: string;

  metadata: Metadata;

  constructor(fields: DocumentInput<Metadata>) {
    this.pageContent = fields.pageContent
      ? fields.pageContent.toString()
      : this.pageContent;
    this.metadata = fields.metadata ?? ({} as Metadata);
  }
}
```

Line 32 should check field.pageContent is undefined, not just truthy, otherwise it uses this.pageContent which is undefined and breaks the contract that pageContent must be a string. Thus in any event, pageContent needs to be initialised to something to fulfill the type contract

### System Info

This will happen on any environment

glorat:
Fixes #5884

### DIFF
--- a/langchain-core/src/documents/document.ts
+++ b/langchain-core/src/documents/document.ts
@@ -20,8 +20,8 @@ export interface DocumentInterface<
  * Interface for interacting with a document.
  */
 export class Document<
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  Metadata extends Record<string, any> = Record<string, any>
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    Metadata extends Record<string, any> = Record<string, any>
 > implements DocumentInput, DocumentInterface
 {
   pageContent: string;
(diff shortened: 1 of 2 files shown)


jacoblee93:
This technically breaks the interface - why do we need to check for this?

jacoblee93:
Thank you! Generally against this kind of defensive check but since it's already there might as well fix it.

PR #5895, opened 2024-06-25
ci[patch]: Adds npm install test for conflicting deps

jacoblee93:
`npm` has rather severe peer dependency resolution requirements, this will verify that we are not violating any of them.

CC @bracesproul


bracesproul (approved):
lgtm, I'm assuming you've tested locally since this won't run in this PR

jacoblee93:
> lgtm, I'm assuming you've tested locally since this won't run in this PR

Hah, fixed. I want this to run on every PR.

PR #5905, opened 2024-06-26
community[patch]: cheerio loader fetch headers type error

versole:
Fix: cheerio loader headers type is not fetch request headers


bracesproul (commented):
Good catch! Thanks for pushing up this commit

PR #5914, opened 2024-06-27
ci[patch]: fix Unit Tests (LangChain Integrations) Skipping Test Jobs

rameshkrishna:
Unit Tests (LangChain Integrations) Workflow is Skipping Jobs due if condition failure. 

Fixes Issue #5913


bracesproul (commented):
Cab you push up some temp commits in an integration to verify this change works as expected?

libs/langchain-cloudflare/src/vectorstores.ts
```diff
@@ -53,7 +53,7 @@ export class CloudflareVectorizeStore extends VectorStore {
     const { index, textKey, ...asyncCallerArgs } = args;
     if (!index) {
       throw new Error(
-        "Must supply a Vectorize index binding, eg { index: env.VECTORIZE }"
+        "Must supply a Vectorize indexx bindingg, eg { index: env.VECTORIZE }"
```
bracesproul, line 56:
```suggestion
        "Must supply a Vectorize index binding, eg { index: env.VECTORIZE }"
```

bracesproul (approved):
lgtm, thank you!

PR #5916, opened 2024-06-27
core[patch]: support image url prompt with mustache

a1ooha:
Fix bug with image url prompt with mustache.

image


bracesproul (approved):
lgtm, thanks!

PR #5919, opened 2024-06-27
openai[patch]: Set stream_usage to false by default for Azure

jacoblee93:
CC @ccurme


deranga:
@jacoblee93 it seems that streamUsage is broken in 0.2.8, and its not possible to set streamUsage when instantiating the class. Can't see how its related to the commit attached to this, but couldn't find any other obvious culprits in 0.2.7 or 0.2.8 releases.

jacoblee93:
Hey @deranga sorry about this - this would be part of `@langchain/openai` here, which is on 0.2.1:

https://www.npmjs.com/package/@langchain/openai

PR #5928, opened 2024-06-28
cohere[patch]: update embeddings and rerank

Anirudh31415926535:
This PR makes the model parameter mandatory for CohereEmbeddings class and the CohereRerank class.
Along with that, adds a embed method for the CohereEmbeddings class, and some minor improvements and docs updates.


libs/langchain-cohere/src/embeddings.ts
```diff
@@ -70,8 +75,16 @@ export class CohereEmbeddings
       token: apiKey,
     });
     this.model = fieldsWithDefaults?.model ?? this.model;
+
+    if (!this.model) {
```
jacoblee93, line 79: Would prefer to just make this a required param and ship a new minor version release along with your tool calling stuff

libs/langchain-cohere/src/embeddings.ts
```diff
@@ -70,8 +75,16 @@ export class CohereEmbeddings
...
+      throw new Error(
+        "Model not specified for CohereEmbeddings instance. Please provide a model name from the options here: https://docs.cohere.com/reference/embed"
+      );
+    }
+
     this.batchSize = fieldsWithDefaults?.batchSize ?? this.batchSize;
-    this.inputType = fieldsWithDefaults?.inputType;
+    this.embeddingTypes =
+      fieldsWithDefaults?.embeddingTypes ?? this.embeddingTypes;
```
jacoblee93, line 87: Do we need to remove `inputType` from `CohereEmbeddingsParams` as well?

libs/langchain-cohere/src/embeddings.ts
```diff
@@ -137,6 +154,25 @@ export class CohereEmbeddings
     }
   }
 
+  async embed(
```
jacoblee93, line 157: Why add this? Would prefer not to extend base class' API unnecessarily.

jacoblee93:
Thanks, see comments!

Let's gear up for a minor version release along with your other PR.

libs/langchain-cohere/src/embeddings.ts
Anirudh31415926535: This method is supported in langchain-python, and also allows to override the inputType parameter to call the embed function

PR #5961, opened 2024-07-02
redis[minor]: Uses string instead of list in RedisChatHistory

itamarhaber:
This is a premature optimization after having reviewed the code. The same logic applies to the Python API.

Given that the history is append-only, read-or-delete-all, using a Redis String instead of a List is cheaper and more performant.

This PR affects the proposed change, and also uses a transaction for atomicity if TTL needs to be set.


jacoblee93:
Seems reasonable - will need to be part of a minor bump though.

jacoblee93:
Actually - I think better to stick to the same format as the Python library. Some people do use both libs and expect consistent serialization. Will ping you on Slack.

PR #5962, opened 2024-07-02
Changes history insertion order to avoid a `reverse` operation

itamarhaber:
This is a premature optimization after having reviewed the code. I believe the same logic applies to the Python API.

Instead of storing the list in reverse order, and then re-reversing on read, we can append to the end of the list.

This PR affects the proposed change.
Note that it isn't required if #5961 is accepted.


jacoblee93:
Let's go with #5961 if anything

PR #5963, opened 2024-07-02
redis[patch]: Adds a connection pool to RedisChatHistory

itamarhaber:
This is a premature optimization after having reviewed the code.

When supplied with a config instead of a client instance, a new connection will be opened each time a history is loaded.

Instead, we can reuse opened clients by keeping them in a "pool".

This PR affects the proposed change.


jacoblee93:
Nice, thanks!

jacoblee93:
Actually - I think better to stick to the same format as the Python library. Some people do use both libs and expect consistent serialization. Will ping you on Slack.

PR #5969, opened 2024-07-03
multiple[patch]: upgrade uuid version

chentschel:
Upgrades current uuid version to support v6 generation.


jacoblee93:
Thank you!

PR #5989, opened 2024-07-05
Fix memory issue for large number of child chunks on large documents

clemenspeters:
Fixes OOM issues. The docker container of our index manager is killed on specific documents after using more than 2GB of memory. For large documents the ParentDocumentRetriever can create lots of parent and child documents. 

While this change seems to fix the memory issues, I'm still investigating another issue on the same large document at the moment: `Uncaught exception: error: terminating connection due to idle-session timeout`.

@PetersClemens

### DIFF
--- a/langchain/src/retrievers/parent_document.ts
+++ b/langchain/src/retrievers/parent_document.ts
@@ -135,6 +135,18 @@ export class ParentDocumentRetriever extends MultiVectorRetriever {
     return parentDocs.slice(0, this.parentK);
   }
 
+  async _storeDocuments(parentDoc: Document, childDocs: Document[], addToDocstore: boolean) {
+    if (this.childDocumentRetriever) {
+      await this.childDocumentRetriever.addDocuments(childDocs);
+    }
+    else {
+      await this.vectorstore.addDocuments(childDocs);
+    }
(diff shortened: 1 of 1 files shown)


langchain/src/retrievers/parent_document.ts
```diff
@@ -197,16 +207,7 @@ export class ParentDocumentRetriever extends MultiVectorRetriever {
...
-      fullDocs[parentDocId] = parentDoc;
-    }
-    if (this.childDocumentRetriever) {
-      await this.childDocumentRetriever.addDocuments(embeddedDocs);
-    } else {
-      await this.vectorstore.addDocuments(embeddedDocs);
-    }
-    if (addToDocstore) {
-      await this.docstore.mset(Object.entries(fullDocs));
+      await this._storeDocuments(parentDoc, taggedSubDocs, addToDocstore);
```
jacoblee93, line 210: Flagging that this does make it a bit more likely that we'll get into half-finished states

jacoblee93 (approved):
Thank you!

jacoblee93:
Couldn't push to your branch so merged in #6012 after running `yarn format`

PR #5992, opened 2024-07-07
core[minor]: Use run tree as stored value in async local storage

jacoblee93:


langchain-core/src/runnables/config.ts
```diff
@@ -1,31 +1,13 @@
-import {
-  type BaseCallbackConfig,
-  CallbackManager,
-  ensureHandler,
-} from "../callbacks/manager.js";
+import { CallbackManager, ensureHandler } from "../callbacks/manager.js";
 import { AsyncLocalStorageProviderSingleton } from "../singletons/index.js";
+import { RunnableConfig } from "./types.js";
```
eyurtsev, line 3: Python location is in config.py rather than types.py

jacoblee93: Exposed in the same places, this is just to avoid a circular import

langchain-core/src/tracers/tracer_langchain.ts
```diff
@@ -147,4 +147,49 @@ export class LangChainTracer
       return undefined;
     }
   }
+
+  convertToRunTree(id: string): RunTree | undefined {
+    // create a run tree from a run map
+    const runTreeMap: Record<string, RunTree> = {};
+    const runTreeList: [id: string, dotted_order: string | undefined][] = [];
+    for (const [id, run] of this.runMap) {
+      // TODO: this loses object reference equality
```
jacoblee93, line 156: Can we compare by id or some other identifier instead?

langchain-core/src/singletons/index.ts
```diff
@@ -21,14 +23,52 @@ const mockAsyncLocalStorage = new MockAsyncLocalStorage();
 class AsyncLocalStorageProvider {
   getInstance(): AsyncLocalStorageInterface {
     return (
-      (globalThis as any).__lc_tracing_async_local_storage ??
+      (globalThis as any).__lc_tracing_async_local_storage_v2 ??
```
jacoblee93, line 26: @dqbd should we use a symbol here too?

PR #6000, opened 2024-07-08
partners[patch]: Fix flaky token counting tests

bracesproul:


libs/langchain-anthropic/src/tests/chat_models.int.test.ts
```diff
@@ -13,6 +13,16 @@ import {
...
 
+async function sleep(ms = 1000): Promise<void> {
+  return new Promise<void>((resolve) => {
+    setTimeout(resolve, ms);
+  });
+}
+
+afterEach(async () => {
+  await sleep();
+});
```
bracesproul, line 24: b/c anthropic rate limits. will remove if we can get our limits increased

bracesproul: removed here https://github.com/langchain-ai/langchainjs/pull/6003

PR #6005, opened 2024-07-08
core[minor],openai[patch],langchain[patch]: Allow tool functions to input `ToolCall` / return `ToolMessage`

bracesproul:
TODO:
- update tool return type to be: string default. ToolMessage if a tool call is passed in.
  - ideally: if tool call is passed and tool call id is passed, return ToolMessage, else return str

Followup PR will wrap all integrations with the `createToolCall`/`createToolCallChunk` func

Tools can now:
- take `ToolCall` as an input.
- return `ToolMessage`

Updated generics to reflect that.
Added a third generic to `tool` func wrapper if you want to specify for the input arg of the `func` field. (it will infer if no generics were prev passed)


langchain-core/src/callbacks/base.ts
```diff
@@ -197,7 +198,7 @@ abstract class BaseCallbackHandlerMethodsClass {
    * Called at the end of a Tool run, with the tool output and the run ID.
    */
   handleToolEnd?(
-    output: string,
+    output: string | ToolMessage,
```
eyurtsev, line 201: I think this should be an any type -- a tool should be allowed to return anything, and we shouldn't be limiting the return type in anyway
```suggestion
    output: any,
```

jacoblee93: How would we handle things that aren't a string or tool message? Type check and e.g. JSON.stringify?

PR #6007, opened 2024-07-08
core[minor]: Add raw_output field to ToolMessage

bracesproul:


langchain-core/src/messages/tool.ts
```diff
@@ -8,6 +8,16 @@ import {
...
 
 export interface ToolMessageFieldsWithToolCallId extends BaseMessageFields {
+  /**
+   * The raw output of the tool.
+   *
+   * **Not part of the payload sent to the model.** Should only be specified if it is
+   * different from the message content, i.e. if only a subset of the full tool output
+   * is being passed as message content.
+   *
+   * `raw_output` is excluded when serializing the tool message.
```
jacoblee93, line 18: Why?

jacoblee93:
We'll populate this where relevant in a followup PR?

langchain-core/src/messages/tool.ts
bracesproul: was a copy pasta from PY but looks like they dropped that.

PR #6029, opened 2024-07-10
core[minor]: RunnableToolLike

bracesproul:
Adds new `RunnableToolLike` class for converting tools -> runnables.
Updates base `bindTools` implementation to take `RunnableToolLike`
Updates all integrations which implement `bindTools` to also accept `RunnableToolLike`
Updates convert to oai tool/function to also accept `RunnableToolLike` (this should mean no additional work is required to accept this tool type in integrations)

docs pr: https://github.com/langchain-ai/langchainjs/pull/6030
part of #6031


langchain-core/package.json
```diff
@@ -74,6 +74,7 @@
     "prettier": "^2.8.3",
     "release-it": "^15.10.1",
     "rimraf": "^5.0.1",
+    "ts-jest": "^29.1.0",
```
jacoblee93, line 77: What's this?

langchain-core/src/runnables/base.ts
```diff
@@ -2783,3 +2804,83 @@ export class RunnablePick<
...
+  name: string;
+
+  description?: string;
+
+  schema: RunInput;
+
+  config?: RunnableConfig;
+}
+
+export class RunnableToolLike<
```
jacoblee93, line 2821: Should this extend `RunnableLambda` instead?

langchain-core/src/runnables/base.ts
```diff
@@ -2783,3 +2804,83 @@ export class RunnablePick<
...
+  RunInput extends z.ZodType = z.ZodType,
+  RunOutput = unknown
+> extends RunnableBinding<z.infer<RunInput>, RunOutput> {
+  lc_runnable_tool_like: true;
```
jacoblee93, line 2825: We already have and use `lc_name`:

```
  static lc_name() {
    return "RunnableToolLike";
  }
```

Then do a constructor check in the type guard, no need for a new property here

langchain-core/src/runnables/base.ts
```diff
@@ -2783,3 +2804,83 @@ export class RunnablePick<
...
+ * @param {string | undefined} [fields.name] The name of the tool. If not provided, it will default to the name of the runnable.
+ * @param {string | undefined} [fields.description] The description of the tool. If not provided, it will default to `Takes {schema}` where `schema` is a JSON string representation of the input schema.
+ * @param {z.ZodType<RunInput>} [fields.schema] The Zod schema for the input of the tool. Infers the Zod type from the input type of the runnable.
+ * @returns {RunnableToolLike<z.ZodType<RunInput>, RunOutput>} An instance of `RunnableToolLike` which is a runnable that can be used as a tool.
+ */
+export function convertRunnableToTool<RunInput, RunOutput>(
+  runnable: Runnable<RunInput, RunOutput>,
+  fields: {
+    name?: string;
+    description?: string;
```
jacoblee93, line 2871: Why is this optional?

langchain-core/src/runnables/base.ts
```diff
@@ -2783,3 +2804,83 @@ export class RunnablePick<
...
+  }
+}
+
+/**
+ * Generate a placeholder description of a runnable
+ */
+const _getDescriptionFromRunnable = <RunInput = unknown>(
+  schema: RunInput
+): string => {
+  return `Takes ${JSON.stringify(schema, null, 2)}`;
```
jacoblee93, line 2851: This is self-evident from the schema. Description should be required or default to empty string

jacoblee93:
Thanks! See comments.

Fewer changes than I'd have thought.

langchain-core/src/runnables/base.ts
bracesproul: Spoke to nuno and he suggested `RunnableBinding`, which a few other runnables extend too

langchain-core/src/runnables/base.ts
bracesproul: We're able to generate those from the runnable, matches py

langchain-core/src/runnables/base.ts
bracesproul: Agree with everything except empty string default. Pushed PR to just be optional.

Issue #6045, opened 2024-07-11
ChatAnthropic streaming with tools not working and instead forces non-stream response


Moe03:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain.js documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain.js rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

The following is a more advanced example but it is to showcase the full problem, I've also implemented a fix for it in a pr.
```ts
import { HumanMessage, SystemMessage } from "@langchain/core/messages";
import { ChatAnthropic } from "@langchain/anthropic";
import { DynamicStructuredTool } from "@langchain/core/tools";
import { z } from "zod";
import { ChatOpenAI } from "@langchain/openai";
import { AgentExecutor, createToolCallingAgent } from "langchain/agents";
import {
    ChatPromptTemplate,
    MessagesPlaceholder
} from "@langchain/core/prompts";

async function main() {

    const defaultTools = [
        new DynamicStructuredTool({
            name: "GetWeather",
            description: "Return the weather.",
            schema: z.object({
                query: z.string().describe("The city to return weather for,."),
            }),
            func: async (input) => {
                console.log(`STARTING TOOL: GetWeather`)
                // just an example
                return 'sunny';
            },
        }),
    ]

    const chatPrompt = ChatPromptTemplate.fromMessages([
        new MessagesPlaceholder("messages"),
        new MessagesPlaceholder("agent_scratchpad"),
    ]);

    const messagesHistory = [
        new SystemMessage("you are a helpful assistant!"),
        new HumanMessage("Whats the weather like in Cairo?")
    ]

    let llm;

    if (false) {
        llm = new ChatOpenAI({
            model: 'gpt-4o',
            apiKey: OAI_API_KEY,
            // streaming: true,
        });
    } else {
        llm = new ChatAnthropic({
            apiKey: ANTHROPIC_API_KEY,
            model: "claude-3-5-sonnet-20240620",
        });
    }

    const agent = await createToolCallingAgent({
        llm,
        tools: defaultTools,
        prompt: chatPrompt,
    });

    const agentExec = new AgentExecutor({
        agent,
        tools: defaultTools,
    }).withConfig({ runName: "Agent" });

    let llmStream = await agentExec.streamEvents(
        {
[...]

PR #6047, opened 2024-07-11
Fix ChatAnthropic forcing non-streamed response with tools

Moe03:
Have adjusted the ChatAnthopic class to enable streaming with tools properly instead of it previously forcing any request with tools to be non-streamed, also works perfectly with agentExecutor with was our usecase that we had to find a workaround for.

I've kept the prvious if statment code for maintainers to have as reference/edit however they want, also fixed little typos in readme: libs/langchain-anthropic

Fixes # (issue)
https://github.com/langchain-ai/langchainjs/issues/6045


Moe03:
I think this pr is ready now, we'll be using it in production at https://voiceglow.org :)

jacoblee93:
Hey @Moe03! Thanks for this PR - we have a specific way we're tackling it in #6056 and #6057 so will close this for now. But much appreciated!

PR #6056, opened 2024-07-12
anthropic[minor]: Implement actual anthropic tool call streaming

bracesproul:
Depends on https://github.com/langchain-ai/langchainjs/pull/6057 merging & releasing

Adds proper streaming tool call support to `ChatAnthropic`. Now it will
- yield tool call chunks as they come in from anthropic
- populate `tool_calls` in the final message with the tool calls

For backcompat:
- still yields tool calls in anthropic format inside the `content` field with `type: "tool_use"`
- `JSON.parse()`s the tool call input args inside the `content` block and only yield the final chunk so the `input` field contains valid JSON


langchain-core/src/messages/base.ts
bracesproul, line 1: Ignore this file, this is merged from https://github.com/langchain-ai/langchainjs/pull/6057

DevDeepakBhattarai:
It seems this exact feature is also missing from ChatGroq and ChatVertexAI. When ever we call the models with tools they don't stream.

jacoblee93:
Thanks for flagging - can you open an issue?

I think when we initially implemented these integration packages, streaming tool calls weren't supported, but if they are now we should support.

wh1pp3rz:
This is now broken when using `streamEvents` with ChatAnthropic, it;s giving the warning: `The "tool_calls" field on a message is only respected if content is a string.` and throws the exception: `Error: Unsupported message content format`.

DevDeepakBhattarai:
> Thanks for flagging - can you open an issue?
> 
> I think when we initially implemented these integration packages, streaming tool calls weren't supported, but if they are now we should support.

Sure, I have opened #6099

danny-avila:
Still having streaming issues with ChatAnthropic using streamEvents and v2 in particular.

I saw an open issue in the langgraph repo so I posted there. This may be involve `@langchain/anthropic` or how streamEvents is handling a bound ChatAnthropic.

More details:
https://github.com/langchain-ai/langgraphjs/issues/253#issuecomment-2246384664

PR #6057, opened 2024-07-12
core[minor]: Allow concatenation of messages with multi part content

bracesproul:
Port of [#22002](https://github.com/langchain-ai/langchain/pull/22002/)


jacoblee93:
Thank you!

PR #6059, opened 2024-07-13
community[minor] HuggingFaceTransformersEmbeddings enhancements 

toboid:
`HuggingFaceTransformersEmbeddings` uses a `feature-extraction` pipeline from `@xenova/transformers`. 
This change adds support for providing arguments (`pretrainedOptions`) to the creation of the pipeline and when executing the pipeline (`pipelineOptions`). Also bumps `@xenova/transformers` to the latest.

- Upgrade @xenova/transformers peer dep from `^2.5.4` -> `^2.17.2`
- Add optional `pretrainedOptions` to constructor args object `HuggingFaceTransformersEmbeddingsParams`. This is passed to @xenova/transformers when creating a pipeline.
- Add optional `pipelineOptions`  to constructor args object `HuggingFaceTransformersEmbeddingsParams`. This is passed to the pipeline when executing it. There prev. arguments which where hardcoded remain the default for backwards compatibility.


bracesproul (approved):
Nice! Thanks for pushing this up, looks like a good improvement to add.

Additionally, it would be great if you could add a small section in the [docs](https://js.langchain.com/v0.2/docs/integrations/text_embedding/transformers) showing this off, but not required.

Granted CI passes, I'll land this EOD today.

jacoblee93:
Thank you!

toboid:
> Nice! Thanks for pushing this up, looks like a good improvement to add.
> 
> Additionally, it would be great if you could add a small section in the [docs](https://js.langchain.com/v0.2/docs/integrations/text_embedding/transformers) showing off the new options/an example showing why people might want to use them, but not required.
> 
> Granted CI passes, I'll land this EOD today.

Thanks for landing. Yes sure, i'll submit a follow-up with docs

PR #6061, opened 2024-07-15
langchain[patch]: Bug fix for ParentDocumentRetriever

clemenspeters:
Fix for bug introduced in https://github.com/langchain-ai/langchainjs/pull/5989 (sorry about this).

Fixes error: The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received undefined TypeError [ERR_INVALID_ARG_TYPE]: The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received undefined

@PetersClemens

### DIFF
--- a/langchain/src/retrievers/parent_document.ts
+++ b/langchain/src/retrievers/parent_document.ts
@@ -136,7 +136,7 @@ export class ParentDocumentRetriever extends MultiVectorRetriever {
   }
 
   async _storeDocuments(
-    parentDoc: Document,
+    parentDoc: Record<string, Document>,
     childDocs: Document[],
     addToDocstore: boolean
   ) {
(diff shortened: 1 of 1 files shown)


langchain/src/retrievers/parent_document.ts
```diff
@@ -210,7 +210,11 @@ export class ParentDocumentRetriever extends MultiVectorRetriever {
             metadata: { ...subDoc.metadata, [this.idKey]: parentDocId },
           })
       );
-      await this._storeDocuments(parentDoc, taggedSubDocs, addToDocstore);
+      await this._storeDocuments(
+        { parentDocId: parentDoc },
```
bracesproul, line 214: Did you mean this?
```suggestion
        { [parentDocId]: parentDoc },
```

Without the `[]` it will make the key `parentDocId`. By wrapping it in brackets it'll use the actual value of the `parentDocId` variable.

bracesproul (commented):
One comment, other than that this fix lgtm

langchain/src/retrievers/parent_document.ts
clemenspeters: Thanks for the thorough review @bracesproul 🙏 
I guess there should really be some test for this 🙈

bracesproul: Could you add a test for this method? Would help catch future issues.

jacoblee93:
Oof thanks

PR #6062, opened 2024-07-15
core[patch]: Adds support for custom events

jacoblee93:


langchain-core/src/callbacks/base.ts
```diff
@@ -259,6 +260,19 @@ abstract class BaseCallbackHandlerMethodsClass {
     tags?: string[]
   ): // eslint-disable-next-line @typescript-eslint/no-explicit-any
   Promise<any> | any;
+
+  handleCustomEvent?(
+    eventName: string,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    payload: any,
```
eyurtsev, line 267: `payload` vs. `data`?

langchain-core/src/callbacks/base.ts
```diff
@@ -259,6 +260,19 @@ abstract class BaseCallbackHandlerMethodsClass {
...
+    runId: string,
+    parentRunId?: string,
+    tags?: string[],
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    metadata?: Record<string, any>,
+    name?: string
```
eyurtsev, line 273: What is name?

langchain-core/src/callbacks/base.ts
```diff
@@ -259,6 +260,19 @@ abstract class BaseCallbackHandlerMethodsClass {
     tags?: string[]
   ): // eslint-disable-next-line @typescript-eslint/no-explicit-any
   Promise<any> | any;
+
+  handleCustomEvent?(
+    eventName: string,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    payload: any,
+    runId: string,
+    parentRunId?: string,
```
eyurtsev, line 269: Drop parentRunId b/c user can already define it, but we actually don't provide the correct one

langchain-core/src/callbacks/dispatch/web.ts
```diff
@@ -0,0 +1,68 @@
...
+        "inside a tool or a RunnableLambda).",
+        `\n\nIf you continue to see this error, please import from "@langchain/core/callbacks/dispatch/web"`,
+        "and explicitly pass in a config parameter.",
+      ].join(" ")
+    );
+  }
+  await callbackManager.handleCustomEvent?.(
+    name,
+    payload,
+    parentRunId,
```
eyurtsev, line 62:
```suggestion
    parentRunId, // Add a doc-string here to say it's being passed as the run id
```

langchain-core/src/callbacks/manager.ts
```diff
@@ -860,6 +860,46 @@ export class CallbackManager
     );
   }
 
+  async handleCustomEvent?(
+    eventName: string,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    payload: any,
+    runId: string,
+    _parentRunId?: string,
```
eyurtsev, line 868: Remove the `parentRunId` on the dispatch?

langchain-core/src/callbacks/base.ts
jacoblee93: `runName`

langchain-core/src/callbacks/base.ts
jacoblee93: These are just positional args, will show in tooltips but not affect anything else.

jacoblee93: Will update anyway

PR #6089, opened 2024-07-16
core[minor]: Add RemoveMessage class

bracesproul:
For LangGraph
Port of https://github.com/langchain-ai/langchain/pull/23636


jacoblee93:
Ah nice thanks

PR #6133, opened 2024-07-18
community[minor], azure-cosmosdb[major]: add new integration for Azure CosmosDB for NoSQL

sinedied:
### @langchain/community
- chore: deprecate Azure CosmosDB for MongoDB vCore vector store in favor of the new package integration

### @langchain/azure-cosmosdb

- feat: new package
- feat: add updated Azure CosmosDB for MongoDB vCore integration (new name: `AzureCosmosDBMongoDBvCoreVectorStore` instead of `AzureCosmosDBVectorStore`
- feat: add new Azure CosmosDB for NoSQL integration

### docs

- docs: updated docs and examples to use the new @langchain/azure-cosmosdb package


aayush3011:
@sinedied We need to add support for "OFFSET LIMIT" query filters as well. And they can only be used together for NoSQL, not separately.

And for the Mongo vCore integration can we also add support for vector-HNSW and make it as default. vCore also added filtered vector search. 

[HNSW](https://learn.microsoft.com/en-us/azure/cosmos-db/mongodb/vcore/vector-search#create-a-vector-index-using-hnsw)
[vCore Filtered Vector Search](https://learn.microsoft.com/en-us/azure/cosmos-db/mongodb/vcore/vector-search#filtered-vector-search-preview)

sinedied:
@aayush3011 Thanks for the feedback!

I've updated the NoSQL filters integration to allow OFFSET LIMIT in addition to WHERE clause, by making it more generic.

Regarding HNSW for vCore, this can be implemented in a separate PR, as this one is focused on CosmosDB NoSQL integration (and it already quite big 😅 ). The vCore integration here was just moved from the @langchain/community package, no features were changed.

sinedied:
Hi @jacoblee93 could you have a look? This PR should be complete now.

Sorry for the size, but it should be easier to update then with a dedicated package.

jacoblee93:
Yes will do thank yoU!

TODO
```diff
@@ -0,0 +1,37 @@
+# TODO:
```
jacoblee93, line 1: Could you remove this file?

docs/core_docs/docs/integrations/platforms/microsoft.mdx
```diff
@@ -100,20 +100,36 @@ See a [usage example](/docs/integrations/vectorstores/azure_aisearch).
...
 ```
 
+### Azure Cosmos DB for NoSQL
+
+> [Azure Cosmos DB for NoSQL](https://learn.microsoft.com/azure/cosmos-db/nosql/) provides support for querying items with flexible schemas and native support for JSON. It now offers vector indexing and search. This feature is designed to handle high-dimensional vectors, enabling efficient and accurate vector search at any scale. You can now store vectors directly in the documents alongside your data. Each document in your database can contain not only traditional schema-free data, but also high-dimensional vectors as other properties of the documents. 
+
+<IntegrationInstallTooltip></IntegrationInstallTooltip>
+
+```bash npm2yarn
+npm install @langchain/azure-cosmosdb
```
jacoblee93, line 110: Depending on how many integrations you're planning - do you think it makes sense to have a single `@langchain/azure` package with peer dependencies on different packages?

libs/langchain-azure-cosmosdb/package.json
```diff
@@ -0,0 +1,93 @@
...
+    "format": "prettier --config .prettierrc --write \"src\"",
+    "format:check": "prettier --config .prettierrc --check \"src\"",
+    "move-cjs-to-dist": "yarn lc-build --config ./langchain.config.js --move-cjs-dist",
+    "create-entrypoints": "yarn lc-build --config ./langchain.config.js --create-entrypoints",
+    "check-tree-shaking": "yarn lc-build --config ./langchain.config.js --tree-shaking"
+  },
+  "author": "LangChain",
+  "license": "MIT",
+  "dependencies": {
+    "@azure/cosmos": "4.0.1-beta.3",
```
jacoblee93, line 41: Should this be looser?

libs/langchain-azure-cosmosdb/src/azure_cosmosdb_mongodb_vcore.ts
```diff
@@ -0,0 +1,483 @@
...
+ * - the `mongodb` NPM package installed
+ * - a connection string associated with a MongoDB VCore Cluster
+ *
+ * You do not need to create a database or collection, it will be created
+ * automatically.
+ *
+ * You also need an index on the collection, which is by default be created
+ * automatically using the `createIndex` method.
+ */
+export class AzureCosmosDBMongoDBvCoreVectorStore extends VectorStore {
```
jacoblee93, line 76: This is a very intimidating name - how about just `CosmosDBvCoreVectorStore`

libs/langchain-azure-cosmosdb/src/azure_cosmosdb_nosql.ts
```diff
@@ -0,0 +1,522 @@
...
+
+  /**
+   * Initializes the AzureCosmosDBNoSQLVectorStore by connecting to the database.
+   * @param client The CosmosClient to use for connecting to the database.
+   * @param databaseName The name of the database to use.
+   * @param containerName The name of the collection to use.
+   * @param initOptions Initialization options for the database and container.
+   * @returns A promise that resolves when the AzureCosmosDBNoSQLVectorStore has been initialized.
+   */
+  private async init(
```
jacoblee93, line 437: Might be good to expose this as `initialize`?

libs/langchain-azure-cosmosdb/src/azure_cosmosdb_mongodb_vcore.ts
```diff
@@ -0,0 +1,483 @@
...
+  }
+
+  /**
+   * Initializes the AzureCosmosDBMongoDBvCoreVectorStore by connecting to the database.
+   * @param client The MongoClient to use for connecting to the database.
+   * @param databaseName The name of the database to use.
+   * @param collectionName The name of the collection to use.
+   * @returns A promise that resolves when the AzureCosmosDBMongoDBvCoreVectorStore has been initialized.
+   */
+  private async init(
```
jacoblee93, line 411: See below

jacoblee93:
A fair bit of Cosmos DB specific things I'm not familiar with so am less helpful with best practices around that but left a few comments!

libs/langchain-azure-cosmosdb/package.json
sinedied: Currently only the latest beta SDK has support for the vector search, so I'll update it when the final release is available. AFAIK beta versions don't update anyways, so I don't think making it looser would change anything.

docs/core_docs/docs/integrations/platforms/microsoft.mdx
sinedied: If you're ok with that, I think smaller packages might be more manageable and easier to maintain, as opposed to one big package. I don't think we'll create too many packages, though I don't have the plans of all product teams 🙂

libs/langchain-azure-cosmosdb/src/azure_cosmosdb_mongodb_vcore.ts
sinedied: I agree with the long name being intimidating 😅 
Though I'm not sure it can be much better here, since vCore is a *tier* of the CosmosDB *MongoDB* API. So vCore without MongoDB might conflict with another API tier.

`AzureCosmosDBMongoDBVectorStore` might be the shortest we could do, as supporting a different tier could eventually be done with an option in case we need it. WDYT?

libs/langchain-azure-cosmosdb/src/azure_cosmosdb_nosql.ts
sinedied: Done. I've added a public `initialize` method, and defer its call to the first time it's needed, so it can either be explicitly called after constructor, or it will be called automatically when any of the other methods needs it.

libs/langchain-azure-cosmosdb/src/azure_cosmosdb_mongodb_vcore.ts
sinedied: @jacoblee93 I've renamed everything to `AzureCosmosDBMongoDBVectorStore`, not the best but still slightly better, and coherent with the other integration. Hope it's fine like this?

sinedied:
@jacoblee93 I've updated the PR and answered all your comments, let me know if it's good for you?

libs/langchain-azure-cosmosdb/src/azure_cosmosdb_mongodb_vcore.ts
jacoblee93: Sure! Generally up to you, just wanted to raise it

libs/langchain-azure-cosmosdb/src/azure_cosmosdb_mongodb.ts
```diff
@@ -0,0 +1,496 @@
...
+    const client = dbConfig.client || this.client!;
+    const databaseName = dbConfig.databaseName ?? "documentsDB";
+    const collectionName = dbConfig.collectionName ?? "documents";
+    this.indexName = dbConfig.indexName ?? "vectorSearchIndex";
+    this.textKey = dbConfig.textKey ?? "textContent";
+    this.embeddingKey = dbConfig.embeddingKey ?? "vectorContent";
+    this.indexOptions = dbConfig.indexOptions ?? {};
+
+    // Deferring initialization to the first call to `initialize`
+    this.initialize = () => {
```
jacoblee93, line 145: Would prefer this as a static method like this:

https://v02.api.js.langchain.com/classes/langchain_community_vectorstores_pgvector.PGVectorStore.html#initialize

Since I'm not sure this will show up well in API refs, but won't block on it

jacoblee93:
Thanks! It's failing lint + format but I can have a look.

Just FYI - we're also doing a bunch of work right now to standardize our integration docs/make our API refs more usable:

https://js.langchain.com/v0.2/docs/integrations/vectorstores/pinecone

jacoblee93:
Thank you!

jacoblee93:
Live here: https://www.npmjs.com/package/@langchain/azure-cosmosdb

sinedied:
> Thanks! It's failing lint + format but I can have a look.
> 
> Just FYI - we're also doing a bunch of work right now to standardize our integration docs/make our API refs more usable and maintainable:
> 
> https://js.langchain.com/v0.2/docs/integrations/vectorstores/pinecone

Thanks! I'll look into updating the docs to the new format in the next PR!

libs/langchain-azure-cosmosdb/src/azure_cosmosdb_mongodb.ts
sinedied: Would you be ok with a middle ground approach?
- make the current `initialize` method I made internal
- add a new static `initialize` method the calls constructor + init/connect the db connection
- keep the "automatic" init when calling any VS method in case it wasn't done before (for compatibility / simplicity)
- explain how the init works in the docs

My reasoning: I'd like to make the migration from the previous implementation painless (ie keep the auto init) and also make it work for folks who skim quickly through docs and just call the constructor without using the `initialize` method

PR #6142, opened 2024-07-18
scripts[patch]: Fix import path for build script

bracesproul:
`dist_build` is not shipped in the released package code


bracesproul:
Merged in #6188

PR #6145, opened 2024-07-19
docs[patch]: Typo fixed in Upstash Ratelimit Callback

Ibkdir:
In the Upstash Ratelimit Callback docs, `limiter` should be set to `Ratelimit.fixedWindow(10, "60 s")` instead of `Ratelimit.fixedWindow(1, "60 s")`.

This is because the comment above it says "10 requests per window, where window size is 60 seconds:". The current docs have it as 1 request every 60 seconds

If you visit the Upstash docs, you see the correct syntax is as follows:

UpstashDocs


bracesproul (approved):
Thank you!

PR #6149, opened 2024-07-19
docs[minor]: Replace deprecated code examples with LCEL & AgentExecutor

Srijan-D:
Description: Added LangChain v0.2 support For the LangChain AgentExecutor using react agent executor which manages a state that is defined by a list of messages in a similar way to the AgentExecutor.  Removed classes as well like LLMChain which were deprecated from LangChain v0.1.

Used the LangChain Expression Language (LCEL) instead of LLMChain deprecated class


bracesproul (approved):
nice! Thanks for doing this

bracesproul (commented):
hey looks like the format CI check is failing, could you run `yarn format` inside `./examples` and push up those changes files

Srijan-D:
Sure let me check :)

Srijan-D:
Done I have pushed it.

PR #6150, opened 2024-07-19
docs[patch]: remove duplicate content from sql QA tutorial page

Sachin-chaurasiya:
This PR will remove the duplicate content from the SQL QA tutorial doc page.

**Before**
image

**After**

image


bracesproul (approved):
Thank you for catching and fixing this!

PR #6171, opened 2024-07-22
langchain[minor]: Generic chat models

bracesproul:
Todo:
- [x] tests
- [x] docs

Updated packages which had dependencies on `langchain` to no longer depend. All instances had `langchain` as a dep for tests, so no functionality was lost by doing this. In most cases it was for agent tests, so we need to figure out a new way to test integrations with agents.

I am aware langchain latest/lowest deps CI test is failing, will need a slight refactor


docs/core_docs/docs/how_to/index.mdx
```diff
@@ -71,11 +71,13 @@ These are the core building blocks you can use when building applications.
 - [How to: cache model responses](/docs/how_to/chat_model_caching)
 - [How to: create a custom chat model class](/docs/how_to/custom_chat)
 - [How to: get log probabilities](/docs/how_to/logprobs)
+- [How to: let your end users choose their model](/docs/how_to/chat_models_universal_init/)
```
baskaryan, line 74: would delete, we just forgot to in py

bracesproul, line 74:
```suggestion
```

langchain/langchain.config.js
```diff
@@ -65,6 +65,8 @@ export const config = {
     "chains/retrieval": "chains/retrieval",
     "chains/sql_db": "chains/sql_db/index",
     "chains/graph_qa/cypher": "chains/graph_qa/cypher",
+    // chat models
+    "chat_models": "chat_models/base",
```
jacoblee93, line 69: We talked about this before

langchain/package.json
```diff
@@ -2166,5 +2225,8 @@
       "require": "./schema/prompt_template.cjs"
     },
     "./package.json": "./package.json"
+  },
+  "resolutions": {
```
jacoblee93, line 2229: What is this?

langchain/src/chat_models/base.ts
```diff
@@ -0,0 +1,864 @@
...
+  | "azure_openai"
+  | "cohere"
+  | "google_vertexai"
+  | "google_genai"
+  | "ollama"
+  | "together"
+  | "fireworks"
+  | "mistralai"
+  | "groq"
+  | "bedrock";
```
jacoblee93, line 47: You don't need to list these twice - something like `export type ChatModelProvider = typeof keyof _SUPPORTED_PROVIDERS` should work

langchain/src/chat_models/base.ts
```diff
@@ -0,0 +1,864 @@
...
+  "together",
+  "fireworks",
+  "mistralai",
+  "groq",
+  "bedrock",
+];
+
+// This is needed so when `bindTools` is called on a configurable model, the typing is correct.
+abstract class BaseChatModelWithBindTools extends BaseChatModel {
+  abstract override bindTools(
```
jacoblee93, line 66: If we have to do this, it should throw if the inner model doesn't support `.bindTools`

langchain/src/chat_models/base.ts
```diff
@@ -0,0 +1,864 @@
...
+    | ToolDefinition
+    | RunnableToolLike
+  )[];
+}
+
+async function _initChatModelHelper(
+  model: string,
+  modelProvider?: string,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  kwargs: Record<string, any> = {}
```
jacoblee93, line 102: Let's move away from calling it `kwargs`, let's do `params` instead

langchain/src/chat_models/base.ts
```diff
@@ -0,0 +1,864 @@
...
+ *     model: "claude-3-5-sonnet-20240620",
+ *   },
+ * });
+ * ```
+ *
+ * @example Create a fully configurable model with a default model and a config prefix
+ * ```typescript
+ * import { initChatModel } from "langchain/chat_models";
+ *
+ * const configurableModelWithDefault = await initChatModel("gpt-4", {
```
jacoblee93, line 694: Why not `openai.gpt-4`?

jacoblee93: Just feels odd to have such an important param alongside random config

langchain/src/chat_models/base.ts
```diff
@@ -0,0 +1,864 @@
...
+      }
+    })();
+
+    return IterableReadableStream.fromReadableStream(stream.readable);
+  }
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export interface InitChatModelFields extends Partial<Record<string, any>> {
+  modelProvider?: string;
```
jacoblee93, line 572: Why is this optional?

langchain/src/chat_models/base.ts
```diff
@@ -0,0 +1,864 @@
...
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          initializedModel = await (initializedModel as any)[method](...args);
+        }
+      }
+    }
+
+    return initializedModel;
+  }
+
+  bindTools(
```
jacoblee93, line 325: `override`

langchain/src/chat_models/base.ts
```diff
@@ -0,0 +1,864 @@
...
+  configPrefix?: string;
+  /**
+   * Methods which should be called after the model is initialized.
+   * The key will be the method name, and the value will be the arguments.
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  queuedMethodOperations?: Record<string, any>;
+}
+
+class _ConfigurableModel<
```
jacoblee93, line 249: Wait lol what is this

langchain/src/chat_models/base.ts
```diff
@@ -0,0 +1,864 @@
...
+    options: Partial<CallOptions> & {
+      version: "v1" | "v2";
+      encoding?: "text/event-stream" | undefined;
+    },
+    streamOptions?: Omit<EventStreamCallbackHandlerInput, "autoClose">
+  ): IterableReadableStream<StreamEvent | Uint8Array> {
+    const stream = new TransformStream();
+    const writer = stream.writable.getWriter();
+
+    void (async () => {
```
jacoblee93, line 552: Use an iterator

langchain/src/chat_models/base.ts
```diff
@@ -0,0 +1,864 @@
...
+      config: mergedConfig,
+      bound: newConfigurableModel,
+    });
+  }
+
+  async invoke(
+    input: RunInput,
+    options?: CallOptions
+  ): Promise<AIMessageChunk> {
+    const model = await this._model(options);
```
jacoblee93, line 430: nit: why are we reloading this every time?

libs/langchain-aws/src/tests/embeddings.int.test.ts
```diff
@@ -58,27 +58,25 @@ test("Test BedrockEmbeddings.embedDocuments with passed region and credentials",
...
-  expect(resultOneMetadatas).toEqual([{ id: 2 }]);
-
-  const resultTwo = await vectorStore.similaritySearch("hello world", 2);
-  const resultTwoMetadatas = resultTwo.map(({ metadata }) => metadata);
-  expect(resultTwoMetadatas).toEqual([{ id: 2 }, { id: 3 }]);
-
-  const resultThree = await vectorStore.similaritySearch("hello world", 3);
-  const resultThreeMetadatas = resultThree.map(({ metadata }) => metadata);
-  expect(resultThreeMetadatas).toEqual([{ id: 2 }, { id: 3 }, { id: 1 }]);
+// TODO: langchain dependency breaks CI. Should add a `FakeVectorStore` in core & import here to fix.
```
jacoblee93, line 61: Probably worth doing now

langchain/langchain.config.js
bracesproul: ah shit, sorry about that - blindly copying the py file path 🙃

langchain/package.json
bracesproul: For the circular dep, not sure I need it anymore though. Will try without it

langchain/src/chat_models/base.ts
bracesproul: This class extends `Runnable` so I don't think that is necessary

langchain/src/chat_models/base.ts
bracesproul: Runtime call options require it

langchain/src/chat_models/base.ts
bracesproul: can be passed as runtime arg

langchain/src/chat_models/base.ts
bracesproul: discussed offline, marking resolved

langchain/src/chat_models/base.ts
bracesproul: discussed offline, marking resolved

langchain/src/chat_models/base.ts
bracesproul: Now we need it lol

langchain/src/chat_models/base.ts
bracesproul: discussed offline, marking resolved

langchain/langchain.config.js
```diff
@@ -65,6 +65,8 @@ export const config = {
     "chains/retrieval": "chains/retrieval",
     "chains/sql_db": "chains/sql_db/index",
     "chains/graph_qa/cypher": "chains/graph_qa/cypher",
+    // chat models
+    "chat_models": "chat_models/configurable",
```
jacoblee93, line 69: I'm ok with `chat_models/configurable` as the entrypoint but remember to keep naming here constant

I've started relying on it for some things in the build script

langchain/package.json
```diff
@@ -694,6 +705,15 @@
     "@gomomento/sdk": "^1.51.1",
     "@gomomento/sdk-core": "^1.51.1",
     "@gomomento/sdk-web": "^1.51.1",
+    "@langchain/anthropic": "^0.2.6",
```
jacoblee93, line 708: I wouldn't upper bound these since we are very unopinionated on what you pass in - `*` is good

langchain/package.json
```diff
@@ -609,7 +613,14 @@
     "@gomomento/sdk": "^1.51.1",
     "@gomomento/sdk-core": "^1.51.1",
     "@jest/globals": "^29.5.0",
-    "@langchain/cohere": "^0.0.8",
+    "@langchain/anthropic": "^0.2.8",
```
jacoblee93, line 616: Should these be `workspace:*`?

langchain/langchain.config.js
jacoblee93: Also fine with `chat_models/universal` but check w/ Bagatur about serialization

langchain/src/chat_models/configurable.ts
```diff
@@ -0,0 +1,877 @@
...
+import {
+  type LogStreamCallbackHandlerInput,
+  type RunLogPatch,
+  type StreamEvent,
+} from "@langchain/core/tracers/log_stream";
+import { type StructuredToolInterface } from "@langchain/core/tools";
+import { CallbackManagerForLLMRun } from "@langchain/core/callbacks/manager";
+import { ChatResult } from "@langchain/core/outputs";
+
+// TODO: remove once `EventStreamCallbackHandlerInput` is exposed in core.
```
jacoblee93, line 32: Should we just expose it?

jacoblee93: You can also extract the types from the base class I think

langchain/src/chat_models/configurable.ts
```diff
@@ -0,0 +1,877 @@
...
+      }
+      case "azure_openai": {
+        const { AzureChatOpenAI } = await import("@langchain/openai");
+        return new AzureChatOpenAI({ model, ...params });
+      }
+      case "cohere": {
+        const { ChatCohere } = await import("@langchain/cohere");
+        return new ChatCohere({ model, ...params });
+      }
+      case "google_vertexai": {
```
jacoblee93, line 125: Is this for Python parity? Would double check

Otherwise would make slightly more sense to do `google-vertexai` to match the package names

langchain/langchain.config.js
jacoblee93: Oh wait - actually yeah let's do `chat_models/universal` or `configurable`. We need to be very careful that this is in its own entrypoint with nothing else since it's Node only.

langchain/src/chat_models/configurable.ts
```diff
@@ -0,0 +1,877 @@
...
+ *
+ * @param {string} modelName - The name of the model to infer the provider for.
+ * @returns {string | undefined} The inferred model provider name, or undefined if unable to infer.
+ *
+ * @example
+ * _attemptInferModelProvider("gpt-4"); // returns "openai"
+ * _attemptInferModelProvider("claude-2"); // returns "anthropic"
+ * _attemptInferModelProvider("unknown-model"); // returns undefined
+ */
+export function _attemptInferModelProvider(
```
jacoblee93, line 206: Oh ok cool

langchain/src/chat_models/configurable.ts
```diff
@@ -0,0 +1,877 @@
...
+    const model = await this._model(options);
+    return model.invoke(input, options);
+  }
+
+  async stream(
+    input: RunInput,
+    options?: CallOptions
+  ): Promise<IterableReadableStream<AIMessageChunk>> {
+    const model = await this._model(options);
+    const config = ensureConfig(options);
```
jacoblee93, line 444: Probably harmless but do you need `ensureConfig(options)` here?

jacoblee93: I would put it above in `invoke` for consistency

langchain/src/chat_models/configurable.ts
```diff
@@ -0,0 +1,877 @@
...
+
+  async *transform(
+    generator: AsyncGenerator<RunInput>,
+    options: CallOptions
+  ): AsyncGenerator<AIMessageChunk> {
+    const model = await this._model(options);
+    const config = ensureConfig(options);
+
+    for await (const chunk of generator) {
+      yield* model.transform(
```
jacoblee93, line 502: Can this just be:

```ts
yield* model.transform(generator, config);
```

langchain/langchain.config.js
bracesproul: Okay can update, this deviates from PY but ig it's not that big an issue

jacoblee93: Yeah I think we have to in this case since we should reserve the stock `chat_models` entrypoint for something more generic

langchain/package.json
bracesproul: Would be better, but this will still default to the local code, and updating to `workspace:*` will require a significant update to the latest/lowest dep CI test which I don't really want to do atm

langchain/src/chat_models/configurable.ts
bracesproul: Planning on exposing but needs to be a separate PR

langchain/src/chat_models/configurable.ts
bracesproul: yes py uses `google_vertexai`. I'll update to allow both

bracesproul: eh actually, I'm just going to make it only `google-vertexai`/`google-genai`. Exact parity for that field isn't very important.

langchain/src/chat_models/configurable.ts
bracesproul: works, replacing with that

docs/core_docs/docs/how_to/chat_models_universal_init.mdx
```diff
@@ -0,0 +1,66 @@
...
+This feature is only intended to be used in Node environments. Use in non Node environments or with bundlers is not guaranteed to work and not officially supported.
+:::
+
+:::tip Supported models
+See the [initChatModel()](https://v02.api.js.langchain.com/functions/langchain_chat_models_configurable.initChatModel.html) API reference for a full list of supported integrations.
+
+Make sure you have the integration packages installed for any model providers you want to support. E.g. you should have `@langchain/openai` installed to init an OpenAI model.
+:::
+
+:::info Requires `langchain >= 0.2.11`
```
jacoblee93, line 19: nit: Use this for standardization:

```mdx
:::caution Compatibility
`initChatModel` requires `langchain>=0.2.11`. See [this guide](/docs/how_to/installation/#installing-integration-packages) for some considerations to take when upgrading.
:::
```

docs/core_docs/docs/how_to/index.mdx
```diff
@@ -76,6 +76,7 @@ These are the core building blocks you can use when building applications.
 - [How to: stream tool calls](/docs/how_to/tool_streaming)
 - [How to: few shot prompt tool behavior](/docs/how_to/tool_calling#few-shotting-with-tools)
 - [How to: force a specific tool call](/docs/how_to/tool_choice)
+- [How to: init any model in one line](/docs/how_to/chat_models_universal_init/)
```
jacoblee93, line 79: This isn't quite accurate since it doesn't support all models - @baskaryan any other ideas on phrasing or should we just go with it

langchain/src/chat_models/universal.ts
```diff
@@ -0,0 +1,870 @@
...
+  "fireworks",
+  "mistralai",
+  "groq",
+  "bedrock",
+] as const;
+
+export type ChatModelProvider = (typeof _SUPPORTED_PROVIDERS)[number];
+
+// This is needed so when `bindTools` is called on a configurable model, the typing is correct.
+abstract class BaseChatModelWithBindTools<
```
jacoblee93, line 54: Do we need this anymore? Can't `_ConfigurableModel` just extend `BaseChatModel` directly?

langchain/src/chat_models/universal.ts
```diff
@@ -0,0 +1,870 @@
...
+ *
+ * @param {string} modelName - The name of the model to infer the provider for.
+ * @returns {string | undefined} The inferred model provider name, or undefined if unable to infer.
+ *
+ * @example
+ * _attemptInferModelProvider("gpt-4"); // returns "openai"
+ * _attemptInferModelProvider("claude-2"); // returns "anthropic"
+ * _attemptInferModelProvider("unknown-model"); // returns undefined
+ */
+export function _attemptInferModelProvider(
```
jacoblee93, line 206: nit: `_attemptToInferModelProvider` or just `_inferModelProvider`

langchain/src/chat_models/universal.ts
```diff
@@ -0,0 +1,870 @@
...
+    ) {
+      const config = Array.isArray(options) ? options[0] : options;
+      const model = await this._model(config);
+      return model.batch(inputs, config, {
+        ...batchOptions,
+        returnExceptions: batchOptions?.returnExceptions ?? false,
+      });
+    } else {
+      // If multiple configs, use the base Runnable.batch implementation
+      return super.batch(inputs, options, batchOptions);
```
jacoblee93, line 490: Why?

jacoblee93: Generally would prefer to preserve base model behavior

jacoblee93:
A few more nits

langchain/src/chat_models/universal.ts
bracesproul: Refactored to just call `super.batch` and not implement any logic ourselves. Left over from PY port (not sure why they do it)

docs/core_docs/docs/how_to/chat_models_universal_init.mdx
```diff
@@ -0,0 +1,66 @@
+# How to init any model in one line
+
+import CodeBlock from "@theme/CodeBlock";
+
```
jacoblee93, line 4: Add prereq section

jacoblee93: Also specify only for chat models

docs/core_docs/docs/how_to/chat_models_universal_init.mdx
```diff
@@ -0,0 +1,66 @@
...
+Many LLM applications let end users specify what model provider and model they want the application to be powered by.
+This requires writing some logic to initialize different ChatModels based on some user configuration.
+The `initChatModel()` helper method makes it easy to initialize a number of different model integrations without having to worry about import paths and class names.
+
+:::caution
```
jacoblee93, line 9: I would roll all these callouts into one or two

langchain/src/chat_models/universal.ts
```diff
@@ -0,0 +1,827 @@
...
+  CallOptions extends ConfigurableChatModelCallOptions = ConfigurableChatModelCallOptions
+>(
+  model?: string,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  fields?: Partial<Record<string, any>> & {
+    modelProvider?: string;
+    configurableFields?: string[] | "any";
+    configPrefix?: string;
+  }
+): Promise<_ConfigurableModel<RunInput, CallOptions>> {
```
jacoblee93, line 784: nit: should this just return something typed as `BaseChatModel`? `_ConfigurableModel` is an implementation detail

jacoblee93: Remember to update overload signatures too

jacoblee93: Also should docstring go above all overloads

bracesproul: Can't because `bindTools` on `BaseChatModel` is optional, so we'd need to re-implement that `BaseChatModelWithBindTools` class I had before.

jacoblee93: Hmm ok

PR #6174, opened 2024-07-23
build[patch]: Add way to escape side effects within an entrypoint

jacoblee93:
Also properly throw if a side effect is detected

### DIFF
--- a/libs/langchain-scripts/src/build_v2.ts
+++ b/libs/langchain-scripts/src/build_v2.ts
@@ -441,16 +441,16 @@ async function checkTreeShaking(config: LangChainConfig) {
   );
   const externals = listExternals(packageJson, config?.internals ?? []);
   const entrypoints = listEntrypoints(packageJson);
-  const consoleLog = console.log;
-  /** @type {Map<string, { log: string; hasSideEffects: boolean; }>} */
+  const consoleInfo = console.info;
+  /** @type {Map<string, { log: string; hasUnexpectedSideEffects: boolean; }>} */
   const reportMap = new Map();
 
   for (const entrypoint of entrypoints) {
     let sideEffects = "";
 
-    console.log = function (...args) {
+    console.info = function (...args) {
       const line = args.length ? args.join(" ") : "";
-      if (line.trim().startsWith("First side effect in")) {
+      if (line.includes("First side effect in")) {
         sideEffects += `${line}\n`;
       }
     };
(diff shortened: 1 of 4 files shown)


libs/langchain-scripts/src/build_v2.ts
```diff
@@ -441,16 +441,16 @@ async function checkTreeShaking(config: LangChainConfig) {
   );
   const externals = listExternals(packageJson, config?.internals ?? []);
   const entrypoints = listEntrypoints(packageJson);
-  const consoleLog = console.log;
-  /** @type {Map<string, { log: string; hasSideEffects: boolean; }>} */
+  const consoleInfo = console.info;
+  /** @type {Map<string, { log: string; hasUnexpectedSideEffects: boolean; }>} */
   const reportMap = new Map();
```
bracesproul, line 446: This can be replaced w/ an actual generic passed to Map. Any jsdocs like this in this file are leftover from the days when this was a `.js` file
```suggestion
  const reportMap = new Map<string, { log: string; hasUnexpectedSideEffects: boolean; }>();
```

libs/langchain-scripts/src/build_v2.ts
```diff
@@ -461,17 +461,26 @@ async function checkTreeShaking(config: LangChainConfig) {
       experimentalLogSideEffects: true,
     });
 
+    let hasUnexpectedSideEffects = sideEffects.length > 0;
+    if (hasUnexpectedSideEffects) {
+      const entrypointContent = fs.readFileSync(entrypoint);
```
bracesproul, line 466: why not async?

libs/langchain-scripts/src/build_v2.ts
```diff
@@ -461,17 +461,26 @@ async function checkTreeShaking(config: LangChainConfig) {
...
+      // Allow escaping side effects strictly within code directly
+      // within an entrypoints
```
bracesproul, line 468:
```suggestion
      // Allow escaping side effects strictly within code directly
      // within an entrypoint
```
I think extra "s"?

libs/langchain-scripts/src/build_v2.ts
```diff
@@ -480,7 +489,7 @@ async function checkTreeShaking(config: LangChainConfig) {
   }
 
   if (failed) {
-    process.exit(1);
+    throw new Error("Tree shaking checks failed.");
```
bracesproul, line 492: Would confirm we're not catching these errors upstream (I don't think so, but wouldn't be surprised)

bracesproul (approved):

libs/langchain-scripts/src/build_v2.ts
jacoblee93: Not worth it

libs/langchain-scripts/src/build_v2.ts
jacoblee93: Doesn't appear so

PR #6217, opened 2024-07-25
mistralai[patch]: Translate tool call ids to mistral compat format

bracesproul:
Port of https://github.com/langchain-ai/langchain/pull/24668

I couldn't use the built in `crypto` module because it is not avail in all js environments, so instead I implemented my own method.


libs/langchain-mistralai/src/chat_models.ts
```diff
@@ -586,6 +601,7 @@ export class ChatMistralAI<
     const tokenUsage: TokenUsage = {};
     const params = this.invocationParams(options);
     const mistralMessages = convertMessagesToMistralMessages(messages);
+    console.dir(mistralMessages, { depth: null });
```
bracesproul, line 604:
```suggestion
```

PR #6232, opened 2024-07-27
docs[minor]: Simplify example code by getting rid of unnecessary else/if statements

kostasx:
Removing unnecessary else/if statements to make the JS code easier to read and comprehend as per the [(No-else-return) ESLint rule](https://eslint.org/docs/latest/rules/no-else-return).


bracesproul (approved):
thanks for this!

PR #6245, opened 2024-07-29
anthropic[patch]: Prevent double-counting of token usage

tofuliang:
This PR addresses an issue where the token usage was being incorrectly calculated. The problem stemmed from redundant token usage accumulation within the package, despite the langchain-core already handling this logic. Consequently, the final token usage count was inflated.

**The fix:**

This PR removes the unnecessary token usage accumulation. By relying on the existing logic in langchain-core, we ensure accurate and consistent token usage tracking across all integrations.

### DIFF
--- a/libs/langchain-anthropic/src/chat_models.ts
+++ b/libs/langchain-anthropic/src/chat_models.ts
@@ -67,8 +67,7 @@ type AnthropicMessageStreamEvent = Anthropic.MessageStreamEvent;
 type AnthropicRequestOptions = Anthropic.RequestOptions;
 
 export interface ChatAnthropicCallOptions
-  extends BaseChatModelCallOptions,
-    Pick<AnthropicInput, "streamUsage"> {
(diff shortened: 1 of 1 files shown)


libs/langchain-anthropic/src/chat_models.ts
```diff
@@ -367,11 +344,6 @@ export interface AnthropicInput {
    */
   invocationKwargs?: Kwargs;
 
-  /**
-   * Whether or not to include token usage data in streamed chunks.
-   * @default true
-   */
-  streamUsage?: boolean;
```
bracesproul, line 374: Deleting this would be a breaking change, we need this + only streaming usage metadata if this is set to true.

bracesproul (changes requested):

bracesproul (commented):
Also, I'm unable to reproduce the duplication issue. Can you add a test which verifies this is a bug and your PR fixes it?

PR #6247, opened 2024-07-29
fix: fix mapStoredMessageToChatMessage to check on generic role when creating ChatMessage

karlomedallo:
- Fixes mapStoredMessageToChatMessage to check on generic role when creating ChatMessage
- Added test to ensure that when an instance of ChatMessage is converted to StoredMessage, it can be converted back.


bracesproul (approved):
Thanks for catching this!

PR #6250, opened 2024-07-29
aws[patch] fix: Prioritize tool calls in BedRock responses, ensures t…

tofuliang:
This function prioritizes tool calls in Bedrock responses.

It prioritizes `tool_calls` over `content` in responses from Bedrock chat models, ensuring that tool calls are handled correctly even when the response includes both text and tool calls.

This addresses an issue where Claude would sometimes include both `tool_calls` and `content` in its responses, leading to the tool calls being ignored. By prioritizing `tool_calls`, we ensure that these are always handled correctly. 

Also, it ensures that streaming works correctly even when the LLM does not support it.


libs/langchain-aws/src/common.ts
```diff
@@ -82,7 +82,26 @@ export function convertToConverseMessages(messages: BaseMessage[]): {
     .map((msg) => {
       if (msg._getType() === "ai") {
         const castMsg = msg as AIMessage;
-        if (typeof castMsg.content === "string" && castMsg.content !== "") {
+        if (castMsg.tool_calls && castMsg.tool_calls.length) {
```
bracesproul, line 85: Instead of mapping over it, calling `.reduce` would likely be a better method, so we can keep all content fields. Could you push up this refactor? If not I can push up in a little

bracesproul (changes requested):

bracesproul:
Going to close this PR due to merge conflicts. Please re-open off of `main` if you'd like to get this merged in! Thank you.

PR #6252, opened 2024-07-29
docs[patch]: Fix Code in Managing Conversation History

sarangan12:
Please refer https://github.com/langchain-ai/langchainjs/discussions/6246 for the details. The code mentioned in the site gives build error. The code change in this PR fixes the issue.


docs/core_docs/docs/tutorials/chatbot.ipynb
```diff
@@ -526,7 +526,8 @@
     "import type { BaseMessage } from \"@langchain/core/messages\";\n",
     "import { RunnablePassthrough, RunnableSequence } from \"@langchain/core/runnables\";\n",
     "\n",
-    "const filterMessages = ({ chat_history }: { chat_history: BaseMessage[]; }) => {\n",
+    "const filterMessages = (input: Record<string, unknown>) => {\n",
```
bracesproul, line 529: Prefer `Record<string, any>`, not as type safe but for docs it makes things easier.
```suggestion
    "const filterMessages = (input: Record<string, any>) => {\n",
```

docs/core_docs/docs/tutorials/chatbot.ipynb
```diff
@@ -526,7 +526,8 @@
...
+    "  const { chat_history } = input as { chat_history: BaseMessage[] };\n",
```
bracesproul, line 530:
```suggestion
    "  const { chat_history }: { chat_history: BaseMessage[] } = input;\n",
```

bracesproul (approved):
Thank you!

docs/core_docs/docs/tutorials/chatbot.ipynb
sarangan12: @bracesproul The previous suggestion (from unknown to any) will work fine. But, this line will cause the same issue. 

```
Property 'chat_history' is missing in type 'Record<string, any>' but required in type '{ chat_history: BaseMessage[]; }'.ts(2741)
tutorial2_updated.ts(31, 31): 'chat_history' is declared here.
```

So, I believe we should have it as:

```js
const { chat_history } = input as { chat_history: BaseMessage[] };
```

bracesproul: Jeez you're right, that's annoying. I pushed a commit doing this, which is a little more concise and does not throw any type errors `const filterMessages = (input: { chat_history: BaseMessage[] }) => input.chat_history.slice(-10);`

docs/core_docs/docs/tutorials/chatbot.ipynb
```diff
@@ -526,7 +526,8 @@
     "import type { BaseMessage } from \"@langchain/core/messages\";\n",
     "import { RunnablePassthrough, RunnableSequence } from \"@langchain/core/runnables\";\n",
     "\n",
-    "const filterMessages = ({ chat_history }: { chat_history: BaseMessage[]; }) => {\n",
+    "const filterMessages = (input: Record<string, any>) => {\n",
+    "  const { chat_history }: { chat_history: BaseMessage[] } = input;\n",
```
bracesproul, line 530:
```suggestion
    "const filterMessages = (input: { chat_history: BaseMessage[] }) => input.chat_history.slice(-10);\n",
```

docs/core_docs/docs/tutorials/chatbot.ipynb
sarangan12: @bracesproul  Actually, even with the new line, the code will still cause build error. But, in a different place. If you add the new line, it will error out in the following lines:

```js
    RunnablePassthrough.assign({
      chat_history: filterMessages,  // <-- Error here
    }),
    prompt,
    model,
  ]).pipe(parser);
```


The following error will be reported:

```
Type '(input: { chat_history: BaseMessage[]; }) => BaseMessage[]' is not assignable to type 'RunnableLike<Record<string, unknown>, BaseMessage[]>'.
  Type '(input: { chat_history: BaseMessage[]; }) => BaseMessage[]' is not assignable to type 'RunnableFunc<Record<string, unknown>, BaseMessage[]>'.
    Types of parameters 'input' and 'input' are incompatible.
      Property 'chat_history' is missing in type 'Record<string, unknown>' but required in type '{ chat_history: BaseMessage[]; }'.ts(2322)
tutorial2_updated.ts(30, 36): 'chat_history' is declared here.
tutorial2_updated.ts(34, 7): The expected type comes from property 'chat_history' which is declared here on type 'RunnableMapLike<Record<string, unknown>, { chat_history: BaseMessage[]; }>'
```

bracesproul: Thanks for notifying me of this. I'll just push a larger commit adding proper types throughout this entire doc. Give me a few min to do so.

Issue #6214, opened 2024-07-25
webloader IgnorePaths unexpected


DipFlip:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain.js documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain.js rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

import { GithubRepoLoader } from "@langchain/community/document_loaders/web/github";

export const run = async () => {
  const loader = new GithubRepoLoader(
    "https://github.com/DipFlip/rag_test_repo",
    { branch: "main", recursive: true, accessToken: "ghp_A1B2C3D4E5F6a7b8c9d0", unknown: "warn", ignorePaths: ["*", "!**/*.txt"] }
  );
  const docs = await loader.load();
  console.log({ docs });

### Description

When using the .gitignore syntax with a `GithubRepoLoader` to collect files from a repo, I don't seem to get the expected behavior.
By giving `ignorePaths: ["*", "!**/*.txt"]` I expect to ignore all files except `txt` files in any folders but I only get back `txt` files from the root of the repository and `txt` files in subfolders are ignored.

### System Info

yarn info v1.22.22
warning ../package.json: No license field
{
  name: 'langchain',
  'dist-tags': {
    next: '0.2.3-rc.0',
    latest: '0.2.11'
  },
  versions: [
    '0.0.4',
    '0.0.5',
    '0.0.6',
    '0.0.7',
    '0.0.8',
    '0.0.9',
    '0.0.10',
    '0.0.11',
    '0.0.12',
    '0.0.13',
    '0.0.14',
    '0.0.15',
    '0.0.16',
    '0.0.17',
    '0.0.18',
    '0.0.19',
    '0.0.20',
    '0.0.21',
    '0.0.22',
    '0.0.23',
    '0.0.24',
    '0.0.25',
    '0.0.26',
    '0.0.27',
    '0.0.28',
    '0.0.29',
    '0.0.30',
    '0.0.31',
    '0.0.32',
    '0.0.33',
    '0.0.34',
    '0.0.35',
    '0.0.36',
    '0.0.37',
    '0.0.38',
    '0.0.39',
    '0.0.40',
    '0.0.41',
    '0.0.42',
    '0.0.43',
    '0.0.44',
    '0.0.45',
    '0.0.46',
[...]

PR #6255, opened 2024-07-29
community[patch]: avoid shouldIgnore on folders for GithubRepoLoader

DipFlip:
This PR avoids the `shouldIgnore` check on directories when looping over files in a github repo that is loaded using the `GithubRepoLoader`. 

The issue with checking `shouldIgnore` on a directory is that a directory path might match an ignore pattern while its content does not. For instance, `ignorePaths: ["*", "!*/", "!**/*.txt"]` results in `ignore.ignores("subfolder") == true` but `ignore.ignores("subfolder/file.txt") == false`. Before this fix, the `GithubRepoLoader` will miss `subfolder/file.txt` and any other files in subfolders that should be included. My solution here is to avoid the `shouldIgnore` test on directories and only check `shouldIgnore` on files. This means all directories will be traversed in a recursive load but only the correct files will be loaded. 

I was unsure how to add a test case for this as an accessToken is required to do recursive collection of github repos.

Fixes https://github.com/langchain-ai/langchainjs/issues/6214

### DIFF
--- a/libs/langchain-community/src/document_loaders/web/github.ts
+++ b/libs/langchain-community/src/document_loaders/web/github.ts
@@ -510,7 +510,7 @@ export class GithubRepoLoader
     >[] = [];
 
     for (const file of files) {
-      if (this.shouldIgnore(file.path, file.type)) {
+      if (file.type !== "dir" && this.shouldIgnore(file.path, file.type)) {
         continue;
       }
       if (file.type === "file" && file.size === 0) {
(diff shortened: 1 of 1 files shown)


bracesproul (approved):
Thanks for fixing this! Would it be possible/easy to push up a test for this as well? If not no worries.

DipFlip:
@bracesproul thanks for checking it. I'd be happy to add a test. What do I use as `accessToken`? Is there some variable I can access in `libs/langchain-community/src/document_loaders/tests/github.int.test.ts` to enable recursive github load?

DipFlip:
> > @bracesproul thanks for checking it. I'd be happy to add a test. What do I use as `accessToken`? Is there some variable I can access in `libs/langchain-community/src/document_loaders/tests/github.int.test.ts` to enable recursive github load?
> 
> Only add a test if there's a way to mock the API request. If there's no way to do it without an API key then don't bother adding a test.

Ok, I don't think it can be tested without an access token so this PR is ready to go.

PR #6257, opened 2024-07-29
anthropic[patch]: Proper handling of multi tool result messages

bracesproul:
todo: update merge adjacent util func to only merge tool_result types.


libs/langchain-anthropic/src/utils/message_inputs.ts
```diff
@@ -0,0 +1,267 @@
...
+      Array.isArray(previous.content) &&
+      Array.isArray(current.content)
+    ) {
+      previous.content = previous.content.concat(current.content);
+      return acc;
+    }
+
+    return [...acc, current];
+  }, []);
+}
```
bracesproul, line 182: cc @jacoblee93 this is the new util which fixes the bug

libs/langchain-anthropic/src/utils/message_inputs.ts
```diff
@@ -0,0 +1,267 @@
...
+      }
+    } else {
+      return {
+        role,
+        content: _formatContent(message.content),
+      };
+    }
+  });
+  return {
+    messages: _mergeAdjacentMessages(formattedMessages),
```
bracesproul, line 264: and @jacoblee93 it's called here

PR #6259, opened 2024-07-29
docs[patch]: Support hidden docs cells and TS validation in notebooks

jacoblee93:
CC @bracesproul


bracesproul (approved):
nice! Thanks for doing this. Side question, roughly how much time does this add to the build process (just curious)

jacoblee93:
It's just some extra `fs` operations so probably negligible and only for docs.

Will set something up to run the validate command on new notebooks only

PR #6268, opened 2024-07-30
core[minor]: Add status field to tool

bracesproul:
Added new `status` field to tools.

TODO in separate PR:
- [update anthropic to use this field](https://github.com/langchain-ai/langchain/pull/24628/files#diff-e2dea7070f0cc79e3fc72e08f7a809d2ae17ae8924794bd917c72d37dc65cc1c)
- doc section on `status` field.
- update LangGraph `ToolNode` to use `status` field.

### DIFF
--- a/langchain-core/src/tools/index.ts
+++ b/langchain-core/src/tools/index.ts
@@ -45,14 +45,6 @@ export interface ToolParams extends BaseLangChainParams {
    * @default "content"
    */
   responseFormat?: ResponseFormat;
-  /**
-   * Whether or not to continue when an error is thrown during
-   * tool invocation. If set to true, this will set the status of `ToolMessage`
-   * to "error" and return the error message as the content.
-   * @default false
-   * @version 0.2.19
-   */
-  continueOnError?: boolean;
 }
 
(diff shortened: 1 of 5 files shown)


langchain-core/src/tools/index.ts
```diff
@@ -212,7 +212,8 @@ export abstract class StructuredTool<
     let result;
     try {
       result = await this._call(parsed, runManager, config);
-    } catch (e) {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
jacoblee93, line 215: Changes here seem fine but unrelated?

jacoblee93:
Looks good!

Only thing I'll flag is that streamed chunks will always have `success` set to true while tool messages may not

bracesproul:
> Looks good!
> 
> Only thing I'll flag is that streamed chunks will always have `success` set to true while tool messages may not

Going to push a refactor to leave undefined when concatenating unless it's passed.

langchain-core/src/tools/index.ts
bracesproul: forgot to remove when I deleted the error handling code from this PR.

PR #6282, opened 2024-07-30
core[patch]: Allow any module to emit a custom event

jacoblee93:
Using:

```ts
await runManager?.handleCustomEvent("some_test_event", {
  someval: true,
});
```

Some concerns about misuse but I think generally useful as an escape hatch.

CC @afirstenberg

### DIFF
--- a/langchain-core/src/utils/testing/index.ts
+++ b/langchain-core/src/utils/testing/index.ts
@@ -320,6 +320,8 @@ export interface FakeChatInput extends BaseChatModelParams {
 
   /** Time to sleep in milliseconds between responses */
   sleep?: number;
+
+  emitCustomEvent?: boolean;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


eyurtsev:
I'm not convinced that this makes sense given that it creates a 2nd way to do the same thing, but in a way that works in fewer cases. The current dispatch API is able to figure out what is the correct parent id context without requiring a run manager to be passed explicitly.

langchain-core/src/utils/testing/index.ts
```diff
@@ -369,9 +374,15 @@ export class FakeListChatModel extends BaseChatModel {
 
   async _generate(
     _messages: BaseMessage[],
-    options?: this["ParsedCallOptions"]
+    options?: this["ParsedCallOptions"],
+    runManager?: CallbackManagerForLLMRun
   ): Promise<ChatResult> {
     await this._sleepIfRequested();
+    if (this.emitCustomEvent) {
```
eyurtsev, line 381: In python this is doable using the regular dispatch API since a run tree will already exist. Why is this not the case here?

eyurtsev: OK i was wrong this is not dooable on the python side either

eyurtsev (approved):

PR #6291, opened 2024-07-31
docs[patch]: add clarification

austinm911:
Adds clarification and link to this external vendor


jacoblee93:
Thanks, good shout!

PR #6305, opened 2024-08-01
core: Add signal/timeout options to RunnableConfig

nfcampos:
- Handled by all built-in runnables
- Handled by all utility methods in base runnable, which should propagate to basically all runnables

Fixes # (issue)


langchain-core/src/runnables/base.ts
```diff
@@ -430,13 +444,23 @@ export abstract class Runnable<
...
     try {
-      outputs = await func.call(
+      const promise = func.call(
         this,
         inputs,
         optionsList,
         runManagers,
         batchOptions
       );
+      outputs = optionsList?.[0]?.signal
```
jacoblee93, line 454: Nice

langchain-core/src/runnables/remote.ts
```diff
@@ -214,11 +214,12 @@ function deserialize<RunOutput>(str: string): RunOutput {
   return revive(obj);
 }
 
-function removeCallbacks(
+function removeCallbacksAndSignal(
   options?: RunnableConfig
-): Omit<RunnableConfig, "callbacks"> {
+): Omit<RunnableConfig, "callbacks" | "signal"> {
```
jacoblee93, line 219: Do we have to handle `timeout` here too?

nfcampos: No, timeout can safely be passed to JSON.stringify

langchain-core/src/runnables/base.ts
```diff
@@ -378,7 +382,17 @@ export abstract class Runnable<
     delete config.runId;
     let output;
     try {
-      output = await func.call(this, input, config, runManager);
+      const promise = func.call(this, input, config, runManager);
+      output = options?.signal
+        ? await Promise.race([
+            promise,
+            new Promise<never>((_, reject) => {
```
jacoblee93, line 389: Ah, edge case but what if someone passes an already aborted signal/there's some race condition?

Feel like there should be a check?

langchain-core/src/utils/signal.ts
```diff
@@ -0,0 +1,17 @@
+export async function raceWithSignal<T>(
+  promise: Promise<T>,
+  signal?: AbortSignal
+): Promise<T> {
+  if (signal === undefined) {
+    return promise;
+  }
+  if (signal.aborted) {
```
jacoblee93, line 8: @nfcampos If we have this check here, do we need all the `signal.throwIfAborted()` calls elsewhere?

PR #6306, opened 2024-08-01
core[patch]: Allow dynamic tools to be initialized with JSON schema

jacoblee93:
If you init a tool with JSON schema, it will not currently validate/infer any types.

Types are starting to get pretty hairy in general but I've added some tests around them.

CC @bracesproul @nfcampos

### DIFF
--- a/langchain-core/src/tools/index.ts
+++ b/langchain-core/src/tools/index.ts
@@ -386,6 +386,9 @@ export class DynamicTool extends Tool {
  * description, designed to work with structured data. It extends the
  * StructuredTool class and overrides the _call method to execute the
  * provided function when the tool is called.
+ *
+ * Schema can be passed as Zod or JSON schema. The tool will not validate
+ * input if JSON schema is passed.
  */
 export class DynamicStructuredTool<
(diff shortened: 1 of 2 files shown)


debajyoti-truefoundry:
@jacoblee93 I tried using this feature with the ReAct agent but realized the JSON schema was not passed on to the OpenAI request.

https://github.com/dkundel/langchainjs/blob/9ec46a2f4d4598bbca4b97034b63cde4146d8809/langchain-core/src/tools/index.ts#L432

```
    '  "tools": [\n' +
    '    {\n' +
    '      "type": "function",\n' +
    '      "function": {\n' +
    '        "name": "postModel",\n' +
    '        "description": "Model",\n' +
    '        "parameters": {\n' +
    '          "type": "object",\n' +
    '          "properties": {},\n' +
    '          "additionalProperties": true,\n' +
    '          "$schema": "http://json-schema.org/draft-07/schema#"\n' +
    '        }\n' +
    '      }\n' +
    '    },\n' 
```

Even though there will be no validation, the JSON Schema will at least be forwarded to the model invocation. It will be helpful if you let me know what is the expected behaviour here.

PR #6308, opened 2024-08-01
community[patch]: add clientParams option to Chroma vectorstore init args

### LINKED ISSUE #6296: Allow Auth w/ Chroma
### Checked other resources

### Example Code

I do not see any way to add auth via the chroma wrapper here:
https://github.com/langchain-ai/langchainjs/blob/993ce7a822bd11e125419bff84a9951bd58e66e5/langchain/src/vectorstores/chroma.ts#L109C1-L112C8

_No response_

### Description

The following issue was closed, but doesn't seem like it was ever fixed
https://github.com/langchain-ai/langchainjs/issues/3441

### System Info

mac, osx, latest version of node, etc...

jeasonnow:
Fixes #6296


libs/langchain-community/src/vectorstores/chroma.ts
```diff
@@ -67,12 +73,16 @@ export class Chroma extends VectorStore {
     return "chroma";
   }
 
-  constructor(embeddings: EmbeddingsInterface, args: ChromaLibArgs) {
+  constructor(
+    embeddings: EmbeddingsInterface,
+    args: ChromaLibArgs & { clientParams?: Omit<ChromaClientParams, "path"> }
```
bracesproul, line 78: Could you instead make this a single exported interface `ChromaArgs`

bracesproul: Actually, just update `ChromaLibArgs` to include this `clientParams` field in both options of the type. That should be enough

bracesproul (changes requested):
One small comment, otherwise this looks good

jacoblee93:
Thanks for this!

@bracesproul before we ship a new community version with this can we make sure there's no peer dep conflicts from `npm install`?

jacoblee93:
Thank you!

PR #6323, opened 2024-08-01
langchain[minor]: use langsmith sdk for prompts functionality instead of langchain hub

madams0013:


bracesproul (approved):
lgtm

langchain/src/hub.ts
```diff
@@ -13,10 +13,26 @@ import { load } from "./load/index.js";
 export async function push(
   repoFullName: string,
   runnable: Runnable,
-  options?: HubPushOptions & ClientConfiguration
+  options?: {
+    apiKey?: string;
+    apiUrl?: string;
+    parentCommitHash?: string;
+    newRepoIsPublic?: boolean;
```
jacoblee93, line 20: Better thing to do would be to deprecate them like this:

```
/** @deprecated Use isPublic instead. */
newRepoIsPublic?: boolean;
isPublic?: boolean;
```

langchain/src/hub.ts
```diff
@@ -13,10 +13,30 @@ import { load } from "./load/index.js";
...
+    readme?: string;
+    tags?: string[];
+  }
 ) {
   const client = new Client(options);
-  return client.push(repoFullName, JSON.stringify(runnable), options);
+  const payloadOptions = {
+    object: runnable,
+    parentCommitHash: options?.parentCommitHash,
+    isPublic: options?.isPublic,
```
jacoblee93, line 34: Oh, you still need to respect the old values:

```
isPublic: options?.isPublic ?? options?.newRepoIsPublic,
```

langchain/src/hub.ts
```diff
@@ -13,10 +13,30 @@ import { load } from "./load/index.js";
...
+    description: options?.description,
```
jacoblee93, line 35: Same here, see above

jacoblee93 (approved):

jacoblee93:
Thanks!

PR #6341, opened 2024-08-02
community[patch]: Add namespace to allow indexing multiple tables within the same schema

clemenspeters:
Issue:

We're working on a tenant based SaaS platform where we use separate tables for each tenant (to avoid infinitely growing tables).

The `createHnswIndex` function translates into a query like 

```
CREATE INDEX IF NOT EXISTS vector_embedding_hnsw_idx
        ON tenant_123_index USING hnsw ((vector::vector(1536)) vector_cosine_ops)
        WITH (
            m=16,
            ef_construction=64
        );
```
where the index name is always `vector_embedding_hnsw_idx`.
Within the same schema in PostgreSQL, the index name must be unique across all tables.

This means it’s created for the first tenant, but not for anyone else since then the index with that static name already exists.

This PR allows to pass an optional `namespace` string to `createHnswIndex` which is used a prefixing the index name.

### DIFF
--- a/libs/langchain-community/src/vectorstores/pgvector.ts
+++ b/libs/langchain-community/src/vectorstores/pgvector.ts
@@ -712,10 +712,12 @@ export class PGVectorStore extends VectorStore {
         throw new Error(`Unknown distance strategy: ${this.distanceStrategy}`);
     }
 
-    const createIndexQuery = `CREATE INDEX IF NOT EXISTS ${prefix}${this.vectorColumnName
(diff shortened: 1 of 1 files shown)


bracesproul (approved):
lgtm, thanks for pushing this update

bracesproul (commented):
The format CI action is failing, could you run `yarn format` from inside `libs/langchain-community` and push up the changes? Thanks

PR #6373, opened 2024-08-04
Convert anthropic token reporting to expected langchain format for tracing

mejackreed:
Previously, llmoutput takes the Anthropic response but it is not utilized by the tracing to provide token metrics due to inconsistency with various API responses. This PR aligns llmoutput to be consistent with other providers for langchain parsability in its Tracer and callbacks. 

Happy to do this a different way if there is a more preferred approach. (e.g custom response parser for compatibility to an internal api?)

### DIFF
--- a/libs/langchain-anthropic/src/chat_models.ts
+++ b/libs/langchain-anthropic/src/chat_models.ts
@@ -512,8 +512,17 @@ export class ChatAnthropicMessages<
       additionalKwargs
     );
     // eslint-disable-next-line @typescript-eslint/no-unused-vars
-    const { role: _role, type: _type, ...rest } = additionalKwargs;
-    return { generations, llmOutput: rest };
+    const { role: _role, type: _type, usage, ...rest } = additionalKwargs;
+    return {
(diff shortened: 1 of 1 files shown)


bracesproul (commented):
Is this a regression where we used to provide it via this field, or are you passing it here so you can access it on your end? If it's the later, we suggest just fetching it from the `usage_metadata` field on `AIMessage`

mejackreed:
Thanks for the pointer @bracesproul . So for a consumer of this data (AIMessage), specifically I'm thinking [`DatadogLLMObsTracer`](https://github.com/langchain-ai/langchainjs/blob/c00dd9cac4b54ee6b028aa870c2e996c66343205/libs/langchain-community/src/experimental/callbacks/handlers/datadog.ts) would you expect that consumer to determine the Chat provider type and determine how to report those usage metrics independently?

jacoblee93:
@mejackreed it feels like it would make more sense to just update the Datadog tracer to use the new standard fields?

PR #6406, opened 2024-08-06
aws[minor]: Account for content strings and tool use

bracesproul:


libs/langchain-aws/src/common.ts
```diff
@@ -207,7 +200,8 @@ export function convertToConverseMessages(messages: BaseMessage[]): {
       } else {
         throw new Error(`Unsupported message type: ${msg._getType()}`);
       }
-    });
+    })
+    .flat();
```
bracesproul, line 204:
```suggestion
    });
```

PR #6438, opened 2024-08-07
openai[minor]: Add support for json schema response format

### LINKED ISSUE #6477: Unable to use the newly launched Structured Outputs feature of OpenAI, `response_format: { type: "json_schema", json_schema: {"strict": true, "schema": ...} }`
### Checked other resources

### Example Code

``` typescript
 const TranslationSchema = z.object({
  translatedText: z.string(),
  entities: z.array(
    z.object({
      ner: z.string(),
      label: z.string(),
      translated: z.string().optional(),
    })
  ),
});
 ...
 
    return RunnableSequence.from([
      {
        sourceLanguage: () => formattingObject.sourceLanguage,
        targetLanguage: () => formattingObject.targetLanguage,
        genre: () => formattingObject.genre,
        textToTranslate: () => formattingObject.textToTranslate,
        untranslatedNerEntities: () => formattingObject.untranslatedNerEntities,
      },
      prompt,
      (promptValue: any) => {
        const boundModel = this.model.bind({
          response_format: zodResponseFormat(TranslationSchema, "translation"),
        });
        return boundModel.invoke(promptValue);
      },
    ]);

_No response_

### Description

Unable to use the newly launched Structured Outputs feature of OpenAI, `response_format: { type: "json_schema", json_schema: {"strict": true, "schema": ...} }`

The type of response_format does not support json_schema, and the return value does not contain the JSON data defined by json_schema

### System Info

osx/bun

### LINKED ISSUE #6479: Zod to JSONSchema conversion doesn't use openai custom logic for openai
### Checked other resources

### Example Code

OpenAI's new structured output support requires compiling zod to JSONSchema, which is something that LangChain already knows how to do. However -- OpenAI's code client uses a specialized, forked version of `zod-to-json-schema` internally to do this conversion in order to improve the compatibility of the generated JSON schemas with what is supported server side in their API. Specifically, they don't use relative references, which LangChain does, because they aren't supported server side. OpenAI's client also handles nullability differently.

You can see their custom converter here: https://github.com/openai/openai-node/blob/cc13af9fa7e76e774d3132bba2427bb0176bd622/src/helpers/zod.ts#L11-L19

and some commits that added the custom logic:  https://github.com/openai/openai-node/commit/fc22483173b521b0b194264e6d975c3517c06216 and https://github.com/openai/openai-node/commit/fc22483173b521b0b194264e6d975c3517c06216

As it stands now, LangChain's conversation of tool or structured output schemas doesn't use this fancy logic, and breaks in many more cases. If you re-use a chunk of schema within a bigger, outer schema, LangChain triggers an OpenAI 400 error, and the OpenAI client doesn't.

It'd be great if LangChain could be a client of OpenAI's `zodResponseFormat` or `zodFunction` helpers instead of doing its own conversion, or something like that.

_No response_

### Description

I'm trying to use LangChain to a call a tool with the same inner bit of Zod schema repeated more than once, and it fails with OpenAI's new `strict: true` parameter when OpenAI's client doesn't.

### System Info

`@langchain/core 0.2.21`
`openai 4.55.3`
`node v22.2.0`
`pnpm v8.12.0`

### PR DESCRIPTION
**Pull request by bracesproul**
port of https://github.com/langchain-ai/langchain/pull/25123/files

Closes #6477
Closes #6479

Requires a release of `@langchain/core` and bumping min core version of oai

### CODE REVIEW

**Comment by david8z** [comment]
Hello @bracesproul is there any example on how should this be used?

**Comment by david8z** [comment]
> Hello @bracesproul is there any example on how should this be used?

Got it working:

```javascript
  model: "gpt-4o-mini",
}).bind({
  response_format: {
    type: "json_schema",
    json_schema: {
      name: "schemaName",
      schema: zodSchema,
    },
  },
});

const parser = StructuredOutputParser.fromZodSchema(zodSchema);
const typedResponse = await prompt.pipe(model).pipe(parser)
```

bracesproul:
@david8z sweet! I also have an example [here](https://v02.api.js.langchain.com/classes/_langchain_openai.ChatOpenAI.html#:~:text=JSON%20Schema%20Structured%20Output) showing how to use it with `withStructuredOutput`

nguyenthinhit996:
Hi  @bracesproul , nice to meet you. I'm currently working on LangChain.js with the Gmail Toolkit (specifically GmailCreateDraft), along with LangSmith and OpenAI. I noticed an issue with the function calling when using optional fields defined with Zod. Although fields like cc and bcc are marked as optional in the toolkit, it seems that the function is requiring all fields to be included. After investigating, I found that the zodToJsonSchema function automatically inserts some default values, which causes the resulting JSON schema to not align with the expected requirements.


image
image
image

PR #6450, opened 2024-08-07
community[patch]: support modelName param and update default fireworks model to Llama 3.1 8B

henryweng03:
- ChatFireworks manifest sometimes uses `modelName` to define the model, which incorrectly leads to defaulting to the default model
- Existing default model has been deprecated by Fireworks, causing it to error out. Update default model to llama-3.1-8b


jacoblee93:
Thank you!

jacoblee93:
Thank you!

PR #6413, opened 2024-08-06
community[patch]: PGVectorStore: Add dimensions parameter to initialize function

clemenspeters:
When using `createHnswIndex` the index is created in the database, but if the vector column doesn't have a fixed dimension, the query planner doesn't make use of the index and the query is still slow.

Example usage:
```
  const dimensions = 1536;
  const pgVectorStore = await PGVectorStore.initialize(embeddings, pgConfig, dimensions)

  await pgVectorStore.createHnswIndex({
    dimensions,
    efConstruction: 64,
    m: 16,
  });
  ```
  
  @PetersClemens

PR #6452, opened 2024-08-07
Extends 6413

jacoblee93:
Extends #6413 

CC @clemenspeters


libs/langchain-community/src/vectorstores/pgvector.ts
```diff
@@ -163,17 +163,19 @@ export class PGVectorStore extends VectorStore {
...
    * @param embeddings - Embeddings instance.
-   * @param fields - `PGVectorStoreArgs` instance.
+   * @param fields - `PGVectorStoreArgs` instance
+   * @param fields.dimensions Number of dimensions in your vector data type. For example, use 1536 for OpenAI's `text-embedding-3-small`. If not set, indexes like HNSW might not be used during query time.
    * @returns A new instance of `PGVectorStore`.
    */
   static async initialize(
     embeddings: EmbeddingsInterface,
-    config: PGVectorStoreArgs
+    config: PGVectorStoreArgs & { dimensions?: number }
```
clemenspeters, line 172: Would it make sense to add the optional dimensions to the PGVectorStoreArgs type directly? 🤔

jacoblee93: Yeah I can now see how it's not convenient - I can just make the constructor `Omit` dimensions

PR #6459, opened 2024-08-08
google[minor]: Move Vertex embeddings to integration package

jacoblee93:
Will need to break this up into different PRs to make it pass CI

CC @afirstenberg


libs/langchain-community/src/utils/googlevertexai-connection.ts
```diff
@@ -212,7 +212,9 @@ export class GoogleVertexAILLMConnection<
     }
 
     const projectId = await this.client.getProjectId();
-
+    console.log(
```
afirstenberg, line 215: Minor nit - remove console.log. 😁

PR #6468, opened 2024-08-08
community[patch]: Update JSDoc on bedrock chat

bracesproul:


libs/langchain-community/src/chat_models/bedrock/index.ts
```diff
@@ -14,15 +14,325 @@ export interface BedrockChatFields
...
+ * export BEDROCK_AWS_SECRET_ACCESS_KEY="your-aws-secret-access-key"
+ * export BEDROCK_AWS_ACCESS_KEY_ID="your-aws-access-key-id"
+ * ```
+ *
+ * ## [Constructor args](/classes/langchain_community_chat_models_bedrock.BedrockChat.html#constructor)
+ *
+ * ## [Runtime args](/interfaces/langchain_community_chat_models_bedrock_web.BedrockChatCallOptions.html)
+ *
+ * Runtime args can be passed as the second argument to any of the base runnable methods `.invoke`. `.stream`, `.batch`, etc.
+ * They can also be passed via the `.bind`, or the second arg in the `.bindTools` method, like shown in the example below:
```
bracesproul, line 34:
```suggestion
* They can also be passed via `.bind`, or the second arg in `.bindTools`, like shown in the examples below:
```

libs/langchain-community/src/chat_models/bedrock/web.ts
```diff
@@ -154,100 +154,310 @@ export interface BedrockChatFields
...
+ *
+ * ## Key args
+ *
+ * ### [Init args](/classes/langchain_community_chat_models_bedrock.BedrockChat.html#constructor)
+ *
+ * ### [Runtime args](/interfaces/langchain_community_chat_models_bedrock_web.BedrockChatCallOptions.html)
+ *
+ * > See full list of supported init args and their descriptions in the [`constructor`](/classes/langchain_community_chat_models_bedrock.BedrockChat.html#constructor) section.
+ *
+ * ## Examples
```
bracesproul, line 177:
```suggestion
 * ```
 *
 * ## [Constructor args](/classes/langchain_community_chat_models_bedrock.BedrockChat.html#constructor)
 *
 * ## [Runtime args](/interfaces/langchain_community_chat_models_bedrock_web.BedrockChatCallOptions.html)
 *
 * Runtime args can be passed as the second argument to any of the base runnable methods `.invoke`. `.stream`, `.batch`, etc.
* They can also be passed via `.bind`, or the second arg in `.bindTools`, like shown in the examples below:
 *
 * ```typescript
 * // When calling `.bind`, call options should be passed via the first argument
 * const llmWithArgsBound = llm.bind({
 *   stop: ["\n"],
 *   tools: [...],
 * });
 *
 * // When calling `.bindTools`, call options should be passed via the second argument
 * const llmWithTools = llm.bindTools(
 *   [...],
 *   {
 *     stop: ["stop on this token!"],
 *   }
 * );
 * ```
 *
 * ## Examples
```

PR #6480, opened 2024-08-09
anthropic[patch]: Allow 8K maxtokens header for Anthropic while using tools

renebrandel:
Fixes # (issue)

Right now, you can't use tools and Anthropic's 8K maxTokens feature. I root-caused the issue to langchain always setting the "tools-2024-04-04" header, when tools are used. As far as I know, this tools beta header is deprecated and not in use anymore. This change allows customers to set the maxTokens feature and use tools with Anthropic.

This is aligned with the Python implementation of Langchain, which doesn't conditionally overwrite the headers based on the availability of tools. https://github.com/langchain-ai/langchain/blob/master/libs/partners/anthropic/langchain_anthropic/chat_models.py


jacoblee93:
CC @bracesproul but this looks good to me

renebrandel:
@jacoblee93 - I fixed the build error. Silly oversight on my end earlier. Sorry for the churn.

jacoblee93:
No worries thank you!

jacoblee93:
Can also land ourselves if CI is burdensome

renebrandel:
I'm just running through all tests locally now and all are green. So I think this build is good to go in CI.

renebrandel:
@jacoblee93 looks like the docs and api reference checks have passed. LMK if there's any more edits needed on my end to kick off the CI.

bracesproul:
@renebrandel just kicked off CI again, hold off on updating the branch with main for now since that'll require us to run CI again.

renebrandel:
Sounds good thanks for the update and context.

renebrandel:
@bracesproul looks like it's ready to merge? Anything else I can assist with? Changelog authoring etc?

jacoblee93:
Thank you!

PR #6486, opened 2024-08-09
docs[minor]: Add missing {context} variable in example

kostasx:
The example included under the deprecated `RetrievalQAChain` method is missing the required `{context}` variable.

The `createStuffDocumentsChain` _"expects a dictionary as input with a list of `Document`s being passed under the "context" key."_ as per the [specs](https://github.com/in-tech-gration/langchainjs/blob/d76cd5aabff3c37348fc34a0461dea0b61e1be28/langchain/src/chains/combine_documents/stuff.ts#L33).

If one tries to replicate the example without including the `{context}` variable in the prompt template, the following error is produced:

> `prompt must include a "context" variable`

Which is based on `DOCUMENTS_KEY = "context";`

I believe the example must include the required variable to avoid any confusion and streamline the process of migrating to the alternative syntax.


jacoblee93:
Thanks!

PR #6488, opened 2024-08-10
community[patch]: Support for specifying api url for firecrawl document loader

### LINKED ISSUE #6489: Add support for specifying api url for FirecrawlLoader
### Privileged issue

### Issue Content

Currently the FirecrawlLoader does not support self hosted instances of Firecrawl. This change would allow users to specify the api url of the self hosted Firecrawl instance. This feature is already supported by the original `@mendable/firecrawl-js` package.

wahpiangle:
Fixes #6489 

JS Version of langchain-ai/langchain/pull/24747

This PR would provide support for users to use self-hosted instances of Firecrawl.

The FirecrawlApp class already has support to pass this argument. See https://github.com/mendableai/firecrawl/blob/main/apps/js-sdk/firecrawl/src/index.ts#L9

### DIFF
--- a/libs/langchain-community/src/document_loaders/web/firecrawl.ts
+++ b/libs/langchain-community/src/document_loaders/web/firecrawl.ts
@@ -19,7 +19,7 @@ interface FirecrawlLoaderParameters {
   apiKey?: string;
 
   /**
-   * API URL for Firecrawl. If not provided, the default value is "https://api.firecrawl.dev".
+   * API URL for Firecrawl.
    */
   apiUrl?: string;
   /**
--- a/libs/langchain-community/src/document_loaders/web/firecrawl.ts
+++ b/libs/langchain-community/src/document_loaders/web/firecrawl.ts
@@ -50,7 +50,7 @@ interface FirecrawlDocument {
 export class FireCrawlLoader extends BaseDocumentLoader {
   private apiKey: string;
 
-  private apiUrl: string;
+  private apiUrl?: string
 
   private url: string;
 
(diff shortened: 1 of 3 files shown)


libs/langchain-community/src/document_loaders/web/firecrawl.ts
```diff
@@ -56,6 +62,7 @@ export class FireCrawlLoader extends BaseDocumentLoader {
     super();
     const {
       apiKey = getEnvironmentVariable("FIRECRAWL_API_KEY"),
+      apiUrl = "https://api.firecrawl.dev",
```
jacoblee93, line 65: Let's just leave this unset by default? Then we can only need to pass it to the underlying SDK if it is overridden?

jacoblee93 (changes requested):
Let's remove the default value so that if the official SDK changes no update will be needed here

libs/langchain-community/src/document_loaders/web/firecrawl.ts
```diff
@@ -78,7 +86,7 @@ export class FireCrawlLoader extends BaseDocumentLoader {
    * @throws An error if the data could not be loaded.
    */
   public async load(): Promise<DocumentInterface[]> {
-    const app = new FirecrawlApp({ apiKey: this.apiKey });
+    const app = new FirecrawlApp({ apiKey: this.apiKey, apiUrl: this.apiUrl });
```
jacoblee93, line 89: Would prefer:

```ts
const params: typeof ConstructorParams<FirecrawlApp> = { apiKey: this.apiKey };
if (this.apiUrl !== undefined) {
  params.apiUrl = this.apiUrl;
}
const app = new FirecrawlApp(params);
```

jacoblee93:
Thank you!

PR #6494, opened 2024-08-12
docs[minor]: Update base model in `ChatFriendli` ChatModel

seuha516:
## Description
- Update `ChatFriendli`'s base model from `llama-2-13b-chat` to `meta-llama-3-8b-instruct`.


jacoblee93:
Would generally prefer not to have a default in the future and have people specify the model but this is fine for now

PR #6519, opened 2024-08-13
google-genai[minor]: Add support for json output

juji:
Support JSON output with google-genai, by adding a `json: true` to instance initialization.

Example:
```ts
  model: "gemini-1.5-flash",
  json: true
})
```

Usage Example:
```ts
const messages = [
  new SystemMessage(`
    Translate the following English text into Italian. 

    Reply with JSON, using the following JSON schema:
    {"translation": "string"}
  `),
  new HumanMessage("hi, i am superman"),
];

  model: "gemini-1.5-flash",
  json: true
})

const parser = new JsonOutputParser()
const result = await model.pipe(parser).invoke(messages);
console.log(result)

// {"translation": "Ciao, sono Superman"}
```


libs/langchain-google-genai/src/chat_models.ts
```diff
@@ -244,6 +247,8 @@ export class ChatGoogleGenerativeAI
 
   streamUsage = true;
 
+  json = false
```
bracesproul, line 250: If this is only accessed inside the class constructor we don't need to make it a class property.
```suggestion
```

libs/langchain-google-genai/src/chat_models.ts
```diff
@@ -167,6 +167,9 @@ export interface GoogleGenerativeAIChatInput
 
   /** Whether to stream the results or not */
   streaming?: boolean;
+
+  /** Whether the response will be json or not */
+  json?: boolean;
```
bracesproul, line 172:
```suggestion
  /**
   * Whether or not to force the model to respond with JSON.
   * @default false
   */
  json?: boolean;
```

libs/langchain-google-genai/src/chat_models.ts
```diff
@@ -310,6 +315,8 @@ export class ChatGoogleGenerativeAI
 
     this.streaming = fields?.streaming ?? this.streaming;
 
+    this.json = fields?.json ?? false
```
bracesproul, line 318:
```suggestion
```
drop since it doesn't need to be a class property

libs/langchain-google-genai/src/chat_models.ts
```diff
@@ -321,6 +328,7 @@ export class ChatGoogleGenerativeAI
           temperature: this.temperature,
           topP: this.topP,
           topK: this.topK,
+          ...this.json ? { responseMimeType: "application/json" } : {}
```
bracesproul, line 331:
```suggestion
          ...fields?.json ? { responseMimeType: "application/json" } : {}
```
nit

bracesproul (commented):
Thank you for pushing these updates! It would be great if you could also push up a test verifying this mode works as intended. Thank you!

jacoblee93:
Thank you!

PR #6539, opened 2024-08-15
pinecone[patch]: Fix document ID not getting set when returned from PineconeStore

nick-w-nick:
When querying for documents using the `PineconeStore` via the various search methods, the documents getting returned from Pinecone currently have their `id` property set to `undefined`.

This is because `id` is omitted when the result is converted to a document prior to being returned, as shown below for each query method respectively:

`similaritySearchVectorWithScore`
https://github.com/langchain-ai/langchainjs/blob/abdcd1fa3b74f066e4a7cbe3688a99f719f8f75b/libs/langchain-pinecone/src/vectorstores.ts#L424-L426

`maxMarginalRelevanceSearch`
https://github.com/langchain-ai/langchainjs/blob/abdcd1fa3b74f066e4a7cbe3688a99f719f8f75b/libs/langchain-pinecone/src/vectorstores.ts#L476-L478

This PR adds a new private method named `_formatMatches` that accepts an array of matches (`ScoredPineconeRecord`) from the `QueryResponse` interface. This method performs the same logic that was duplicated across the query methods, just with some added deconstruction safety rails as well as the addition of the `id` property.

Result from `similaritySearchWithScore` **before** this change (same for `similaritySearchVectorWithScore`):
```
[
  [
    Document {
      pageContent: 'document text',
      metadata: [Object],
      id: undefined
    },
    0.81118
  ]
]
```

Result from `similaritySearchWithScore` **after** this change (same for `similaritySearchVectorWithScore`):
```
[
  [
    Document {
      pageContent: 'document text',
      metadata: [Object],
      id: '91522887-3be3-4937-9ea7-eb235b991a4c'
    },
    0.81118
  ]
]
```

Result from `similaritySearch` **before** this change (same as `maxMarginalRelevanceSearch`):
```
[
  [
    Document {
      pageContent: 'document text',
      metadata: [Object],
      id: undefined
    }
  ]
]
```

Result from `similaritySearch` **after** this change (same as `maxMarginalRelevanceSearch`):
```
[
  [
    Document {
      pageContent: 'document text',
      metadata: [Object],
      id: '91522887-3be3-4937-9ea7-eb235b991a4c'
    }
  ]
]
```

### DIFF
--- a/libs/langchain-pinecone/src/tests/vectorstores.int.test.ts
+++ b/libs/langchain-pinecone/src/tests/vectorstores.int.test.ts
@@ -1,6 +1,7 @@
 /* eslint-disable no-process-env */
 /* eslint-disable @typescript-eslint/no-non-null-assertion */
 /* eslint-disable no-promise-executor-return */
+/* eslint-disable @typescript-eslint/no-explicit-any */
 import { describe, expect, test } from "@jest/globals";
 import { faker } from "@faker-js/faker";
 import { Pinecone } from "@pinecone-database/pinecone";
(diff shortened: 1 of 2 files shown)


bracesproul (changes requested):
thanks for pushing this up, well needed update! Could you also push up a commit adding a test for this? Thank you!

jacoblee93:
Thank you!

PR #6540, opened 2024-08-15
aws[patch]: Use abort signal when available with SDK requests

plisy:
When using the Converse APIs in `ChatBedrockConverse` the `AbortSignal` wasn't being passed in. 

This prevented the cancellation from working.


jacoblee93:
Thank you very much!

jacoblee93 (approved):

PR #6549, opened 2024-08-16
core[patch]: Pass input to invocation for JSON schema tools

dkundel:
#6306 introduced the ability to pass JSON Schema instead of Zod instances into the `schema` of `tool()` or `DynamicStructuredTool`. This would skip the input validation on the Langchain side but it introduced two bugs that this PR addresses:
1. Even though the tool get appropriately triggered it would always pass `{}` as input since the fallback Zod schema that was used during initialization would map to `z.object({})` which when used with `schema.parseAsync(input)` results in all properties being dropped. To get around this we need to use `passthrough()` to tell Zod not to drop any additional properties. 
2. The `tool()` function is checking if a schema is not an object Zod schema and then returns a `DynamicTool` instance instead of a `DynamicStructuredTool` instance. This results in the resulting Tool execution always passing `undefined` as input. Instead `tool()` needs to check that if the input `schema` is defined that it's not an instance of a ZodType before checking any other conditions.

This is my first contribution so I'm not sure if I missed anything but the tests seemed to pass.

My Twitter handle is `dkundel`.


jacoblee93 (approved):
Thanks for catching! Just ran `lint` and `format` but will merge and ship shortly

PR #6571, opened 2024-08-19
docs[minor]: Add doc section on structured tool params

bracesproul:


docs/core_docs/docs/how_to/custom_tools.ipynb
```diff
@@ -32,7 +32,51 @@
...
+    ":::\n",
+    "```\n",
+    "\n",
+    "The simplest way to create a tool is through the [`StructuredToolParams`](https://api.js.langchain.com/interfaces/_langchain_core.tools.StructuredToolParams.html) schema. Every chat model which supports tool calling in LangChain accepts binding tools to the model through this schema. This schema has only three fields\n",
+    "\n",
+    "- `name` - The name of the tool.\n",
+    "- `schema` - The schema of the tool, defined with a Zod object.\n",
+    "- `description` (optional) - A description of the tool.\n",
+    "\n",
+    "This schema does not include a function to pair with the tool, and for this reason it should only be used in situations where you want to generate structured output using an LLM."
```
bracesproul, line 59:
```suggestion
    "This schema does not include a function to pair with the tool, and for this reason it should only be used in situations where the generated output does not need to be passed as the input argument to a function."
```

PR #6575, opened 2024-08-20
baidu-qianfan[patch]: Fix the issue of missing request parameters

Versirity:
This PR is to fix the issue where only the `messages` parameter was sent: The generated request: 
`{ stream: false, user_id: undefined, temperature: undefined, top_p: undefined, penalty_score: undefined, system: 'Translate the following from English into Chinese!', messages: [ { role: 'user', content: 'hello?' } ] } `
is correct, but when sending the request, only the `messages` parameter was included, and the other parameters were not sent.

```
{
  stream: false,
  user_id: undefined,
  temperature: undefined,
  top_p: undefined,
  penalty_score: undefined,
  system: 'Translate the following into italian:',
  messages: [ { role: 'user', content: 'hi!' } ]
}
####error return（system parameters were not sent)
{
  words: 'Hello there! I am doing well and feeling quite alright. I am a virtual artificial intelligence language model and I am always here to assist you. What about you? How are you today?'   
}
####correct return
{ words: 'Ciao!' }
```


bracesproul (approved):
lgtm, thank you!

PR #6582, opened 2024-08-21
google-common[minor]: export GoogleAISafetyError type

marsal:
Minor patch to export GoogleAISafetyError type
Useful in case you don't want to modify the GoogleAISafeyHandler workflow, but need to evaluate in your code if this type of exception is being thrown


afirstenberg:
Ack! I didn't export this? My bad! Was definitely meant to be available.
@jacoblee93 and @bracesproul - LGTM

bracesproul (approved):
lgtm, thank you!

PR #6612, opened 2024-08-23
core[patch]: Prevent ConsoleCallbackHandler from throwing TypeError when logging

benjamincburns:
Prior to this change, when `onToolEnd` is called, it assumed the value of `run.outputs.output` is always of type `string`. In cases when it wasn't a string, this would cause a `TypeError` to be thrown.

Note that it appeared that the same issue could occur in `onToolStart`, so I corrected that case as well.


jacoblee93:
Ah thank you!

PR #6646, opened 2024-08-27
feat(core): always provide message ID inferred from run ID if not present while streaming

dqbd:
Fixes # (issue)

### DIFF
--- a/langchain-core/src/language_models/chat_models.ts
+++ b/langchain-core/src/language_models/chat_models.ts
@@ -264,6 +264,10 @@ export abstract class BaseChatModel<
           callOptions,
           runManagers?.[0]
         )) {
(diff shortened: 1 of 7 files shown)


langchain-core/src/language_models/chat_models.ts
```diff
@@ -264,6 +264,10 @@ export abstract class BaseChatModel<
           callOptions,
           runManagers?.[0]
         )) {
+          if (chunk.message.id == null) {
+            const runId = runManagers?.at(0)?.runId;
```
jacoblee93, line 268: Oh cool didn't know this existed

jacoblee93 (approved):
Thank you!

PR #6657, opened 2024-08-28
feat(core): support method field in StructuredPrompt

henryweng03:
Fixes # (issue)


benjamincburns:
Thanks for the contribution!

PR #6659, opened 2024-08-29
Bump pyodide version to latest, allow installing latest as peer dependency

benjamincburns:
Prior to this change, the `langchain` package's peer dependency on `pyodide` was restricted to `^0.24.1`. Because this is an unstable release, this causes peer dependency conflict warnings in user code that depends on the latest version (0.26.2).

This change updates the range in the peer dependency to allow peer versions from the previously-defined version until the current latest minor version (allowing for patch updates), and sets the dev dependency version to the caretted latest version, to ensure that tests still pass with this version.

I also audited the [Pyodide Changelog](https://pyodide.org/en/stable/project/changelog.html) for breaking changes that might affect the `PythonInterpreterTool` (the only direct use of `pyodide` that I can find). While there weren't any logic changes that would break the direct use of `pyodide` in LangChain code, there was [one breaking change](https://github.com/pyodide/pyodide/pull/4342) to the input params object of the `loadPyodide` function in [0.25.0](https://pyodide.org/en/stable/project/changelog.html#version-0-25-0).

With respect to whether these changes cause any breakages in the LangChain API, I think that's a bit open to interpretation, and it would depend on how you classify breaking changes to experimental tools, and whether you regard interface structures that are "forwarded" from a peer dependency to be part of the LangChain public API.

I mention this because [`PythonInterpreterToolParams` is defined](https://github.com/langchain-ai/langchainjs/blob/5d63f397c02f232d2a3729ed34461e209cfb0350/langchain/src/experimental/tools/pyinterpreter.ts#L4-L7) as a conjunction of the `loadPyodide` params object, meaning that breaking changes to that type do effectively cause breaking changes to `PythonInterpreterToolParams`. That said, even setting aside that `PythonInterperterTool` is an experimental tool and this likely doesn't matter for that reason, I'd argue that by setting the version range to continue to allow versions that matched the old range, these changes aren't actually breaking, as the consumer of LangChain can decide whether or not they want a version of pyodide from before or after these breakages.

For good measure I also checked comments for links to pyodide docs (e.g. in `examples/src/tools/pyinterpreter.ts`) and those are still valid and live.


benjamincburns:
Argh, I seem to have accidentally included a change from my other PR - will drop that commit. Sorry about that!

PR #6536, opened 2024-08-15
community[patch]: fix: ElasticVectorSearch: exclude metadata filters …

choeller:
…not working due to syntax error in filter creation

Fixes #6535

PR #6660, opened 2024-08-29
fix(community): ElasticVectorSearch: exclude metadata filters not working due changing to term instead of terms

LaurensTsestigh:
The fix merged in #6536 broke our code due to the following error:

`parsing_exception: [term] query does not support array of values`

As we also need to allow an array, I made some changes to allow both term and terms.

### DIFF
--- a/libs/langchain-community/src/vectorstores/elasticsearch.ts
+++ b/libs/langchain-community/src/vectorstores/elasticsearch.ts
@@ -343,10 +343,9 @@ export class ElasticVectorSearch extends VectorStore {
           },
         });
       } else if (condition.operator === "exclude") {
+        const toExclude = { [metadataField]: condition.value }
         must_not.push({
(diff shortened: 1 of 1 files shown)


jacoblee93:
Hey sorry about this and the delay here - will test and have a look.

It might be nice to add an integration test here to prevent breakages in the future.

jacoblee93:
Added in https://github.com/langchain-ai/langchainjs/pull/6688 and will merge there. Thank you very much!

jacoblee93:
Merged in #6688, thank you again!

LaurensTsestigh:
> Hey sorry about this and the delay here - will test and have a look.
> 
> It might be nice to add an integration test here to prevent breakages in the future. I can look into it.

No problem at all, thanks for reviewing and merging!
Have a nice day

PR #6664, opened 2024-08-30
fix(openai): Check `configuration.apiKey` in `ChatOpenAI` constructor

ezra-chu:
OpenAI SDK's `ClientOptions` includes `apiKey`. It looks like `@langchain/openai`'s `ChatOpenAI` constructor does not honor that field. 

image

Referenced in the docs [here](https://v02.api.js.langchain.com/interfaces/_langchain_openai.ClientOptions.html).


jacoblee93 (approved):
Thank you! Apologies for the delay.

PR #6671, opened 2024-09-01
Update score_threshold.ts

emehmet:
fix: Updated ScoreThresholdRetriever to filter results with scores below the minimum similarity score

Fixes # (issue)


jacoblee93:
This is breaking - we could add a way to specify sort order though!

PR #6672, opened 2024-09-01
Update score_threshold.ts

emehmet:
Fix: Initialize minSimilarityScore before using it in the constructor to avoid reference error.

Fixes # (issue)

### DIFF
--- a/langchain/src/retrievers/score_threshold.ts
+++ b/langchain/src/retrievers/score_threshold.ts
@@ -17,7 +17,7 @@ export type ScoreThresholdRetrieverInput<V extends VectorStore> = Omit<
 export class ScoreThresholdRetriever<
   V extends VectorStore
 > extends VectorStoreRetriever<V> {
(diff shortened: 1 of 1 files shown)


jacoblee93 (commented):
It's typed as required in `ScoreThresholdRetrieverInput`?

langchain/src/retrievers/score_threshold.ts
```diff
@@ -17,7 +17,7 @@ export type ScoreThresholdRetrieverInput<V extends VectorStore> = Omit<
 export class ScoreThresholdRetriever<
   V extends VectorStore
 > extends VectorStoreRetriever<V> {
-  minSimilarityScore: number;
+  minSimilarityScore: number = 0; // Initialize minSimilarityScore
```
bracesproul, line 20:
```suggestion
  minSimilarityScore = 0;
```
nit, no need for comment or type anymore

bracesproul (approved):
thanks for pushing this!

PR #6674, opened 2024-09-02
feat: add arcjet redact integration

e-moran:
This PR introduces the [Arcjet](https://arcjet.com/) Redact langchainjs integration, it provides local sensitive information redaction and un-redaction capabilities using the [@arcjet/redact](https://www.npmjs.com/package/@arcjet/redact) package. Data that is redacted using Arcjet never leaves your machine, ensuring best in class performance and privacy.


e-moran:
Here's what the docs look like:
image

docs/core_docs/src/theme/FeatureTables.js
```diff
@@ -261,6 +261,13 @@ const FEATURE_TABLES = {
         apiLink:
           "https://api.python.langchain.com/en/latest/llms/langchain_ai21.llms.AI21LLM.html#langchain_ai21.llms.AI21LLM",
       },
+      {
+        name: "Arcjet Redact",
+        link: "arcjet",
+        package: "langchain-arcjet",
+        // apiLink:
```
e-moran, line 268: Was unsure what to do with this apiLink field as we don't have a Python version of this package.

docs/core_docs/docs/integrations/llms/arcjet.mdx
```diff
@@ -0,0 +1,34 @@
+import CodeBlock from "@theme/CodeBlock";
```
jacoblee93, line 1: We prefer standardization along the lines of this template:

https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-scripts/src/cli/docs/templates/llms.ipynb

Happy to update it though after other comments are addressed

examples/src/llms/arcjet.ts
```diff
@@ -0,0 +1,42 @@
+import {
+  ArcjetRedact
+} from "@langchain/community/llms/arcjet";
+import { OpenAI } from "@langchain/openai";
+
+// Create an instance of another LLM for Arcjet to wrap
+const openai = new OpenAI({
```
jacoblee93, line 7: We probably will want this to work with chat models - most state of the art models are chat models rather than LLMs.

libs/langchain-community/src/llms/arcjet.ts
```diff
@@ -0,0 +1,96 @@
...
+  contextWindowSize?: number;
+
+  detect?: Detect;
+
+  replace?: (entity: ValidEntities<Detect>[number]) => string | undefined;
+
+  constructor(options: ArcjetRedactOptions<Detect>) {
+    super(options);
+
+    if (!options.llm) {
```
jacoblee93, line 42: Can we rely on TypeScript for some of these?

jacoblee93 (commented):
Thanks for this!

Looks solid overall - only thing is that most people use chat models rather than text-in/text-out LLMs these days (gpt-4, 3.5-sonnet, mistral-large are all chat models).

I can merge this more or less as is but it might be better to make this a chat model wrapper instead?

examples/src/llms/arcjet.ts
e-moran: We are planning on adding a second integration for chat models, I'm planning on adding that in a second PR though since it is a separate piece of work.

libs/langchain-community/src/llms/arcjet.ts
e-moran: Yeah, I've removed all of the ones that aren't handled by TS

docs/core_docs/docs/integrations/llms/arcjet.mdx
e-moran: I've migrated this to jupyter

libs/langchain-community/langchain.config.js
```diff
@@ -91,6 +91,7 @@ export const config = {
     // llms
     "llms/ai21": "llms/ai21",
     "llms/aleph_alpha": "llms/aleph_alpha",
+    "llms/arcjet": "llms/arcjet",
```
jacoblee93, line 94: Will need to add these under `requiresOptionalDependency` below but I can do that

libs/langchain-community/langchain.config.js
```diff
@@ -163,6 +164,7 @@ export const config = {
     "vectorstores/zep_cloud": "vectorstores/zep_cloud",
     // chat_models
     "chat_models/alibaba_tongyi": "chat_models/alibaba_tongyi",
+    "chat_models/arcjet": "chat_models/arcjet",
```
jacoblee93, line 167: Same see above

libs/langchain-community/package.json
```diff
@@ -213,6 +213,7 @@
     "youtubei.js": "^9.1.0"
   },
   "peerDependencies": {
+    "@arcjet/redact": "^v1.0.0-alpha.23",
```
jacoblee93, line 216: Should this be more relaxed?

jacoblee93: Also needs to be under `devDependencies` too but I can add that

jacoblee93 (commented):
Looks good, small question on deps but I can fix it up and merge this weekend.

Thanks for your patience!

libs/langchain-community/package.json
e-moran: No that's actually the first version of the package. While we're in alpha we're just using common versioning for all of our npm packages, so `@arcjet/redact` starts with `v1.0.0-alpha.32`

davidmytton: It starts with `v1.0.0-alpha.23` not v1.0.0-alpha.32`.

davidmytton:
Anything else you need from us to get this over the line @jacoblee93 ? We're aiming to announce the redaction features on Thursday 💪

jacoblee93:
Reviewing now!

jacoblee93:
Hey @davidmytton I couldn't push to this branch but created #6725 with build and CJS fixes. Also removed the `const` in the generic since it's not supported in older TS versions - would prefer not to break them.

Let me know what you think! Will merge later.

davidmytton:
Looks good - thanks for the edits!

jacoblee93:
Merged there, closing this!

PR #6676, opened 2024-09-02
Rerank Documents Based on Recency Bias on Document Metadata Date

Mraj23:
Fixes # (issue): https://github.com/langchain-ai/langchain/discussions/18729 

This PR implements a retriever that performs reranking based on receny of a documents published date.

This retreiver differs from the `TimeWeightedVectorStoreRetriever` that performs retrieval based on the **retrieved date** of the document, not its **published date**

### DIFF
--- a/langchain/src/retrievers/recency_ranked.ts
+++ b/langchain/src/retrievers/recency_ranked.ts
@@ -1,0 +1,66 @@
+import { BaseRetriever } from "@langchain/core/retrievers";
+import { VectorStoreInterface } from "@langchain/core/vectorstores";
+import { Document } from "@langchain/core/documents";
+
+export interface RecencyRankedRetrieverConfig {
+  vectorStore: VectorStoreInterface;
+  k: number;
+  recencyWeight?: number;
+}
+
+export class RecencyRankedRetriever extends BaseRetriever {
+  static lc_name() {
+    return "RecencyRankedRetriever";
+  }
+
+  lc_namespace = ["langchain", "retrievers", "recency_ranked"];
+  
+  private vectorStore: VectorStoreInterface;
+
+  private k: number;
+
+  private recencyWeight: number;
+
+  constructor(config: RecencyRankedRetrieverConfig) {
+    super();
+    this.vectorStore = config.vectorStore;
+    this.k = config.k;
+    this.recencyWeight = config.recencyWeight ?? 0.3;
+  }
+
+  async getRelevantDocuments(query: string): Promise<Document[]> {
+    const relevantDocs = await this.vectorStore.similaritySearchWithScore(query, 15);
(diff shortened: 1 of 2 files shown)


langchain/src/retrievers/recency_ranked.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  constructor(config: RecencyRankedRetrieverConfig) {
+    super();
+    this.vectorStore = config.vectorStore;
+    this.k = config.k;
+    this.top_k = config.top_k ?? config.k;
+    this.recencyWeight = config.recencyWeight;
+  }
+
+  async getRelevantDocuments(query: string): Promise<Document[]> {
```
jacoblee93, line 35: You should override `_getRelevantDocuments` rather than the non-underscored method:

https://js.langchain.com/v0.2/docs/how_to/custom_retriever

langchain/src/retrievers/recency_ranked.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  private top_k: number;
+
+  private recencyWeight: number;
+
+  constructor(config: RecencyRankedRetrieverConfig) {
+    super();
+    this.vectorStore = config.vectorStore;
+    this.k = config.k;
+    this.top_k = config.top_k ?? config.k;
```
jacoblee93, line 31: Convention is to use camel case

jacoblee93 (commented):
Thanks for this!

Would it make more sense to add a different mode to the existing `TimeWeightedVectorStoreRetriever`?

langchain/src/retrievers/recency_ranked.ts
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  private recentDocumentRanker(
+    documents: [Document, number][],
+    topK: number,
+    recencyWeight: number
+  ): [Document, number][] {
+    if (documents.length === 0) return [];
+
+    if (!documents.every(([doc, _]) => doc.metadata.date instanceof Date)) {
```
jacoblee93, line 52: Would suggest storing this as an ISO string -  not all (most?) vector stores will not deserialize dates

christian-bromann:
Thank you @Mraj23 for the contribution 🙏 Unfortunately at this point we no longer accept new additions to the langchain package anymore. I suggest the best approach is to:

- create your own repository to distribute LangChain integrations, e.g. `https://github.com/Mraj23/langchain-retriever`
- publish the package to NPM as e.g. `langchain-retriever`
- comment here so we can add it to the list of recommended integrations

Our team is still working on finding the ideal way to recommend integration packages like that to our community, if you have any feedback here, let me know. Please reach out if I can support with above steps and let me know when you have the package published.

Thank you!

PR #6685, opened 2024-09-04
feat(openai): baseUrl for DallEAPIWrapper

wanoo21:
This simple pull request adds the `baseUrl` to DallEAPIWrapper. It can be used for Cloudflare AI Gateway or other use cases.


jacoblee93:
Thank you!

PR #6701, opened 2024-09-06
Docs: Remove inline markdown links before truncating text

dinghar:
Note: I wasn't able to find a ticket for this, let me know if you'd like me to file one.

The `IndexTable` component truncates item descriptions at a fixed length. Some descriptions contain markdown links, which add to the total length, but are not rendered. This results in some very brief descriptions (example below from the VectorStores table).

This PR cleans markdown links from the strings before truncating them.

Screenshot 2024-09-05 at 5 19 30 PM


jacoblee93:
Thanks for this @dinghar!

I'm not sure this is an improvement - I think the proper fix would be to somehow be able to specify this field manually.

PR #6708, opened 2024-09-06
feat(groq): Add responseFormat to ChatGroqInput

allohamora:
This PR introduces the `responseFormat` field to `ChatGroqInput`, allowing `response_format` to be passed directly from the constructor. This provides a way to use `response_format` in scenarios where we cannot pass it as a call option, for example:
```ts
const chat = new ChatGroq({ responseFormat: { type: 'json_object' } });

  {
    message: new RunnablePassthrough(),
  },
  prompt,
  chat,
  outputParser,
]);

const response = await chain.invoke(message);
```

### DIFF
--- a/libs/langchain-groq/src/chat_models.ts
+++ b/libs/langchain-groq/src/chat_models.ts
@@ -73,7 +73,7 @@ export interface ChatGroqCallOptions extends BaseChatModelCallOptions {
   headers?: Record<string, string>;
   tools?: ChatGroqToolType[];
   tool_choice?: OpenAIClient.ChatCompletionToolChoiceOption | "any" | string;
-  response_format?: { type: "json_object" };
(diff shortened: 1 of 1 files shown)


jacoblee93:
Thanks for this!

I think doing this be equivalent?

```ts
const model = new ChatGroq();

const modelWithResponseFormat = model.bind({ response_format: "json_object" });
```

If so, would prefer to just add more examples of doing that.

allohamora:
> Thanks for this!
> 
> I think doing this be equivalent?
> 
> ```ts
> const model = new ChatGroq();
> 
> const modelWithResponseFormat = model.bind({ response_format: "json_object" });
> ```
> 
> If so, would prefer to just add more examples of doing that.

Yes, looks like an equivalent but with Runnable type instead of BaseChatModel
image

PR #6615, opened 2024-08-23
feat(anthropic): langchain-anthropic support custom create anthropic client

rxliuli:
Support for custom Anthropic clients using @langchain/anthropic, which will allow the use of @anthropic-ai/vertex-sdk and @anthropic-ai/bedrock-sdk, and also support custom Anthropic, such as supporting a modified version of the edge runtime.

```ts
import { expect, it } from 'vitest'
import { authenticate } from '../authenticate'
import { ChatAnthropic } from '@langchain/anthropic'
import { AnthropicVertex } from '@anthropic-ai/vertex-sdk'
import { GoogleAuth } from 'google-auth-library'
import { AnthropicVertexWeb } from '../AnthropicVertex'

it('Call Anthropic Vertex on nodejs', async () => {
  const client = new AnthropicVertex({
    region: import.meta.env.VITE_VERTEX_ANTROPIC_REGION,
    projectId: import.meta.env.VITE_VERTEX_ANTROPIC_PROJECTID,
    googleAuth: new GoogleAuth({
      credentials: {
        client_email: import.meta.env
          .VITE_VERTEX_ANTROPIC_GOOGLE_SA_CLIENT_EMAIL!,
        private_key: import.meta.env
          .VITE_VERTEX_ANTROPIC_GOOGLE_SA_PRIVATE_KEY!,
      },
      scopes: ['https://www.googleapis.com/auth/cloud-platform'],
    }),
  })
  const chat = new ChatAnthropic({
    apiKey: 'test',
    model: import.meta.env.VITE_VERTEX_ANTROPIC_MODEL,
    createClient: (() => client) as any,
  })
  const response = await chat.invoke([['human', 'Hello!']])
  console.log(response)
  expect(response).not.undefined
})
it('Call Anthropic Vertex on edge runtime', async () => {
  const token = await authenticate({
    clientEmail: import.meta.env.VITE_VERTEX_ANTROPIC_GOOGLE_SA_CLIENT_EMAIL!,
    privateKey: import.meta.env.VITE_VERTEX_ANTROPIC_GOOGLE_SA_PRIVATE_KEY!,
  })
  const client = new AnthropicVertexWeb({
    region: import.meta.env.VITE_VERTEX_ANTROPIC_REGION,
    projectId: import.meta.env.VITE_VERTEX_ANTROPIC_PROJECTID,
    accessToken: token.access_token,
  })
  const chat = new ChatAnthropic({
    apiKey: 'test',
[...]

PR #6716, opened 2024-09-07
fix(anthropic): Do not check for apiKey existence when createClient parameters are present

rxliuli:
Fixes https://github.com/langchain-ai/langchainjs/pull/6615#issuecomment-2334981613

### DIFF
--- a/libs/langchain-anthropic/src/chat_models.ts
+++ b/libs/langchain-anthropic/src/chat_models.ts
@@ -949,9 +949,6 @@ export class ChatAnthropicMessages<
   ): Promise<Anthropic.Message> {
     if (!this.batchClient) {
       const options = this.apiUrl ? { baseURL: this.apiUrl } : undefined;
-      if (!this.apiKey && !this.createClient) {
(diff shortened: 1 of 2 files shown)


jacoblee93:
Thank you! Good catch, sorry for missing this.

libs/langchain-anthropic/src/chat_models.ts
```diff
@@ -949,7 +949,7 @@ export class ChatAnthropicMessages<
   ): Promise<Anthropic.Message> {
     if (!this.batchClient) {
       const options = this.apiUrl ? { baseURL: this.apiUrl } : undefined;
-      if (!this.apiKey) {
+      if (!this.apiKey && !this.createClient) {
```
jacoblee93, line 952: I think `this.createClient` will always be defined since we have a default - can update

jacoblee93: Honestly, I don't really think we need a second check here. Will remove.

PR #6736, opened 2024-09-11
fix(langchain): Fix local file store traversal issue

jacoblee93:


langchain/src/storage/tests/file_system.test.ts
```diff
@@ -88,4 +88,19 @@ describe("LocalFileStore", () => {
...
+    const encoder = new TextEncoder();
+    const store = await LocalFileStore.fromPath(secondaryRootPath);
+    const value1 = new Date().toISOString();
+    await expect(
+      store.mset([["../foo", encoder.encode(value1)]])
+    ).rejects.toThrowError();
+    await expect(
+      store.mset([["/foo", encoder.encode(value1)]])
+    ).rejects.toThrowError();
+    await expect(store.mget(["../foo"])).rejects.toThrowError();
```
eyurtsev, line 102: maybe add test for `\` as well?

jacoblee93: Yeah good call

PR #6743, opened 2024-09-12
feat(langchain): Allow to passthrough model kwargs via createToolCallingAgent

OlegIvaniv:
This PR adds optional `kwargs` parameter to `createToolCallingAgent` which are passed to `llm.bindTools`. This is useful when user needs to specify some model provider parameters, for example to always require the model to use a tool(`{ tool_choice: 'required' }`)


langchain/src/agents/tool_calling/index.ts
```diff
@@ -40,6 +40,10 @@ export type CreateToolCallingAgentParams = {
    * allowing streaming of intermediate steps. Defaults to true.
    */
   streamRunnable?: boolean;
+  /**
+   * Optional kwargs to pass to the underlying model.
+   */
+  kwargs?: Record<string, unknown>;
```
bracesproul, line 46: Lets use the base interface, combined w/ `Record<string, any>` (not 100% sure if you need the `Partial<>`)
```suggestion
  kwargs?: BaseChatModelCallOptions & Partial<Record<string, any>>;
```

bracesproul (changes requested):
left one small comment, other than that this looks good!
Once fixed please re-request my review

OlegIvaniv:
Thanks, @bracesproul! Can you have another look, please?

OlegIvaniv:
Ping @bracesproul :P

jacoblee93:
Thanks for this @OlegIvaniv!

You should now be able to pass a `model.bindTools([...], { tool_choice: "required" })` directly into the factory method - would that be sufficient for your use case?

OlegIvaniv:
> Thanks for this @OlegIvaniv!
> 
> You should now be able to pass a `model.bindTools([...], { tool_choice: "required" })` directly into the factory method - would that be sufficient for your use case?

Hi, @jacoblee93. Yes, that would indeed solve it. Although the syntax feels a bit awkward now as you'd need to pass tools to both `model.bindTools`, then to `createToolCallingAgent`, and finally to `new AgentExecutor`, it's something I can live with. :D Thanks!

PR #6744, opened 2024-09-12
fix(langchain): Prevent exception when using plan & execute agent with no tools

OlegIvaniv:
When checking if the first tool is a dynamic structured tool in [getDefaultStepExecutor](https://github.com/hwchase17/langchainjs/blob/2ece692a4f1a10ee2ef4e117c1357bf132de0255/langchain/src/experimental/plan_and_execute/agent_executor.ts#L135) we should check if there are any tools. Passing an empty tools array results in exception since in `isDynamicStructuredTool`, we access `tool.constructor.lc_name`.


langchain/src/experimental/plan_and_execute/tests/plan_and_execute.int.test.ts
```diff
@@ -67,3 +67,21 @@ test.skip("Run agent with a sequential math problem", async () => {
...
+    verbose: true,
+  });
+  const executor = await PlanAndExecuteAgentExecutor.fromLLMAndTools({
+    llm: model,
+    tools: [],
+  });
+
+  // @eslint-disable-next-line/@typescript-eslint/ban-ts-comment
+  // @ts-expect-error unused var
+  const result = await executor.call({
```
bracesproul, line 84:
```suggestion
  await executor.call({
```
nit, if we're not using the var no need to declare it

bracesproul (approved):
lgtm, thanks for pushing this up!

PR #6756, opened 2024-09-13
fix(google): Fixed - TypeError: Cannot read properties of undefined (reading 'generationInfo')

### LINKED ISSUE #6371: google-common [bug]: Gemini response with no parts triggers error
### Privileged issue

### Issue Content

I am occasionally getting the following error during processing:
```
>  /Users/prisoner/Documents/project/functions/node_modules/@langchain/google-common/dist/utils/gemini.cjs:504
>                  generationInfo: ret[ret.length - 1].generationInfo,
>                                                      ^
>
>  TypeError: Cannot read properties of undefined (reading 'generationInfo')
>      at responseToChatGenerations (/Users/prisoner/Documents/project/functions/node_modules/@langchain/google-common/dist/utils/gemini.cjs:504:53)
>      at responseToChatResult (/Users/prisoner/Documents/projectfunctions/node_modules/@langchain/google-common/dist/utils/gemini.cjs:562:25)
>      at safeResponseTo (/Users/prisoner/Documents/project/functions/node_modules/@langchain/google-common/dist/utils/gemini.cjs:374:25)
>      at safeResponseToChatResult (/Users/prisoner/Documents/projectfunctions/node_modules/@langchain/google-common/dist/utils/gemini.cjs:570:12)
>      at ChatGoogle._generate (/Users/prisoner/Documents/project/functions/node_modules/@langchain/google-common/dist/chat_models.cjs:270:62)
```

Looking at the code, I'm guessing that the earlier call
```
const parts = responseToParts(response);
```
is returning an empty array, which is legitimate, but usually only happens when the safety handler would be triggered, so the safety handler should have caught it first.

Several things need to be addressed:
* `ret[ret.length - 1].generationInfo` should have a optional property accessor
  * We should probably just short circuit if no parts are returned
* Evaluate the `DefaultSafetyHandler` to see if other finish reasons should trigger a problem. (Perhaps as part of another issue)

Platform: Vertex AI
Models: 
* gemini-1.5-pro-001
* gemini-1.5-flex-001

@jacoblee93 and @bracesproul - Please assign this to me

pksharmapokhrel:
Fixes #6371 (issue)

While using ChatVertexAI, when the model throws RECITATION error and the response does not contains parts. Since the condition ```ret.every((item) => typeof item.message.content === "string")``` (Line 628) is true for ret = [], so it fails when we try to access "generationInfo" property of the first element of the ret in line ```generationInfo: ret[ret.length - 1].generationInfo``` (Line 658)

image

 Adding the checking ```ret.length > 0``` alongside the existing condition, this will fix the issue.


jacoblee93:
Thanks fro the PR!

@afirstenberg could you have a look as well if you have time?

afirstenberg:
I think this is an ok solution.

I would rather the solution we discussed in #6371 to be done, since I think it is easier and more obvious what is going on.

After line 626 add the lines:
```
    if (parts.length === 0) {
      return [];
    }
```

Most important - I think either should be added already and a release cut with it.

afirstenberg:
LGTM

Suggest merge and release.

jacoblee93:
Thank you!

PR #6768, opened 2024-09-13
docs: Fix Firecrawl v0 version in the docs

nickscamara:
Firecrawl integration is currently on v0 - which is supported until version 0.0.36.

@rafaelsideguide is working on a pr for v1 but meanwhile we should fix the docs.


jacoblee93:
Thanks!

PR #6803, opened 2024-09-15
fix(langchain): Remove circular peer dep in langchain on @langchain/community

jacoblee93:


jsumners-nr:
When will this be released to npmjs.com?

jacoblee93:
> When will this be released to npmjs.com?

It's live in langchain@0.3.2 and the latest 0.2.x!

jsumners-nr:
This is what I am seeing:

image

jacoblee93:
This PR only applies to `langchain`, core was fixed in 0.3.1.

PR #6832, opened 2024-09-17
fix(community): bump `@upstash/ratelimit` version

CahidArda:
Bumping Upstash Ratelimit version to benefit from the number of improvements we have made.


jacoblee93:
Flagging that this will break existing users but your call

libs/langchain-community/src/callbacks/handlers/upstash_ratelimit.ts
```diff
@@ -148,7 +148,7 @@ class UpstashRatelimitHandler extends BaseCallbackHandler {
     _name?: string
   ): Promise<void> {
     if (this.tokenRatelimit) {
-      const remaining = await this.tokenRatelimit.getRemaining(this.identifier);
+      const { remaining } = await this.tokenRatelimit.getRemaining(this.identifier);
```
jacoblee93, line 151: May be some way to make this backwards compatible and allow `"^1.1.3 || ^2.0.3"` in deps above

bracesproul: @CahidArda I just pushed up a commit allowing those versions

bracesproul (approved):

CahidArda:
Hi @jacoblee93,

can we go ahead with this PR?

jacoblee93:
Yes, thanks for your patience!

PR #6856, opened 2024-09-20
fix(langchain): Passing only a model string to initChatModel returns a chat model

bracesproul:


christian-bromann:
I will go ahead and close this due to inactivity. @bracesproul please ping me if there is still interest landing this change.

PR #6876, opened 2024-09-24
feat(anthropic,openai): Add payload formatting utils to Anthropic and OpenAI

jacoblee93:
Fixes # (issue)

### DIFF
--- a/libs/langchain-anthropic/src/tests/prompts.int.test.ts
+++ b/libs/langchain-anthropic/src/tests/prompts.int.test.ts
@@ -3,7 +3,7 @@ import { pull } from "langchain/hub";
 
 import { convertPromptToAnthropic } from "../utils/prompts.js";
 
-test("basic traceable implementation", async () => {
+test("Convert hub prompt to Anthropic payload and invoke", async () => {
   const prompt = await pull("jacob/joke-generator");
   const formattedPrompt = await prompt.invoke({
     topic: "cats",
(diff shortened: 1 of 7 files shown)


libs/langchain-anthropic/src/tests/prompts.int.test.ts
```diff
@@ -0,0 +1,25 @@
+import Anthropic from "@anthropic-ai/sdk";
+import { pull } from "langchain/hub";
+
+import { convertPromptToAnthropic } from "../utils/prompts.js";
+
+test("basic traceable implementation", async () => {
```
hinthornw, line 6: naming

libs/langchain-openai/src/tests/prompts.int.test.ts
```diff
@@ -0,0 +1,22 @@
+import OpenAI from "openai";
+import { pull } from "langchain/hub";
+
+import { convertPromptToOpenAI } from "../utils/prompts.js";
+
+test("basic traceable implementation", async () => {
```
hinthornw, line 6: ditto - naming

hinthornw (approved):

libs/langchain-anthropic/src/tests/prompts.int.test.ts
jacoblee93: Oops

PR #6910, opened 2024-09-30
fix(community): supabase comparison to string function

lakshyaag12:
# Summary
This PR fixes a minor error in the `SupabaseTranslator` file, specifically in the `visitComparisonAsString()` function that takes in a Comparison object and converts it to a format suitable for RPC.

# Details
- The change fixes an extra `}` that causes the API call to Supabase to return no rows.


jacoblee93:
Thank you!

Issue #6913, opened 2024-10-01
Column VEC_VECTOR has the wrong length


prashantrakheja:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain.js documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain.js rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```
  const args: HanaDBArgs = {
    connection: client,
    tableName: "my_hana_table",
  };
  const vectorStore = new HanaDB(embeddings, args);
  await vectorStore.initialize();
  ```

### Error Message and Stack Trace (if applicable)

[2024-09-30T20:00:34.304Z] ERROR    Error: Column VEC_VECTOR has the wrong length: 0
    at HanaDB.checkColumn (file:///Users/Desktop/code/typescript/node_modules/@langchain/community/dist/vectorstores/hanavector.js:245:23)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async HanaDB.initialize (file:///Users/Desktop/code/typescript/node_modules/@langchain/community/dist/vectorstores/hanavector.js:133:9)
    at async file:///Users/Desktop/code/typescript/dist/index.js:212:5

### Description

While trying to initialise the HANA db with langchain js library, I'm running into this issue.
 

### System Info

 "langchain": "^0.3.2"
 "@langchain/community": "^0.3.3"
 
 platform - mac M1 Max

jacoblee93:
Will ship a new release tomorrow 👍  thanks again!

PR #6914, opened 2024-10-01
fix(community): Column VEC_VECTOR has the wrong length

prashantrakheja:
* Description: This PR fixes an issue with SAP HANA Cloud QRC03 version. In that version the number to indicate no length being set for a vector column changed from -1 to 0. The change in this PR support both behaviours (old/new).
* Dependencies: No dependencies have been introduced.

The same issue has been fixed in the python library as well:
https://github.com/langchain-ai/langchain/pull/23516

Fixes # (issue)
https://github.com/langchain-ai/langchainjs/issues/6913


prashantrakheja:
cc @cinqisap @jacoblee93

jacoblee93:
Thank you!

PR #6924, opened 2024-10-02
feat(openai): include raw data in embedQuery method

CarlosSLoureiro:
I would like to contribute with this small feature because it is a need I have and perhaps others may have as well.

We have the option to obtain the full response data from OpenAI in completion methods such as `withStructuredOutput`, for example, but we don't have it in this embedding generation method. I need to obtain the data on the number of tokens sent without necessarily having another library estimate this data for me.

Here’s an example:

image

Additionally, this adjustment will not cause any changes to the existing interface unless we explicitly the `includeRaw` in the options param.

### DIFF
--- a/libs/langchain-community/src/vectorstores/tests/singlestore.int.test.ts
+++ b/libs/langchain-community/src/vectorstores/tests/singlestore.int.test.ts
@@ -1,7 +1,7 @@
 /* eslint-disable no-process-env */
 /* eslint-disable import/no-extraneous-dependencies */
 import { test, expect } from "@jest/globals";
-import { OpenAIEmbeddings } from "@langchain/openai";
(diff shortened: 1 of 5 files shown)


XStarlink:
Hi @CarlosSLoureiro , thanks for this very nice addition!

I was looking through the whole langchain-js github to see how openai embeddings were handled because like you I need to get the usages when I do embeddings:

`const embeddings = await openaiEmbeddings.embedDocuments(texts)`

I'd like to know the total cost of tokens to track my users' consumption in my saas.

Is it possible for you to add what you've done to the embedDocuments method? 🙏

CarlosSLoureiro:
Hi @XStarlink , that's a great suggestion, thanks. When I have some free time I'll make this adjustment and open another PR.

PR #6937, opened 2024-10-04
fix: allow numberic value for keepAlive property

mok-liee:
Hi, the ollama API is allowing to use numeric values.

If somone **trying** to use a valid numeric value eg. `-1` to the ChatOllama `keepAlive` prop via:

```js
new ChatOllama({
  baseUrl: ollamaEndpoint,
  model: model.model,
  temperature: 0.7,
  keepAlive: "-1"
})
```
Then the api will fail with `Bad Request` -> `missing unit in duration \"-1\"`.

I think it would be better to add `number` to the valid `keepAlive` types.
Then it is possible to call it like this:

```js
new ChatOllama({
  baseUrl: ollamaEndpoint,
  model: model.model,
  temperature: 0.7,
  keepAlive: -1
})
```

Here are some refs to the official ollama docs:

> If you're using the API, use the keep_alive parameter with the /api/generate and /api/chat endpoints to set the amount of time that a model stays in memory. The keep_alive parameter can be set to:
> 
> a duration string (such as "10m" or "24h")
> a number in seconds (such as 3600)
> any negative number which will keep the model loaded in memory (e.g. -1 or "-1m")
> '0' which will unload the model immediately after generating a response

https://github.com/ollama/ollama/blob/f40bb398f6423ae82386820f0abfe279045771b7/docs/faq.md#how-do-i-keep-a-model-loaded-in-memory-or-make-it-unload-immediately

### DIFF
--- a/libs/langchain-community/src/embeddings/ollama.ts
+++ b/libs/langchain-community/src/embeddings/ollama.ts
@@ -21,7 +21,7 @@ interface OllamaEmbeddingsParams extends EmbeddingsParams {
   headers?: Record<string, string>;
 
   /** Defaults to "5m" */
-  keepAlive?: string | number;
+  keepAlive?: string;
 
   /** Advanced Ollama API request parameters in camelCase, see
    * https://github.com/jmorganca/ollama/blob/main/docs/modelfile.md#valid-parameters-and-values
--- a/libs/langchain-community/src/embeddings/ollama.ts
+++ b/libs/langchain-community/src/embeddings/ollama.ts
@@ -21,7 +21,7 @@ interface OllamaEmbeddingsParams extends EmbeddingsParams {
   headers?: Record<string, string>;
 
   /** Defaults to "5m" */
-  keepAlive?: string;
+  keepAlive?: string | number;
 
   /** Advanced Ollama API request parameters in camelCase, see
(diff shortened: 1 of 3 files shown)


libs/langchain-community/src/embeddings/ollama.ts
```diff
@@ -21,7 +21,7 @@ interface OllamaEmbeddingsParams extends EmbeddingsParams {
   headers?: Record<string, string>;
 
   /** Defaults to "5m" */
-  keepAlive?: string;
+  keepAlive?: string | number;
```
bracesproul, line 24: The community implementations have been deprecated in favor of the `@langchain/ollama` package. 

```suggestion
  keepAlive?: string;
```

libs/langchain-community/src/utils/ollama.ts
```diff
@@ -7,7 +7,7 @@ export interface OllamaInput {
   f16KV?: boolean;
   frequencyPenalty?: number;
   headers?: Record<string, string>;
-  keepAlive?: string;
+  keepAlive?: string | number;
```
bracesproul, line 10: see above
```suggestion
  keepAlive?: string;
```

PR #6944, opened 2024-10-07
feat(docs,core): Adds error reference pages and populate codes on errors in core

jacoblee93:
@efriis


docs/core_docs/docs/troubleshooting/errors/GRAPH_RECURSION_LIMIT.mdx
```diff
@@ -0,0 +1,14 @@
+# GRAPH_RECURSION_LIMIT
+
+Your LangGraph app reached the maximum number of steps before hitting a stop condition.
+
+This is usually a sign of an infinite loop, but complex graphs may hit this naturally.
```
efriis, line 5: should we have an example repro of how to trigger this error for mental pattern matching?

docs/core_docs/docs/troubleshooting/errors/INVALID_GRAPH_UPDATE.mdx
```diff
@@ -0,0 +1,40 @@
+# INVALID_GRAPH_UPDATE
+
+A LangGraph app received an unexpected return value from a node and failed to update its state at the end of a step.
```
efriis, line 3: are we certain this is the only condition that creates this? can it also be an input with a thread?

jacoblee93: Hmm probably not, might be worth splitting the error more granularly

jacoblee93: Yeah let's actually split this into different cases and keep this focused on parallel updates

PR #6950, opened 2024-10-08
core[patch], openai[patch]: add UsageMetadata details

baskaryan:


libs/langchain-openai/src/tests/chat_models.standard.int.test.ts
```diff
@@ -65,3 +111,26 @@ test("ChatOpenAIStandardIntegrationTests", async () => {
   const testResults = await testClass.runTests();
   expect(testResults).toBe(true);
 });
+
+async function invoke(
```
bracesproul, line 115: I like this pattern! Might be useful to add to the base standard tests library as an exported util. Not important to add for now though

jacoblee93:
Thank you!

PR #6963, opened 2024-10-10
feat(community): Return document ids when querying pgvector store

elliots:
Currently the ids are not returned when querying a pgvector store. This adds them to the returned document.


jacoblee93:
Looks good, thanks!

PR #6967, opened 2024-10-11
feat(core): Adds support for context variables

jacoblee93:
Modifies the `RunTree` object to maintain backwards compatibility with older LangSmith instances

### DIFF
--- a/langchain-core/src/context.ts
+++ b/langchain-core/src/context.ts
@@ -4,6 +4,8 @@ import {
   _CONTEXT_VARIABLES_KEY,
   AsyncLocalStorageProviderSingleton,
 } from "./singletons/index.js";
+import { RunTree } from "langsmith";
+import { isRunTree } from "langsmith/run_trees";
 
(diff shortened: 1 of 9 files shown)


langchain-core/src/context.ts
```diff
@@ -0,0 +1,126 @@
...
+ * // undefined, because context variable has not been set yet
+ * console.log(getContextVariable("foo"));
+ *
+ * // Final return value is "baz"
+ * const result = await runnable.invoke({});
+ * ```
+ *
+ * @param name The name of the context variable.
+ */
+export function getContextVariable(name: PropertyKey) {
```
dqbd, line 123: _question_: this is automatically typed as `any` or `unknown`?

dqbd (approved):

langchain-core/src/context.ts
jacoblee93: Automatically typed as `any` but can make it explicit.

PR #6979, opened 2024-10-13
fix(docs): usage web-streams-polyfill

electroluxcode:
it is useless when `import "web-streams-polyfill/es6"` ，When `web-streams-polyfill` is at version 3, 

it should be referenced like in the file `langchain-core\src\utils\tests\polyfill_stream.test.ts`. That is, import "web-streams-polyfill".

usage_polyfill


jacoblee93:
Thanks for flagging!

Should we instead fix support for v4?

electroluxcode:
> Thanks for flagging!
> 
> Should we instead fix support for v4?

Sure. In version 4, only need to change the import method to `import "web-streams-polyfill/polyfill"`

jacoblee93:
Thank you!

PR #6997, opened 2024-10-16
fix(anthropic): Pass back full Anthropic content for stream events

jacoblee93:


libs/langchain-anthropic/src/tests/chat_models-tools.int.test.ts
```diff
@@ -202,6 +203,73 @@ test("Can bind & stream AnthropicTools", async () => {
...
+    "What is the weather in London today?",
+    {
+      version: "v2",
+    }
+  );
+
+  const chatModelStreamEvents = [];
+  for await (const event of eventStream) {
+    if (event.event === "on_chat_model_stream") {
+      console.log(event);
```
dqbd, line 261: _nit_: remove

dqbd (approved):

libs/langchain-anthropic/src/tests/chat_models-tools.int.test.ts
jacoblee93: Oops

PR #6999, opened 2024-10-17
feat(google-vertexai): Support Non-Google and Model Garden models in Vertex AI - Anthropic integration

### LINKED ISSUE #2562: Add support for Vertex AI Model Garden / Custom Models
See langchain-ai/langchain#8850 and langchain-ai/langchain#9367 and will talk to @lkuligin about their implementation.

The general idea will be to provide a standard way to either access instances of these models at a designated endpoint directly, or create a subclass that can do so. (The request and response format have a standard, but it is a vague standard in some places.)

### LINKED ISSUE #6207: google-common [minor]: Custom events containing request/response JSON
### Privileged issue

### Issue Content

It would be useful, particularly for debugging purposes, if there were custom callback events that include the JSON of the exact request and response, and possibly other attributes such as the endpoint sent to, response code, etc.

See https://js.langchain.com/v0.2/docs/how_to/callbacks_custom_events/

afirstenberg:
Fixes #2562 
Fixes #6207 

The ultimate objective is to add full support to Vertex AI for the various 3P models that are provided via API as well as custom user endpoints they may deploy for personal models

This will be done through an API object that provides methods that both turn the LangChain objects into model-API compatible objects that will be formatted in JSON as well as take the JSON result and convert it to LangChain objects.

In this first batch, we get the Gemini support integrated with this API object and add Anthropic / Claude support.

This work was supported by Google Cloud Credits provided by Google.


afirstenberg:
@bracesproul @jacoblee93 - I believe this is ready for review and, if things look good, integration and release. 

This is just the Anthropic updates. As the other supported models are ready, I'll be doing separate PRs for them. (I don't want to hold things up.)

afirstenberg:
Hmmm... I see the docs have failed, but I don't see why.

libs/langchain-google-common/src/chat_models.ts
```diff
@@ -398,8 +351,10 @@ export abstract class ChatGoogleBase<AuthOptions>
                 usage_metadata: usageMetadata,
               }),
             });
-      yield chunk;
-      await runManager?.handleLLMNewToken(chunk.text);
+      if (chunk) {
+        yield chunk;
+        await runManager?.handleLLMNewToken(chunk.text);
```
jacoblee93, line 356: Unrelated nit: good to pass the full generation chunk here for other things

For example:

https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-openai/src/chat_models.ts#L1427

libs/langchain-google-common/src/connection.ts
```diff
@@ -178,6 +179,22 @@ export abstract class GoogleHostConnection<
     return "gcp";
   }
 
+  get location(): string {
+    return this._location ?? this.computedLocation;
+  }
+
+  get computedLocation(): string {
```
jacoblee93, line 186: Any point to having a base implementation here? One less level of indirection to just put it in `location` above

libs/langchain-google-common/src/connection.ts
```diff
@@ -178,6 +179,22 @@ export abstract class GoogleHostConnection<
...
+    return "us-central1";
+  }
+
+  get endpoint(): string {
+    return this._endpoint ?? this.computedEndpoint;
+  }
+
+  get computedEndpoint(): string {
```
jacoblee93, line 194: See above

libs/langchain-google-common/src/connection.ts
```diff
@@ -245,6 +288,19 @@ export abstract class GoogleAIConnection<
     }
   }
 
+  get computedLocation(): string {
+    switch (this.apiName) {
+      case "google":
+        return super.computedLocation;
```
jacoblee93, line 294: Why isn't this just `super.location`?

libs/langchain-google-common/src/connection.ts
```diff
@@ -245,6 +288,19 @@ export abstract class GoogleAIConnection<
...
+      case "anthropic":
+        return "us-east5";
```
jacoblee93, line 296: Should we be hardcoding this here?

If it's only available in one region now, would prefer to have this configurable or even not have a default at all and just have it documented

libs/langchain-google-common/src/connection.ts
```diff
@@ -298,141 +382,216 @@ export abstract class AbstractGoogleLLMConnection<
...
-  structuredToolToFunctionDeclaration(
-    tool: StructuredToolParams
-  ): GeminiFunctionDeclaration {
-    const jsonSchema = zodToGeminiParameters(tool.schema);
-    return {
-      name: tool.name,
-      description: tool.description ?? `A function available to call.`,
-      parameters: jsonSchema,
-    };
+export class GoogleRequestLogger extends GoogleRequestCallbackHandler {
```
jacoblee93, line 504: This is intended for public use? Or just your own debugging?

libs/langchain-google-common/src/utils/anthropic.ts
```diff
@@ -0,0 +1,723 @@
...
+  GeminiTool,
+  GoogleAIAPI,
+  GoogleAIModelParams,
+  GoogleAIModelRequestParams,
+  GoogleAIToolType,
+  GoogleLLMResponse,
+} from "../types.js";
+
+export function getAnthropicAPI(config?: AnthropicAPIConfig): GoogleAIAPI {
+  // function notImplemented(): never {
```
jacoblee93, line 54: nit: remove

libs/langchain-google-common/src/utils/anthropic.ts
```diff
@@ -0,0 +1,723 @@
...
+    }
+  }
+
+  /**
+   * Normalize the AIMessageChunk.
+   * If the fields are just a string - use that as content.
+   * If the content is an array of just text fields, turn them into a string.
+   * @param fields
+   */
+  function newAIMessageChunk(fields: string | AIMessageFields): AIMessageChunk {
```
jacoblee93, line 87: Think it's worth reusing some logic from `@langchain/anthropic`?

https://github.com/langchain-ai/langchainjs/tree/main/libs/langchain-anthropic/src/utils

Could even make it a dep

jacoblee93 (commented):
Overall looks fantastic - left some nits/comments and will try myself before cutting a release

libs/langchain-google-common/src/chat_models.ts
afirstenberg: I think I'm missing something about the nit. What about the "full generation chunk" isn't being passed here?

libs/langchain-google-common/src/connection.ts
afirstenberg: Initially for my debugging, but there have been cases where people have reported problems that I couldn't duplicate. Being able to easily have them add this logger to the callbacks to see what the request and response were would make it easier to create a test and a fix.

So it is rare that anyone would need to use it, but invaluable to have in place if they do.

libs/langchain-google-common/src/utils/anthropic.ts
afirstenberg: Whoops. Yup. Removing. Thanks for that catch.

libs/langchain-google-common/src/utils/anthropic.ts
afirstenberg: I considered it, and even tried for a bit, but ultimately decided not to for a few reasons:
* It adds an unnecessary dependency in the case where people don't intend to use Claude on GCP. Including pulling in the Anthropic library.
* Some of the types aren't exactly compatible (I don't remember which ones now - I just remember starting down this road and running into issues).

The code you highlighted here has more to do with how my logic was generating the content objects and how there are expectations in parts of the base code that `AIMessageChunk.content` should be a string. So this normalizes it in those cases. (Which are almost all the cases for AI Messages historically.)

libs/langchain-google-common/src/connection.ts
afirstenberg: Mostly this is how location and endpoint evolved and come from a few requirements:
* The endpoint is usually, but not always, location based. So it may need to be specified separately from location.
* Other elements of the path may have location. So location may need to be specified separately from endpoint.
* In the simple cases, I didn't want people to have to specify both, but usually just location. But have them be able to if necessary.

Most resources are available at us-central1. But, apparently not all. So I wanted subclasses to be able to set their default location if appropriate.

And I wanted to be able to support default locations. Requiring a developer to have to know more magic values makes for less manageable code. Especially if we can avoid it. (We support default values all over the place.)

libs/langchain-google-common/src/connection.ts
afirstenberg: See above.

libs/langchain-google-common/src/connection.ts
afirstenberg: `computedLocation` is typically only called from `location`. Keeping it calling the same method on the superclass keeps it clean and avoids additional logic processing.

libs/langchain-google-common/src/connection.ts
afirstenberg: See above about making magic values as unnecessary as possible.

Some of the Anthropic models are available in other regions, but all are available in us-east5. And none, mysteriously, are available in us-central1.

I'd like to keep it as a default, but clearly have it available to be settable.

libs/langchain-google-common/src/utils/anthropic.ts
afirstenberg: (To be clear, however, I'm open minded on this. I don't like duplicating code.)

afirstenberg:
Thanks! Very much appreciate the feedback.
I've explained my thought process on the points above, but happy to discuss them further.

Any thoughts on the problems with docs?

afirstenberg:
@jacoblee93 - Thanks! Very much appreciate the feedback.
I've explained my thought process on the points above, but happy to discuss them further.

Any thoughts on the problems with docs?

libs/langchain-google-common/src/chat_models.ts
afirstenberg: Ah! Got it! Fixed.

libs/langchain-google-common/src/types.ts
```diff
@@ -45,10 +57,59 @@ export interface GoogleConnectionParams<AuthOptions>
   platformType?: GooglePlatformType;
 }
 
+export enum GoogleAISafetyCategory {
```
jacoblee93, line 60: Ah we can't use enum for tree shaking reasons - can we just use a plain object?

afirstenberg: Will do!

PR #7012, opened 2024-10-17
feat(openai): Support audio output

bracesproul:
TODO:
- run standard int tests
- test azure
- add docs

### DIFF
--- a/libs/langchain-openai/src/chat_models.ts
+++ b/libs/langchain-openai/src/chat_models.ts
@@ -385,6 +385,26 @@ export interface ChatOpenAICallOptions
    * @version 0.2.6
    */
   strict?: boolean;
+  /**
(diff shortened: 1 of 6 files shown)


libs/langchain-openai/src/chat_models.ts
```diff
@@ -1190,6 +1207,8 @@ export class ChatOpenAI<
       seed: options?.seed,
       ...streamOptionsConfig,
       parallel_tool_calls: options?.parallel_tool_calls,
+      audio: this.audio,
```
jacoblee93, line 1210: May be slightly safer to add these fields only if they are present - not sure how Azure or other model proxies will react to adding `audio: undefined`

jacoblee93 (commented):
Should add this to `CallOptions` and add a test for `.bind`

Issue #5128, opened 2024-04-17
Google Gemini + DynamoDB chat memory fails by getting an empty string chunk


albertodeavila:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain.js documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain.js rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```typescript 

const llm = new ChatGoogleGenerativeAI({
	apiKey: googleAIApiKey,
	model: "gemini-pro",
	maxOutputTokens: 2048,
	safetySettings: [
		{
			category: HarmCategory.HARM_CATEGORY_HARASSMENT,
			threshold: HarmBlockThreshold.BLOCK_LOW_AND_ABOVE,
		},
	],
	temperature: temperature ? parseFloat(temperature) : 0.7,
        frequencyPenalty: frequencyPenalty ? parseFloat(frequencyPenalty) : 0,
        streaming: true
})

const memory = new BufferMemory({
      chatHistory: new DynamoDBChatMessageHistory({
	      tableName: CHAT_HISTORY_TABLE_NAME,
	      partitionKey: HASH_KEY,
	      sessionId: getChatHistoryId(),
	      config: {
		      region: REGION_AWS,
	      },
      }),
})

const chain = new ConversationChain({llm, memory})
await chain.call({
	input: prompt,
	callbacks: [
		{
			handleLLMNewToken(token: string) {
				aiOutput += token
				responseStream.write(token)
			}
		}
	]
})
```

### Error Message and Stack Trace (if applicable)

Error message: 
> Error during prompt execution Error: output values have 1 keys, you must specify an output key or pass only 1 key as output

Stack trace: 
```
    at getOutputValue 
    at BufferMemory.saveContext 
    at process.processTicksAndRejections 
    at async ConversationChain.invoke 
```

### Description

When using Langchain to execute prompts agains Google Gemini Pro + using DynamoDB-Backed Chat Memory, we are getting an error, cause Google Gemini is responding an empty string and dynamoDB is trying to store it, but that generates an exception. That was working few weeks ago, maybe because Google Gemini wasn't responding with an empty string, but I think it should be fixed on Langchain. 

So the error comes from this method: 
```typescript
var getOutputValue = (outputValues, outputKey) => {
[...]

PR #7019, opened 2024-10-18
fix(core): Handle empty model output in memory

burivuhster:
## Summary

This PR addresses the issue that happens when a model is used together with a memory, and LLM produces an empty output (empty string).

Because of the coercion to boolean, saving the history to memory fails when the output is an empty string:
https://github.com/langchain-ai/langchainjs/blob/391bc4645036fd0c2c41b5ed35236919732edf2a/langchain-core/src/memory.ts#L93C1-L98C4

Here is the minimal reproducible example:
```js
const { ChatOpenAI } = require("@langchain/openai");
const { initializeAgentExecutorWithOptions } = require("langchain/agents");
const { SerpAPI } = require("langchain/tools");
const { Calculator } = require("@langchain/community/tools/calculator");
const { BufferMemory } = require("langchain/memory");

// Initialize the language model
const model = new ChatOpenAI({ temperature: 0, model: 'gpt-4o-mini' });

// Define tools
const tools = [
  new Calculator(),
];

// Initialize memory
const memory = new BufferMemory({
  memoryKey: "chat_history",
  returnMessages: true,
});

async function main() {
  // Initialize the agent
  const executor = await initializeAgentExecutorWithOptions(tools, model, {
    agentType: "chat-conversational-react-description",
    verbose: true,
    memory: memory,
  });

  // Run the agent
  const result1 = await executor.call({
    input: "Answer with an empty string without any explanations and without any additional characters",
  });
  console.log(result1.output);
}

main().catch(console.error);
```

It throws an exception: `Error: output values have 1 keys, you must specify an output key or pass only 1 key as output`

## Related issues
- https://github.com/langchain-ai/langchainjs/issues/5128

### DIFF
--- a/langchain-core/src/memory.ts
+++ b/langchain-core/src/memory.ts
@@ -90,7 +90,7 @@ export const getOutputValue = (
   outputKey?: string
 ) => {
   const value = getValue(outputValues, outputKey);
-  if (!value) {
+  if (!value && value !== "") {
     const keys = Object.keys(outputValues);
     throw new Error(
       `output values have ${keys.length} keys, you must specify an output key or pass only 1 key as output`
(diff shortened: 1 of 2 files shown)


jacoblee93:
Thanks for this! Happy to merge but note that we now suggest using other things for memory:

https://js.langchain.com/docs/versions/migrating_memory/
https://langchain-ai.github.io/langgraphjs/how-tos/persistence/

PR #7025, opened 2024-10-18
fix(community): AILanguageModel interface

hemanth:
Using latest interface definitions from prompt-api spec.

### DIFF
--- a/libs/langchain-community/src/experimental/llms/chrome_ai.ts
+++ b/libs/langchain-community/src/experimental/llms/chrome_ai.ts
@@ -5,25 +5,82 @@ import { GenerationChunk } from "@langchain/core/outputs";
 import { IterableReadableStream } from "@langchain/core/utils/stream";
 import { BaseLLMParams, LLM } from "@langchain/core/language_models/llms";
 
-export interface AI {
-  canCreateTextSession(): Promise<AIModelAvailability>;
-  createTextSession(options?: AITextSessionOptions): Promise<AITextSession>;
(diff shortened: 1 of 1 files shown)


libs/langchain-community/src/experimental/llms/chrome_ai.ts
```diff
@@ -5,25 +5,82 @@ import { GenerationChunk } from "@langchain/core/outputs";
...
 }
 
-export interface AITextSession {
-  prompt(input: string): Promise<string>;
-  promptStreaming(input: string): ReadableStream;
+export interface AILanguageModel extends EventTarget {
+  prompt(input: AILanguageModelPromptInput, options?: AILanguageModelPromptOptions): Promise<String>;
+  promptStreaming(input: AILanguageModelPromptInput, options?: AILanguageModelPromptOptions): ReadableStream;
+
+  countPromptTokens(input: AILanguageModelPromptInput, options?: AILanguageModelPromptOptions): Promise<number>;
```
hemanth, line 17: I have updated the interface to the latest definition, we may implement these if it make sense.

jacoblee93: Given this will likely change again, is it worth adding all of this if we're not using it?

jacoblee93: Eh it's experimental I suppose so it's fine

jacoblee93:
Thank you!

PR #7027, opened 2024-10-19
feat: setenv fn used to configure Transformers.js env

electroluxcode:
refer:https://huggingface.co/docs/transformers.js/main/en/api/env#env

When I want to set the cache configuration of the transformer, I find that there seems to be a lack of methods to change the configuration of `Transformers.js`

usage
```js
let hgEmbeddings= new HuggingFaceTransformersEmbeddings({ model: "Xenova/all-MiniLM-L6-v2" })
hgEmbeddings.setEnv(KEY,VALUE)
```


jacoblee93:
Thanks for this!

Is there any advantage to exposing this on the LangChain class vs. just importing the object from Transformers.js directly?

electroluxcode:
Haha, I overcomplicated the problem. I will close this MR. Thank you for your reply

PR #7033, opened 2024-10-21
feat(cosmosdbnosql): Add Semantic Cache Integration

aditishree1:
This PR adds following changes:

- adds Azure Cosmos DB NoSQL SemanticCache Integration.
- adds docs and samples for Semantic Cache


libs/langchain-azure-cosmosdb/src/azure_cosmosdb_nosql.ts
```diff
@@ -78,7 +78,7 @@ export interface AzureCosmosDBNoSQLConfig
   readonly metadataKey?: string;
 }
 
-const USER_AGENT_PREFIX = "langchainjs-azure-cosmosdb-nosql";
+const USER_AGENT_SUFFIX = "LangChain-CDBNoSQL-VectorStore-JavaScript";
```
sinedied, line 81: Could you explain why you made this change?
The naming is following a common pattern we have for all our JS integrations, which makes it easier to use filters: `<framework>-<integration_name>`
If we need a distinction, it would be best to use `<framework>-<integration>-<type>`, ie `langchainjs-azure-cosmosdb-nosql-vectorstore`

libs/langchain-azure-cosmosdb/src/caches.ts
```diff
@@ -0,0 +1,171 @@
...
+import { EmbeddingsInterface } from "@langchain/core/embeddings";
+import { CosmosClient, CosmosClientOptions } from "@azure/cosmos";
+import { DefaultAzureCredential } from "@azure/identity";
+import { getEnvironmentVariable } from "@langchain/core/utils/env";
+import {
+  AzureCosmosDBNoSQLConfig,
+  AzureCosmosDBNoSQLVectorStore,
+} from "./azure_cosmosdb_nosql.js";
+
+const USER_AGENT_SUFFIX = "LangChain-CDBNoSQL-SemanticCache-JavaScript";
```
sinedied, line 18: If possible, it would be preferable to use `langchainjs-azure-cosmosdb-nosql-semanticcache` (see previous comment)

libs/langchain-azure-cosmosdb/src/azure_cosmosdb_nosql.ts
```diff
@@ -68,7 +68,7 @@ export interface AzureCosmosDBNoSQLInitOptions {
  */
 export interface AzureCosmosDBNoSQLConfig
   extends AzureCosmosDBNoSQLInitOptions {
-  readonly client?: CosmosClient;
+  client?: CosmosClient;
```
sinedied, line 71: The client should not be overridable by the user aside from the constructor, please keep it read-only. When creating the new client in the semantic cache, you can use the constructor to use it.

libs/langchain-azure-cosmosdb/src/caches.ts
```diff
@@ -0,0 +1,171 @@
...
+    } else {
+      this.config = dbConfig;
+    }
+    this.embeddings = embeddings;
+  }
+
+  private getLlmCache(llmKey: string) {
+    const key = getCacheKey(llmKey);
+    if (!this.cacheDict[key]) {
+      this.cacheDict[key] = new AzureCosmosDBNoSQLVectorStore(
```
sinedied, line 103: In the current state of this implementation, this use the same default container name as for the VectorStore, which can be problematic:

For example, if a user uses default values and have a vector store for RAG and semantic cache, the results will get mixed up.

I suggest 2 changes:
- Add a metadata to indicate that the document is used for semantic caching (so users wants to use a single table, they can use filters to distinguish cache vs documents)
- Use a different container name for the semantic caching by default, for example `vectorSearchContainer` to avoid conflicts

sinedied (changes requested):

libs/langchain-azure-cosmosdb/src/azure_cosmosdb_nosql.ts
aayush3011: @sinedied we are following a user agent pattern across all ai integrations, and we came up with "LangChain-CDBNoSQL-VectorStore-JavaScript"

libs/langchain-azure-cosmosdb/src/caches.ts
aayush3011: @sinedied we are following a user agent pattern across all ai integrations, and we came up with "LangChain-CDBNoSQL-VectorStore-JavaScript"

libs/langchain-azure-cosmosdb/src/caches.ts
aditishree1: added different default database and container name for semantic cache

libs/langchain-azure-cosmosdb/src/caches.ts
```diff
@@ -0,0 +1,172 @@
...
+import { DefaultAzureCredential } from "@azure/identity";
+import { getEnvironmentVariable } from "@langchain/core/utils/env";
+import {
+  AzureCosmosDBNoSQLConfig,
+  AzureCosmosDBNoSQLVectorStore,
+} from "./azure_cosmosdb_nosql.js";
+
+const USER_AGENT_SUFFIX = "LangChain-CDBNoSQL-SemanticCache-JavaScript";
+const DEFAULT_DATABASE_NAME = "CosmosNoSqlCacheDB";
+const DEFAULT_CONTAINER_NAME = "CosmosNoSqlCacheContainer";
```
sinedied, line 20:
```suggestion
const DEFAULT_CONTAINER_NAME = "semanticCacheContainer";
```
Aligning on other integrations

libs/langchain-azure-cosmosdb/src/caches.ts
```diff
@@ -0,0 +1,172 @@
...
+import { CosmosClient, CosmosClientOptions } from "@azure/cosmos";
+import { DefaultAzureCredential } from "@azure/identity";
+import { getEnvironmentVariable } from "@langchain/core/utils/env";
+import {
+  AzureCosmosDBNoSQLConfig,
+  AzureCosmosDBNoSQLVectorStore,
+} from "./azure_cosmosdb_nosql.js";
+
+const USER_AGENT_SUFFIX = "LangChain-CDBNoSQL-SemanticCache-JavaScript";
+const DEFAULT_DATABASE_NAME = "CosmosNoSqlCacheDB";
```
sinedied, line 19:
```suggestion
const DEFAULT_DATABASE_NAME = "semanticCacheDB";
```
Aligning on other integrations.

@aayush3011 I was thinking that it might be even better for the DB to use a default common one for the VectorSearch and SemanticCache integration? In that case, you could just remove the default DB name here and let it fall back to the vector search one if it's not defined.

This way users would by default have a unique DB, with one container for each use case, simplifying the setup.

sinedied (changes requested):
I think it's best to align the default container naming with the other integrations.
For the DB and user agents let's discuss this offline

docs/core_docs/docs/integrations/platforms/microsoft.mdx
```diff
@@ -132,6 +132,24 @@ See a [usage example](/docs/integrations/vectorstores/azure_cosmosdb_mongodb).
 import { AzureCosmosDBMongoDBVectorStore } from "@langchain/azure-cosmosdb";
 ```
 
+## Semantic Cache
+
+### Azure Cosmos DB NoSQL Semantic Cache
+
+> The Semantic Cache feature is supported with Azure Cosmos DB for NoSQL integration, enabling users to retrieve cached responses based on semantic similarity between the user input and previously cached results. It leverages AzureCosmosDBNoSQLVectorStore, which stores vector embeddings of cached prompts. These embeddings enable similarity-based searches, allowing the system to retrieve relevant cached results.
```
sinedied, line 139: nit: a link to `AzureCosmosDBNoSQLVectorStore` doc would be nice here

docs/core_docs/docs/integrations/semantic_caches/azure_cosmosdb_nosql.mdx
```diff
@@ -0,0 +1,40 @@
+# Azure Cosmos DB NoSQL Semantic Cache
+
+> The Semantic Cache feature is supported with Azure Cosmos DB for NoSQL integration, enabling users to retrieve cached responses based on semantic similarity between the user input and previously cached results. It leverages AzureCosmosDBNoSQLVectorStore, which stores vector embeddings of cached prompts. These embeddings enable similarity-based searches, allowing the system to retrieve relevant cached results.
```
sinedied, line 3: nit: a link to AzureCosmosDBNoSQLVectorStore doc would be nice here

libs/langchain-azure-cosmosdb/src/caches.ts
```diff
@@ -0,0 +1,172 @@
...
+import { CosmosClient, CosmosClientOptions } from "@azure/cosmos";
+import { DefaultAzureCredential } from "@azure/identity";
+import { getEnvironmentVariable } from "@langchain/core/utils/env";
+import {
+  AzureCosmosDBNoSQLConfig,
+  AzureCosmosDBNoSQLVectorStore,
+} from "./azure_cosmosdb_nosql.js";
+
+const USER_AGENT_SUFFIX = "langchainjs-cdbnosql-semanticcache-javascript";
+const DEFAULT_DATABASE_NAME = "semanticCacheDB";
```
sinedied, line 19:
```suggestion
```
I thought we agreed that we should use the same default DB (but with a different container) as for vector search to simplify the setup? In that case, we should fall back to the default one of the vector search implementation if it's not defined.

libs/langchain-azure-cosmosdb/src/caches.ts
```diff
@@ -0,0 +1,172 @@
...
+        } as CosmosClientOptions);
+      }
+    } else {
+      client = dbConfig.client;
+    }
+
+    this.config = {
+      ...dbConfig,
+      client,
+      databaseName: dbConfig.databaseName ?? DEFAULT_DATABASE_NAME,
```
sinedied, line 95:
```suggestion
      databaseName: dbConfig.databaseName,
```
If param is not defined, will fall back to vector search default

sinedied (changes requested):

sinedied (approved):

jacoblee93:
Looks good to me! Only nit is to put docs under `caches/` vs `semantic_caches`. I can have a look.

jacoblee93:
One more thing you may want to consider - happy to merge but you may want to return `undefined` in certain cases

aayush3011 (approved):
LGTM, thanks @aditishree1 .

aditishree1:
> One more thing you may want to consider - happy to merge but you may want to return `undefined` in certain cases if the score is not high enough

make sense. Any suggestion on what should be the threshold score?

jacoblee93:
> > One more thing you may want to consider - happy to merge but you may want to return `undefined` in certain cases if the score is not high enough
> 
> 
> 
> make sense. Any suggestion on what should be the threshold score?

No good intuition - would try it with some things you'd expect to be cached and allow the user to override it

libs/langchain-azure-cosmosdb/src/caches.ts
```diff
@@ -0,0 +1,171 @@
...
+   * Retrieves data from the cache.
+   *
+   * @param prompt The prompt for lookup.
+   * @param llmKey The LLM key used to construct the cache key.
+   * @returns An array of Generations if found, null otherwise.
+   */
+  public async lookup(prompt: string, llmKey: string) {
+    const llmCache = this.getLlmCache(llmKey);
+
+    const results = await llmCache.similaritySearch(prompt, 1);
```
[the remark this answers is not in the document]
sinedied: Thanks for the catch!

aditishree1:
Hi @jacoblee93 @sinedied @aayush3011, I have added similarity score threshold in the last commit. I have set the default threshold as 0.6 by trying out some prompts. Also added the tests. Pls review.

examples/src/caches/azure_cosmosdb_nosql/azure_cosmosdb_nosql.ts
```diff
@@ -11,7 +11,21 @@ const config: AzureCosmosDBNoSQLConfig = {
...
   connectionString: "<CONNECTION_STRING>",
 };
-const cache = new AzureCosmosDBNoSQLSemanticCache(embeddings, config);
+
+/**
+ * Sets the threshold similarity score for returning cached results based on vector distance.
+ * Cached output is returned only if the similarity score meets or exceeds this threshold;
+ * otherwise, a new result is generated. Default is 0.6, adjustable via the constructor
+ * to suit various distance functions and use cases.
+ * (see: https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/query/vectordistance).
```
sinedied, line 20:
```suggestion
 * (see: https://learn.microsoft.com/azure/cosmos-db/nosql/query/vectordistance).
```

libs/langchain-azure-cosmosdb/src/tests/caches.int.test.ts
```diff
@@ -31,33 +31,42 @@ function vectorEmbeddingPolicy(
         path: "/embedding",
         dataType: "float32",
         distanceFunction,
-        dimensions: 0,
+        dimensions: 505,
```
sinedied, line 34: Would be best to pass embeddings as input of the `vectorEmbeddingPolicy` function, and generate a test embedding to get the length. This way the tests can be run with any embedding model

sinedied (approved):
Looks good, I put a minor comment regarding the tests to make it work with any embedding model

aditishree1:
Hi @jacoblee93, I have added changes to set a threshold score. Please have a look.

jacoblee93:
Awesome, thanks will merge and ship a new package version once it's in!

PR #7036, opened 2024-10-21
feat(community): ElasticVectorSearch: add a not_exists filter 

LaurensTsestigh:
We have recently implemented a new functionality in our application where the customer is able to archive items.
As we want to exclude archived items by default from our vectorstore search we need a filter to add a statement to the must_not  filtering being a must_not exists and then we pass ```{ field: 'archivedAt', operator: 'not_exists' }``` .
If we then want to include archived items later one, we can just not pass this filter.

I think by adding this custom check, we add a lot of flexibility to the similarity search as well.

### DIFF
--- a/libs/langchain-community/src/vectorstores/elasticsearch.ts
+++ b/libs/langchain-community/src/vectorstores/elasticsearch.ts
@@ -342,6 +342,12 @@ export class ElasticVectorSearch extends VectorStore {
             field: metadataField,
           },
         });
+      } else if (condition.operator === "not_exists") {
+        must_not.push({
+          exists: {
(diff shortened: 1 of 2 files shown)


jacoblee93:
Looks good - could you add an integration test illustrating desired behavior?

LaurensTsestigh:
> Looks good - could you add an integration test illustrating desired behavior?

Sure, I've added/adapted the tests!

jacoblee93:
Thank you!

LaurensTsestigh:
> Thank you!

My pleasure, thank you for taking your time!

PR #7044, opened 2024-10-22
feat: add MeiliSearch as a VectorStore provider

SrilalS:
Adding MeiliSearch as a VectorStore provider to LangchainJS. This is written to be feature-compatible with the python implementation of Meilisearch.

Compatible with Meilisearch 1.6.0 or newer.
Node-Only Implementation.

- [ x ] Build Passes.
- [ x ] Documentation Done.

TODO:
- [ ] Index Management within the library.
- [ ] Able to run health checks.
- [ ] Improve Documentation to include more comprehensive use cases.


jacoblee93:
Thanks for your patience, having a look now!

SrilalS:
Thank you for checking it out and for the changes.
The library itself works fine in our codebase, but I noticed the `tsconfig` is different so this must be an issue with type rules. 
in the last commit have made it to follow the rules set in it. 

I can't build this repo locally first as it having an issue with Windows. I'll try with WSL.

manylov:
what is current status?

hntrl:
Hey @SrilalS, sorry to revive an old PR!

PR #7053, opened 2024-10-24
feat(community): Add delete and allow default row id in libsql 

cdierkens:
@jacoblee93 Sorry to tag directly, just noticed you were working here last!

## Changes

- used named parameters throughout
- in the `addVectors` function we've used the `vector(:embedding)` libsql function
  - see: https://docs.turso.tech/features/ai-and-embeddings#functions
-  returned every row id on insert
- ensure that we return a `String` wherever id's are returned per the `Document` type
- implemented the delete all and delete function
- allow creating and searching from table with an implicit ROWID column

## Could use Help

I'd suspect that we should use `OpenAIEmbeddings` throughout so we can run all tests against a local `libsql` db and a remote hosted db. Right now I used  `OllamaEmbeddings` in all the new test because the `OPENAI_API_KEY` in my .env was not picked up.  `OllamaEmbeddings` with the default model are `1024` dimensions while `OpenAIEmbeddings` are 1536. I don't see a create table statement in the existing test, so I'm not sure if it was created manually with 1536, or that's something that Turso does automatically.

## Other things I wouldn't mind adding

- filtering (can just replicate exact and in matches like other sql adapters)
- specifying the index in the constructor with a default

### DIFF
--- a/libs/langchain-community/src/vectorstores/libsql.ts
+++ b/libs/langchain-community/src/vectorstores/libsql.ts
@@ -1,7 +1,7 @@
-import type { Client } from "@libsql/client";
-import { VectorStore } from "@langchain/core/vectorstores";
-import type { EmbeddingsInterface } from "@langchain/core/embeddings";
 import { Document } from "@langchain/core/documents";
+import type { EmbeddingsInterface } from "@langchain/core/embeddings";
+import { VectorStore } from "@langchain/core/vectorstores";
+import type { Client, InStatement } from "@libsql/client";
 
 /**
  * Interface for LibSQLVectorStore configuration options.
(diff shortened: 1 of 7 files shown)


cdierkens:
Also, let me know if I should create issues to go along with this PR.

jacoblee93:
Hey @cdierkens! Issue would be nice but not necessary. 

I was going off the setup instructions here:

https://js.langchain.com/docs/integrations/vectorstores/libsql/#setup-the-table-and-index

I had some difficulty with deletions - there was some kind of shadow index I couldn't figure out how to clear

cdierkens:
> I was going off the setup instructions here:

@jacoblee93 Yeah I had issues with those instructions as well.  Following these helped alot:
https://docs.turso.tech/features/ai-and-embeddings

jacoblee93:
Oh I updated them in my previous PR and they should work now!

cdierkens:
@jacoblee93 What can I do to keep this moving. I'm available to help in any way.

jacoblee93:
You're good, will look today or tomorrow!

libs/langchain-community/src/vectorstores/libsql.ts
```diff
@@ -145,6 +140,26 @@ export class LibSQLVectorStore extends VectorStore {
...
 
+  /**
+   * Deletes vectors from the store.
+   * @param {Object} params - Delete parameters.
+   * @param {string[] | number[]} [params.ids] - The ids of the vectors to delete.
+   * @returns {Promise<void>}
+   */
+  async delete(params: { ids?: string[] | number[] }): Promise<void> {
+    if (!params.ids) {
+      await this.db.execute(`DELETE FROM ${this.table}`);
```
jacoblee93, line 151: We usually gate this on a `deleteAll` flag - will add

jacoblee93 (approved):
Thank you for your patience!

cdierkens:
> Thank you for your patience!

No worries!

PR #7057, opened 2024-10-24
 feat(cosmosdbnosql): Add Chat History Integration

aditishree1:
This PR adds Azure Cosmos DB NoSQL Chat History Integration with tests and samples.


libs/langchain-azure-cosmosdb/src/chat_histories.ts
```diff
@@ -0,0 +1,166 @@
...
+    }
+  }
+
+  private async initializeContainer(): Promise<void> {
+    const { database } = await this.client.databases.createIfNotExists({
+      id: this.databaseName,
+    });
+    const { container } = await database.containers.createIfNotExists({
+      id: this.containerName,
+      partitionKey: "/user_id",
```
sinedied, line 132: If this for Python compatibility?
As in JS, we usually name things with cameCase, ie `/userId`

libs/langchain-azure-cosmosdb/src/chat_histories.ts
```diff
@@ -0,0 +1,166 @@
...
+    } else {
+      return new CosmosClient({
+        endpoint,
+        aadCredentials: input.credentials ?? new DefaultAzureCredential(),
+        userAgentSuffix: USER_AGENT_SUFFIX,
+      } as CosmosClientOptions);
+    }
+  }
+
+  private async initializeContainer(): Promise<void> {
```
sinedied, line 126: This method is called every time messages needs to be accessed, so we don't want to make the calls to init the DB and container if not needed. Please use a boolean attribute so it's only called once.

libs/langchain-azure-cosmosdb/src/chat_histories.ts
```diff
@@ -0,0 +1,166 @@
...
+      messages,
+    });
+  }
+
+  async clear(): Promise<void> {
+    this.messageList = [];
+    await this.initializeContainer();
+    await this.container.item(this.sessionId, this.userId).delete();
+  }
+}
```
sinedied, line 166: Could we also have a method to clear all sessions for a user?

libs/langchain-azure-cosmosdb/src/chat_histories.ts
```diff
@@ -0,0 +1,166 @@
...
+
+/**
+ * Class for storing chat message history with Cosmos DB NoSQL. It extends the
+ * BaseListChatMessageHistory class and provides methods to get, add, and
+ * clear messages.
+ *
+ * @example
+ * ```typescript
+ * const chatHistory = new AzureCosmsosDBNoSQLChatMessageHistory({
+ *     sessionId: new ObjectId().toString(),
```
sinedied, line 38: ObjectId() is for MongoDB, a simple string is enough here for both params

libs/langchain-azure-cosmosdb/src/chat_histories.ts
```diff
@@ -0,0 +1,166 @@
...
+const USER_AGENT_SUFFIX = "langchainjs-cdbnosql-chathistory-javascript";
+const DEFAULT_DATABASE_NAME = "chatHistoryDB";
+const DEFAULT_CONTAINER_NAME = "chatHistoryContainer";
+
+/**
+ * Type for the input to the `AzureCosmosDBNoSQLChatMessageHistory` constructor.
+ */
+export interface AzureCosmosDBNoSQLChatMessageHistoryInput {
+  sessionId: string;
+  userId: string;
```
sinedied, line 20: Could we make the userId param optional here, with a fallback to a default one? (like `anonymous` or whatever)?
This would allow the integration to work in context where no user auth is available. Not a big use case I agree, but other history integrations don't force a userId requirement.

sinedied (changes requested):

libs/langchain-azure-cosmosdb/src/chat_histories.ts
aditishree1: I kept that same as python. But, not sure if python compatibility is required here.

libs/langchain-azure-cosmosdb/src/chat_histories.ts
aditishree1: I'm adding "anonymous" as default for now. But should it be different for different users (e.g. we could add some random digits after "user" or "anonymous")? Considering we are adding method for clearing all sessions for a user. And if someone would pass "anonymous" in this method, it would clear sessions for other users too.

jacoblee93:
Please ping when you'd like a final review!

aditishree1:
Hi @jacoblee93, its ready for review.

examples/src/memory/azure_cosmosdb_nosql.ts
```diff
@@ -0,0 +1,47 @@
...
+const memory = new BufferMemory({
+  chatHistory: new AzureCosmsosDBNoSQLChatMessageHistory(input),
+});
+
+const model = new ChatOpenAI({
+  model: "gpt-3.5-turbo",
+  temperature: 0,
+});
+
+const chain = new ConversationChain({ llm: model, memory });
```
jacoblee93, line 24: We generally don't suggest using this - prefer `RunnableWithMessageHistory` instead:

https://js.langchain.com/docs/versions/migrating_memory/#runnablewithmessagehistory-and-basechatmessagehistory

Note that we are suggesting LangGraph.js for apps that require memory in general

jacoblee93 (commented):
Seems fine overall, new docs should not use the deprecated `ConversationChain` though. Can you switch the example?

aditishree1:
> Seems fine overall, new docs should not use the deprecated `ConversationChain` though. Can you switch the example?

Hi, I have updated the example.

aditishree1:
Hey @jacoblee93 let me know if any more changes are required. Thanks.

jacoblee93 (approved):
Thank you!

PR #7069, opened 2024-10-25
fix(qdrant): Support custom payload also when using fromExistingColle…

JustDoItSascha:
…ction (#2756)

Fixes # (issue)

### DIFF
--- a/libs/langchain-qdrant/src/vectorstores.ts
+++ b/libs/langchain-qdrant/src/vectorstores.ts
@@ -25,15 +25,16 @@ export interface QdrantLibArgs {
   apiKey?: string;
   collectionName?: string;
   collectionConfig?: QdrantSchemas["CreateCollection"];
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  customPayload?: Record<string, any>[];
+  customPayload?: QdrantCustomPayload;
   contentPayloadKey?: string;
   metadataPayloadKey?: string;
 }
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type QdrantCustomPayload = Record<string, any>;
+
 export type QdrantAddDocumentOptions = {
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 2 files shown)


libs/langchain-qdrant/src/vectorstores.ts
```diff
@@ -113,18 +118,18 @@ export class QdrantVectorStore extends VectorStore {
...
    * vectors to the database.
    * @param documents Array of `Document` instances to be added to the Qdrant database.
-   * @param documentOptions Optional `QdrantAddDocumentOptions` which has a list of JSON objects for extra querying
+   * @param customPayloads Optional `QdrantCustomPayload` array which has a list of JSON objects for extra querying.
    * @returns Promise that resolves when the documents have been added to the database.
    */
   async addDocuments(
     documents: Document[],
-    documentOptions?: QdrantAddDocumentOptions
+    customPayloads?: QdrantCustomPayload[],
```
jacoblee93, line 126: This is breaking isn't it?

Would prefer to keep the wrapping

JustDoItSascha: Sure, feel free to keep it 🫣 I understand the urge, but in my opinion I'm always for the correct way instead producing the legacy code of tomorrow. It's yet already complicated enough with the customPayloads on document level and the customPayload on dbConfig level.

But I understand that you don't want to have breaking changes, even a simple one like this.

jacoblee93: Yeah it's a fine line - I'd err towards keeping it for now with a note to change it in the next minor bump.

JustDoItSascha:
Sounds reasonable :)

Jacob Lee ***@***.***> schrieb am Di., 29. Okt. 2024, 22:04:

christian-bromann:
@JustDoItSascha sorry for the delay on this, we started to pick up the stale issues one by one. Mind rebasing the PR on top of current `main`?

hntrl:
Closing this due to inactivity

PR #7096, opened 2024-10-28
fix(anthropic): Fix multipart tool message

gingerhendrix:
Fixes handling multi-part (e.g. image and text) `ToolMessage`s.

For Anthropic these need to be handled as `HumanMessage`s with a `type: "tool_result"` MessageContentComplex, which itself can have a multipart content part (formatted according to anthropic's requirements).

https://x.com/gingerhendrixai


bracesproul (approved):
looks good granted @jacoblee93 confirms it passes integration tests!

jacoblee93:
Thank you!

Issue #7013, opened 2024-10-17
Missing ID in VectorStores


kwei-zhang:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain.js documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain.js rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

Example code:
```
import { Chroma } from "@langchain/community/vectorstores/chroma";
import { OpenAIEmbeddings } from "@langchain/openai";

const embeddings = new OpenAIEmbeddings({
  model: "text-embedding-3-small",
});

const vectorStore = new Chroma(embeddings, {
  collectionName: "a-test-collection",
  collectionMetadata: {
    "hnsw:space": "cosine",
  },
});

const document1 = {
  pageContent: "The powerhouse of the cell is the mitochondria",
  metadata: { source: "https://example.com" },
};

const documents = [document1];
await vectorStore.addDocuments(documents, { ids: ["1"] });
const similaritySearchResults = await vectorStore.similaritySearch("pwoerhouse of the cell");

console.log(similaritySearchResults);
```

### Error Message and Stack Trace (if applicable)

```
[
  Document {
    pageContent: 'The powerhouse of the cell is the mitochondria',
    metadata: { source: 'https://example.com' },
    id: undefined
  }
]
```

### Description

The same issue as shown in [Python](https://github.com/langchain-ai/langchain/issues/26860).
There is already pull request [here](https://github.com/langchain-ai/langchainjs/pull/6963) that fixes one of the vector. But the issue occurs in other vectorstores.

Example code fix in Chroma db
`libs/langchain-community/src/vectorstores/chroma.ts line:351`:
```
            ...
            results.push([
                new Document({
                    pageContent: firstDocuments?.[i] ?? "",
                    metadata,
                }),
                firstDistances[i],
            ]);
```
to
```
            results.push([
                new Document({
                    pageContent: firstDocuments?.[i] ?? "",
                    metadata,
                    id: firstIds[i],
                }),
                firstDistances[i],
            ]);
```

If it's possible we would like to perform same solution to other vectorstores.

### System Info

```
dependencies:
@langchain/openai: >=0.1.0 <0.4.0        jsonpointer: ^5.0.1                      uuid: ^10.0.0                            
[...]

PR #7120, opened 2024-10-30
fix(community): Add id in chroma search result

KevinZJN:
Fixes # [7013](https://github.com/langchain-ai/langchainjs/issues/7013)


jacoblee93:
Thank you!

PR #7133, opened 2024-10-31
[docs] Basic tutorial: Vector stores and retrievers

SkSirius:
As per discussion: #7132 

- Added new tutorial to langchainjs documentation for vector stores and retrievers (same as python)


SkSirius:
@jacoblee93 How can I get access to Vercel deployment details? I can't fix it unless I can have a glance at the error

hntrl:
Going to close this PR due to inactivity. We're also collecting our docs efforts on https://github.com/langchain-ai/docs, so if we want to pick this up again it will happen there. Thanks!

Issue #6994, opened 2024-10-16
Error encountered when using LlamaCpp class


chantal-rose:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain.js documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain.js rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```typescript 
import { LlamaCpp } from "@langchain/community/llms/llama_cpp";

// downloaded from https://huggingface.co/mradermacher/Meta-Llama-3.1-8B-Instruct-GGUF/resolve/main/Meta-Llama-3.1-8B-Instruct.Q4_K_M.gguf?download=true
const llamaPath = "/path/to/your/model.gguf"; 

const model = new LlamaCpp({ modelPath: llamaPath });

// Use the model as needed

```

### Error Message and Stack Trace (if applicable)

file:///Users/cdgamarose/projects/poc/node-llama-cpp-project/node_modules/node-llama-cpp/dist/evaluator/LlamaModel/LlamaModel.js:42
    constructor({ modelPath, gpuLayers, vocabOnly = false, useMmap, useMlock, checkTensors, onLoadProgress, loadSignal, metadataOverrides }, { _llama, _fileInfo, _fileInsights, _defaultContextFlashAttentionOptionEnabled, _defaultContextFlashAttention, _flashAttentionSupported }) {
                                                                                                                                               ^

TypeError: Cannot destructure property '_llama' of 'undefined' as it is undefined.
    at new LlamaModel (file:///Users/cdgamarose/projects/poc/node-llama-cpp-project/node_modules/node-llama-cpp/dist/evaluator/LlamaModel/LlamaModel.js:42:144)
    at createLlamaModel (file:///Users/cdgamarose/projects/poc/node-llama-cpp-project/node_modules/@langchain/community/dist/utils/llama_cpp.js:13:12)
    at new LlamaCpp (file:///Users/cdgamarose/projects/poc/node-llama-cpp-project/node_modules/@langchain/community/dist/llms/llama_cpp.js:87:23)
    at file:///Users/cdgamarose/projects/poc/node-llama-cpp-project/langchain.js:4:15

Node.js v20.18.0

### Description

I am trying to create a desktop RAG app using langchain.js that uses node llama cpp.

### System Info

Node.js v20.18.0
[...]

PR #7135, opened 2024-11-01
fix(community) : Upgrade node-llama-cpp to be compatible with version 3 

rd4cake:
[Work in Progress]

Draft PR to address the following issue: https://github.com/langchain-ai/langchainjs/issues/6994
Using asynchronous function to load models from node-llama-cpp breaks adherence to the default way of instantiating components ( chatmodels, embeddings, llms, etc ).

Here is an example:

before with node-llama-cpp v2
```typescript
const llamaPath = "/path/to/your/model.gguf"; 

const model = new LlamaCpp({ modelPath: llamaPath });
```

after with node-llama-cpp v3
```typescript
const llamaPath = "/path/to/your/model.gguf"; 

const model = await LlamaCpp.llmInit({ modelPath: llamaPath });
```
Question for the maintainers:
Our implementation changes the way LlamaCpp model is instantiated. Is it this fine moving forward?


Jacky3003:
@jacoblee93 was wondering if it was possible for you to take a look at what we have so far for our implementation, it would be greatly appreciated.

examples/src/embeddings/llama_cpp_basic.ts
```diff
@@ -2,7 +2,7 @@ import { LlamaCppEmbeddings } from "@langchain/community/embeddings/llama_cpp";
 
 const llamaPath = "/Replace/with/path/to/your/model/gguf-llama2-q4_0.bin";
 
-const embeddings = new LlamaCppEmbeddings({
+const embeddings = await LlamaCppEmbeddings.embeddingsInit({
```
jacoblee93, line 5: Would prefer these all to be named `.initialize` for consistency

libs/langchain-community/src/chat_models/llama_cpp.ts
```diff
@@ -87,20 +91,35 @@ export class ChatLlamaCpp extends SimpleChatModel<LlamaCppCallOptions> {
     return "ChatLlamaCpp";
   }
 
-  constructor(inputs: LlamaCppInputs) {
+  private constructor(inputs: LlamaCppInputs) {
```
jacoblee93, line 94: Does this make inheritance impossible? Should it be `protected` instead?

jacoblee93: And is there no sync API anymore?

Jacky3003: Unfortunately, there isn't a synchronous API anymore so the current implementation is the best workaround right now. This is both outlined in the [docs](https://node-llama-cpp.withcat.ai/guide/chat-session) and the [source code](https://github.com/withcatai/node-llama-cpp/blob/master/src/bindings/Llama.ts#L314). And yea it does make inheritance impossible, so will change to protected, as long as the person using the library is unable to access the constructor directly, all should be fine.

jacoblee93 (approved):
Seems to run for me, thank you!

jacoblee93:
CC @nigel-daniels

jacoblee93:
Just one merge to main should be fine! If you keep doing it, it requires CI to run again.

PR #7137, opened 2024-11-01
docs: Update pdf_qa.ipynb

apuravchauhan:
Refer the correct variable name

In the tutorial the, llm is instantiated under the variable name **`model`**. But while creating the createStuffDocumentsChain in the following steps, the variable name referred is `llm`.

The issue comes in because the variable llm exists in python documentation and the same is copied here in js documentation
Fixes # (issue)


jacoblee93:
Thank you!

PR #7141, opened 2024-11-02
docs(core): VectorStore and Retriever: types, interfaces, classes

SkSirius:
Discussion #7142 

- Added TypeDoc comments for retriever and vector store types, classes, and interfaces.


langchain-core/src/vectorstores.ts
```diff
@@ -337,6 +769,41 @@ export abstract class VectorStore
...
+   * @returns
+   *    - A configured `VectorStoreRetriever` instance based on the provided parameters.
+   *
+   * @example
+   * // Basic usage with a `k` value:
+   * const retriever = asRetriever(5);
+   *
+   * @example
+   * // Usage with a configuration object:
+   * const retriever = asRetriever({
```
jacoblee93, line 798: Example is wrong - this is an instance method

Also should wrap in markdown blocks for syntax highlighting in API refs

jacoblee93 (commented):
Thanks for this! Small comment about some of the code examples

SkSirius:
@jacoblee93 Updated, good catch, thanks.

jacoblee93:
Are you using an LLM for some of this? Would love to learn more about your process - I did one a long time ago:

https://github.com/jacoblee93/auto-docstrings

But never kept it maintained

SkSirius:
Oh, I'll have a look, thanks! LLM - yes, but not to autogenerate it. I usually write it on my own and use llm to validate and polish up my comments as I'm not a native speaker. And to come up with examples (when I'm lazy).

Issue #6734, opened 2024-09-11
StructuredOutputParser can't handle nested triple backticks


hnykda:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain.js documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain.js rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```typescript
import { StructuredOutputParser } from 'langchain/output_parsers';
import { z } from 'zod';

async function main() {
  const c =
    '{  "revisedOutput": "long markdown text with example ```javascript\\nfunction(){}``` here"}';

  const parser = StructuredOutputParser.fromZodSchema(
    z.object({
      revisedOutput: z.string(),
    }),
  );
  // the following line will throw an error
  const response = await parser.parse(c);
  console.log(response);
}

main();
```

### Error Message and Stack Trace (if applicable)
```
/Users/dan/work/varuna/code/varuna/node_modules/langchain/dist/output_parsers/structured.cjs:83
            throw new output_parsers_1.OutputParserException(`Failed to parse. Text: "${text}". Error: ${e}`, text);
                  ^

OutputParserException [Error]: Failed to parse. Text: "{  "revisedOutput": "long markdown text with example ```javascript\nfunction(){}``` here"}". Error: SyntaxError: Unexpected token 'j', "javascript"... is not valid JSON
    at StructuredOutputParser.parse (/Users/dan/work/varuna/code/varuna/node_modules/langchain/dist/output_parsers/structured.cjs:83:19)
    at main (/Users/dan/work/varuna/code/varuna/app/dev/parse.tsx:15:33)
    at <anonymous> (/Users/dan/work/varuna/code/varuna/app/dev/parse.tsx:19:1)
    at Object.<anonymous> (/Users/dan/work/varuna/code/varuna/app/dev/parse.tsx:19:6)
    at Module._compile (node:internal/modules/cjs/loader:1546:14)
    at Object.transformer (/Users/dan/work/varuna/code/varuna/node_modules/tsx/dist/register-C1urN2EO.cjs:2:1122)
    at Module.load (node:internal/modules/cjs/loader:1317:32)
    at Module._load (node:internal/modules/cjs/loader:1127:12)
    at TracingChannel.traceSync (node:diagnostics_channel:315:14)
[...]

PR #7144, opened 2024-11-03
fix: structured output parser

anadi45:
Fixes #([6734](https://github.com/langchain-ai/langchainjs/issues/6734))

1. Capture JSON object using outer curly brackets.
2. Fix existing integration tests

### DIFF
--- a/langchain/src/output_parsers/structured.ts
+++ b/langchain/src/output_parsers/structured.ts
@@ -101,9 +101,9 @@ ${JSON.stringify(zodToJsonSchema(this.schema))}
    */
   async parse(text: string): Promise<z.infer<T>> {
     try {
-      const jsonRegex = /```(?:json)?\n([\s\S]*?)\n```/;
+      const jsonRegex = /{[\s\S]*}/;
       const match = text.match(jsonRegex);
-      const jsonString = match?.[1]?.trim() ?? text.trim();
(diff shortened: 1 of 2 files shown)


jacoblee93:
Hey @anadi45, thanks for this!

Am a bit nervous about swapping this out wholesale since LLMs could perhaps output curly braces in other contexts and I wouldn't want to break anyone depending on the old behavior - what is the issue you are encountering?

anadi45:
Hi @jacoblee93, thank you so much for the review! I'm new to open source and wanted to contribute to LangChain since we use it at our organization. I chose this [issue](https://github.com/langchain-ai/langchainjs/issues/6734) as a starting point where the parser currently breaks if there are nested triple backticks. 

I agree that changing the parsing logic entirely could be risky. To maintain backward compatibility, I could incorporate my logic within a catch block, and only throw an error if both approaches fail. I’ll also explore alternative ways to handle this issue. 

If you have any other suggestions, I’d be happy to incorporate them! Regardless, I'll add more test cases to cover these scenarios.

jacoblee93:
Hey @anadi45, sounds good - please ping again when ready.

anadi45:
@jacoblee93 should I close this? Looks like https://github.com/langchain-ai/langchainjs/pull/7199 already solved this issue.

PR #7151, opened 2024-11-05
fix(community): Fix logic issue

clemenspeters:
The ?? operator is the nullish coalescing operator, which only checks for null or undefined. However, the expression before it `getEnvironmentVariable("LANGCHAIN_VERBOSE") === "true"` is a boolean expression and will never be null or undefined.

@PetersClemens

Fixes # (issue)

### DIFF
--- a/libs/langchain-community/src/vectorstores/pgvector.ts
+++ b/libs/langchain-community/src/vectorstores/pgvector.ts
@@ -270,8 +270,6 @@ export class PGVectorStore extends VectorStore {
     } else {
       this._verbose = config.verbose;
     }
-
-
   }
 
   get computedTableName() {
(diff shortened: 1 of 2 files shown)


jacoblee93:
Thank you!

libs/langchain-community/src/vectorstores/pgvector.ts
```diff
@@ -527,9 +535,8 @@ export class PGVectorStore extends VectorStore {
 
     const queryString = `
       DELETE FROM ${this.computedTableName}
-      WHERE ${collectionId ? "collection_id = $2 AND " : ""}${
-      this.idColumnName
-    } = ANY($1::uuid[])
+      WHERE ${collectionId ? "collection_id = $2 AND " : ""}${this.idColumnName
```
jacoblee93, line 538: Can you run `yarn format`? Seems like it's blocking CI

I don't have permission to merge to your branch

clemenspeters: Sorry about that! I ran it on the wrong branch before 🙈

PR #7167, opened 2024-11-06
pinecone[patch]: Update `@pinecone/pinecone-database` version to resolve type errors

nick-w-nick:
Due to a recent `@pinecone/pinecone-database` release (`v4.0.0`), there is a type collision between `@langchain/pinecone` and `@pinecone/pinecone-database`, since `@langchain/pinecone` still uses `v3.0.0`.

Here is the issue in the `pinecone-ts-client` repository which mentions this bug: https://github.com/pinecone-io/pinecone-ts-client/issues/309

Updating the `@langchain/pinecone` module to use the `v4.0.0` version of `@pinecone/pinecone-database` fixes the issue.


jacoblee93:
If this isn't breaking - can we make this `^3.0.0 || ^4.0.0`?

nick-w-nick:
@jacoblee93 That's a great idea! Just made the update, let me know if that works for you.

jacoblee93:
Thank you!

PR #7169, opened 2024-11-07
feat(google-genai): Context Caching

### LINKED ISSUE #5841: google-common google-genai [feature]: Context Caching
### Privileged issue

### Issue Content

Gemini now allows a developer to create a context cache with the system instructions, contents, tools, and model information already set, and then reference this context as part of a standard query. It must be explicitly cached (ie - it is not automatic as part of a request or reply) and a cache expiration can be set (and later changed).

It does not appear to be supported in Vertex AI at this time.

Open issues:
* Best paradigm to add to cache or integrate with LangChain history system
* Best paradigm to reference

References:
* AI Studio / genai: https://ai.google.dev/gemini-api/docs/caching?lang=node
* REST: https://ai.google.dev/api/rest/v1beta/cachedContents#CachedContent
* LangChain: https://github.com/langchain-ai/langchain/issues/23259

KevinZJN:
Fixes #5841 

Implemented conetxt caching feature for google-genai, we now allow user to caching a file and create gen-ai model base on the cached content


KevinZJN:
Hi @jacoblee93, we've implemented the foundational structure for context caching with files. Could you take a look and let us know if the code aligns with our intended design? Thank you!

libs/langchain-google-genai/src/tests/context_caching.int.test.ts
```diff
@@ -0,0 +1,79 @@
...
+        ],
+      },
+    ],
+    ttlSeconds: 300,
+  });
+
+  model.enableCachedContent(cachedContent);
+}, 10 * 60 * 1000); // Set timeout to 10 minutes to upload file
+
+test("Test Google AI", async () => {
```
jacoblee93, line 70: I think this should be part of the same test? Otherwise it will fail in isolation

jacoblee93:
Looks reasonable to me - @afirstenberg can you have a look? It looks similar to some work you've done on Vertex.

We will also want to write up some docs!

libs/langchain-google-genai/src/tests/context_caching.int.test.ts
chaunguyenm: Thank you for taking the time to review this. Regarding your comment, could you please clarify how to reproduce the test failure? My understanding is that the `beforeAll` hook, which is handling the setup and file uploading, runs before any tests, so I'm not entirely sure what you mean by suggesting that they should be part of the same test.

chaunguyenm:
> Looks reasonable to me - @afirstenberg can you have a look? It looks similar to some work you've done on Vertex.
> 
> We will also want to write up some docs!

We will be working on the docs now, but please keep us updated if there's any suggestion on the design or the tests. Thank you both!

chaunguyenm:
@jacoblee93 @afirstenberg We currently add a wrapper around `GoogleAIFileManager` and `GoogleAICacheManager` to support context caching, but the wrapper doesn't provide additional functionalities. This is because we are not sure if we can rely on users installing and using `google/generative-ai/server` package directly. Do you have a suggestion on how we can structure this? Thanks a lot.

libs/langchain-google-genai/src/tests/context_caching.int.test.ts
```diff
@@ -0,0 +1,79 @@
...
+let fileResult: UploadFileResponse;
+
+beforeAll(async () => {
+  // Download video file and save in src/tests/data
+  // curl -O https://storage.googleapis.com/generativeai-downloads/data/Sherlock_Jr_FullMovie.mp4
+  const displayName = "Sherlock Jr. video";
+
+  const filename = fileURLToPath(import.meta.url);
+  const dirname = path.dirname(filename);
+  const pathToVideoFile = path.join(dirname, "/data/Sherlock_Jr_FullMovie.mp4");
```
jacoblee93, line 22: Can we use a smaller file for this test? Maybe one of the files that already exist in the repo?

libs/langchain-google-genai/src/tests/context_caching.int.test.ts
jacoblee93: Ah sorry

jacoblee93:
Thank you! Removed the intermediate class you added in favor of just the raw managers

It wasn't exported anyway

jacoblee93 (approved):

jacoblee93:
Also renamed to `useCachedContent`

Thanks again, will go live shortly!

PR #7184, opened 2024-11-11
feat(community): Adds an HTML loader for URLS

### LINKED ISSUE #2467: Possible mistake in the documentation | Document Transformers | Mozilla Readability
Hi folks, I've been playing with the framework, and while following the [documentation on retrievers](https://js.langchain.com/docs/modules/data_connection/document_transformers/integrations/mozilla_readability) I found the following case.

You mention in the docs to first load the page with `cheerio`:

```typescript
const loader = new CheerioWebBaseLoader(
  "https://news.ycombinator.com/item?id=34817881"
);

const splitter = RecursiveCharacterTextSplitter.fromLanguage("html");
const transformer = new MozillaReadabilityTransformer();

const sequence = splitter.pipe(transformer);

const newDocuments = await sequence.invoke(docs);

console.log(newDocuments);
```

However, your implementation of the `CheerioWebBaseLoader` takes the page content and turns it into text (therefore stripping HTML tags)

```typescript
 async load() {
        const $ = await this.scrape();
        const text = $(this.selector).text();
        const metadata = { source: this.webPath };
        return [new Document({ pageContent: text, metadata })];
    }
```

I wonder if this is indeed what you've intended to do and from the code in the documentation, I figured it might not be.

You are using `const transformer = new MozillaReadabilityTransformer();` to later retrieve the page's content, but Mozilla readability requires access to the HTML structure (e.g. `article` tag). However, given that the previous step stripped all HTML the reader won't be able to function properly.

I found that writing a custom loader that extends yours helps a bit. Here is the class I use:

```typescript
class CustomCheerioWebBaseLoader extends CheerioWebBaseLoader {
  /**
   * Extracts the text content from the loaded document using the selector
   * and creates a Document instance with the extracted text and metadata.
   * It returns an array of Document instances.
   * @returns A Promise that resolves to an array of Document instances.
   */
  async loadHtml(): Promise<Document[]> {
    const $ = await this.scrape();
    const html = $(this.selector).html() || "";
    const metadata = { source: this.webPath };
    return [new Document({ pageContent: html, metadata })];
  }
}
```

However, I think the code in the documentation will still produce suboptimal results. The code uses `const splitter = RecursiveCharacterTextSplitter.fromLanguage("html");` and the order of processing the document is:
1. Split
2. Process with readability

But when the split happens first it leaves a lot of noise (HTML data irrelevant to the actual loaded article content). Also, it means that readability will try to analyze chunks of HTML and I found it doesn't handle this well.

Hope those thoughts will be useful, cheers!

philnash:
I recently discovered the same issue that was reported in #2467. The documentation and examples suggest chaining the `CheerioWebBaseLoader` with `MozillaReadabilityTransformer` to load and transform HTML documents. However, the `CheerioWebBaseLoader` uses [Cheerio's `text` method to extract the text](https://github.com/langchain-ai/langchainjs/blob/05e5813715150cd69d9e384924818562e3b7c1fa/libs/langchain-community/src/document_loaders/web/cheerio.ts#L144C34-L144C40) from the HTML document, or provided selector. This is great if you want the text, but the MozillaReadabilityTransformer needs to act on HTML.

I have added an `HTMLWebBaseLoader` that simply uses `fetch` to get an HTML document and returns the full HTML content.

I've also updated the `MozillaReadabilityTransformer` example to use the `HTMLWebBaseLoader` instead of the `CheerioWebBaseLoader`.

Fixes #2467

I'm [@philnash](https://twitter.com/philnash) if you're still doing shout outs :)


jacoblee93 (approved):
Thanks for this! Looks great.

libs/langchain-community/src/document_loaders/web/cheerio.ts
```diff
@@ -7,36 +7,21 @@ import type {
...
+import { WebBaseLoaderParams, WebBaseLoader } from "./web_base_loader.js";
 
 /**
  * Represents the parameters for configuring the CheerioWebBaseLoader. It
- * extends the AsyncCallerParams interface and adds additional parameters
- * specific to web-based loaders.
+ * extends the WebBaseLoaderParams interface and adds additional parameters
+ * specific to loading with Cheerio.
  */
-export interface WebBaseLoaderParams extends AsyncCallerParams {
```
jacoblee93, line 20: Actually one small thing - this is exported, so on the off-chance someone is importing and using it, would prefer to keep the old export and deprecate here.

jacoblee93:
Sorry, noticed one small thing. Can you also run `yarn format`?

libs/langchain-community/src/document_loaders/web/cheerio.ts
philnash: Ah yep, fair enough. Will sort that out now. :+1:

philnash:
Force pushed to fix the merge conflict.

libs/langchain-community/src/document_loaders/web/cheerio.ts
```diff
@@ -5,38 +5,27 @@ import type {
...
- * specific to web-based loaders.
+ * @deprecated Either import the CheerioWebBaseLoaderParams from @langchain/community/document_loaders/web/cheerio
+ * or use the WebBaseLoaderParams from @langchain/community/document_loaders/web/html.
  */
-export interface WebBaseLoaderParams extends AsyncCallerParams {
-  /**
-   * The timeout in milliseconds for the fetch request. Defaults to 10s.
-   */
-  timeout?: number;
+export { WebBaseLoaderParams };
```
jacoblee93, line 16: This is still technically breaking since we're removing `selector?: SelectorType;` from it but I think this is fine

philnash:
@jacoblee93 can we trigger the jobs again to see if this is looking good? Thanks!

jacoblee93 (approved):
Thank you! Sorry for the delay.

PR #7192, opened 2024-11-12
fix[community]: Fixed missing tool_choice in call options of chat IBM

FilipZmijewski:
Fixed missing tool_choice in call options of chat model.
Was not aware of this issue until actually looking into langgraph implementation, maybe there could be an option to avoid such problems in the future?

### DIFF
--- a/libs/langchain-community/src/chat_models/ibm.ts
+++ b/libs/langchain-community/src/chat_models/ibm.ts
@@ -41,6 +41,7 @@ import {
   TextChatResultChoice,
   TextChatResultMessage,
   TextChatToolCall,
+  TextChatToolChoiceTool,
   TextChatUsage,
 } from "@ibm-cloud/watsonx-ai/dist/watsonx-ai-ml/vml_v1.js";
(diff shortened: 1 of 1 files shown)


FilipZmijewski:
@jacoblee93 This is very simple fix if you could have a look at it, please. Did not manage to see the issue until updating package in langgraph based project. And receiving and error that types on ```llm``` do not match. Thank you.
Screenshot 2024-11-12 at 10 19 36
Fixed in this PR.

libs/langchain-community/src/chat_models/ibm.ts
```diff
@@ -86,6 +87,7 @@ export interface WatsonxCallOptionsChat
   extends Omit<BaseLanguageModelCallOptions, "stop">,
```
jacoblee93, line 87: This should extend `BaseChatModelCallOptions` to prevent this in the future

jacoblee93:
Thanks! See comment but happy to merge and cut for now

FilipZmijewski:
yes, checked locally with langgraph, will fix accordingly to your comment in the next PR then

PR #7204, opened 2024-11-14
chore(langchain): add throw err for JsonOutputFunctionsParser.parse

Makoq:
When use the method of `JsonOutputFunctionsParser.parse(text: string)` with  input text in non-strict mode JSON string, throw description of error information to notice the developers.

It's relatively common in developing and debugging that input content is non-strict mode JSON.


jacoblee93 (approved):
Thank you!

PR #7199, opened 2024-11-13
fix(langchain): Fix structured parser with triple backticks, adds tests

johnguirgis:
Fixes #6734

Corrects an issue where the output parser naïvely attempts to parse contents of triple backticks as JSON.

1. Output parser falls back to parse the entire string when extracting from within triple backticks fails
2. Adds tests to account for different backtick positions in JSONs

PR #7232, opened 2024-11-19
fix(langchain): Fix structured parser error handling

burivuhster:
Previously, StructuredOutputParser would sometimes throw JSON.parse errors instead of the original Zod schema validation errors. 

This PR modifies the fallback mechanism to only trigger for JSON parsing failures, while preserving Zod's schema validation errors.

Relevant links:
- #7199

### DIFF
--- a/langchain/src/output_parsers/structured.ts
+++ b/langchain/src/output_parsers/structured.ts
@@ -104,15 +104,23 @@ ${JSON.stringify(zodToJsonSchema(this.schema))}
       const json = text.includes("```")
         ? text.trim().split(/```(?:json)?/)[1]
         : text.trim();
+
       return await this.schema.parseAsync(JSON.parse(json));
     } catch (e) {
-      try {
-        return await this.schema.parseAsync(JSON.parse(text.trim()));
-      } catch (e2) {
-        throw new OutputParserException(
-          `Failed to parse. Text: "${text}". Error: ${e2}`,
-          text
-        );
+      // eslint-disable-next-line no-instanceof/no-instanceof
+      if (e instanceof SyntaxError && e.message.includes("JSON")) {
+        // In case the error is JSON.parse related, try to parse the text directly
+        try {
+          return await this.schema.parseAsync(JSON.parse(text.trim()));
+        } catch (e2) {
+          throw new OutputParserException(
+            `Failed to parse. Text: "${text}". Error: ${e2}`,
+            text
+          );
+        }
+      } else {
+        // Otherwise, throw the original error (e.g. ZodError)
+        throw e;
       }
     }
   }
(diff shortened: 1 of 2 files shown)


langchain/src/output_parsers/structured.ts
```diff
@@ -104,15 +104,23 @@ ${JSON.stringify(zodToJsonSchema(this.schema))}
...
     } catch (e) {
-      try {
-        return await this.schema.parseAsync(JSON.parse(text.trim()));
-      } catch (e2) {
-        throw new OutputParserException(
-          `Failed to parse. Text: "${text}". Error: ${e2}`,
-          text
-        );
+      // eslint-disable-next-line no-instanceof/no-instanceof
+      if (e instanceof SyntaxError && e.message.includes("JSON")) {
```
bracesproul, line 111: Can't use `instanceof` as it's not available in all JS environments.

langchain/src/output_parsers/structured.ts
```diff
@@ -104,15 +104,23 @@ ${JSON.stringify(zodToJsonSchema(this.schema))}
...
+        // In case the error is JSON.parse related, try to parse the text directly
+        try {
+          return await this.schema.parseAsync(JSON.parse(text.trim()));
+        } catch (e2) {
+          throw new OutputParserException(
+            `Failed to parse. Text: "${text}". Error: ${e2}`,
```
bracesproul, line 117: If `e2` is an object, it will not be rendered properly in this message. Please add a check beforehand which stringifies the error if it's an object

This was an issue before, but we should fix now.

bracesproul (changes requested):
thanks for doing this! I have two small comments, but other than that, lgtm!

Please re-request my review once you've implemented them

langchain/src/output_parsers/structured.ts
burivuhster: Do you think it would be better to change this condition to something like `if (e && (e as Error).name !== "ZodError")`?

jacoblee93: Unfortunately minifiers can also mess with that 😬

It looks like `SyntaxError` is a native object though?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SyntaxError

So probably ok to do the instanceof check

langchain/src/output_parsers/structured.ts
burivuhster: Hey, it took a while 😅 Can you please take a look again?

burivuhster:
@benjamincburns can you please take a look?

hntrl:
I believe this was fixed with the recent zod 4 changes. @burivuhster could you open a new issue if you're still encountering this with the latest package versions?

PR #7233, opened 2024-11-19
await the generation in the no cache branch

### LINKED ISSUE #7172: Gemini JSON Mode Not Working
### Checked other resources

### Example Code

```typescript
      model: "gemini-1.5-flash-001",
      apiKey: env.GOOGLE_GEMINI_API_KEY,
    });

const format = z.object({
  joke: z.string(),
});

const modelWithParser = this.model.withStructuredOutput(
  format,
  {
    method: "jsonMode",
  },
);

const result = await modelWithParser.invoke("Tell me a joke");
```

Error: ChatGoogleGenerativeAI only supports "functionCalling" as a method.
    at ChatGoogleGenerativeAI.withStructuredOutput

### Description

Despite this closed ticket that seems to mention support for Gemini structured outputs via JSON mode, I still get that error. It seems maybe PR that closed it just implemented support for a general JSON response type?

### System Info

@langchain/openai: >=0.1.0 <0.4.0        openapi-types: ^12.1.3                   
@langchain/textsplitters: >=0.0.0 <0.2.0 p-retry: 4                               
js-tiktoken: ^1.0.12                     uuid: ^10.0.0                            
js-yaml: ^4.1.0                          yaml: ^2.2.1                             
jsonpointer: ^5.0.1                      zod-to-json-schema: ^3.22.3              
langsmith: ^0.2.0                        zod: ^3.22.4

kevinnayar:
Fixes #7172

### DIFF
--- a/langchain-core/src/language_models/chat_models.ts
+++ b/langchain-core/src/language_models/chat_models.ts
@@ -629,7 +629,7 @@ export abstract class BaseChatModel<
     runnableConfig.callbacks = runnableConfig.callbacks ?? callbacks;
 
     if (!this.cache) {
-      return this._generateUncached(baseMessages, callOptions, runnableConfig);
+      return await this._generateUncached(baseMessages, callOptions, runnableConfig);
     }
 
     const { cache } = this;


kevinnayar:
Oh I'm sorry, I attached the wrong issue but I cannot seem to locate the one I was viewing earlier and am trying to address. The attached issue is related to `ChatGoogleGenerativeAI` and `withStructuredOutput`, but the one I was referencing was related to `ChatVertexAI`. I'll keep digging for it.

Without awaiting the return on this method, the response when calling `model.withStructuredOutput(...)` is consistently the following error:
```
TypeError: Cannot read properties of undefined (reading 'text')
```

In this call stack sequence `invoke` -> `generatePrompt` -> `generate`, the last method doesn't return a value which results in this error. I believe this is because `generate`'s return type is a promise/non-promise union and therefore, it should be awaited.
```
generate(...): Promise<{
    generations: never[][];
    llmOutput: any;
} | {
    generations: any[];
    llmOutput: {};
}>
```

jacoblee93:
Gotcha. The only difference the `await` makes here after a `return` is in the call stack if an error were to occur:

https://stackoverflow.com/questions/38708550/difference-between-return-await-promise-and-return-promise

kevinnayar:
Yep, understand what awaits does. I did think it was weird that it resolves the issue when this minor edit of the dist file. I'll do a bit more debugging and create a more meaningful bug first. The issue does persist without it. Thanks.

Issue #5069, opened 2024-04-11
google-genai [feature]: Support Gemini system messages


afirstenberg:
### Privileged issue

- [X] I am a LangChain maintainer, or was asked directly by a LangChain maintainer to create an issue here.

### Issue Content

See https://github.com/langchain-ai/langchain-google/issues/129

https://ai.google.dev/docs/system_instructions
https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/gemini#request_body (for reference only)

This isn't implemented for all models, so also add `convertSystemMessageToHumanContent` to be compatible with Python. Also see the warning message generated at https://github.com/langchain-ai/langchain-google/blob/62ba6ca5e39174d467e47c83119c8bb4275128f6/libs/vertexai/langchain_google_vertexai/chat_models.py#L175

shannon-ab:
Hi @afirstenberg @jacoblee93 we are students from the University of Toronto and we are interested in contributing to this issue. Would it be possible for us to take it on? If so, are there any specific guidelines or suggestions you would like us to follow as we work on it?

jacoblee93:
Hey @shannon-ab! I'm not sure the status of this but you can check out general guidelines here:

https://github.com/langchain-ai/langchainjs/blob/main/CONTRIBUTING.md

But in general feel free to give it a best shot and let us refine it!

shannon-ab:
Hi @jacoblee93! Thank you for your response and the guidelines. We have an issue analysis document that outlines the code changes for this issue which you can find here: https://docs.google.com/document/d/1YDUxJVI9NPu7nPObFppPfkdn1ZSR5BsosZdUkuDNB6I/edit?tab=t.0#heading=h.6w32w3mbbyvh
We plan to have a PR ready by early to mid-November. I saw from the guidelines attached that we should assign the issue to ourselves once we start working on it, however I am unable to assign it to myself, could you please assign this issue to me? Thanks again!

afirstenberg:
@shannon-ab - Great to see your group taking this on, and it is good to see you have a preliminary design document. Some notes on your analysis:
* PaLM has been decommissioned (see https://ai.google.dev/palm_docs/deprecation). There is no need to reference it anymore.
* In general, it may be useful to see how the google-common package is doing things. (The google-common package is doing things with REST, and not Google's library, and it aims to support both the AI Studio and the Vertex AI endpoint, but it may provide some insight about how to address some issues.)
[...]

PR #7235, opened 2024-11-20
feat(google-genai): Support Gemini system instructions

chrisnathan20:
Fixes [#5069](https://github.com/langchain-ai/langchainjs/issues/5069)

Currently for Google Gen AI, system message (it can only be the first element in list of messages) are prepended to the following human message which is used to steer the model's behavior. The handling of system messages is the same although some newer gen ai models support direct invocation of system messages via the systemInstructions field per https://ai.google.dev/gemini-api/docs/system-instructions?lang=node

This PR provides support for models that allow direct system instructions by providing an alternative method of handling system messages when possible or set via convertSystemMessageToHumanContent. The changes in this PR are modeled after the changes in google-common to support system instructions (https://github.com/langchain-ai/langchainjs/pull/5089)

High Level Outline of Changes and Additions:
1. added optional convertSystemMessageToHumanContent which overrides dependency on model to use new system message logic or not. This is also done to match the Generative AI package on Python.
2. added method computeUseSystemInstruction to determine whether model supports system Instructions or not. List of supported and unsupported models taken from google-common's with deprecated models (palm) removed.
3. Added new optional boolean parameter to convertBaseMessagesToContent which defaults to false. Setting this to true triggers new handling of system messages where system messages are treated as independent messages and not prepended to following user message while maintaining existing restrictions.
4. Prior to sending prompt for response generation, if prompt's first message is an independent system message, then it is removed from the actual prompt to be sent and set as the system instruction of the model. It is important to note that an independent system message can only be the first message in prompt and can only exist if new system message handling in convertBaseMessagesToContent is triggered which can only be for model that supports system instructions.

Test Cases created:
1. Given: Input has single system message followed by one user message 
When: convertBaseMessagesToContent is invoked with 3rd parameter set to true
Then: the System message is removed from prompt and passed as systemInstruction field instead and actualPrompt would only have 1 user message under the role of user

2. Given: Input has single system message followed by one user message 
When: convertBaseMessagesToContent is invoked with 3rd parameter set to false
Then: the System message is not removed from prompt as actualPrompt would only have the system message prepended to the user message under the role of user

3. Given: Input has a system message that is not the first message 
When: convertBaseMessagesToContent is invoked with 3rd parameter set to true
Then: convertBaseMessagesToContent should raise the error - "System message should be the first one"

4. Given: Input has a system message that is not the first message 
When: convertBaseMessagesToContent is invoked with 3rd parameter set to false
Then: convertBaseMessagesToContent should raise the error - "System message should be the first one"

5. Given: Input has multiple system messages
When: convertBaseMessagesToContent is invoked with 3rd parameter set to true
Then: convertBaseMessagesToContent should raise the error - "System message should be the first one"

6. Given: Input has multiple system messages
When: convertBaseMessagesToContent is invoked with 3rd parameter set to false
Then: convertBaseMessagesToContent should raise the error - "System message should be the first one"

7. Given: Input has no system message and one user message
When: convertBaseMessagesToContent is invoked with 3rd parameter set to true
Then: actualPrompt would only have the user message under the role of user

8. Given: Input has no system message and one user message
When: convertBaseMessagesToContent is invoked with 3rd parameter set to false
Then: actualPrompt would only have the user message under the role of user

9. Given: Input has no system message and multiple user messages
When: convertBaseMessagesToContent is invoked with 3rd parameter set to true
Then: actualPrompt would only have the first user message followed by the next user messages as separate messages

10. Given: Input has no system message and multiple user messages
When: convertBaseMessagesToContent is invoked with 3rd parameter set to false
Then: actualPrompt would only have the first user message followed by the next user messages as separate messages

Credits:
Issue selection -  @shannon-ab 
Implementation - @chrisnathan20 @garychen2002 
Testing - @shannon-ab @martinl498


chrisnathan20:
Hi @jacoblee93 and @afirstenberg - this is our PR to support system instructions for capable Generative AI models based on the implementation plan shared under the issue. If there are any points of concern or improvements, please do not hesitate to let us know. Thank you and have a great rest of your day!

jacoblee93 (approved):
Looks good, apologies for the delay in review!

chrisnathan20:
Thanks for the review @jacoblee93! I noticed that the formatting check failed but when I tried to do _yarn lint_ inside the latest _libs\langchain-google-genai_, it doesn't raise any concerns or errors. Is there anything that needs to be done from our side?

chrisnathan20:
Thanks for the review @jacoblee93! I also want to let you know that I pushed a new commit to fix formatting issues in case there is anything else that needs to be done from your end

jacoblee93:
Nope all good, about to pull and try it out now

jacoblee93:
Thank you!

chrisnathan20:
@jacoblee93 Thanks for the review!

PR #7236, opened 2024-11-20
feat(community): Incorporate BM25 score in the results

felipediel:
In the BM25Retriever, include the score in each document's metadata.

The score is useful for performing reranking, combining BM25 with other algorithms.

### DIFF
--- a/libs/langchain-community/src/retrievers/bm25.ts
+++ b/libs/langchain-community/src/retrievers/bm25.ts
@@ -53,6 +53,13 @@ export class BM25Retriever extends BaseRetriever {
 
     scoredDocs.sort((a, b) => b.score - a.score);
 
-    return scoredDocs.slice(0, this.k).map((item) => item.document);
+    return scoredDocs.slice(0, this.k).map((item) => new Document({
+      ...item.document.id && { id: item.document.id },
+      pageContent: item.document.pageContent,
+      metadata: {
+        score: item.score,
+        ...item.document.metadata,
+      }
+    }));
   }
 }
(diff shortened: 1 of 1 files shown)


libs/langchain-community/src/retrievers/bm25.ts
```diff
@@ -53,6 +53,13 @@ export class BM25Retriever extends BaseRetriever {
...
 
-    return scoredDocs.slice(0, this.k).map((item) => item.document);
+    return scoredDocs.slice(0, this.k).map((item) => new Document({
+      ...item.document.id && { id: item.document.id },
+      pageContent: item.document.pageContent,
+      metadata: {
+        score: item.score,
+        ...item.document.metadata,
+      }
+    }));
```
felipediel, line 63: I would prefer to use the ranked document structure `{document: Document, score: number}`; however, I believe this change could disrupt client applications that rely on the existing structure. By adding the score to the metadata instead, we introduce the feature in a non-disruptive manner.

libs/langchain-community/src/retrievers/bm25.ts
```diff
@@ -53,6 +53,13 @@ export class BM25Retriever extends BaseRetriever {
 
     scoredDocs.sort((a, b) => b.score - a.score);
 
-    return scoredDocs.slice(0, this.k).map((item) => item.document);
+    return scoredDocs.slice(0, this.k).map((item) => new Document({
+      ...item.document.id && { id: item.document.id },
+      pageContent: item.document.pageContent,
+      metadata: {
+        score: item.score,
```
jacoblee93, line 60: Will rename this to `bm25Score` to avoid conflict

libs/langchain-community/src/retrievers/bm25.ts
jacoblee93: Yes thanks for this!

jacoblee93:
Let's actually make it configurable to really avoid disrupting anyone

jacoblee93 (approved):

PR #7242, opened 2024-11-22
feat(azure-cosmosdb): add session context and retrieve all sessions for a user

sinedied:
- Added generic `context` object saved along the session. This can be used for example to give a title or save any additional properties along with a session.
- Added method to retrieve all sessions for a user (without the messages)
- Update to latest CosmosDB SDK (had to fix a few types in vector search integrations)

@jacoblee93 Not directly related to the PR, but I've found some memory integration that uses `RunnableWithMessageHistory` like ours, and other using `BufferMemory` which one is the recommended approach? Thanks

### DIFF
--- a/libs/langchain-azure-cosmosdb/src/azure_cosmosdb_nosql.ts
+++ b/libs/langchain-azure-cosmosdb/src/azure_cosmosdb_nosql.ts
@@ -15,7 +15,9 @@ import {
   IndexingPolicy,
   SqlParameter,
   SqlQuerySpec,
+  VectorEmbedding,
   VectorEmbeddingPolicy,
+  VectorIndex,
 } from "@azure/cosmos";
 import { DefaultAzureCredential, TokenCredential } from "@azure/identity";
 
@@ -186,7 +188,7 @@ export class AzureCosmosDBNoSQLVectorStore extends VectorStore {
           distanceFunction: "cosine",
           // Will be determined automatically during initialization
           dimensions: 0,
-        },
+        } as VectorEmbedding,
       ];
     }
 
@@ -195,7 +197,7 @@ export class AzureCosmosDBNoSQLVectorStore extends VectorStore {
         {
           path: "/vector",
           type: "quantizedFlat",
-        },
+        } as VectorIndex,
       ];
     }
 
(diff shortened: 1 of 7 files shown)


aditishree1 (commented):
nit: Should we add this in example or doc-comment to let the user know that they have the option to add context in input?

Otherwise, this PR LGTM.

aditishree1:
hey @sinedied, Jacob suggested to prefer RunnableWithMessageHistory. 
 https://github.com/langchain-ai/langchainjs/pull/7057#discussion_r1837250292

sinedied:
> nit: Should we add this in example or doc-comment to let the user know that they have the option to add context in input?
> 
> Otherwise, this PR LGTM.

Good callout! I'll update the example with the new methods.

sinedied:
@aditishree1 I completed the code example with the new APIs usage.

I also had to fix a few types in the vector search integration due to the updated Cosmos SDK.

aditishree1 (approved):

jacoblee93:
> which one is the recommended approach?

We have been leaning more and more on LangGraph as the memory abstraction of choice (you are defining a state with whatever keys you want) but if you need to stay in LangChain then `RunnableWithMessageHistory`

libs/langchain-azure-cosmosdb/src/azure_cosmosdb_nosql.ts
```diff
@@ -195,7 +197,7 @@ export class AzureCosmosDBNoSQLVectorStore extends VectorStore {
         {
           path: "/vector",
           type: "quantizedFlat",
-        },
+        } as VectorIndex,
```
jacoblee93, line 200: Maybe `as const` works but not a huge deal

libs/langchain-azure-cosmosdb/src/chat_histories.ts
```diff
@@ -201,4 +218,36 @@ export class AzureCosmsosDBNoSQLChatMessageHistory extends BaseListChatMessageHi
       await this.container.item(userSession.id, userId).delete();
     }
   }
+
+  async getAllSessionsForUser(userId: string): Promise<ChatSession[]> {
```
jacoblee93, line 222: Should this be a util instead? Doesn't seem to really fit on this class

libs/langchain-azure-cosmosdb/src/azure_cosmosdb_nosql.ts
sinedied: Tried it but it still complained 😞

sinedied:
@jacoblee93 @aditishree1 I've reworked how the helper methods works so it makes more sense, and fixed handling of existing context in a session.

jacoblee93 (approved):
Thanks! Looks good

PR #7249, opened 2024-11-23
Update Pinecone indexing example to show use of deletionProtection

aulorbe:
Now that LangChain uses the newest version of the Pinecone client, let's show users how to use (or not) `deletionProtection` in the examples :) 

More about deletion protection here https://docs.pinecone.io/guides/indexes/prevent-index-deletion#disable-deletion-protection

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1208759468706823


jacoblee93 (approved):
Thank you!

Issue #7095, opened 2024-10-28
Port ArxivRetriever to LangChainJS


bracesproul:
### Discussed in https://github.com/langchain-ai/langchainjs/discussions/7077

<div type='discussions-op-text'>

<sup>Originally posted by **boni-teppanyaki** October 26, 2024</sup>
### Checked

- [X] I searched existing ideas and did not find a similar one
- [X] I added a very descriptive title
- [X] I've clearly described the feature request and motivation for it

### Feature request

Add an ArxivRetriever component to LangChainJS, enabling JavaScript-based applications to query and retrieve research papers directly from arXiv. This retriever would allow users to pull academic content by setting parameters such as titles, authors, topics and abstracts, as well as through natural language queries or article identifiers.

### Motivation

Incorporation of arXiv will extend LangChainJS's utility for research, question-answering, summarization, and other use cases by retrieving and utilizing academic resources from the arXiv platform. Currently, arXiv support is available only in LangChain Python. We would like to port this feature so that we can use it in js version.

### Proposal (If applicable)

We would like to develop this feature and contribute it as a community package.</div>

cc @boni-teppanyaki

PR #7250, opened 2024-11-24
feat(community): Port ArxivRetriever to LangChainJS

boni-teppanyaki:
- **Description:**
  This PR ports ArxivRetriever from LangChain Python to LangChainJS.  An ArxivRetriever component would grant users access to up-to-date research directly from arXiv. This retriever would allow users to pull full text and summaries through natural language queries or arXiv identifiers. Specifically, we have added
  - Arxiv utils & retriever implementation,
  - documentation for ArxivRetriever,
  - integration tests for ArxivRetriever,
  - example usages for ArxivRetriever. 

- **Issue:**
  This PR addresses https://github.com/langchain-ai/langchainjs/issues/7095


libs/langchain-community/src/utils/arxiv.ts
```diff
@@ -0,0 +1,211 @@
+import axios from 'axios'; // For HTTP requests
```
jacoblee93, line 1: Let's just use `fetch`

Googlogogo: Hi, sorry for the late response, but our team have just modified the file to use ```fetch``` instead of ```axios```. We have tested our code and the test cases still passed. Thank you!

libs/langchain-community/langchain.config.js
```diff
@@ -193,6 +193,7 @@ export const config = {
     // retrievers
     "retrievers/amazon_kendra": "retrievers/amazon_kendra",
     "retrievers/amazon_knowledge_base": "retrievers/amazon_knowledge_base",
+    "retrievers/arxiv": "retrievers/arxiv",
```
jacoblee93, line 196: Need to add this under `requiresOptionalDeps` below

jacoblee93:
Thanks!

Issue #7143, opened 2024-11-03
Tool Support for Hedera Integration


aks-456:
### Privileged issue

- [X] I am a LangChain maintainer, or was asked directly by a LangChain maintainer to create an issue here.

### Issue Content

This is a continuation of the discussion post #7097 . 
We would like to add support for interaction with the Hedera network, including smart contracts, wallet interactions, and access to information through the Mirror Node API. This improvement allows LangChain users to interact with the Hedera network to provide access to verified data, make and record transactions.

We intend to make extensive use of the Hedera REST API, providing features including but not necessarily limited to:

Query Accounts and their balances
List NFTs and Tokens owned by an account
Make and Schedule transactions
Query information from Smart Contracts including execution results and state.

The above could theoretically be equally accomplished with a Document Loader, but the reason for implementation as a Tool is that we also intend to allow agentic behaviour by providing the AI with the ability to interface with the Hedera network:

Prerequisite: The Agent must be provided with a Hedera Account on the same network as defined in baseURL; functioning analogous to an API key, and distinct from the traditional definition of an account as it is closer to a blockchain address/wallet.

Query balance of the provided wallet
Read the Transaction History
Make Transactions
Interact with Smart Contracts (Depending on the implementation of the contract)

Since the tool is multifunctional, we intend to follow a similar general structure to the Gmail tool described in https://js.langchain.com/docs/integrations/tools/gmail/

i.e. We intend to define a directory for our tool that defines each functionality we implement, with each file defining a subset of functionality relating to either smart contracts, wallets, or the REST API.

Since some of these actions are sensitive, e.g. making transactions, which could cost real money, we intend to implement a permissioning system such that upon initialisation of the tool, its capabilities must be defined by the developer initialising it and providing it to their agent. These permissions will be the following:

-> REST API permissions
-> Smart Contract Query permissions
-> Smart Contract Create/Update permissions
-> Wallet Read Permissions (Query Balance, etc.)
-> Wallet Write Permissions (Posting Transactions, etc.)

An agent provided with a tool with sufficient permissions may then call a given tool like so:

```ts
import { initializeAgentExecutorWithOptions } from "langchain/agents";
import { OpenAI } from "@langchain/openai";
import {
  HederaQueryBalance,
  HederaTransactionHistory,
  HederaSendTransaction,
  HederaSmartContractQuery,
} from "@langchain/community/tools/hedera";
import { StructuredTool } from "@langchain/core/tools";

export async function run() {
[...]

PR #7253, opened 2024-11-24
Tool Support for Hedera Integration

aks-456:
This is related to issue #7143 Tool Support for Hedera Integration. 

**Description**

This PR adds support for Hedera tools such as:
* Creating an account
* Deleting an account
* Querying account balance
* Initiating HBAR transfers between accounts


syntaxsec:
We're working on potentially adding ECDSA and Smart Contracts support as well, but from what we can tell this is just about complete as a wallet-interfacing tool on its own.

We'd love feedback and suggestions.

Xpereze:
Hi @jacoblee93, could you take a look at this? it would be greatly appreciated!

docs/core_docs/docs/integrations/tools/hedera.mdx
jacoblee93, line 1: Could you use this template?

https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-scripts/src/cli/docs/templates/tools.ipynb

examples/src/tools/hedera.ts
```diff
@@ -0,0 +1,82 @@
...
+      new HederaCreateAccount(hederaParams),
+      new HederaDeleteAccount(hederaParams),
+    ];
+
+    // Setup the agent to use the hedera tool
+    const prompt = await pull<ChatPromptTemplate>(
+        "hwchase17/openai-functions-agent"
+      );
+      
+      const agent = await createOpenAIFunctionsAgent({
```
jacoblee93, line 44: Let's use `createReactAgent` from `@langchain/langgraph` here instead

libs/langchain-community/package.json
```diff
@@ -35,6 +35,7 @@
   "author": "LangChain",
   "license": "MIT",
   "dependencies": {
+    "@hashgraph/sdk": "^2.53.0",
```
jacoblee93, line 38: Should be a peer + dev dep, not a direct dependency:

https://github.com/langchain-ai/langchainjs/blob/main/.github/contributing/INTEGRATIONS.md#third-party-dependencies

examples/src/tools/hedera.ts
MuhannadJam: @jacoblee93 would it be possible to use ```createStructuredChatAgent``` instead?

jacoblee93: No, let's show the latest recommended practices

libs/langchain-community/package.json
syntaxsec: Will be done by EOD Monday. Are there any other changes we need to make?

hntrl:
Hey @aks-456, sorry to revive an old PR!

PR #7254, opened 2024-11-24
fix(community): PrismaVectorStore handle empty array in filter

shan-mx:
When an an empty array is passed to IN or NOTIN operator in PrismaVectorStore's filter, it will be directly passed to Prisma.join and cause
 **TypeError: Expected `join([])` to be called with an array of multiple elements, but got an empty array**

This PR adds a separate logic for handling empty arrays in SQL queries:

* [`libs/langchain-community/src/vectorstores/prisma.ts`](diffhunk://#diff-5b221be6896a52539d6efcdf50be7c50df05b6a025c8737730b782b672c90164R440-R449): Added a condition to check if the array is empty and handle `IN` and `NOT IN` operators appropriately. For empty arrays, `IN ()` will return `FALSE` and `NOT IN ()` will return `TRUE`.

### DIFF
--- a/libs/langchain-community/src/vectorstores/prisma.ts
+++ b/libs/langchain-community/src/vectorstores/prisma.ts
@@ -437,6 +437,16 @@ export class PrismaVectorStore<
                   )}`
                 );
               }
+
+              if (value.length === 0) {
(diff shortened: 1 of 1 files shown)


jacoblee93 (approved):
Thank you!

PR #7256, opened 2024-11-24
feat(community): Add property extraction for Nodes and Relationships in LLMGraphTransformer

gvasilei:
Adds support for extracting Node and Relationship properties on `LLMGraphTransformer`, similar to the Python version. A key distinction is that arbitrary property extraction is not supported. The caller needs to specify the property keys in advance.  

This is my first contribution here, so please let me know if I've missed anything, or if there is any feedback. Thank you!


jacoblee93:
Thank you! Tagging in @easwee or @tomasonjo to have a look

libs/langchain-community/tsconfig.json
```diff
@@ -5,8 +5,8 @@
     "rootDir": "./src",
     "target": "ES2021",
     "lib": ["ES2021", "ES2022.Object", "DOM"],
-    "module": "ES2020",
-    "moduleResolution": "nodenext",
+    "module": "NodeNext",
```
tomasonjo, line 8: Probably don't want to change tsconfig file

tomasonjo (approved):
LGTM

libs/langchain-community/tsconfig.json
gvasilei: Indeed! I was using Typescript 5.2 locally and faced an issue with module being `ES2020`. Reverted.

jacoblee93: Yeah, you have to use TS 5.1. It's bad and we need to fix it at some point.

gvasilei: Yes, that would be good. In any case, I've reverted that particular change. @jacoblee93 do you see any other areas for feedback?

jacoblee93:
Thank you both!

Issue #5073, opened 2024-04-11
google-common [Feature]: Grounding with Google Search and Vertex AI Search


afirstenberg:
### Privileged issue

- [X] I am a LangChain maintainer, or was asked directly by a LangChain maintainer to create an issue here.

### Issue Content

See https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/grounding

See https://github.com/langchain-ai/langchain-google/issues/146

afirstenberg:
Do we have a standard way of defining non-FunctionCall tools? (@jacoblee93)

jacoblee93:
> Do we have a standard way of defining non-FunctionCall tools? (@jacoblee93)

Not at the moment - thinking so far has been to always support the model-specific format and make the `.tool_calls` format on `AIMessage` expressive enough to handle the various possibilities.

afirstenberg:
My thinking has been that there would be a function call version of it (for most models to use), but also a function call that a specific model could check for and then use.

So a Gemini model, for example, might check for a function `toGeminiTool()` or something and, if it exists, call it and use those results instead.

williamc99:
Hi @afirstenberg and @jacoblee93, we are students from the University of Toronto and we are interested in contributing to this issue.  We noticed that a similar issue: [langchain-ai/langchain-google/issues/146](https://github.com/langchain-ai/langchain-google/issues/146) was recently completed, so we were wondering if this issue is still open for contribution.  If so, we would like to implement a similar solution based on the existing one from langchain-google.  Thanks! 😄

jacoblee93:
Hey @williamc99! I think this should work if you call `.bindTools()` with Google's format but haven't verified. Would be a nice one to add if you're willing to take it on!

williamc99:
No worries, my team and I will get started right away!

afirstenberg:
@williamc99 - Please keep me in the loop on what you and your team are seeing or if you have questions.

When I first looked at this, I had several thoughts around this:
* Google is using Tool Calls to implement *both* function calling and other tools. So functions are a subset of all tools.
* As such, the `ToolCallRaw` (as it stands now) type isn't appropriate for other Gemini tools, such as the grounding tools.
  * So my thought would be to rename `ToolCallRaw` to `FunctionCallToolRaw`
  * We might then create other type definitions for `GoogleRetrievalToolRaw` and `CodeExecutionToolRaw` and others that may come up in the future.
[...]

PR #7280, opened 2024-11-27
feat(google-common): Grounding with Google Search and Vertex AI Search

renathossain:
We’ve implemented grounding with Google Search and Vertex AI Search. To use this feature, simply call `.bindTools([searchRetrievalTool])` on the model and pass in the appropriate `searchRetrievalTool` object, as defined in the [Google Cloud documentation on grounding](https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/grounding), depending on the type of grounding you want to use.

Here are the implementation details:

- We’ve added the necessary `googleSearchRetrieval` and `retrieval` parameters to `GeminiTool`.
- `convertToGeminiTools` and `formatTools` previously supported only one tool type (`functionDeclarations`), but we’ve refactored them to handle these two new tools.
- The `structuredToolsToGeminiTools` function has been removed; its logic is now consolidated within `formatTools`.

We've also added test cases and documentation changes.

Twitter handle: @renathossain

Fixes [#5073](https://github.com/langchain-ai/langchainjs/issues/5073)


libs/langchain-google-common/src/utils/common.ts
```diff
@@ -62,32 +62,38 @@ function processToolChoice(
...
-  ];
+  const geminiTools: GeminiTool[] = [];
   tools.forEach((tool) => {
     if (
       "functionDeclarations" in tool &&
       Array.isArray(tool.functionDeclarations)
     ) {
       const funcs: GeminiFunctionDeclaration[] = tool.functionDeclarations;
-      geminiTools[0].functionDeclarations?.push(...funcs);
+      geminiTools.push({ functionDeclarations: [...funcs] });
```
jacoblee93, line 72: I vaguely remember this not working - can you add a test using multiple traditional tools?

jacoblee93 (commented):
Thank you, and sorry for the delayed review!

See one comment but looks fine otherwise

jacoblee93:
CC @afirstenberg for a peek as well!

libs/langchain-google-common/src/utils/common.ts
renathossain: Hello @afirstenberg and @jacoblee93,

Thank you for the review! We’re glad to hear that you appreciate our code contribution. We’ll make sure to add the additional test cases as requested. Please note that libs/langchain-google-vertexai/src/tests/chat_models.int.test.ts already includes test cases for grounding.

afirstenberg: I think there are some known bugs on Google's side about having multiple tools at once - but Google considers this a bug on their end. It *should* be possible to submit multiple tools at once and mixing functions and other tools.

libs/langchain-google-common/src/types.ts
```diff
@@ -309,6 +309,8 @@ export interface GeminiContent {
 
 export interface GeminiTool {
   functionDeclarations?: GeminiFunctionDeclaration[];
+  googleSearchRetrieval?: object;
+  retrieval?: object;
```
afirstenberg, line 313: I would have like to see these as actual types or classes rather than as `object`, but I don't think this should hold off on merging.

afirstenberg:
Generally looks good and I'm sorry I missed this two weeks ago.

I do suggest making actual types for the possible attributes for `GeminiTool`, but I don't think that should hold up a merge.

LGTM

libs/langchain-google-common/src/utils/common.ts
jacoblee93: Yes this breaks multiple functions, will revert partially

jacoblee93 (approved):
Thanks so much!

jacoblee93:
Google has changed the field name for Gemini 2.0, fixing now

PR #7290, opened 2024-11-29
chore(community): Bump faker dev dep to 8.4.1

mgiorgino-iobeya:
This simply bumps faker dev dependency in community dir (to be used in another PR to come).


jacoblee93 (approved):
Thanks!

PR #7298, opened 2024-11-30
chore(core): externalize @cfworker/json-schema dependency

jdanyow:
Back in https://github.com/langchain-ai/langchainjs/pull/1878, the dependency was inlined to avoid portability issues, which have since been resolved.


jacoblee93:
Thank you and thank you for your awesome work!

jacoblee93:
Ah seems it's still failing our export tests:

```
/app/libs/langchain-core/dist/output_parsers/transform.cjs:4
const json_schema_1 = require("@cfworker/json-schema");
                      ^

Error [ERR_REQUIRE_ESM]: require() of ES Module /app/node_modules/@cfworker/json-schema/dist/cjs/index.js from /app/libs/langchain-core/dist/output_parsers/transform.cjs not supported.
index.js is treated as an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which declares all .js files in that package scope as ES modules.
Instead either rename index.js to end in .cjs, change the requiring code to use dynamic import() which is available in all CommonJS modules, or change "type": "module" to "type": "commonjs" in /app/node_modules/@cfworker/json-schema/package.json to treat all .js files as CommonJS (using .mjs for all ES modules instead).

    at Object.<anonymous> (/app/libs/langchain-core/dist/output_parsers/transform.cjs:4:23)
    at Object.<anonymous> (/app/libs/langchain-core/dist/output_parsers/bytes.cjs:4:24)
    at Object.<anonymous> (/app/libs/langchain-core/dist/output_parsers/index.cjs:18:14)
    at Object.<anonymous> (/app/libs/langchain-core/output_parsers.cjs:1:18)
    at Object.<anonymous> (/app/libs/langchain-openai/dist/chat_models.cjs:11:26)
    at Object.<anonymous> (/app/libs/langchain-openai/dist/index.cjs:21:14)
    at Object.<anonymous> (/app/libs/langchain-openai/index.cjs:1:18)
    at Object.<anonymous> (/app/src/require.cjs:2:20) {
  code: 'ERR_REQUIRE_ESM'
}
```

jdanyow:
hmm- my bad- I'll look into it and post back with an update soon. We had just added a `require` export to the package so I thought it was finally resolved.
https://github.com/cfworker/cfworker/blob/main/packages/json-schema/package.json#L28

jacoblee93:
If helpful - after a lot of trial and error we ended up on this:

https://github.com/langchain-ai/langchainjs/blob/main/langchain-core/package.json#L298

I do not know we need all of it since we're using it internally (our users won't need the CF worker types, for example)

jdanyow:
I ended up switching my build to https://www.npmjs.com/package/tshy, which simplified things, and seems to produce a solid dual ESM/CommonJS distributable.

The package no longer has any dependencies, other than devDependencies, hopefully that eliminates the cloudflare worker types related issues.

jacoblee93:
Giving it another shot now thanks!

jacoblee93:
Good enough for our CI! Thank you so much.

PR #7300, opened 2024-11-30
feat(community): Added Reddit integration with tool and document loader

MirrorLimit:
Adds feature as per discussion #7043

**Description:**
The JavaScript version of LangChain currently lacks several features available in the Python version. This PR aims to bridge that gap by adding a Reddit integration with the following features:

- Reddit tool for searching posts from a subreddit or from a Reddit user
- Reddit document loader that loads Reddit posts as documents from a subreddit or from a Reddit user and formats them into a structured document format
- API wrapper for the Reddit integration to communicate to Reddit with

Made together with @HyphenHook, @baoyng, and @erinL168


libs/langchain-community/src/utils/reddit.ts
```diff
@@ -0,0 +1,145 @@
+import dotenv from "dotenv";
+import { AsyncCaller } from "@langchain/core/utils/async_caller";
+
+dotenv.config();
```
nick-w-nick, line 4:
```suggestion
import "dotenv/config";
import { AsyncCaller } from "@langchain/core/utils/async_caller";
```

This could be changed to a side-effect import since the default configuration isn't being changed.

nick-w-nick: @MirrorLimit I actually don't see `process.env` being used here at all, can this import be removed?
```suggestion
import { AsyncCaller } from "@langchain/core/utils/async_caller";
```

libs/langchain-community/src/utils/reddit.ts
```diff
@@ -0,0 +1,145 @@
...
+      onFailedAttempt: (error) => {
+        console.error("Attempt failed:", error.message);
+      },
+    });
+  }
+
+  private async authenticate() {
+    if (this.token) return;
+
+    const authString = btoa(`${this.clientId}:${this.clientSecret}`);
```
nick-w-nick, line 51: @MirrorLimit `btoa` is usually a browser-specific function IIRC and isn't meant to be used in Node. Here is the proper way to create the base64-encoded string.

```suggestion
    const authString = Buffer.from(`${this.clientId}:${this.clientSecret}`).toString('base64');
```

libs/langchain-community/src/utils/reddit.ts
```diff
@@ -0,0 +1,145 @@
...
+        throw new Error(
+          `Error authenticating with Reddit: ${response.statusText}`
+        );
+      }
+
+      const data = await response.json();
+      this.token = data.access_token;
+    } catch (error) {
+      console.error("Error authenticating with Reddit:", error);
+    }
```
nick-w-nick, line 77: If I am reading this correctly, if it encounters an issue, it will throw an error with the text 

`Error authenticating with Reddit: ${response.statusText}`

but will then do `console.error("Error authenticating with Reddit:", error);`, which will output something like:
```
Error authenticating with Reddit:
  Error authenticating with Reddit: "invalid auth token"
```

I'd suggest just throwing the status text from Reddit instead and then logging out the authentication error in the catch instead.

```suggestion
      if (!response.ok) {
        throw new Error(response.statusText);
      }

      const data = await response.json();
      this.token = data.access_token;
    } catch (error) {
      console.error("Error authenticating with Reddit:", error);
    }
```

libs/langchain-community/src/utils/reddit.ts
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  private async makeRequest(
+    endpoint: string,
+    params: Record<string, any> = {}
+  ): Promise<any> {
+    await this.authenticate();
+
+    const url = new URL(`${this.baseUrl}${endpoint}`);
```
nick-w-nick, line 86: Updated to use the URL constructor as intended
```suggestion
    const url = new URL(endpoint, this.baseUrl);
```

libs/langchain-community/src/utils/reddit.ts
```diff
@@ -0,0 +1,145 @@
...
+      return await response.json();
+    });
+  }
+
+  async searchSubreddit(
+    subreddit: string,
+    query: string,
+    sort: "new",
+    limit: 10,
+    time: "all"
```
nick-w-nick, line 118: Are these meant to be defaults?

```suggestion
    sort = "new",
    limit = 10,
    time = "all"
```

libs/langchain-community/src/utils/tests/reddit.test.ts
```diff
@@ -0,0 +1,185 @@
...
+
+    await redditAPIWrapper["authenticate"](); 
+
+    // Validate that the fetch was called correctly with the correct headers and URL
+    expect(global.fetch).toHaveBeenCalledWith(
+      "https://www.reddit.com/api/v1/access_token",
+      expect.objectContaining({
+        method: "POST",
+        headers: expect.objectContaining({
+          Authorization: expect.stringContaining("Basic"), // Checks if Basic auth is used
```
nick-w-nick, line 50: @MirrorLimit What are your thoughts on including the base64 string generation in the test where you set what the generated token should be and then verify that it was properly generated and used in the request?

libs/langchain-community/src/document_loaders/web/reddit.ts
```diff
@@ -0,0 +1,133 @@
...
+    clientId = getEnvironmentVariable("REDDIT_CLIENT_ID") as string,
+    clientSecret = getEnvironmentVariable("REDDIT_CLIENT_SECRET") as string,
+    userAgent = getEnvironmentVariable("REDDIT_USER_AGENT") as string,
+    searchQueries,
+    mode,
+    categories = ["new"],
+    numberPosts = 10,
+  }: RedditAPIConfig & {
+    searchQueries: string[];
+    mode: string;
```
nick-w-nick, line 57: Would be a lot nicer if these were typed instead of just a string since other values with cause errors to be thrown.

```suggestion
    mode: "subreddit" | "username";
```

libs/langchain-community/src/document_loaders/web/reddit.ts
```diff
@@ -0,0 +1,133 @@
...
+    this.clientId = clientId;
+    this.clientSecret = clientSecret;
+    this.userAgent = userAgent;
+    this.redditApiWrapper = new RedditAPIWrapper({
+      clientId: this.clientId,
+      clientSecret: this.clientSecret,
+      userAgent: this.userAgent,
+    });
+    this.searchQueries = searchQueries;
+    this.mode = mode;
```
nick-w-nick, line 71: Would be a lot nicer if these were typed instead since different values cause errors to be thrown.

```suggestion

export type SearchMode = "subreddit" | "username";

/**
 * Class representing a document loader for loading Reddit posts. It extends
 * the BaseDocumentLoader and implements the RedditAPIConfig interface.
 * @example
 * ```typescript
 * const loader = new RedditPostsLoader({
 *   clientId: "REDDIT_CLIENT_ID",
 *   clientSecret: "REDDIT_CLIENT_SECRET",
 *   userAgent: "REDDIT_USER_AGENT",
 *   searchQueries: ["LangChain", "Langchaindev"],
 *   mode: "subreddit",
 *   categories: ["hot", "new"],
 *   numberPosts: 5
 * });
 * const docs = await loader.load();
 * ```
 */
export class RedditPostsLoader
  extends BaseDocumentLoader
  implements RedditAPIConfig
{
  public clientId: string;

  public clientSecret: string;

  public userAgent: string;

  private redditApiWrapper: RedditAPIWrapper;

  private searchQueries: string[];

  private mode: SearchMode;

  private categories: string[];

  private numberPosts: number;

  constructor({
    clientId = getEnvironmentVariable("REDDIT_CLIENT_ID") as string,
    clientSecret = getEnvironmentVariable("REDDIT_CLIENT_SECRET") as string,
    userAgent = getEnvironmentVariable("REDDIT_USER_AGENT") as string,
    searchQueries,
    mode,
    categories = ["new"],
    numberPosts = 10,
  }: RedditAPIConfig & {
    searchQueries: string[];
    mode: SearchMode;
    categories?: string[];
    numberPosts?: number;
  }) {
    super();
    this.clientId = clientId;
    this.clientSecret = clientSecret;
    this.userAgent = userAgent;
    this.redditApiWrapper = new RedditAPIWrapper({
      clientId: this.clientId,
      clientSecret: this.clientSecret,
      userAgent: this.userAgent,
    });
    this.searchQueries = searchQueries;
    this.mode = mode;
```

libs/langchain-community/src/document_loaders/web/reddit.ts
```diff
@@ -0,0 +1,133 @@
...
+          posts = await this.redditApiWrapper.fetchUserPosts(
+            query,
+            category,
+            this.numberPosts
+          );
+        } else {
+          throw new Error(
+            "Invalid mode: please choose 'subreddit' or 'username'"
+          );
+        }
```
nick-w-nick, line 105: This _could_ be a switch statement, which is technically faster and more consistent with what is used elsewhere in the library.

```suggestion
        switch (this.mode) {
          case "subreddit":
            posts = await this.redditApiWrapper.searchSubreddit(
              query,
              "*",
              category,
              this.numberPosts
            );
            break;
          case "username":
            posts = await this.redditApiWrapper.fetchUserPosts(
              query,
              category,
              this.numberPosts
            );
            break;
          default:
            throw new Error(
              "Invalid mode: please choose 'subreddit' or 'username'"
            );
        }
```

libs/langchain-community/src/tools/reddit.ts
```diff
@@ -0,0 +1,123 @@
+import { getEnvironmentVariable } from "@langchain/core/utils/env"; //"../../../../langchain-core/src/utils/env.js";
```
nick-w-nick, line 1: removed unused comment

```suggestion
import { getEnvironmentVariable } from "@langchain/core/utils/env";
```

libs/langchain-community/src/tools/reddit.ts
```diff
@@ -0,0 +1,123 @@
...
+   * @param {string} query The search query to be sent to reddit
+   * @description Function to retrieve posts based on a search query
+   * @returns the search results from using the API wrapper
+   */
+  async _call(query: string): Promise<any> {
+    const apiWrapper = new RedditAPIWrapper({
+      clientId: this.clientId,
+      clientSecret: this.clientSecret,
+      userAgent: this.userAgent,
+    });
```
nick-w-nick, line 90: Can you make the API wrapper client a property on the class so it doesn't need to re-auth every time and can re-use the existing client?

Here's an example of how it was implemented in the Discord tool:
https://github.com/langchain-ai/langchainjs/blob/53feadef511b3b5838327ebe8a8ec2e904b13fcc/libs/langchain-community/src/tools/discord.ts#L63-L84

libs/langchain-community/src/tools/tests/reddit.int.test.ts
```diff
@@ -0,0 +1,28 @@
+import { test, expect } from "@jest/globals";
+//import { Document } from "@langchain/core/documents";
+//import { RedditPostsLoader } from "../web/reddit.js";
```
nick-w-nick, line 3: removed unused comments

```suggestion
```

christian-bromann:
I am going to close this due to inactivity. Our team is currently revisiting the strategy on integration support. I would recommend to consider publishing this integration under your own name and push it on your own GitHub. This allows you to move forward without being dependent on us reviewing it. We will try find ways to highlight community contributions like this and make them discoverable.

PR #7303, opened 2024-12-02
feat(community): add FileChatMessageHistory

sinedied:
## Changes

Added `FileChatMessageHistory` to allow chat history based on a local JSON file. The implementation optionally supports:
- storing and listing sessions for a given user ID
- storing additional context metadata alongside the messages history

The implementation is currently only compatible with Node.js, but the storage backend can be easily extended to support other environments.

The purpose of this module is to have a simple way to store chat histories locally, without having to rely on an external DB, for experimentations, local development, or local-first LLM scenarios.


jacoblee93:
I've moved it to the proper export (`stores/message/file_system`) but otherwise looks great. Thanks for contributing @sinedied!

jacoblee93 (approved):

PR #7307, opened 2024-12-03
fix(azure-cosmosdb): Add caches for Azure CosmosDB vCore [comments resolved]

crisjy:
Fixes # (issue)


jacoblee93 (approved):
Sorry about the delayed review!

jacoblee93:
Actually - would love to tag @sinedied for a second look!

libs/langchain-azure-cosmosdb/src/caches.ts
```diff
@@ -189,3 +195,151 @@ export class AzureCosmosDBNoSQLSemanticCache extends BaseCache {
...
+  private cacheDict: { [key: string]: AzureCosmosDBMongoDBVectorStore } = {};
+
+  private readonly client: MongoClient | undefined;
+
+  private vectorDistanceFunction: string;
+
+  constructor(
+    embeddings: EmbeddingsInterface,
+    dbConfig: AzureCosmosDBMongoDBConfig,
+    similarityScoreThreshold: number = 0.0
```
jacoblee93, line 232: Should we combine this third arg with the second?

sinedied: This is currently implemented as a separate param in the Cosmos NoSQL version, so it would make sense to keep the same here?

I would more be wondering about `0.0` being the default threshold here, could we keep `0.6` like the other one?

libs/langchain-azure-cosmosdb/src/caches.ts
```diff
@@ -189,3 +195,151 @@ export class AzureCosmosDBNoSQLSemanticCache extends BaseCache {
...
+  ) {
+    super();
+
+    const connectionString =
+      dbConfig.connectionString ??
+      getEnvironmentVariable("AZURE_COSMOSDB_MONGODB_CONNECTION_STRING");
+
+    if (!dbConfig.client && !connectionString) {
+      throw new Error(
+        "AzureCosmosDBMongoDBVectorStore client or connection string must be set."
```
sinedied, line 242:
```suggestion
        "AzureCosmosDBMongoDBSemanticCache client or connection string must be set."
```

libs/langchain-azure-cosmosdb/src/caches.ts
```diff
@@ -189,3 +195,151 @@ export class AzureCosmosDBNoSQLSemanticCache extends BaseCache {
...
+      );
+    }
+
+    if (!dbConfig.client) {
+      this.client = new MongoClient(connectionString!, {
+        appName: "langchainjs",
+      });
+    }
+
+    this.config = {
```
sinedied, line 252: It would be best to use a different default collection name here than the one used of the vector store, like `semanticCacheContainer`

libs/langchain-azure-cosmosdb/src/caches.ts
```diff
@@ -189,3 +195,151 @@ export class AzureCosmosDBNoSQLSemanticCache extends BaseCache {
...
+      queryEmbedding,
+      1,
+      this.config.indexOptions?.indexType
+    );
+    if (!results.length) return null;
+
+    const generations = results
+      .flatMap(([document, score]) => {
+        const isSimilar =
+          (this.vectorDistanceFunction === "COS" &&
```
sinedied, line 295: This should be `AzureCosmosDBMongoDBSimilarityType.L2` instead of `"COS"` in both checks here

sinedied (changes requested):
Thanks for your contribution!

Could you restructuring the current file structure to: 
```
- src/
  - caches/
    - nosql.ts
    - mongodb.ts
```

The same could be done for the tests, so you could duplicate the existing test that exist for the NoSQL version to have one for the MongoDB version.

libs/langchain-azure-cosmosdb/src/tests/caches/caches_nosql.init.test.ts
jacoblee93, line 1: I think you meant `.int.test.ts`?

libs/langchain-azure-cosmosdb/src/tests/caches/caches_nosql.init.test.ts
```diff
@@ -0,0 +1,245 @@
+/* eslint-disable no-process-env */
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+import {
+    CosmosClient,
```
jacoblee93, line 5: Remember to run `yarn format`

libs/langchain-azure-cosmosdb/src/tests/caches/caches_nosql.init.test.ts
jacoblee93: Also seems there's some merge conflict - can you make sure we didn't lose a test?

crisjy: done

libs/langchain-azure-cosmosdb/src/caches.ts
crisjy: done

crisjy:
@jacoblee93  @sinedied  comments and conflict resolved, also add test for mongo cache. Could you help review and approve this PR? Thanks a lot

jacoblee93 (approved):
Thank you!

PR #7317, opened 2024-12-04
feat(community): Introduce callbacks to IBM Watsonx SDK with CallbackHandlers

FilipZmijewski:
This PR introduces callbacks that are passed to our Watsonx sdk


libs/langchain-community/src/chat_models/ibm.ts
```diff
@@ -631,19 +630,25 @@ export class ChatWatsonx<
   async *_streamResponseChunks(
     messages: BaseMessage[],
     options: this["ParsedCallOptions"],
-    _runManager?: CallbackManagerForLLMRun
+    _runManager?: WatsonxCallbackManagerForLLMRun
```
jacoblee93, line 633: Don't think you want to type it this way

Are you able to just use this instead?

https://js.langchain.com/docs/how_to/callbacks_custom_events

FilipZmijewski: How would you recommend to type it? We are extending class in this case. I don't think dispatching custom events help much. We just want to pass the callbacks that we create to our SDK, so nothing but passing the callbacks happens on langchain side. The case is that it would be good to still pass it in ```callbacks``` property

FilipZmijewski:
@jacoblee93 We have discussed this case and proceeded with following implementation that will be much simpler. Converted this to draft and created new PR. If you could have a look at it and provide feedback if that is a better case: https://github.com/langchain-ai/langchainjs

PR #7329, opened 2024-12-06
feat(community): Introduce callbacks to IBM Watsonx SDK

FilipZmijewski:
This PR introduces callbacks that are passed to our Watsonx sdk


jacoblee93:
Think it'd be nicer to map your internal events to our custom events to avoid a developer needing to use two different systems

But this is ok, thank you!

jacoblee93 (approved):

jacoblee93:
CI is throwing some errors (possibly due to the version bump?)

```
src/document_compressors/ibm.ts(169,5): error TS2322: Type '({ index: number; relevanceScore: number; input: { text: string; }; } | { index: number; relevanceScore: number; input?: undefined; })[]' is not assignable to type '{ index: number; relevanceScore: number; input?: string | undefined; }[]'.
```

FilipZmijewski:
This is correct, version bump issue which I somehow did not see. Fixed now.

FilipZmijewski:
Also, I am having troubles understanding what "mapping events to out custom events" means, if you could provide any slice of documentation for that maybe I would be able to understand that a little bit more. Thank you!

jacoblee93:
> Also, I am having troubles understanding what "mapping events to out custom events" means, if you could provide any slice of documentation for that maybe I would be able to understand that a little bit more. Thank you!

Pass some custom callbacks into your SDK

Then use `dispatchCustomEventWeb` (or I think you can call `handleCustomEvent` on the callback manager directly):

https://js.langchain.com/docs/how_to/callbacks_custom_events/

User then can hook into it

PR #7330, opened 2024-12-07
feat(core): add registerConfigureHook

ibolmo:
Follows similar pattern in LangChain (Python)

This allows langchain devs to register a callback without having to pass them explicitly in each invocation. 

@ibolmo


langchain-core/src/callbacks/manager.ts
```diff
@@ -1,26 +1,27 @@
 import { v4 as uuidv4 } from "uuid";
 import { AgentAction, AgentFinish } from "../agents.js";
-import type { ChainValues } from "../utils/types/index.js";
+import { getConfigureHooks, getContextVariable } from "../context.js";
```
ibolmo, line 3: I can revert these changes. My editor auto formats the imports.

langchain-core/src/callbacks/manager.ts
```diff
@@ -1252,6 +1253,32 @@ export class CallbackManager
...
+      const createOne =
+        envVar && getEnvironmentVariable(envVar) === "true" && handlerClass;
+      if (getContextVariable(contextVar) || createOne) {
+        const varHandler = ensureHandler(
+          (getContextVariable(contextVar) as
+            | BaseCallbackHandler
+            | CallbackHandlerMethods
+            | undefined) || handlerClass!
+        );
+        if (
```
ibolmo, line 1272: Slightly changed from the original implementation. I'll

langchain-core/src/context.ts
```diff
@@ -129,3 +133,25 @@ export function getContextVariable<T = any>(name: PropertyKey): T | undefined {
...
+  if (config.envVar && !config.handlerClass) {
+    throw new Error(
+      "If envVar is set, handlerClass must also be set to a non-None value."
+    );
+  }
+
+  setContextVariable(LC_CONFIGURE_HOOKS_KEY, [...getConfigureHooks(), config]);
+};
+
+type ConfigureHook = {
```
jacoblee93, line 152: Let's export this for portability

langchain-core/src/context.ts
```diff
@@ -129,3 +133,25 @@ export function getContextVariable<T = any>(name: PropertyKey): T | undefined {
...
+  contextVar: string;
+  inheritable: boolean;
+  handlerClass?: BaseCallbackHandler | CallbackHandlerMethods;
```
jacoblee93, line 155: I don't think we want to accept a raw methods object here because then we can't dedupe by `name` later?

langchain-core/src/callbacks/manager.ts
```diff
@@ -1252,6 +1254,32 @@ export class CallbackManager
...
+      handlerClass,
+      envVar,
+    } of getConfigureHooks()) {
+      const createOne =
+        envVar && getEnvironmentVariable(envVar) === "true" && handlerClass;
+      if (getContextVariable(contextVar) || createOne) {
+        const varHandler = ensureHandler(
+          (getContextVariable(contextVar) as
+            | BaseCallbackHandler
+            | CallbackHandlerMethods
```
jacoblee93, line 1270: See below, let's just accept `BaseCallbackHandler`

langchain-core/src/callbacks/manager.ts
```diff
@@ -1252,6 +1254,32 @@ export class CallbackManager
...
+            | undefined) || handlerClass!
```
jacoblee93, line 1271: nit: use nullish coalescing for convention

langchain-core/src/callbacks/manager.ts
```diff
@@ -1252,6 +1254,32 @@ export class CallbackManager
...
+      contextVar,
+      inheritable,
+      handlerClass,
+      envVar,
+    } of getConfigureHooks()) {
+      const createOne =
+        envVar && getEnvironmentVariable(envVar) === "true" && handlerClass;
+      if (getContextVariable(contextVar) || createOne) {
+        const varHandler = ensureHandler(
+          (getContextVariable(contextVar) as
```
jacoblee93, line 1268: nit: Would prefer a type guard here for `BaseCallbackHandler ` instead of a cast

jacoblee93 (commented):
Thanks for the PR - looks good overall!

See comments for some nits, can apply myself later too.

langchain-core/src/context.ts
```diff
@@ -129,3 +133,25 @@ export function getContextVariable<T = any>(name: PropertyKey): T | undefined {
   const runTree = AsyncLocalStorageProviderSingleton.getInstance().getStore();
   return runTree?.[_CONTEXT_VARIABLES_KEY]?.[name];
 }
+
+const LC_CONFIGURE_HOOKS_KEY = Symbol("lc:configure_hooks");
+
+export const getConfigureHooks = () =>
+  getContextVariable<ConfigureHook[]>(LC_CONFIGURE_HOOKS_KEY) || [];
+
+export const registerConfigureHook = (config: ConfigureHook) => {
```
jacoblee93, line 142: Short docstring with a usage example would be nice if you're up for it

jacoblee93:
Ah ok one more thing I realized - we need to be a little bit more clever so we don't pull in `AsyncLocalStorage` automatically for `callbacks/manager` and break web environments

jacoblee93:
Need to fix a circular dep issue, will aim to get this in tomorrow or midweek

jacoblee93:
May want to add a docs page later outside of API refs

jacoblee93:
Thanks again!

PR #7357, opened 2024-12-12
fix(community): file system chat history

sinedied:
## Changes

- Fixes `getAllSessions()` method not return `id` property
- Fixes doc


jacoblee93 (approved):
Thanks!

PR #7360, opened 2024-12-13
MariaDB vector store implementation.

rusher:
This is the MariaDB vectorstore implementation. see https://mariadb.org/projects/mariadb-vector/

This PR contains a proposal to unify filtering data. I might have not understood stuctured query well, but goal of this framework is to permit unifying behavior and it doesn't feel this part has succeed in it. 

I'll detail what i've submitted, i'll let you judge if accurate or not.  
There is now a new package "@langchain/core/filter"

A builder permit to create the filter : 
```Typescript
const b = new FilterExpressionBuilder();

// a = 2015
let exp = b.eq("a", 2015); 

// a IN [2015,2018]
exp = b.in("a", [2015, 2018]); 

// name = 'martin' AND (firstname = 'john' OR firstname = 'jack
exp = b.and(
            b.eq("name", "martin"),
            b.group(b.or(b.eq("firstname", "john"), b.eq("firstname", "jack")))
        ); 

// the opposite : just add b.not(xx), and converter will transform query
// name != 'martin' OR (firstname != 'john' AND firstname != 'jack')
exp = b.not(
          b.and(
            b.eq("name", "martin"),
            b.group(b.or(b.eq("firstname", "john"), b.eq("firstname", "jack")))
          )
        ); 
```
(Ideally, a text parser using antlr to generate the expression from a standardized query would next the next step)

When requiring the corresponding SQL to a filter expression, each vector store needs to implement a filter implementation, extending BaseFilterExpressionConverter. This prevent SQL injection, unify behavior, and permit any kind of filtering, like IN, NOT, gouping ...

(a database filtering implementation must be probably maximum 20 significant lines of code, mainly because each database use his specific JSON functions)


rusher:
Can i still help to get this PR merged ?

langchain-core/src/filter.ts
jacoblee93, line 1: This is interesting but too broadly scoped

Let's put it in a separate PR

libs/langchain-community/langchain.config.js
```diff
@@ -134,6 +134,7 @@ export const config = {
     "vectorstores/hanavector": "vectorstores/hanavector",
     "vectorstores/lancedb": "vectorstores/lancedb",
     "vectorstores/libsql": "vectorstores/libsql",
+    "vectorstores/mariadb": "vectorstores/mariadb",
```
jacoblee93, line 137: You will also need to add this to `requiresOptionalDependencies` below

libs/langchain-community/src/vectorstores/mariadb.ts
```diff
@@ -0,0 +1,760 @@
...
+
+    // Either all documents have ids or none of them do to avoid confusion.
+    if (ids !== undefined && ids.length !== vectors.length) {
+      throw new Error(
+        "The number of ids must match the number of vectors provided."
+      );
+    }
+    await this.loadCollectionId();
+
+    const insertQuery = `INSERT INTO ${this.computedTableName}(${
```
jacoblee93, line 557: nit: can we escape column names too/disallow nonalphanumeric?

jacoblee93 (commented):
Really sorry about the delay - let's move the core work you've done into `community` for now and make it MariaDB specific

I think it is interesting but there are a lot of filters to generalize across

libs/langchain-community/src/vectorstores/mariadb.ts
rusher: table name and columns names are all escaped to prevent any injection. see `escapeId` function

libs/langchain-community/langchain.config.js
rusher: right, adding

langchain-core/src/filter.ts
rusher: i would like too, but i don't now how to makes mariadb implementation without this filter other leed to injection.

Maybe a version like Dictionnaly in python ( like filter={ 'id': {'$in': [1, 5, 2, 9]} } ). It doesn't seem appropriate too.

what is the solution ? makes a different PR for the filter part, and wait for it for now ?

rusher:
(i've add the missing requiresOptionalDependencies and rebased to current version)

langchain-core/src/filter.ts
jacoblee93: Let's just move it to `langchain-community/utils/mariadb.ts` for now

jacoblee93:
Sorry for the delay again, let's just move the core changes to community

Generally speaking we don't allow core changes and community changes in the same PR, and are much more selective about what goes in core

langchain-core/src/filter.ts
rusher: just to be sure we're on the same page: are you going to move this yourself in langchain-community/utils/mariadb.ts or should I move the filter part in mariadb.ts?

rusher: @jacoblee93 ?

robertsilen:
Hey @jacoblee93 , any update on this? :) Cheers, Robert

jacoblee93:
Hey @robertsilen, see my comment above - we don't add core concepts in the same PR as `@langchain/community`

rusher:
submit correction without filter part. There is now only the mariadb vector store

jacoblee93:
Hey sorry will look today

jacoblee93 (approved):
Thank you all for your patience! Looks great

rusher:
ok, thanks to work on this

PR #7361, opened 2024-12-13
fix(google-gauth): Remove logging of error responses

jarib:
It should be up to the user if they want to log error responses or not. Prompts logged could contain sensitive information.


jacoblee93 (approved):
Thanks! Will add this to `gauth-web` as well.

PR #7366, opened 2024-12-15
fix(core): Make sure to return a callback manager if configure hooks

ibolmo:
Set a campaign manager, if we know there are callback managers we'd like to use

This preserves existing behavior except for when library creators want handlers registered with the `registerConfigureHook` function. In that case, we want to make sure we return a callback manager.

Twitter: @ibolmo

### DIFF
--- a/langchain-core/src/callbacks/manager.ts
+++ b/langchain-core/src/callbacks/manager.ts
@@ -1,31 +1,31 @@
 import { v4 as uuidv4 } from "uuid";
 import { AgentAction, AgentFinish } from "../agents.js";
-import type { ChainValues } from "../utils/types/index.js";
-import { LLMResult } from "../outputs.js";
-import {
-  BaseCallbackHandler,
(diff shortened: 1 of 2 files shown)


langchain-core/src/callbacks/manager.ts
```diff
@@ -1276,8 +1276,12 @@ export class CallbackManager
         handler = new (handlerClass as any)({});
       }
       if (handler !== undefined) {
-        if (!callbackManager?.handlers.some((h) => h.name === handler!.name)) {
-          callbackManager?.addHandler(handler, inheritable);
+        if (!callbackManager) {
```
jacoblee93, line 1279: Since this can potentially create a callback manager - let's move this whole block above where we add the `inheritableTags` and `inheritableMetadata` so that those are added properly

jacoblee93 (commented):
Small nit, can apply myself if you don't have time

PR #7373, opened 2024-12-16
docs: updates text-to-html example

philnash:
The `text-to-html` transformer example was using the `CheerioWebBaseLoader` to fetch HTML, but the `CheerioWebBaseLoader` already strips HTML using Cheerio's `.text()` function. So the `html-to-text` transformer wasn't really doing anything.

This commit updates the example to use the `HTMLWebBaseLoader` instead, to give the `html-to-text` transformer something to do in its own example.


jacoblee93 (approved):
Thank you for following up here!

PR #7374, opened 2024-12-16
community[minor]: Remove Dria retriever

erhant:
[Dria](https://dria.co/) has been previously added with https://github.com/langchain-ai/langchainjs/pull/4302

While Dria is an active project, it is no longer doing retriever services, and shall be removed from LangChain as such.


jacoblee93:
Thanks for cleaning up

Is the service completely gone now? Or would it be better to just remove from docs and add a warning?

efriis:
hey can we all chat on https://github.com/langchain-ai/langchain/pull/28687 ?

it's unusual to ask to remove docs from langchain for a service that still documents the integration in their own docs, especially because this langchain integration is still documented in the dria docs https://dria.co/docs/integrations/langchain

christian-bromann:
Closing in favor of https://github.com/langchain-ai/langchainjs/pull/8896

PR #7400, opened 2024-12-18
fix(google-common): add tool_use param when using structured outputs

kashifmin:
When using `.withStructuredOutput` we need to force tool use. If not, the model tends produce an empty response and the chain eventually returns `undefined`.
1. Modified tool_use param in gemini utils to handle structured output
2. withStructuredOutput will now bind tool_use parameter
3. anthropic utils already handles the tool use param by setting the tool use object to the function name if provided


kashifmin:
@afirstenberg Please take a look when possible. Currently structured output breaks intermittently since tool_choice param is not set correctly.

jacoblee93 (approved):
Ah thank you for this - this is a good idea

PR #7406, opened 2024-12-19
fix(google-common, google-vertexai): Invalid URL for embeddings model

### LINKED ISSUE #7390: VertexAIEmbeddings don't work with europe-west9 location
### Checked other resources

### Example Code

``` typescript
  const embedder = new VertexAIEmbeddings({
    model: 'text-multilingual-embedding-002',
    location: 'europe-west9',
    authOptions: {
      keyFilename: config.vector_store_key_filename,
      scopes: [config.vector_store_scopes],
    },
  });
```

GaxiosError: Publisher Model `projects/${projectId}/locations/europe-west9/publishers/unknown/models/text-multilingual-embedding-002` not found.

### Description

After updating when I use europe-west8 or  europe-west9 I have the GaxiosError with unknown publisher. But it work with us-central1 or europe-west1.

### System Info

node : v21.7.2
MacOS : 15.1.1

afirstenberg:
The URL path for Vertex AI typically includes a publisher name.
For LLMs, this publisher name is determined based on model name.
For embeddings, the publisher name never matched, so was assigned a name of "unknown".

Oddly enough, in some regions and with some embedding models - this was fine. But in other regions, this returned an error.

Fix is to make sure that all embedding models (at least for now) are considered published by Google.

Fixes #7390

### DIFF
--- a/libs/langchain-google-common/src/embeddings.ts
+++ b/libs/langchain-google-common/src/embeddings.ts
@@ -38,6 +38,11 @@ class EmbeddingsConnection<
     return "predict";
   }
 
+  get modelPublisher(): string {
+    // All the embedding models are currently published by "google"
+    return "google";
+  }
+
   async formatData(
     input: GoogleEmbeddingsInstance[],
     parameters: GoogleAIModelRequestParams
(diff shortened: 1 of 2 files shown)


jacoblee93 (approved):
Thank you very much!

PR #7407, opened 2024-12-19
feat(langchain-core): Add support for `boolean` metadata attributes in `FunctionalTranslator`

nick-w-nick:
While adding support for `boolean` as a type in the `castValue` function to allow them to be used in metadata filters that are created within a `StructuredQuery`, I noticed that this would cause the `FunctionalTranslator` implementation to have a gap in functionality, where booleans were technically supported as filter inputs, but would not be handled correctly during the query translation without modifications.

To remedy this, I've added a new utility method called `getAllowedComparatorsForType` which returns the allowed comparators that can be used with a given type. This enables guards around using incorrect combinations of values and comparators.

Another added benefit of this utility is that we can easily get the list of comparators for each expected type, making it possible to generate tests for every comparator for each data type automatically. I've gone ahead and added an MVP set of tests for the `FunctionalTranslator`, limited to the `getAllowedComparatorsForType` and `visitComparison` methods, to not overcrowd the PR with unrelated changes.

### DIFF
--- a/langchain-core/src/structured_query/base.ts
+++ b/langchain-core/src/structured_query/base.ts
@@ -119,7 +119,7 @@ export class BasicTranslator<
     } else {
       throw new Error("Unknown comparator or operator");
     }
-    return `$${func}`;
+    return `${func}`;
   }
 
   /**
--- a/langchain-core/src/structured_query/base.ts
+++ b/langchain-core/src/structured_query/base.ts
@@ -119,7 +119,7 @@ export class BasicTranslator<
     } else {
       throw new Error("Unknown comparator or operator");
     }
-    return `${func}`;
+    return `$${func}`;
   }
 
   /**
--- a/langchain-core/src/structured_query/base.ts
+++ b/langchain-core/src/structured_query/base.ts
@@ -100,7 +100,7 @@ export class BasicTranslator<
         this.allowedComparators.indexOf(func as Comparator) === -1
       ) {
         throw new Error(
-          `Comparator ${func} not allowed. Allowed operators: ${this.allowedComparators.join(
+          `Comparator ${func} not allowed. Allowed comparators: ${this.allowedComparators.join(
             ", "
           )}`
         );
--- a/langchain-core/src/structured_query/tests/functional.test.ts
+++ b/langchain-core/src/structured_query/tests/functional.test.ts
@@ -1,0 +1,303 @@
+import { Document } from "../../documents/document.js";
+import { FunctionalTranslator } from "../functional.js";
+import { Comparators, Visitor } from "../ir.js";
+
+describe("FunctionalTranslator", () => {
+  const translator = new FunctionalTranslator();
+
+  describe("getAllowedComparatorsForType", () => {
+    test("string", () => {
+      expect(translator.getAllowedComparatorsForType("string")).toEqual([
+        Comparators.eq,
+        Comparators.ne,
+        Comparators.gt,
+        Comparators.gte,
+        Comparators.lt,
+        Comparators.lte,
+      ]);
+    });
+    test("number", () => {
+      expect(translator.getAllowedComparatorsForType("number")).toEqual([
+        Comparators.eq,
+        Comparators.ne,
+        Comparators.gt,
+        Comparators.gte,
+        Comparators.lt,
+        Comparators.lte,
+      ]);
+    });
+    test("boolean", () => {
+      expect(translator.getAllowedComparatorsForType("boolean")).toEqual([
+        Comparators.eq,
+        Comparators.ne,
+      ]);
+    });
+    test("unsupported", () => {
+      expect(() =>
+        translator.getAllowedComparatorsForType("unsupported")
+      ).toThrow("Unsupported data type: unsupported");
+    });
+  });
+
+  describe("visitComparison", () => {
+    const attributeByType = {
+      string: "stringValue",
+      number: "numberValue",
+      boolean: "booleanValue",
+    };
+    
+    describe("returns true or false for valid comparisons", () => {
+      const inputValues = {
+        string: "value",
+        number: 1,
+        boolean: true,
+      };
+      
+      const validDocumentsByComparator: { [key in string]: Document<Record<string, any>>[] } = {
+        [Comparators.eq]: [
+          new Document({
+            pageContent: "",
+            metadata: {
+              stringValue: "value",
+              numberValue: 1,
+              booleanValue: true,
+            },
+          }),
+        ],
+        [Comparators.ne]: [
+          new Document({
+            pageContent: "",
+            metadata: {
+              stringValue: "not-value",
+              numberValue: 0,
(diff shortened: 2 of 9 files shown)


langchain-core/src/structured_query/base.ts
```diff
@@ -100,7 +100,7 @@ export class BasicTranslator<
         this.allowedComparators.indexOf(func as Comparator) === -1
       ) {
         throw new Error(
-          `Comparator ${func} not allowed. Allowed operators: ${this.allowedComparators.join(
+          `Comparator ${func} not allowed. Allowed comparators: ${this.allowedComparators.join(
```
nick-w-nick, line 103: Noticed a random typo, thought I'd fix it while in the area.

langchain/src/chains/query_constructor/tests/query_chain.int.test.ts
```diff
@@ -33,6 +33,10 @@ test("Query Chain Test", async () => {
       new Comparison(Comparators.lt, "length", 90),
     ])
   );
+  const sq6 = new StructuredQuery(
+    "",
+    new Comparison(Comparators.eq, "isReleased", true)
+  );
```
nick-w-nick, line 39: This isn't _technically_ a required test addition since I didn't modify the actual Comparison logic, but since I added support for `boolean` types in `Comparison`, I thought it was worthwhile to include it here just for posterity.

langchain/src/chains/query_constructor/tests/query_parser.test.ts
```diff
@@ -17,6 +17,7 @@ const correctQuery = new StructuredQuery(
     ]),
     new Comparison(Comparators.lt, "length", 180),
     new Comparison(Comparators.eq, "genre", "pop"),
+    new Comparison(Comparators.eq, "hasLyrics", true),
```
nick-w-nick, line 20: Same as [this comment](https://github.com/langchain-ai/langchainjs/pull/7407/files#r1891285697)

langchain-core/src/structured_query/tests/functional.test.ts
nick-w-nick, line 1: As mentioned in the PR description, I only added the bare minimum tests needed to ensure the modification to the existing functionality did not break the `visitComparison` logic.

nick-w-nick, line 1: As mentioned in the PR description, I only added the bare minimum tests needed to ensure the modification to the existing functionality did not break the `visitComparison` logic.

langchain-core/src/structured_query/tests/functional.test.ts
```diff
@@ -0,0 +1,310 @@
+import { test, expect } from "@jest/globals";
+import { Document } from "../../documents/document.js";
+import { FunctionalTranslator } from "../functional.js";
+import { Comparators, Visitor } from "../ir.js";
+
+describe("FunctionalTranslator", () => {
+  const translator = new FunctionalTranslator();
+
+  describe("getAllowedComparatorsForType", () => {
```
nick-w-nick, line 9: As mentioned in the PR description, I only added the bare minimum tests needed to ensure the modification to the existing functionality did not break the `visitComparison` logic.

langchain-core/src/structured_query/tests/functional.test.ts
```diff
@@ -0,0 +1,310 @@
...
+        ],
+      };
+      
+      const stringComparators = translator.getAllowedComparatorsForType("string");
+      for (const comparator of stringComparators) {
+        const attribute = attributesByType.string;
+        const value = inputValuesByAttribute[attribute];
+        const validDocuments = validDocumentsByComparator[comparator];
+        for (const validDocument of validDocuments) {
+          test(`${value} -> ${comparator} -> ${validDocument.metadata[attribute]}`, () => {
```
nick-w-nick, line 211: When a test fails, the generated test name looks like `input_value -> comparator -> document_value`, like shown in the below screenshot for a failed `eq` comparator check, where the input value is `value` and the document value is `invalid-value`.

image

langchain-core/src/structured_query/tests/functional.test.ts
```diff
@@ -0,0 +1,246 @@
...
+      };
+      
+      function generateComparatorTestsForType(type: "string" | "number" | "boolean") {
+        const comparators = translator.getAllowedComparatorsForType(type);
+        for (const comparator of comparators) {
+          const attribute = attributesByType[type];
+          const value = inputValuesByAttribute[attribute];
+          const validDocuments = validDocumentsByComparator[comparator];
+          for (const validDocument of validDocuments) {
+            test(`${value} -> ${comparator} -> ${validDocument.metadata[attribute]}`, () => {
```
nick-w-nick, line 212: When a test fails, the generated test name looks like `input_value -> comparator -> document_value`, like shown in the below screenshot for a failed `eq` comparator check, where the input value is `value` and the document value is `invalid-value`.

image

langchain-core/src/structured_query/base.ts
jacoblee93: Thanks!

jacoblee93 (approved):
Looks reasonable to me!

jacoblee93:
@nick-w-nick seems like it's breaking some types in `community`:

```
Error: src/structured_query/supabase.ts(225,7): error TS2345: Argument of type 'string | number | boolean' is not assignable to parameter of type 'string | number'.
  Type 'boolean' is not assignable to type 'string | number'.
Error: src/structured_query/supabase.ts(243,44): error TS2345: Argument of type 'string | number | boolean' is not assignable to parameter of type 'string | number'.
Error: src/structured_query/vectara.ts(122,7): error TS2345: Argument of type 'string | number | boolean' is not assignable to parameter of type 'Value'.
```

Any way we can make this backwards compatible?

nick-w-nick:
@jacoblee93 The type issues have been resolved, could you re-run the CI? Thanks!

langchain/src/chains/query_constructor/parser.ts
```diff
@@ -94,7 +94,7 @@ export class QueryTransformer {
               return new Comparison(
                 funcName as Comparator,
                 traverse(node.args[0]) as string,
-                traverse(node.args[1]) as string | number
+                traverse(node.args[1]) as string | number | boolean
```
jacoblee93, line 97: If this change is required, will this will mean that `langchain` will no longer support older versions of core without your other changes?

Is there a way we can do this while only changing core?

nick-w-nick: After looking into this, `TraverseType` already expects `boolean`, it's just that the typecasting on `traverse(node.args[1])` is forcibly setting it to expect `string | number`.

To allow backward compatibility, I've reverted the typecasting to be what it was originally, while still allowing `boolean` return values from `traverse`.

jacoblee93:
Thank you for your patience! Will do a release today

Issue #7365, opened 2024-12-14
DOC: <Issue related to docs/integrations/llms/llama_cpp.mdx>


Jacky3003:
### Checklist

- [X] I added a very descriptive title to this issue.
- [X] I included a link to the documentation page I am referring to (if applicable).

### Issue with current documentation:

With the new update to `node-llama-cpp` being bumped up to version 3, the integration docs for its coresponding LLM component in LangChain should use some updates in the numbering and model being used in the examples in order to avoid confusion. The following docs as a result should be updated:
- https://js.langchain.com/docs/integrations/llms/llama_cpp/#guide-to-installing-llama2
- https://js.langchain.com/docs/integrations/chat/llama_cpp/
- https://js.langchain.com/docs/integrations/text_embedding/llama_cpp/

### Idea or request for content:

It would be nice for the version numbering to be changed from 2 to 3. For the LLM's documentation specifically, it should be investigated if the guide to installing a llama 2 model (which should now be llama 3 in this instance) should be updated as well. I plan to work on this issue.

PR #7426, opened 2024-12-24
fix(docs): Fix outdated documentation for llama cpp

Jacky3003:
Fixes [#7365](https://github.com/langchain-ai/langchainjs/issues/7365)

The updated documentation also changes the tutorial for installing llama 2 based on the changes needed to install llama 3 instead, based on updates made to the following two repos:
- https://github.com/meta-llama/llama-models
- https://github.com/ggerganov/llama.cpp


jacoblee93 (approved):
Thank you so much!

PR #7428, opened 2024-12-25
fix(aws): fix error message output

hujiulong:
When executing this logic, an error occurs:
```
TypeError: Iterator value text is not an entry object
```

The code here should use `Object.entries()` instead of `Object.values()`


jacoblee93 (approved):
Oof, thanks!

PR #7436, opened 2024-12-27
feat(azure-cosmosdb): add session context for a user mongodb

crisjy:
Fixes # (issue)


jacoblee93 (commented):
Seems reasonable to me!

@sinedied do you mind giving this a look?

libs/langchain-azure-cosmosdb/src/chat_histories/mongodb.ts
```diff
@@ -152,4 +161,51 @@ export class AzureCosmosDBMongoChatMessageHistory extends BaseListChatMessageHis
...
+    try {
+      await this.collection.updateOne(
+        { [ID_KEY]: this.sessionId },
+        {
+          $set: { context },
+        },
+        { upsert: true }
+      );
+    } catch (error) {
+      console.log("Error setting context", error);
```
sinedied, line 208: Would be better to throw so these errors can be identified

libs/langchain-azure-cosmosdb/src/chat_histories/mongodb.ts
```diff
@@ -152,4 +161,51 @@ export class AzureCosmosDBMongoChatMessageHistory extends BaseListChatMessageHis
...
+
+    return chatSessions;
+  }
+
+  async clearAllSessions() {
+    await this.initialize();
+    try {
+      await this.collection.deleteMany({});
+    } catch (error) {
+      console.log("Error clearing sessions:", error);
```
sinedied, line 182: Would be better to throw so these errors can be identified

libs/langchain-azure-cosmosdb/src/chat_histories/mongodb.ts
```diff
@@ -152,4 +161,51 @@ export class AzureCosmosDBMongoChatMessageHistory extends BaseListChatMessageHis
...
+      context: doc.context || {},
+    }));
+
+    return chatSessions;
+  }
+
+  async clearAllSessions() {
+    await this.initialize();
+    try {
+      await this.collection.deleteMany({});
```
sinedied, line 180: This would delete all sessions for all users, which might not be the intended behavior... I think you missed the user handling here

sinedied (changes requested):
Thanks @crisjy !
The context handling looks fine here, but you're missing user handling, which is needed to list/delete sessions otherwise you have no distinction of users.

What we did in the other CosmosDB implementation is make `userId` parameter optional in the DB options and set it to `anonymous` by default to keep compatibility with other implementation (or when user handling is not needed).

jacoblee93:
@crisjy let me know when this is ready for another look!

libs/langchain-azure-cosmosdb/src/chat_histories/mongodb.ts
crisjy: throw errors for this

libs/langchain-azure-cosmosdb/src/chat_histories/mongodb.ts
crisjy: yep, add user id as handle for sessions, thank you for your comment!

crisjy:
> @crisjy let me know when this is ready for another look!

@jacoblee93 @sinedied resolve comments for this pr, sorry for delay, could you help me to have a look? thank you!

libs/langchain-azure-cosmosdb/src/chat_histories/mongodb.ts
```diff
@@ -150,6 +165,60 @@ export class AzureCosmosDBMongoChatMessageHistory extends BaseListChatMessageHis
...
-    await this.collection.deleteOne({ [ID_KEY]: this.sessionId });
+    await this.collection.deleteOne({
+      [ID_KEY]: this.sessionId,
+      [ID_USER]: this.userId,
+    });
+  }
+
+  async getAllSessions(): Promise<ChatSessionMongo[]> {
+    await this.initialize();
+    const documents = await this.collection.find().toArray();
```
sinedied, line 176: Missing user filter here

libs/langchain-azure-cosmosdb/src/chat_histories/mongodb.ts
sinedied: I think you forgot to update this call to filter it for the current user only

sinedied (changes requested):
Thanks, the update with userId support looks great!

Though you forgot to add the filter for the current userId on the getAllSessions() and clearAllSessions() methods, as it should apply only to the current users. Should be a quick fix :)

libs/langchain-azure-cosmosdb/src/chat_histories/mongodb.ts
crisjy: add user id as handler to clear sessions

crisjy:
@sinedied @jacoblee93  comments resolved, could you help me have another look? Thank you!

sinedied (approved):

sinedied:
LGTM, thanks!

crisjy:
@jacoblee93 Hi jacob, could you help me have another look at this PR, thanks!

libs/langchain-azure-cosmosdb/src/chat_histories/mongodb.ts
```diff
@@ -150,6 +165,64 @@ export class AzureCosmosDBMongoChatMessageHistory extends BaseListChatMessageHis
...
+    this.context = document?.context || this.context;
+    return this.context;
+  }
+
+  async setContext(context: Record<string, unknown>): Promise<void> {
+    await this.initialize();
+
+    try {
+      await this.collection.updateOne(
+        { [ID_KEY]: this.sessionId },
```
jacoblee93, line 217: Do we need `[ID_USER]: this.userId,` here too?

I assume this is sufficient in this case to just pass id?

jacoblee93 (approved):
Thank you!

PR #7438, opened 2024-12-27
feat(community): add mmr search to pgvector

anadi45:
Adds new features to pgvector class
1. Add flag to include vectors in similarity search
2. `maxMarginalRelevanceSearch` implementation


anadi45:
@jacoblee93 Please take a look when possible.

libs/langchain-community/src/vectorstores/pgvector.ts
```diff
@@ -609,12 +613,14 @@ export class PGVectorStore extends VectorStore {
...
    * @param filter - Optional filter to apply to the search.
+   * @param includeEmbedding Whether to include the embedding vectors in the results.
    * @returns Promise that resolves with an array of tuples, each containing a `Document` and its similarity score.
    */
   async similaritySearchVectorWithScore(
     query: number[],
     k: number,
-    filter?: this["FilterType"]
+    filter?: this["FilterType"],
+    includeEmbedding?: boolean
```
jacoblee93, line 623: Changing the default signature (even adding params) makes migrations a lot trickier - can we factor this out into a private method?

Sort of like we did with Supabase:

https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-community/src/vectorstores/supabase.ts#L280

jacoblee93 (commented):
Thanks! One small comment but mostly looks good

libs/langchain-community/src/vectorstores/pgvector.ts
anadi45: @jacoblee93 thanks for the review. Have made the suggested changes.

jacoblee93 (approved):
Thank you! Just missing a test case but I can add one.

PR #7448, opened 2024-12-30
feat(google-common): Google Search grounding tool formatter

afirstenberg:
Based on results provided by the Google Search Grounding tool, format the result text.

There is a `BaseGoogleSearchOutputParser` which is abstract and which handles a lot of the work.

Two specific output formatters are provided:
- `SimpleGoogleSearchOutputParser` which is just a simple implementation
- `MarkdownGoogleSearchOutputParser` which returns (mostly) markdown that matches how AI Studio tends to format search grounded results.


libs/langchain-google-common/src/output_parsers.ts
```diff
@@ -0,0 +1,261 @@
...
+    });
+    return ret;
+  }
+}
+
+export class MarkdownGoogleSearchOutputParser extends BaseGoogleSearchOutputParser {
+  // FIXME: What should this be?
+  lc_namespace: string[] = ["google_common", "output_parsers"];
+
+  segmentPrefix(
```
jacoblee93, line 209: Let's make all these methods `protected` or `private`

libs/langchain-google-common/src/output_parsers.ts
```diff
@@ -0,0 +1,261 @@
...
+      return Promise.resolve(text);
+    }
+
+    return Promise.resolve(this.annotateText(text, grounding));
+  }
+}
+
+export class SimpleGoogleSearchOutputParser extends BaseGoogleSearchOutputParser {
+  // FIXME: What should this be?
+  lc_namespace: string[] = ["google_common", "output_parsers"];
```
jacoblee93, line 167: This is fine, won't serialize these anyway

Just put it on the base abstract class

libs/langchain-google-common/src/output_parsers.ts
```diff
@@ -0,0 +1,261 @@
...
+  }
+
+  annotateText(text: string, grounding: GroundingInfo): string {
+    const prefix = this.textPrefix(text, grounding) ?? "";
+    const suffix = this.textSuffix(text, grounding) ?? "";
+    const body = this.annotateTextSegments(text, grounding);
+    return `${prefix}${body}${suffix}`;
+  }
+
+  parseResult(
```
jacoblee93, line 150: You can just make this method `async` so you don't need the `Promise.resolve`s below

jacoblee93 (commented):
Some small nits but otherwise looks nice!

Is this the accepted way to show sources?

afirstenberg:
> Some small nits but otherwise looks nice!

Nits addressed.

> Is this the commonly accepted way to show sources vs. just having the developer do it themselves?

The markdown one mimics what AI Studio provides, but the base class makes it pretty flexible so people can create their own formats if necessary (such as a pure HTML one).

libs/langchain-google-common/src/output_parsers.ts
```diff
@@ -0,0 +1,266 @@
...
+
+type GroundingInfo = {
+  metadata: GeminiGroundingMetadata;
+  supports: GeminiGroundingSupport[];
+};
+
+export abstract class BaseGoogleSearchOutputParser extends BaseLLMOutputParser<string> {
+  lc_namespace: string[] = ["google_common", "output_parsers"];
+
+  generationToGroundingInfo(
```
jacoblee93, line 21: Should we make all these protected too?

Makes e.g. API refs hard to navigate if the user isn't intended to call them directly

afirstenberg: I'm fine with all of them protected. I tended to think to make them open if people did want to use them natively, but I can't imagine that being important.

I'll take care of it later if you don't get to it.

jacoblee93 (approved):

jacoblee93:
Thanks!

PR #7471, opened 2025-01-06
fix(google-common): Search grounding formatting

afirstenberg:
Problem: 
If using a subclass of `BaseGoogleSearchOutputParser` where the input doesn't have grounding info, it attempts to dereference undefined.

Solution:
Return the input text, unchanged, if there is no grounding info to annotate.

Fix and tests for same.

### DIFF
--- a/libs/langchain-google-common/src/output_parsers.ts
+++ b/libs/langchain-google-common/src/output_parsers.ts
@@ -144,7 +144,7 @@ export abstract class BaseGoogleSearchOutputParser extends BaseLLMOutputParser<s
    * @param grounding
    */
   protected searchSuggestion(grounding: GroundingInfo): string {
-    return grounding.metadata.searchEntryPoint?.renderedContent ?? "";
+    return grounding?.metadata?.searchEntryPoint?.renderedContent ?? "";
   }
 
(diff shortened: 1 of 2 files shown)


libs/langchain-google-common/src/output_parsers.ts
```diff
@@ -198,7 +198,7 @@ export class SimpleGoogleSearchOutputParser extends BaseGoogleSearchOutputParser
 
   protected textSuffix(_text: string, grounding: GroundingInfo): string {
     let ret = "\n";
-    const chunks: GeminiGroundingChunk[] = grounding.metadata.groundingChunks;
+    const chunks: GeminiGroundingChunk[] = grounding?.metadata?.groundingChunks ??[];
```
jacoblee93, line 201: nit: run `yarn format`

jacoblee93 (commented):
Agh thanks. Remember to run `yarn format`!

libs/langchain-google-common/src/output_parsers.ts
afirstenberg: Doh! Fixed

jacoblee93:
Thanks!

PR #7472, opened 2025-01-06
feat(google-common): Gemini logprobs

afirstenberg:
Add support for OpenAI compatible logprobs, including:
- logprobs boolean model parameter
- topLogprobs numeric model parameter
- logprobs returned in `response_metadata` in the same format OpenAI uses
- updated documentation to indicate availability

Add support for token selection updates in model parameters:
- presencePenalty
- frequencyPenalty

Added tests for same.
- Note that the tests for logprobs are essential since Google is highly restrictive on its use.
- Note that the penalty test is somewhat non-deterministic


libs/langchain-google-common/src/chat_models.ts
```diff
@@ -189,8 +189,16 @@ export abstract class ChatGoogleBase<AuthOptions>
 
   topK = 40;
 
+  presencePenalty = undefined;
```
jacoblee93, line 192: Don't need to init as `undefined` - should have their actual typing

libs/langchain-google-common/src/chat_models.ts
```diff
@@ -189,8 +189,16 @@ export abstract class ChatGoogleBase<AuthOptions>
...
+
+  frequencyPenalty = undefined;
+
   stopSequences: string[] = [];
 
+  logprobs: false;
```
jacoblee93, line 198: Did you mean to set `logprobs = false`?

and `topLogprobs = 0`?

libs/langchain-google-common/src/utils/gemini.ts
```diff
@@ -674,6 +678,52 @@ export function getGeminiAPI(config?: GeminiAPIConfig): GoogleAIAPI {
...
+  function candidateToLogprobs(
+    candidate: GeminiResponseCandidate
+  ): LogprobContent | undefined {
+    const logprobs: GeminiLogprobsResult = candidate?.logprobsResult;
+    const chosenTokens: GeminiLogprobsResultCandidate[] =
+      logprobs?.chosenCandidates ?? [];
+    const topTokens: GeminiLogprobsTopCandidate[] =
+      logprobs?.topCandidates ?? [];
+    const content: Logprob[] = [];
+    for (let co = 0; co < chosenTokens.length; co += 1) {
```
jacoblee93, line 715: nit: what does `co` mean?

jacoblee93 (commented):
Thanks! Some nits

libs/langchain-google-common/src/utils/gemini.ts
afirstenberg: Counter. (Throwback to my BASIC days.)

libs/langchain-google-common/src/chat_models.ts
afirstenberg: Doh! Yeah. Thanks for catching this.

afirstenberg:
Fixed the type declarations. Thanks!

jacoblee93 (approved):
Thanks!

PR #7477, opened 2025-01-07
fix(community): update YoutubeLoader implementation

sinedied:
## Changes

- Removes `youtube-transcript` dependency
- Updates `youtubei.js` dependency and made peer dependency versioning lenient
- Refactored `YoutubeLoader` to use `youtubei.js` Innertube API instead or relying on `youtube-transcript` web scraping (may solve some issues regarding [YT IP banning](https://github.com/Kakulukian/youtube-transcript/issues/11#issuecomment-2244224076))
- Added test for YoutubeLoader

As YT transcript loading is often broken and requires frequent package updates, this change will (hopefully) simplify the maintenance by now relying on a single package that can be updated independently of @langchain/community package.

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -28,7 +28,6 @@
     "readline": "https://deno.land/x/readline@v1.1.0/mod.ts",
     "uuid": "npm:/uuid",
     "youtubei.js": "npm:/youtubei.js",
-    "youtube-transcript": "npm:/youtube-transcript",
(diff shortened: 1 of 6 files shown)


jacoblee93 (approved):
Thank you!

jacoblee93:
I think we will also need to update docs but I can do that in a followup

PR #7479, opened 2025-01-07
feat(core): Allow checking tool_calls on any BaseMessage without type casts

jacoblee93:
We should no longer need to do casts like this:

```ts
const prettyPrint = (message: BaseMessage) => {
  let txt = `[${message._getType()}]: ${message.content}`;
  if (
    isAIMessage(message) && (message as AIMessage)?.tool_calls?.length || 0 > 0
  ) {
    const tool_calls = (message as AIMessage)?.tool_calls
      ?.map((tc) => `- ${tc.name}(${JSON.stringify(tc.args)})`)
      .join("\n");
    txt += ` \nTools: \n${tool_calls}`;
  }
  console.log(txt);
};
```

Instead:

```ts
const prettyPrint = (message: BaseMessage) => {
  let txt = `[${message._getType()}]: ${message.content}`;
  if (message.tool_calls?.length) {
    const tool_calls = message.tool_calls
      ?.map((tc) => `- ${tc.name}(${JSON.stringify(tc.args)})`)
      .join("\n");
    txt += ` \nTools: \n${tool_calls}`;
  }
  console.log(txt);
};
```

### DIFF
--- a/langchain-core/src/messages/base.ts
+++ b/langchain-core/src/messages/base.ts
@@ -173,6 +173,23 @@ function stringifyWithDepthLimit(obj: any, depthLimit: number): string {
 }
 
 /**
+ * A call to a tool.
+ * @property {string} name - The name of the tool to be called
+ * @property {Record<string, any>} args - The arguments to the tool call
+ * @property {string} [id] - If provided, an identifier associated with the tool call
+ */
+export type ToolCall = {
+  name: string;
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  args: Record<string, any>;
+
+  id?: string;
+
+  type?: "tool_call";
+};
+
+/**
  * Base class for all types of messages in a conversation. It includes
  * properties like `content`, `name`, and `additional_kwargs`. It also
  * includes methods like `toDict()` and `_getType()`.
@@ -218,6 +235,8 @@ export abstract class BaseMessage
    * provided by the provider/model which created the message.
    */
   id?: string;
+
+  tool_calls?: never[] | ToolCall[];
 
   /**
    * @deprecated Use .getType() instead or import the proper typeguard.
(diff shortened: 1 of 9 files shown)


langchain-core/src/messages/base.ts
```diff
@@ -219,6 +236,8 @@ export abstract class BaseMessage
    */
   id?: string;
 
+  tool_calls?: never[] | ToolCall[];
```
efriis, line 239:
```suggestion
  tool_calls?: ToolCall[];
```
I'm likely misunderstanding typescript `never[]` type but shouldn't this just be this, and it'll be undefined in most cases?

jacoblee93: No, I still want this to throw a type error:

```ts
if (isHumanMessage(msg)) {
  // This should error
  message.tool_calls?.[0].args
}
```

jacoblee93:
So this will mess up a few automatic type inference situations like this:

```ts
// messages is inferred as SystemMessage[]
const messages = [new SystemMessage("foo")];
messages.push(new AIMessage("bar"));
```

```
error TS2345: Argument of type 'AIMessage' is not assignable to parameter of type 'HumanMessage'.
  Types of property 'tool_calls' are incompatible.
    Type 'ToolCall[] | undefined' is not assignable to type 'never[] | undefined'.
      Type 'ToolCall[]' is not assignable to type 'never[]'.
        Type 'ToolCall' is not assignable to type 'never'.
```

I still think it's worth it since you can do `const messages: BaseMessage[] = [...]` or just use dict (as most people should be doing) syntax and it's just this one type narrowing situation that causes issues. Would love thoughts.

jacoblee93:
Yeah think I need to find a way to make this non-breaking...

langchain-core/src/messages/base.ts
efriis: why? shouldn't it always be undefined it it's a humanmessage?

nfcampos (approved):

christian-bromann:
Checked in with @jacoblee93 on this and agreed to close this out as we are in the effort of revamping the message formats in v1.

PR #7481, opened 2025-01-08
fix(groq): Fix Max Retries for Groq

weakit:
The `maxRetries` field is not respected with the Groq integration. This PR fixes this issue.

By default, the Groq SDK automatically retries any request twice.   
So with a default of 6 max retries, this results in 6 * 3 = 18 retries before the request is passed on to a fallback provider or failing.


jacoblee93:
Thank you - nice catch!

PR #7485, opened 2025-01-08
docs: Add docs for Upstash Vector built-in embeddings support

ytkimirti:
This PR adds docs about the [built-in embedding model](https://upstash.com/docs/vector/features/embeddingmodels) support to the Upstash vector integration.

There already were docs for the embedding support, however it probably got pruned when the docs being unified.

This is the PR adding embedding support and docs for reference: #5150


jacoblee93 (approved):
Thanks!

PR #7486, opened 2025-01-08
weaviate[minor]: Add Document className as metadata

rolando-cruz:
### Description

- Added `className` as part of Document metadata.

### Why

- If we have multiple weaviate classes and use wrapper retrievers like `EnsembleRetriever`, we don't know where the docs are coming from.

### DIFF
--- a/libs/langchain-weaviate/src/vectorstores.ts
+++ b/libs/langchain-weaviate/src/vectorstores.ts
@@ -322,7 +322,7 @@ export class WeaviateStore extends VectorStore {
             pageContent: text,
             metadata: {
               ...rest,
-              collectionName: this.indexName
+              collectionName: this.indexName,
             },
             id: _additional.id,
           }),
--- a/libs/langchain-weaviate/src/vectorstores.ts
+++ b/libs/langchain-weaviate/src/vectorstores.ts
@@ -322,7 +322,7 @@ export class WeaviateStore extends VectorStore {
             pageContent: text,
             metadata: {
               ...rest,
-              className: this.indexName
+              collectionName: this.indexName
             },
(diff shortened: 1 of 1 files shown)


libs/langchain-weaviate/src/vectorstores.ts
```diff
@@ -320,7 +320,10 @@ export class WeaviateStore extends VectorStore {
         documents.push([
           new Document({
             pageContent: text,
-            metadata: rest,
+            metadata: {
+              ...rest,
+              className: this.indexName
```
jacoblee93, line 325: Should we just call it `indexName` instead of `className`?

rolando-cruz: We could but the idea is to export the `className` metadata field from `weaviate`. Since we are exporting this field in the `Document` interface `metadata` field, I would think is even safer to use `className` to avoid any field name collision. 

PD: Thanks for the fast feedback!

jacoblee93: In their docs I see:

https://weaviate.io/developers/weaviate/manage-data/collections

```
Newer Weaviate documentation discuses "collections." Older Weaviate documentation refers to "classes" instead. Expect to see both terms throughout the documentation.
```

Maybe `collectionName` is more appropriate?

rolando-cruz:
Hi @jacoblee93 can you take a look? Looking to try to get this in the next release

rolando-cruz:
Hi @jacoblee93 can you take a look? Looking to try to get this in the next release

jacoblee93 (approved):
Thank you!

PR #7496, opened 2025-01-10
feat(anthropic): Add Anthropic PDF support (document type) in invoke

adhambadr:
Since `claude-3-5-sonnet-20240620` [PDF Support](https://docs.anthropic.com/en/docs/build-with-claude/pdf-support) has been added to Anthropic's message types. 

You are able to send the PDF document to Anthropic and they do text extraction, Image conversion and supply the LLM with both (Text + Screenshot) of each page to do deep dive analysis, text extraction and more. Its pretty neat and handy especially in doing structured output and I added support for it in the Langchain Ecosystem as right now using document type throws an unsupported type error before passing it to the LLM. 

I added the source type `document` support as well as simplifying the source object to just pass the base64 or the object exactly as in Anthropic's documentation. 
I added a working example inside `yarn example examples/src/prompts/pdf_document.ts`

Here is an example usage: 
```Typescript
import { ChatAnthropic } from "@langchain/anthropic";

    model: "claude-3-5-sonnet-20240620",
 // Key 
});

// Local file
const file = fs.readFileSync("test.pdf");
const base64 = Buffer.from(file).toString("base64");
// Or Load remotely (web environment): 
const res = await fetch("https://www.w3.org/WAI/ER/tests/xhtml/testfiles/resources/pdf/dummy.pdf")
const buffer = await res.arrayBuffer();
const base64 = Buffer.from(buffer).toString("base64");

const prompt = "Summarize for me the contents of this document"; 
const {content} = await llm.invoke([ 
  {
     role : "user",
     content :  [
        {
          type: "text",
          text: prompt,
        },
        {
          type: "document",
          source: base64,
        }
      ]
   }
]);

console.log(content);

```

It's my first PR to this project so apologies if I missed something crucial, feedback or improvements are welcomed, as all as the shoutout to my [twitter](https://x.com/adhambadr2) 

Supported models as of Jan 2025: 

1. `claude-3-5-sonnet-20240620`
2. `claude-3-5-sonnet-20241022`


jacoblee93:
Ah nice!

libs/langchain-anthropic/src/utils/message_inputs.ts
```diff
@@ -131,6 +131,20 @@ function _formatContent(content: MessageContent) {
           source,
           ...(cacheControl ? { cache_control: cacheControl } : {}),
         };
+      } else if (contentPart.type === "document") {
+        // PDF
+        return {
+          type: "document",
+          source:
+            typeof contentPart.source === "string"
```
jacoblee93, line 139: Let's just allow Anthropic's format directly here

Things are still changing quickly (you could imagine more media types soon) and allowing devs to pass the raw content directly gives us the most flexibility

jacoblee93 (commented):
Thanks for flagging this - see comment!

libs/langchain-anthropic/src/utils/message_inputs.ts
adhambadr: Sure up to you, I already included that option the line after it, if the source is not a string (an Object) it gets passed as it is. 
```javascript
source: typeof contentPart.source === "string" ? { .. } : contentPart.source
```
So you can either pass the base64 as a string in the source or provide the whole source object with whatever values you need inside (given you set the type as document. 

```javascript
const {content} = await llm.invoke([ 
  {
     role : "user",
     content :  [
        {
          type: "text",
          text: prompt,
        },
        {
          type: "document",
          source: { 
            media_type:'application/pdf',
            type : 'base64',
            source: base64String
        }
      ]
   }
]);
```

camwardy:
Thanks for adding this @adhambadr, hopefully this can get merged soon as it'd be really useful for us!

libs/langchain-anthropic/src/utils/message_inputs.ts
jacoblee93: Still, I think making it default to PDF could be confusing in the future. I can make the change myself.

jacoblee93:
Thank you!

adhambadr:
> Thank you!

thanks a lot for the clean up (removing console, unnecessary import etc)!

JoeABCDEF:
@jacoblee93 Is it possible to provide such functionality for @langchain/openai as well?

PR #7510, opened 2025-01-13
fix(langchain): Fix ZeroShotAgent createPrompt with correct formatted tool names

thu2004:
Fix ZeroShotAgent createPrompt with correct formatted tool names

### DIFF
--- a/langchain/src/agents/mrkl/index.ts
+++ b/langchain/src/agents/mrkl/index.ts
@@ -128,7 +128,7 @@ export class ZeroShotAgent extends Agent {
       .map((tool) => `${tool.name}: ${tool.description}`)
       .join("\n");
 
-    const toolNames = tools.map((tool) => tool.name);
+    const toolNames = tools.map((tool) => tool.name).join("\n");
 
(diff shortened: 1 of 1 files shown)


thu2004:
Screenshot 2025-01-13 at 10 31 32
Here is where I discovered the issue. Instead of 
```
Action: the action to take, should be one of [["query-sql","info-sql","list-tables-sql","query-checker"]] , 
```
it should be 
```
Action: the action to take, should be one of ["query-sql","info-sql","list-tables-sql","query-checker"]
```

thu2004:
Hi @jacoblee93, Can you take a look at this PR? I'd appreciate your feedback.

jacoblee93:
Thanks for this!

Yeah I think the extra brackets are probably not optimal, though I think separating by newline is also weird

How about changing to comma separated instead and trying that out? Let's keep the quotes too

So:

```ts
const toolNames = tools.map((tool) => `"${tool.name}"`).join(", ");
```

thu2004:
@jacoblee93 Thank you for the feedback. I updated the code with your input.

jacoblee93 (approved):
Thank you!

PR #7521, opened 2025-01-14
feat(community): Provide fallback relationshipType in case it is not present in graph_transformer

hlminh2000:
Although difficult to reproduce, I've ran into a few times when `relationshipType` was undefined and so the extraction fails. 

In these cases I'd still like to have the relationship being captured in my knowledge graph, so it seems appropriate to provide the transformer with a fallback value it can use, in this case defaulted to `"unknown"`. This also allows for the value to be overridden with something else in cases where `"unknown"` already means something in an existing context.

I'm happy to contribute to anything that may need to be updated to reflect this info, like documentations, just need a pointer :)

### DIFF
--- a/libs/langchain-community/src/experimental/graph_transformers/llm.ts
+++ b/libs/langchain-community/src/experimental/graph_transformers/llm.ts
@@ -201,7 +201,11 @@ function mapToBaseNode(node: any): Node {
 }
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-function mapToBaseRelationship({ fallbackRelationshipType }: { fallbackRelationshipType: string | null }) {
+function mapToBaseRelationship({
+  fallbackRelationshipType,
+}: {
+  fallbackRelationshipType: string | null;
+}) {
   return function (relationship: any): Relationship {
     return new Relationship({
       source: new Node({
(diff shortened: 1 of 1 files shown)


libs/langchain-community/src/experimental/graph_transformers/llm.ts
```diff
@@ -254,6 +258,7 @@ export class LLMGraphTransformer {
     strictMode = true,
     nodeProperties = [],
     relationshipProperties = [],
+    fallbackRelationshipType = "unknown,
```
jacoblee93, line 261: Think you're missing a quote here!

jacoblee93 (commented):
Thanks for this!

CC @tomasonjo for a look as well

tomasonjo:
Actually, can you make the default fallback relationship type Null, so the default behavior isn't changed. When the fallback relationship is configured, then you can use it as fallback.

Also describe the new attribute in docstring please.

hlminh2000:
@tomasonjo @jacoblee93 thank you both for your reviews. Addressed and pushed!

jacoblee93:
Will wait for your final approval @tomasonjo

tomasonjo:
Lgtm

jacoblee93:
Thank you both!

PR #7535, opened 2025-01-16
feat(openai): Make only AzureOpenAI respect Azure env vars, remove class defaults, update withStructuredOutput defaults

jacoblee93:
summary of changes:

- moves azure-specific properties from ChatOpenAI to AzureChatOpenAI
- same for embeddings and llm
- removes default values (defaults to undefined; let openai populate default)
- update default `withStructuredOutput` method to be `jsonSchema` for models that support it


jacoblee93:
CC @sinedied

libs/langchain-openai/src/chat_models.ts
```diff
@@ -1150,30 +1148,20 @@ export class ChatOpenAI<
...
-    this.azureOpenAIBasePath =
-      fields?.azureOpenAIBasePath ??
-      getEnvironmentVariable("AZURE_OPENAI_BASE_PATH");
+    this.azureOpenAIBasePath = fields?.azureOpenAIBasePath;
+
+    this.azureOpenAIEndpoint = fields?.azureOpenAIEndpoint;
 
     this.organization =
       fields?.configuration?.organization ??
       getEnvironmentVariable("OPENAI_ORGANIZATION");
```
sinedied, line 1164: Isn'it it missing setting the endpoint property here, like the others?

sinedied (commented):
Can I ask why only moving the env vars and not all azure related properties, since it's a breaking change anyways? (Unless it's that's part of a multi-PR change, then ignore that). Just wondering what the goal is with this change, as it keeps the Azure compatibility with the base OpenAI class when using config property but breaks it when using env vars?

I might be wrong on this one, but should also be `lc_secrets()` updated in the Azure implementation since the env vars are moved?

@sarangan12 any thoughts?

jacoblee93:
> Can I ask why only moving the env vars and not all azure related properties, since it's a breaking change anyways? (Unless it's that's part of a multi-PR change, then ignore that). Just wondering what the goal is with this change, as it keeps the Azure compatibility with the base OpenAI class when using config property but breaks it when using env vars?

Yeah I think this is probably a good idea. CC @efriis

libs/langchain-openai/src/chat_models.ts
jacoblee93: Will just remove entirely

jacoblee93:
The minor bump will also enable #7564

jacoblee93:
@sinedied @efriis @ccurme have to verify tests but WDYT

Will make a different branch that I'll merge this into as well as the switch to `json_schema` for `withStructuredOutput` and treat that all as breaking changes for 0.4.0

libs/langchain-openai/src/chat_models.ts
```diff
@@ -1006,7 +1003,7 @@ export class ChatOpenAI<
     CallOptions extends ChatOpenAICallOptions = ChatOpenAICallOptions
   >
   extends BaseChatModel<CallOptions, AIMessageChunk>
-  implements OpenAIChatInput, AzureOpenAIInput
+  implements Partial<OpenAIChatInput>
```
efriis, line 1006: was this a bug before or is there something we're doing to not require all inputs anymore

libs/langchain-openai/src/chat_models.ts
```diff
@@ -1043,28 +1039,24 @@ export class ChatOpenAI<
...
       apiKey: "openai_api_key",
-      azureOpenAIApiVersion: "azure_openai_api_version",
-      azureOpenAIApiKey: "azure_openai_api_key",
-      azureOpenAIApiInstanceName: "azure_openai_api_instance_name",
-      azureOpenAIApiDeploymentName: "azure_openai_api_deployment_name",
     };
   }
 
-  temperature = 1;
+  temperature?: number;
```
efriis, line 1045: just confirming passing undefined = default here, not null (pretty sure this is the case but typescript always makes this fun)

libs/langchain-openai/src/embeddings.ts
```diff
@@ -65,11 +61,11 @@ export interface OpenAIEmbeddingsParams extends EmbeddingsParams {
  */
 export class OpenAIEmbeddings
   extends Embeddings
-  implements OpenAIEmbeddingsParams, AzureOpenAIInput
+  implements OpenAIEmbeddingsParams
 {
-  modelName = "text-embedding-ada-002";
+  modelName: string;
```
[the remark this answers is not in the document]
jacoblee93: Interesting

libs/langchain-openai/src/chat_models.ts
```diff
@@ -1137,43 +1111,10 @@ export class ChatOpenAI<
...
     this.organization =
       fields?.configuration?.organization ??
       getEnvironmentVariable("OPENAI_ORGANIZATION");
 
-    this.azureOpenAIEndpoint =
-      fields?.azureOpenAIEndpoint ??
-      getEnvironmentVariable("AZURE_OPENAI_ENDPOINT");
-
     this.modelName = fields?.model ?? fields?.modelName ?? this.model;
     this.model = this.modelName;
```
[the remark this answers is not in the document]
jacoblee93: Yeah, I just realized I never marked `modelName` as deprecated 😭

libs/langchain-openai/src/chat_models.ts
```diff
@@ -1043,28 +1039,24 @@ export class ChatOpenAI<
...
 
   n = 1;
 
   logitBias?: Record<string, number>;
 
-  modelName = "gpt-3.5-turbo";
+  modelName: string;
 
-  model = "gpt-3.5-turbo";
+  model: string;
```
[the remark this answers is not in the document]
jacoblee93: Yeah I can remove from class

libs/langchain-openai/src/chat_models.ts
jacoblee93: I may need to just not specify it when invoking

libs/langchain-openai/src/chat_models.ts
jacoblee93: To not require all inputs

libs/langchain-openai/src/chat_models.ts
jacoblee93: Seems fine

libs/langchain-openai/src/legacy.ts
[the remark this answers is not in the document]
jacoblee93: Sure

jacoblee93:
Going to try to ship 0.4.0 today - @sinedied any other thoughts?

libs/langchain-openai/src/azure/chat_models.ts
```diff
@@ -264,7 +266,7 @@ import {
  * const Joke = z.object({
  *   setup: z.string().describe("The setup of the joke"),
  *   punchline: z.string().describe("The punchline to the joke"),
- *   rating: z.number().optional().describe("How funny the joke is, from 1 to 10")
+ *   rating: z.number().nullable().describe("How funny the joke is, from 1 to 10")
```
efriis, line 269: does openai just like nullable better?

libs/langchain-openai/src/azure/chat_models.ts
```diff
@@ -425,12 +427,27 @@ import {
...
+
+  azureADTokenProvider?: () => Promise<string>;
+
+  azureOpenAIApiInstanceName?: string;
+
+  azureOpenAIApiDeploymentName?: string;
+
+  azureOpenAIBasePath?: string;
+
+  azureOpenAIEndpoint?: string;
```
efriis, line 442: do these match python / should we take this as an opportunity to make the variable names match, or keep as is to make as little breaking as possible?

libs/langchain-openai/src/azure/chat_models.ts
```diff
@@ -450,21 +474,43 @@ export class AzureChatOpenAI extends ChatOpenAI {
...
+    this.azureOpenAIApiDeploymentName =
+      fields?.azureOpenAIApiDeploymentName ??
+      fields?.deploymentName ??
+      getEnvironmentVariable("AZURE_OPENAI_API_DEPLOYMENT_NAME");
 
-    super(newFields);
+    this.azureOpenAIApiVersion =
+      fields?.azureOpenAIApiVersion ??
+      fields?.openAIApiVersion ??
+      getEnvironmentVariable("AZURE_OPENAI_API_VERSION");
```
efriis, line 499: this feels funky - should we just update `this.openAIApiVersion`?

libs/langchain-openai/src/azure/chat_models.ts
```diff
@@ -450,21 +474,43 @@ export class AzureChatOpenAI extends ChatOpenAI {
...
+      getEnvironmentVariable("AZURE_OPENAI_API_KEY");
+
+    this.azureOpenAIApiInstanceName =
+      fields?.azureOpenAIApiInstanceName ??
+      getEnvironmentVariable("AZURE_OPENAI_API_INSTANCE_NAME");
+
+    this.azureOpenAIApiDeploymentName =
+      fields?.azureOpenAIApiDeploymentName ??
+      fields?.deploymentName ??
+      getEnvironmentVariable("AZURE_OPENAI_API_DEPLOYMENT_NAME");
```
efriis, line 494: this is not in py but we should add it in py

libs/langchain-openai/src/chat_models.ts
```diff
@@ -83,7 +81,7 @@ import {
 } from "./utils/openai-format-fndef.js";
 import { _convertToOpenAITool } from "./utils/tools.js";
 
-export type { AzureOpenAIInput, OpenAICallOptions, OpenAIChatInput };
```
efriis, line 86: will removing this export mess some people up or better to have all azure exports from azure file directly?

libs/langchain-openai/src/embeddings.ts
```diff
@@ -65,12 +62,13 @@ export interface OpenAIEmbeddingsParams extends EmbeddingsParams {
...
   extends Embeddings
-  implements OpenAIEmbeddingsParams, AzureOpenAIInput
+  implements Partial<OpenAIEmbeddingsParams>
 {
-  modelName = "text-embedding-ada-002";
-
   model = "text-embedding-ada-002";
 
+  /** @deprecated Use "model" instead */
+  modelName: string;
```
efriis, line 70: does this need a default value to not throw an error - or ability to make undefined

libs/langchain-openai/src/llms.ts
```diff
@@ -14,20 +14,11 @@ import type {
...
 
-export type {
-  AzureOpenAIInput,
-  OpenAICallOptions,
-  OpenAIInput,
-  OpenAIChatCallOptions,
-};
-
-export { OpenAIChat };
+export type { AzureOpenAIInput, OpenAICallOptions, OpenAIInput };
```
efriis, line 21: should remove azure one if philosophy is to hide azure stuff within azure files

efriis (approved):
looks good - some small nits if helpful but you understand better than me here

libs/langchain-openai/src/azure/chat_models.ts
jacoblee93: Yeah not sure why, they just don't respect it. Guess not an issue with Pydantic.

libs/langchain-openai/src/azure/chat_models.ts
jacoblee93: Yeah I think too big a break

libs/langchain-openai/src/azure/chat_models.ts
jacoblee93: WDYM?

libs/langchain-openai/src/chat_models.ts
jacoblee93: Yeah I think taking the opportunity to separate things would be nice

libs/langchain-openai/src/azure/chat_models.ts
jacoblee93: I don't even understand what any of these do

libs/langchain-openai/src/embeddings.ts
jacoblee93: You just have to set it in the constructor (which I do)

libs/langchain-openai/src/llms.ts
jacoblee93: Yeah good call thanks

PR #7542, opened 2025-01-17
fix(community): make sure guardrailConfig can be added even with anthropic models

madmed88:
Fixes # 7541


jacoblee93 (approved):
Thank you! Note that you should generally be using the Converse API through the `@langchain/aws` package

PR #7543, opened 2025-01-17
chore(openai): update azure user agent

sinedied:
## Changes

Fix library name and versioning, and add details about the Node.js version and platform in the user agent


libs/langchain-openai/src/azure/chat_models.ts
```diff
@@ -504,9 +505,7 @@ export class AzureChatOpenAI extends ChatOpenAI {
 
       params.defaultHeaders = {
         ...params.defaultHeaders,
-        "User-Agent": params.defaultHeaders?.["User-Agent"]
-          ? `${params.defaultHeaders["User-Agent"]}: langchainjs-azure-openai-v2`
-          : `langchainjs-azure-openai-v2`,
+        "User-Agent": `langchainjs-azure-openai/2.0.0 (node.js/${process.version}; ${process.platform}; ${process.arch}) ${params.defaultHeaders?.["User-Agent"] || ""}`
```
johnpapa, line 508: should the version be pulled from a variable, so this doesnt have to be reset on future releases? @sinedied ?

jacoblee93: Unfortunately we don't have a way to do this at the moment 😞

sinedied: @johnpapa the `2.0.0` is just to distinguish from the previous (deprecated) package implementation, it's not really a proper versioning as it won't change. It would be nice to send along the @langchain/openai package version though as Jacob mentioned it doesn't seem to be possible currently: from what I understand, we can't pull the version from `package.json` as the lib can also be ran from the browser where there's no FS access like with Node.js.

jacoblee93: Yeah we would need to manually some process to bump a statically specified version on release. We do it here:

https://github.com/langchain-ai/langsmith-sdk/blob/main/js/src/index.ts#L21

I just haven't had time to set it up

jacoblee93 (approved):
Thanks!

libs/langchain-openai/src/azure/chat_models.ts
```diff
@@ -1,3 +1,4 @@
+import process from "node:process";
```
jacoblee93, line 1: Oh hmm actually this will break non-node environments - can we use this instead? https://github.com/langchain-ai/langchainjs/blob/main/langchain-core/src/utils/env.ts#L40

jacoblee93 (changes requested):

libs/langchain-openai/src/azure/chat_models.ts
sinedied: My bad, I forgot about this!
I sent an update using `getEnv` that fixes that

jacoblee93 (approved):
Thanks!

libs/langchain-openai/src/azure/chat_models.ts
```diff
@@ -555,11 +556,16 @@ export class AzureChatOpenAI extends ChatOpenAI {
...
+      if (env === "node" || env === "deno") {
+        env = `(${env}/${process.version}; ${process.platform}; ${process.arch})`;
+      }
+
       params.defaultHeaders = {
         ...params.defaultHeaders,
-        "User-Agent": params.defaultHeaders?.["User-Agent"]
-          ? `${params.defaultHeaders["User-Agent"]}: langchainjs-azure-openai-v2`
-          : `langchainjs-azure-openai-v2`,
+        "User-Agent": `langchainjs-azure-openai/2.0.0 (${env}) ${
```
jacoblee93, line 566: Actually, here you will have an extra space if no default user agent. How about:

```
const specifiedUserAgent = params.defaultHeaders?.["User-Agent"];

params.defaultHeaders = {
        ...params.defaultHeaders,
        "User-Agent": `langchainjs-azure-openai/2.0.0 (${env})${specifiedUserAgent ? ` ${specifiedUserAgent}` : ""}`
};
```

jacoblee93:
Also looks like lint is failing:

```ts
Error:   6:24  error  '@langchain/core/utils/env' imported multiple times  import/no-duplicates
Error:   7:40  error  '@langchain/core/utils/env' imported multiple times  import/no-duplicates
```

jacoblee93:
@sinedied happy to merge as is, but LMK if you want to do this: https://github.com/langchain-ai/langchainjs/pull/7543#discussion_r1943669555

libs/langchain-openai/src/azure/chat_models.ts
sinedied: Thanks for the catch, I'll fix that!

sinedied:
@jacoblee93 I think all issues are now resolved, could you have a look? Thanks

PR #7555, opened 2025-01-19
feat(community): Adds graph_document to export list

hlminh2000:
# Changes
Adds `graphs/graph_document` to the export list in `package.json`
# Why
I'm trying to build a custom knowledge graph and having access to the classes exported by `graphs/graph_document` would help greatly.


jacoblee93:
It's a bit more involved than this, can fix up and export though!

jacoblee93 (approved):
Thanks!

PR #7561, opened 2025-01-20
feat(community): improve support for Tavily search tool args

hugoleborso:
I needed a property of the API (`includeDomains`) that was not in the `TavilySearchAPIRetrieverFields`, so I decided to update all of the body params from the [Tavily API reference](https://docs.tavily.com/docs/rest-api/api-reference).

Let me know if you have remarks, this is my first contribution 😇

### DIFF
--- a/libs/langchain-community/src/tools/tavily_search.ts
+++ b/libs/langchain-community/src/tools/tavily_search.ts
@@ -170,23 +170,23 @@ export class TavilySearchResults extends Tool {
 
   protected kwargs: Record<string, unknown> = {};
 
-  protected includeImages = false;
+  protected includeImages?: boolean;
 
-  protected includeImageDescriptions = false;
+  protected includeImageDescriptions?: boolean;
 
-  protected includeAnswer = false;
+  protected includeAnswer?: boolean;
 
-  protected includeRawContent = false;
+  protected includeRawContent?: boolean;
 
-  protected includeDomains: string[] = [];
+  protected includeDomains?: string[];
 
-  protected excludeDomains: string[] = [];
+  protected excludeDomains?: string[];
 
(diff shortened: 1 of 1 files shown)


libs/langchain-community/src/tools/tavily_search.ts
```diff
@@ -93,11 +170,41 @@ export class TavilySearchResults extends Tool {
...
+
+  protected includeAnswer = false;
+
+  protected includeRawContent = false;
+
+  protected includeDomains: string[] = [];
+
+  protected excludeDomains: string[] = [];
+
+  protected searchDepth: "basic" | "deep" = "basic";
```
jacoblee93, line 185: Let's avoid defaults at this level when possible

hugoleborso: Hi @jacoblee93, I did this because these are actually the default settings of the API doc, and also because `maxResults ` already had a default value set corresponding to the default value in the doc.

If you want me too I can still change it of course, but I think it makes sense as is.

Let me know what you think!

jacoblee93: We should ideally let the backend set this in case they change best practices, would prefer to have things unset if it won't cause issues (which it wasn't before)

hugoleborso: @jacoblee93 done, let me know if this fit your needs, and thanks for all the quick responses !

jacoblee93 (approved):
Thank you!

libs/langchain-community/src/tools/tavily_search.ts
jacoblee93: Thank you, old defaults have started causing issues for OpenAI so it's top of mind right now

PR #7574, opened 2025-01-23
chore(ollama): Export `OllamaEmbeddingsParams` interface

gurisko:
We are already exporting `ChatOllamaInput` in `@langchain/ollama`. It would be great to also export `OllamaEmbeddingsParams` for embeddings instance.


jacoblee93 (approved):
Whoops, thank you!

PR #7616, opened 2025-01-29
fix(openai): Allow o1 to stream a single chunk

bracesproul:
cc @jacoblee93

### DIFF
--- a/libs/langchain-openai/src/chat_models.ts
+++ b/libs/langchain-openai/src/chat_models.ts
@@ -1002,8 +1002,7 @@ export class ChatOpenAI<
     this.reasoningEffort = fields?.reasoningEffort;
 
     if (this.model === "o1") {
-      this.disableStreaming =
-        fields?.disableStreaming != null ? fields?.disableStreaming : true;
+      this.disableStreaming = fields?.disableStreaming ?? true;
     }
 
     this.streaming = fields?.streaming ?? false;
--- a/libs/langchain-openai/src/chat_models.ts
+++ b/libs/langchain-openai/src/chat_models.ts
@@ -1451,7 +1451,6 @@ export class ChatOpenAI<
     const messagesMapped: OpenAICompletionParam[] =
       _convertMessagesToOpenAIParams(messages, this.model);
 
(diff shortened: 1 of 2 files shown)


libs/langchain-openai/src/chat_models.ts
```diff
@@ -1002,7 +1002,8 @@ export class ChatOpenAI<
     this.reasoningEffort = fields?.reasoningEffort;
 
     if (this.model === "o1") {
-      this.disableStreaming = true;
+      this.disableStreaming =
+        fields?.disableStreaming != null ? fields?.disableStreaming : true;
```
bracesproul, line 1006: this is the one area which is technically breaking

jacoblee93: Only if someone's passing `disableStreaming: false` right? I think fine

jacoblee93: You can just do `fields?.disableStreaming ?? true`

bracesproul: ya `??` would work

libs/langchain-openai/src/chat_models.ts
```diff
@@ -1310,6 +1311,15 @@ export class ChatOpenAI<
     options: this["ParsedCallOptions"],
     runManager?: CallbackManagerForLLMRun
   ): AsyncGenerator<ChatGenerationChunk> {
+    if (this.streaming === false) {
```
jacoblee93, line 1313: Could we fix this in `core` instead?

jacoblee93: Here: https://github.com/langchain-ai/langchainjs/blob/main/langchain-core/src/language_models/chat_models.ts#L241

bracesproul: `streaming` doesn't exist on `BaseChatModel`

jacoblee93: Ah ok - why can't we just set `disableStreaming: true`?

PR #7621, opened 2025-01-29
feat(google-genai): Add support for fileUri in media type in Google GenAI

hongkongkiwi:
Adds support for fileUri type which accepts Google AI Filemanager URIs.

This allows uploading large files and passing the uri as per the Gemini documentation without having to send inline data.


jacoblee93 (approved):
Thanks!

PR #7631, opened 2025-01-30
Update fileData & inlineData to kebab case as per REST API requirements

hongkongkiwi:
Google Gemini HTTP API (which is what we are using in this module) actually requires file_data and inline_data to be in kebab case. This is different to the SDKs which use Snake case. 

It is possible to use Snake case for some things but with mixed results (e.g. inlineData works, but fileData does not), however officially both should be kebab (see Shell docs): [https://ai.google.dev/api/files](https://ai.google.dev/api/files) and have been tested working.

Previously, I had contributed a fix which added the fileData #7621 , but it was using snake case which is wrong and didn't actually work consistently.

So this fix converts both inline_data and file_data to kebab case, which is what the HTTP API wants. Google docs are sparse, but REST/Shell examples always show kebab case for these fields.


hongkongkiwi:
Although this works, I misunderstood that we do actually use the official sdk client.

PR #7633, opened 2025-01-30
feat[community]: Add chat deployment to IBM chat class

FilipZmijewski:
This PR adds chat deployment options to Watsonx IBM class allowing user to use langchain with their deployed models

### DIFF
--- a/libs/langchain-community/src/chat_models/ibm.ts
+++ b/libs/langchain-community/src/chat_models/ibm.ts
@@ -405,8 +405,8 @@ export class ChatWatsonx<
       ls_provider: "watsonx",
       ls_model_name: this.model,
       ls_model_type: "chat",
-      ls_temperature: params?.temperature ?? undefined,
-      ls_max_tokens: params?.maxTokens ?? undefined,
+      ls_temperature: params.temperature ?? undefined,
+      ls_max_tokens: params.maxTokens ?? undefined,
     };
   }
 
--- a/libs/langchain-community/src/chat_models/ibm.ts
+++ b/libs/langchain-community/src/chat_models/ibm.ts
@@ -33,6 +33,7 @@ import {
 } from "@langchain/core/outputs";
 import { AsyncCaller } from "@langchain/core/utils/async_caller";
 import {
+  DeploymentsTextChatParams,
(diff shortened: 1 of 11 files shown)


libs/langchain-community/src/chat_models/ibm.ts
```diff
@@ -380,8 +405,8 @@ export class ChatWatsonx<
       ls_provider: "watsonx",
       ls_model_name: this.model,
       ls_model_type: "chat",
-      ls_temperature: params.temperature ?? undefined,
-      ls_max_tokens: params.maxTokens ?? undefined,
+      ls_temperature: params?.temperature ?? undefined,
```
jacoblee93, line 408: nit: why would `params` be optional?

libs/langchain-community/src/chat_models/ibm.ts
```diff
@@ -425,37 +452,44 @@ export class ChatWatsonx<
...
+      this.topLogprobs = fields?.topLogprobs;
+      this.maxTokens = fields?.maxTokens ?? this.maxTokens;
+      this.presencePenalty = fields?.presencePenalty;
+      this.topP = fields?.topP;
+      this.timeLimit = fields?.timeLimit;
+      this.responseFormat = fields?.responseFormat ?? this.responseFormat;
+      this.streaming = fields?.streaming ?? this.streaming;
+      this.n = fields?.n ?? this.n;
+      this.model = fields?.model ?? this.model;
+    } else this.idOrName = fields?.idOrName;
```
jacoblee93, line 487: Don't think you're guaranteed to be setting `model` in the constructor anymore?

Should make the instance property optional then?

jacoblee93 (commented):
Happy to merge as is but see comments

libs/langchain-community/src/chat_models/ibm.ts
FilipZmijewski: Correct, done

libs/langchain-community/src/chat_models/ibm.ts
FilipZmijewski: Managed to to it another way, fixed now

PR #7635, opened 2025-01-31
feat(community): Add support for Amazon Aurora DSQL memory message

jl4nz:
Based on postgresql integration, but modified to support DSQL compatibility. 

1. The `id` column in PostgreSQL is SERIAL auto-incrementent, and DSQL is UUID using the database function `gen_random_uuid`.
2. A `created_at` column is created to track the order and history of the messages.
4. The `message` column in PostgreSQL is JSONB, and DSQL is TEXT with Javascript parsing handling


libs/langchain-community/src/stores/message/aurora_dsql.ts
```diff
@@ -0,0 +1,183 @@
...
+  BaseMessage,
+  StoredMessage,
+  mapChatMessagesToStoredMessages,
+  mapStoredMessagesToChatMessages,
+} from '@langchain/core/messages';
+import pg from 'pg';
+
+/**
+ * Type definition for the input parameters required when instantiating a
+ * PostgresChatMessageHistory object.
```
jacoblee93, line 12: Update naming

jacoblee93 (commented):
Sorry about the delay!

Could this/should this just subclass `PostgresChatMessageHistory`?

jl4nz:
> Sorry about the delay!
> 
> Could this/should this just subclass `PostgresChatMessageHistory`?

Hey thanks!

I think is possible and I though about it a bit, but the DSQL is kinda special (is not 100% all postgres features) and if something gets added to the parent class of postgres, it might break this one. Pretty sure the test will let us know, but some of these are disabled. For example the postgresql test is with: `describe.skip(...)`

I'm not sure how the test are run for libs that have credential requirements (AWS)... , so IMO if test will be run, yes, makes sense to be subclass, otherwise, better separated. Anyway, open to modify it after this discussion.

jacoblee93:
Hmm ok, can keep separate!

jacoblee93 (approved):

jacoblee93:
Thank you!

wesleyandrade05:
Hi!

I have been trying to use `AuroraDsqlChatMessageHistory` on my project, but I keep getting the following error:

```
unsupported mode. please use CREATE INDEX ASYNC.
    at /var/task/index.cjs:72186:15
    at async AuroraDsqlChatMessageHistory.createIndex (/var/task/index.cjs:97776:5)
    at async AuroraDsqlChatMessageHistory.ensureTable (/var/task/index.cjs:97766:7)
    at async AuroraDsqlChatMessageHistory.getMessages (/var/task/index.cjs:97788:5)
    at async RunnableWithMessageHistory._enterHistory (/var/task/index.cjs:77337:22)
    at async /var/task/index.cjs:23493:27
 ```

After deep-diving the error, I see this is related to the function `createIndex`, in particular, my Aurora DSQL table is not accepting `CREATE INDEX`, it just accepts `CREATE INDEX ASYNC`, even when the table is empty. I was wondering if you saw this issue at any point, and if this might be related to some recent update to Aurora DSQL. I also just see it mentioning about `CREATE INDEX ASYNC` on its documentation, so I wonder if we should change to that.

Thank you! Please let me know if you need any more details

jl4nz:
Probably a change after DSQL moved from preview. I can work on it next week

jl4nz:
@wesleyandrade05 https://github.com/langchain-ai/langchainjs/pull/8486 added async here.

PR #7644, opened 2025-02-04
fix(community): Store and return ids supabase vectorstore

bracesproul:
Currently, any `id` set on `Document` is lost when you insert via `fromDocuments`

This PR will now start setting them in the document metadata under a special key `__document_id`


jacoblee93:
Let's just pull from the `id` column and update setup docs for different types

PR #7664, opened 2025-02-06
fix(experimental): openai assistant attachments

zachary-nguyen:
Fixing openai assistants thread run. Parameters to create and run a thread do not match the latest from openai. 
This results in a unknown parameter error when trying to provide a file.

See docs:
https://platform.openai.com/docs/api-reference/runs/createThreadAndRun#runs-createthreadandrun-thread

Langchain Docs:
https://js.langchain.com/v0.1/docs/modules/agents/agent_types/openai_assistant/#use-file-in-ai-assistant

Fixes [# (issue)](https://github.com/langchain-ai/langchainjs/issues/7665)

### DIFF
--- a/langchain/src/experimental/openai_assistant/index.ts
+++ b/langchain/src/experimental/openai_assistant/index.ts
@@ -115,7 +115,7 @@ export class OpenAIAssistantRunnable<
           {
             role: "user",
             content: input.content,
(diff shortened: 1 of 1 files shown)


jacoblee93:
Thank you!

PR #7679, opened 2025-02-10
fix(community): hide console errors on RecursiveUrlLoader

jfromaniello:
Hide console errors from JSDOM on RecursiveUrlLoader.


jacoblee93 (approved):
Thanks!

PR #7695, opened 2025-02-14
fix(redis): Add TTL support for redis vector store

AllenFang:
hi team, want to address this [ask](https://github.com/langchain-ai/langchainjs/discussions/7632). It's a tiny change and python already supported it.  In addition, I also tested it in local redis stack.


jacoblee93 (approved):
Thank you!

PR #7704, opened 2025-02-15
fix(google-common): Eliminate hard-coded default values in favor of model-based defaults

afirstenberg:
This eliminates the hard coded default values for:
* temperature
* maxOutputTokens
* topP
* topK
and makes sure that values that aren't set aren't even sent in the request.

### DIFF
--- a/libs/langchain-google-common/src/chat_models.ts
+++ b/libs/langchain-google-common/src/chat_models.ts
@@ -181,13 +181,13 @@ export abstract class ChatGoogleBase<AuthOptions>
 
   modelName = "gemini-pro";
 
-  temperature = 0.7;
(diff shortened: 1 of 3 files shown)


jacoblee93 (approved):
Thanks for this!

Will make a minor bump as this is technically breaking

jacoblee93:
Please remember to run `yarn format` as well!

afirstenberg:
> Will make a minor bump as this is technically breaking

The changes for Express Mode are also probably considered breaking, so both at once certainly makes sense.

> Please remember to run `yarn format` as well!

Semicolon added. {:

jacoblee93:
Thank you!

PR #7706, opened 2025-02-17
fix(google-common): Handle multiple function calls and complex parts/generations

### LINKED ISSUE #7705: Gemini Flash 2 not making function calls In Langgraphjs
### Checked other resources

### Example Code

@afirstenberg 

Code used for binding the tools to the agent
``` javascript
const plannerModel = planningModel.bindTools([
          tools.get_resource_schema,
        ]);

```

Here is the code for where I setup the ChatVertexAI 

```javascript
const vertexAIModel = new ChatVertexAI({
        modelName: "gemini-2.0-flash-001",
        maxOutputTokens: 8192,
        response_mime_type: "application/json",
        serviceAccount: path.join(spice.parent_app_dir, "google-services.json"),
        temperature: 1,
      });
```

_No response_

### Description

@afirstenberg 

Chat VertexAI is not making tool calls. I am binding it the same way I did for the other model but its just not making the tool calls. 

Here is now I setup the ChatVertexAI. 
```javascript 
const vertexAIModel = new ChatVertexAI({
        modelName: "gemini-2.0-flash-001",
        maxOutputTokens: 8192,
        response_mime_type: "application/json",
        serviceAccount: path.join(spice.parent_app_dir, "google-services.json"),
        temperature: 1,
      });

 const plannerModel = vertexAIModel.bindTools([
          tools.get_resource_schema,
        ]);

```

Here is where I bind to the tools. This works for for the openai models perfectly.

Here is an example response it would give me 
Okay, I need to provide a breakdown of users by group. This means I need to access the `user` and `usergroup` resources and understand how they are related.

Here's my plan:

1.  **Get the schema for the `user` resource.** This will tell me what fields are available for users, including any fields that link to user groups.
2.  **Get the schema for the `usergroup` resource.** This will tell me what fields are available for user groups, which I'll need to display the group names.
3.  **Analyze the schemas to find the relationship between `user` and `usergroup`.**  I'll look for a field in the `user` resource that references the `usergroup` resource (e.g., a `usergroup_id`).
4.  **Request data from the data fetcher.** I will use the `perform_term_analysis` tool to get the count of users for each user group. I will group by the `usergroup` name and count the number of users in each group.
5.  **Prepare the response.** I'll format the data into a table showing the user group name and the number of users in each group.

Its supposed to actually make the tool calls instead its stating what it would do. I told it which tools it has access too in the prompt that why its making reference to the tools. but it does not seem like the binding is working. 

### System Info

"@google-cloud/vertexai": "^1.9.3",
 "@langchain/anthropic": "^0.3.13",
    "@langchain/community": "^0.3.30",
    "@langchain/core": "^0.3.40",
    "@langchain/google-vertexai": "^0.1.8",
    "@langchain/groq": "^0.1.3",
    "@langchain/langgraph": "^0.2.46",
    "@langchain/openai": "^0.4.4",

afirstenberg:
Cleanup responseToChatGenerations() so it is more readable.
The combineGenerations() function has much of its work, but also splits things up so that content, text, toolCalls, and kwargs are now combined in a more reasonable way.

The previous implementation only handled cases where there was one function call and there was no text content along with it. And it did it in a huge loop over all the parts that was difficult to untangle.
These changes clean things up, separate some of the logic so it is clearer what is going on, and handles multiple parts better. It also sets things up so future changes can handle new part types easier.

Fixes #7705


libs/langchain-google-common/src/utils/gemini.ts
```diff
@@ -799,9 +798,18 @@ export function getGeminiAPI(config?: GeminiAPIConfig): GoogleAIAPI {
   function partToChatGeneration(part: GeminiPart): ChatGeneration {
     const message = partToMessageChunk(part);
     const text = partToText(part);
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const generationInfo: Record<string, any> = {};
+
+    // Handle some of the "reasoning" elements
+    // if (Object.hasOwn(part, "thought")) {
```
jacoblee93, line 805: Should we uncomment?

libs/langchain-google-common/src/utils/gemini.ts
```diff
@@ -867,49 +875,168 @@ export function getGeminiAPI(config?: GeminiAPIConfig): GoogleAIAPI {
...
+
+    // Build the message and the generation chunk to return
+    const message = new AIMessageChunk({
+      content: combinedContent,
+      additional_kwargs: kwargs,
+      usage_metadata: usageMetadata,
+      tool_calls: combinedToolCalls.tool_calls,
+      invalid_tool_calls: combinedToolCalls.invalid_tool_calls,
+    });
+    console.log(message);
```
jacoblee93, line 993: Stray log

libs/langchain-google-common/src/utils/gemini.ts
```diff
@@ -867,49 +875,168 @@ export function getGeminiAPI(config?: GeminiAPIConfig): GoogleAIAPI {
...
+  function combineText(gen: ChatGeneration[]): string {
+    return gen.map((item) => item.text ?? "").join("");
+  }
+
+  function combineToolCalls(gen: ChatGeneration[]): AIMessageChunk {
+    const ret = new AIMessageChunk("");
+
+    gen.forEach((item: ChatGeneration) => {
+      const message: AIMessageChunk = item?.message as AIMessageChunk;
+      ret.tool_calls!.push(...(message?.tool_calls ?? []));
```
jacoblee93, line 924: Is this really what we want to do vs. just chunks?

Will each `item` have unique tool calls?

We do have a `concat` method internally that handles this:

```ts
import { concat } from "@langchain/core/utils/stream";

const combinedChunk = concat(chunk1, chunk2);
```

jacoblee93 (commented):
Thank you! See comments

libs/langchain-google-common/src/utils/gemini.ts
afirstenberg: The reasoning elements have been removed from Gemini (at least for the moment), with no word if that is a permanent removal or not. The code I needed was from the branch I was working on for that, however.

This should probably just be removed for now, tho part of me is hopeful that they'll wise up and I can uncomment this.

afirstenberg: I've deleted this comment

libs/langchain-google-common/src/utils/gemini.ts
afirstenberg: Stray log removed

libs/langchain-google-common/src/utils/gemini.ts
afirstenberg: I think it is (although I've changed it to use `concat()`. Certainly this seems to be at least part of the issue that was reported.

I've added a comment to make it more clear that the `AIMessageChunk` isn't really the objective here - it is just a convenient object to carry all the tool calling related arrays.

afirstenberg:
> Thank you! See comments

And thank you for reviewing! Hopefully these three updates address all the issues.

PR #7726, opened 2025-02-19
fix(core): Fix issue in .d.ts typing for TextEncoder

### LINKED ISSUE #7247: 'TextEncoder' refers to a value, but is being used as a type here. Did you mean 'typeof TextEncoder'?
### Checked other resources

### Example Code

```ts
// bytes.d.ts
import { BaseTransformOutputParser } from "./transform.js";
/**
 * OutputParser that parses LLMResult into the top likely string and
 * encodes it into bytes.
 */
export declare class BytesOutputParser extends BaseTransformOutputParser<Uint8Array> {
    static lc_name(): string;
    lc_namespace: string[];
    lc_serializable: boolean;
    protected textEncoder: TextEncoder;
    parse(text: string): Promise<Uint8Array>;
    getFormatInstructions(): string;
}
```

Found 1 error in `node_modules/@langchain/core/dist/output_parsers/bytes.d.ts`:

### Description

Building my typescript project fails after installing @langchain/core. Have tried deleting node_modules and reinstalling all packages with `npm install` 

Adding `typeof TextEncoder` like the error suggests to the file in  `node_modules/@langchain/core/dist/output_parsers/bytes.d.ts` does indeed fix this problem, so I believe this is an error in the npm release itself.

### System Info

@langchain 0.3.6
@langchain/core 0.3.18
platform: Window
Node v22.11.0
NPM v10.9.0

danielkatz:
Fixes #7247 without loosing the typing

This pull request makes a small but important change to the `BytesOutputParser` class in the `langchain-core` package. The change involves updating the type of the `textEncoder` property to use a more specific type.

`globalThis.TextEncoder` (when not imported from `node:util`) is a value, not a type. In this case, the type is actually `InstanceType<typeof TextEncoder>`.

* [`langchain-core/src/output_parsers/bytes.ts`](diffhunk://#diff-d38c7f72b0e15d5088141cc1fcbaa2dfb397a6107d941ce71e780e972e05529bL16-R16): Updated the `textEncoder` property to use `InstanceType<typeof TextEncoder>` instead of `any` for better type safety.

### DIFF
--- a/langchain-core/src/output_parsers/bytes.ts
+++ b/langchain-core/src/output_parsers/bytes.ts
@@ -13,9 +13,7 @@ export class BytesOutputParser extends BaseTransformOutputParser<Uint8Array> {
 
   lc_serializable = true;
 
-  // TODO: Figure out why explicit typing is needed
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  protected textEncoder: any = new TextEncoder();
+  protected textEncoder: InstanceType<typeof TextEncoder> = new TextEncoder();
 
   parse(text: string): Promise<Uint8Array> {
     return Promise.resolve(this.textEncoder.encode(text));


jacoblee93:
Nice! Makes sense

Any chance this would miss some other edge case? Given it's internal facing seems like some risk for not much reward

danielkatz:
Thanks for the feedback.

I encountered the same issue and, after researching, found the #7247 issue without a real solution.
Once I figured it out for myself, I decided to submit this PR to help others. I’ve been using this solution in production without encountering any related problems.

jacoblee93:
Ok thank you!

PR #7733, opened 2025-02-21
fix(google-genai): Fix Google Genai usage token

### LINKED ISSUE #7444: google-genai: Streaming mode returns empty tokenUsage object
### Checked other resources

## Steps to Reproduce

1.    Initialize a ChatGoogleGenerativeAI instance with streaming mode enabled (streaming: true).
2.   Call a public method like invoke and pass callbacks with `handleLLMEnd` that includes a console.log to inspect the `llmOutput` field.
3.   Observe that the `estimatedTokenUsage` object in the `llmOutput` field is empty ({}).

Example code to reproduce:
```typescript
    model: "gemini-1.5-flash",
    temperature: 1.0,
    streaming: true,
  });

  const result = await model.pipe(parser).invoke(formattedPrompt, {
    callbacks: [
      {
        handleLLMEnd: (output) => {
          console.log(output);
        },
      },
    ],
  });
```

{
  generations: [ [ [ChatGenerationChunk] ] ],
  llmOutput: { estimatedTokenUsage: {} }
}

### Description

When using the ChatGoogleGenerativeAI class in streaming mode, the tokenUsage object is always returned as an empty object, even though the usage metadata is expected to be updated during the streaming process.

From analyzing the code, it appears that tokenUsage is not being updated within the _generate method in streaming mode. The chunk data from _streamResponseChunks includes usageMetadata, but it is not being added to the tokenUsage object.

### System Info

node v20.18.1

beowulf11:
Fix corrupted token count (output & total) for Google Genai provider in streaming mode.

Problem was that when using streaming mode we assume `candidatesTokenCount` and `totalTokenCount` are present. They are not initially so the value gets corrupted (`NaN`) and this error carries forward.

Fixes #7444

### DIFF
--- a/libs/langchain-google-genai/src/chat_models.ts
+++ b/libs/langchain-google-genai/src/chat_models.ts
@@ -905,21 +905,21 @@ export class ChatGoogleGenerativeAI
         options.streamUsage !== false
       ) {
         const genAIUsageMetadata = response.usageMetadata as {
-          promptTokenCount: number;
-          candidatesTokenCount: number;
-          totalTokenCount: number;
+          promptTokenCount: number | undefined;
+          candidatesTokenCount: number | undefined;
+          totalTokenCount: number | undefined;
         };
         if (!usageMetadata) {
           usageMetadata = {
-            input_tokens: genAIUsageMetadata.promptTokenCount,
-            output_tokens: genAIUsageMetadata.candidatesTokenCount,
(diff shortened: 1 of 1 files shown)


jacoblee93 (approved):
Thank you, sorry for the delayed review!

beowulf11:
Np, Thanks for the approval.

PR #7740, opened 2025-02-23
feat(community): Support google cloud storage document loader

AllenFang:
self explanatory, tested in local. Our platform is GCP, it's a bit helpful to developer if langchain could support cloud storage. 
Please review. thanks!

### DIFF
--- a/libs/langchain-community/src/document_loaders/web/google_cloud_storage.ts
+++ b/libs/langchain-community/src/document_loaders/web/google_cloud_storage.ts
@@ -1,0 +1,99 @@
+import { Storage, StorageOptions } from '@google-cloud/storage'
+import * as os from "node:os";
+import * as path from "node:path";
+import * as fsDefault from "node:fs";
+import { BaseDocumentLoader } from "@langchain/core/document_loaders/base";
+import { Document } from "@langchain/core/documents";
+import { UnstructuredLoaderOptions, UnstructuredLoader } from "../fs/unstructured.js";
+
+
+/**
+ * Represents the parameters for the GoogleCloudStorageLoader class. It includes
+ * properties such as the GCS bucket, file destination, the options for the UnstructuredLoader and
+ * the options for Google Cloud Storage
+ */
+export type GcsLoaderConfig = {
+  fs?: typeof fsDefault;
+  bucket: string;
+  file: string;
+  unstructuredLoaderOptions: UnstructuredLoaderOptions;
+  storageOptions: StorageOptions
+}
+
+/**
+ * A class that extends the BaseDocumentLoader class. It represents a
+ * document loader for loading files from a google cloud storage bucket.
+ * @example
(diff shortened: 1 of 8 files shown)


libs/langchain-community/src/document_loaders/web/google_cloud_storage.ts
```diff
@@ -0,0 +1,99 @@
...
+    super();
+    this._fs = fs;
+    this.bucket = bucket;
+    this.file = file;
+    this.unstructuredLoaderOptions = unstructuredLoaderOptions;
+    this.storageOptions = storageOptions;
+  }
+
+  async load(): Promise<Document<Record<string, any>>[]> {
+    const tempDir = this._fs.mkdtempSync(
```
jacoblee93, line 69: nit: could we do this without the filesystem as a passthrough?

jacoblee93 (approved):
Thank you! Small question but can merge as is

libs/langchain-community/src/document_loaders/web/google_cloud_storage.ts
```diff
@@ -0,0 +1,104 @@
...
+ * Represents the parameters for the GoogleCloudStorageLoader class. It includes
+ * properties such as the GCS bucket, file destination, the options for the UnstructuredLoader and
+ * the options for Google Cloud Storage
+ */
+export type GcsLoaderConfig = {
+  fs?: typeof fsDefault;
+  bucket: string;
+  file: string;
+  unstructuredLoaderOptions: UnstructuredLoaderOptions;
+  storageOptions: StorageOptions;
```
afirstenberg, line 22: StorageOptions seems to be used just for authentication. Does this also work with Application Default Credentials? If so - this should be an optional field.

libs/langchain-community/src/document_loaders/web/google_cloud_storage.ts
AllenFang: I just follow the s3 loader solution, but I think it is possible if `UnstructuredLoader` support buffer.

I check the source code, it seems like `UnstructuredLoader` support either file path or buffer, so in theory, for s3, azure and gcp storage, we can just read the file as buffer and assign to `UnstructuredLoader`

I can take a look and make an enhancement when I have time @jacoblee93

libs/langchain-community/src/document_loaders/web/google_cloud_storage.ts
AllenFang: @afirstenberg  you are right, if I remember correctly, the `Storage` can accept nothing and it will use the Application Default Credentials as default auth approach.

AllenFang: fixed: https://github.com/langchain-ai/langchainjs/pull/7791

PR #7750, opened 2025-02-24
feat(anthropic): Support claude 3.7 sonnet & extended thinking

benjamincburns:


libs/langchain-anthropic/src/chat_models.ts
```diff
@@ -766,6 +782,30 @@ export class ChatAnthropicMessages<
...
+        throw new Error("topK is not supported when thinking is enabled");
+      }
+      if (this.topP !== -1) {
+        throw new Error("topP is not supported when thinking is enabled");
+      }
+      if (this.temperature !== 1) {
+        throw new Error(
+          "temperature is not supported when thinking is enabled"
+        );
+      }
```
benjamincburns, line 796: we're using magic values for these rather than leaving them `undefined` in the instance vars, so I had to drop them here, and I didn't want to drop them silently.

libs/langchain-anthropic/src/chat_models.ts
```diff
@@ -766,6 +782,30 @@ export class ChatAnthropicMessages<
...
+
+      return {
```
jacoblee93, line 798: Do we need a separate return statement here?

jacoblee93 (approved):
Seems reasonable, one question

libs/langchain-anthropic/src/chat_models.ts
benjamincburns: ha, I was clarifying as you were asking: https://github.com/langchain-ai/langchainjs/pull/7750/files/eba780ee7ff498774ac885a16c7bec271142e51e#r1968553519

jacoblee93: Ah ok

Maybe we should also ship a minor version that drops the defaults but that can wait

jacoblee93 (approved):
LGTM if tests pass

PR #7754, opened 2025-02-25
fix(deepseek): wrong home link for @langchain/deepseek

AllenFang:
the `Homepage` link from [npm repository](https://www.npmjs.com/package/@langchain/deepseek) seems wrong


jacoblee93 (approved):
Oof, thanks!

PR #7768, opened 2025-02-27
feat(aws): support reasoning blocks for claude 3.7

benjamincburns:


benjamincburns:
reminder that I'll need to bump the version of core that AWS depends on when #7769 gets released.

libs/langchain-aws/src/types.ts
```diff
@@ -19,3 +19,26 @@ export type BedrockToolChoice =
   | ToolChoice.AutoMember
   | ToolChoice.ToolMember;
 export type ChatBedrockConverseToolType = BindToolsInput | BedrockTool;
+
+export type MessageContentReasoningBlockReasoningText = {
+  type: "reasoningContent";
```
jacoblee93, line 24: These are our internal representations?

Why don't we just preserve the ones returned from Bedrock?

jacoblee93: Ok - only other concern then is to try to keep it inline with changes in Python as well

benjamincburns: Not sure who's picking up the python one, but we're consciously _not_ forcing structural similarity between the two libs anymore unless there's a good reason to (e.g. they need to interop). Winds up causing too many pythonic conventions that feel weird in JS

jacoblee93 (approved):

PR #7771, opened 2025-02-27
feat(runnables): invoke runnable without inputs

saharis9988:
Currently, triggering a chain or runnable without parameters results in a TypeScript error. However, there are valid use cases where we want to allow this behavior.
This PR introduces a proper solution to enable parameter-less execution without relying on `@ts-ignore` comments.
Just a recommendation for improvement.
Let me know if you have any suggestions or alternative recommendations! 🚀


jacoblee93:
Thanks for the PR @saharis9988!

This seems niche and I don't think making the input seem optional is good in most cases. Could you instead change the generic `RunInput` somehow to infer `undefined` if it's created from an empty function or something?

What is the use case for this in general?

saharis9988:
Hi @jacoblee93 
sorry for the late response, was a little busy in the lat two weeks.
I have a scenario in which I need to trigger an api request without any parameters (depends on the current date).
Then I make some transformations on the data returns back and I use the data as PromptTemplate params.
So this chain is not including any additional args.
I can escape the error but I dont want to add ts escape comments to my production code.

saharis9988:
Hi @jacoblee93,
Sorry for the late response — I've been a bit busy over the past two weeks.
I have a scenario where I need to trigger an API request without any parameters (based on the current date). After that, I perform some transformations on the returned data and use it as parameters in a PromptTemplate. As a result, this chain doesn't include any additional arguments.
I can suppress the error using TypeScript escape comments, but I'd prefer to avoid that in production code..

jacoblee93:
Can you just pass an empty object?

saharis9988:
It works, but it’s not very clear.
From a DX perspective, I would expect the framework to handle this automatically.
We mainly need this for RunnableLambdas. If we want the type to be more dynamic based on the scenario, we could use type inference to automatically determine the `RunInput` from the first argument passed to the aRunnableFunc` and conditionally enable optional input when no arguments are provided.
Alternatively, we could use an empty literal object as the default for the chain invocation.
But isn’t this a good enough reason to switch to optional inputs?

saharis9988:
It works, but it’s not very clear.
From a DX perspective, I would expect the framework to handle this automatically.
We mainly need this for `RunnableLambda`. If we want the type to be more dynamic based on the scenario, we could use type inference to automatically determine the RunInput from the first argument passed to the `RunnableFunc` and conditionally enable optional input when no arguments are provided.
Alternatively, we could use an empty literal object as the default for the chain invocation.
But isn’t this a good enough reason to switch to optional inputs?

saharis9988:
@jacoblee93 WDYT?
image

jacoblee93:
If you can fix the typing properly so it's not optional for the majority of use cases I could get on board with it, but wouldn't want to degrade the typing for others

PR #7776, opened 2025-02-27
fix(core): Adds fallback for context vars

jacoblee93:
CC @ibolmo


langchain-core/src/singletons/async_local_storage/index.ts
```diff
@@ -78,6 +78,9 @@ class AsyncLocalStorageProvider {
       previousValue !== undefined &&
       previousValue[_CONTEXT_VARIABLES_KEY] !== undefined
     ) {
+      if (runTree === undefined) {
```
ibolmo, line 81: confirmed this did the trick for me

ibolmo (approved):

jacoblee93:
Thank you for raising!

PR #7779, opened 2025-02-28
feat(libs): add @langchain/milvus

zhanshuyou:


jacoblee93:
Hey @zhanshuyou, sorry for the delay!

Would you be up for self-publishing and managing your own integration package? Would allow you to ship changes more quickly with us out of the loop (though if you CC someone on the team on PRs we'd of course be happy to have a look, and you could also add us as maintainers to any private repo).

We're then happy to deprecate the `@langchain/community` implementation and update docs to point at your new repo.

PR #7780, opened 2025-02-28
Google oauth support + fix tools for API v1

deepu105:
- Adds support for OAuth access tokens. This is a much better option for use in applications for end users to use their own email account via federated Oauth token exchange using IdPs supporting the feature
- Fixes the search, get message and get thread tools to work properly with google gmail API v1

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,3 +48,5 @@ docs/api_refs/typedoc.json
 credentials.json
 **/typedoc.json
 .vercel
+
+bun.lock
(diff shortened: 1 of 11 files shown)


deepu105:
@benjamincburns same as the calender PR. I assumed the tool can be used in a singleton as well so made the access token a string or function so that token refresh and validation can be handled on the agent side

benjamincburns (approved):

PR #7781, opened 2025-02-28
feature(community): Enable using chat and llms without providing project/space/development id (lightweight engine IBM)

FilipZmijewski:
Providing small changes in order to allow user to use inference methods in both LLM and chat class

### DIFF
--- a/libs/langchain-community/src/chat_models/ibm.ts
+++ b/libs/langchain-community/src/chat_models/ibm.ts
@@ -465,11 +465,6 @@ export class ChatWatsonx<
     )
       throw new Error("Maximum 1 id type can be specified per instance");
 
-    if (!("projectId" in fields || "spaceId" in fields || "idOrName" in fields))
-      throw new Error(
-        "No id specified! At least id of 1 type has to be specified"
-      );
-
     if ("model" in fields) {
       this.projectId = fields?.projectId;
       this.spaceId = fields?.spaceId;
@@ -564,13 +559,18 @@ export class ChatWatsonx<
   scopeId():
     | { idOrName: string }
     | { projectId: string; modelId: string }
-    | { spaceId: string; modelId: string } {
+    | { spaceId: string; modelId: string }
+    | { modelId: string } {
     if (this.projectId && this.model)
       return { projectId: this.projectId, modelId: this.model };
     else if (this.spaceId && this.model)
       return { spaceId: this.spaceId, modelId: this.model };
     else if (this.idOrName) return { idOrName: this.idOrName };
-    else throw new Error("No scope id provided");
+    else if (this.model)
+      return {
+        modelId: this.model,
+      };
+    else throw new Error("No id or model provided!");
   }
 
   async completionWithRetry<T>(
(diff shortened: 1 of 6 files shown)


libs/langchain-community/src/chat_models/ibm.ts
```diff
@@ -465,11 +465,6 @@ export class ChatWatsonx<
     )
       throw new Error("Maximum 1 id type can be specified per instance");
 
-    if (!("projectId" in fields || "spaceId" in fields || "idOrName" in fields))
-      throw new Error(
-        "No id specified! At least id of 1 type has to be specified"
-      );
```
benjamincburns, line 471: Can we keep this check and only bypass it when the lightweight engine is in use?

libs/langchain-community/src/llms/ibm.ts
```diff
@@ -184,11 +184,6 @@ export class WatsonxLLM<
     )
       throw new Error("Maximum 1 id type can be specified per instance");
 
-    if (!("projectId" in fields || "spaceId" in fields || "idOrName" in fields))
-      throw new Error(
-        "No id specified! At least id of 1 type has to be specified"
-      );
```
benjamincburns, line 190: Same question here

libs/langchain-community/src/llms/tests/ibm.test.ts
```diff
@@ -156,17 +153,6 @@ describe("LLM unit tests", () => {
...
-      const testPropsServiceUrl: WatsonxInputLLM = {
-        serviceUrl: process.env.WATSONX_AI_SERVICE_URL as string,
-      };
-      expect(
-        () =>
-          new WatsonxLLM({
-            ...testPropsServiceUrl,
-            ...fakeAuthProp,
-          })
-      ).toThrowError();
```
benjamincburns, line 169: Assuming it's possible to check for the lightweight engine, you could keep this logic here, but pass whatever is necessary to detect that use

libs/langchain-community/src/chat_models/ibm.ts
FilipZmijewski: So that would enforce us to use a flag ex. ```lightweight: true```. Unless user uses a lightweight engine the API is going to throw an error "No spaceId nor projectId is present" so this is taken care of there.
Also, we want to match with our Python Langchain SDK implementation. Assuming whenever id is not passed user chooses lightweight engine. If id of any type is mistakenly not passed the API will throw an error letting user know.

benjamincburns: @FilipZmijewski just to clarify, are you saying that the API _does_ check this and therefore the check in our code is unnecessary? The benefit of having the check in our code is that it causes us to throw a nice/readable error. Is there no other way that we can infer that the user is using the lightweight engine?

By the way, feel free to reply in your native language if you're more comfortable that way. LLMs are very good at translating these days 😄

FilipZmijewski: @benjamincburns We would like to match the solution with what we have in SDK (both python and node.js) and also langchain IBM python meaning "no id provided -> lightweight engine in use". In case of lightweight engine unavailable the API will throw an error, quite clear to understand, `Missing either space_id or project_id or wml_instance_crn`. One thing that could be added is disabling error check depending on `serviceUrl` provided but this is not ideal since not all `IBM watsonx.ai software` instalations will allow user to use lightweight engine. 

Haha, thanks. I enjoy speaking/writing in English so that works!

benjamincburns (approved):
Thanks for the clarifications!

Issue #7778, opened 2025-02-28
Error in Claude 3.7 Sonnet thinking mode when using tool calls


kishan0725:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain.js documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain.js rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

This is the conversation JSON I got after the initial invoke call
```
[
  {
    "type": "system",
    "data": {
      "content": "system message goes here...",
      "additional_kwargs": {},
      "response_metadata": {}
    }
  },
  {
    "type": "human",
    "data": {
      "content": "human message goes here...",
      "additional_kwargs": {},
      "response_metadata": {}
    }
  },
  {
    "type": "ai",
    "data": {
      "content": [
        {
          "type": "reasoning_content",
          "reasoningText": {
            "signature": "EqoBCkgIARABGAIiQLtO0oqIjC3WgiutslITJmn9qsGko9Ecx2b7DbWRUJwBgyyjHLzK9Xm8Ij4edIYG0DYApxlGJgqEFX0bHU7HgvJAOw+fC87q6k=",
            "text": "reasoning text goes here..."
          }
        },
        {
          "type": "text",
          "text": "ai text goes here..."
        }
      ],
      "tool_calls": [
        {
          "id": "tooluse_xEX1OcCYT0eUrOhwIZsEWg",
          "name": "search_files",
          "args": {
            "path": "/Users/kishan/Documents/codebase",
            "pattern": "FileName\\.aspx$"
          },
          "type": "tool_call"
        }
      ],
      "response_metadata": {
        "$metadata": {
          "httpStatusCode": 200,
          "requestId": "1ed632ba-9688-4f4c-a173-46028d110508",
          "attempts": 1,
          "totalRetryDelay": 0
        },
        "metrics": {
          "latencyMs": 10655
        },
        "stopReason": "tool_use",
        "usage": {
          "cacheReadInputTokenCount": 0,
          "cacheReadInputTokens": 0,
          "cacheWriteInputTokenCount": 0,
          "cacheWriteInputTokens": 0,
          "inputTokens": 4215,
          "outputTokens": 332,
          "totalTokens": 4547
        }
      },
      "usage_metadata": {
        "input_tokens": 4215,
[...]

PR #7785, opened 2025-03-01
fix(aws): fix content block ordering for claude reasoning on bedrock

benjamincburns:
Fixes: #7778

### DIFF
--- a/libs/langchain-aws/src/tests/chat_models.int.test.ts
+++ b/libs/langchain-aws/src/tests/chat_models.int.test.ts
@@ -469,3 +469,57 @@ test("Test reasoning_content blocks multiturn streaming", async () => {
   // test a second time to make sure that we've got input translation working correctly
   await doStreaming(streamingMessages);
 });
+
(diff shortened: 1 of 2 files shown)


libs/langchain-aws/src/tests/chat_models.int.test.ts
```diff
@@ -469,3 +469,56 @@ test("Test reasoning_content blocks multiturn streaming", async () => {
...
+  // console.log("result.tool_calls?.[0]", result.tool_calls?.[0]);
+
+  expect(typeof result.tool_calls![0]).toBe("object");
+  expect(result.tool_calls![0].name).toBe("weather_poet");
+
+  expect(typeof result.tool_calls![0].id).toBe("string");
+  expect(result.tool_calls![0].id!.length).toBeGreaterThan(0);
+
+  expect(typeof result.tool_calls![0].args).toBe("object");
+  expect(result.tool_calls![0].args.location).toBe("London");
```
benjamincburns, line 513: eh, checking this feels like it could get flaky if the model decides to go above and beyond and e.g. specify "London, England". I'll relax this check and just look for `string`.

PR #7788, opened 2025-03-01
Google oauth support for Google calender tool

deepu105:
- Added access token support via Google OAuth
- Added keyfile support for consistency with Gmail tool
- Fix prompts since output was not in correct format


deepu105:
@benjamincburns let me know if there is anything to be addressed here. I assumed the tool can be used in a singleton as well so made the access token a string or function so that token refresh and validation can be handled on the agent side

benjamincburns (approved):

benjamincburns:
Looks great - I appreciate the ability to set the accessKey using a function - should make this very flexible. Thanks for the contribution @deepu105!

PR #7789, opened 2025-03-02
fix(community): adding structured output parser for Google Calendar

John2360:
Adding a structured output parser for Google Calendar view and event creation. This removes the possibility that the LLM returns more than just the formatted JSON and causes a JSON parsing error.


jacoblee93 (commented):
Thanks for the PR!

If we're going to do this - may be best to try to use `.withStructuredOutput` and throw for models that don't support it?

https://js.langchain.com/docs/how_to/structured_output/#the-.withstructuredoutput-method

John2360:
> Thanks for the PR!
> 
> If we're going to do this - may be best to try to use `.withStructuredOutput` and throw for models that don't support it?
> 
> https://js.langchain.com/docs/how_to/structured_output/#the-.withstructuredoutput-method

It makes sense to me! I went ahead and pushed a new commit with those changes.

jacoblee93 (approved):
Thank you! Sorry for the delay!

PR #7793, opened 2025-03-03
feat (google-common,vertex,etc): Support Claude 3.7 Thinking via Vertex AI

afirstenberg:
This adds direct support for Anthropic Claude 3.7 Thinking through Vertex AI


libs/langchain-google-vertexai/src/tests/chat_models.int.test.ts
```diff
@@ -655,113 +656,231 @@ describe("Express Gemini Chat", () => {
...
+      const request = recorder?.request ?? {};
+      const data = request?.data;
+      expect(data).toHaveProperty("tools");
+      expect(data.tools).toHaveLength(1);
+
+      expect(result.tool_calls).toHaveLength(1);
+      expect(result.tool_calls?.[0]).toBeDefined();
+      expect(result.tool_calls?.[0].name).toBe("get_weather");
+      expect(result.tool_calls?.[0].args).toHaveProperty("location");
     });
```
benjamincburns, line 738: On the bedrock integration I initially forgot to add a test that sent the `ToolMessage` back to the LLM after a tool call. This caused us to overlook an issue where the message translation logic was placing `thinking` content blocks after tool calls in the API-specific messages.

It may be worth updating this test, and the one below it, to send the `ToolMessage` back to ensure that this API error doesn't occur here, too. The content block ordering on AI messages is an upstream requirement from Anthropic, so unless Google is handling it in their proxy, I'm assuming it's possible to encounter that error here, too.

See #7785 for reference.

benjamincburns (changes requested):
Looks good overall, just ideally would need one relatively small addition to the tests.

libs/langchain-google-vertexai/src/tests/chat_models.int.test.ts
afirstenberg: I've added new tests to check for this

afirstenberg:
> Looks good overall, just ideally would need one relatively small addition to the tests.

I added two new tests to check for the ordering issue.
Along the way - I discovered a bug in how the tool response is sent and corrected that.

benjamincburns (approved):

benjamincburns:
Looks great - thanks for keeping after this!

PR #7794, opened 2025-03-03
tests(google-vertex,webauth): Verify code works with gemini-2.0-flash-lite-001

afirstenberg:
No code changes - just tests.

Verify code works with gemini-2.0-flash-lite-001 (or that failures are documented and understood).

(Flash Lite doesn't support every feature, and these are documented.)


benjamincburns (approved):

benjamincburns:
@afirstenberg happy to merge this one as-is - just curious if it's worth updating the integrations doc page to reflect these details, or at least whether these details are easily discovered from outbound links from that page?

PR #7822, opened 2025-03-10
feat(community): Permit bedrock application inference profiles

bioshazard:
Fixes #7809 

Bedrock Chat Model did not have a way to invoke Application Inference Profiles. Solving this was complicated at first... how do you keep all that modelId-derived metadata parsing?

So I added a `modelAlias` field that simply overrides `this.model` to be used in the /invoke URL. It expects to be URL encoded. Worked great in my local testing. You can find me on Twitter [@bios_hazard](https://x.com/bios_hazard)


AllenFang:
invoke approach is complicated indeed but this patch looks hacky and like a workaround. I will see what I can do when i have time unless langchain team is okay for this patch. 🙏

bioshazard:
Thanks, I think it's actually pretty clean given the meta data is parsed from the model id. All that is unavailable in the interface id arn. Curious to hear how you could do it another way with that limitation. I expect you'd need to do an intermediate lookup or supply a second field as I have.

libs/langchain-community/src/utils/bedrock/index.ts
```diff
@@ -117,6 +117,12 @@ export interface BaseBedrockInput {
   */
   model: string;
 
+  /** Optional URL Encoded model alias to use. Necessary for invoking an Application Inference Profile
+      For example, "arn%3Aaws%3Abedrock%3Aus-east-1%3A1234567890%3Aapplication-inference-profile%2Fabcdefghi", will override this.model in final /invoke URL call.
+      Must still provide `model` as normal modelId to benefit from all the metadata.
+  */
+  modelAlias?: string;
```
jacoblee93, line 124: Should we call this `applicationInferenceProfile`?

jacoblee93 (commented):
Thanks! Small question

libs/langchain-community/src/utils/bedrock/index.ts
bioshazard: Works for me. Figured to make it generic tho only application inference profile is the only use case today.

bioshazard: @jacoblee93 friendly bump, hoping to stay w/ LangChain as we migrated to App Inf profiles. Gotta deliver this soon. Thanks!

jacoblee93: Will try to get this in today thanks!

jacoblee93 (approved):
Thanks! Though note to others who might find this that it is better to switch to Bedrock's newer Converse API instead in `@langchain/aws`

bioshazard:
> Thanks! Though note to others who might find this that it is better to switch to Bedrock's newer Converse API instead in `@langchain/aws`

Does that package support App Inf profiles yet? I probably need to pitch the same PR as I use that lib for embeddings.

jacoblee93:
Thank you!

PR #7826, opened 2025-03-11
fix(aws): fix `Invalid reasoning content` exception on empty reasoning chunk

ChenNima:
When invoking the DeeSeek R1 model on Bedrock, there exists a possibility that the model may generate reasoningContent with an  empty text string.

```json
{
  "reasoningContent": {
    "text": ""
  }
}
```

Modify the code to prevent the `Invalid reasoning content` exception from being thrown in this scenario.

### DIFF
--- a/libs/langchain-aws/src/common.ts
+++ b/libs/langchain-aws/src/common.ts
@@ -563,7 +563,7 @@ export function bedrockReasoningDeltaToLangchainPartialReasoningBlock(
   | MessageContentReasoningBlockReasoningTextPartial
   | MessageContentReasoningBlockRedacted {
   const { text, redactedContent, signature } = reasoningContent;
-  if (text || typeof text === 'string') {
+  if (typeof text === 'string') {
     return {
       type: "reasoning_content",
       reasoningText: { text },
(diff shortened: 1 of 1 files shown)


libs/langchain-aws/src/common.ts
```diff
@@ -563,7 +563,7 @@ export function bedrockReasoningDeltaToLangchainPartialReasoningBlock(
   | MessageContentReasoningBlockReasoningTextPartial
   | MessageContentReasoningBlockRedacted {
   const { text, redactedContent, signature } = reasoningContent;
-  if (text) {
+  if (text || typeof text === 'string') {
```
jacoblee93, line 566: nit: can we just do the `typeof` check?

jacoblee93 (commented):
Thanks for this! Small question

jacoblee93 (approved):
Just made the change, thank you!

libs/langchain-aws/src/common.ts
ChenNima: Yes, I believe the "typeof text" check suffices. I just made the minimum change necessary to resolve the issue, while striving to avoid disrupting any other scenarios.

PR #7829, opened 2025-03-11
Use the config value semanticConfigurationName in AzureAISearchVectorStore

keiohtani:
`semanticConfigurationName` is an argument that AzureAISearchVectorStore takes at the initialization, but in contrast to what you would expect from the argument's name, it is not used anywhere, and it's confusing. This PR includes the change to use the parameter as the semantic configuration name for Azure AI Search instead of the static text. 

Here is the interface for the option parameter. 

```ts
export interface AzureAISearchQueryOptions {
  readonly type?: AzureAISearchQueryType;
  readonly semanticConfigurationName?: string;
}
```

### DIFF
--- a/libs/langchain-community/src/vectorstores/azure_aisearch.ts
+++ b/libs/langchain-community/src/vectorstores/azure_aisearch.ts
@@ -484,7 +484,7 @@ export class AzureAISearchVectorStore extends VectorStore {
       top: k,
       queryType: "semantic",
       semanticSearchOptions: {
-        configurationName: "semantic-search-config",
+        configurationName: this.options.semanticConfigurationName,
       },
     });
 
@@ -663,7 +663,7 @@ export class AzureAISearchVectorStore extends VectorStore {
         defaultConfigurationName: "semantic-search-config",
         configurations: [
           {
-            name: "semantic-search-config",
+            name: this.options.semanticConfigurationName,
             prioritizedFields: {
               contentFields: [
                 {


libs/langchain-community/src/vectorstores/azure_aisearch.ts
```diff
@@ -484,7 +484,7 @@ export class AzureAISearchVectorStore extends VectorStore {
       top: k,
       queryType: "semantic",
       semanticSearchOptions: {
-        configurationName: "semantic-search-config",
+        configurationName: this.options.semanticConfigurationName,
```
jacoblee93, line 487: I think this should keep the default to avoid breaking changes?

jacoblee93:
Thanks! Let's maybe go with:

```
this.options.semanticConfigurationName ?? "semantic-search-config"
```

sinedied:
@keiohtani Thanks for catching this! Though the original intent behind this parameter is not the one you're pushing here: it's aimed to be able to use a different semantic configuration if you're using an existing schema, and not meant as a way to rename the default schema configuration.
We're still discussing what would be the best way to support custom schema in this integration.

keiohtani:
> @keiohtani Thanks for catching this! Though the original intent behind this parameter is not the one you're pushing here: it's aimed to be able to use a different semantic configuration if you're using an existing schema, and not meant as a way to rename the default schema configuration. We're still discussing what would be the best way to support custom schema in this integration.

Exactly. In my team, we are using the custom configuration name and at the moment, the only way for us to modify this value is to replace the string with our custom name after the build, and it would be great if we can avoid such a hacky solution. Updated the PR with the suggestion. Please let me know if anything else is needed to merge!

libs/langchain-community/src/vectorstores/azure_aisearch.ts
```diff
@@ -663,7 +663,7 @@ export class AzureAISearchVectorStore extends VectorStore {
         defaultConfigurationName: "semantic-search-config",
         configurations: [
           {
-            name: "semantic-search-config",
+            name: this.options.semanticConfigurationName ?? "semantic-search-config",
```
sinedied, line 666: This one should not change, as it's meant to be the default one: if the name changes here, the default schema will break.
This parameter was only meant (for now) to be used alongside a custom schema, not to override the default schema

jacoblee93: Bump on this @keiohtani!

I will apply this in a few days to move on on this.

jacoblee93:
Thank you!

jacoblee93 (approved):

PR #7831, opened 2025-03-12
feat[openai]: responses API

dqbd:
Fixes # (issue)


libs/langchain-openai/src/chat_models.ts
```diff
@@ -200,7 +202,127 @@ export function _convertMessagesToOpenAIParams(
...
+      }
+
+      if (isAIMessage(message) && !!message.tool_calls?.length) {
+        return message.tool_calls.map(
+          (toolCall): OpenAIResponsesInputItem => ({
+            type: "function_call",
+            name: toolCall.name,
+            arguments: JSON.stringify(toolCall.args),
+            call_id: toolCall.id!,
+            id: toolCall.id!,
```
baskaryan, line 238: these are different i believe. 'id' is the object id, 'call_id' is what you'll use to refer to this tool call when returning the output

Screenshot 2025-03-11 at 5 58 04 PM

baskaryan: 'call_id' maps to what we store as toolCall.id

baskaryan: for now was thinking of throwing the new 'id' for all tool calls in AIMessage.additionalKwargs['openai_tool_call_ids'] or something like that

baskaryan: smth like
```python
            if tool_calls := lc_msg.tool_calls:
                if not lc_msg.additional_kwargs.get("__openai_function_call_ids__"):
                    raise ValueError(...)
                function_call_ids = lc_msg.additional_kwargs["__openai_function_call_ids__"]
                for tool_call in tool_calls:
                    function_call = {
                        "type": "function_call",
                        "name": tool_call["name"],
                        "arguments": tool_call["arguments"],
                        "call_id": tool_call["id"],
                        "id": function_call_ids[tool_call["id"]],
                    }
                    updated_payload["input"].append(function_call)

```

langchain/package.json
```diff
@@ -457,6 +457,7 @@
     "openai": "^4.41.1",
     "peggy": "^3.0.2",
     "prettier": "^2.8.3",
+    "reflect-metadata": "^0.2.2",
```
jacoblee93, line 460: Is this an accident?

libs/langchain-openai/src/chat_models.ts
```diff
@@ -200,7 +202,375 @@ export function _convertMessagesToOpenAIParams(
...
+    };
+  } else if (chunk.type === "response.refusal.done") {
+    additional_kwargs.refusal = chunk.refusal;
+  } else {
+    return null;
+  }
+
+  return new ChatGenerationChunk({
+    // TODO: why do I need to do this?
+    text: content.map((part) => part.text).join(""),
```
jacoblee93, line 509: Legacy reasons, plus I think `onLLMNewToken` pulls this out

langchain/package.json
dqbd: Nope! Turns out that to make tests pass, just needed to add that dep

PR #7842, opened 2025-03-13
feat(community): add deepseek to allowed bedrock models

jarib:
See https://aws.amazon.com/blogs/aws/deepseek-r1-now-available-as-a-fully-managed-serverless-model-in-amazon-bedrock/

IMHO this validation of model providers is redundant. Bedrock will let the user know if the model is invalid. Strictly enforcing it in client libs makes no sense to me.

### DIFF
--- a/libs/langchain-community/src/chat_models/bedrock/web.ts
+++ b/libs/langchain-community/src/chat_models/bedrock/web.ts
@@ -72,6 +72,7 @@ const ALLOWED_MODEL_PROVIDERS = [
   "cohere",
   "meta",
   "mistral",
+  "deepseek",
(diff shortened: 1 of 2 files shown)


jacoblee93:
I agree with you - want to just remove them?

jacoblee93:
Actually I notice this is not present in `@langchain/aws` which is what most people should be using going forwards - let's just merge this as is

jarib:
Ah, I wasn't aware of @langchain/aws. Will look into it. Fixed the formatting failure here anyway :)

jacoblee93 (approved):
Thank you!

PR #7844, opened 2025-03-14
fixed bug when tool call throws Exception that like [llm/error] [1:llm:ChatOpenAI] [399ms] LLM run errored with error: "400 [Bad Request] Validation error for body application/json: No schema matches\n\nError: 400 [Bad Request] Validation error for body application/json: No schema matches

awol2005ex:
fixed bug when tool call throws Exception that like [llm/error] [1:llm:ChatOpenAI] [399ms] LLM run errored with error: "400 [Bad Request] Validation error for body application/json: No schema matches\n\nError: 400 [Bad Request] Validation error for body application/json: No schema matches

### DIFF
--- a/libs/langchain-openai/src/chat_models.ts
+++ b/libs/langchain-openai/src/chat_models.ts
@@ -166,13 +166,13 @@ export function _convertMessagesToOpenAIParams(
     }
     if (message.additional_kwargs.function_call != null) {
       completionParam.function_call = message.additional_kwargs.function_call;
-      completionParam.content = null;
+      completionParam.content = "";
     }
     if (isAIMessage(message) && !!message.tool_calls?.length) {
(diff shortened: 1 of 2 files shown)


jacoblee93 (approved):
Thanks!

libs/langchain-openai/src/chat_models.ts
```diff
@@ -166,13 +166,13 @@ export function _convertMessagesToOpenAIParams(
     }
     if (message.additional_kwargs.function_call != null) {
       completionParam.function_call = message.additional_kwargs.function_call;
-      completionParam.content = null;
+      completionParam.content = "";
```
jacoblee93, line 169: Actually, would prefer if we don't change OpenAI in this PR. Can you revert?

jacoblee93 (approved):
Thanks!

PR #7845, opened 2025-03-14
feat(google-common): add responseModalities parameter for multimodal output.

tofuliang:
feat: add responseModalities parameter for multimodal output.


jacoblee93:
Thanks! Seems fine to me - CC @afirstenberg

jacoblee93 (approved):

libs/langchain-google-common/src/types.ts
```diff
@@ -123,6 +123,16 @@ export interface GoogleAISafetySetting {
 
 export type GoogleAIResponseMimeType = "text/plain" | "application/json";
 
+export const GoogleAIResponseModality = {
+  ModalityUnspectified: "MODALITY_UNSPECIFIED",
```
afirstenberg, line 127: Unspecified isn't necessary. It is a result of their generated code, but it should never be used.

afirstenberg:
Thank you! Appreciate you doing this before I had a chance to.

Before we release, I think we need at least two sets of things:
* Update to the documentation
* At least one test with the mocked response in the google-common package and one integration test in each of google-webauth and google-vertex.

If you don't, I'll try to get to them later this weekend.

libs/langchain-google-common/src/types.ts
```diff
@@ -229,6 +239,12 @@ export interface GoogleAIModelParams {
    * logprobs must be set to true if this parameter is used.
    */
   topLogprobs?: number;
+
+  /**
+   * The modalities of the response.
+   */
+  responseModalities?: GoogleAIModelModality[];
```
afirstenberg, line 246: At least in the IntelliJ IDEA IDE, this is represented as a type of `string[] | unknown` which doesn't help with code completion.

For others, we've made the equivalent of `GoogleAIModelModality` something like `"TEXT" | "IMAGE" | "AUDIO" | string`. Is that reasonable to do here?

Thoughts, @jacoblee93 or @benjamincburns?

afirstenberg:
I couldn't update the PR directly, but here's a diff that fixes some formatting and includes a test.

I didn't update documentation since I have a question about our approach.

[pr-7845.diff.txt](https://github.com/user-attachments/files/19277028/pr-7845.diff.txt)

libs/langchain-google-common/src/types.ts
jacoblee93: Yeah I'd go with that!

maruhidd:
I need it.

jacoblee93:
Thank you both!

PR #7846, opened 2025-03-14
fix(RunnableLambda): set default value for input parameter in `invoke…

saharis9988:
continues #7771 (pull request)


saharis9988:
Summary:
I initially implemented dynamic inference of function arguments to update the argument types of the `invoke` function. The core of the implementation involved the following utility types:
```
type HasAtleastOneArg<T extends (...args: any[]) => unknown> = 
  Parameters<T> extends [arg:infer FirstArg,...args:unknown[]] ? FirstArg : false;

type ShloudIncludeParams<T extends (...args: unknown[]) => any,ReturnType = any,AdditionalParams extends unknown[] = unknown[]> = HasAtleastOneArg<T> extends false ? () => ReturnType : (value:HasAtleastOneArg<T>,...args:AdditionalParams) => ReturnType;
```
However, this approach required modifying the base class, which was problematic because we only needed this support in the `RunnableLambda` child class.
To address this, I refined the solution to ensure it remains compatible with the base type while specifically targeting this use case. 
This solution avoids breaking changes and maintains flexibility where needed.

jacoblee93:
If it builds I suppose this might be okay

saharis9988:
@jacoblee93 build succeeded

jacoblee93:
I think we would want this for `.stream()` as well

CC @benjamincburns what do you think?

benjamincburns:
@jacoblee93 `RunInput` isn't constrained to extend a type to which `{}` is assignable, so this change isn't valid. I don't think we want to add that constraint, either.

saharis9988:
Hi @benjamincburns 🙂
An empty literal object is a valid RunInput.
We’re just providing a default value when `invoke` is called with no arguments (meaning the lambda function doesn’t receive any arguments). Alternatively, we could initialize it with an empty string or `undefined`, as those would also be valid.
Additionally, if your function receives a primitive value, you can still pass it without any issues—the type system won’t prevent it.

benjamincburns:
Will this value be passed to the function if the user forgets to pass an input?

benjamincburns:
@saharis9988 sorry for my brief replies before. I appreciate what you're trying to do here, but unfortunately this change isn't versatile enough to work for the broad range of cases that `RunnableLambda` has to support.

For example, compare the behaviour of the following code before and after the change in this PR:

```ts
const rl = RunnableLambda.from((val?: number) => (val ?? 0) + 1);
// @ts-expect-error - empty input is fine here
const rlResult = await rl.invoke();
expect(rlResult).toEqual(1);
```

The type checker is unhappy, but the code executes fine prior to your change.

I haven't tested this, but I suspect that the correct way to fix this would be to add overload to `Runnable` that treats `input` as optional in the case where `RunInput extends undefined`, like in the following example:

```ts
export abstract class Runnable<
    RunInput = any,
    RunOutput = any,
    CallOptions extends RunnableConfig = RunnableConfig
  >
  extends Serializable
  implements RunnableInterface<RunInput, RunOutput, CallOptions>
{

  ...

  abstract invoke(
    input?: RunInput,
    options?: Partial<CallOptions>
  ): RunInput extends undefined ? Promise<RunOutput> : never;

  abstract invoke(
    input: RunInput,
    options?: Partial<CallOptions>
  ): Promise<RunOutput>;

  ...
}
```

Note that returning `never` in the case where `RunInput` _doesn't_ extend `undefined` makes it so that the argument is always required by the TS compiler in cases when `RunInput` is non-optional.

benjamincburns:
Closing, per the above comment.

@saharis9988 if you want to try out the solution I proposed in a new PR, with updates to the affected types, I'd be much obliged.

This would technically be a breaking change, but it'd only break types that extend `Runnable`, arguably for the better.

cc @jacoblee93 @nfcampos (just in case this has been discussed previously and ruled out for some reason)

benjamincburns:
The other way to do this would be to replace `invoke/stream/streamEvents` with a `withConfig` method that returns a callable that returns something like a [`Promiventerator`](https://github.com/benjamincburns/promiventerator), but that would be a big, big breaking change.

saharis9988:
Hi @benjamincburns,
First of all, I initially tried adding another overload as you suggested, but it ended up breaking all the extended child classes. Overloads work at the type level, but the implementation overload is not exposed externally, which limits their effectiveness in this case.
I believe we should find a solution that doesn’t break the API. 
The scenario you pointed out is valid, and it highlights that `{}` isn't the right default parameter initializer.
But what about explicitly passing `undefined`? This approach would align with TypeScript's type checking while avoiding unintended behavior.

On a separate note, I have a personal request regarding the PR process. This is the second time I’ve had to open a PR for the same issue. I actively work with your framework daily and want to contribute to improving it, but having to reopen PRs makes the process more difficult. In the future, I’d appreciate it if you could wait for my response before closing a PR.

benjamincburns:
Hi @saharis9988,

Sorry if closing the PR feels harsh, and I should have thought through my remarks more carefully last night before replying, as this is a topic that has been considered extensively both internally and externally.

Regarding the PR state, bear in mind that we use PRs a bit like a todo list to track incomplete/pending work. I opted to close this one rather than submitting a review with changes requested because there aren't any changes that I can reasonably request to make it so that we'd be able to move forward with this PR.

Unless you're willing to put in a _monumental_ amount of effort on this, I don't see a path forward for you to resolve the specific problem of allowing an invocation of `RunnableLambda` with zero arguments. This is because there is no default value that can be specified for `input` that wouldn't also require additional breaking changes to be made to `Runnable`. As you're likely aware, the "blast radius" associated with this sort of change would be huge, so we'd only want to do this if the value to making such a change would far outstrip the cost.

To address your proposal, a default value of `undefined` for `input` _does_ feel closer to correct, as it captures what's actually passed when the user passes zero arguments, but then we would need to change the signature of `invoke` (and the other related methods) to be `abstract invoke(input: RunInput | undefined, options?: Partial<CallOptions>): Promise<RunOutput>;`.  Alternatively, we'd need to constrain the definition of `RunInput` to be `RunInput extends undefined`.

We'd need to choose one of these breaking changes regardless of the default value you decide to specify for `input`. Doing anything other than this invalidates the core value proposition of implementing LangChain JS in TypeScript.

For what it's worth, I appreciate that always having to pass some value to `invoke`, `stream`, etc is annoying. We use `Runnable` and its descendants internally every day. We'd _love_ to improve the developer experience in this area, but the reality is that it's just not possible without a breaking change. But really if we're going to make a breaking change to something as core to LangChain as `Runnable` we should consider it quite carefully and produce a substantially improved design rather than just addressing one specific issue.

I'm a bit biased, as it's my proposal, but so far the best design that I've seen is the one that I mentioned briefly [in this comment](https://github.com/langchain-ai/langchainjs/pull/7846#issuecomment-2751035335). It cleanly separates the invocation of user code from the configuration of the execution runtime, and it allows for Runnable invocations that accept any number of arguments. It also gets rid of the combinatorial "arms race" that we have around the various `invoke`/`stream`/`streamEvents` methods by returning a single object from a single method that allows the user to decide how they'd prefer to consume the result of the task (either as an awaitable `Promise`, as an `AsyncGenerator`, or as an `EventEmitter`).

I'm happy to keep considering any less costly alternative changes that you might propose (you may think of something that we haven't!), but in the interim this PR needs to stay closed, as there's no further pending actions for either you or our team at this stage.

saharis9988:
First, thanks for the detailed comment—I really appreciate the thoughtfulness in your response.
The goal of this PR was to address the issue specifically for RunnableLambda, as the pain point is particularly significant there. 
I understand that the problem is broader, but my focus here was to improve the DX for this specific case without introducing breaking changes. 
I agree that a more comprehensive solution might be needed in the long run, and I’d be happy to contribute to that effort.
That said, I didn’t fully understand your point about `RunInput` needing to extend `undefined`. 
Currently, the default type for RunInput is `any`, and there are no restrictions on it. 
You can see this [here](https://github.com/langchain-ai/langchainjs/blob/main/langchain-core/src/runnables/base.ts#L120).
Since `any` can be assigned **any** value—including `undefined`—and there are no generic constraints that prevent this, we can safely initialize the argument as `undefined` in this specific case without affecting other parts of the system. Given that this change applies only to `RunnableLambda`, I don’t see why we can’t proceed with this quick win for now to improve DX without introducing any breaking changes..

@jacoblee93, I’d love to hear your thoughts on this as well.

benjamincburns:
Replying from my phone, so I'll be somewhat brief. The _default_ type for RunInput is `any`, but the type in use will be inferred from the signature of the `func`. If that function doesn't have a first arg that extends `undefined`, a default input of `undefined` breaks the inferred constraints on `input`.

In order to make the types be constrained we'd need to force users to only construct a `RunnableLambda` with a `func` that has a first argument that is unioned with `undefined`. The only way to do that is to make the breaking changes that I mentioned in my previous comment.

We also can't really accept a solution that changes the signature of `RunnableLambda` without considering `Runnable`. `RunnableLambda` and friends are really only meant to be typed as such in the code that produces them. Consumer code is meant to depend on `Runnable`. They must implement the same interface for key methods like `invoke` and `stream` or it breaks the `Runnable` abstraction.

saharis9988:
So if I understand you correctly, you're saying that after the change I suggest, TypeScript's type inference won't correctly recognize the type of input when it's passed as the first argument to the function, unless the function's parameter type explicitly is unioned with undefined.

So in that case, you expect the TypeScript compiler to throw an error?
```typescript
const runnable = RunnableLambda.from((input: number) => input);

...

await runnable.invoke(); // you say TypeScript should complain here since `undefined` isn't part of the input type

```

saharis9988:
So if I understand you correctly, you're saying that after the change I suggest, TypeScript's type inference won't correctly recognize the type of input when it's passed as the first argument to the function, unless the function's parameter type explicitly is unioned with undefined.

So in that case, you expect the TypeScript compiler to throw an error?
```typescript
const runnable = RunnableLambda.from((input: number) => input);
// you say TypeScript should complain here since `undefined` isn't part of the input type
```

benjamincburns:
Moreover, in the code snippet you gave, we would require TypeScript to produce an error if we were imposing a default input of `undefined`. If it didn't, we've made a mistake in our specification of the library's public interfaces.

benjamincburns:
Apologies. I deleted a comment because my remarks weren't clear. I said in that comment that whether or not the compiler raises an error doesn't matter, but that's not what I meant.

What I meant was that the lack of an error in the example that you gave doesn't mean that the change you are recommending is acceptable.  Rather, you should infer precisely the opposite. The lack of the type check error is why we would need to make additional breaking changes in order to correctly specify the type requirements for `Runnable`.

This is ultimately because changing this signature to specify a default value is a breaking change. For any default value that you might specify, I can come up with an example runnable that works fine today that the default value would break.

saharis9988:
@benjamincburns Still waiting for your example..

I agree that for any value other than `undefined`, this change would break the behavior. However, I don’t see how it would break for `undefined`...

Let me know if I'm missing something 🙏

saharis9988:
The goal of this PR was to resolve the optional-like behavior for RunnableLambda specifically.
This is how TypeScript infers the type after my change, and it aligns with our expectations:
image

I understand that in the base Runnable interface, the first argument is mandatory, but TypeScript does not complain because, in TypeScript, a parameter with a default value is still considered required from the type system's perspective. This means we are not breaking the contract—we are simply extending the behavior only for this specific case where it is needed.
That said, I won’t continue to push the discussion further, as I see that this isn't a direction you want to take. Ultimately, it’s your decision as the maintainer.
However, I strongly believe that this issue must be addressed as soon as possible, before the framework moves out of its zero major version..
CC: @jacoblee93

jacoblee93:
Hey @saharis9988, will add it to our list when there's next a breaking change but agree with Ben about potential blast radius/benefit ratio at the moment. Thanks for the respectful discussion!

PR #7851, opened 2025-03-16
feat(google-common): Gemma 3 support for AI Studio

afirstenberg:
AI Studio supports gemma-3-27b-it through the AI Studio API.
This provides basic support through the `google-gauth` and `google-webauth` packages.

Supported:
* Text invocation
* Streaming response
* System instructions (not natively supported in gemma - implemented by creating pseudo human and model messages).

Known issues (with AI Studio, not apparently with LangChainJS):
* Image support
* Tool support


docs/core_docs/docs/integrations/platforms/google.mdx
```diff
@@ -153,7 +154,43 @@ Click [here](/docs/integrations/chat/google_vertex_ai) for the `@langchain/googl
 
 The value of `image_url` must be a base64 encoded image (e.g., `data:image/png;base64,abcd124`).
 
-### Non-Gemini Models
+### Gemma
+
+Access the `gemma-3-27b-it` model through AI Studio using the `ChatGoogle` class.
+(This class is a superclass of the [`ChatVertexAI`](/docs/integrations/chat/google_vertex_ai)
```
jacoblee93, line 160: I'd prefer if we just stick to using via Google VertexAI and then switch all docs over to the superclasses if we are confident we want to do this rather than doing it piecemeal

libs/langchain-google-common/src/utils/common.ts
```diff
@@ -214,6 +217,7 @@ export function validateModelParams(
   const model = testParams.model ?? testParams.modelName;
   switch (modelToFamily(model)) {
     case "gemini":
+    case "gemma": // TODO: Are we sure?
```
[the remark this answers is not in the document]
afirstenberg: We're at least sure when it comes to AI Studio.

docs/core_docs/docs/integrations/platforms/google.mdx
afirstenberg: The issue is that Gemma-as-a-Service isn't available on Vertex. Just on AI Studio.

People can use the Vertex packages, but force using AI Studio with a setting (which undoes the only thing these packages do).

I'd love to get to the superclasses, personally... but I think that's a bigger question than just getting some fixes in place and some minimal documentation to support Gemma on the AI Studio API.

jacoblee93: Got it - thank you!

jacoblee93 (approved):
Thanks!

PR #7852, opened 2025-03-17
feat(google-cloud-sql): Add VectorStore, ChatMessageHistory, Loader for Google Cloud SQL

averikitsch:
Fixes # (issue)


libs/langchain-google-cloud-sql-pg/src/indexes.ts
```diff
@@ -0,0 +1,35 @@
...
+    this.searchFunction = searchFunction;
+    this.indexFunction = indexFunction;
+  }
+}
+
+/**
+ * Enumerator of the Distance strategies.
+ */
+export class DistanceStrategy extends StrategyMixin {
+  public static EUCLIDEAN =  new StrategyMixin("<->", "l2_distance", "vector_l2_ops");
```
jacoblee93, line 19: Don't forget to run `yarn format`

libs/langchain-google-cloud-sql-pg/src/indexes.ts
```diff
@@ -0,0 +1,35 @@
...
+    this.operator = operator;
+    this.searchFunction = searchFunction;
+    this.indexFunction = indexFunction;
+  }
+}
+
+/**
+ * Enumerator of the Distance strategies.
+ */
+export class DistanceStrategy extends StrategyMixin {
```
jacoblee93, line 18: I'm a bit confused by this pattern - why have `DistanceStrategy` itself extend the mixin?

libs/langchain-google-cloud-sql-pg/src/chatMessageHistory.ts
```diff
@@ -0,0 +1,143 @@
...
+   * @param {string} schemaName Schema name for the chat message history table. Default: "public".
+   * @returns PostgresChatMessageHistory instance.
+   */
+  static async create(
+    engine: PostgresEngine,
+    sessionId: string,
+    tableName: string,
+    schemaName: string = "public"
+  ) {
+    const query = `SELECT column_name, data_type FROM information_schema.columns WHERE table_name = '${tableName}' AND table_schema = '${schemaName}'`;
```
jacoblee93, line 46: May be worth a small note in docs emphasizing that `tableName` and `schemaName` in these files are not escaped and to not pass in end-user input here

libs/langchain-google-cloud-sql-pg/src/loader.ts
```diff
@@ -0,0 +1,180 @@
...
+      } else if (format === "json") {
+        formatFunc = jsonFormatter;
+      } else {
+        formatFunc = textFormatter;
+      }
+    }
+
+    let queryStmt = query;
+    if (!queryStmt) {
+      queryStmt = `SELECT * FROM "${schemaName}"."${tableName}"`;
```
jacoblee93, line 109: Probably good to set some kind of limit here?

libs/langchain-google-cloud-sql-pg/src/vectorStore.ts
```diff
@@ -0,0 +1,423 @@
...
+          valuesStmt += `, :${metadataColumn}`;
+          values[metadataColumn] = metadata[metadataColumn]
+          delete extra[metadataColumn]
+        } else {
+          valuesStmt += " ,null"
+        }
+      }
+
+      // Add JSON column and/or close statement
+      stmt += this.metadataJsonColumn ? `, ${this.metadataJsonColumn})` : ")";
```
jacoblee93, line 285: OOC why not just support `metadataJsonColumn` vs spreading metadata into other columns?

jacoblee93 (commented):
Thanks for this! Looks good overall, mostly nits and some suggestions for simplifying

Another option I'll float is self-publishing the package itself outside of the LangChain repo and just merging docs here - that way you don't have to wait on me in the review loop

libs/langchain-google-cloud-sql-pg/src/loader.ts
averikitsch: The user can use the table argument to pull the whole table. I will add a note that we suggest using the lazyLoad.

libs/langchain-google-cloud-sql-pg/src/vectorStore.ts
averikitsch: This pattern follows database best practices for relational databases by using individual columns versus all data in a JSON column. This allows for low latency filtering and type checking.

libs/langchain-google-cloud-sql-pg/src/engine.ts
```diff
@@ -0,0 +1,271 @@
...
+  }
+
+  /**
+   * Create a Cloud SQL table to store chat history.
+   *
+   * @param tableName Table name to store chat history
+   * @param schemaName Schema name to store chat history table
+   */
+
+  async initChatHistoryTable(tableName: string, schemaName: string = "public"): Promise<void> {
```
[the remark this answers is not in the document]
averikitsch: We believe that users should be aware of the database tables they are creating and this is not abstracted away which can cause confusion for long term maintenance of the vector store. This pattern allows separation of workflows ie a specific user can create the table and set up the vector store with specific permissions, then the vector store can have reduce permissions when running in an application.

libs/langchain-google-cloud-sql-pg/src/engine.ts
```diff
@@ -0,0 +1,271 @@
...
+    if (typeof this.dataType !== "string") {
+      throw Error("Column data_type must be type string");
+    }
+
+  }
+}
+
+const USER_AGENT = "langchain-google-cloud-sql-pg-js";
+
+export class PostgresEngine {
```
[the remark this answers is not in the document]
averikitsch: I will add docstrings. The engine is a shared object so the connection pool is shared. This is a postgres best practice.

libs/langchain-google-cloud-sql-pg/src/indexes.ts
averikitsch: This allows us to create an enum like object where we can really easily call the get the different attributes

libs/langchain-google-cloud-sql-pg/src/vectorStore.ts
jacoblee93: And the JSON column is just to make onboarding easier?

libs/langchain-google-cloud-sql-pg/src/engine.ts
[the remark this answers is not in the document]
jacoblee93: Fair enough!

libs/langchain-google-cloud-sql-pg/src/engine.ts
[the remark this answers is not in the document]
jacoblee93: Ok, sounds good. Yeah definitely better to do it this way for prod, just more pieces and classes to set up for getting started

libs/langchain-google-cloud-sql-pg/src/vectorStore.ts
averikitsch: Yes and we don't want users to lose metadata if not specifically defined

averikitsch:
@jacoblee93 I've force pushed to fix the merge conflict. I've also added a couple more index methods to libs/langchain-google-cloud-sql-pg/src/vectorStore.ts.

nahidvay123483 (commented):
Ndjend 3ueniebdi3 djdnid
D
Ejeniren
Dneiek uejei3nodnjdkdmdhdn(.didndlndbd) djend djebjenej ejenienuenirn uenod
Ebueneuebgneoebhdjehdoe.dh hrneoenrgienuebubfjeoendv
D🇧🇩bangladcshsneiwns euneeibeie privahensjs ueenoene

libs/langchain-google-cloud-sql-pg/src/chat_message_history.ts
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    const requiredColumns = ["id", "session_id", "data", "type"];
+
+    if (!requiredColumns.every((x) => columnNames.includes(x))) {
+      throw new Error(
+        `Table '${schemaName}'.'${tableName}' has incorrect schema.
+        Got column names ${columnNames} but required column names ${requiredColumns}.\n
+        Please create table with following schema: \nCREATE TABLE '${schemaName}'.'${tableName}' (
+        \n    id SERIAL AUTO_INCREMENT PRIMARY KEY,
```
jacoblee93, line 79: nit: do yo uneed the newlines here? It's already a multiline string

libs/langchain-google-cloud-sql-pg/src/engine.ts
```diff
@@ -0,0 +1,315 @@
...
+    return new PostgresEngine(PostgresEngine._createKey, engine);
+  }
+
+  /**
+   * Create a PostgresEngine instance from an Knex instance.
+   *
+   * @param engine knex instance
+   * @returns PostgresEngine instance from a knex instance
+   */
+  static async fromEngine(engine: knex.Knex) {
```
jacoblee93, line 182: nit: `PostgresEngine.fromEngine` is a little strange

jacoblee93 (approved):
Thank you! Will ship, won't block on nits since I've been slow.

jacoblee93:
Hey @averikitsch, getting a bunch of test failures in CI like this:

```
```

averikitsch:
> Hey @averikitsch, getting a bunch of test failures in CI like this:
> 
> ```
>   ● ChatMessageHistory methods › should clear all messages added to the store
> 
>     connect ECONNREFUSED 127.0.0.1:5432
> ```
> 
> We can either add mocking or change them to integration tests (rename to `x.int.test.ts`)

I will update the name and copy over the docker container file https://github.com/langchain-ai/langchainjs/blob/54b639d57c47921244b5a02083c3a61c9a4ebb97/libs/langchain-community/src/vectorstores/tests/pgvector/docker-compose.yml

jacoblee93:
Just to keep it lighterweight - could we try using https://node.testcontainers.org/modules/postgresql/?

averikitsch:
> Just to keep it lighterweight - could we try using https://node.testcontainers.org/modules/postgresql/?

Yes, sgtm

libs/langchain-google-cloud-sql-pg/src/engine.ts
averikitsch: This adds a lot of flexibility for developers to use a connection pool they are already using and have configured for their application

jacoblee93: Ya just the naming is funny

averikitsch: You are right, this didn't translate well from python. I have updated it to be `fromPool`

averikitsch:
@jacoblee93 could you help me with the yarn.lock conflict?

jacoblee93:
Thanks for getting this to the finish line!

jacoblee93:
Will ship later today and ping here! Thanks so much

PR #7853, opened 2025-03-17
fix(azure-cosmosdb): Update user agent of AzureCosmosDBMongoDBSemanticCache and AzureCosmosDBMongoChatMessageHistory

fatmelon:
Add UA in order to track the usage of Azure Mongo vCore in langchainjs


jacoblee93 (approved):
Thanks! FYI @sinedied

PR #7865, opened 2025-03-19
Add Mem0 Memory Integration

whysosaket:
Adds Mem0 Memory Adapter

[Mem0](https://github.com/mem0ai/mem0) is a self-improving memory layer for LLM applications, enabling personalized AI experiences that save costs and delight users.


jacoblee93 (approved):
Looks good, thanks!

Though keep in mind that we generally prefer people use LangGraph rather than these older `Memory` classes

whysosaket:
> Looks good, thanks!
> 
> Though keep in mind that we generally prefer people use LangGraph rather than these older `Memory` classes

Hi @jacoblee93, thank you for merging this. I understand why people would prefer LangGraph, and I'm looking forward to contributing to it.

PR #7874, opened 2025-03-22
fix(core): Add typing in ChatMessagePromptTemplate.fromTemplate

gouroujo:
When using the static method fromTemplate, this PR allows the type of the input to be strictly enforced.

When using the following example :

```typescript
  const chatPrompt = ChatPromptTemplate.fromTemplate<{foo: string, bar: string}>("Hello {foo}, I'm {bar}");
  const messages = await chatPrompt.formatPromptValue({
    foo: "Foo",
    bar: "Bar",
  });
```

If an input is missing you got a type error `Property 'foo' is missing in type '{ bar: string; }' but required in type 'TypedPromptInputValues<{ foo: string; bar: string; }>'`

Screenshot 2025-03-21 at 18 19 32

I have reused the declarations found in `PromptTemplate.fromTemplate` with the eslint-disabled for the banned type `Symbol` to stay consistent.

Let me know if you have any feedback.

Thanks for the great work !

### DIFF
--- a/langchain-core/src/prompts/chat.ts
+++ b/langchain-core/src/prompts/chat.ts
@@ -336,13 +336,13 @@ export class ChatMessagePromptTemplate<
     return new ChatMessage(await this.prompt.format(values), this.role);
   }
 
-  static fromTemplate(
-    template: string,
-    role: string,
-    options?: { templateFormat?: TemplateFormat }
-  ) {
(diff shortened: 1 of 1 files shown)


jacoblee93:
Thanks! You shouldn't have to do this in most cases but good to allow.

PR #7879, opened 2025-03-24
feat(google-common): Allow passing cachedContent params via Google Vertex

jacoblee93:
CC @afirstenberg


libs/langchain-google-common/src/types.ts
```diff
@@ -257,6 +257,16 @@ export interface GoogleAIModelRequestParams extends GoogleAIModelParams {
    * If empty, any one of the provided functions are called.
    */
   allowed_function_names?: string[];
+
+  /**
+   * Used to specify a previously created context cache to use with generation.
+   * Should be of the form: "projects/PROJECT_NUMBER/locations/LOCATION/cachedContents/CACHE_ID",
+   *
```
afirstenberg, line 264: What is the format for this value for the AI Studio platform?

libs/langchain-google-common/src/types.ts
```diff
@@ -257,6 +257,16 @@ export interface GoogleAIModelRequestParams extends GoogleAIModelParams {
...
+   * See these guides for more information on how to use context caching:
+   * https://cloud.google.com/vertex-ai/generative-ai/docs/context-cache/context-cache-create
```
afirstenberg, line 266: My biggest concern is that I don't think there is a way create or maintain the cache through LangChainJS - just use it.

afirstenberg (commented):
Overall, I don't see any real issues with the update. Most of the questions more discuss the future of this feature, but that shouldn't stop it from going in.
One question about code style.

libs/langchain-google-common/src/types.ts
```diff
@@ -257,6 +257,16 @@ export interface GoogleAIModelRequestParams extends GoogleAIModelParams {
...
+   * https://cloud.google.com/vertex-ai/generative-ai/docs/context-cache/context-cache-use
+   */
+  cachedContent?: string;
```
afirstenberg, line 269: Maybe `cachedContentId` is a better term? Since it isn't the content itself?

libs/langchain-google-common/src/types.ts
jacoblee93: Will add a node that this is for Vertex

libs/langchain-google-vertexai/src/tests/chat_models.int.test.ts
```diff
@@ -625,6 +625,22 @@ describe.each(testGeminiModelNames)("GAuth Gemini Chat (%s)", (modelName) => {
   });
 });
 
+test("Context caching", async () => {
+  const model = new ChatVertexAI({
+    model: "gemini-1.5-pro-002",
+    location: "us-east5",
```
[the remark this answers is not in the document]
jacoblee93: Didn't seem to work with 2.0 Flash when I tried it. Said it "wasn't stable".

libs/langchain-google-common/src/utils/common.ts
```diff
@@ -175,6 +175,10 @@ export function copyAIModelParamsInto(
     ret.tools = convertToGeminiTools(tools as Record<string, any>[]);
   }
 
+  if (options?.cachedContent) {
+    ret.cachedContent = options.cachedContent;
+  }
+
```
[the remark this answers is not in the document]
jacoblee93: This one doesn't really make sense to pass into the constructor IMO

libs/langchain-google-common/src/types.ts
jacoblee93: I wanted to stick with the Google field name

PR #7890, opened 2025-03-24
feat(community): TavilySearchResults: Add `apiUrl` field

svennergr:
We'd like to use a proxy between our app and the Tavily API. This PR adds an optional `apiUrl` field to the `TavilySearchResults` class in order to control the API host used.


jacoblee93 (approved):
Thank you, sorry for the delay!

PR #7895, opened 2025-03-25
fix(community): Tool indexing in Watsonx chat class

FilipZmijewski:
This fixes an issue that occurred when n>1 in chat call


jacoblee93 (approved):
Thanks!

PR #7902, opened 2025-03-26
test(google): Gemini 2.5

afirstenberg:
* Add Gemini 2.5 Pro Experimental to test runs against AI Studio
* Adjust tests to remove Max Token setting for streaming tests since it was triggering no output on "thinking" model. 
* Fix bug in placement of multimodal output testing.

Note that tests weren't done against Vertex since it isn't supported there yet.

No code changes.
**No need for new release.**

### DIFF
--- a/libs/langchain-google-webauth/src/tests/chat_models.int.test.ts
+++ b/libs/langchain-google-webauth/src/tests/chat_models.int.test.ts
@@ -339,6 +339,17 @@ const testGeminiModelNames = [
     platformType: "gcp",
     apiVersion: "v1",
   },
+  {
+    modelName: "gemini-2.5-pro-exp-03-25",
+    platformType: "gai",
+    apiVersion: "v1beta",
+  },
+  // Gemini 2.5 not available on Vertex yet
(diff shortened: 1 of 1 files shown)


jacoblee93 (approved):
Thank you! Sorry for the delay

afirstenberg:
AUGH! Just before I was going to push an update. {:
(Which is still just tests, no actual code change.)

damsorian:
@afirstenberg do you know how to use image generation like Imagen3 with Google Vertex and LangChain.JS ? there are a module wrapper for that? or only for text generation?

afirstenberg:
> @afirstenberg do you know how to use image generation like Imagen3 with Google Vertex and LangChain.JS ? there are a module wrapper for that? or only for text generation?

@damsorian - There are a lot of answers here, depending on what you're actually looking for. But in the comments of an unrelated, closed, patch don't seem like the best place for it. Perhaps open a new issue to specifically raise the issue with details of what you're looking to do, and we can discuss the best approach?

PR #7907, opened 2025-03-26
fix: handle subgraph nesting better in graph_mermaid

jasonphillips:
Currently, `graph_mermaid` can throw an error when subgraphs are deeply nested, due to its way of tracking subgraph names and its order of processing edges. We encountered this in production with a large graph which had a couple of levels of subgraphs and many edges.

See the nodes/edges fixture added in `graph_mermaid.test.ts`, which is a minimal reproduction of one part of our graph, and will throw the following error on main branch:

```
Found duplicate subgraph 'fooGenericSearch' -- this likely means that you're reusing a subgraph node with the same name. Please adjust your graph to have subgraph nodes with unique names
```

The graph (and original langgraph code that generated it) is valid, and doesn't have a duplicate subgraph, but the subgraph/prefix handling misrecognizes it as duplicate when traversing and tracking the set of subgraphs.

This PR fixes that, and can now render the graph correctly; and also pre-orders the edges so that subgraphs will appear in their correct nesting, which wasn't necessarily the case before.

Without the additional edge reordering, the test graph here was not properly nested:
image

due to incorrect ordering of subgraphs (excerpt):
```
      subgraph fooTool
      subgraph fooGenericSearch
         ...
      end
      subgraph fooSearchSource
         ...
      end
        ...
      end
```

With the fix, it is, thanks to ordering edges by depth before traversal:
image

```
          subgraph fooTool
          subgraph fooSearchSource
          subgraph fooGenericSearch
              ...
          end
              ...
          end
              ...
          end
```

I was able to test this rendering on our very large application graph and there were no unintended side effects / regressions; would be happy to have even more test cases for example graphs if needed--this particular script had little in the way of existing tests.

### DIFF
--- a/langchain-core/src/runnables/graph_mermaid.ts
+++ b/langchain-core/src/runnables/graph_mermaid.ts
@@ -97,20 +97,36 @@ export function drawMermaid(
 
   const seenSubgraphs = new Set<string>();
 
+  // sort prefixes by path length for correct nesting
+  function sortPrefixesByDepth(prefixes: string[]): string[] {
+    return [...prefixes].sort((a, b) => {
+      return a.split(":").length - b.split(":").length;
+    });
+  }
+
   function addSubgraph(edges: Edge[], prefix: string): void {
     const selfLoop = edges.length === 1 && edges[0].source === edges[0].target;
     if (prefix && !selfLoop) {
       const subgraph = prefix.split(":").pop()!;
-      if (seenSubgraphs.has(subgraph)) {
-        throw new Error(
-          `Found duplicate subgraph '${subgraph}' -- this likely means that ` +
-            "you're reusing a subgraph node with the same name. " +
-            "Please adjust your graph to have subgraph nodes with unique names."
-        );
+
+      if (seenSubgraphs.has(prefix)) {
+        return;
       }
 
-      seenSubgraphs.add(subgraph);
+      seenSubgraphs.add(prefix);
       mermaidGraph += `\tsubgraph ${subgraph}\n`;
+    }
+
+    // all nested prefixes for this level, sorted by depth
+    const nestedPrefixes = sortPrefixesByDepth(
+      Object.keys(edgeGroups).filter(
+        (nestedPrefix) =>
+          nestedPrefix.startsWith(`${prefix}:`) && nestedPrefix !== prefix
+      )
+    );
+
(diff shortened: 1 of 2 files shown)


jacoblee93:
Nice catch, thank you!

langchain-core/src/runnables/graph_mermaid.ts
```diff
@@ -97,22 +97,38 @@ export function drawMermaid(
...
     if (prefix && !selfLoop) {
       const subgraph = prefix.split(":").pop()!;
-      if (seenSubgraphs.has(subgraph)) {
-        throw new Error(
-          `Found duplicate subgraph '${subgraph}' -- this likely means that ` +
-            "you're reusing a subgraph node with the same name. " +
-            "Please adjust your graph to have subgraph nodes with unique names."
-        );
+
+      if (seenSubgraphs.has(prefix)) {
```
jacoblee93, line 112: We should probably still throw in this case?

jacoblee93 (commented):
I think this makes sense - CC @nfcampos

langchain-core/src/runnables/graph_mermaid.ts
jasonphillips: Good point; I restored this, and also needed to add one more protection to the recursion to prevent reaching deep paths twice

langchain-core/src/runnables/graph_mermaid.ts
```diff
@@ -97,22 +97,44 @@ export function drawMermaid(
...
         );
       }
 
-      seenSubgraphs.add(subgraph);
+      seenSubgraphs.add(prefix);
       mermaidGraph += `\tsubgraph ${subgraph}\n`;
     }
 
+    // all nested prefixes for this level, sorted by depth
+    const nestedPrefixes = sortPrefixesByDepth(
```
dqbd, line 125: Will need to take a closer look, but shouldn't all prefixes passed into `sortPrefixesByDepth` have the same length?

christian-bromann:
@jasonphillips thanks for your contribution 🙏 our team has been focused on working on LangChain v1 and unfortunately couldn't give PRs as much attention as they need. We are in an effort to picking up where we left off. I will take this PR and discuss it with the team. Please stay tuned!

christian-bromann (approved):
LGTM 👍 

Thanks a lot!

PR #7910, opened 2025-03-27
fix(community): add optional chaining to JiraDocumentConverter to prevent undefined errors

thomHayner:
This pull request adds optional (?) chaining to certain values which may be undefined to handle errors.


jacoblee93 (approved):
Thank you!

PR #7917, opened 2025-03-27
feat(core): add `orchestratorAbortBehavior` option to `RunnableConfig`

benjamincburns:
This feature allows the consumer of a `Runnable` to define how the orchestration layer (i.e. the LangChain core generic `Runnable` implementations, like `RunnableLambda`, and LangGraph's task execution logic) behaves when the abort signal is fired.

This is necessary because JavaScript doesn't provide any mechanism for forcibly terminating an asynchronous task. The best you can do is hand the task itself an `AbortSignal` and allow it to decide when to abort. The only alternative to this would be to simply abandon the promise instead of blocking until it settles.

Prior to this change, (and by default under this change) the execution layer in LangChain will abandon a `Promise` and throw under most circumstances when the `AbortSignal` is triggered. This means that asynchronous work in progress that does not monitor the `AbortSignal` will continue to execute, but the result of that work will not be captured.

The change introduced by this feature allows the consumer of a runnable to specify how eagerly the execution runtime should handle aborts. The options are:

- `"throw_immediately"` (default): The execution runtime will abandon an active asynchronous task and throw when `RunnableConfig.signal` requests an abort.
- `"complete_pending"`: The execution runtime will *not* abandon active asynchronous tasks. Instead, the `AbortSignal` is checked before starting additional asynchronous tasks, and the execution runtime throws on abort only if it would otherwise start new asynchronous work.
- `"passthrough"`: The execution runtime will *not* check the `AbortSignal` at all in this case, but the signal will be passed through to the asynchronous tasks as per usual. This mode delegates the responsibility for early termination entirely to the asynchronous tasks themselves. This mode also makes it possible for execution runtimes external to LangChain core (like LangGraph) can request an abort while still allowing in-progress chains to complete so their state can be captured and replayed on resume.

### DIFF
--- a/langchain-core/src/runnables/base.ts
+++ b/langchain-core/src/runnables/base.ts
@@ -6,10 +6,11 @@ import {
   type TraceableFunction,
   isTraceableFunction,
 } from "langsmith/singletons/traceable";
-import type {
-  RunnableInterface,
-  RunnableBatchOptions,
-  RunnableConfig,
+import {
+  type RunnableInterface,
+  type RunnableBatchOptions,
+  type RunnableConfig,
+  OrchestratorAbortBehavior,
 } from "./types.js";
 import { CallbackManagerForChainRun } from "../callbacks/manager.js";
 import {
@@ -34,7 +35,7 @@ import {
   pipeGeneratorWithSetup,
   AsyncGeneratorWithSetup,
 } from "../utils/stream.js";
-import { raceWithSignal } from "../utils/signal.js";
+import { checkAbortSignal, raceWithSignal } from "../utils/signal.js";
 import {
   DEFAULT_RECURSION_LIMIT,
   ensureConfig,
@@ -354,6 +355,7 @@ export abstract class Runnable<
         runId: options.runId,
         timeout: options.timeout,
         signal: options.signal,
+        orchestratorAbortBehavior: options.orchestratorAbortBehavior,
       });
     }
     const callOptions = { ...(options as Partial<CallOptions>) };
@@ -367,6 +369,7 @@ export abstract class Runnable<
     delete callOptions.runId;
     delete callOptions.timeout;
     delete callOptions.signal;
+    delete callOptions.orchestratorAbortBehavior;
     return [runnableConfig, callOptions];
   }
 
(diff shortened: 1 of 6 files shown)


langchain-core/src/runnables/base.ts
```diff
@@ -395,8 +398,13 @@ export abstract class Runnable<
     delete config.runId;
     let output;
     try {
+      // To support OrchestratorAbortBehavior.COMPLETE_PENDING, we need to check the signal here and
+      // throw if it aborted rather than starting additional tasks
+      if (checkAbortSignal(config)) {
+        config?.signal?.throwIfAborted();
+      }
```
benjamincburns, line 405: Bit of a self nit, but I don't love this pattern for this utility function. The `checkAbortSignal` function itself will throw on abort if `throw_immediately` is set (default). Maybe it should have a required flag that tells it if there are pending active tasks?

langchain-core/src/runnables/base.ts
```diff
@@ -515,6 +523,15 @@ export abstract class Runnable<
...
     try {
+      const resolvedOrchestratorAbortBehavior =
+        options?.orchestratorAbortBehavior ??
+        OrchestratorAbortBehavior.THROW_IMMEDIATELY;
+
+      const signal =
+        resolvedOrchestratorAbortBehavior ===
+        OrchestratorAbortBehavior.THROW_IMMEDIATELY
+          ? options?.signal
+          : undefined;
```
benjamincburns, line 534: This special case was necessary here to avoid a breaking change to the `pipeGeneratorWithSetup` function. I _think_ it's meant to be an internal utility, but nothing marks it as internal and it's exported from a file that's included in the package exports, so I think we should regard it as public even if that wasn't the original intent.

langchain-core/src/runnables/base.ts
```diff
@@ -1912,12 +1934,21 @@ export class RunnableSequence<
             ),
           })
         );
-        nextStepInput = await raceWithSignal(promise, options?.signal);
+        nextStepInput = await raceWithSignal(promise, config);
+        if (!isAsyncIterable(nextStepInput) && checkAbortSignal(config)) {
```
benjamincburns, line 1938: Are there other cases that might indicate an active async task? this should cover async iterables, async generators, and `ReadableStream`.

PR #7934, opened 2025-03-30
feat(community): add include_image_descriptions for Tavily search request

AllenFang:
* add `include_image_descriptions` param for Tavily search request
* only attach the images in metadata when `include_image_descriptions` is set to true. See https://docs.tavily.com/documentation/api-reference/endpoint/search#response-images


libs/langchain-community/src/retrievers/tavily_search_api.ts
```diff
@@ -44,6 +45,8 @@ export class TavilySearchAPIRetriever extends BaseRetriever {
 
   includeImages = false;
 
+  includeImageDescriptions = false;
```
jacoblee93, line 48: nit: do we need to have a default here at all? Can we just have it `undefined`?

jacoblee93:
Going to close this in favor of #7957

PR #7935, opened 2025-03-31
fix(google-genai): Fix Empty Properties Error in Google Generative AI Request, add ids to tool calls

eug:
This PR fixes an issue where an empty properties object in a function declaration for Google Generative AI caused a `400 Bad Request error`. The update ensures that when a tool's schema has an empty properties `object`, the function declaration omits the parameters field entirely.

Issue:
```
GoogleGenerativeAIFetchError: [GoogleGenerativeAI Error]: Error fetching from https://generativelanguage.googleapis.com/v1beta/models/gemini-2.0-flash:generateContent: [400 Bad Request] * GenerateContentRequest.tools[0].function_declarations[0].parameters.properties: should be non-empty for OBJECT type
```

Changes:
- Added a check for jsonSchema.properties being empty.
- If empty, the function declaration will only include name and description.

Fixes # (issue)


jacoblee93:
Agh, thanks for this - will add an integration test

Although I could have sworn I'd tried this and it worked ok

jacoblee93:
Thanks! Will add a test and merge

eug:
Thanks for the help @jacoblee93 with the testing!

I believe it makes sense to have passed on your previous tests because having a property with *empty object* or *undefined property* should lead to the same semantics, but this provider seems to be strict when parsing these properties.

jacoblee93:
Found another issue where simple back and forth with the model would not populate tool call ids and Google function names in some circumstances, pushed a fix for that too here

CC @ccurme

ccurme (approved):

PR #7937, opened 2025-04-01
feat(community): move `api_key` from body to header for Tavily tool

svennergr:
As far as I can see, the authorization in Tavily moved from request body to `Authorization` header: https://docs.tavily.com/documentation/api-reference/endpoint/search#authorization-authorization

This change makes it easier to proxy requests together with `apiUrl`, because the proxy service would only need to rewrite headers and not change request bodies.


svennergr:
cc @jacoblee93 - would be cool if you could take a look. TY

PR #7961, opened 2025-04-03
feat(anthropic): Add image url support to anthropic

meng-tsai:
This PR updates the `langchain-anthropic` package to support image inputs provided as URLs, in addition to the existing base64 encoding support.

According to the latest [Anthropic documentation](https://docs.anthropic.com/en/api/messages-examples#vision), Claude 3 models can accept image inputs via a URL source: `{"type": "image", "source": {"type": "url", "url": "..."}}`.

This change modifies the `_formatImage` utility function in `libs/langchain-anthropic/src/utils/message_inputs.ts` to detect if the input string is a valid URL. If it is, it formats the message payload accordingly; otherwise, it falls back to the base64 parsing logic.

This resolves the limitation where LangchainJS previously only accepted base64-encoded images, allowing users to leverage the full capabilities of the Anthropic API for multimodal inputs more easily.

```
Test Suites: 1 skipped, 56 passed, 56 of 57 total
Tests:       12 skipped, 323 passed, 335 total
Snapshots:   39 passed, 39 total
Time:        5.396 s
Ran all test suites.
```

### DIFF
--- a/libs/langchain-anthropic/src/utils/message_inputs.ts
+++ b/libs/langchain-anthropic/src/utils/message_inputs.ts
@@ -26,20 +26,39 @@ import {
 function _formatImage(imageUrl: string) {
   const regex = /^data:(image\/.+);base64,(.+)$/;
   const match = imageUrl.match(regex);
-  if (match === null) {
+
+  // Check if the imageUrl is a data URI
+  if (match) {
+    return {
+      type: "base64",
+      media_type: match[1] ?? "",
+      data: match[2] ?? "",
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } as any;
+  }
+
+  // Check if the imageUrl is a URL
+  try {
+    // Check if the imageUrl is a valid URL.
+    // If it is, then we assume it is a URL reference.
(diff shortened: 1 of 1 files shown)


libs/langchain-anthropic/src/utils/message_inputs.ts
```diff
@@ -26,20 +26,39 @@ import {
...
-  if (match === null) {
+
+  // Check if the imageUrl is a data URI
+  if (match) {
+    return {
+      type: "base64",
+      media_type: match[1] ?? "",
+      data: match[2] ?? "",
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } as any;
```
jacoblee93, line 37: Do we need this cast?

meng-tsai: I'm just following the cast in the previous implementation.

benjamincburns:
Hi @meng-tsai & @menghuant - thanks for the submission. Unfortunately we need a different approach to fix this issue. Will close this and push a correct fix in just a moment. Sorry that we're unable to accept!

PR #7976, opened 2025-04-07
fix(ci): Remove unused debug action

jacoblee93:


eyurtsev (approved):
looks good, don't think we ever used it

PR #7997, opened 2025-04-09
docs(community): removed @param tag for non-existent function parameter

alexTayanovsky:
Removed outdated `@param chunkIndex` tag from JSDoc comments that references non-existent function parameter.


jacoblee93 (approved):
Thanks!

PR #8005, opened 2025-04-10
fix(google-genai): handle tool_calls when the message contains other content

andantonyan:
This update modifies convertMessageContentToParts to ensure that tool_calls on a message are processed even when the message also contains content (either as a string or an array). Previously, the presence of content could lead to tool_calls being overlooked. This change correctly converts both textual/array content and any accompanying tool_calls into their respective Google AI Part structures within the same message conversion.


jacoblee93:
CC @afirstenberg if you have a minute!

exi:
Whats the status on this? I'm seeing the issue in when using Librechat with a Gemini-based agent. Leading to failed requests because langchain-google-genai is dropping the tool messages.

danny-avila:
@benjamincburns @afirstenberg hoping to see this merged, lgtm!

benjamincburns:
@andantonyan if you're still around & able would you be willing to add one or more tests for this? I'm afraid that the complexity here is a bit beyond what I'm comfortable merging without tests. That said, if @afirstenberg is able to review and is happy w/ the change, then I'm happy to merge & release.

andantonyan:
Hi @benjamincburns ,

Thanks for your comment and the review.

Regarding the tests, what kind of tests would you like to see added? I'll be available from mid next week and will be happy to allocate some time for this.

As you mentioned, let's also wait to hear from @afirstenberg – perhaps if they are happy with the changes, additional tests won't be needed for merging.

Thanks

benjamincburns:
> what kind of tests would you like to see added?

Ideally readable unit tests that target each branch of `convertMessageContentToParts`, but I'll take what I can get. I think the bare minimum would be a set of tests (ideally unit tests) that fail prior to your change, pass after it, and would be sufficient to prevent this issue from regressing.

andantonyan:
Hi @benjamincburns 

I've added the unit tests for `convertMessageContentToParts` as you suggested. They cover the key scenarios for handling AIMessages with both content and tool_calls, plus relevant regressions.

Ready for another look when you get a chance.

benjamincburns (approved):

benjamincburns:
Thanks for the contribution @andantonyan!

PR #8020, opened 2025-04-15
refactor(messages): Add return ToolCall type from messages

Xm0onh:
I needed to declare `ToolCall` for one of the structures in part of my project. I noticed this type is not included in the export message package.

### DIFF
--- a/langchain-core/src/messages/index.ts
+++ b/langchain-core/src/messages/index.ts
@@ -16,4 +16,5 @@ export {
   type InvalidToolCall,
   isToolMessage,
   isToolMessageChunk,
+  type ToolCall,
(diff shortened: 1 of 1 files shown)


langchain-core/src/messages/index.ts
```diff
@@ -16,4 +16,5 @@ export {
   type InvalidToolCall,
   isToolMessage,
   isToolMessageChunk,
+  type ToolCall,
```
jacoblee93, line 19: This conflicts with the `ToolCall` declared in `base.ts` - you should import it like this:

```ts
import { type ToolCall } from "@langchain/core/messages/tool";
```

PR #8021, opened 2025-04-15
fix: LangChain tracer should send session name with update requests

angus-langchain:
### Purpose
Currently, patch requests don't carry information about the project they are associated with. This causes upstream issues on ingestion when the patch is sent separately and we don't know the trace tier of the project, and thus, assign it the default tracing project tier. The result of this is that we upgrade the s3 urls post ingestion API: resulting in higher storage costs and an increase in class A operations.

### Changes
1. Adds `session_id` to update run callback in the LangChainTracer
2. Bumps LangSmith version since this requires a schema change


hinthornw (commented):
Change seems reasonable to me. Probably don't need to update thedep though?

langchain-core/src/tracers/tracer_langchain.ts
```diff
@@ -105,6 +105,7 @@ export class LangChainTracer
       dotted_order: run.dotted_order,
       parent_run_id: run.parent_run_id,
       extra: run.extra,
+      session_name: this.projectName,
```
jacoblee93, line 108: Do we not need to pass this on runCreate as well?

jacoblee93 (commented):
One question and yeah do we need to update the dep?

Not a huge deal just curious

langchain-core/src/tracers/tracer_langchain.ts
angus-langchain: We already pass in session on runCreate: https://github.com/langchain-ai/langchainjs/pull/8021/files#diff-e886d71179cbd7c0f2438e1fe3adc93aab47ac7a1693b8aab6cba894afbef564L82

PR #8022, opened 2025-04-15
fix(google-common): Fix warning about system message in Vertex Anthropic

### LINKED ISSUE #8011: ChatVertexAI logs "Unknown BaseMessage type: system " when using SystemMessage with Claude models
### Checked other resources

### Example Code

```javascript
  modelName: 'claude-3-7-sonnet@20250219',
  temperature: 0.2,
});

const messages = [
  new SystemMessage('Answer only in italian'),
  new HumanMessage('What is the moon?')
];

```

This produces the following console output:

```
Unknown BaseMessage type: system SystemMessage {
"content": "Answer only in italian",
"additional_kwargs": {},
"response_metadata": {}
}
```

### Description

When using SystemMessage with Claude models via the ChatVertexAI integration, the code works but produces log warnings about "Unknown BaseMessage type: system".

Despite the warning, the code works correctly and the Claude model does respect the system message, but the log output is concerning.

## Expected behavior
The system should handle SystemMessage correctly without warnings when using Claude models via VertexAI.

## Actual behavior
The system logs a warning about "Unknown BaseMessage type: system" but still processes the request correctly.

### System Info

"langchain": "0.3.21"
"@langchain/core": "0.3.44",
"@langchain/google-vertexai": "0.2.4",
"@langchain/anthropic": "0.3.17",

afirstenberg:
System Message types aren't supposed to be handled in `baseToAnthropicMessage()`, but they were generating a warning there. Make sure there isn't a warning and add a test to confirm.

Fixes #8011

### DIFF
--- a/libs/langchain-google-common/src/utils/anthropic.ts
+++ b/libs/langchain-google-common/src/utils/anthropic.ts
@@ -618,6 +618,9 @@ export function getAnthropicAPI(config?: AnthropicAPIConfig): GoogleAIAPI {
         return aiMessageToAnthropicMessage(base as AIMessage);
       case "tool":
         return toolMessageToAnthropicMessage(base as ToolMessage);
+      case "system":
+        //  System messages are handled in formatSystem()
+        return undefined;
       default:
         console.warn(`Unknown BaseMessage type: ${type}`, base);
         return undefined;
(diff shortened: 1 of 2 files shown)


jacoblee93 (approved):
Thanks!

PR #8024, opened 2025-04-16
fix(langchain): Remove extra }} from structured output format instructions

lx-0:
Fixes extra }} from structured output format instructions


jacoblee93 (approved):
Thank you!

You should generally not be using this in favor of [`.withStructuredOutput`](https://js.langchain.com/docs/how_to/structured_output/) but nice catch.

PR #7943, opened 2025-04-02
fix(anthropic): Fix Anthropic tool calling for tools with no args

jacoblee93:
Fixes #7872

PR #8025, opened 2025-04-16
fix: don't drop tool input when converting messages to Anthropic payload

sd2k:
As mentioned in https://github.com/langchain-ai/langchainjs/pull/7943#issuecomment-2807991300, Anthropic's tool use inputs are Javascript objects at this point rather than JSON strings, so we only need to parse the input if it's a string.

Without this change, the tool input is silently dropped when converting messages due to the try/catch.


Altaflux:
Thanks @sd2k, this looks like it correctly addresses the issue.

jacoblee93:
Thanks for this! Will dig in

jacoblee93:
So if I'm understanding correctly - we had bad logic here the whole time and we were reliant on the no-op?

jacoblee93:
Yikes - apologies for whiffing on this folks :/

Thank you for adding a test.

PR #8026, opened 2025-04-16
fix(docs): fix parameter casing for useResponsesAPI -> useResponsesApi

24601:
useResponsesAPI was used where it should be useResponsesApi, small but important b/c it's enough to be throwing off AI tooling, but also is where docs are generated from, and threw me (as a soft carbon based lifeform) off as well when reading it with my old school eyes/manual coding process (shame, I know...).


jacoblee93 (approved):
Thank you sir!

Issue #8039, opened 2025-04-18
max_completion_tokens is required for OpenAI o4-mini.


JasonJHKang:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain.js documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain.js rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

const llm  = new ChatOpenAI({
    model: "o4-mini-2025-04-16",
    maxTokens: 8192,
});

const response = await llm.invoke("Hello?");

### Error Message and Stack Trace (if applicable)

BadRequestError: 400 Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead.
    at APIError.generate (/Users/workspace/test/node_modules/openai/error.js:45:20)
    at OpenAI.makeStatusError (/Users/workspace/test/node_modules/openai/core.js:302:33)
    at OpenAI.makeRequest (/Users/workspace/test/node_modules/openai/core.js:346:30)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async /Users/workspace/test/node_modules/@langchain/openai/dist/chat_models.cjs:2036:29
    at async RetryOperation._fn (/Users/workspace/test/node_modules/p-retry/index.js:50:12) {
  status: 400,
  error: {
    message: "Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead.",
    type: 'invalid_request_error',
    param: 'max_tokens',
    code: 'unsupported_parameter'
  },
  code: 'unsupported_parameter',
  param: 'max_tokens',
  type: 'invalid_request_error',
  attemptNumber: 1,
  retriesLeft: 6
}

### Description

max_completion_tokens is required for OpenAI o4-mini model.

const llm  = new ChatOpenAI({
  model: "o4-mini-2025-04-16",
  maxCompletionTokens: 8192,
});
The code above also throws same error.

### System Info

Node.js v22.12.0

dudebot:
this is because o4 is not recognized as a reasoning model:

https://github.com/langchain-ai/langchainjs/blob/275f387de09452a1f711afd7ffc958c0190642a6/libs/langchain-openai/src/chat_models.ts#L727

should be 
function isReasoningModel(model) {
[...]

PR #8040, opened 2025-04-18
o4 reasoning model support (#8039)

dudebot:
Fixes [#8039](https://github.com/langchain-ai/langchainjs/issues/8039)

sorry in advance if it doesn't follow contributing guidelines. I have to rush out the door so just let 4o analyze them on short notice


bracesproul (commented):
thanks a ton for pushing this up!!

PR #8043, opened 2025-04-18
Implement DictPromptTemplate

nfcampos:
Fixes # (issue)

### DIFF
--- a/langchain-core/src/prompts/chat.ts
+++ b/langchain-core/src/prompts/chat.ts
@@ -357,7 +357,14 @@ interface _TextTemplateParam {
 }
 
 function isTextTemplateParam(param: object): param is _TextTemplateParam {
-  return "text" in param;
+  if (param === null || typeof param !== "object" || Array.isArray(param)) {
+    return false;
+  }
+  return (
+    Object.keys(param).length === 1 &&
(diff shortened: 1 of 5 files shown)


langchain-core/src/prompts/chat.ts
```diff
@@ -355,11 +356,36 @@ interface _TextTemplateParam {
   text?: string | Record<string, any>;
 }
 
+function isTextTemplateParam(param: object): param is _TextTemplateParam {
```
bracesproul, line 359: `unknown` instead of `object` maybe?

langchain-core/src/prompts/chat.ts
```diff
@@ -355,11 +356,36 @@ interface _TextTemplateParam {
...
+    typeof param.text === "string"
+  );
+}
+
 interface _ImageTemplateParam {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   image_url?: string | Record<string, any>;
 }
 
+function isImageTemplateParam(param: object): param is _ImageTemplateParam {
```
bracesproul, line 375: same

bracesproul (approved):

PR #8053, opened 2025-04-22
fix(langchain): Expose _ConfigurableModel in exports

bracesproul:


langchain/src/chat_models/universal.ts
```diff
@@ -238,7 +238,10 @@ interface ConfigurableModelFields extends BaseChatModelParams {
   queuedMethodOperations?: Record<string, any>;
 }
 
-class _ConfigurableModel<
+/**
+ * Internal class used to create chat models.
```
benjamincburns, line 242: Not sure if this is set up for langchain yet, but it's good practice to use the `@internal` tsdoc tag on internal structures, that way we don't need to treat them as public for the purpose of breaking changes.

benjamincburns:
```suggestion
 * Internal class used to create chat models.
 *
 * @internal
```

PR #8056, opened 2025-04-23
fix(core): Infer tool return types, with related tool type fixes

### LINKED ISSUE #8010: `tool().invoke()` widens to `any` always
### Checked other resources

### Example Code

[TypeScript Playground Link](https://www.typescriptlang.org/play/?target=11#code/JYWwDg9gTgLgBAbzjCEA2cC+cBmUIhwBEAAmgIYB2A5gMYAW5wlA9LdAKYsroDORAbgBQoSLERwAXllz5CAckkQAJvOFD2lXvBABXNDGBg0ATzgBeZKjQAKIXDg2k5ADRwARlgBcE8j8q6IO4cUAIe-oHBUFgAlBFBIRYAfIj2DnAsAFSZaemZcACy+obGZuRwVMoeAHS5DpksdVAcMLpQlBVw+e7CDpguaQh1lOQgHD5EegZGpkQD6XDKHLy0UEaGEJQTRdOlyADuEHABCVD88+kr9BwgflLVEO4AVhy0ME51DneS1SdRNjELgt3D4fn8QgCgbELpghDF1JptHBmrxihYKvsmDpijMTNVmAA3CAAaw4TjuAEY3CC4BTYsIWCwFgA9AD8QkZxwghloHDg9Ag+zgwHgvHIJl4cAABlQTFKAIRwXgC-TKSjyeAijx87RrGhwAA+cAAKtYCssxdQOKygA)

But in summary:

```ts
import { z } from 'zod';

const multiply = tool(
  ({ a, b }): number => a * b,
  {
    name: "multiply",
    description: "Multiply two numbers",
    schema: z.object({
      a: z.number(),
      b: z.number(),
    }),
  }
);

const result = await multiply.invoke({a: 1, b: 1 });
//    ^? this is `any` instead of something useful
```

_No response_

### Description

My hope was that a manual tool invocation would be somewhat type-safe, but it resolves to `any`.

https://github.com/langchain-ai/langchainjs/blob/626247f65e88fc6a8d1f592d5f38680fc1ac3923/langchain-core/src/tools/index.ts#L610-L640

Unfortunately `ToolReturnType` is typed as `any`, which seems to erase all the hard work the types are doing! I'm not sure why this would be.

### System Info

```
@langchain/community 0.3.34
└── langchain 0.3.21
agentevals 0.0.4
├── langchain 0.3.21
└─┬ openevals 0.0.3
  └── langchain 0.3.21
langchain 0.3.21
```

hntrl:
Fixes #8010. Instead of just using an `any` alias as the return type to invoke, this adds a generic `ToolOutputT` type param on the tool classes that decides the return type of invoke. 

A later enhancement could be to restrict the return type of the provided function if "content_and_artifact" is provided as the response format (i.e. instead of `any[]`, the return type is forced to be `extends [any, any]` in the tool function)

cc @benjamincburns


benjamincburns:
Hi @hntrl - looks like you might've forgotten to do a top-level `yarn build` - looks like the [exports tests are failing due to a type error in the mistral provider](https://github.com/langchain-ai/langchainjs/actions/runs/14625119435/job/41050709347?pr=8056#step:5:2301).

benjamincburns:
> Hi @hntrl - looks like you might've forgotten to do a top-level `yarn build`? The [exports tests appear to be failing](https://github.com/langchain-ai/langchainjs/actions/runs/14625119435/job/41050709347?pr=8056#step:5:2301) due to type errors in the provider libs.

Ah quick correction. I think the exports tests are actually just smoke testing bundled versions of the project, and they tend to catch breakages. I suspect this may be a case where the code in question is making invalid assumptions about the return type of `StructuredTool`, so it's maybe a false positive, but you'd need to evaluate.

Apologies for the confusion on my part - still learning the finer details about our build/test pipeline process.

libs/langchain-aws/src/tests/chat_models.int.test.ts
```diff
@@ -513,12 +513,7 @@ test("Test ChatBedrockConverse can respond to tool invocations with thinking ena
   expect(typeof result.tool_calls![0].args.location).toBe("string");
   expect(result.tool_calls![0].args.location.length).toBeGreaterThan(0);
 
-  messages.push(
-    new ToolMessage({
-      tool_call_id: result.tool_calls![0].id!,
-      content: await tools[0].invoke(result.tool_calls![0]),
-    })
-  );
+  messages.push(await tools[0].invoke(result.tool_calls![0]));
```
hntrl, line 516: StructuredTool.invoke will return a tool message when a tool call is passed, so we don't have to create a new tool message here

hntrl:
> > Hi @hntrl - looks like you might've forgotten to do a top-level `yarn build`? The [exports tests appear to be failing](https://github.com/langchain-ai/langchainjs/actions/runs/14625119435/job/41050709347?pr=8056#step:5:2301) due to type errors in the provider libs.
> 
> Ah quick correction. I think the exports tests are actually just smoke testing bundled versions of the project, and they tend to catch breakages. I suspect this may be a case where the code in question is making invalid assumptions about the return type of `StructuredTool`, so it's maybe a false positive, but you'd need to evaluate.
> 
> Apologies for the confusion on my part - still learning the finer details about our build/test pipeline process.

No worries and thanks for the callout! I think you're right in that the failing test was a false positive. The most recent commits I pushed fix the CI failures.

[a0da856](https://github.com/langchain-ai/langchainjs/pull/8056/commits/a0da856d0eecb8506e631ee87c95e0f37f184454) fixes the failing test from the latest CI run. I added comments explaining the change.

[3cbaceb](https://github.com/langchain-ai/langchainjs/pull/8056/commits/3cbacebc8f44a43c5a57afad25bdd9f864fb64eb) fixes some other build failures I found that come from changing the interface type specifically. From what I can gather, changing that type has a (maybe unnecessary?) rippling effect through a bunch of provider packages, so I'll avoid trying to tackle that for now. At least this way the tool method works as expected when imported from the core package directly.

langchain-core/src/tools/types.ts
```diff
@@ -183,12 +183,12 @@ export interface StructuredToolInterface<
...
    */
-  call<TArg extends StructuredToolCallInput<SchemaT, SchemaInputT>>(
-    arg: TArg,
+  call(
+    arg: StructuredToolCallInput<SchemaT, SchemaInputT>,
     configArg?: Callbacks | RunnableConfig,
     /** @deprecated */
     tags?: string[]
-  ): Promise<TArg extends ToolCall ? ToolMessage : ToolOutputT>;
+  ): Promise<ToolOutputT>;
```
benjamincburns, line 191: Still assessing, but I'm not sure that this change is valid.

It's a bit confusing because we don't have `StructuredTool` actually implementing `StructuredToolInterface` (I forget why that is), but the latter is meant to reflect the public interface of the former. Same goes for `ToolInterface` and `Tool`.

benjamincburns: I think additionally all of those should be assignable to `StructuredToolInterface` (lack of type params is intentional on this and all of the above - default params should be wide enough to allow this).

hntrl: Makes sense. Two paths I can see going from here (both of which I'm happy to do, more just a matter of preference):
(1) Regress the tool classes to return just `ToolOutputT` instead of switching based on input (which doesn't really have any impact since it was just returning an `any` alias before, at least this way it's an optional type param on the classes + interfaces that we can use to control the output type)
(2) Bring the output type switching back to the interface and fix the side effects in the provider packages

Option 1 gets this PR merged quicker, option 2 is more "feature complete".

Unrelated but I'd be interested to know why that inheritance isn't there.

hntrl: I think the diff makes it confusing but this change specifically is reverting a previous change I made. On main the return type is set to `Promise<ToolReturnType>`

benjamincburns: > Unrelated but I'd be interested to know why that inheritance isn't there.

We've been iterating in this area recently (see #7973), so the types are a bit more well-aligned between the classes and interfaces now than when they initially came together. Looking into these issues myself, I see one bug that I introduced in that PR, where I should've aligned the constraint for `SchemaT` on `ToolInterface` with the constraint for `SchemaT` on `Tool`. I _think_ that's all that's necessary to make `Tool` implement `ToolInterface` and `StructuredTool` implement `StructuredToolInterface`.

> Two paths I can see going from here (both of which I'm happy to do, more just a matter of preference):
> 
> (1) Regress the tool classes to return just ToolOutputT instead of switching based on input (which doesn't really have any impact since it was just returning an any alias before, at least this way it's an optional type param on the classes + interfaces that we can use to control the output type)

My concern here would be that we'd drop the `ToolMessage` type from the output, or am I mistaken? That would trigger cases where we return something that's not expressed by the return type. Unfortunately I think it'd be kinder to consumers to regress all the way back to `any` as it'd be better to have no "documentation" of types here than to be misleading.

> (2) Bring the output type switching back to the interface and fix the side effects in the provider packages

I think this would be the preferred outcome, with a bias toward better alignment of the type hierarchy and against using casts to fix the errors outside of `@langchain/core` (if possible). I can probably pick this up Monday if you don't have the time for it, though.

hntrl: > My concern here would be that we'd drop the ToolMessage type from the output, or am I mistaken? That would trigger cases where we return something that's not expressed by the return type.

Good catch. I guess I was thinking more-so on the lines of consuming from the interface end instead of the consumer end (i.e. not creating tools using the `tool` method) where the output type is always implicitly any. Agree this would be a bad change.

I'll have some more time to work on this tomorrow. If you don't hear from me by Monday then feel free to work off this PR

hntrl:
@benjamincburns Some updates on my findings:

It turns out the error I was trying to corner has to do more with how typescript computes type inheritance rather than integration problems. When I add the conditional return types back to the interfaces ([284f3e5](https://github.com/langchain-ai/langchainjs/pull/8056/commits/284f3e53edc319171f66afc836e621a2d6ad348d)), the error that occurs downstream is with the prebuilt agents and how it passes tools around. I haven't been able to address this yet so I imagine CI is failing atm.

The crux of the error is `DynamicStructuredTool<...> is not assignable to StructuredToolInterface`.

If you try to add the interface inheritance directly on the class, the error comes from the `call` method where it's saying that the two arg types are incompatible. `<TArg extends StructuredToolCallInput<...>>(...) is not assignable to <TArg extends unknown>(...)`

It's almost like when evaluating if the dynamic tool conforms to the interface, TypeScript is computing the `SchemaT` param on the interface (which will always fall down to `unknown` bc of how the tool call input type is defined), assuming that's the reference type for the generic in the call method, and complaining that `StructuredToolCallInput != unknown`. If that's the case, this doesn't lend itself well to comparing generic methods like we're doing here to get the conditional return type. 

This is just specifically for `DynamicStructuredTool`. I can add inheritance from `StructuredToolInterface` to `Tool` and `DynamicTool` without any problems.

langchain-core/src/tools/index.ts
```diff
@@ -166,34 +172,52 @@ export abstract class StructuredTool<
...
+    TArg extends StructuredToolCallInput<SchemaT, SchemaInputT>,
+    TConfig extends Callbacks | ToolRunnableConfig | undefined
+  >(
+    arg: TArg,
+    configArg?: TConfig,
     /** @deprecated */
     tags?: string[]
-  ): Promise<ToolReturnType> {
-    let parsed = arg;
+  ): Promise<ToolReturnType<TArg, TConfig, ToolOutputT>> {
```
benjamincburns, line 183: Unfortunately I think this return type is invalid here - `call` can't return `ToolMessage` - only `invoke` can.

benjamincburns: Argh, it's more complex than that, even - if the output content is an object that contains `lc_direct_tool_output: true` we always output the `content` without wrapping in a `ToolMessage` - fun fun.

benjamincburns:
Rebased on current `main` and linearized history, which dropped a couple of duplicated commits.

benjamincburns:
Build passes locally - will merge if CI goes green 🤞.

benjamincburns:
Realized that I didn't have type tests for the tool methods (namely `call` and `invoke`). Added those and found some more issues. Still need to add them for `Tool`, however.

langchain-core/src/tools/index.ts
hntrl: > `call` returns `ToolMessage` only if the `toolCall` property of the `config` is set and contains a nonempty ID.

Doesn't `call` also cover the case of returning a tool message if `arg` is a tool call? The continuation i'm following is after the validation logic (starting at `let toolCallId: string | undefined;`), toolCallId gets assigned if arg is a tool call, then `_formatToolOutput` returns a ToolMessage if toolCallId is defined.

I could very well be missing something here...

langchain-core/src/tools/index.ts
```diff
@@ -283,6 +320,18 @@ export abstract class Tool extends StructuredTool<
...
+    TConfig extends ToolRunnableConfig | undefined
+  >(
+    input: TArg,
+    config?: TConfig
+  ): Promise<ToolReturnType<NonNullable<TArg>, TConfig, ToolOutputT>> {
+    return super.invoke(input, config) as Promise<
+      ToolReturnType<NonNullable<TArg>, TConfig, ToolOutputT>
+    >;
+  }
+
```
hntrl, line 334: This new override is causing CI to fail. On main there is no invoke override for `tool`, it just assumes the signature from extending StructuredTool. Removing this gets my build to work locally

The TArg union is covered by all of the types in StructuredToolCallInput (except for undefined), but I'm lost as to why undefined is acceptable as input. Wouldn't an undefined input fail validation every time?

benjamincburns:
> > call returns ToolMessage only if the toolCall property of the config is set and contains a nonempty ID.

> Doesn't call also cover the case of returning a tool message if arg is a tool call? The continuation i'm following is after the validation logic (starting at let toolCallId: string | undefined;), toolCallId gets assigned if arg is a tool call, then _formatToolOutput returns a ToolMessage if toolCallId is defined.

> I could very well be missing something here...

Nah you're right - I meant to go back and strike out that comment. In the end I wound up doubling back on this and making the typing a bit more explicit as you saw, based on the devX that I was seeing when writing the tests.

langchain-core/src/tools/index.ts
benjamincburns: Unfortunately `undefined` was part of the input signature here originally. If I were writing this from scratch I definitely wouldn't allow `undefined` input, but we're trying to avoid making input types any narrower than they were before this PR, as otherwise this would be a breaking change.

I suppose there's a valid argument that narrowing output types is breaking here in the strictest sense, as the change can introduce new compiler errors where there weren't any previously. However I think it's reasonable to consider a change to be not breaking if the only compiler errors it introduces are ones that expose structural and/or logical errors that existed prior to the change in question.

benjamincburns: I did some spelunking and it's been that way since at least [2 years ago](https://github.com/langchain-ai/langchainjs/blame/da9b1d922170a0b2e237d233b43da6a0f89e7150/langchain-core/src/tools.ts#L209).

IIRC I added the `invoke` override to align the invoke input type with the `call` input type for adherence to the interface. Maybe that wasn't necessary, though. 🤔

langchain-core/src/tools/tests/types.test.ts
```diff
@@ -18,14 +19,14 @@ const testDynamicTool = new DynamicTool({
...
-  schema: z.string(),
+  func: async (input: { input: string }) => `test ${input.input}`,
+  schema: z.object({ input: z.string() }),
 });
 
 const testDynamicStructuredToolWithZodEffects = new DynamicStructuredTool({
   name: "test",
   description: "test",
-  func: async (input: string) => `test ${input}`,
+  func: async (input: { input: string }) => `test ${input.input}`,
```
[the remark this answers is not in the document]
benjamincburns: Yeah - I think I must've modified this in error. Will double check by adding a test that validates the output on invoke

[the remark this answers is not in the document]
benjamincburns: Yeah you were right - good catch!

benjamincburns:
> Still need to add them for `Tool`, however.

This was silly - `Tool` is abstract, and `DynamicTool` extends it - I think this is complete now, provided CI is green.

benjamincburns:
Thanks a ton for all your help on this one @hntrl! It was a pleasure working with you to get this over the finish line!

PR #8081, opened 2025-04-27
feat(tavily): use chunks per source arg

anadi45:
Adds missing `chunks_per_source` parameter in the latest tavily search class


benjamincburns:
Thanks for the contribution. I had to add one small additional change so the new config field would be serialized correctly into the request, otherwise LGTM!

PR #8089, opened 2025-04-28
Upgrading weaviate client to v3.5.2

dee-at-weaviate:
Fixes # (issue)


benjamincburns:
Just pushed some fixes. Please avoid using sleeps in tests - if they are necessary to make the test pass they are almost always an indication of a concurrency bug. In this case there were `await` keywords missing in both the library code and the test code.

libs/langchain-weaviate/src/tests/translator.int.test.ts
```diff
@@ -220,56 +231,50 @@ test.skip("Weaviate Vector Store Self Query Retriever Test With Default Filter O
...
 
-  const query4 = await selfQueryRetriever.getRelevantDocuments(
-    "Wau wau wau wau hello gello hello?"
-  );
-  // console.log(query4); // query4 has to return documents, since the default filter takes over with
-  expect(query4.length).toEqual(7);
+    const query4 = await selfQueryRetriever.invoke(
+      "Wau wau wau wau hello gello hello?"
+    );
+    // console.log(query4); // query4 has to return documents, since the default filter takes over with
```
benjamincburns, line 270:
```suggestion
    // query4 has to return documents, since the default filter takes over
```

libs/langchain-weaviate/src/tests/translator.int.test.ts
```diff
@@ -367,51 +372,62 @@ test.skip("Weaviate Vector Store Self Query Retriever Test With Default Filter A
...
 
-  const query4 = await selfQueryRetriever.getRelevantDocuments(
-    "Wau wau wau wau hello gello hello?"
-  );
-  // console.log(query4); // query4 has to return empty array, since the default filter takes over with and filter
-  expect(query4.length).toEqual(0);
+    const query4 = await selfQueryRetriever.invoke(
+      "Wau wau wau wau hello gello hello?"
+    );
+    // console.log(query4); // query4 has to return empty array, since the default filter takes over with and filter
```
benjamincburns, line 424:
```suggestion
    // query4 has to return empty array, since the default filter takes over with and filter
```

examples/src/indexes/vector_stores/weaviate_fromTexts.ts
```diff
@@ -1,10 +1,10 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
-import weaviate, { ApiKey } from "weaviate-ts-client";
+import weaviate, { ApiKey } from "weaviate-client";
 import { WeaviateStore } from "@langchain/weaviate";
 import { OpenAIEmbeddings } from "@langchain/openai";
 
 export async function run() {
-  // Something wrong with the weaviate-ts-client types, so we need to disable
+  // Something wrong with the weaviate-client types, so we need to disable
```
benjamincburns, line 7: Probably not something that needs to change as part of this PR, but why does `@langchain/weaviate` require the SDK's `client` be passed in at all? Why isn't it initialized internally based on constructor args passed to `WeaviateStore` and similar?

dee-at-weaviate: Thanks Benjamin! That's a fair point and I can look at fixing it in another PR. I don't know the reason why this design was used in the older version and I tried to stay as close as possible to it. But there is actually no good reason for it.

libs/langchain-weaviate/src/tests/translator.int.test.ts
```diff
@@ -75,51 +91,46 @@ test.skip("Weaviate Self Query Retriever Test", async () => {
...
-  );
-  // console.log(query2, query3, query4); // query4 has to return empty array
-  expect(query4.length).toBe(0);
+    const query2 = await selfQueryRetriever.invoke(
+      "Which movies are rated higher than 8.5?"
+    );
+    // this query isn't correctly converted by Langchain selfQuery
+    // const _query3 = await selfQueryRetriever.invoke(
+    //   "Which movies are directed by Greta Gerwig?"
+    // );
```
benjamincburns, line 124: Is there an open issue that captures whatever is broken here?

libs/langchain-weaviate/tsconfig.json
```diff
@@ -3,11 +3,13 @@
...
     "outDir": "../dist",
     "rootDir": "./src",
-    "target": "ES2021",
-    "lib": ["ES2021", "ES2022.Object", "DOM"],
-    "module": "ES2020",
-    "moduleResolution": "nodenext",
+    "target": "ESNext",
+    // "lib": ["ES2021", "ES2022.Object", "DOM"],
+    "module": "NodeNext",
+    "moduleResolution": "NodeNext",
```
benjamincburns, line 9: Unfortunately with our current build process it's important to keep these as they were. If you were seeing errors here it'd likely be due to VSCode using its internal TS version rather than the workspace version. You can hit `ctrl-shift-p` (or `cmd-shift-p` on mac) and type something like "TypeScript version" into the command palette and it'll let you select the workspace version instead.

benjamincburns: With these changes I expect the CJS build will be broken.

tsconfig.json
```diff
@@ -7,8 +7,8 @@
       "ES2022.Object",
       "DOM"
     ],
-    "module": "ES2020",
-    "moduleResolution": "nodenext",
+    "module": "NodeNext",
+    "moduleResolution": "NodeNext",
```
benjamincburns, line 11: Same here.

benjamincburns (changes requested):
Needs a couple changes - for future reference please avoid modifying tsconfigs. The current setup allows for hybrid ESM/CJS builds and it's sensitive to module/target changes.

benjamincburns (approved):

libs/langchain-weaviate/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@langchain/weaviate",
-  "version": "0.1.1",
+  "version": "0.2.0",
```
benjamincburns, line 3: In the future please don't change version numbers in PRs. We communicate the type of change via [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/) messages, and then use these to decide how to modify version numbers when we release.

libs/langchain-weaviate/src/tests/translator.int.test.ts
dee-at-weaviate: It wouldn't convert to a filter, will open an issue for it, thanks Ben

PR #8100, opened 2025-04-30
fix(prisma): filter on enum value

### LINKED ISSUE #7953: PrismaVectorStore.similaritySearch fails when filtering on enum column: "operator does not exist: Type ~~ text"`
### Checked other resources

### Example Code

```typescript
import { PrismaVectorStore } from "@langchain/community/vectorstores/prisma";
import { PrismaClient } from "@prisma/client";

const db = new PrismaClient();

// This works fine:
const modelVectorStore = PrismaVectorStore.withModel<Model>(db).create(
  embeddings,
  {
    prisma: PrismaClient,
    tableName: "Model",
    vectorColumnName: "vector",
    columns: {
      vectorContent: PrismaVectorStore.ContentColumn,
      type: true, // type is an enum column in the DB
    },
  },
);

// This throws an error:
const results = await modelVectorStore.similaritySearch(
  `${sourceType}: ${metric}`,
  5,
  {
    type: {
      equals: type, // `type` is an enum value
    },
  },
);

```

Raw query failed. Code: `42883`. Message: `ERROR: operator does not exist: "Type" ~~ text
HINT: No operator matches the given name and argument types. You might need to add explicit type casts.`

### Description

I'm trying to use the PrismaVectorStore from @langchain/community to perform a similarity search with a filter on an enum column (type).

I expect the query to filter correctly based on the type enum.

Instead, I get a PostgreSQL error: operator does not exist: "Type" ~~ text.

It seems that the generated SQL is trying to apply a LIKE (~~) operator between a text and an enum, which is not supported without an explicit cast. This likely occurs because the Prisma filter is translated directly into a raw query without accounting for the enum type in the database.

### System Info

"@langchain/community": "^0.3.28",
"@langchain/core": "0.3.19",
"@langchain/langgraph": "0.2.23",
"@langchain/langgraph-checkpoint": "^0.0.13",
"@langchain/mistralai": "^0.2.0",
"@langchain/openai": "0.3.14",
"@langchain/textsplitters": "^0.1.0",

alenoir:
This PR fixes an issue where filtering on enum values using Prisma would fail due to a type mismatch. The fix ensures enum values are correctly cast to text in the default SQL query, allowing filters to work as expected for enums.

Context: When using where filters with enums in a vector store powered by Prisma, the generated SQL failed due to a type mismatch (enum vs. text). This cast to ::text resolves the issue.

Fixes #7953

### DIFF
--- a/libs/langchain-community/src/vectorstores/prisma.ts
+++ b/libs/langchain-community/src/vectorstores/prisma.ts
@@ -455,7 +455,7 @@ export class PrismaVectorStore<
             case OpMap.isNotNull:
               return this.Prisma.sql`${colRaw} ${opRaw}`;
             default:
-              return this.Prisma.sql`${colRaw} ${opRaw} ${value}`;
+              return this.Prisma.sql`${colRaw}::text ${opRaw} ${value}`;
           }
         })
       ),


benjamincburns:
@alenoir - thanks for the contribution - by any chance would you be able to add a test that demonstrates the fix?

PR #8112, opened 2025-05-02
feat(utils/testing): add tool-call streaming support to FakeStreamingChatModel

### LINKED ISSUE #8083: Feature Request: Add tool call support to `FakeStreamingChatModel`
I'm trying to write some tests to make sure our client is reconciling:

1. LLM token streams (e.g. `mode=messages`)
2. State updates (e.g. `mode=updates`)

So that we can both stream responses nicely to the User and also reliably restore the conversation from persistence and "show the same thing".

I've been playing around with the [`FakeToolCallingChatModel`](https://github.com/langchain-ai/langgraphjs/blob/main/libs/langgraph/src/tests/utils.ts#L147) but it behaves strangely w/ `AIMessageChunk`s, emitting them `on_chat_model_end` which according to testing the actual implementation and the [event reference](https://python.langchain.com/docs/how_to/streaming/#event-reference) is not quite right.

I guess this is because its a Fake and not designed for this use case!

There seems to be a [`FakeStreamingChatModel`](https://github.com/langchain-ai/langchainjs/blob/main/langchain-core/src/utils/testing/index.ts#L215) but that doesn't work with tool calling. Smushing the two implementations together (`FakeToolCallingChatModel` and `FakeStreamingLLM`) has not worked so far. But could that work if you know what you're doing?

Do you have any advice on trying to test this scenario?

RUPOJS:
Added tool-call streaming requested in #8083.

Now it accepts chunks (array of AIMessageChunk) and toolStyle.
Adds .bindTools() to attach provider-style tool specs.
Streams given chunks verbatim; falls back to original char-by-char logic.

Fixes #8083


langchain-core/src/utils/testing/index.ts
```diff
@@ -212,30 +214,89 @@ export class FakeChatModel extends BaseChatModel {
...
 
-export class FakeStreamingChatModel extends BaseChatModel {
+export class FakeStreamingChatModel extends BaseChatModel<FakeStreamingChatModelCallOptions> {
   sleep?: number = 50;
 
   responses?: BaseMessage[];
 
+  chunks?: AIMessageChunk[] = [];
+
+  toolStyle?: "openai" | "anthropic" | "bedrock" | "google";
```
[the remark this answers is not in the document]
benjamincburns: If not, we should either make this non-nullable, or specify a default option.

benjamincburns (changes requested):
Had a few questions

langchain-core/src/utils/testing/index.ts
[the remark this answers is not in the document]
benjamincburns: Looks like the lack of default is causing CI to fail. Be sure to run `yarn format && yarn lint && yarn build && yarn test` before commit/push.

langchain-core/src/runnables/tests/runnable.test.ts
```diff
@@ -585,4 +592,57 @@ describe("runId config", () => {
...
+        tool_calls: [
+          {
+            id: "call_1",
+            type: "tool_call",
+            name: "echo",
+            args: { input: "hello" },
+          },
+        ],
+      }),
+      new AIMessageChunk({ content: "!" }),
```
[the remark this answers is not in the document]
RUPOJS: Good idea—updated the test to build the chunks programmatically:

```
const text = "Hi!";
const chunks = [...text].map(c => new AIMessageChunk({ content: c }));
```
Much cleaner than enumerating each char.
Let me know if you’d like me to add or modify anything else!

langchain-core/src/utils/testing/index.ts
```diff
@@ -212,30 +214,96 @@ export class FakeChatModel extends BaseChatModel {
...
+
+    const toolDicts = merged.map((t) => {
+      switch (this.toolStyle) {
+        case "openai":
+          return {
+            type: "function",
+            function: {
+              name: t.name,
+              description: t.description,
+              parameters: zodToJsonSchema(t.schema),
```
benjamincburns, line 261: tool schema are not always zod - tools can be defined as JSON schema

langchain-core/src/utils/testing/index.ts
```diff
@@ -212,30 +214,96 @@ export class FakeChatModel extends BaseChatModel {
...
-      sleep?: number;
-      responses?: BaseMessage[];
-      thrownErrorString?: string;
-    } & BaseLLMParams
-  ) {
-    super(fields);
-    this.sleep = fields.sleep ?? this.sleep;
-    this.responses = fields.responses;
-    this.thrownErrorString = fields.thrownErrorString;
+  private tools: StructuredTool[] = [];
```
benjamincburns, line 228: should depend on the interface here

benjamincburns (changes requested):
couple more, sorry

langchain-core/src/utils/testing/index.ts
RUPOJS: Hi @benjamincburns , thanks for pointing out this. I'm now convertin only Zod schemas with `toJsonSchema` helper (existing) and plain JSON schemas are forwarded untouched.

langchain-core/src/utils/testing/index.ts
RUPOJS: Done. tools is now typed as (StructuredTool | ToolSpec)[], so we accept both full StructuredTool instances and lightweight { name, description, schema } objects.

@benjamincburns 

Let me know if you’d like any further tweaks/changes!

benjamincburns (approved):

benjamincburns:
Thanks for the contribution, @RUPOJS! This will go out with the next release of `@langchain/core`.

PR #8118, opened 2025-05-05
fix(tavily): Fix unit tests 

abogacki:
When trying to create a release for tavily package, i've noticed that recent PR to the @langchain/tavily package created issues with tests: https://github.com/langchain-ai/langchainjs/pull/8081
- the default parameters were not updated with `chunksPerSource`
- adjustment in the camelCase conversion method created regression


benjamincburns:
Thanks for catching this @abogacki!

PR #8123, opened 2025-05-06
fix(google-common): don't set includeThoughts

### LINKED ISSUE #8119: Cannot set the include thoughts parameter for Google Gemini
### Checked other resources

- [ ] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

In gemini.ts, `includeThoughts` parameter for when the reasoning is enabled is hardcoded. We should expose that parameter in `GoogleAIModelRequestParams`.

Gemini.ts -> formatGenerationConfig function
```typescript
  // Add thinking configuration if explicitly set
  if (typeof parameters.maxReasoningTokens !== "undefined") {
    ret.thinkingConfig = {
      thinkingBudget: parameters.maxReasoningTokens,
      includeThoughts: true,
    };
  }
```
Also, if maxReasoningTokens is explicitly set to 0, this code still sets `includeThoughts` to be true, which causes the following error: `Unable to submit request because Thinking_config.include_thoughts is only enabled when thinking is enabled..`

_No response_

### Description

I'd like to be able to set `includeThoughts` parameter for gemini reasoning models. Right now, it's hardcoded to be 'true'.

### System Info

langchain@0.3.24 | MIT | deps: 12 | versions: 319
MacOS
Node version: v23.7.0

iltenahmet:
Fixes #8119.

- Enables includeThoughts parameter to be set by the user.
- Adds thinking configuration if maxReasoningTokens is not undefined AND not zero. (previously we weren't checking for zero)
- Adds tests

### DIFF
--- a/libs/langchain-google-common/src/utils/gemini.ts
+++ b/libs/langchain-google-common/src/utils/gemini.ts
@@ -1373,7 +1373,8 @@ export function getGeminiAPI(config?: GeminiAPIConfig): GoogleAIAPI {
     ) {
       ret.thinkingConfig = {
         thinkingBudget: parameters.maxReasoningTokens,
-        includeThoughts: parameters.includeThoughts ?? true,
+        // TODO: Expose this configuration to the user once google fully supports it
+        includeThoughts: false,
       };
     }
 
@@ -1586,16 +1587,6 @@ export function validateGeminiParams(params: GoogleAIModelParams): void {
         );
       }
     }
-  }
-
-  if (
-    typeof params.includeThoughts !== "undefined" &&
-    (typeof params.maxReasoningTokens === "undefined" ||
-      params.maxReasoningTokens === 0)
-  ) {
-    throw new Error(
-      "`includeThoughts` can only be set when thinking is enabled"
-    );
   }
 
   if (
--- a/libs/langchain-google-common/src/utils/gemini.ts
+++ b/libs/langchain-google-common/src/utils/gemini.ts
@@ -1367,10 +1367,13 @@ export function getGeminiAPI(config?: GeminiAPIConfig): GoogleAIAPI {
     }
 
     // Add thinking configuration if explicitly set
-    if (typeof parameters.maxReasoningTokens !== "undefined") {
+    if (
+      typeof parameters.maxReasoningTokens !== "undefined" &&
+      parameters.maxReasoningTokens !== 0
+    ) {
       ret.thinkingConfig = {
         thinkingBudget: parameters.maxReasoningTokens,
-        includeThoughts: true,
+        includeThoughts: parameters.includeThoughts ?? true,
       };
     }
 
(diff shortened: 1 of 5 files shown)


afirstenberg:
I'm staring at the code, and I'm not sure why I included `includeThoughts` at all, since it is currently ignored by Gemini. (And I'm irritated at myself that I left it in.)

My suggestion would be:
* Do not offer this as another parameter
* Remove the line `includeThoughts:true` completely.

iltenahmet:
@afirstenberg Thanks for the reply, I think they recently stopped ignoring it for vertex api, because suddenly we started getting reasoning steps in the response.

afirstenberg:
> @afirstenberg Thanks for the reply, I think they recently stopped ignoring it for vertex api, because suddenly we started getting reasoning steps in the response.

Yeah, I'm just finding out details as well. There are also some other details that are changing that I'll try to put in an issue to track soon.

I might suggest we make the fix that sets `includeThoughts` to **false** by default as a rapid fix while we work out the right strategy to get the thinking tokens represented. 
Does that sound reasonable to you, @iltenahmet? (cc: @benjamincburns)

iltenahmet:
> Yeah, I'm just finding out details as well. There are also some other details that are changing that I'll try to put in an issue to track soon.

Sounds good to me will update this pr.

benjamincburns (approved):

benjamincburns:
Thanks for the quick turnaround on this @iltenahmet! I'll get this released ASAP (likely within a few hours, as I have some meetings coming up)

libs/langchain-google-common/src/utils/gemini.ts
```diff
@@ -1367,10 +1367,14 @@ export function getGeminiAPI(config?: GeminiAPIConfig): GoogleAIAPI {
     }
 
     // Add thinking configuration if explicitly set
-    if (typeof parameters.maxReasoningTokens !== "undefined") {
+    if (
+      typeof parameters.maxReasoningTokens !== "undefined" &&
+      parameters.maxReasoningTokens !== 0
+    ) {
       ret.thinkingConfig = {
         thinkingBudget: parameters.maxReasoningTokens,
```
LoganDickey, line 1375: Forgive me if this is the wrong place to comment (or outright incorrect) - but per https://ai.google.dev/gemini-api/docs/thinking I believe that thinking is set to on by default, and must be set to 0 explicitly to disable.

Is there a way to disable thinking with these models that I'm missing?

> "Thinking is on by default in both the API and AI Studio because the 2.5 series models have the ability to automatically decide when and how much to think based on the prompt"

afirstenberg: ... bah.
We introduced a bug and we and the tests didn't catch it.
See #8145 

You **should** be able to set `maxReasoningTokens` to 0 to turn it off.

PR #8128, opened 2025-05-07
feat(google-common), feat(core): Improve token counting

### LINKED ISSUE #8127: Google Gemini 2.5 token counting
### Privileged issue

### Issue Content

In both Gemini 2.5 Flash and Pro, there are three types of token counts:
* `promptTokenCount` (ie - input tokens)
* `thoughtsTokenCount` (ie - reasoning tokens, which count against output tokens)
* `candidatesTokenCount` (ie - the visible output tokens)

Currently, LangChainJS uses just the `promptTokenCount` and `candidatesTokenCount`. It ignores the `thoughtsTokenCount`.

Updates likely needed in the google-common library in:
* gemini.ts: `combineGenerations()`
* gemini.ts: `responseToGenerationInfo()`
* chat_models.ts: `*_streamResponseChunks()`
I don't see any tests that specifically check for it, but that may not be a bad idea.

The question is - what should be the correct value set for `output_tokens`? (ie - how is this handled for other models?)
* If the total of `thoughtsTokenCount` and `candidatesTokenCount`, is there a way they should be broken out as well?
* If this is just `candidatesTokenCount`, then no change is necessary, but it would make sense to include `thoughtsTokenCount` somewhere in a standard way.

(@benjamincburns - please assign to me, tho I need input on the last question I asked, or feel free to make the change yourself if you have the time.)

afirstenberg:
***Note that this includes changes in langchain-core***

Done:
* Changes to `InputTokenDetails` and `OutputTokenDetails` that split out the common modalities field into a new type. Added several modalities based on those available from Gemini.
* Changes in Gemini output processing to handle "thinking" token count, which is not included in the output token count.
* Add the modalities/details from Gemini to the token details
* Tests
  * Note that some tests fail because of a bug in the AI Studio API. I'm working with Google to resolve.

Fixes #8127

### DIFF
--- a/libs/langchain-google-common/src/utils/gemini.ts
+++ b/libs/langchain-google-common/src/utils/gemini.ts
@@ -20,7 +20,8 @@ import {
   isDataContentBlock,
   convertToProviderContentBlock,
   InputTokenDetails,
-  OutputTokenDetails, ModalitiesTokenDetails,
+  OutputTokenDetails,
+  ModalitiesTokenDetails,
 } from "@langchain/core/messages";
 import {
   ChatGeneration,
@@ -864,7 +865,8 @@ export function getGeminiAPI(config?: GeminiAPIConfig): GoogleAIAPI {
   ): void {
     modalityTokenCounts?.forEach((modalityTokenCount) => {
       const { modality, tokenCount } = modalityTokenCount;
-      const modalityLc: keyof ModalitiesTokenDetails = modality.toLowerCase() as keyof ModalitiesTokenDetails;
+      const modalityLc: keyof ModalitiesTokenDetails =
+        modality.toLowerCase() as keyof ModalitiesTokenDetails;
       const currentCount = details[modalityLc] ?? 0;
       // eslint-disable-next-line no-param-reassign
       details[modalityLc] = currentCount + tokenCount;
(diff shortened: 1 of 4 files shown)


libs/langchain-google-common/src/utils/gemini.ts
```diff
@@ -855,6 +858,57 @@ export function getGeminiAPI(config?: GeminiAPIConfig): GoogleAIAPI {
...
+
+      const input_token_details: InputTokenDetails = {};
+      addModalityCounts(usageMetadata.promptTokensDetails, input_token_details);
+
+      const output_token_details: OutputTokenDetails = {};
+      addModalityCounts(
+        usageMetadata?.candidatesTokensDetails,
+        output_token_details
+      );
+      if (usageMetadata?.thoughtsTokenCount) {
```
benjamincburns, line 896:
```suggestion
      if (usageMetadata?.thoughtsTokenCount != null) {
```

else we'll miss zeros

benjamincburns (approved):
LGTM - can live with the integration test failure for now since it doesn't impact CI, but I suppose if it's a lasting issue we should probably either relax the constraint or make it conditional

benjamincburns:
(and thanks!)

PR #8134, opened 2025-05-07
fix(community): mark Azure blob storage connection string as secrets

sinedied:
Make sure connection string is hidden when serialization occurs


benjamincburns (approved):

benjamincburns:
Thanks for the contribution!

PR #8141, opened 2025-05-08
fix(core, anthropic): handling for undefined tool outputs

### LINKED ISSUE #8140: "TypeError: Cannot read properties of undefined (reading 'map')" when tool_result message has undefined content
### Checked other resources

### Example Code

The following code:

```typescript
const result = await boundLLM.invoke(
    [{ role: 'system', content: systemPrompt }, ...state.messages],
    {
      configurable: {
        model: 'claude-3-7-sonnet-20250219',
        modelProvider: 'anthropic',
      }
    }
  );
```

```
2025-05-07T18:39:49.396Z	08228b2e-80c8-481e-b653-e1219deea749	ERROR	TypeError: Cannot read properties of undefined (reading 'map')
    at _formatContent (/var/task/app.js:161819:35)
    at _ensureMessageContents (/var/task/app.js:161790:24)
    at _convertMessagesToAnthropicPayload (/var/task/app.js:161894:26)
    at ChatAnthropic._streamResponseChunks (/var/task/app.js:162682:35)
    at _streamResponseChunks.next (<anonymous>)
    at ChatAnthropic._generateUncached (/var/task/app.js:99905:30)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async ChatAnthropic.invoke (/var/task/app.js:99804:24)
    at async RunnableCallable.callModel [as func] (/var/task/app.js:310196:18)
    at async RunnableCallable.invoke (/var/task/app.js:299799:21) {
  pregelTaskId: '078324e2-2315-5b92-967a-adaab7cfd4e9'
}
```

### Description

- I'm building a ReAct agent with tool node.
- It works well in a ReAct agent with a toolNode until the last messages has no content. e.g.,

```javascript
{
  lc_serializable: true,
  lc_kwargs: {
    tool_call_id: "toolu_01GGEm8vnmAkUG8nnb9yJ11k",
    name: "webSearch",
    additional_kwargs: {},
    response_metadata: {},
    id: "run-aa36fe19-8d1a-4ec8-800f-e7d2fc693d8f-tool-toolu_01GGEm8vnmAkUG8nnb9yJ11k"
  },
  lc_namespace: [
    "langchain_core",
    "messages"
  ],
  name: "webSearch",
  additional_kwargs: {},
  response_metadata: {},
  id: "run-aa36fe19-8d1a-4ec8-800f-e7d2fc693d8f-tool-toolu_01GGEm8vnmAkUG8nnb9yJ11k",
  lc_direct_tool_output: true,
  tool_call_id: "toolu_01GGEm8vnmAkUG8nnb9yJ11k"
}
```

Notice that this happens occasionally even the `webSearch` tool provides response like "Nothing found". the error stacktrace indicates that it's a problem in `@langchain/anthropic` package.

### System Info

`@langchain/anthropic@0.3.20`
npm 10.9.2
linux x86_64.

shawnzhu:
Fixes #8140

### DIFF
--- a/libs/langchain-anthropic/src/utils/message_inputs.ts
+++ b/libs/langchain-anthropic/src/utils/message_inputs.ts
@@ -116,7 +116,10 @@ function _ensureMessageContents(
             content: [
               {
                 type: "tool_result",
-                content: _formatContent(message.content),
+                // rare case: message.content could be undefined
+                ...(message.content != null
+                  ? { content: _formatContent(message.content) }
+                  : {}),
                 tool_use_id: (message as ToolMessage).tool_call_id,
               },
             ],
--- a/langchain-core/src/tools/index.ts
+++ b/langchain-core/src/tools/index.ts
@@ -723,7 +723,7 @@ function _formatToolOutput<TOutput extends ToolOutputType>(params: {
 
 function _stringify(content: unknown): string {
   try {
-    return JSON.stringify(content, null, 2);
+    return JSON.stringify(content, null, 2) ?? "";
   } catch (_noOp) {
     return `${content}`;
   }
--- a/langchain-core/src/tools/tests/tools.test.ts
+++ b/langchain-core/src/tools/tests/tools.test.ts
@@ -56,6 +56,34 @@ test("Tool works if responseFormat is content_and_artifact and returns a tuple",
   expect(toolResult).toBe("msg_content");
 });
 
+test("ToolMessage content coerces to empty string when tool returns undefined", async () => {
+  const weatherSchema = z.object({
+    location: z.string(),
+  });
+
+  const toolCall = {
+    id: "testid",
+    args: { location: "San Francisco" },
+    name: "weather",
+    type: "tool_call",
+  } as const;
+
+  const weatherTool = tool(
+    () => {
+      return undefined;
+    },
+    {
+      name: "weather",
+      schema: weatherSchema,
+    }
+  );
+
+  const toolResult = await weatherTool.invoke(toolCall);
+
+  expect(toolResult).toBeInstanceOf(ToolMessage);
+  expect(toolResult).toHaveProperty("content", "");
+});
+
 test("Does not return tool message if responseFormat is content_and_artifact and returns a tuple and a tool call with no id is passed in", async () => {
   const weatherSchema = z.object({
     location: z.string(),


libs/langchain-anthropic/src/tests/chat_models.test.ts
```diff
@@ -189,3 +189,58 @@ test("Can properly format anthropic messages when given two tool results", async
...
+    new ToolMessage({
+      name: "get_weather",
+      tool_call_id: "weather_call",
+      content: "It is currently 24 degrees with hail in San Francisco.",
+    }),
+    new ToolMessage({
+      name: "calculator",
+      tool_call_id: "calculator_call",
+      // @ts-expect-error this could happen in runtime for tool_result message
+      content: undefined,
```
benjamincburns, line 224: We use types to constrain what values we expect the structure to contain. The fact that you needed `@ts-expect-error` here tells me that we've either over-constrained the type of `ToolMessage.content`, or that this change is handling an invalid case.

I'm not necessarily opposed to accepting the change, but if it's possible in the current version of `@langchain/core` to define a tool that returns `undefined` without unsafe casts and without type errors, we also need to update the type of `ToolMessage.content` to make it optional.

To that end, can you please update this test to use a tool definition that demonstrates a valid cases where a LangChain style tool (aka `StructuredTool` and its descendants) returns `undefined`?

benjamincburns: Also just generally we only use `@ts-expect-error` in negative cases for type tests - using it anywhere else almost always means that the test predicate is invalid.

benjamincburns (changes requested):

libs/langchain-anthropic/src/tests/chat_models.test.ts
shawnzhu: see https://github.com/langchain-ai/langchainjs/pull/8141/commits/5c7dc47837d8baebfcc901eda471fea0739d9bb9

if the RunnableFunc in a tool returns `undefined`, the ToolMessage.content will be `undefined`

benjamincburns: Thanks for adding that, @shawnzhu!

Upon checking, some LLM providers (e.g. Anthropic) do indeed allow tool responses with no content present (the equivalent of a `ToolMessage` with `undefined` content).

> The fact that you needed @ts-expect-error here tells me that we've either over-constrained the type of ToolMessage.content, or that this change is handling an invalid case.

Seems it's the first one, then.

The most correct fix would be to revise `BaseMessage` and related (e.g. `BaseMessageChunk`) to have optional `content`, and then revise `HumanMessage`, `SystemMessage`, and `AIMessage` (along with their chunked counterparts) to narrow their `content` fields such that they're required (as they are today).

The unfortunate side effect of this is that it may cause some follow-on effects in code that doesn't currently assume that `content` may be `undefined`. 😮‍💨 

To avoid those potential breakages it may be more pragmatic to revise [the `_formatToolOutput` function](https://github.com/langchain-ai/langchainjs/blob/3767147e96002018507229b437845af1cead29ef/langchain-core/src/tools/index.ts#L692-L722) to replace `ToolMessage.content` with `""` when the `content` argument is `undefined`. The unfortunate downside of this option would be that `undefined` and `""` are semantically different outputs, and I suspect at least some LLM providers who allow optional tool response content do differentiate how they present those two values back to the LLM.

libs/langchain-anthropic/src/utils/message_inputs.ts
```diff
@@ -116,7 +116,8 @@ function _ensureMessageContents(
             content: [
               {
                 type: "tool_result",
-                content: _formatContent(message.content),
+                // rare case: message.content could be undefined
+                ...(message.content != null ? { content: _formatContent(message.content) } : {}),
```
benjamincburns, line 120: Slightly modified from the original submission. Anthropic's API reference shows `content` here as being an _optional_ (but not nullable) field of type `string | object[]`. The changes in `core` will make this case no longer possible, but this should handle it for older versions that weren't coalescing `undefined` tool output to `""`.

libs/langchain-anthropic/src/tests/chat_models.test.ts
benjamincburns: Wound up taking the latter approach.

benjamincburns (approved):

nfcampos (approved):

benjamincburns:
Thanks for raising this, @shawnzhu

shawnzhu:
@benjamincburns Thanks for the thorough thinking!

PR #8142, opened 2025-05-08
fix(redis): handle both numDocs and num_docs in RedisVectorStore

UrielCh:
The redis store do not use the correct name for the field numDocs

Some other redis conector may convert snake case to camelcase, so I keep the two syntax.

the official redis client use the snake case.

https://github.com/redis/node-redis/blob/bb7845dfe39e385d9e02966d54efb211addb79e0/packages/search/lib/commands/INFO.ts#L25


christian-bromann (approved):
LGTM 👍 

Thanks for your contribution!

PR #8143, opened 2025-05-08
fix(aws): handle image_url object format in block type extraction

caro-cc:
It looks like with the current version, a small bug was introduced: after the recent refactoring, the handling of image_url was changed to assume it's always a string. However, image_url can also be an object with a url property, which seems to be causing errors. I'm reintroducing the previous logic to handle both cases properly

### DIFF
--- a/libs/langchain-aws/src/common.ts
+++ b/libs/langchain-aws/src/common.ts
@@ -297,7 +297,7 @@ function convertLangChainContentBlockToConverseContentBlock<
   }
 
   if (block.type === "image_url") {
-    return extractImageInfo(block.image_url);
+    return extractImageInfo(typeof block.image_url === "string" ? block.image_url : block.image_url.url);
   }
 
   if (block.type === "document" && block.document !== undefined) {
--- a/libs/langchain-aws/src/common.ts
+++ b/libs/langchain-aws/src/common.ts
@@ -297,7 +297,11 @@ function convertLangChainContentBlockToConverseContentBlock<
   }
 
   if (block.type === "image_url") {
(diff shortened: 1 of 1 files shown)


benjamincburns:
Thanks for flagging this!

danny-avila:
Will there be a release for `@langchain/aws` to include this change? I believe this is affecting my app:

```
base64.match is not a function","pregelTaskId":"ade22b10-77dd-5fdb-8cc7-67eedf269664","stack":"TypeError: base64.match is not a function
at extractImageInfo (/node_modules/@langchain/aws/dist/common.cjs:139:32)
at convertLangChainContentBlockToConverseContentBlock (/node_modules/@langchain/aws/dist/common.cjs:176:16)
at /node_modules/@langchain/aws/dist/common.cjs:257:34
at Array.map (<anonymous>)
at convertHumanMessageToConverseMessage (/node_modules/@langchain/aws/dist/common.cjs:257:23)
at /node_modules/@langchain/aws/dist/common.cjs:327:20
at Array.map (<anonymous>)
at convertToConverseMessages (/node_modules/@langchain/aws/dist/common.cjs:322:10)
at ChatBedrockConverse._streamResponseChunks (/node_modules/@langchain/aws/dist/chat_models.cjs:686:96)
at _streamResponseChunks.next (<anonymous>)"
```

https://github.com/langchain-ai/langchainjs/blob/aeac259a15f68c222108b6090ffb692e0f667fd6/libs/langchain-aws/src/common.ts#L227-L230

https://github.com/danny-avila/LibreChat/issues/7354

caro-cc:
@danny-avila in the meantime (until the next release), we are using the previous version (0.1.8), which is not affected by this issue

PR #8154, opened 2025-05-09
test(mongodb): Add tooling to use local Atlas for tests

baileympearson:
Hi! 

This PR intends to make development of the MongoDB integration simpler by utilizing local Atlas if no remote Atlas cluster is configured.  The PR takes inspiration from a [similar PR in langchaingo](https://github.com/tmc/langchaingo/pull/1025), which uses testcontainers to manage the local Atlas' container.

This PR is structured to preserve the existing behavior of the tests is unchanged when a MONGODB_ATLAS_URI is provided.  When it isn't, a local Atlas cluster is launched before tests and destroyed after the tests.  I've modified the tests a bit to work on local atlas (namely, we need to create the search index manually during the test run).


fivemeyestore:
Great implementation of this feature. I particularly like the approach to error handling.

benjamincburns (approved):

benjamincburns:
Thanks, @baileympearson!

Issue #8126, opened 2025-05-06
@langchain/google-genai: Error "Unknown / unsupported author: <agent_name>"


HarshavardhanNetha:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain.js documentation with the integrated search.
- [x] I am sure that this is a bug in LangChain.js rather than my code.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```typescript
try {
      const systemPrompt = promptHelper.test.v2;
      const llm = new ChatGoogleGenerativeAI({
        model: AIService.modelNames.geminiPro,
        temperature: 0.2,
      });

      // const llm = new ChatOpenAI({
      //   model: "gpt-4o",
      //   temperature: 0.2,
      // });

      const workflow = createSupervisor({
        llm: llm,
        agents: [
          this.projectManagerAgent(),
          this.taskManagerAgent(),
          this.crmAgent(),
        ],
        outputMode: "last_message",
        prompt: systemPrompt,
        stateSchema: AIService.GraphState,
        // tools,
      });

      const graph = workflow.compile({
        checkpointer: this.getCheckPoint(),
      });

      const result = await graph.invoke(
        {
          messages: [new HumanMessage(query)],
          query,
        },
      );

      const response = result.messages[result.messages.length - 1];
      let jsonResponse;
      ...
}
...
```

```typescript
static projectManagerAgent = () => {
    // const GraphState = Annotation.Root({
    //   messages: Annotation({
    //     reducer: (x: any, y: any) => x.concat(y),
    //   }),
    // });

    const tools = [
      createProjectTool,
      editProjectTool,
    ];

    const systemPrompt = promptHelper.projectManagerTool.v1;

    const llm = new ChatGoogleGenerativeAI({
      model: AIService.modelNames.geminiPro,
      temperature: 0.2,
    });

    // const llm = new ChatOpenAI({
    //   model: "gpt-4o",
    //   temperature: 0.2,
    // });

    const reactAgent = createReactAgent<typeof AIService.GraphState>({
      llm: llm,
      tools,
      prompt: systemPrompt,
      name: "AI_PM",
    });

    const graph = reactAgent;
    return graph;
  };
```

### Error Message and Stack Trace (if applicable)

Error: Unknown / unsupported author: AI_PM
[...]

PR #8157, opened 2025-05-10
fix(google-genai): convertAuthorToRole: add support for agent names

sklinkert:
When using supervisor or swarm, the current implementation throws `Unknown / unsupported author: <agent_name>` because the author is the name of the agent.

https://github.com/langchain-ai/langchainjs/issues/8126

### DIFF
--- a/libs/langchain-google-genai/src/utils/common.ts
+++ b/libs/langchain-google-genai/src/utils/common.ts
@@ -74,7 +74,9 @@ export function convertAuthorToRole(
     case "function":
       return "function";
     default:
-      throw new Error(`Unknown / unsupported author: ${author}`);
+      // When using supervisor or swarm, the author will be the name of the agent
+      // and the role must be "system"
+      return "system";
   }
 }
 


thinhnguyennt7 (approved):

thinhnguyennt7:
Nice nice, this PR is looking good. I encountered a similar issue while working on Gemini model support. Could we merge this, please?

benjamincburns:
I'm afraid that this fix isn't valid. The effect of this change would be to silently drop the `AIMessage.name` field and replace it with a system message in cases when the `name` doesn't match to a role that is recognized by google.

It's actually a desirable property to throw here. If we don't throw, users who expect that the generated result will be affected by the value of `BaseMessage.name` (e.g. in the case of multi-agent prebuilts like supervisor & swarm) won't have anything that tells them why the model can't differentiate between the various roles at play.

I have a partial fix for this, specifically for prebuilts, that I'm hoping to get out today via langchain-ai/langgraphjs#1174. That will provide an option that inlines the agent name into the message content for providers who don't include the `name` field during generation.

Issue #8136, opened 2025-05-07
Model name mismatch between Langsmith and Hub pull function for Google Vertex AI


yahav-levy:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain.js documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain.js rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

## Minimal Reproducible Example

```typescript
import { pull } from 'langchain/hub/node';

async function demonstrateIssue() {
    try {
        // This prompt uses ChatVertexAI in Langsmith
        const result = await pull("your-username/your-prompt", {
            includeModel: true
        });
        
        // The pull fails to load the model because it's looking for "ChatGoogleVertexAI"
        // instead of "ChatVertexAI" which is what Langsmith returns
        console.log(result);
    } catch (error) {
        console.error(error);
    }
}
```

## Expected vs Actual Behavior

Expected:
- The prompt should be pulled with the Google Vertex AI model properly loaded

Actual:
- The model fails to load because of a name mismatch:
  - Langsmith returns: `"ChatVertexAI"`
  - Hub's pull function checks for: `"ChatGoogleVertexAI"`

## Relevant Code from LangChain

Here's the relevant code from `hub/node` showing the mismatch:

```typescript
// In hub/node
else if (modelName === "ChatGoogleVertexAI") {  // <-- Expects "ChatGoogleVertexAI"
    modelClass = (await import("@langchain/google-vertexai")).ChatVertexAI;
}
```

The imported class is actually `ChatVertexAI` from `@langchain/google-vertexai`, but the check is looking for a different name than what Langsmith provides.

## Steps to Reproduce
1. Create a prompt in Langsmith using Google Vertex AI (which will be saved as "ChatVertexAI")
2. Try to pull the prompt using the hub's pull function with `includeModel: true`
3. Observe that the model fails to load due to the name mismatch

## Error Details

Received unknown model name from prompt hub: "ChatVertexAI"
Warning: Received unknown model name from prompt hub: "ChatVertexAI"
Error: Invalid namespace: $.kwargs.last.kwargs.bound -> {
"lc": 1,
"type": "constructor",
"id": ["langchain", "chat_models", "vertexai", "ChatVertexAI"],
"kwargs": {
"temperature": 0,
"max_tokens": 1024,
[...]

PR #8159, opened 2025-05-11
fix(langchain): vertex ai model name pulled from hub

anadi45:
Fixes # (https://github.com/langchain-ai/langchainjs/issues/8136)

1. `ChatGoogleVertexAI` was incorrect and should be `ChatVertexAI`
2. Fixed langchain doc urls

### DIFF
--- a/langchain/src/hub/base.ts
+++ b/langchain/src/hub/base.ts
@@ -106,7 +106,7 @@ export function generateModelImportMap(
       importMapKey = "chat_models__anthropic";
     } else if (modelLcName === "ChatAzureOpenAI") {
       importMapKey = "chat_models__openai";
-    } else if (modelLcName === "ChatGoogleVertexAI") {
+    } else if (modelLcName === "ChatVertexAI") {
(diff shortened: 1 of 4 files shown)


langchain/src/hub/base.ts
```diff
@@ -106,7 +106,7 @@ export function generateModelImportMap(
       importMapKey = "chat_models__anthropic";
     } else if (modelLcName === "ChatAzureOpenAI") {
       importMapKey = "chat_models__openai";
-    } else if (modelLcName === "ChatGoogleVertexAI") {
+    } else if (modelLcName === "ChatVertexAI") {
```
benjamincburns, line 109: probably neither here nor there w.r.t. this change, but no support for vertex web?

afirstenberg: Both of them use the same `lc_name` and Class name, and python doesn't distinguish between the two.

Certainly part of a much bigger discussion that we need to have at some point.

benjamincburns:
Thanks for the fix, @anadi45!

PR #8161, opened 2025-05-12
feat: chaining of methods

Torrus168:
The core idea is to override methods like `withConfig` so they always return a new instance of model class they were invoked in, allowing us to chain model-specific methods and always have access to all model-specific functionality.


benjamincburns:
Closing for now, as discussed via Slack.

Issue #8172, opened 2025-05-14
ChromeAI not working with the latest Chrome


jtpio:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain.js documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain.js rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

**package.json**

```json
{
  "name": "langchainjs-minimal-example-chromeai",
  "private": true,
  "version": "0.0.1",
  "type": "module",
  "scripts": {
    "dev": "vite --debug",
    "build": "vite build",
    "preview": "vite preview"
  },
  "dependencies": {
    "@langchain/community": "^0.3.30",
    "@langchain/core": "^0.3.40"
  },
  "devDependencies": {
    "vite": "^5.0.8"
  }
}
```

**main.js**

```js
import { ChromeAI } from "@langchain/community/experimental/llms/chrome_ai";

const model = new ChromeAI({
  temperature: 0.5,
  topK: 40,
});

document.getElementById('generate').addEventListener('click', async () => {
  const resultDiv = document.getElementById('result');
  resultDiv.textContent = 'Generating poem...';
  
  try {
    const response = await model.invoke("Write me a short poem please");
    resultDiv.textContent = response;
  } catch (error) {
    resultDiv.textContent = `Error: ${error.message}`;
  }
});
```

**index.html**

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="UTF-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
    <title>LangChain.js ChromeAI Example</title>
    <style>
      body {
        font-family: system-ui, -apple-system, sans-serif;
        max-width: 800px;
        margin: 0 auto;
        padding: 2rem;
      }
      button {
        padding: 0.5rem 1rem;
        margin: 1rem 0;
      }
      #result {
        white-space: pre-wrap;
        padding: 1rem;
        background: #f5f5f5;
        border-radius: 4px;
      }
    </style>
  </head>
  <body>
    <h1>LangChain.js ChromeAI Example</h1>
    <button id="generate">Generate Poem</button>
    <div id="result"></div>
    <script type="module" src="/main.js"></script>
  </body>
</html>
```

### Description

[...]

PR #8173, opened 2025-05-14
fix(community): update ChromeAI to use the new `window.LanguageModel`

jtpio:
Fixes https://github.com/langchain-ai/langchainjs/issues/8172


jtpio:
The patch seems to be working fine when trying with the demo app from https://github.com/jtpio/chromeai-langchainjs-demo published to GitHub Pages:

https://jtpio.github.io/chromeai-langchainjs-demo/

https://github.com/user-attachments/assets/4cd8b21a-250c-465b-81bc-cb4247991d0e

benjamincburns (approved):
Thanks!

jtpio:
Thanks @benjamincburns for the review!

Indeed the failure looks odd and first didn't seem to be related. But the failure doesn't seem to be happening on other PRs or commits on `main`, so not sure either. Maybe one more CI kick would help?

benjamincburns:
> Maybe one more CI kick would help?

Certainly can't hurt!

benjamincburns:
Well that's fun. Surely whatever caused it to fail previously is gone and won't ever happen again... 😬 

Guess I'll merge, though - thanks for the contribution!

jtpio:
Thanks!

PR #8206, opened 2025-05-20
Fix the problem of wrong entry file pointing

helloqian12138:
Fixes # (issue)


benjamincburns:
This is not a valid change, sorry. To be honest, it's not clear to me what this is trying to achieve. Further, if you were to have attempted a `yarn build` this change would've been overwritten by the build process.

If you're having a problem, please raise an issue. Whatever problem you may be having, this is not the fix.

PR #8209, opened 2025-05-20
Add support for GPT-4.1 token counting and max output tokens

KoreanThinker:
I encountered an error when trying to use GPT-4.1:
culate number of tokens, falling back to approximate count Error: Unknown model

This PR adds support for GPT-4.1 by updating the token counting logic and setting the correct max output token size (32768), based on the official documentation:
https://platform.openai.com/docs/models/gpt-4.1

Let me know if any further changes are needed!

### DIFF
--- a/langchain-core/package.json
+++ b/langchain-core/package.json
@@ -37,7 +37,7 @@
     "ansi-styles": "^5.0.0",
     "camelcase": "6",
     "decamelize": "1.2.0",
-    "js-tiktoken": "^1.0.12",
+    "js-tiktoken": "^1.0.20",
     "langsmith": "^0.3.29",
     "mustache": "^4.2.0",
     "p-queue": "^6.6.2",
--- a/langchain-core/src/language_models/base.ts
+++ b/langchain-core/src/language_models/base.ts
@@ -44,6 +44,10 @@ export const getModelNameForTiktoken = (modelName: string): TiktokenModel => {
     return "gpt-4o";
   }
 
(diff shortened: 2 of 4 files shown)


benjamincburns:
Hi @KoreanThinker - thanks for bringing this to my attention. The warning that you're referring to is caused by using an older version of the `tiktoken` library. Bump the version of `tiktoken` in your `package-lock.json`, `yarn.lock`, or `pnpm.lock` file and the error should go away.

As for the change you've proposed here, I'm afraid that I can't accept it, as the code in this area appears to be a bit broken to me. Specifically, the `getModelContextSize` function should be returning the maximum _input_ context size allowed for the model in question. For `gpt-4.1`, this would be `1_047_576`, not `32_768`. The latter number is the maximum output token count for a single generation, which leads me to the next issue...

The second problem that I see here is that `calculateMaxTokens` is broken. It's used to set the `max_tokens` property in the completions API, which limits the number of tokens the model can generate for the given request. The correct way to calculate this would be to return either the count of remaining input tokens given the current prompt size, or the maximum allowed token output for the given model, whichever is smaller.

Sadly there are also two _other_ problems that I see here. First, the current set of models prior to your change will return invalid model context sizes for some current OpenAI models, (e.g. `gpt-4o` would return 4097). Second, all of this code really doesn't belong in core.

Sorry that I can't accept, but thanks for bringing this all to my attention. I'll create a new issue for all of this.

PR #8235, opened 2025-05-22
feat(google-genai): Add baseUrl property to embeddings

schrothbn:
Small addition of the baseUrl property that already exists on the google-genai chat model to the embeddings. This should enable the use of proxies / custom base urls as it's already supported on the chat model.


OlegIvaniv:
@hntrl Any thoughts on this one? Would appreciate if we could get this in as it's a blocker for us

PR #8248, opened 2025-05-23
# Add Support for Anthropic's Built-in Tools

### LINKED ISSUE #8215: Feature: support anthropic's new web search tool
### Privileged issue

### Issue Content

See https://docs.anthropic.com/en/docs/agents-and-tools/tool-use/web-search-tool

bleafman:
# Add Support for Anthropic's Built-in Tools

Adds support for Anthropic's built-in tools (web search, text editor, bash) to bring feature parity with the Python implementation.

- Added support for `server_tool_use`, `web_search_tool_result`, and `web_search_result` content blocks
- Extended message formatting to handle server tool content blocks  
- Added test coverage

Fixes #8215

Twitter Handle: @BrandonLeafman

### DIFF
--- a/libs/langchain-anthropic/src/chat_models.ts
+++ b/libs/langchain-anthropic/src/chat_models.ts
@@ -100,6 +100,26 @@ function isAnthropicTool(tool: any): tool is Anthropic.Messages.Tool {
   return "input_schema" in tool;
 }
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function isBuiltinTool(
+  tool: any
+): tool is
+  | Anthropic.Messages.ToolBash20250124
+  | Anthropic.Messages.ToolTextEditor20250124
+  | Anthropic.Messages.WebSearchTool20250305 {
+  return (
+    typeof tool === "object" &&
+    tool !== null &&
+    typeof tool.type === "string" &&
+    typeof tool.name === "string" &&
+    ((tool.type === "bash_20250124" && tool.name === "bash_20250124") ||
+      (tool.type === "text_editor_20250124" &&
+        tool.name === "text_editor_20250124") ||
+      (tool.type === "web_search_20250305" &&
+        tool.name === "web_search_20250305"))
+  );
+}
+
 /**
  * Input to AnthropicChat class.
  */
@@ -720,11 +740,14 @@ export class ChatAnthropicMessages<
    */
   formatStructuredToolToAnthropic(
     tools: ChatAnthropicCallOptions["tools"]
-  ): Anthropic.Messages.Tool[] | undefined {
+  ): Anthropic.Messages.ToolUnion[] | undefined {
     if (!tools || !tools.length) {
       return undefined;
     }
     return tools.map((tool) => {
+      if (isBuiltinTool(tool)) {
+        return tool;
+      }
       if (isAnthropicTool(tool)) {
         return tool;
       }
--- a/libs/langchain-anthropic/src/tests/chat_models-built-in-tools.int.test.ts
+++ b/libs/langchain-anthropic/src/tests/chat_models-built-in-tools.int.test.ts
@@ -1,0 +1,187 @@
+import { test, expect } from "@jest/globals";
+import { ChatAnthropic } from "../chat_models.js";
+import { HumanMessage, AIMessage } from "@langchain/core/messages";
+import { _convertMessagesToAnthropicPayload } from "../utils/message_inputs.js";
+
+const chatModel = new ChatAnthropic({
+  model: "claude-3-5-sonnet-20241022",
+  temperature: 0,
+  // Enable built-in tools (web search, text editor, bash)
+  // Note: This requires API access to server tools
+});
+
+test("Server Tools Integration - Web Search", async () => {
+  // Test that we can handle a conversation with web search tool usage
+  const messages = [
+    new HumanMessage({
+      content:
+        "Search for the latest news about TypeScript 5.7 release and summarize what you find.",
+    }),
+  ];
+
+  const response = await chatModel.invoke(messages);
+
+  console.log("Response content:", JSON.stringify(response.content, null, 2));
+
+  // The response should be an AIMessage
+  expect(response).toBeInstanceOf(AIMessage);
+  expect(response.content).toBeDefined();
+
+  // The response should contain meaningful content about TypeScript
+  expect(
+    typeof response.content === "string"
+      ? response.content
+      : JSON.stringify(response.content)
+  ).toMatch(/TypeScript|typescript/i);
+
+  console.log("✅ Successfully handled web search request");
+}, 30000); // 30 second timeout for API call
+
+test("Server Tools Integration - Message Round Trip", async () => {
+  // Test that we can properly parse messages with server tool content blocks
+  const conversation = [
+    new HumanMessage({
+      content:
+        "What are the latest developments in AI research? Please search for current information.",
+    }),
+  ];
+
+  try {
+    const response1 = await chatModel.invoke(conversation);
+
+    console.log("First response:", JSON.stringify(response1.content, null, 2));
+
+    // Add the AI response to conversation
+    conversation.push(response1);
+
+    // Continue the conversation
+    conversation.push(
+      new HumanMessage({
+        content:
+          "Based on your search, what are the most promising areas for future research?",
+      })
+    );
+
+    const response2 = await chatModel.invoke(conversation);
+
+    console.log("Second response:", JSON.stringify(response2.content, null, 2));
+
+    // Both responses should be valid
+    expect(response1).toBeInstanceOf(AIMessage);
+    expect(response2).toBeInstanceOf(AIMessage);
+
+    console.log(
+      "✅ Successfully completed multi-turn conversation with server tools"
+    );
+  } catch (error) {
+    // If server tools aren't available, the test should still not crash due to unsupported content format
+    if (
+      error instanceof Error &&
+      error.message.includes("Unsupported message content format")
+    ) {
+      throw new Error(
+        "❌ REGRESSION: 'Unsupported message content format' error returned - this should be fixed!"
+      );
+    }
+
+    // Other errors (like API access issues) are expected and should not fail the test
+    console.log(
+      "⚠️  Server tools may not be available for this API key, but no format errors occurred"
+    );
+  }
+}, 45000); // 45 second timeout for longer conversation
+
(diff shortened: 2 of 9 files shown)


dikaioai:
Please merge this!

dikaioai (approved):

libs/langchain-anthropic/src/chat_models.ts
```diff
@@ -100,6 +100,26 @@ function isAnthropicTool(tool: any): tool is Anthropic.Messages.Tool {
   return "input_schema" in tool;
 }
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function isBuiltinTool(
+  tool: any
```
benjamincburns, line 105: nit: use `unknown` - that's why we have the ESLint error

bleafman: Done!

libs/langchain-anthropic/src/tests/chat_models-built-in-tools.int.test.ts
```diff
@@ -0,0 +1,187 @@
...
+import { _convertMessagesToAnthropicPayload } from "../utils/message_inputs.js";
+
+const chatModel = new ChatAnthropic({
+  model: "claude-3-5-sonnet-20241022",
+  temperature: 0,
+  // Enable built-in tools (web search, text editor, bash)
+  // Note: This requires API access to server tools
+});
+
+test("Server Tools Integration - Web Search", async () => {
```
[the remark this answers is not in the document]
benjamincburns: To be clear, we don't want built-in tools to be enabled by default.

libs/langchain-anthropic/src/tests/chat_models-built-in-tools.int.test.ts
```diff
@@ -0,0 +1,187 @@
...
+        text: "I'll search for that information.",
+      },
+      {
+        type: "server_tool_use",
+        id: "toolu_01ABC123",
+        name: "web_search",
+        input: {
+          query: "latest AI developments",
+        },
+      },
```
benjamincburns, line 109: Have you seen how the built-in tool responses are formatted by the OpenAI provider when using the responses API? Would it be possible to make this structure look similar to that? This will cut down on the number of breaking changes we'll need to make when we eventually go to standardize this type of output.

benjamincburns (changes requested):
Thanks for submitting this, @bleafman! I had some questions, and a few requested changes. Hopefully it's nothing too troublesome, but let me know if you need me to clarify anything.

libs/langchain-anthropic/src/chat_models.ts
```diff
@@ -100,6 +100,26 @@ function isAnthropicTool(tool: any): tool is Anthropic.Messages.Tool {
...
+  return (
+    typeof tool === "object" &&
+    tool !== null &&
+    typeof tool.type === "string" &&
+    typeof tool.name === "string" &&
+    ((tool.type === "bash_20250124" && tool.name === "bash_20250124") ||
+      (tool.type === "text_editor_20250124" &&
+        tool.name === "text_editor_20250124") ||
+      (tool.type === "web_search_20250305" &&
+        tool.name === "web_search_20250305"))
```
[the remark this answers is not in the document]
bleafman: Good catch / callout! The value for `type` in custom tools is `'custom' | null`, however I made a mistake here and the name is actually closer to the "category" of the tool. 

So `ToolBash20250124.name === bash`, `WebSearchTool20250305.name === "web_search"`, etc. 

We can just key off of those and avoid making the assertion too fragile (assuming anthropic keeps the convention for future versions of their tools).

libs/langchain-anthropic/src/chat_models.ts
```diff
@@ -100,6 +100,26 @@ function isAnthropicTool(tool: any): tool is Anthropic.Messages.Tool {
   return "input_schema" in tool;
 }
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function isBuiltinTool(
+  tool: any
+): tool is
+  | Anthropic.Messages.ToolBash20250124
+  | Anthropic.Messages.ToolTextEditor20250124
+  | Anthropic.Messages.WebSearchTool20250305 {
```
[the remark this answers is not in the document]
benjamincburns: I wonder if `export type AnthropicBuiltInTool = Anthropic.Messages.Tool & { type : null }` wouldn't be enough to resolve `AnthropicBuiltInTool` to what we'd be after, without having to explicitly redefine. We like to derive types (and schema) from the provider SDK when possible, as it makes forward compatibility a bit easier when they make additive changes.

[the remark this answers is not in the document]
bleafman: I pushed up a change that uses:

```ts
export type AnthropicBuiltInToolUnion = Exclude<
  Anthropic.Messages.ToolUnion,
  Anthropic.Messages.Tool
>;
```

This separates out the custom tool and _should_ make it forward compatible since it's just deriving from the SDK types (theoretically Anthropic could put something unexpected in the union, but this at least the type inference to be "Everything besides the custom tools").

Let me know what you think about that one.


`export type AnthropicBuiltInTool = Anthropic.Messages.Tool & { type : null }`` doesn't work because the built-in tools all have string literals with they timestamp (ex. `text_editor_20250124`) for the their `type`. The built-in tools and the custom tools don't actually overlap on the `type` property. 

`type` actually would let you narrow, but typing "Any string that isn't literal `'custom'`"  is pretty gross/kluggy 😓

libs/langchain-anthropic/src/tests/chat_models-built-in-tools.int.test.ts
hntrl: @bleafman I would think that's OK. We have some opinions about how we want to standardize those citation/annotation outputs, but that can be for a later iteration.
(we're doing something similar with the annotations key for openai [here](https://github.com/langchain-ai/langchainjs/blob/d766c35db8c7f06a68539faa32049eb9cfa65612/libs/langchain-openai/src/chat_models.ts#L572-L573))

bleafman:
Alright! I think this is ready for review again.

A few notes/questions:

1. Limited down the scope to just the web_search tool, testing all the other tools was going to be too much and the other built-in tools are in beta anyway and may change (or be removed entirely).

2. The `isBuiltinTool` currently allows the bash tool and the text editor tool. I can remove that if we want to avoid the risk around "all observable behaviors of your system will be depended on by somebody"

3. When testing in my app, I noticed that the web search tool streams much more of the response in parallel* compared to non-server tool calls / regular Claude responses. This isn't an issue, more just a note that the behavior for the built in tool will output the same but feels a bit different when streaming vs. a standard bound tool since it's happening on the Anthropic server-side.

*This is speculation, but [a recent breakdown](https://x.com/badlogicgames/status/1929312803799576757) I read about Claude Code seems to indicate that the search results are summarized by another LLM, so it'd make sense that it's streaming these responses in parallel under the hood.

libs/langchain-anthropic/src/tests/chat_models-built-in-tools.int.test.ts
bleafman: @hntrl putting a quick bump on this just because I'm going out of town in a few days and don't want to be a blocker.

Let me know if you have any feedback!

hntrl: looking now, thank you!

libs/langchain-anthropic/src/chat_models.ts
```diff
@@ -102,6 +103,20 @@ function isAnthropicTool(tool: any): tool is Anthropic.Messages.Tool {
...
+function isBuiltinTool(tool: unknown): tool is AnthropicBuiltInToolUnion {
+  return (
+    typeof tool === "object" &&
+    tool !== null &&
+    "type" in tool &&
+    "name" in tool &&
+    typeof tool.type === "string" &&
+    typeof tool.name === "string" &&
+    (tool.name === "bash" ||
+      tool.name === "str_replace_editor" ||
```
hntrl: I'd trend towards only allowing the web search tool if you're reducing scope for this PR, and we can have a connective effort to support the rest of these tools. I'd rather have an explicit failure when I go to pass in one of these tools rather than trying to debug unexpected behaviors.

hntrl (changes requested):
@bleafman Apologies for the delay! Only change I'd make is surrounding us allowing the other built in tools. There's a lockfile issue when I tried merging from main but that should be a trivial fix.

andruhon:
Hi! I have a question about this
I created an instance with set up like this:
```
    model: 'claude-sonnet-4-20250514',
  }).bindTools([
    {
      type: 'web_search_20250305',
      name: 'web_search',
      max_uses: 10,
    },
  ]);
```
but it does not seem to have the bindTools function after default tools were bound.

Consequentially when I am later trying to create react agent

```
    this.agent = createReactAgent({
      llm,
      tools,
      checkpointSaver,
    });
```

It explodes

```
file:///C:/wrk/gaunt-sloth-assistant/node_modules/@langchain/langgraph/dist/prebuilt/react_agent_executor.js:240
                throw new Error(`llm ${llm} must define bindTools method.`);
                      ^

Error: llm [object Object] must define bindTools method.
    at getModelRunnable (file:///C:/wrk/gaunt-sloth-assistant/node_modules/@langchain/langgraph/dist/prebuilt/react_agent_executor.js:240:23)                                                                                       
    at async RunnableCallable.callModel [as func] (file:///C:/wrk/gaunt-sloth-assistant/node_modules/@langchain/langgraph/dist/prebuilt/react_agent_executor.js:289:31) {
```

Am I doing something wrong? I checked dist in node_modules and the code from this PR seems to be there.

"@langchain/anthropic": "^0.3.23",
"@langchain/core": "^0.3.61",
"@anthropic-ai/sdk": "^0.52.0",

andruhon:
Sorry for silly question. I think I figured this out myself, and this appears to be a correct way:

```javascript
createReactAgent({
      llm,
      tools: [
        {
          type: 'web_search_20250305',
          name: 'web_search',
          max_uses: 10,
        },
      ],
      checkpointSaver,
    });
```

it works and *returns search results* in the stream, but *sometimes* later explodes with:

```
BadRequestError: 400 {"type":"error","error":{"type":"invalid_request_error","message":"messages.2.content.0: unexpected `tool_use_id` found in `tool_result` blocks: srvtoolu_abracadabra. Each `tool_result` block must have a corresponding `tool_use` block in the previous message."}}
```

PR #8249, opened 2025-05-24
feat: Query Embedding Caching

YasharF:
This change brings the JS/TypeScript implementation for caching embeddings closer to feature parity with Python LangChain, while preserving backward compatibility for current users.

- Add support for caching query embeddings in `CacheBackedEmbeddings`.  This implementation adds an optional `queryEmbeddingStore` parameter to cache query embeddings separately from document embeddings.
- Update names for consistency: `documentEmbeddingStore`and `queryEmbeddingStore`.
- Add new tests.
- Update documentation to cover query embedding caching and provide additional clarifications.

_Python version features NOT included in this change:_
- Using UUID-based key encoding is not included since it would be a breaking change for existing cache stores.
- Batch process is not included as it requires changes outside the cache_backed related code.

Fixes https://github.com/langchain-ai/langchainjs/discussions/8149

### DIFF
--- a/docs/core_docs/docs/how_to/caching_embeddings.mdx
+++ b/docs/core_docs/docs/how_to/caching_embeddings.mdx
@@ -20,13 +20,65 @@ The cache backed embedder is a wrapper around an embedder that caches embeddings
 
 The text is hashed and the hash is used as the key in the cache.
 
-The main supported way to initialized a `CacheBackedEmbeddings` is the `fromBytesStore` static method. This takes in the following parameters:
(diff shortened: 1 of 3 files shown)


christian-bromann:
@YasharF thanks for your contribution 🙏 our team has been focused on working on LangChain v1 and unfortunately couldn't give PRs as much attention as they need. We are in an effort to picking up where we left off. I will take this PR and discuss it with the team. Please stay tuned!

christian-bromann:
Thanks for the contributions @YasharF. I will go ahead and close this. At this point we don't actively maintain many of the primitives within the former `langchain` now `@langchain/classic` package anymore. I recommend to just copy the original source and use the code directly within your project.

Issue #8221, opened 2025-05-21
OpenAIEmbeddings does not allow setting encoding_format, causing incompatibility with LM Studio (returns float[], not base64)


GregorBiswanger:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain.js documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain.js rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

**WORKS** with LM Studio if encoding_format is set manually (value doesn't matter)

```
const openAiClient = new OpenAIClient(createLMStudioConfig());

async function getEmbeddings(text) {
  const response = await openAiClient.embeddings.create({
    model: 'none',
    input: text,
    encoding_format: 'base64',
  });

  return response.data[0].embedding;
}
```

**DOES NOT WORK** with OpenAIEmbeddings – no way to set encoding_format
```
  const embeddings = new OpenAIEmbeddings({ ...createLMStudioConfig(), model: '' });

  const vectorStore = await QdrantVectorStore.fromExistingCollection(embeddings, {
    url: 'http://localhost:6333',
    collectionName: 'eu-ai',
  });
```

### Error Message and Stack Trace (if applicable)

[Running] node "c:\projects\rag-sample\index.mjs"
✅ Server running at http://localhost:11434
file:///c:/projects/rag-sample/node_modules/@qdrant/openapi-typescript-fetch/dist/esm/fetcher.js:169
                throw new fun.Error(err);
                      ^

ApiError: Bad Request
    at Object.fun [as searchPoints] (file:///c:/projects/rag-sample/node_modules/@qdrant/openapi-typescript-fetch/dist/esm/fetcher.js:169:23)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async QdrantClient.search (file:///c:/projects/rag-sample/node_modules/@qdrant/js-client-rest/dist/esm/qdrant-client.js:165:26)
    at async QdrantVectorStore.similaritySearchVectorWithScore (file:///c:/projects/rag-sample/node_modules/@langchain/qdrant/dist/vectorstores.js:165:25)
    at async QdrantVectorStore.similaritySearch (file:///c:/projects/rag-sample/node_modules/@langchain/core/dist/vectorstores.js:260:25)
[...]

PR #8256, opened 2025-05-26
feat(openai): encoding_format support for embeddings

anadi45:
Fixes # (https://github.com/langchain-ai/langchainjs/issues/8221)

1. This would return base64 string and not decode it into number[]. (openai-node sdk behaviour)


anadi45:
@hntrl please check

libs/langchain-openai/src/embeddings.ts
```diff
@@ -43,6 +43,12 @@ export interface OpenAIEmbeddingsParams extends EmbeddingsParams {
    * See: https://github.com/openai/openai-python/issues/418#issuecomment-1525939500
    */
   stripNewLines?: boolean;
+
+  /**
+   * The format to return the embeddings in. Can be either `float` or
+   * [`base64`](https://pypi.org/project/pybase64/).
```
hntrl, line 55: nit, but can we remove this backlink to pypi base64? we don't need a link really, base64 is universally understood IMO

hntrl (changes requested):
Only thing I'll note is that base64 outputs aren't going to be consistent with the types of the [base embeddings class](https://github.com/langchain-ai/langchainjs/blob/main/langchain-core/src/embeddings.ts) which probably isn't what we want (it indicates that embeddings are represented in num arrays, but oAI allows base64 outputs). CI will probably pass just because their latest types are reflective of the encoding format and we haven't bumped the sdk version yet. So a couple of things would need some attention before we could merge this:

(1) Bump the openai sdk version to latest to properly surface param types
(2) Take another look at the base class to see if we can better manage output types (i'd be amenable to introducting a generic w/ a default on the base class that determines output type)

hntrl:
will rerun when gcp reverts 🙈

libs/langchain-openai/src/embeddings.ts
anadi45: removed

hntrl:
@anadi45 Happy to merge this if we give some thought to what I described above. Out of curiosity what's the use case you have for base64 embeddings?

anadi45:
Hi @hntrl,

I've updated the OpenAI SDK to the latest version, but it appears that the type definitions have not been updated accordingly within the package.

Regarding the alternative approach of updating the base embedding class to use generic types — I did explore that direction. However, I encountered a number of dependencies, particularly around the vector store classes, which made me think if they break existing features..

anadi45:
@hntrl I saw this as an open issue, and thought of solving it. I personally don't have a use case for this.

hntrl:
@anadi45 gotcha, I missed the linked issue! Thanks for taking the time to contribute! 

I'll call out that the sdk version is up to date on main (we had some other things we needed from the latest openAI version). I imagine if the types haven't updated in the oAI sdk but it's reflected in the docs then it's just a change that got missed/ is waiting to land (there's more of those than you think). Maybe it's possible to have an infill type in the class signature that accepts the proper encoding format, then we can just nix it whenever openAI gets around to adding it?

If we make it a generic with a default then I imagine that won't be a breaking change. Could you give more insight into why you think it would be breaking?

```diff
- export interface EmbeddingsInterface {
+ export interface EmbeddingsInterface<TOutput = number[]> { 
  ...
-  embedDocuments(documents: string[]): Promise<number[][]>;
+  embedDocuments(documents: string[]): Promise<TOutput[]>;
  ...
-  embedQuery(document: string): Promise<number[]>;
+  embedQuery(document: string): Promise<TOutput>;
}
```

anadi45:
@hntrl have added generic type to base class

hntrl:
apologies, this got buried! I've rebased and continued this on in #8916

PR #8261, opened 2025-05-27
feat: zod 4 support

hntrl:
Adds support for the newly released zod 4 API in langchain where it's applicable

(( **DO NOT MERGE** )) until these are complete
- [x] PR that (1) removes the zod-to-json-schema dep everywhere and (2) uses these new zod types gets made. CI will be failing here until that gets merged in (#8273)
  - [ ] that PR should also be reviewed/merged into here separately
  - [ ] CI passes after that merge
- [ ] PR inside of langgraph that bumps core dep and uses these new interop utilities


langchain-core/src/tools/index.ts
```diff
@@ -579,30 +588,24 @@ export function tool<
...
 >(
   func: RunnableFunc<SchemaOutputT, ToolOutputT, ToolRunnableConfig>,
   fields: ToolWrapperParams<SchemaT>
-): SchemaT extends z.ZodString
-  ? DynamicTool<ToolOutputT>
-  : SchemaT extends ZodObjectAny
-  ? DynamicStructuredTool<SchemaT, SchemaOutputT, SchemaInputT, ToolOutputT>
-  :
-      | DynamicStructuredTool<SchemaT, SchemaOutputT, SchemaInputT, ToolOutputT>
-      | DynamicTool<ToolOutputT> {
```
hntrl, line 596: This conditional return type is unnecessary as it gets covered by the overloads above. Had to make this change because adding the InteropZodType layer to the return type broke typescript's recursion limit. I've added type tests later in this PR to add a signal check for this

langchain-core/src/utils/json_schema.ts
```diff
@@ -12,9 +12,12 @@ export { deepCompareStrict, Validator } from "@cfworker/json-schema";
  * @param schema - The schema to convert.
  * @returns The converted schema.
  */
-export function toJsonSchema(schema: z.ZodType | JSONSchema): JSONSchema {
-  if (isZodSchema(schema)) {
-    return zodToJsonSchema(schema) as JSONSchema;
+export function toJsonSchema(schema: InteropZodType | JSONSchema): JSONSchema {
```
dqbd, line 15: Think that we want to export this as an `@langchain/core` entrypoint so that we can start migrating `langchain` as well

hntrl: do you mean this function being its own distinctive entrypoint? There's already an exports for json_schema
https://github.com/langchain-ai/langchainjs/blob/945e99a801a9eb0c331cd58ed9f6eb86eb70b23b/langchain-core/package.json#L569

langchain-core/src/utils/json_schema.ts
```diff
@@ -1,7 +1,7 @@
-import type { z } from "zod";
+import * as z4 from "zod/v4";
```
dqbd: I think we need to import from `zod/v4/core`?

hntrl:
Closing in favor of #8317

PR #8269, opened 2025-05-29
fix(mistralai): handle both Zod and JSON schemas in tool conversion

christian-bromann:
## Description

Fixes a bug where using MCP (Model Context Protocol) tools with `ChatMistralAI` would throw a `TypeError: Cannot read properties of undefined (reading 'typeName')` error.

fixes #8171

## Problem

The `_convertToolToMistralTool` function was assuming all tool schemas were Zod schemas and unconditionally calling `zodToJsonSchema()` on them. However, MCP tools from `@langchain/mcp-adapters` provide tools with JSON schemas already, not Zod schemas. This mismatch caused the `zodToJsonSchema` function to fail when trying to access the `typeName` property on a plain JSON schema object.

**Error Stack Trace:**
```
TypeError: Cannot read properties of undefined (reading 'typeName')
    at parseDef (/node_modules/.pnpm/zod-to-json-schema@3.24.3_zod@3.24.2/node_modules/zod-to-json-schema/dist/cjs/parseDef.js:22:77)
    at zodToJsonSchema (/node_modules/.pnpm/zod-to-json-schema@3.24.3_zod@3.24.2/node_modules/zod-to-json-schema/dist/cjs/zodToJsonSchema.js:22:45)
    at _convertToolToMistralTool (/node_modules/.pnpm/@langchain+mistralai@0.2.0_@langchain+core@0.3.55_openai@4.85.4_ws@8.18.1_zod@3.24.2__/node_modules/@langchain/mistralai/dist/chat_models.cjs:261:70)
```

## Solution

Modified the `_convertToolToMistralTool` function to:

1. Check if the tool schema is a Zod schema using the existing `isZodSchema` helper function
2. Only call `zodToJsonSchema()` if it's actually a Zod schema
3. Use the schema directly if it's already a JSON schema (like MCP tools)

**Code Changes:**
```typescript
// Before
parameters: zodToJsonSchema(tool.schema),

// After  
const parameters = isZodSchema(tool.schema) 
  ? zodToJsonSchema(tool.schema)
  : tool.schema;
```

## Testing

- ✅ Added unit test covering both Zod schema tools and JSON schema tools
- ✅ Verified existing unit tests still pass
- ✅ Test specifically covers `DynamicStructuredTool` with JSON schema (simulating MCP tools)
- ✅ Integration tests pass (where API keys are available)

## Compatibility

This change maintains backward compatibility with existing Zod-based tools while adding support for JSON schema-based tools from MCP adapters.

**Affected Packages:**
[...]

PR #8272, opened 2025-05-29
fix(community): handle both Zod and JSON schemas in Watson tool conversion

christian-bromann:
## Description

Fixes the same `TypeError: Cannot read properties of undefined (reading 'typeName')` error in `ChatWatsonx` that was fixed for `ChatMistralAI` in #8269.

## Problem

The `_convertToolToWatsonxTool` function had the same issue as MistralAI - it was assuming all tool schemas were Zod schemas and unconditionally calling `zodToJsonSchema()` on them. MCP tools and other `DynamicStructuredTool` instances already have JSON schemas, causing the conversion to fail.

## Solution

Applied the same fix as PR #8269:
- Added `isZodSchema()` check before attempting schema conversion
- Only call `zodToJsonSchema()` for actual Zod schemas
- Pass through JSON schemas unchanged

**Changes:**
- Modified `_convertToolToWatsonxTool()` in `libs/langchain-community/src/chat_models/ibm.ts`
- Added unit test covering both schema types

## Testing

- [x] Unit tests pass
- [x] Backward compatibility maintained for existing Zod tools
- [x] Forward compatibility verified for JSON schema tools

This enables MCP (Model Context Protocol) tools to work with `ChatWatsonx`, just like the fix for `ChatMistralAI`.

## Related

- Same root cause and solution as https://github.com/langchain-ai/langchainjs/pull/8269
- No breaking changes

### DIFF
--- a/libs/langchain-community/src/chat_models/ibm.ts
+++ b/libs/langchain-community/src/chat_models/ibm.ts
@@ -139,12 +139,17 @@ function _convertToolToWatsonxTool(
     if ("type" in tool) {
       return tool as WatsonXAI.TextChatParameterTools;
     }
+    // Check if schema is a Zod schema or already a JSON schema
+    const parameters = isZodSchema(tool.schema)
+      ? zodToJsonSchema(tool.schema)
+      : tool.schema;
+
     return {
       type: "function",
       function: {
         name: tool.name,
         description: tool.description ?? "Tool: " + tool.name,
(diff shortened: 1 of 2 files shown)


hntrl:
Thanks for the fix @christian-bromann!
I've pulled this commit under #8273. We're making an effort to support the newly released zod 4 API that comes with some modifications to how we serialize schemas, and this falls under that umbrella.

PR #8273, opened 2025-05-29
chore: use new zod utils in every package

hntrl:
Sibling PR for https://github.com/langchain-ai/langchainjs/pull/8316.

This is all apart of an effort to support zod v4 and lay the groundwork to use more standardized validation interfaces


libs/langchain-openai/package.json
```diff
@@ -37,11 +37,10 @@
   "dependencies": {
     "js-tiktoken": "^1.0.12",
     "openai": "^4.96.0",
-    "zod": "^3.25.32",
-    "zod-to-json-schema": "^3.22.3"
+    "zod": "3.25.32"
```
tobymac-siro, line 40: Why was this pinned to this specific zod version?

PR #8280, opened 2025-05-30
chor: update dependency: mongodb

YasharF:
There are multiple versions of mongodb across the monorepo including an older 5.x version.  The PR upgrades them all to the latest version.

Some of the code is going to be deprecated in the near future as part of the move to @langchain/mongodb, but since that is going to be in some time from now, it is probably worth keeping the dependency for it up-to-date for the time being.


dqbd:
Thank you @YasharF!

PR #8282, opened 2025-05-30
fix(mcp-adapters): guard against undefined schema when binding tools

simozampa:
**Summary**
When loading MCP tools, we dynamically derive content-type literals from the Zod schema at `CallToolResultSchema.shape.content`. If that schema is wrapped in `ZodDefault` (or similar), the old code
```typescript
const callToolResultContentSchema = CallToolResultSchema.shape.content.element;
```
fails because `.element` isn’t defined on the wrapper. 
This PR adds an unwrapSchema() helper that peels off `ZodDefault`, `ZodOptional`, `ZodNullable`, and `ZodEffects` layers before accessing `.element`, ensuring we always get the underlying array schema.

**Changes**
- Introduce `unwrapSchema(schema)` to recursively strip common Zod wrappers
- Replace direct `.shape.content.element` access with:
    ```typescript 
    const rawContent = CallToolResultSchema.shape.content;
    const arraySchema = unwrapSchema(rawContent);
    const callToolResultContentSchema = arraySchema.element;
    ```
- No behavioral change when no wrappers are present; only adds resilience for schemas using `.default()`, `.optional()`, `.nullable()`, or `.transform()` wrappers.

### DIFF
--- a/libs/langchain-mcp-adapters/src/types.ts
+++ b/libs/langchain-mcp-adapters/src/types.ts
@@ -7,6 +7,10 @@ import {
   ZodRawShape,
   UnknownKeysParam,
   Primitive,
+  ZodDefault,
+  ZodOptional,
+  ZodNullable,
+  ZodEffects,
 } from "zod";
 import type { OAuthClientProvider } from "@modelcontextprotocol/sdk/client/auth.js";
 import { CallToolResultSchema } from "@modelcontextprotocol/sdk/types.js";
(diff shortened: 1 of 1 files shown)


hntrl (changes requested):
Hey @simozampa! Thanks for the PR!

There seems to be an underlying issue where the cjs/esm and types exports of mcp disagree with each other. The type checks here passed for a release because we're relying on the sdk to expose the schema (a pattern that lets us be forward-compatible), but the underlying type associated with the sdk schema is `z.ZodArray<...>`, not `z.ZodDefault<z.ZodArray<>>` even though that's how it's exported in cjs/esm. I'll raise an issue in their repo here in a bit. I don't anticipate that this mismatch will stick.

I think you're on the right track of accommodating this, but this implementation needs better type safety. I've added review that uses a more forgiving approach to determining defaults that I worked up. I also don't think we need to accommodate transform & optional & nullable since that just seems like we're making too broad of an assumption about how that exported schema will look going into the future.

If you wouldn't mind also adding a test that tests that the `MultiServerMcpClient` can be constructed (without the type errors being thrown) then this PR will be in a really good spot. You can create a new directory & test file in `libs/langchain-mcp-adapters/src/tests` and the test scripts should pick up automatically. If I don't hear from you by EOD then I'll work on picking up where you left off to get a patch release out quickly.

I see that you're a first time contributor (🥳), so I'll make a call out that it's a good idea to run `yarn build && yarn format && yarn lint` before pushing up commits as this will catch errors before it hits CI.

(fixes #8284)

hntrl:
Scratch that, I may have gotten ahead of myself. They made a change to use a default zod schema and pushed it as a patch bump, so this started breaking as soon as that release went out. All we need here is to bump the sdk version in package.json and target the element properly:

```
const callToolResultContentSchema =
  CallToolResultSchema.shape.content._def.innerType.element;
```

https://github.com/modelcontextprotocol/typescript-sdk/releases/tag/1.11.5

MilanCancarevic:
@simozampa thanks for the fix!

PR #8288, opened 2025-05-30
Ensure the GithubRepoLoader path is URI Encoded when fetching repo files

dylandignan:
Currently, if you are processing a GithubRepo with a file or directory name with characters like `%5F` (which is allowed by Github) this loader will return 404. Instead, we need to URI Encode the path to ensure it can be reached.


christian-bromann:
Duplicate of https://github.com/langchain-ai/langchainjs/pull/8847

PR #8251, opened 2025-05-26
feat(mongodb): add MongoDB LLM cache

YasharF:
This PR is adds `MongoDBCache` and `MongoDBAtlasSemanticCache` functionality for feature parity with Python LangChain.

- Add `MongoDBCache` and `MongoDBAtlasSemanticCache` classes to `@langchain/mongodb` for key-value and semantic LLM caching.
- Implement integration tests for both cache types.
- Update documentation:
  - Add guide for using MongoDB as a key-value cache.
  - Add guide for using MongoDB Atlas as semantic cache.
  - Added additional information and clarification on semantic vs. key-value caching.
- Fix test DB usage to consistently use `langchain_test` collection for all MongoDB tests and clean up the DB after running the tests when not using a container.

Partial Fix for https://github.com/langchain-ai/langchainjs/discussions/7048 (fixes the mongoDB use case) 

x: YasharF3
linkedin.com/in/yasharf

PR #8295, opened 2025-06-02
test(mongodb): add more comprehensive testing to MongoDB integration 

baileympearson:
We (MongoDB) have started running the langchainjs MongoDB integration periodically in our own CI to get early alerting if something were to break.  We'd like to also make sure the tests provide good coverage of the integration.

This PR updates the existing MongoDB integration test suite to more exhaustively test the integration.

### DIFF
--- a/libs/langchain-mongodb/src/storage.ts
+++ b/libs/langchain-mongodb/src/storage.ts
@@ -182,9 +182,7 @@ export class MongoDBStore extends BaseStore<string, Uint8Array> {
       regexPattern = `^${this._getPrefixedKey(".*")}`;
     }
 
-    let totalDocsYielded = 0;
-
-    let cursor = await this.collection
+    const cursor = this.collection
       .find(
         {
           [this.primaryKey]: { $regex: regexPattern },
@@ -193,29 +191,10 @@ export class MongoDBStore extends BaseStore<string, Uint8Array> {
           batchSize: this.yieldKeysScanBatchSize,
         }
       )
-      .toArray();
+      .map((key) => this._getDeprefixedKey(key[this.primaryKey]));
 
-    for (const key of cursor) {
-      yield this._getDeprefixedKey(key[this.primaryKey]);
-    }
-    totalDocsYielded += cursor.length;
-    while (cursor.length !== 0) {
-      cursor = await this.collection
-        .find(
-          {
-            [this.primaryKey]: { $regex: regexPattern },
-          },
-          {
-            batchSize: this.yieldKeysScanBatchSize,
-            skip: totalDocsYielded,
-          }
-        )
-        .toArray();
-
-      for (const key of cursor) {
-        yield this._getDeprefixedKey(key[this.primaryKey]);
-      }
-      totalDocsYielded += cursor.length;
+    for await (const document of cursor) {
+      yield document;
     }
   }
 }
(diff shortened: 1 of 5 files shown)


libs/langchain-mongodb/src/storage.ts
```diff
@@ -193,29 +191,10 @@ export class MongoDBStore extends BaseStore<string, Uint8Array> {
...
-          }
-        )
-        .toArray();
-
-      for (const key of cursor) {
-        yield this._getDeprefixedKey(key[this.primaryKey]);
-      }
-      totalDocsYielded += cursor.length;
+    for await (const document of cursor) {
+      yield document;
```
baileympearson, line 197: This was a drive-by cleanup.  When iterating a MongoDB cursor, we will automatically fetch batches of size `batchSize` when we need more documents.  So by the time this for-await loop is exhausted, there are no more documents left in the cursor.  The below loop would then do nothing.

hntrl: We probably won't issue a release for just this, but this is a welcome change nonetheless!

hntrl (approved):

libs/langchain-mongodb/src/tests/vectorstores.int.test.ts
```diff
@@ -240,8 +246,676 @@ test("MongoDBAtlasVectorSearch upsert", async () => {
...
+
+    let deletes = 0;
+    client.on("commandStarted", (event) => {
+      if (event.commandName === "delete") {
+        deletes += 1;
+      }
+    });
+
+    // @ts-expect-error: Mock embeddings
+    const vectorStore = new PatchedVectorStore(embeddings, {
```
YasharF, line 755: This test was hanging for me and failing after a timeout even after increasing the timeout to 2 minutes. I suspect it is because PatchedVectorStore waits for all documents to be queryable, which doesn’t seem to work [edited]. I changed this to use MongoDBAtlasVectorSearch instead of PatchedVectorStore and the test passed.

```
  delete method
    ✓ removes documents by ids (357 ms)
    ✓ handles large number of ids by chunking (11260 ms)
    ✓ ignores non-existent ids (368 ms)
```

YasharF: I did more digging and confirmed that this code is fine—it’s not where the bug originates. However, it does reveal the bug elsewhere. Switching to `MongoDBAtlasVectorSearch` allows the test to proceed because `MongoDBAtlasVectorSearch` doesn't hang _**if there is no search index**_, skipping the infinite loop in `PatchedVectorStore`.

The actual bug is in this line:
https://github.com/langchain-ai/langchainjs/blob/0a6f745553dae08e9b3acd4e56a0c6ca2a021654/libs/langchain-mongodb/src/tests/vectorstores.int.test.ts#L45

This issue occurs when someone is using `MONGODB_ATLAS_URI` to run tests against a DB that is not in a test container. When `MONGODB_ATLAS_URI` is provided, the test creates a collection, but this line causes the test to skip index creation for it.  Since the index doesn't exist, `this.similaritySearch("sandwich", documents.length);` in `PatchedVectorStore` doesn't return anything, which puts us in an infinite for loop.

I’ll be adding a fix as part of PR https://github.com/langchain-ai/langchainjs/pull/8251 in the next few hours since I am already addressing other test issues for such setup in there.

Issue #8132, opened 2025-05-07
Title: ERR_REQUIRE_ESM with @langchain/google-webauth in NestJS (CommonJS) Project


ialberquilla:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain.js documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain.js rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```typescript
        return new ChatGoogle({
          model: 'gemini-2.5-flash-preview-04-17',
          maxReasoningTokens: 0,
          apiVersion: 'v1beta',
          platformType: 'gai',  
          safetySettings: setSafetySettings(),
          apiKey: configService.get<string>('GOOGLE_AI_API_KEY'), 
        });
```

### Error Message and Stack Trace (if applicable)

/root/repo/agent/backend/node_modules/@langchain/google-webauth/dist/auth.cjs:6
const google_1 = require("web-auth-library/google");
                 ^

Error [ERR_REQUIRE_ESM]: require() of ES Module /root/repo/agent/backend/node_modules/web-auth-library/dist/google/index.js from /root/repo/agent/backend/node_modules/@langchain/google-webauth/dist/auth.cjs not supported.
Instead change the require of index.js in /root/repo/agent/backend/node_modules/@langchain/google-webauth/dist/auth.cjs to a dynamic import() which is available in all CommonJS modules.
    at Object.<anonymous> (/root/repo/agent/backend/node_modules/@langchain/google-webauth/dist/auth.cjs:6:18)
    at Object.<anonymous> (/root/repo/data-agent/backend/node_modules/@langchain/google-webauth/dist/chat_models.cjs:5:19)
    at Object.<anonymous> (/root/repo/data-agent/backend/node_modules/@langchain/google-webauth/dist/index.cjs:17:14)
    at Object.<anonymous> (/root/repo/data-agent/backend/node_modules/@langchain/google-webauth/index.cjs:1:18)
    // ... (further stack trace related to project structure)
    at Object.<anonymous> (/root/repo/data-agent/backend/dist/main.js:4:22) {
  code: 'ERR_REQUIRE_ESM'
}

### Description

- I'm trying to use the langchain library to: integrate Google services that rely on @langchain/google-webauth within a NestJS (TypeScript, CommonJS) backend application.

[...]

PR #8315, opened 2025-06-05
chore: upgrade to TypeScript 5.8

dqbd:
Using similar tactics as in https://github.com/langchain-ai/langgraphjs/pull/1216, with addition of following:

1. Fix new TypeScript errors that are actually valid bugs (unreachable code after `??`)
2. Dedupe packages such as `mongo`, `dpdm` etc. to avoid TypeScript resolution errors whenever possible (exception is Zep since it imports `langchain` as peerDep)
3. Add `@ts-ignore` flags to package imports that export types that are incompatible with CJS (@langchain/google-webauth, @langchain/ollama). `@langchain/ollama` seems to properly work with CJS, whereas `@langchain/google-webauth` does not export CJS at all. 
4. Fix default imports such as @langchain/anthropic, Neo4j etc. (`dep.default` -> `dep`)


libs/langchain-community/src/stores/message/mongodb.ts
```diff
@@ -27,13 +28,15 @@ export interface MongoDBChatMessageHistoryInput {
 export class MongoDBChatMessageHistory extends BaseListChatMessageHistory {
   lc_namespace = ["langchain", "stores", "message", "mongodb"];
 
-  private collection: Collection<MongoDBDocument>;
+  private collection: Collection<{ messages: StoredMessage[] }>;
```
hntrl: is this intentional? Is the typescript bump causing this to throw type errors?

langchain-core/src/utils/async_caller.ts
```diff
@@ -94,7 +94,9 @@ export class AsyncCaller {
     this.onFailedAttempt =
       params.onFailedAttempt ?? defaultFailedAttemptHandler;
 
-    const PQueue = "default" in PQueueMod ? PQueueMod.default : PQueueMod;
+    const PQueue = (
+      "default" in PQueueMod ? PQueueMod.default : PQueueMod
+    ) as typeof PQueueMod;
```
hntrl, line 99: can we just ditch this entirely? This reads we're trying to conditionally target cjs vs. esm at runtime, but wouldn't that be totally redundant?

examples/src/use_cases/human_in_the_loop/accept-feedback.ts
```diff
@@ -5,8 +5,8 @@ import { callToolList, model } from "./helpers.js";
 // Use readline to ask the user for approval
 function askQuestion(question: string): Promise<string> {
   const rl = readline.createInterface({
-    input: process.stdin,
-    output: process.stdout,
+    input: process.stdin as unknown as NodeJS.ReadableStream,
+    output: process.stdout as unknown as NodeJS.WritableStream,
```
hntrl, line 9: could we just add @types/node to examples/tsconfig.json?

examples/src/guides/expression_language/zep/zep_cloud_message_history.ts
```diff
@@ -36,6 +36,7 @@ const chainWithHistory = new RunnableWithMessageHistory({
   runnable: chain,
   getMessageHistory: (sessionId) =>
     new ZepCloudChatMessageHistory({
+      // @ts-expect-error - Mismatched ZepClient due to resolution issues
```
hntrl: are we sure these error messages are something that we can ignore? I'm not buddy-buddy with the build process yet, but if this is the result of the `moduleResolution` change it prob won't hurt to investigate this more

libs/langchain-azure-openai/package.json
```diff
@@ -16,7 +16,7 @@
     "build": "yarn turbo:command build:internal --filter=@langchain/azure-openai",
     "build:internal": "yarn lc_build --create-entrypoints --pre --tree-shaking",
     "lint:eslint": "NODE_OPTIONS=--max-old-space-size=4096 eslint --cache --ext .ts,.js src/",
-    "lint:dpdm": "dpdm --exit-code circular:1 --no-warning --no-tree src/*.ts src/**/*.ts",
+    "lint:dpdm": "dpdm --skip-dynamic-imports circular --exit-code circular:1 --no-warning --no-tree src/*.ts src/**/*.ts",
```
hntrl, line 19: what dependencies are circular? I'm assuming is just checking at the import level and not evaluating actual symbols

libs/langchain-google-webauth/src/auth.ts
```diff
@@ -9,6 +9,8 @@ import {
   getAccessToken,
   getCredentials,
   Credentials,
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore CJS type resolution workaround
```
hntrl, line 13: this gives me pause. are we testing that these libs are importable from cjs?

libs/langchain-community/src/stores/message/mongodb.ts
dqbd: Did pin the `mongodb` version to `6.11.0` to avoid having duplicate packages, which caused this type error. Treating this as a no-op since the types of private fields are omitted from `.d.ts`

afirstenberg:
I saw the comment about `@langchain/google-webauth`. Will this address the problems that have been documented in #8132 ?

libs/langchain-google-webauth/src/auth.ts
afirstenberg: I don't have an answer to @hntrl 's  question, but see comment below about #8132 and if there is anything we're doing with this update that might also fix it.

libs/langchain-azure-openai/package.json
dqbd: Previous version of `dpdm` did not detect circular dependencies caused by dynamic imports. An example of such circular import can be found here: https://github.com/langchain-ai/langchainjs/blob/f9a3ef28ddc499be3c73a8c1d983324c22535f65/langchain-core/src/prompts/base.ts#L201

(`src/prompts/base.ts` <-> `src/prompts/string.ts` <-> `src/prompts/prompt.ts`)

dqbd:
> I saw the comment about `@langchain/google-webauth`. Will this address the problems that have been documented in #8132 ?

I don't think so, since `web-auth-library` does not ship with CJS, only ESM unfortunately. However, I think `@langchain/google-gauth` should be working fine with CJS though.

libs/langchain-google-webauth/src/auth.ts
dqbd: Added CJS tests for `ollama` and `google-gauth`. `google-webauth` does not support CJS due to upstream dep.

examples/src/use_cases/human_in_the_loop/accept-feedback.ts
dqbd: Afaict this is not an issue with lack of `@types/node`, rather than due to weird issues with ReadableStream != ReadableStream.

```
Conversion of type 'ReadStream & { fd: 0; }' to type 'ReadableStream' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Type 'ReadStream & { fd: 0; }' is missing the following properties from type 'ReadableStream': off, removeAllListeners, setMaxListeners, getMaxListeners, and 4 more.
```

langchain-core/src/utils/async_caller.ts
dqbd: I'd rather err on side of caution since this is the async caller we're dealing with?

examples/src/guides/expression_language/zep/zep_cloud_message_history.ts
dqbd: This is primarily due to the fact that `@getzep/zep-cloud` contains a peer dependency on `langchain`, which in turn causes the package to be present _twice_.

Fixed by making `@getzep/zep-cloud` a peer dep

libs/langchain-community/src/stores/message/mongodb.ts
YasharF: I'm not sure which way is better, but I have a slightly different approach to fix the issue when bumping the mongodb version in https://github.com/langchain-ai/langchainjs/pull/8280/files .  That open PR bumps mongdb to the latest (6.17).  Hope it helps.

libs/langchain-community/src/vectorstores/tests/azure_cosmosdb.test.ts
```diff
@@ -104,7 +104,12 @@ test("AzureCosmosDBVectorStore deletes documents", async () => {
   expect(mockCollection.deleteMany).toHaveBeenCalledTimes(1);
   expect(mockCollection.deleteMany).toHaveBeenCalledWith({});
 
-  await store.delete({ ids: ["id1234567890", "id2345678901"] });
+  await store.delete({
+    ids: [
+      Buffer.from("id1234567890").toString("hex"),
+      Buffer.from("id2345678901").toString("hex"),
+    ],
```
YasharF: Instead of doing a conversion in code, you can probably just use valid mongodb ids like in the other PR https://github.com/langchain-ai/langchainjs/pull/8280/files#diff-338159b176a18cb4e8217186ae8413997b376739fb0b0301a54ef56abfec3599

```
await store.delete({
    ids: ["507f1f77bcf86cd799439011", "507f1f77bcf86cd799439012"],
  });
```

The reason that the original test case stops working after the mongodb version bump is because mongodb 6 removed 12 length string support from ObjectId constructors.
Release note: https://github.com/mongodb/js-bson/blob/main/HISTORY.md#600-2023-08-24
Change PR: https://github.com/mongodb/js-bson/pull/601

afirstenberg:
> > I saw the comment about `@langchain/google-webauth`. Will this address the problems that have been documented in #8132 ?
> 
> I don't think so, since `web-auth-library` does not ship with CJS, only ESM unfortunately. However, I think `@langchain/google-gauth` should be working fine with CJS though.

We've never gotten a report on the gauth version having issues, so I wasn't worried about that.
Unfortunately, the reason we have so many libraries is because the gauth version doesn't work in some environments, so we needed to provide another auth path.

Anyway - outside this issue, apparently. Thanks!

libs/langchain-google-webauth/src/auth.ts
hntrl: Would it make sense/ would it be low lift to just ditch cjs exports for this package? That way dependents on the package are getting errors from typescript instead of having to re-discover this when they go to use it

At the very least we might add something into the package readme to say it isn't compatible with cjs 👀

libs/langchain-community/src/vectorstores/tests/azure_cosmosdb.test.ts
hntrl: this way of initializing object id's is still valid though yeah?

hntrl:
> Unfortunately, the reason we have so many libraries is because the gauth version doesn't work in some environments

@afirstenberg I'm assuming the other environments in question is web?

libs/langchain-community/src/vectorstores/tests/azure_cosmosdb.test.ts
YasharF: I find it more helpful to have explicit and readable values instead of transformations, specially if later something breaks and needs to be debugged.  With that said, they both do end up with the required 24-char HEX strings as objectIDs.

hntrl: Where this is just compat fixes for tests I'm not sure if it totally matters (we do drive-by changes like this all the time in tests when we upgrade SDKs, so if something breaks it's pretty trivial). I'd have different opinions if it was public docs since we want to model best practices there.

PR #8316, opened 2025-06-05
patch(core): expose bindings for Zod 4 from core

dqbd:
The purpose of this PR is to centralize a number of utilities for working with zod 3 & 4 schemas, and make some non-invasive changes to core APIs that allows for zod 4 support in a non-breaking way.

This is a sibling PR for #8273, which is where these utilities actually get used. These PRs are divvied up since the core changes need to happen independently to allow us to pin a release to all the integrations packages that want to consume these new utilities. Once this PR lands, we can cut a core release and update the peer dependency requirement for all affected integrations packages in #8273 (which will fix the CI failures there)

Some criteria was used when making these changes:
- cutting a release of `@langchain/core` wouldn't break applications since this has core API changes
- zod 4 can be "opt-in" for integrations packages + LG to avoid any release dominos


langchain-core/src/tools/index.ts
```diff
@@ -71,7 +82,7 @@ export { ToolInputParsingException };
  * Base class for Tools that accept input of any shape defined by a Zod schema.
  */
 export abstract class StructuredTool<
-    SchemaT extends ToolInputSchemaBase = ToolInputSchemaBase,
```
hntrl, line 74: 🙈

yarn.lock
hntrl, line 1: This might seem like a misnomer for the zod 4 rollout, but test-exports-tsc is trying to shallow evaluate two different versions of zod (between langchain-core and langchain-openai) which quickly hits typescript recursion limits.

I've tested this independently and confirmed that zod deps disagreeing like this won't impact in a breaking way. `test-exports-tsc` fails because it resolves deps in a non-traditional way which causes the branching zod deps to resolve differently then how it does in practice.

libs/langchain-openai/package.json
hntrl, line 1: This might seem like a misnomer for the zod 4 rollout, but test-exports-tsc is trying to shallow evaluate two different versions of zod (between langchain-core and langchain-openai) which quickly hits typescript recursion limits.

I've tested this independently and confirmed that zod deps disagreeing like this won't impact in a breaking way. `test-exports-tsc` fails because it resolves deps in a non-traditional way which causes the branching zod deps to resolve differently then how it does in practice.

(there was a previously passing CI run before this change, but an intermittent release of langchain-openai in-between then and now pinned deps differently afaict)

hntrl, line 1: CI passes w/ this change, but this might be more indicative of a bigger problem with how package managers are going to reconcile differing zod deps when core is bumped

hntrl: confirmed this with a core release candidate (`@langchain/core@0.3.58-rc.0`), this shouldn't be an issue

PR #8317, opened 2025-06-05
feat(core): integrate zod 4 in backwards compatible manner

dqbd:


langchain-core/src/language_models/chat_models.ts
hntrl, line 1: My big question would be if we have to add overloads like this everywhere that involves zod types. I guess I'm not totally against it (it lets us ship zod 4 support in a better fashion), but I think we need to make a commitment towards abstracting schema types like we've been talking about rather than ballooning these types every time an upgrade like this happens.

Part of my hope with hoisting everything to a single type is it gives us a single entrypoint to modify what kind of schemas we accept rather than inferring that from zod directly which (theoretically) means this would be the only breaking change we'd need to involve on the path to becoming schema lib agnostic. I can't say that with 100% certainty though, so if we don't want to sit on this any longer than I'm all for this!

At any rate we will encounter a breaking change regarding schemas at some point -- as soon as zod@4 comes out we have to roll back on these changes. Maybe a better candidate for the breaking change is the introduction of a standard schema lib (like a langchain xsschema 👀)

hntrl (commented):
Does this work against an integration upgrade?

dqbd:
> Does this work against an integration upgrade?

At quick glance & small test we can add an additional override to handle zod 4 schemas

langchain-core/src/language_models/chat_models.ts
dqbd: Agreed, let's make this for a next minor bump?

PR #8321, opened 2025-06-06
Fix param in langchain-google-genai README

MilanCancarevic:
I was updating package '@langchain/google-genai' to 0.2.10 from 0.1.8 and noticed that docs are mentioning old param `modelName`. By looking at typings from `ChatGoogleGenerativeAI` i see that it has been renamed to `model`

`GoogleGenerativeAIEmbeddingsParams` has alias for both so this raises question should alias be supported for `ChatGoogleGenerativeAI` - if yes i would like to work on it

### DIFF
--- a/libs/langchain-google-genai/README.md
+++ b/libs/langchain-google-genai/README.md
@@ -53,7 +53,7 @@ import { ChatGoogleGenerativeAI } from "@langchain/google-genai";
 import { HumanMessage } from "@langchain/core/messages";
 
 const model = new ChatGoogleGenerativeAI({
-  modelName: "gemini-pro",
+  model: "gemini-pro",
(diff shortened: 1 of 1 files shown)


hntrl:
Thanks for the change @MilanCancarevic!

Do you mean you want to see both `modelName` and `model` as params for `ChatGoogleGenerativeAI`? I'd call out that `modelName` is a broad deprecation that we have across all of langchain which will be removed in docs/examples/code in a later version. Some packages are more ahead than others in removing old deprecations in that regard.

MilanCancarevic:
@hntrl I would prefer using `model` just so it's consistent, but since you are guys are already working on it then no changes are needed, thanks for making great framework

PR #8323, opened 2025-06-06
feat: add typed model IDs for chat and embedding models across providers

christian-bromann:
## Description

This PR enhances type safety for model IDs in core OpenAI and Anthropic integrations by leveraging official SDK types. This provides better IntelliSense support and compile-time validation while maintaining full backward compatibility.

### Type Definitions Added
- **`AnthropicMessagesModelId`**: Uses `Anthropic.Model` from the official Anthropic SDK
- **`OpenAIChatModelId`**: Uses `OpenAIClient.ChatModel` from the official OpenAI SDK  
- **`OpenAIEmbeddingModelId`**: Uses `OpenAIClient.EmbeddingModel` from the official OpenAI SDK
- **`OpenAIImageModelId`**: Uses `OpenAIClient.ImageModel` from the official OpenAI SDK

## Benefits

1. **Enhanced Developer Experience**: Full IntelliSense support with autocomplete for all supported models
2. **Type Safety**: Compile-time validation prevents typos in model names
3. **Automatic Updates**: Model types stay current with official SDK releases
4. **Documentation**: All type definitions include links to official documentation
5. **Backward Compatibility**: All types include `(string & NonNullable<unknown>)` for custom models

## Implementation Strategy

- **Official SDK First**: Leverages types directly from `@anthropic-ai/sdk` and `openai` packages
- **Zero Maintenance**: Model lists automatically update with SDK updates
- **Flexible**: Supports custom model names via string union types
- **Non-Breaking**: Existing code continues to work without changes

## Backward Compatibility

✅ **No breaking changes** - all existing code continues to work unchanged  
✅ **Custom model support** - string union allows any custom model name  
✅ **Deprecation handled** - maintains support for deprecated `modelName` parameters

## Testing

- All existing tests pass
- Type checking validates model ID usage across all affected components
- IntelliSense provides comprehensive autocomplete for supported models
- Custom model names work seamlessly alongside typed models


hntrl (changes requested):
Thanks for the PR @christian-bromann!

I'm not too sure if we want to inline all available models in a union type that we maintain. What i'm thinking is it could lead to some false negatives where someone _assumes_ that we don't support a model when a new one releases because it isn't typed (even though we are just passing the request onto the first party sdk).

I'd redirect you to try and see if the provider SDKs we use export a string union with all of their model types and use that instead. That way whenever the provider SDK upgrades, our types also upgrade. (it could be an 'ugly' type that just picks it off invoke params from the SDK)

christian-bromann:
@hntrl good call! I was able to find proper types within the OpenAI and Anthropic SDKs. Shall I revert changes for the other packages? This could indeed be maintenance nightmare.

An alternative would be to use the Open Router Models API ([ref](https://openrouter.ai/api/v1/models)) to generate types from that response into an extra package we consume everywhere. I don't think the added complexity justifies the dev x gain here though.

Thoughts?

hntrl:
1. Unless we use an sdk for other packages that has a model type, then go ahead and revert them.
2. On the open router API, yeah probably not.

christian-bromann:
> 1. then go ahead and revert them.

Done.

hntrl (approved):

PR #8324, opened 2025-06-06
openai[minor]: upgrade sdk and support new built-in tools

hntrl:
Adds support for OpenAI's remote mcp, code interpreter, and image generation built-in tools supported by the responses API. Also bumps the openai sdk to surface these new types back through the library


docs/core_docs/docs/integrations/chat/openai.ipynb
```diff
@@ -817,6 +817,199 @@
...
+        "const response = await llmWithMcp.invoke(\n",
+        "    \"What transport protocols does the 2025-03-26 version of the MCP spec (modelcontextprotocol/modelcontextprotocol) support?\"\n",
+        ");"
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {},
+      "source": [
+        "<details>\n",
```
ccurme, line 928: This doesn't seem to render correctly in the [preview](https://langchainjs-docs-git-hunter-openai-built-ins-langchain.vercel.app/docs/integrations/chat/openai/#built-in-tools).

libs/langchain-openai/src/tests/chat_models_responses.int.test.ts
```diff
@@ -319,6 +318,185 @@ test("Test file search", async () => {
...
+          approve: true,
+        });
+      }
+    }
+  }
+
+  const response2 = await model.invoke(
+    [new HumanMessage({ content: approvals })],
+    {
+      previous_response_id: response.response_metadata.id,
```
ccurme, line 414: I'm curious if passing the full message history works here instead of using `previous_response_id`.

Using the MCP tool from what I see generates multiple reasoning items. In python we're just storing one of them in `additional_kwargs`. So when we attempt to pass back message history we get BadRequestError because we're missing reasoning items. I'm working on a fix, which would likely involve a breaking change to langchain-openai.

ccurme (approved):

libs/langchain-openai/src/tests/chat_models_responses.int.test.ts
hntrl: I see what you mean about MCP emitting multiple reasoning blocks. Looks like we're doing the same thing here of just keeping the latest block when we're iterating over output items. Though I can just pass the output back without pinning it to the response ID and not get any explicit errors (maybe a server-end change from oAI fixed your issue?).

In any case this means we're losing reasoning context when looping in MCP, so I'm interested to hear what you fix entails.

PR #8329, opened 2025-06-08
feat: add service_tier to openai langchain chat

samuelscheit:
OpenAI [introduced](https://platform.openai.com/docs/guides/flex-processing?api-mode=responses) Flex processing which
> provides significantly lower costs for [Chat Completions](https://platform.openai.com/docs/api-reference/chat) or [Responses](https://platform.openai.com/docs/api-reference/responses) requests in exchange for slower response times and occasional resource unavailability. It is ideal for non-production or lower-priority tasks such as model evaluations, data enrichment, or asynchronous workloads.

The python counterpart already has the ability to specify a service_tier but the JS library is missing it, which this PR adds

### Example
```ts
const llm = new ChatOpenAI({
	model: "o4-mini",
	apiKey: process.env.OPENAI_API_KEY,
	service_tier: "flex",
	timeout: 60 * 1000 * 30,
});

llm.invoke("Hi", {
	service_tier: "default", // service_tier can be overridden when invoking llm
});
```

Keep in mind that:
> Flex processing is in beta, and currently only available for [o3](https://platform.openai.com/docs/models/o3) and [o4-mini](https://platform.openai.com/docs/models/o4-mini) models.

### DIFF
--- a/docs/core_docs/docusaurus.config.js
+++ b/docs/core_docs/docusaurus.config.js
@@ -143,7 +143,7 @@ const config = {
     ({
       announcementBar: {
         content:
-          '<strong>We are growing and hiring for multiple roles for LangChain, LangGraph and LangSmith. <a href="https://www.langchain.com/careers" target="_blank" rel="noopener noreferrer"> Join our team!</a></strong>',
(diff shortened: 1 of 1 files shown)


hntrl:
Thanks for the PR @SamuelScheit! I had to make some changes but didn't have write access to your remote. This PR is superseded by #8421

PR #8340, opened 2025-06-10
feat(openai): add file URL preprocessing for message handling

mantrakp04:
Fixes # (issue)


mantrakp04:
(this is a temp fix, a good fix would be using serde but this one works for now)

hntrl:
Thanks for the PR @mantrakp04! (and sorry for the delay!)

Out of curiosity what's the use case here? Is there any reason to not reach for the openai files api?

hntrl:
Going to close this since there hasn't been a response, and I'm not sure we'd want to onshore entire images in base64 and pass those around instead of just deferring to the model provider to handle the logic

Issue #8336, opened 2025-06-10
trimMessages tokenCounter can't process complex messages


Diluka:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain.js documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain.js rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```ts
import 'npm:dotenv/config'; // Load environment variables from .env file

import {
  AIMessage,
  HumanMessage,
  SystemMessage,
  trimMessages,
} from 'npm:@langchain/core/messages';
import { ChatOpenAI } from 'npm:@langchain/openai';

const messages = [
  new SystemMessage({content:[{text:"you're a good assistant, you always respond with a joke.",type:"text"}]}),
  new HumanMessage({content:[{text:"i wonder why it's called langchain",type:"text"}]}),
  new AIMessage(
    {content:[{text:'Well, I guess they thought "WordRope" and "SentenceString" just didn\'t have the same ring to it!',type:"text"}]},
  ),
  new HumanMessage({content:[{text:'and who is harrison chasing anyways',type:"text"}]}),
  new AIMessage(
    {content:[{text:"Hmmm let me think.\n\nWhy, he's probably chasing after the last cup of coffee in the office!",type:"text"}]},
  ),
  new HumanMessage({content:[{text:'what do you call a speechless parrot',type:"text"}]}),
];

const llm = new ChatOpenAI();

const trim = trimMessages({
  maxTokens: 45,
  tokenCounter: async (msgs) => {
    console.log(msgs)
    const tokenCounts = await Promise.all(
      msgs.map((msg) => llm.getNumTokens(msg.content)),
    );
    console.log('Token counts:', tokenCounts);
    return tokenCounts.reduce((sum, count) => sum + count, 0);
  },
  strategy: 'last',
  includeSystem: true,
  allowPartial: false,
  endOn: 'human',
});

const trimmed = await trim.invoke(messages);

console.log(
  trimmed
    .map((x) =>
      JSON.stringify(
        {
          role: x._getType(),
          content: x.content,
        },
        null,
        2,
      ),
    )
    .join('\n\n'),
);
```

### Description

[...]

PR #8341, opened 2025-06-10
fix(core): implement proper token counting for mixed content in getNumTokens

### LINKED ISSUE #8310: getNumTokens only handles string values
### Checked other resources

### Example Code

```ts
    const messages: [BaseMessage] = [
      new HumanMessage({
        content: [
          {
            type: "text",
            text: "What is this image?",
          },
          {
            type: "image_url",
            image_url: {
              url: "https://www.w3.org/MarkUp/Test/xhtml-print/20050519/tests/jpeg420exif.jpg",
            },
            detail: "high",
          },
        ],
      }),
    ];

    const inputTokens = await model.getNumTokensFromMessages(messages);
```

_No response_

### Description

It is a TODO comment in `langchain-core/src/language_models/base.ts`

```ts
  async getNumTokens(content: MessageContent) {
    // TODO: Figure out correct value.
    if (typeof content !== "string") {
      return 0;
    }
```

When calling `getNumTokensFromMessages` and `message.content` is not a string, for example passing `image_url` as an array, the returned value for `textCount` here

```ts
const textCount = await this.getNumTokens(message.content);
```

is 0. So the tokens in the text "What is this image?" is not counted for.

### System Info

langchain/core: 0.3.57

christian-bromann:
## Description

This PR fixes a critical issue in the `BaseLanguageModel.getNumTokens()` method where it would return 0 tokens for any non-string `MessageContent`, effectively ignoring text content in mixed-content messages.

### Problem
Previously, when calling `getNumTokensFromMessages()` with messages containing mixed content (e.g., text + image_url arrays), the text portions were not being counted, leading to inaccurate token usage calculations. The method had a TODO comment and simply returned 0 for any array content:

```typescript
// TODO: Figure out correct value.
if (typeof content !== "string") {
  return 0;
}
```

### Solution
Implemented proper text extraction logic that:
- **Maintains backward compatibility**: String content works exactly as before
- **Extracts text from arrays**: Processes `MessageContentComplex[]` to find and concatenate all text blocks
- **Ignores non-text content**: Image URLs, audio, and other media content are properly ignored for token counting
- **Handles edge cases**: Empty arrays, image-only content, and multiple text blocks

### Example
```javascript
const messages = [
  new HumanMessage({
    content: [
      {
        type: "text",
        text: "What is this image?", // This text is now properly counted
      },
      {
        type: "image_url",
        image_url: {
          url: "https://example.com/image.jpg",
        },
      },
    ],
  }),
];

// Before: getNumTokensFromMessages would undercount due to 0 tokens from getNumTokens
// After: Correctly counts tokens for "What is this image?"
```

### Changes Made
- **Core fix**: Replaced TODO implementation in `getNumTokens()` with proper text extraction
- **Comprehensive tests**: Added test suite covering mixed content, edge cases, and compatibility
- **Test restructuring**: Improved test organization with `describe/it` blocks for better clarity

### Testing
- ✅ All existing tests pass (backward compatibility maintained)
- ✅ New tests cover mixed content scenarios
- ✅ Edge cases tested (empty arrays, image-only content, multiple text blocks)
- ✅ Integration with `getNumTokensFromMessages` verified

This fix ensures accurate token counting for modern chat applications that use rich message content with text, images, and other media types.

Fixes #8310

### DIFF
--- a/langchain-core/src/language_models/base.ts
+++ b/langchain-core/src/language_models/base.ts
@@ -412,12 +412,23 @@ export abstract class BaseLanguageModel<
   private _encoding?: Tiktoken;
 
   async getNumTokens(content: MessageContent) {
-    // TODO: Figure out correct value.
-    if (typeof content !== "string") {
-      return 0;
+    // Extract text content from MessageContent
+    let textContent: string;
+    if (typeof content === "string") {
+      textContent = content;
+    } else {
+      // Content is an array of MessageContentComplex
+      textContent = content
+        .map((item) => {
+          if (typeof item === "string") return item;
+          if (item.type === "text" && "text" in item) return item.text;
+          return "";
+        })
+        .join("");
     }
+
     // fallback to approximate calculation if tiktoken is not available
-    let numTokens = Math.ceil(content.length / 4);
+    let numTokens = Math.ceil(textContent.length / 4);
 
     if (!this._encoding) {
       try {
@@ -436,7 +447,7 @@ export abstract class BaseLanguageModel<
 
     if (this._encoding) {
       try {
-        numTokens = this._encoding.encode(content).length;
+        numTokens = this._encoding.encode(textContent).length;
       } catch (error) {
         console.warn(
           "Failed to calculate number of tokens, falling back to approximate count",
--- a/langchain-core/src/language_models/tests/count_tokens.test.ts
+++ b/langchain-core/src/language_models/tests/count_tokens.test.ts
@@ -1,33 +1,104 @@
-import { test, expect } from "@jest/globals";
+import { describe, it, expect } from "@jest/globals";
 import { calculateMaxTokens, getModelContextSize } from "../base.js";
+import { MessageContent } from "../../messages/base.js";
+import { FakeLLM } from "../../utils/testing/index.js";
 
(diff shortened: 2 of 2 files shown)


hntrl (changes requested):
Better than what we had before but I don't know if this is "end-state" token counting (for instance in python we consider other things like function calls apart of the token count, there's probably other ones we should be considering). Probably at a minimum we re-add the comment to signal a more appropriate solution is needed. It would be interesting to compare different message content instances with gpt-2 tokenizer and this fallback method.

I'll also call out I don't think the blast radius on this is big (I think) since we only use this in the trimMessages utility

https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/messages/utils.py#L86

christian-bromann:
Thanks for reviewing!

> I'll also call out I don't think the blast radius on this is big (I think) since we only use this in the trimMessages utility

There have been 2 reported instances of this failing for users, see #8310 and #8336. I agree with you, this is definitely not perfect but addresses the issues reported. I've added a comment to revisit this section at a later time. Let me know if there is anything else I can do.

hntrl (approved):

hntrl:
thanks @christian-bromann!

PR #8350, opened 2025-06-12
feat: use run tree post/patch

lc-arjun:
Fixes # (issue)

### DIFF
--- a/langchain-core/src/tracers/tracer_langchain.ts
+++ b/langchain-core/src/tracers/tracer_langchain.ts
@@ -87,27 +87,41 @@ export class LangChainTracer
   protected async persistRun(_run: Run): Promise<void> {}
 
   async onRunCreate(run: Run): Promise<void> {
-    const persistedRun: RunCreate2 = await this._convertToCreate(
-      run,
-      this.exampleId
-    );
-    await this.client.createRun(persistedRun);
+    try {
+      const runTree = this.convertToRunTree(run.id);
+      await runTree?.postRun();
+    } catch (error) {
+      console.error(`Error in onRunCreate for run ${run.id}:`, error);
+      // Fallback to original behavior if RunTree approach fails
+      const persistedRun: RunCreate2 = await this._convertToCreate(
+        run,
+        this.exampleId
+      );
+      await this.client.createRun(persistedRun);
+    }
   }
 
   async onRunUpdate(run: Run): Promise<void> {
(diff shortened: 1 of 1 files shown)


langchain-core/src/tracers/tracer_langchain.ts
```diff
@@ -87,27 +87,41 @@ export class LangChainTracer
...
-      run,
-      this.exampleId
-    );
-    await this.client.createRun(persistedRun);
+    try {
+      const runTree = this.convertToRunTree(run.id);
+      await runTree?.postRun();
+    } catch (error) {
+      console.error(`Error in onRunCreate for run ${run.id}:`, error);
+      // Fallback to original behavior if RunTree approach fails
```
jacoblee93: Why the fallback?

hntrl (changes requested):
lgtm, but it's probably important that we bump the langsmith dependency in package.json when those changes land before merging this

langchain-core/src/tracers/tracer_langchain.ts
```diff
@@ -132,62 +108,30 @@ export class LangChainTracer
...
       }
     }
 
     this.client = runTree.client ?? this.client;
+    this.replicas = runTree.replicas ?? this.replicas;
     this.projectName = runTree.project_name ?? this.projectName;
     this.exampleId = runTree.reference_example_id ?? this.exampleId;
   }
 
-  convertToRunTree(id: string): RunTree | undefined {
```
jacoblee93, line 147: Removing this is technically breaking but would be unusual if someone were extending a our tracer

langchain-core/src/tracers/base.ts
```diff
@@ -32,6 +34,36 @@ export interface Run extends BaseRun {
...
+// TODO: Remove and just use base LangSmith Run type
+const convertRunTreeToRun = (runTree?: RunTree): Run | undefined => {
+  if (!runTree) {
+    return undefined;
+  }
+  runTree.events = runTree.events ?? [];
+  runTree.child_runs = runTree.child_runs ?? [];
+  // Important that we return the raw run tree object since the reference
+  // is mutated in other places.
+  return runTree as unknown as Run;
```
jacoblee93, line 52: Sadly because of the difference in `events` typing between LangSmith/LangChain we need this horrible cast

Alternative is that we just remove the base LangChain typing but that's disruptive to subclassers. I don't think `events` is super widely used anyway.

PR #8351, opened 2025-06-12
fix(community): allow any chars in delimited identifiers in hanavector

alumni:
Allows any identifier to be used in SAP HANA vector store, including the ones that contain non-alphanumeric characters.

This would allow the users to pass e.g.: `sap.hdb.container/embeddings_table` as `tableName`, and the special characters won't be stripped away.

```ts
const hanaDb = new HanaDB(embeddingClient, {
  connection,
  tableName: 'sap.hdb.container/embeddings_table',
});
```

### DIFF
--- a/libs/langchain-community/src/vectorstores/hanavector.ts
+++ b/libs/langchain-community/src/vectorstores/hanavector.ts
@@ -133,23 +133,15 @@ export class HanaDB extends VectorStore {
   constructor(embeddings: EmbeddingsInterface, args: HanaDBArgs) {
     super(embeddings, args);
     this.distanceStrategy = args.distanceStrategy || defaultDistanceStrategy;
-    this.tableName = HanaDB.sanitizeName(args.tableName || defaultTableName);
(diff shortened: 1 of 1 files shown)


christian-bromann:
@alumni thanks for the contribution. Are you still interested in getting this landed into Langchain? It generally looks good to me and is only missing some unit tests. Let me know how you would like to move forward here?

hntrl:
rebased in #8913

alumni:
@christian-bromann @hntrl sorry, I'm not that much online lately, so I'm slow to reply. Thanks for taking over the PR.

I'll likely extract the HANA vector store (or the TypeORM one) into a separate package, currently the community package is adding some unwanted dependencies to our large monorepo, and it seems it's simpler to take only the parts that we need.

christian-bromann:
@alumni no worries and yes, that is the direction we are suggesting to folks now. We are working on finding an ideal way to highlight these community contributions, so please let us know once you have something out.

PR #8354, opened 2025-06-12
Add notIn filter support for pgvector

yilinjuang:
Add support for `notIn` filtering to enable negative filters


hntrl:
Thanks for the change @yilinjuang!

PR #8365, opened 2025-06-16
minor(google-gauth): Upgrade google-auth-library to latest major version (v10)

lbennett-stacki:
Upgrades `google-auth-library` dep to latest major version (v10.1.0).

### DIFF
--- a/libs/langchain-google-gauth/package.json
+++ b/libs/langchain-google-gauth/package.json
@@ -36,7 +36,7 @@
   "license": "MIT",
   "dependencies": {
     "@langchain/google-common": "^0.2.10",
-    "google-auth-library": "^9.15.1"
+    "google-auth-library": "^10.1.0"
   },
   "peerDependencies": {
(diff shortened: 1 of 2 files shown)


afirstenberg:
#lgtm

hntrl:
@lbennett-stacki Looks like CI is failing after the upgrade. Do you mind taking a look?

afirstenberg:
To be clear, my lgtm was because tests ran without problems.

Looking at the code, `langchain-google-gauth/src/auth.ts` expects Gaxios handling, which is what "google-auth-library" 9 uses. With version 10, they're moving towards `fetch()` semantics, which means some of the types are broken. 10 is still Gaxios compatible - but apparently not type compatible.

The good news is that `langchain-google-webauth` uses `fetch()` as well and builds off a base class for it. So it may be easy to use this. Investigating further.

See https://github.com/googleapis/google-auth-library-nodejs/pull/1938

PR #8373, opened 2025-06-16
fix(core): Await default client batches when awaiting all callbacks, set default traced project from LangSmith

jacoblee93:
CC @lc-arjun


langchain-core/src/singletons/callbacks.ts
```diff
@@ -60,6 +61,10 @@ export async function consumeCallback<T>(
  * Waits for all promises in the queue to resolve. If the queue is
  * undefined, it immediately resolves a promise.
  */
-export function awaitAllCallbacks(): Promise<void> {
-  return typeof queue !== "undefined" ? queue.onIdle() : Promise.resolve();
+export async function awaitAllCallbacks(): Promise<void> {
+  const defaultClient = getDefaultLangChainClientSingleton();
+  await Promise.all([
```
dqbd: Should it be `allSettled`?

jacoblee93: Yeah probably for safety, good call

jacoblee93: Thanks!

PR #8379, opened 2025-06-17
fix(core): add custom key encoder and deprecate insecureHash

manekinekko:
This PR deprecates the `insecureHash` and introduces a support for Custom Key Encoders for Caching and Indexing.

Note: The `insecureHash` function is currently based on SHA-1 algorithm, which may lead to security issues as discussed with @sinedied and @jacoblee93

### DIFF
--- a/langchain-core/src/caches/tests/in_memory_cache.test.ts
+++ b/langchain-core/src/caches/tests/in_memory_cache.test.ts
@@ -41,12 +41,7 @@ test("InMemoryCache works with complex message types", async () => {
 });
 
 test("InMemoryCache handles default key encoder", async () => {
-  // @ts-expect-error __printInsecureHashWarning is not defined in globalThis
-  globalThis.__printInsecureHashWarning = false;
   const cache = new InMemoryCache();
-  const consoleWarnSpy = jest
-    .spyOn(console, "warn")
-    .mockImplementation(() => {});
 
   await cache.update("prompt1", "key1", [
     {
@@ -58,20 +53,10 @@ test("InMemoryCache handles default key encoder", async () => {
   const result = await cache.lookup("prompt1", "key1");
 
   expect(result).toBeDefined();
-  if (!result) {
-    return;
-  }
-  expect(consoleWarnSpy).toHaveBeenCalled();
-  consoleWarnSpy.mockRestore();
 });
 
 test("InMemoryCache handles custom key encoder", async () => {
-  // @ts-expect-error __printInsecureHashWarning is not defined in globalThis
-  globalThis.__printInsecureHashWarning = false;
   const cache = new InMemoryCache();
-  const consoleWarnSpy = jest
-    .spyOn(console, "warn")
-    .mockImplementation(() => {});
 
   // use fancy hashing algorithm to encode the key :)
   cache.makeDefaultKeyEncoder((prompt, key) => `${prompt}###${key}`);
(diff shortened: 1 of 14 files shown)


langchain-core/src/utils/js-sha1/hash.ts
```diff
@@ -412,6 +412,19 @@ Sha1.prototype.arrayBuffer = function () {
   return buffer;
 };
 
+/**
+ * @deprecated Use `makeDefaultKeyEncoder()` to create a custom key encoder.
+ * This function will be removed in a future version.
+ */
 export const insecureHash = (message) => {
+  console.warn(
```
jacoblee93: Let's just log this once on initial call to avoid flooding the console and add some example code

langchain-core/src/caches/tests/in_memory_cache.test.ts
```diff
@@ -39,3 +39,52 @@ test("InMemoryCache works with complex message types", async () => {
...
+  consoleWarnSpy.mockRestore();
+});
+
+
+test("InMemoryCache handles custom key encoder", async () => {
+  const cache = new InMemoryCache();
+  const consoleWarnSpy = jest.spyOn(console, "warn").mockImplementation(() => {});
+  
+  // use fancy hashing algorithm to encode the key :)
+  cache.makeDefaultKeyEncoder((prompt, key) => `${prompt}###${key}`);
```
jacoblee93, line 62: Lol!

langchain-core/src/utils/js-sha1/hash.ts
manekinekko: Fixed.

manekinekko: > add some example code

Can you elaborate on which example code you'd like to be added?

manekinekko: @jacoblee93 can we maybe link to some docs?

langchain-core/src/utils/js-sha1/hash.ts
```diff
@@ -412,6 +412,22 @@ Sha1.prototype.arrayBuffer = function () {
   return buffer;
 };
 
+/**
+ * @deprecated Use `makeDefaultKeyEncoder()` to create a custom key encoder.
+ * This function will be removed in a future version.
+ */
 export const insecureHash = (message) => {
+  if (!globalThis.__printInsecureHashWarning) {
```
jacoblee93: Any reason not to just set a variable out of function scope? Just to avoid polluting global scope

```ts
let hasLoggedWarning = false;

export const insecureHash = (message) => {
  if (!hasLoggedWarning) {
    hasLoggedWarning = true;
  }
```

manekinekko: oh because I am using that in tests, but I can drop the test that checks for the printed warning.

manekinekko: Fixed.

langchain-core/src/utils/js-sha1/hash.ts
jacoblee93: Ah right... this is used in multiple places

@hntrl maybe let's make a small docs page with agnostic info here?

manekinekko:
Thank you @hntrl and @jacoblee93 for providing improvements to this PR. I think sha256 is agreat start. It's defintely way stronger hash fn than sha1. 

The PR LGTM.

manekinekko:
Thank you @jacoblee93 and @hntrl for taking the time to help reviewing this work. We (Microsoft) appreciate it.

PR #8380, opened 2025-06-17
Update langchain-openai package.json to unpin specifc zod version

tobymac-siro:
[This PR](https://github.com/langchain-ai/langchainjs/pull/8273) updated the zod versions for all langchain libs to `^3.25.32`, but only for `langchain-openai`, it appears that it got pinned to the specific version `3.25.32`.  There is no explanation so it might just be a mistake.  This unpins the specific version of zod.


hntrl:
This was definitely made in error. @tobymac-siro ty for flagging!

PR #8383, opened 2025-06-17
minor(google): Upgrade Upgrade google-auth-library to latest major version (v10)

afirstenberg:
This is #8365 with code changes that fix typing issues and does some code cleanup in the process.

(I don't have permissions to change the original PR.)

cc: @lbennett-stacki @hntrl

### DIFF
--- a/libs/langchain-google-common/src/utils/stream.ts
+++ b/libs/langchain-google-common/src/utils/stream.ts
@@ -307,18 +307,18 @@ export class ReadableAbstractStream implements AbstractStream {
     return this.baseStream.streamDone;
   }
 
-  async run(body: ReadableStream) {
-    const reader = body.getReader();
-    let isDone = false;
-    while (!isDone) {
-      const { value, done } = await reader.read();
-      if (!done) {
+  // Should be a ReadableStream, but the Gaxios Readable stream isn't.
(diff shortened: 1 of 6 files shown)


libs/langchain-google-common/src/utils/stream.ts
```diff
@@ -307,18 +307,18 @@ export class ReadableAbstractStream implements AbstractStream {
...
-    let isDone = false;
-    while (!isDone) {
-      const { value, done } = await reader.read();
-      if (!done) {
+  // Should be a ReadableStream, but the Gaxios Readable stream isn't.
+  // But both should support async iterators, so make sure of that.
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  async run(body: any) {
+    if (typeof body[Symbol.asyncIterator] === "function") {
+      for await (const value of body) {
```
hntrl, line 315: Is it possible to just do something like this?
```suggestion
  async run(body: AsyncIterable<ArrayBuffer>) {
    for await (const value of body) {
```

afirstenberg: *Probably*. But I wanted to put the check there just in case run() gets something that... can't.

PR #8388, opened 2025-06-18
Upgrade zod dependency

brichet:
The current version of `zod` (`3.25.51`) has an issue when building with `"skipLibCheck": false` is specified in `tsconfig.json`.

See https://github.com/colinhacks/zod/issues/4545 and https://github.com/colinhacks/zod/issues/4624 for more constext.

This PR fix it by upgrading the version of `zod`.


brichet:
Right, this change would not impact our dependencies.
I just saw that the culprit was a pinned version in `@langchain/openai`, which has already been addressed in https://github.com/langchain-ai/langchainjs/pull/8380.
Sorry for the noise :smile:

PR #8396, opened 2025-06-19
refactor: update Ollama import and add fetch option to models

Sergeant61:
Adds the setting where we can change the fetch for the ollama.


hntrl:
Thanks for the PR @Sergeant61!

PR #8415, opened 2025-06-23
chore: Add tool annotations to metadata

jonathanfds:
Tools annotations from the MCP Server are not set in the `DynamicStructuredTool` and are completely ignored. 
This change will set the tools annotations within the `metadata` property.


hntrl:
This makes sense. Thanks for the PR @jonathanfds!

PR #8416, opened 2025-06-24
Hunter/net mocks

hntrl:
> [!WARNING]
> This PR isn't intended to be merged yet. There are still some rough edges, but it's being put up now to start collecting feedback.

This PR introduces a new network mocking and recording utility for langchain thats inspired by popular libraries like Python's VCR.py and MSW. This utility, internally called `net-mocks`, allows us to retain real API interactions as HAR files and replay them during test runs.

The core of the utility is built using two main components:
- [`@mswjs/interceptors`](https://github.com/mswjs/interceptors) is a well known low-level library for observing and mutating the network layer in different JS environments, and we're using it here to do exactly that.
- The actual retention of network traffic happens using a purpose-built recording interface based on the [HAR spec](http://www.softwareishard.com/blog/har-12-spec/) (or HTTP Archive Format).

## ⚠️ The problem

The way that we test our integrations today is in two parts: (1) by unit testing internal functions that model our assumptions about the in/out formats from a provider, and (2) using a separate integrations suite that actually calls provider APIs. This approach has a few drawbacks:

- **Running our full test suite gets expensive.** Every time our integration suite runs, we make real API calls to dozens of providers, which costs real money and slows down CI. Because of this, we only run our unit tests as apart of our required CI path.
  - **We only test a subset of interactions.** To keep costs down, we're selective about which integrations we test, meaning many features and edge cases are never validated against the real APIs.
  - **Regressions are easy to miss.** Small changes designed for a specific issue can have unintended side-effects on other integrations, which our unit tests are unable to catch.
- **We're blind to upstream API changes.** If a service we integrate with changes its API ever-so-slightly, our CI passes because we don't run integrations tests as default. We only find out about regressions either when someone reports an issue or we re-run the integrations suite.
- **It's hard to manually maintain network mocks.** Whenever we do mock things out, it involves a developer manually copy-pasting a response into a test file. These mocks are brittle, and they become out-of-date as soon as an API changes.

## ✅ The solution

This new `net-mocks` utility solves these problems by providing a standard testing utility. Here's how it works:

1.  **Record:** When running tests for the first time, `net-mocks` intercepts outgoing HTTP requests and records them, along with the corresponding responses, into a `.har` file. These files are stored in a `__data__` directory alongside the tests.
2.  **Replay:** On subsequent test runs, `net-mocks` intercepts the same HTTP requests but instead of hitting the actual API, it finds a matching request in the `.har` file and returns the recorded response. This makes tests deterministic and runnable offline.
3. **Refresh:** If the underlying API changes, or our tests are slightly modified, then the matching function between the stored and actual request becomes falsy and data is automatically re-fetched (or will reject the test if configured)
    - Mainstream model providers that use stainless as their sdk backing (openAI, anthropic) typically pin a `sdk-version` or `api-version` header that gets sent with requests. Once those change (like by an sdk upgrade), that means that the network activity stored with a test will automatically be refetched. This makes it easy to mocked tests up-to-date with the latest API changes.
    - This tool also involves the concept of "stale" requests. If a request becomes older than a pre-determined interval, then we either refetch the request or reject the test based on the configuration. This just acts as an extra sanity-check to protect from sneaky API changes.

### Why not `{insert-library}`?

There are a couple of off-the-shelf solutions (like core msw/ nock) that solve similar problems to this, but I opted to go for a bespoke solution since none of those aforementioned libraries ticked all of these boxes:

- **They weren't runtime agnostic.** While we don't run tests in different environments today, that would be a nice future thing to have, so I didn't want to close off that option with the introduction of this tool.
- **They were purpose-built for a tangential use-case.** The different tools I assessed are much better suited for a different type of testing. Some of these tools can be co-opted to fit langchain's needs, but they often required brittle interactions with internal methods to get it to work right. 

## 📚 Usage guide

This utility is meant to have an easy API so that it can work as a "drop-in" replacement for our existing test suites.

```ts
import { describe, it, expect, beforeAll } from "vitest";
import { net } from "@langchain/net-mocks";
import { ChatAnthropic } from "../chat_models";

beforeAll(() =>
  net.setupVitest(...)
});

describe("ChatAnthropic", () => {
  it("should work as expected", async () => {
    // This binds the network listener to the current test context
    await net.vcr();

      model: "claude-3-5-sonnet-20240620",
    });
    const message = new HumanMessage("do you know the muffin man");
    const result = await model.invoke([message]);

    expect(result.content).toBeDefined();
  });
});
```

### Options

Options are defined by collapsing a set of default options, the options provided in `net.setupVitest`, and the options provided in `net.vcr`. That resulting config object determines a couple of behaviors:

<details>
<summary>Options config</summary>

```ts
/**
 * Options for configuring network mocking and recording behavior.
 */
type NetOptions = {
  /**
   * Maximum age (in milliseconds) for cached network entries before considered stale.
   * Defaults to 30 days
   */
  maxAge: number;
  /**
   * Strategy for handling stale cache entries:
   * - "reject": Reject the request if the entry is stale.
   * - "warn": Warn but allow the request.
   * - "refetch": Refetch the request from the network.
   * - "ignore": Ignore staleness and use the entry.
   * @default "warn"
   */
  stale: "reject" | "warn" | "refetch" | "ignore";
  /**
   * Strategy for handling unmatched requests:
   * - "reject": Reject unmatched requests.
   * - "warn": Warn but allow unmatched requests.
   * - "fetch": Fetch the request from the network.
   * @default "warn"
   */
  noMatch: "reject" | "warn" | "fetch";
  /**
   * Whether to mimick the timings of the original request.
   * @default true
   */
  useTimings: boolean;
  /** Output file path for saving the archive or mock data. */
  out?: string;
  /** List of header or body keys to redact from the archive for privacy/security. */
  redactedKeys: string[];
};
```
</details>

### Network timings

As apart of the HAR spec, complete archives come with fields that describe the timing of a request/response pair. When `useTimings` is set to `true` in vcr options, the resulting response object returned internally is delayed according to those timings to simulate real network delay.

Additionally when a response comes back with `Content-Type: text/event-stream` (a common format for streaming responses), we store the timings of each chunk as apart of the stored response body. On followup test runs (where `useTimings: true`), we replay the chunks as if we were streaming actual data over the network. This just better helps us model real world environments and detect race conditions if they exist.

### Using dev tools

HAR is a lesser known web standard that has its own place in modern web browser devtools, and its the format that we're using to retain network activity for later runs. This means that we can inspect all network transactions that occured in a test directly in the browser.

https://github.com/user-attachments/assets/36de46b7-d9d6-408a-9d06-4185dd990020


internal/net-mocks/src/utils.ts
```diff
@@ -0,0 +1,69 @@
...
+ * Performs a deep equality check between two values, with special handling for stringified JSON.
+ *
+ * If either value is a string that can be parsed as JSON, it will be parsed before comparison.
+ * This allows for deep equality checks between objects and their JSON string representations.
+ *
+ * @param {unknown} a - The first value to compare. Can be any type.
+ * @param {unknown} b - The second value to compare. Can be any type.
+ * @returns {boolean} True if the values are deeply equal (after normalization), false otherwise.
+ */
+export function deepEqual(a: unknown, b: unknown): boolean {
```
dqbd, line 58: _nit_: I wonder if we need to handle circular values?

internal/net-mocks/src/mock.ts
```diff
@@ -0,0 +1,313 @@
...
+
+    this.hooks?.cleanup(async () => {
+      this.interceptor?.removeAllListeners("request");
+      await Promise.all(this._archivePromises);
+      await store.save(source, archive);
+    });
+  }
+
+  async setupVitest(options?: Partial<NetMockOptions>) {
+    const { onTestFinished, expect } = await import("vitest");
```
dqbd, line 329: _nit_: If we expect to support `jest` down the line, we might want to instead break down into different entrypoints? ie `@langchain/net-mocks/vitest` etc.

internal/net-mocks/src/utils.ts
hntrl: I imagine no since the only thing we're comparing is serial outputs like HAR bodies with Request bodies

(the deep-equal package might do circular values anyway, I didn't dig that deep into it)

internal/net-mocks/src/request.ts
```diff
@@ -0,0 +1,452 @@
...
+): Promise<HARRequest> {
+  const requestUrl = new URL(request.url);
+  const queryParams = new URLSearchParams(requestUrl.search);
+  const queryString: HARQueryString[] = [];
+  for (const [key, value] of queryParams.entries()) {
+    queryString.push({ name: key, value });
+  }
+  let postData: HARPostData | undefined;
+  if (request.body && request.method !== "GET") {
+    const buffer = await consumeBodyStream(request.body);
```
dqbd: Should we mark the consumed body stream for text vs. binary data?

internal/net-mocks/README.md
```diff
@@ -0,0 +1,107 @@
+# @langchain/net-mocks
+
+This is an internal utility used within LangChain to record & mock network activity for use in tests. Here's how it works:
+
+1. **Record:** When running tests for the first time, `net-mocks` intercepts outgoing HTTP requests and records them, along with the corresponding responses, into a `.har` file. These files are stored in a `__data__` directory (by default) alongside the tests.
```
dqbd: _nit_: should we follow the `__snapshots__` convention instead?

internal/net-mocks/src/mock.ts
hntrl: hmm yeah. Do we want that to be the expectation? (i.e. using net mocks is the excuse to migrate to vitest 🤑)

internal/net-mocks/package.json
```diff
@@ -0,0 +1,36 @@
...
+  },
+  "type": "module",
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "@mswjs/interceptors": "^0.39.2",
+    "cookie": "^1.0.2",
+    "deep-equal": "^2.2.3",
+    "vitest": "^3.2.4"
```
dqbd, line 24: Worth making `vitest` a peer dep?

internal/net-mocks/src/request.ts
hntrl: HAR only accepts string contents (i.e. for text). It does support base64 encoding though (will work in).

internal/net-mocks/example/vitest.setup.ts
```diff
@@ -0,0 +1,9 @@
+import { beforeAll } from "vitest";
+import { net } from "../src";
+
+beforeAll(() =>
+  net.setupVitest({
+    redactedKeys: ["x-api-key"],
```
dqbd: _question_: I wonder if we should allow specifying values rather than keys to be omitted? Also I wonder if we should auto-include certain keys or values such as OAI keys etc.

hntrl: yeah I like this more (push protection may have been the only thing saving me from making the pinnacle rookie mistake yesterday)

internal/net-mocks/README.md
hntrl: ~~I called it `__data__` since that's what we call it in the few other places where we have mocking in langchain today (google packages are probably the best example). I won't die on this hill if you think this is better though~~

nvm I like snapshots more

internal/net-mocks/example/vitest.setup.ts
hntrl: Screenshot 2025-06-26 at 9 53 50 AM

also some headers are not deterministic

Issue #8394, opened 2025-06-19
🐛 Bug: AIMessageChunk tool_call_chunks merging fails for fragmented tool call arguments


kinka:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain.js documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain.js rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

Relevant code in @langchain/core/dist/messages/ai.js:
```typescript
for (const toolCallChunk of fields.tool_call_chunks) {
    let parsedArgs = {};
    try {
        parsedArgs = parsePartialJson(toolCallChunk.args || "{}");
        // ... (fails for partial JSON)
        toolCalls.push({ ... });
    } catch (e) {
        invalidToolCalls.push({ ... });
    }
}
```

// Robustly merge tool_call_chunks into a single tool_call
```typescript
function combineMessageChunks(chunks) {
  if (!chunks || chunks.length === 0) return [];
  // Skip leading chunks with empty name (garbage fragments)
  while (!chunks[0].name) chunks.shift();
  const { name, id } = chunks[0];
  let argsStr = '';
  for (const c of chunks) argsStr += c.args || '';
  let args;
  try { args = JSON.parse(argsStr); } catch { args = argsStr; }
  return [{ name, id, args, type: 'tool_call' }];
}
```

### Summary

When using streaming tool calls with some LLMs (e.g., Qwen, some OpenAI/self-hosted models), the model outputs tool_call_chunks where the `args` field is split into many small fragments. LangChain's current merging logic tries to parse each chunk individually, which fails unless each chunk is a complete JSON object. This results in `tool_calls` being empty or incomplete, and only the last chunk (if lucky) is parsed.

**Claude 4 and some Anthropic models do not have this problem because they output each tool_call_chunk as a complete JSON.**

---

### Expected Behavior

- LangChain should **collect and concatenate all tool_call_chunks belonging to the same tool call (by id/name)**, then parse the full JSON string after all chunks are received.
- This should work regardless of how fragmented the model's output is.

---

### Actual Behavior

- LangChain tries to parse each chunk's `args` field individually.
- For fragmented outputs, this always fails except for the last chunk (if it's complete).
[...]

PR #8419, opened 2025-06-26
fix: merging tool_call_chunks grouped by id (issue #8394)

tnoori98:
Fixes: #8394
This pull request handles the merging logic for tool_call_chunks in AIMessageChunk. Right now LangChain attempts to parse each args field in separately, which fails when models like OpenAI or Qwen return fragmented JSON with multiple chunks.

What I did:
- Group tool_call_chunks by id
- Concatenate the args fields per group
- Parsing the full JSON string after the chunks are merged

I implemented a new unit test under src/tests/ai.test.ts to verify the merging behavior


hntrl:
Thanks for the PR @tnoori98! I reworked a few formatting and collapsed tests, but this is otherwise a very welcome change!

tnoori98:
I am glad it helped :) @hntrl

Issue #8408, opened 2025-06-22
Using inferred model to Bedrock chat does not support APAC


arvsxz05:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain.js documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain.js rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```typescript
const chatModel = new BedrockChat({
    model: 'apac.anthropic.claude-sonnet-4-20250514-v1:0',
    region: 'ap-southeast-1',
    maxTokens: 1024,
    temperature: 0,
});
```

### Error Message and Stack Trace (if applicable)

Unknown model provider: 'apac', only these are supported: ai21,anthropic,amazon,cohere,meta,mistral,deepseek

### Description

I wanted to use a Bedrock [inferred model](https://docs.aws.amazon.com/bedrock/latest/userguide/inference-profiles-support.html) from Anthropic in APAC region:
```
apac.anthropic.claude-sonnet-4-20250514-v1:0
```

But when I pass it as a model to `BedrockChat`, I encounter an error:
```
Unknown model provider: 'apac', only these are supported: ai21,anthropic,amazon,cohere,meta,mistral,deepseek
```

Based in the [codes](https://github.com/langchain-ai/langchainjs/blob/ac42886f52a0bb851a281653a77baee019725ea9/libs/langchain-community/src/chat_models/bedrock/web.ts#L55), it seems that `apac` is not part of the expected AWS regions for inference.

Related:
https://github.com/langchain-ai/langchain-aws/issues/291

### System Info

Platform: Linux
Node Version: 18
Yarn Version: 1.22.19

jwasnoggin:
I'm getting a similar error now with "au" as the region: "Unknown model provider: 'au', only these are supported: ai21,anthropic,amazon,cohere,meta,mistral,deepseek".

Can this also be added?

pawel-twardziak:
hi @jwasnoggin the fix is cominig from there https://github.com/langchain-ai/langchainjs/pull/10399

PR #8425, opened 2025-06-27
fix: add missing APAC region prefix in Bedrock models

christian-bromann:
## Add support for APAC region prefix in Bedrock models

- Add "apac" to AWS_REGIONS list in both chat models and LLMs
- Add comprehensive test suite for model provider detection
- Test various regional model prefixes (eu, apac, us-gov, us)
- Ensure proper handling of different model ID formats and providers
- Add validation for unsupported custom model ARNs

This enables support for APAC-specific Bedrock model IDs like: `apac.anthropic.claude-3-5-sonnet-20240620-v1:0`

This change resolves the issue where users in APAC regions couldn't use region-specific Bedrock model identifiers. Previously, attempting to use models with the `apac.` prefix would result in an "Unknown model provider" error because the APAC region prefix wasn't recognized in the AWS_REGIONS configuration.

The fix includes:
- Adding "apac" to the supported AWS regions in both `BedrockChat` and `Bedrock` LLM classes
- Comprehensive test coverage for all regional model prefixes including edge cases
- Validation that ensures proper error handling for unsupported custom model ARNs

Fixes [#8408](https://github.com/langchain-ai/langchainjs/issues/8408)

### DIFF
--- a/libs/langchain-community/src/llms/bedrock/web.ts
+++ b/libs/langchain-community/src/llms/bedrock/web.ts
@@ -28,6 +28,7 @@ const AWS_REGIONS = [
   "ap",
   "af",
   "us-gov",
+  "apac",
 ];
 
 const ALLOWED_MODEL_PROVIDERS = [
--- a/libs/langchain-community/src/llms/bedrock/web.ts
+++ b/libs/langchain-community/src/llms/bedrock/web.ts
@@ -17,6 +17,9 @@ import {
 } from "../../utils/bedrock/index.js";
 import type { SerializedFields } from "../../load/map_keys.js";
 
+/**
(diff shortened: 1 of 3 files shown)


libs/langchain-community/src/llms/bedrock/web.ts
```diff
@@ -28,6 +31,7 @@ const AWS_REGIONS = [
   "ap",
   "af",
   "us-gov",
+  "apac",
```
hntrl, line 35: Could we also add mexico (`mx`)?
```suggestion
  "apac",
  "mx",
```

christian-bromann: Done!

hntrl (approved):

PR #8426, opened 2025-06-27
use bm25 sorter function

AllenFang:
self explanatory, we could just leverage the sort function provided from bm25 function


AllenFang:
never mind, I just figured out why we didn't use the sort param from bm25 function.

PR #8427, opened 2025-06-27
improve bm25 sorting function

AllenFang:
Goal:
* make good use of the `sorter` method signature of `BM25`
* refactor the `BM25` method to solve the problem if sorter param assign, it's very difficult to map the result to original data. 
  * Clearly define the input and output type of `BM25` method
* now the sorter can not only sort by score but also anything in the `docs`

### DIFF
--- a/.changeset/swift-brooms-argue.md
+++ b/.changeset/swift-brooms-argue.md
@@ -1,0 +1,5 @@
+---
+"@langchain/community": patch
+---
+
+improve bm25 sorting function


libs/langchain-community/src/utils/@furkantoprak/bm25/BM25.ts
```diff
@@ -16,21 +16,28 @@ export const getTermFrequency = (term: string, corpus: string) => {
...
+    document.text.includes(term)
   ).length;
   return Math.log(
     (documents.length - relevantDocuments + 0.5) / (relevantDocuments + 0.5) + 1
   );
 };
 
+export interface BMInputDocument<T> {
+  /** The text from the original document */
+  text: string;
```
AllenFang, line 31: let me know if team prefer to use a method to extract the text rather than adding another text field BMInputDocument:

ex: 
```
export interface BMInputDocument<T> {
  extractText: (docs: T) => string;
  /** The original document */
  docs: T;
}
```

christian-bromann:
@AllenFang thanks for the PR 🙏 do you think we can add some unit tests to verify the changes?

libs/langchain-community/src/utils/@furkantoprak/bm25/BM25.ts
```diff
@@ -53,16 +63,16 @@ export type BMSorter = (firstEl: BMDocument, secondEl: BMDocument) => number;
...
  * If this option is provided, the return type will not be an array of scores but an array of documents with their scores.
  */
-export function BM25(
-  documents: string[],
+export function BM25<T>(
+  documents: BMInputDocument<T>[],
   keywords: string[],
   constants?: BMConstants,
-  sorter?: BMSorter
-): number[] | BMDocument[] {
```
christian-bromann, line 61: Wouldn't this be a breaking change?

christian-bromann (changes requested):

christian-bromann:
@AllenFang let me know if you are still interested landing this change.

libs/langchain-community/src/utils/@furkantoprak/bm25/BM25.ts
hntrl: fine since you're really using the BM25Retriever, and this is more of an implementation detail. We also advise people to pin community versions

PR #8433, opened 2025-06-28
fix(core): prevent tool call chunks from merging incorrectly in AIMes…

### LINKED ISSUE #8276: AIMessageChunk incorrectly merged tool_calls
### Checked other resources

## The following code

```ts
aiMessageChunk.concat(aiMessageChunk)
```

## Root cause demo

```ts
function _mergeLists(left, right) {
  debugger;
  if (left === undefined && right === undefined) {
    return undefined;
  } else if (left === undefined || right === undefined) {
    return left || right;
  } else {
    const merged = [...left];
    for (const item of right) {
      console.log("Merging item:", item);
      if (
        typeof item === "object" &&
        "index" in item &&
        typeof item.index === "number"
      ) {
        const toMerge = merged.findIndex(
          (leftItem) => leftItem.index === item.index
        );
        if (toMerge !== -1) {
          merged[toMerge] = _mergeDicts(merged[toMerge], item);
        } else {
          merged.push(item);
        }
      } else if (
        typeof item === "object" &&
        "text" in item &&
        item.text === ""
      ) {
        // No-op - skip empty text blocks
        continue;
      } else {
        merged.push(item);
      }
    }
    return merged;
  }
}

function _mergeDicts(
  left,
  right
) {
  debugger;
  const merged = { ...left };
  for (const [key, value] of Object.entries(right)) {
    if (merged[key] == null) {
      merged[key] = value;
    } else if (value == null) {
      continue;
    } else if (
      typeof merged[key] !== typeof value ||
      Array.isArray(merged[key]) !== Array.isArray(value)
    ) {
      throw new Error(
        `field[${key}] already exists in the message chunk, but with a different type.`
      );
    } else if (typeof merged[key] === "string") {
      if (key === "type") {
        // Do not merge 'type' fields
        continue;
      }
      merged[key] += value;
    } else if (typeof merged[key] === "object" && !Array.isArray(merged[key])) {
      merged[key] = _mergeDicts(merged[key], value);
    } else if (Array.isArray(merged[key])) {
      merged[key] = _mergeLists(merged[key], value);
    } else if (merged[key] === value) {
      continue;
    } else {
      console.warn(
        `field[${key}] already exists in this message chunk and value has unsupported type.`
      );
    }
  }
  return merged;
}

var left = {
  tool_calls: [
    {
      id: "call_50556714",
      function: "[Object]",
      index: 0,
      type: "function",
    },
  ],
};

var right = {
  tool_calls: [
    {
      id: "call_79113425",
      function: "[Object]",
      index: 0,
      type: "function",
    },
  ],
};
var merged = _mergeDicts(left, right);

console.log("Merged result:", merged);

/**
Merged result: {
  tool_calls: [
    {
      id: 'call_50556714call_79113425',
      function: '[Object][Object]',
      index: 0,
      type: 'function'
    }
  ]
}
 */
```

_No response_

### Description

- I’m trying to concat AIMessageChunks from a same stream to get a gathered AIMessage
- But the the GatheredAIMessageChunk’s tool related properties are messed.

### System Info

This doesn’t matter , I believe my code tells why

christian-bromann:
## Description

This PR fixes a bug in `AIMessageChunk.concat()` where tool call chunks with the same index were being incorrectly merged instead of preserved as separate tool calls.

### Problem

When using models like Ollama that return multiple tool calls in chunks, the `_mergeLists` function was merging tool call chunks that had the same `index` property. This resulted in malformed tool calls with concatenated properties:

```javascript
// Before fix - incorrectly merged:
{
  name: "add_new_taskadd_ideas",
  args: '{"tasks":["buy tomatoes","help child with math"]}{"ideas":["read about Angular 19 updates"]}',
  id: "9fb5c937-6944-4173-84be-ad1caee1cedd1c9d8514-874a-4d8a-a9a9-604b9d831b60"
}

// After fix - correctly preserved as separate calls:
[
  {
    name: "add_new_task",
    args: '{"tasks":["buy tomatoes","help child with math"]}',
    id: "9fb5c937-6944-4173-84be-ad1caee1cedd"
  },
  {
    name: "add_ideas", 
    args: '{"ideas":["read about Angular 19 updates"]}',
    id: "5abf542e-87f3-4899-87c6-8f7d9cb6a28d"
  }
]
```

### Solution

- Replaced `_mergeLists()` with simple array concatenation using the spread operator
- Added comprehensive test case to verify tool call chunks remain separate during concatenation
- Removed unused `_mergeLists` import

### Impact

- ✅ Multiple tool calls now work correctly with streaming providers like Ollama
- ✅ Tool call arguments remain valid JSON and can be parsed properly
- ✅ No more `invalid_tool_calls` due to malformed merged data
- ✅ Existing functionality preserved (all tests pass)

### Testing

- Added specific test case for the reported scenario
- All existing AIMessageChunk tests continue to pass
- Verified fix works with the exact example from the issue report

Fixes #8276


hntrl (commented):
There's two things that I'm not totally sold on regarding this:

1. this isn't a breaking change
2. that we don't want to collapse multiple tool call chunks based on index

I think what would solve the original problem is if we add a condition that the ID in the tool call chunk and the index match in order for chunks to collapse inside of `_mergeLists`

christian-bromann:
@hntrl good call, I've updated the PR to fix the bug in the `mergeList` function.

PR #8443, opened 2025-07-01
minor[openai]: break out ChatOpenAI

hntrl:
This PR involves a refactor change to the `ChatOpenAI` API to more closely align with offerings from OpenAI:

1. Breaks out the original implementation of `ChatOpenAI` into two classes: `ChatOpenAICompletions` and `ChatOpenAIResponses`
    - These two classes extend a `BaseChatOpenAI` class to hoist common needs between the two APIs.
2. Adds a new implementation of `ChatOpenAI` that works like any other chat model implementation, except it routes to two inner `completions` and `responses` chat model instances based on existing criteria that existed in `ChatOpenAI` before this change.
3. Replaces existing usage of `ChatOpenAI` as a class extension with `ChatOpenAICompletions` in other provider packages where applicable.

## 💭 Why it's needed

> *Responses and Completions have diverged significantly*

When the responses API was new, the responses logic was in-lined alongside the existing implementation since they weren't distinct enough to justify it existing independently. Fast forward to today, and that has changed: reasoning content, built in tools, zdr, special responses formatting, and other openAI features exist exclusively (or are subtly different) within the responses API. So much so that it's safe to consider them semantically different at this point.

> *The completions format has developed itself into a standard beyond the public offering from OpenAI*

Players like openrouter, deepseek, xAI, etc. all have some form of completions binding to interface with it. This makes it even more important to isolate completions from responses.

  - The current implementation/ inheritance structure would _technically_ allow you to call the responses API against a completions-only chat model 😨

> *The current implementation is based on old assumptions that are awkward to maintain*

Since things are now ordered better into a distinct completions/ responses hierarchy, it will be easier to maintain these two API implementations independently of each other. This also gives us an easier path to introduce additional output formats in the near-future also.

## 🪨 Migration path

Migrating from the old to new `ChatOpenAI` implementations will look exactly the same for dependants. You can use `ChatOpenAI` in the exact same way without making any changes.

(We won't be exporting the new broken out classes externally just yet)

The minor change is justified since there are a number of classes that extend the old chat model implementation (xai, deepseek, fireworks, external oss packages, etc.), and introducing this change on its own will break existing class inheritance. The upgrade path for those cases would look something like:

```diff ts
- export class ChatXAI extends ChatOpenAI<ChatXAICallOptions>
+ export class ChatXAI extends ChatOpenAICompletions<ChatXAICallOptions>
```

## 📝 Other Notes

This is a pretty significant refactor, which means the diff is significantly confusing. You can see the new implementation if you look at the first commit in this PR. The commit right after that replaces the logic in `chat_models.ts`

The only changes here are hierarchical changes, meaning that the I/O of `ChatOpenAI` and the new classes won't change at all. A few key logic changes have been made to support this which will have associated review comments.

CI for lower and latest deps in downstream packages will be failing until we release 0.6

<details>
<summary>A complete refactor change log</summary>

- `ChatOpenAI.createResponseFormat` → `BaseChatOpenAI._getResponseFormat`
- `ChatOpenAI.getReasoningParams` → `BaseChatOpenAI._getReasoningParams`
- `ChatOpenAI._getClientOptions` → `BaseChatOpenAI._getClientOptions`
- `ChatOpenAI.bindTools` → `BaseChatOpenAI.bindTools`
- `ChatOpenAI.responseApiWithRetry` → `ChatOpenAIResponses.completionWithRetry`
- `ChatOpenAI.completionWithRetry`  → `ChatOpenAICompletions.completionWithRetry`
- invocationParams are now separated into their respective classes (used to be one big closure)
- `ChatOpenAI._responseApiGenerate` → `ChatOpenAIResponses._generate`
- `ChatOpenAI._generate` → `ChatOpenAICompletions._generate`
- `ChatOpenAI.withStructuredOutput` → `BaseChatOpenAI.withStructuredOutput`
- `ChatOpenAI.completionWithRetry` & `ChatOpenAI.betaParsedCompletionWithRetry` → `ChatOpenAICompletions.completionWithRetry`
    - this probably needs testing, the decision logic has been changed from looking at the call options to looking at the request directly
- `ChatOpenAI.getEstimatedTokenCountFromPrompt` → `BaseChatOpenAI._getEstimatedTokenCountFromPrompt`
- `ChatOpenAI.getNumTokensFromGenerations` → `BaseChatOpenAI._getNumTokensFromGenerations`
- `ChatOpenAI.getNumTokensFromMessages` → `BaseChatOpenAI.getNumTokensFromMessages`
- `interopZodResponseFormat` hoisted to `utils/openai.ts`
- `ChatOpenAIStructuredOutputMethodOptions` has been replaced with `StructuredOutputMethodOptions`
    - A deprecation was added in 0.2.6 for this
- `ChatOpenAICallOptions` → `BaseChatOpenAICallOptions`
- `ChatOpenAI._combineLLMOutput` → `BaseChatOpenAI._combineLLMOutput`
- `ChatOpenAI._llmType` → `BaseChatOpenAI._llmType`
- `ChatOpenAI.lc_name` → `BaseChatOpenAI.lc_name`
- `ChatOpenAI.callKeys` → `BaseChatOpenAI.callKeys`
- `ChatOpenAI.lc_serializable` → `BaseChatOpenAI.lc_serializable`
- `ChatOpenAI.lc_secrets` → `BaseChatOpenAI.lc_secrets`
- `ChatOpenAI.lc_aliases` → `BaseChatOpenAI.lc_aliases`
- `ChatOpenAI.lc_serializable_keys` → `BaseChatOpenAI.lc_serializable_keys`
- `ChatOpenAI.getLsParams` → `BaseChatOpenAI.getLsParams`
- `ChatOpenAI._identifyingParams` → `BaseChatOpenAI._identifyingParams`
- `ChatOpenAI.identifyingParams` → `BaseChatOpenAI.identifyingParams`
- `ChatOpenAI._convertOpenAIChatCompletionMessageToBaseMessage` → `ChatOpenAICompletions._convertCompletionsMessageToBaseMessage`
- `ChatOpenAI._convertOpenAIDeltaToBaseMessageChunk` → `ChatOpenAICompletions._convertCompletionsDeltaToBaseMessageChunk`
- `TokenUsage` → `OpenAILLMOutput["tokenUsage"]` (token usage inlined)
- `_convertOpenAIResponsesDeltaToBaseMessageChunk` → `ChatOpenAIResponses._convertResponsesDeltaToBaseMessageChunk`
- `_convertOpenAIResponsesMessageToBaseMessage` → `ChatOpenAIResponses._convertResponsesMessageToBaseMessage`
- `_convertMessagesToOpenAIResponsesParams` → `ChatOpenAIResponses._convertMessagesToResponsesParams`
    - the `model` and `zdrEnabled` positional args have been removed for class accessors
- `_reduceChatOpenAIToolsForResponses` → `ChatOpenAIResponses._reduceChatOpenAITools`
- `_convertChatOpenAIToolTypeToOpenAITool` → `BaseChatOpenAI._convertChatOpenAIToolToCompletionsTool`
</details>

## 🧪 Refactor testing 

> *Traces still come across as expected*

Test case:
```ts
import { RunnableLambda, RunnableParallel } from "@langchain/core/runnables";
import { ChatOpenAI } from "../chat_models.js";

async function main() {
  const modelRunnable = (model: ChatOpenAI) =>
    RunnableLambda.from(() =>
      model.invoke(
        "do you know the answer to life, the universe, and everything?"
      )
    );

  const runnable = RunnableParallel.from([
    modelRunnable(new ChatOpenAI({ model: "gpt-4o" })),
    modelRunnable(new ChatOpenAI({ model: "gpt-4o", useResponsesApi: true })),
    modelRunnable(new ChatOpenAI({ model: "o3-mini" })),
  ]);

  await runnable.invoke(undefined);
}

void main();
```

before: [trace](https://smith.langchain.com/public/eddac5fb-3658-480e-b598-c924fcc806dd/r)
after: [trace](https://smith.langchain.com/public/88284b32-8e43-4694-8cb0-3c1eca6de522/r)

> *Existing integration tests work for the new `ChatOpenAI`*

before: [results](https://static-html-worker.hunter-79f.workers.dev/langchain-openai/pr8443/test-report-before-refactor)
after: [results](https://static-html-worker.hunter-79f.workers.dev/langchain-openai/pr8443/test-report-after-refactor)

Some results are variable between the two test runs because integration testing with LLMs is not deterministic (yet)


libs/langchain-openai/src/chat_models.ts
```diff
@@ -2952,460 +2688,739 @@ export class ChatOpenAI<
...
-    const parsedWithFallback = parserAssign.withFallbacks({
-      fallbacks: [parserNone],
-    });
-    return RunnableSequence.from<
-      BaseLanguageModelInput,
-      { raw: BaseMessage; parsed: RunOutput }
-    >([{ raw: llm }, parsedWithFallback]);
+    this.responses = bindChatOpenAIConfig(this.responses, config);
+    this.completions = bindChatOpenAIConfig(this.completions, config);
+    return this;
```
hntrl, line 3412: This isn't pretty, but replacing these functions in the only way to override config options without having to introduce a `RunnableBinding` afaict (which might not work for this implementation).

libs/langchain-openai/src/chat_models.ts
```diff
@@ -2823,64 +2316,307 @@ export class ChatOpenAI<
...
+          );
         }
-      })
-    );
+        // const res =
+      } catch (e) {
+        const error = wrapOpenAIClientError(e);
+        throw error;
+      }
+    });
```
hntrl, line 2477: This was handled before by providing a separate `betaCompletionWithRetry` function that used the parse method instead. The `response_format` passed to that old method existed on options, whereas that response format is fetched directly from the request object. (had to do this because of the way the prop drilling was setup)

libs/langchain-openai/src/chat_models.ts
```diff
@@ -2952,460 +2688,739 @@ export class ChatOpenAI<
...
+   * @deprecated
+   * This getter is deprecated and will be removed in a future release.
+   * Please use {@link completions.getNumTokensFromMessages} directly instead.
+   *
+   * Returns a reference to the `getNumTokensFromMessages` method from the `completions` instance,
+   * which calculates the number of tokens in an array of messages.
+   */
+  get getNumTokensFromMessages() {
+    return this.completions.getNumTokensFromMessages;
   }
```
hntrl, line 3425: `getNumTokensFromMessages` was a public method on the old ChatOpenAI class, and since we're avoiding breaking public APIs this is re-exported here with a deprecation notice. (this still exists on the completions class)

libs/langchain-openai/src/chat_models-new.ts
```diff
@@ -0,0 +1,3424 @@
...
+            request,
+            clientOptions
+          );
+        }
+        // const res =
+      } catch (e) {
+        const error = wrapOpenAIClientError(e);
+        throw error;
+      }
+    });
```
hntrl, line 2602: This was handled before by providing a separate betaCompletionWithRetry function that used the parse method instead. The response_format passed to that old method existed on options, whereas that response format is fetched directly from the request object. (had to do this because of the way the prop drilling was setup)

libs/langchain-openai/src/chat_models-new.ts
```diff
@@ -0,0 +1,3405 @@
...
+          return await this.client.chat.completions.create(
+            request,
+            clientOptions
+          );
+        }
+      } catch (e) {
+        const error = wrapOpenAIClientError(e);
+        throw error;
+      }
+    });
```
hntrl, line 2599: This was handled before by providing a separate betaCompletionWithRetry function that used the parse method instead. The response_format passed to that old method existed on options, whereas that response format is fetched directly from the request object. (had to do this because of the way the prop drilling was setup)

libs/langchain-openai/src/chat_models-new.ts
```diff
@@ -0,0 +1,3404 @@
...
+            runManager
+          );
+        },
+      } as Partial<TClass>);
+    };
+
+    this.responses = bindChatOpenAIConfig(this.responses, config);
+    this.completions = bindChatOpenAIConfig(this.completions, config);
+    return bindChatOpenAIConfig(this, config);
+  }
```
hntrl: This isn't pretty, but replacing these functions in the only way to override config options without having to introduce a RunnableBinding afaict (which prob won't work for this structure).

libs/langchain-openai/src/chat_models-new.ts
```diff
@@ -0,0 +1,3404 @@
...
+          return await this.client.chat.completions.create(
+            request,
+            clientOptions
+          );
+        }
+      } catch (e) {
+        const error = wrapOpenAIClientError(e);
+        throw error;
+      }
+    });
```
hntrl: This was handled before by providing a separate betaCompletionWithRetry function that used the parse method instead. The response_format passed to that old method existed on options, whereas that response format is fetched directly from the request object. (had to do this because of the way the prop drilling was setup)

libs/langchain-openai/src/utils/openai.ts
```diff
@@ -83,3 +91,63 @@ export function formatToOpenAIToolChoice(
...
+    return makeParseableResponseFormat(
+      {
+        type: "json_schema",
+        json_schema: {
+          ...props,
+          name,
+          strict: true,
+          schema: toJSONSchemaV4(zodSchema, {
+            cycles: "ref", // equivalent to nameStrategy: 'duplicate-ref'
+            reused: "ref", // equivalent to $refStrategy: 'extract-to-root'
```
jacoblee93, line 134: Nice, what happens if these optimizations aren't set/how important are they?

libs/langchain-openai/src/chat_models.ts
```diff
@@ -105,31 +102,63 @@ import {
 } from "./utils/openai-format-fndef.js";
 import { _convertToOpenAITool } from "./utils/tools.js";
 
+const _FUNCTION_CALL_IDS_MAP_KEY = "__openai_function_call_ids__";
+
+type ResponsesTool = Exclude<
+  OpenAIClient.Responses.ResponseCreateParams["tools"],
```
jacoblee93, line 108: Is this the same as `NonNullable`?

libs/langchain-openai/src/chat_models.ts
```diff
@@ -341,10 +372,11 @@ const completionsApiContentBlockConverter: StandardContentBlockConverter<{
 };
 
 // Used in LangSmith, export is important here
+// TODO: put this conversion elsewhere
 export function _convertMessagesToOpenAIParams(
   messages: BaseMessage[],
   model?: string
-): OpenAICompletionParam[] {
+): OpenAIClient.Chat.Completions.ChatCompletionMessageParam[] {
```
jacoblee93, line 379: nit: why remove the short alias for this?

libs/langchain-openai/src/chat_models.ts
```diff
@@ -401,2256 +433,1863 @@ export function _convertMessagesToOpenAIParams(
...
-            fallthroughCallTypes.includes(item.type)
-          );
-          if (fallthroughCalls.length > 0) input.push(...fallthroughCalls);
-        }
+  /** @deprecated Use "model" instead */
+  modelName: string;
 
-        return input;
-      }
+  model = "gpt-3.5-turbo";
```
jacoblee93, line 590: If we are minor bumping, should we remove this default at this point?

libs/langchain-openai/src/chat_models.ts
```diff
@@ -401,2256 +433,1863 @@ export function _convertMessagesToOpenAIParams(
...
-      }
-    } else if (item.type === "reasoning") {
-      additional_kwargs.reasoning = item;
-    } else {
-      additional_kwargs.tool_outputs ??= [];
-      additional_kwargs.tool_outputs.push(item);
-    }
-  }
+  /** @deprecated Use "apiKey" instead */
+  openAIApiKey?: string;
```
jacoblee93, line 613: Could probably remove this one too at this point and make it throw a descriptive error if passed into constructor

libs/langchain-openai/src/chat_models.ts
```diff
@@ -401,2256 +433,1863 @@ export function _convertMessagesToOpenAIParams(
...
+    ) {
+      return interopZodResponseFormat(
+        resFormat.json_schema.schema,
+        resFormat.json_schema.name,
+        {
+          description: resFormat.json_schema.description,
+        }
+      );
+    }
+    return resFormat as
```
jacoblee93, line 907: Is this cast needed?

libs/langchain-openai/src/chat_models.ts
```diff
@@ -401,2256 +433,1863 @@ export function _convertMessagesToOpenAIParams(
...
+        ? options?.tool_choice
+        : (() => {
+            const formatted = formatToOpenAIToolChoice(options?.tool_choice);
+            if (typeof formatted === "object" && "type" in formatted) {
+              return { type: "function", name: formatted.function.name };
+            } else {
+              return undefined;
+            }
+          })(),
+      text: (() => {
```
jacoblee93, line 1460: We pass a function into the client?

libs/langchain-openai/src/chat_models.ts
```diff
@@ -401,2256 +433,1863 @@ export function _convertMessagesToOpenAIParams(
...
+      model_name: response.model,
+    };
+
+    const additional_kwargs: {
+      [key: string]: unknown;
+      refusal?: string;
+      reasoning?: OpenAIClient.Responses.ResponseReasoningItem;
+      tool_outputs?: unknown[];
+      parsed?: unknown;
+      [_FUNCTION_CALL_IDS_MAP_KEY]?: Record<string, string>;
```
jacoblee93, line 1644: A bit out of the loop but what is this?

libs/langchain-openai/src/chat_models.ts
```diff
@@ -401,2256 +433,1863 @@ export function _convertMessagesToOpenAIParams(
...
 
-      return;
-    }
+        if (role === "tool") {
+          const toolMessage = lcMsg as ToolMessage;
 
-    const messagesMapped: OpenAICompletionParam[] =
-      _convertMessagesToOpenAIParams(messages, this.model);
+          // Handle computer call output
+          if (additional_kwargs?.type === "computer_call_output") {
```
jacoblee93, line 1896: Not super familiar with our built-in tools support so assuming this is just refactored code

libs/langchain-openai/src/chat_models.ts
```diff
@@ -401,2256 +433,1863 @@ export function _convertMessagesToOpenAIParams(
...
+    reasoning: ChatOpenAIReasoningSummary
+  ): OpenAIClient.Responses.ResponseReasoningItem {
+    // combine summary parts that have the the same index and then remove the indexes
+    const summary = (
+      reasoning.summary.length > 1
+        ? reasoning.summary.reduce(
+            (acc, curr) => {
+              const last = acc.at(-1);
+
+              if (last!.index === curr.index) {
```
jacoblee93, line 2145: Any reason not to be defensive here?

libs/langchain-openai/src/chat_models.ts
```diff
@@ -2823,589 +2444,967 @@ export class ChatOpenAI<
...
+        text: chunk.content,
+        generationInfo,
+      });
+      yield generationChunk;
+      await runManager?.handleLLMNewToken(
+        generationChunk.text ?? "",
+        newTokenIndices,
+        undefined,
+        undefined,
+        undefined,
```
jacoblee93, line 2519: As an aside I hate this so much lol

libs/langchain-openai/src/chat_models.ts
```diff
@@ -2823,589 +2444,967 @@ export class ChatOpenAI<
...
-    });
-    const parsedWithFallback = parserAssign.withFallbacks({
-      fallbacks: [parserNone],
-    });
-    return RunnableSequence.from<
-      BaseLanguageModelInput,
-      { raw: BaseMessage; parsed: RunOutput }
-    >([{ raw: llm }, parsedWithFallback]);
+    this.responses = new ChatOpenAIResponses(fields);
+    this.completions = new ChatOpenAICompletions(fields);
```
jacoblee93, line 3324: Nice

libs/langchain-openai/src/utils/openai.ts
hntrl: Without these the serialized schemas wont be compatible with openai's specific output format

jacoblee93: Damn

libs/langchain-openai/src/chat_models.ts
hntrl: yeah love this. any model in particular? (4.1?)

jacoblee93: Have been removing the default in general and making the user provide it. Not really a single good default choice anymore

libs/langchain-openai/src/chat_models.ts
hntrl: prob just a personal preference thing, but I get more info on what exactly the type is when its exploded like this.

I won't die on this hill though if we prefer aliases, though I would change it since `OpenAICompletionParam` is ambiguous between responses + completions even though its a completions type

jacoblee93: Your call, was just curious

libs/langchain-openai/src/chat_models.ts
hntrl: yes unfortunately (for now). The type guard above this doesn't actually restrict what resFormat can be since it has additional conditions (`resFormat.type === "json_schema"`)

libs/langchain-openai/src/chat_models.ts
hntrl: good question! I have no idea!

This is just apart of refactored code. My guess is this covers an edge case with function calls in a similar fashion to python. Will prob be readdressed when we get to standard outputs

libs/langchain-openai/src/chat_models.ts
hntrl: yeah this just got moved from an old location

libs/langchain-openai/src/chat_models.ts
hntrl: A function that gets immediately called

libs/langchain-openai/src/chat_models.ts
hntrl: This is just relocated code. It's worked so far so I'll prob just leave it and give it another look for standard outputs

libs/langchain-openai/src/chat_models.ts
hntrl: will do this in another PR

PR #8472, opened 2025-07-08
support api key authentication for Bedrock API

0x-fang:
AWS Bedrock has introduced API key support on 7/7, simplifying authentication for the Amazon Bedrock API. To enable this new feature in langchainjs, this update bumps the required "@aws-sdk/client-bedrock-runtime" version to "^3.840.0". It has been tested and worked as expected with new `AWS_BEARER_TOKEN_BEDROCK` environment variable.

- [x] bump up the version of "@aws-sdk/client-bedrock-runtime" in package.json and update the yarn.lock
- [x] update README file to add API key support feature for Bedrock.

### DIFF
--- a/libs/langchain-aws/README.md
+++ b/libs/langchain-aws/README.md
@@ -40,12 +40,19 @@ The field you need depends on the package manager you're using, but we recommend
 
 This package contains the `ChatBedrockConverse` class, which is the recommended way to interface with the AWS Bedrock Converse series of models.
 
-To use, install the requirements, and configure your environment.
+To use, install the requirements, and configure your environment following the traditional authentication methods.
 
 ```bash
 export BEDROCK_AWS_REGION=
 export BEDROCK_AWS_SECRET_ACCESS_KEY=
 export BEDROCK_AWS_ACCESS_KEY_ID=
+```
+
+Alternatively, set the `AWS_BEARER_TOKEN_BEDROCK` environment variable locally for API Key authentication. For additional API key details, refer to [docs](https://docs.aws.amazon.com/bedrock/latest/userguide/api-keys.html).
+
+```bash
+export BEDROCK_AWS_REGION=
+export AWS_BEARER_TOKEN_BEDROCK=
 ```
 
 Then initialize
--- a/libs/langchain-aws/package.json
+++ b/libs/langchain-aws/package.json
@@ -33,7 +33,7 @@
   "license": "MIT",
   "dependencies": {
     "@aws-sdk/client-bedrock-agent-runtime": "^3.755.0",
-    "@aws-sdk/client-bedrock-runtime": "^3.755.0",
+    "@aws-sdk/client-bedrock-runtime": "^3.840.0",
     "@aws-sdk/client-kendra": "^3.750.0",
     "@aws-sdk/credential-provider-node": "^3.750.0"
   },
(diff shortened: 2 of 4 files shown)


danny-avila:
Is there any way to provide the API key without an environment variable? For example, similar to defining `credentials` object and providing `accessKeyId` from elsewhere like a database.

hassiebp:
> Is there any way to provide the API key without an environment variable? For example, similar to defining credentials object and providing accessKeyId from elsewhere like a database.

+1, would be great to decide on runtime what key to use rather than via env var

hntrl:
@danny-avila @hassiebp good callout! Opened an issue to track here: https://github.com/langchain-ai/langchainjs/issues/8532

(would love a PR to add support if you're up for it!)

0x-fang:
@danny-avila @hassiebp actually there is one way with bedrock-runtime-client JS SDK only. you can create a client with token parameter, set authSchemePreference, and use it to call InvokeModel API without an environment variable. 

```
const client = new BedrockRuntimeClient({
  token: { token: "<api-key>" },
  authSchemePreference: ["httpBearerAuth"],
});

const llm = new ChatBedrockConverse({
  model: "<my-model-id>",
  region: process.env.BEDROCK_AWS_REGION,
  client: client,
});

const res = await llm.invoke([
  ["system", "You are a helpful assistant."],
  ["human", "What is capital of France."],
]);

> {
  res: AIMessage {
    "id": "1e3bc241-6077-4bc3-b143-63651dc53911",
    "content": "The capital of France is Paris.",
    "additional_kwargs": {},
    "response_metadata": {
      "$metadata": {
        "httpStatusCode": 200,
        "requestId": "1e3bc241-6077-4bc3-b143-63651dc53911",
        "attempts": 1,
        "totalRetryDelay": 0
      },
      "metrics": {
        "latencyMs": 302
      },
      "stopReason": "end_turn",
      "usage": {
        "inputTokens": 19,
        "outputTokens": 10,
        "totalTokens": 29
      }
    },
    "tool_calls": [],
    "invalid_tool_calls": [],
    "usage_metadata": {
      "input_tokens": 19,
      "output_tokens": 10,
      "total_tokens": 29
    }
  }
}
```

danny-avila:
Thanks @0x-fang! It wasn't immediately clear what the `BedrockRuntimeClient` expected to use an API Key defined at runtime.

In any case, we need a release for `@langchain/aws` or we need to override `@aws-sdk/client-bedrock-runtime`

0x-fang:
Hi @danny-avila the above approach only works with `bedrock-runtime-client` Javascript SDK, but not boto3. If we're using `@langchain/aws` or `@aws-sdk/client-bedrock-runtime`, which internally rely on boto3, we'll need to set env variable to support API key during bedrock API call. The BedrockRuntimeClient in boto3 (e.g., `client = boto3.client('bedrock-runtime')`) does not accept API-Key as a parameter. 

We had raised a PR to bump up the boto3 version in `@langchain/aws` to support the env variable for API key. https://github.com/langchain-ai/langchain-aws/pull/536 Thanks

hntrl:
@danny-avila we missed the release step, apologies! Just pushed out 0.1.12 off of main.

For the issue that got added to address an api key param, would people prefer initializing as @0x-fang described or through an implicit `apiKey` field in the constructor of `ChatBedrockConverse`?

PR #8499, opened 2025-07-14
fix(messages): prevent audio ID concatenation when merging chunks

### LINKED ISSUE #8487: `concat` concatenates audio id twice
### Checked other resources

### Example Code

```ts
import { concat } from '@langchain/core/utils/stream';

async function test() {
        model: 'gpt-4o-audio-preview',
        modalities: ['text', 'audio'],
        audio: {
            format: 'pcm16',
            voice: 'alloy',
        },
    });

    const stream = await model.stream('Hello how are you');
    let result = undefined as AIMessageChunk | undefined;
        result = result ? concat(result, chunk) : chunk;
    }
    //@ts-expect-error test
    console.log(result?.additional_kwargs?.audio?.id)
}

test();
```

This logs `audio_6871323ab2fc8191b2cf516af96cd851audio_6871323ab2fc8191b2cf516af96cd851` which then causes error:
```
400 Assistant audio audio_6871323ab2fc8191b2cf516af96cd851audio_6871323ab2fc8191b2cf516af96cd851 not found.
```
When asking the next message. I'm guessing the audio id should be `audio_6871323ab2fc8191b2cf516af96cd851` instead

_No response_

### Description

- i ask a text message through streaming
- i receive audio chunks
- i concat the chunks using the streaming utility
- i create a new AiMessage by passing the concat result in it 
- the error happens

### System Info

N/A

christian-bromann:
## Description

This PR fixes a bug where audio IDs were being incorrectly concatenated when streaming audio chunks from OpenAI's `gpt-4o-audio-preview` model, causing "Assistant audio not found" errors.

## Problem

When streaming audio chunks, the `_mergeDicts` function in `langchain-core/src/messages/base.ts` was concatenating audio IDs, transforming:
- `"audio_6871323ab2fc8191b2cf516af96cd851"` → `"audio_6871323ab2fc8191b2cf516af96cd851audio_6871323ab2fc8191b2cf516af96cd851"`

This resulted in invalid audio IDs that caused subsequent API requests to fail.

## Solution

Added a targeted fix to preserve the first non-empty value for audio ID fields instead of concatenating them:

```typescript
// Special case: preserve the first non-empty value for audio.id field
// This prevents audio ID duplication when streaming audio chunks
if (key === "id" && merged[key] && merged[key].startsWith("audio_")) {
  continue;
}
```

## Testing

- Added comprehensive unit tests for audio ID preservation
- Verified other audio fields still concatenate correctly
- Ensured non-audio ID fields maintain existing behavior
- All existing message tests pass (51 total)

## Future Considerations

While this addresses the immediate audio ID issue, similar fields may need consideration:
- Request/response IDs (`request_id`, `response_id`, `conversation_id`)
- Timestamps (fields ending with `_at`, `_time`, `timestamp`)
- Unique identifiers (fields ending with `_id`, `uuid`, `guid`)
- Index/position fields (`index`, `position`, `sequence`)
- Status/state fields (`status`, `state`, `phase`)

Fixes #8487

### DIFF
--- a/langchain-core/src/messages/base.ts
+++ b/langchain-core/src/messages/base.ts
@@ -419,6 +419,11 @@ export function _mergeDicts(
         // Do not merge 'type' fields
         continue;
       }
+      // Special case: preserve the first non-empty value for audio.id field
+      // This prevents audio ID duplication when streaming audio chunks
+      if (key === "id" && merged[key].startsWith("audio_")) {
+        continue;
+      }
       merged[key] += value;
     } else if (typeof merged[key] === "object" && !Array.isArray(merged[key])) {
       merged[key] = _mergeDicts(merged[key], value);
--- a/langchain-core/src/messages/tests/base_message.test.ts
+++ b/langchain-core/src/messages/tests/base_message.test.ts
@@ -498,4 +498,85 @@ describe("usage_metadata serialized", () => {
     expect(jsonConcatenatedAIMessageChunk).toContain("output_tokens");
     expect(jsonConcatenatedAIMessageChunk).toContain("total_tokens");
   });
+
+  test("Audio ID should not be concatenated when merging chunks", () => {
+    const audioId = "audio_6871323ab2fc8191b2cf516af96cd851";
+
+    const chunk1 = new AIMessageChunk({
+      content: "Hello",
+      additional_kwargs: {
+        audio: {
+          id: audioId,
+          data: "data1",
+          transcript: "Hello",
+          expires_at: 1729201448,
+        },
+      },
+    });
+
+    const chunk2 = new AIMessageChunk({
+      content: " world",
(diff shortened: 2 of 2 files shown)


hntrl:
Will call out that there is a similar issue when using the image generation tool with openai. Off the top of my head we retain each `partial_image` object for streaming responses, but the intended behavior is to keep the latest one instead of retaining all of them. Similar semantics here where we only want to retain one audio id.

I'm also not sure how much of a fan I am of baking in the assumption that ids prefixed with `audio_` should always be collated this way. This is how OpenAI views the world, but whos to say anthropic and google will in the near future? There's some work to be done in terms of beefing up the base chunk class that can reconcile with these aggregation considerations.

hntrl:
Going to close this as this is being assessed as apart of v1.

PR #8500, opened 2025-07-14
feat: `zod@4.0` support

hntrl:
Fixes # (issue)


witvoet-hamiltonai:
Super helpful, as currently we're getting a lot of TypeError: Cannot read properties of undefined (reading 'parent')

hntrl:
This was identified as a breaking change so we're bundling it apart of our v1 effort. Continued on in #8792

PR #8504, opened 2025-07-15
returns Promise resolving to the response from the insert or upsert operation

### LINKED ISSUE #8503: addDocument() & addVectors() missing response in Milvus
The following awaiting function is returning a response, but the parent function `addDocument() & addVectors()` is set to return type of `void`.
```typescript
 const insertResp = this.autoId
      ? await this.client.insert(params)
      : await this.client.upsert(params);
```
https://github.com/langchain-ai/langchainjs/blob/d05510da7ace1073e3b8cbf41ab4c6f73128704b/libs/langchain-community/src/vectorstores/milvus.ts#L309C1-L312C1

princethirstydevs:
Fixes #8503


hntrl:
Hey @princethirstydevs! Appreciate you taking the time to make a PR.

Just returning the response from the implementation violates the types of the core abstraction, which is something we don't want. I left some comments in the linked issue to maybe help paint a clearer picture.

PR #8534, opened 2025-07-18
feat(tavily-search): adding startDate and endDate params to TavilySearch tool

michaelgriff:
## Summary
Adds `startDate` and `endDate` params to `TavilySearch` tool

Notes
The Start Date and End Date parameters can be used to return search results within a specified time range.

`startDate`: Filters search results to include only content published on or after this date.

`endDate`: Filters search results to include only content published on or before this date.


hntrl (commented):
Is it worth it to include start_date + end_date in TavilySearchParamsBase in utils?

hntrl:
tyvm @michaelgriff!

PR #8545, opened 2025-07-20
fix(cloudflare): similaritySearchVectorWithScore not returning id

stpwin:
Fixes # 
- similaritySearchVectorWithScore not returning id
-  use existing document.id if available when generating documentIds


hntrl (approved):
Thanks for the change @stpwin!

PR #8555, opened 2025-07-23
patch[gemini]: Gemini 2 5 Flash-Lite

afirstenberg:
* Since Gemini 2.5 Flash Lite on Vertex now works in more regions, use the same default region as other models (us-central1). (The previous default region is still valid.)
* Update tests to use release name gemini-2.5-flash-lite


hntrl:
Out of curiosity whats the motivation behind making this a minor bump?

PR #8558, opened 2025-07-23
feat(chroma): Add Chroma Cloud support

itaismith:
* Update Chroma dep to version 3.x.
* Add native support for Chroma Cloud.
* Update docs


libs/langchain-community/src/vectorstores/chroma.ts
```diff
@@ -376,6 +386,12 @@ export class Chroma extends VectorStore {
     const [firstDocuments] = documents;
     const [firstMetadatas] = metadatas;
 
+    if (firstDistances.some((item) => item === null)) {
```
itaismith, line 389: This is a limitation of our typing. The `distances` array entries are `number | null`, even thought they will never be `null`. Will update here when we fix

christian-bromann:
Ported to `v1` branch.

PR #8560, opened 2025-07-24
docs: fix example code of extraction tutorial

tksrmz:
the tutorial doc says that the example code uses `.nullish()` but it actually uses `.optional()`

> Above we used `.nullish()` for the attributes allowing the LLM to output `null` or `undefined` if it doesn't know the answer.

https://js.langchain.com/docs/tutorials/extraction

Fixes # (issue)


hntrl:
Thanks for the fix @tksrmz!

PR #8564, opened 2025-07-24
fix: Relax constraint on empty message content in convertMessageContentToParts

Annhiluc:
This change removes the constraint when building the Gemini API parts message that the string content must be non-empty. This is to ensure that even if empty content is passed along it will be populated as a fully fledged text component when calling the Gemini API.


hntrl:
Hi @Annhiluc, apologies for the delay.

Is this meant to be a draft PR? Would you mind either closing this (+ opening a new issue so we can track this if you're still encountering this), or marking this ready for review?

PR #8575, opened 2025-07-28
feat: expose maxCompletionTokens parameter, fix tests in IBM Watsonx implementation

FilipZmijewski:
Fixes # (issue)


hntrl (commented):
What's the motivation behind changing the input from `input` to `q`? Have to ask since this is technically a breaking change.

FilipZmijewski:
@hntrl So the tool that we are referring to changed its input schema from string to and object with `q` property. Since tools are dynamically loaded then this needed to be adjusted. We are aware of the change but we are just adjusting this. For a use case with an agent this won't cause any damage, only if someone is hard calling the tool.

FilipZmijewski:
Something seems to be failing but not related to this PR 
```Internal Error: Assertion failed: Expected workspace langchain-google-common-ee482c (/home/runner/work/langchainjs/langchainjs/libs/langchain-google-common/package.json) to have been resolved. Run "yarn install" to update the lockfile```

Issue #8283, opened 2025-05-30
Unable to stream output when using OpenAI's Responses API to call the web_search tool


mjzcng:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain.js documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain.js rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```typescript
import { HumanMessage, SystemMessage } from "@langchain/core/messages";
import { END, MessagesAnnotation, START, StateGraph } from "@langchain/langgraph";
import { ChatOpenAI } from "@langchain/openai";

const llm = new ChatOpenAI({
    modelName: "gpt-4.1", 
}).bindTools([
    { type: "web_search_preview" }
]);

const callModel = async (state: typeof MessagesAnnotation.State) => {
    const { messages } = state;
    const response = await llm.invoke(messages);
    return { messages: [response] };
};

const graph = new StateGraph(MessagesAnnotation)
                .addNode("assistant", callModel)
                .addEdge(START, "assistant")
                .addEdge("assistant", END);

const app = graph.compile();

(async () => {
    const messages = [
        new SystemMessage("You are a helpful AI assistant."),
        new HumanMessage("Hello")
    ];

    const stream = await app.stream(
        { messages },
        { streamMode: "messages" }
    );

    for await (const [message, _metadata] of stream) {
        console.log(`${message.getType()} MESSAGE CONTENT: ${JSON.stringify(message)}`);
        // console.log(`${message.getType()} MESSAGE CONTENT: ${message.content}`);
    }
})();
```

### Description

The above output shows only a single, one-time response rather than a streamed output.

[...]

PR #8578, opened 2025-07-28
fix: enable handleLLMNewToken events for useResponsesApi=true

### LINKED ISSUE #8577: Streaming Does not Work with OpenAI Responses API
### Checked other resources

### Example Code

The following code:

```typescript
import { HumanMessage } from '@langchain/core/messages';
import { MessagesAnnotation, StateGraph } from '@langchain/langgraph';

async function testDirectModelStreaming(useResponsesApi: boolean): Promise<void> {
  console.warn(`\n=== Direct Model Streaming (useResponsesApi: ${useResponsesApi ? 'true' : 'false'}) ===`);

    modelName: 'gpt-4',
    temperature: 0.7,
    // streaming: true, // same result with or without this
    useResponsesApi,
  });

  const messages = [new HumanMessage('Say "Hello world" in 10 words.')];
  const stream = model.streamEvents(messages, { version: 'v2' });

  for await (const event of stream) {
    console.warn(event.event);
  }
}

async function testLangGraphStreaming(useResponsesApi: boolean): Promise<void> {
  console.warn(`\n=== LangGraph Streaming (useResponsesApi: ${useResponsesApi ? 'true' : 'false'}) ===`);

    modelName: 'gpt-4',
    temperature: 0.7,
    // streaming: true, // same result with or without this
    useResponsesApi,
  });

  async function callModel(state: typeof MessagesAnnotation.State): Promise<{ messages: unknown[] }> {
    const messages = [new HumanMessage('Say "Hello world" in 10 words.'), ...state.messages];
    return { messages: [response] };
  }

  const workflow = new StateGraph(MessagesAnnotation)
    .addNode('agent', callModel)
    .addEdge('__start__', 'agent')
    .addConditionalEdges('agent', () => '__end__');

  const agent = workflow.compile();
  const eventStream = agent.streamEvents({ messages: [] }, { version: 'v2', streamMode: 'messages' });

  for await (const event of eventStream) {
    console.warn(event.event);
  }
}

async function runTests(): Promise<void> {
  await testDirectModelStreaming(false);
  await testDirectModelStreaming(true);
  await testLangGraphStreaming(true);
  await testLangGraphStreaming(false);
}

runTests().catch(console.error);
```

_No response_

### Description

Token-by-token streaming doesn't work when useResponsesApi: true is enabled in LangGraph applications, even though it works fine when calling the model directly.

## **Root Cause**

The issue is in the `ChatOpenAI` class implementation. When `useResponsesApi: true` is enabled:

1. `ChatOpenAI._streamResponseChunks()` routes to `this.responses._streamResponseChunks()`
2. However, it does **NOT** pass the `runManager` parameter to the responses method
3. This prevents `ChatOpenAIResponses._streamResponseChunks()` from emitting `handleLLMNewToken` events
4. The callback system never receives token-by-token updates

I am working on a fix for this with a test, and I'll submit a PR.

### System Info

```
@langchain/openai@0.6.3
@langchain/core@0.3.66
@langchain/langgraph@0.3.11
```

ninjudd:
- Fix ChatOpenAI._streamResponseChunks to pass runManager to responses._streamResponseChunks
- Fix ChatOpenAIResponses._streamResponseChunks to accept and use runManager parameter
- Add handleLLMEnd event emission for non-streaming responses
- Add regression test to verify handleLLMNewToken events are emitted
- Resolves token-by-token streaming issue when useResponsesApi=true is enabled

Fixes #8577

### DIFF
--- a/libs/langchain-openai/src/chat_models.ts
+++ b/libs/langchain-openai/src/chat_models.ts
@@ -1489,7 +1489,7 @@ export class ChatOpenAIResponses<
 
   async _generate(
     messages: BaseMessage[],
-    options: this["ParsedCallOptions"],
+    options: this["ParsedCallOptions"]
   ): Promise<ChatResult> {
     const invocationParams = this.invocationParams(options);
     if (invocationParams.stream) {
@@ -1604,7 +1604,7 @@ export class ChatOpenAIResponses<
         if (request.text?.format?.type === "json_schema" && !request.stream) {
           return await this.client.responses.parse(request, clientOptions);
         }
-        return await this.client.responses.create(request, clientOptions)
+        return await this.client.responses.create(request, clientOptions);
       } catch (e) {
         const error = wrapOpenAIClientError(e);
         throw error;
--- a/libs/langchain-openai/src/tests/chat_models_responses.int.test.ts
+++ b/libs/langchain-openai/src/tests/chat_models_responses.int.test.ts
@@ -78,7 +78,7 @@ function assertResponse(message: BaseMessage | BaseMessageChunk | undefined) {
 }
 
 test("Test with built-in web search", async () => {
-  const llm = new ChatOpenAI({ model: "gpt-4o-mini" });  
+  const llm = new ChatOpenAI({ model: "gpt-4o-mini" });
 
   // Test invoking with web search
   const firstResponse = await llm.invoke(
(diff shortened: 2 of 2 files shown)


ninjudd:
I believe this PR also fixes https://github.com/langchain-ai/langchainjs/issues/8283

libs/langchain-openai/src/chat_models.ts
```diff
@@ -1539,12 +1540,21 @@ export class ChatOpenAIResponses<
...
         },
       };
+
+      // Emit handleLLMEnd event for non-streaming responses
+      await runManager?.handleLLMEnd({
+        generations: [result.generations],
+        llmOutput: result.llmOutput,
+      });
+
+      return result;
```
hntrl: I don't believe this is necessary? This gets handled upstream from the base chat model (if it doesn't it should, I'll take a look)

https://github.com/ninjudd/langchainjs/blob/fa9f31a90212a7ed41f998e89a59099e971b3981/langchain-core/src/language_models/chat_models.ts#L479-L482

hntrl (commented):
nice catch @ninjudd! I don't believe this is something that was ever supported. The writeup + PR is very much appreciated ❤️

One note I had about handleLLMEnd, and I don't think the params being passed to `handleLLMNewToken` are exactly accurate. I'll pick up on this tomorrow since I think it's a little tricker then what is initially let on.

libs/langchain-openai/src/chat_models.ts
ninjudd: Ah, you may be right! Given that everything but streaming works without this PR, emitting `handleLLMEnd` here is probably not necessary. I can update the PR to remove this if you agree.

ninjudd:
Thanks @hntrl! Any idea when the next release will go out?

Also, where do I track that? Looks like the last tagged release on Github was June 19.

hntrl:
Hey @ninjudd! Apologies for the delay! We haven't dont a good job of posting our releases as of late. This was published under `@langchain/openai==0.6.4`

josiahcoad:
@hntrl  Unfortunately, the same error seems to be present in python

PR #8579, opened 2025-07-28
fix: so we can support string as index params also

Tazer:


hntrl (approved):

hntrl:
Hey @Tazer! Thanks for the PR!

Overall looks good, just some issues with CI that we need to fix before merging

hntrl:
closing in favor of #8592 as we didn't have write permissions to finish this out

PR #8587, opened 2025-07-29
Cb/standard outputs

hntrl:
Fixes # (issue)


langchain-core/src/_standard/content/multimodal.ts
```diff
@@ -1,138 +1,63 @@
-import { ContentBlock } from "./index.js";
+import { ContentBlock as BaseContentBlock } from "./base.js";
 
 // eslint-disable-next-line @typescript-eslint/no-namespace
-export declare namespace MultimodalContentBlocks {
+export declare namespace Multimodal {
+  type DataRecord =
+    | { fileId: string; url?: never; data?: never }
+    | { fileId?: never; url: string; data?: never }
+    | { fileId?: never; url?: never; data: string | Uint8Array };
```
hntrl: possible to add typedoc strings for these?

langchain-core/src/_standard/message.ts
```diff
@@ -10,7 +10,7 @@ export type $MessageType =
 
 export interface $MessageComplex {
   readonly content: {
-    [key: string]: ContentBlock | undefined;
+    [key: string]: unknown;
```
hntrl: dont we want this to always be ContentBlock?

PR #8588, opened 2025-07-30
test(mongodb): use azure openid when available in tests

durran:
This PR updates the MongoDB integration vector store tests to be able to use Azure OpenID, with a fallback to OpenID if the necessary environment variables are not there. This is due to us switching in our own integration test suite. (https://github.com/mongodb-labs/ai-ml-pipeline-testing)

The sister pull request to this is https://github.com/mongodb-labs/ai-ml-pipeline-testing/pull/89 in our other repo.

### DIFF
--- a/libs/langchain-mongodb/package.json
+++ b/libs/langchain-mongodb/package.json
@@ -39,6 +39,7 @@
   },
   "devDependencies": {
     "@jest/globals": "^29.5.0",
+    "@langchain/azure-openai": "workspace:*",
     "@langchain/core": "workspace:*",
     "@langchain/openai": "workspace:*",
     "@langchain/scripts": ">=0.1.0 <0.2.0",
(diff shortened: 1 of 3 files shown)


baileympearson (approved):

libs/langchain-mongodb/package.json
```diff
@@ -39,6 +39,7 @@
   },
   "devDependencies": {
     "@jest/globals": "^29.5.0",
+    "@langchain/azure-openai": "workspace:*",
```
hntrl: We've actually inlined our azure implementation to `@langchain/openai` and we largely don't update that package anymore (we need to do a formal deprecation of it!). Since you're already importing embeddings from `@langchain/openai`, could you remove this dep?

hntrl (changes requested):

hntrl:
Also if you could run `yarn lint && yarn format` from the root to unstick CI that would be much appreciated!

libs/langchain-mongodb/package.json
durran: Dependency removed now.

durran:
> Also if you could run `yarn lint && yarn format` from the root to unstick CI that would be much appreciated!

Done.

hntrl (approved):

PR #8590, opened 2025-07-30
chore(core): centralise import for v1 primitives

christian-bromann:
This patch includes:

- centralize export for new v1 primitives
- add helper functions for detecting message and content block types
- port `StandardContentBlockConverter` to new message format

### DIFF
--- a/langchain-core/src/_standard/content/multimodal.ts
+++ b/langchain-core/src/_standard/content/multimodal.ts
@@ -103,3 +103,23 @@ export declare namespace Multimodal {
 
   export type ContentBlock = Image | Video | Audio | PlainText | File;
 }
+
+export function isMultimodalContentBlock(
+  content_block: unknown
+): content_block is Multimodal.ContentBlock {
+  if (typeof content_block !== "object" || content_block === null) {
+    return false;
+  }
+
+  if (!("type" in content_block) || typeof content_block.type !== "string") {
+    return false;
+  }
+
+  return (
+    content_block.type === "image" ||
+    content_block.type === "video" ||
+    content_block.type === "audio" ||
+    content_block.type === "text-plain" ||
+    content_block.type === "file"
+  );
+}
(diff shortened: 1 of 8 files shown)


langchain-core/src/_standard/message.ts
```diff
@@ -104,3 +104,21 @@ export type SystemMessage<
...
+
+export function isSystemMessage(
+  message: BaseMessage
+): message is SystemMessage {
+  return message.type === "system";
+}
+
+export function isToolMessage(message: BaseMessage): message is ToolMessage {
+  return message.type === "tool";
+}
```
hntrl: think it would be better to type the message params as `message: unknown` and make some more assertions that the thing were testing is _actually_ a message

```suggestion
export function isBaseMessage(message: unknown): message is BaseMessage {
  ...
}

export function isAIMessage(message: unknown): message is AIMessage {
  return isBaseMessage(message) && message.type === "ai";
}

export function isHumanMessage(message: unknown): message is HumanMessage {
  return isBaseMessage(message) && message.type === "user";
}

export function isSystemMessage(
  message: unknown
): message is SystemMessage {
  return isBaseMessage(message) && message.type === "system";
}

export function isToolMessage(message: BaseMessage): message is ToolMessage {
  return isBaseMessage(message) && message.type === "tool";
}
```

langchain-core/src/_standard/content/tools.ts
```diff
@@ -28,6 +28,22 @@ export declare namespace Tools {
     callId: string;
   }
 
+  export interface ToolResult extends BaseContentBlock {
+    /**
+     * Type of the content block
+     */
+    readonly type: "tool_result";
```
hntrl: I think this should be `tool_call_result`? (to match python)

langchain-core/src/_standard/content/multimodal.ts
```diff
@@ -103,3 +103,23 @@ export declare namespace Multimodal {
 
   export type ContentBlock = Image | Video | Audio | PlainText | File;
 }
+
+export function isMultimodalContentBlock(
+  content_block: unknown
+): content_block is Multimodal.ContentBlock {
```
hntrl: nit: camel case

christian-bromann:
@hntrl thanks for the feedback, updated the patch.

hntrl (approved):

PR #8599, opened 2025-08-01
chore(internal): add new package for tsdown build

christian-bromann:
To help review the new build system, this branch was created to separate out the actual build package from the other required changes.


hntrl:
We'll probably want to split these out to individual plugins rather than one big build utility in the near-ish future

PR #8600, opened 2025-08-01
wip(core): standard outputs

hntrl:
Fixes # (issue)


libs/langchain-core/src/_standard/message.ts
```diff
@@ -0,0 +1,512 @@
...
+
+/**
+ * Standard message structured used to define the most basic message structure that's
+ * used throughout the library.
+ *
+ * This is also the message structure that's used when a message structure is not provided.
+ */
+export interface $StandardMessageStructure extends $MessageStructure {
+  /** @internal Discriminator to give TS a hint when evaluating if a type is a standard message structure */
+  __STANDARD_MESSAGE_STRUCTURE: never;
```
dqbd, line 318: Should we use a symbol instead? TypeScript will still discriminate properly but it won't be shown in Intellisense.

Alternatively, would use `~` instead of `__` to avoid bubbling it upwards?

PR #8606, opened 2025-08-01
chore(v1): cleanup indexes and prompts

christian-bromann:
This PR cleans up `indexes` and `prompts`.


christian-bromann:
Let me check in with the Python team first. I will re-raise the PR again.

PR #8608, opened 2025-08-03
chore: rearrange packages

hntrl:


dependency_range_tests/scripts/langchain/test-with-latest-deps.sh
```diff
@@ -17,14 +17,14 @@ cp -r /scripts/langchain/node/!(node_modules|dist|dist-cjs|dist-esm|build|.next|
 
 cd /updater_script
 
-pnpm install --production
+pnpm install
```
christian-bromann: Any reason we need dev dependencies here?
```suggestion
pnpm install --prod
```

hntrl: Will readd, but don't think its super necessary since there's only 1 dep for this install line?

https://github.com/langchain-ai/langchainjs/blob/6dc484bc92e7027edb1f9866d195cf9df2888392/dependency_range_tests/scripts/langchain/node/package.json#L1-L10

PR #8611, opened 2025-08-04
chore(v1): clean up './schema/query_constructor' and './schema/prompt_template' exports

christian-bromann:
This patch cleans up the following exports in `langchain`:

- ./schema/query_constructor
- ./schema/prompt_template

I was not quite sure how to replace `import { formatDocument } from "langchain/schema/prompt_template";` in the example which made me remove the example all together. Any advise on what to use for `formatDocument` instead would be appreciated.


christian-bromann:
Removed in #8624

Issue #8514, opened 2025-07-15
Cannot add requestMetadata to Bedrock Claude Invocation


DarioPanada:
### Checked other resources

- [x] This is a bug, not a usage question. For questions, please use the LangChain Forum (https://forum.langchain.com/).
- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain.js documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain.js rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```
chain.invoke([message], {metadata: {foo: 'bar314'}});
```

### Description

I trying to pass custom metadata to my chain invocation using langchain to interact with AWS Bedrock Claude models (>=3.5).

```
chain.invoke([message], {metadata: {foo: 'bar314'}});
```

`metadata` is supported as per langchain docs (https://v03.api.js.langchain.com/interfaces/_langchain_aws.ChatBedrockConverseCallOptions.html#metadata)

`requesMetadata` is supported on AWS Bedrock SDK calls (https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Converse.html#bedrock-runtime_Converse-request-requestMetadata)

Expectation is that my metadata would be mapped to requestMetadata in the HTTP Request payload.

Observed behaviour is that the metadata is not included in the HTTP Request payload.

Issue appears related to this statement `delete callOptions.metadata;` _separateRunnableConfigFromCallOptions in runnables.base.ts:374 (https://github.com/langchain-ai/langchainjs/blob/main/langchain-core/src/runnables/base.ts#L374C24-L374C32) which deletes metadata in what then becomes the object used to parametrise the command to bedrock.

Manually adding a statement like `callOptions.requestMetadata = callOptions.metadata` before said attribute is deleted seems to fix the issue. Obviously this is not a full solution as not all models would support the requestMetadata attribute, so just adding that statement has potential to break integration with other providers.

### System Info

Node v18.18.2, latest release of langchain

hntrl:
Apologies for the delay @DarioPanada!

`metadata` for a runnable is semantically different than the data that gets passed to a chat model request. The attached PR should let you pass requestMetadata into the config of an invoke call when it gets released:

`chain.invoke([message], { requestMetadata: { foo: 'bar314' } });`

DarioPanada:
[...]

PR #8612, opened 2025-08-04
Fix: Pass user-supplied metadata as requestMetadata to Bedrock API

akintunero:
**Issue**: [#8514](https://github.com/langchain-ai/langchainjs/issues/8514) - `requestMetadata` is not being passed to the Claude Bedrock invocation function.

**Problem**: The `metadata` from call options is being deleted in the base runnable class before it can be passed as `requestMetadata` to the Bedrock API, preventing user-supplied metadata from reaching the AWS Bedrock service.

## Root Cause Analysis

The issue occurs in two places:

1. **`langchain-core/src/runnables/base.ts`** (Line 373):
   ```typescript
   delete callOptions.metadata; // ❌ This prevents metadata from reaching the model
   ```

2. **`libs/langchain-aws/src/chat_models.ts`**:
   The `client.send()` calls don't pass the metadata as `requestMetadata` to the AWS Bedrock API.

### 1. Fix Base Runnable Class

**File**: `langchain-core/src/runnables/base.ts`

**Change**: Remove the line that deletes metadata from callOptions so it can be passed through to the model.

```typescript
// Before (BROKEN)
const callOptions = { ...(options as Partial<CallOptions>) };
delete callOptions.callbacks;
delete callOptions.tags;
delete callOptions.metadata; // ❌ This prevents metadata from reaching the model
delete callOptions.runName;
// ... other deletions

// After (FIXED)
const callOptions = { ...(options as Partial<CallOptions>) };
delete callOptions.callbacks;
delete callOptions.tags;
// delete callOptions.metadata; // ✅ REMOVED - metadata now passes through
delete callOptions.runName;
// ... other deletions
```

### 2. Update Bedrock Implementation

**File**: `libs/langchain-aws/src/chat_models.ts`

**Changes**: Pass `options.metadata` as `requestMetadata` to the `client.send()` method in both streaming and non-streaming methods.

#### Non-streaming method:
```typescript
// Before (BROKEN)
const response = await this.client.send(command, {
  abortSignal: options.signal,
});

// After (FIXED)
const response = await this.client.send(command, {
  abortSignal: options.signal,
  requestMetadata: options.metadata, // ✅ Pass metadata as requestMetadata
});
```

#### Streaming method:
```typescript
// Before (BROKEN)
const response = await this.client.send(command, {
  abortSignal: options.signal,
});

// After (FIXED)
const response = await this.client.send(command, {
  abortSignal: options.signal,
  requestMetadata: options.metadata, // ✅ Pass metadata as requestMetadata
});
```

### Test File: `test_request_metadata_fix.js`

The test demonstrates:
1. **Working Fix**: Metadata is properly passed to Bedrock API
2. **Broken Version**: Metadata is not passed (simulates current behavior)
3. **Verification**: Shows that user-supplied metadata reaches the AWS service

### Expected Behavior

```typescript
// User code
const model = new ChatBedrockConverse({
  model: "anthropic.claude-3-haiku-20240307-v1:0",
  region: "us-west-2",
});

const response = await model.invoke("Hello", {
  metadata: {
    userId: "12345",
    sessionId: "session-67890",
    customField: "test-value"
  }
});

// ✅ Now the metadata reaches the Bedrock API as requestMetadata
```

## Files Modified

1. **`langchain-core/src/runnables/base.ts`**
   - Remove `delete callOptions.metadata;` line
   - Allows metadata to pass through to model implementations

2. **`libs/langchain-aws/src/chat_models.ts`**
   - Add `requestMetadata: options.metadata` to `client.send()` calls
   - Update both `_generateNonStreaming` and `_streamResponseChunks` methods

## Impact

- **Positive**: User-supplied metadata now reaches the AWS Bedrock API
- **Backward Compatible**: No breaking changes to existing APIs
- **Feature Complete**: Works for both streaming and non-streaming calls
- **AWS Compliant**: Follows AWS Bedrock API specification for requestMetadata

## Verification

The fix can be verified by:
1. Running the test file: `node test_request_metadata_fix.js`
2. Checking that metadata appears in AWS CloudTrail logs
3. Verifying that user-supplied metadata reaches the Bedrock service

## Related Issues

- Fixes [#8514](https://github.com/langchain-ai/langchainjs/issues/8514)
- Enables proper metadata handling for AWS Bedrock integration
- Maintains compatibility with other LangChainJS features

### DIFF
--- a/langchain-core/src/runnables/base.ts
+++ b/langchain-core/src/runnables/base.ts
@@ -371,6 +371,7 @@ export abstract class Runnable<
     const callOptions = { ...(options as Partial<CallOptions>) };
     delete callOptions.callbacks;
     delete callOptions.tags;
+    delete callOptions.metadata;
     delete callOptions.runName;
     delete callOptions.configurable;
     delete callOptions.recursionLimit;
--- a/libs/langchain-aws/src/chat_models.ts
+++ b/libs/langchain-aws/src/chat_models.ts
@@ -16,6 +16,7 @@ import type {
   ToolConfiguration,
   GuardrailConfiguration,
   PerformanceConfiguration,
+  ConverseRequest,
 } from "@aws-sdk/client-bedrock-runtime";
 import {
   BedrockRuntimeClient,
@@ -187,6 +188,11 @@ export interface ChatBedrockConverseCallOptions
    * If a tool name is passed, it will force the model to call that specific tool.
    */
   tool_choice?: BedrockConverseToolChoice;
+
+  /**
+   * Key-value pairs that you can use to filter invocation logs.
+   */
+  requestMetadata?: ConverseRequest["requestMetadata"];
 }
 
 /**
@@ -845,11 +851,11 @@ export class ChatBedrockConverse
       modelId: this.model,
       messages: converseMessages,
       system: converseSystem,
+      requestMetadata: options.requestMetadata,
       ...params,
     });
     const response = await this.client.send(command, {
       abortSignal: options.signal,
-      requestMetadata: options.metadata,
     });
     const { output, ...responseMetadata } = response;
     if (!output?.message) {
@@ -886,11 +892,11 @@ export class ChatBedrockConverse
       modelId: this.model,
       messages: converseMessages,
       system: converseSystem,
+      requestMetadata: options.requestMetadata,
       ...params,
     });
     const response = await this.client.send(command, {
       abortSignal: options.signal,
-      requestMetadata: options.metadata,
     });
     if (response.stream) {
       for await (const chunk of response.stream) {


hntrl:
Thanks for the PR @akintunero!

Removing the condition that cleans a runnables metadata seems a little too heavy handed of an approach to fix the issue here. I think a better approach would be to add `requestMetadata` to the call options since `metadata` and `requestMetadata` have semantic differences.

```diff
- chain.invoke([message], {metadata: {foo: 'bar314'}});
+ chain.invoke([message], {requestMetadata: {foo: 'bar314'}});
```

Thoughts?

akintunero:
> Thanks for the PR @akintunero!
> 
> Removing the condition that cleans a runnables metadata seems a little too heavy handed of an approach to fix the issue here. I think a better approach would be to add `requestMetadata` to the call options since `metadata` and `requestMetadata` have semantic differences.
> 
> ```diff
> - chain.invoke([message], {metadata: {foo: 'bar314'}});
> + chain.invoke([message], {requestMetadata: {foo: 'bar314'}});
> ```
> 
> Thoughts?



> Thanks for the PR @akintunero!
> 
> Removing the condition that cleans a runnables metadata seems a little too heavy handed of an approach to fix the issue here. I think a better approach would be to add `requestMetadata` to the call options since `metadata` and `requestMetadata` have semantic differences.
> 
> ```diff
> - chain.invoke([message], {metadata: {foo: 'bar314'}});
> + chain.invoke([message], {requestMetadata: {foo: 'bar314'}});
> ```
> 
> Thoughts?

Thanks for the feedback! You're absolutely right - removing the metadata cleaning condition entirely would be too broad and could have unintended side effects.
The requestMetadata approach makes much more sense semantically. It clearly distinguishes between:
metadata: Persistent metadata that gets attached to the runnable and flows through the entire execution
requestMetadata: Request-specific metadata that should be included in the current invocation
This maintains the existing behavior while providing a clean way to pass request-specific metadata. I'll update the PR to implement this approach instead.

PR #8615, opened 2025-08-04
feat(groq): add `json_schema` support

janzheng:
This PR adds support for Groq's new structured outputs and adds support for `json_schema` for `model.withStructuredOutput(ResponseFormatter)` and 

```
  model: "gpt-4o-mini",
}).bind({
  response_format: { type: "json_object" },
});
```

This also adds a passing integration test for structured outputs in `chat_models.standard.int.test.ts`


libs/langchain-groq/src/tests/chat_models.standard.int.test.ts
```diff
@@ -55,11 +56,62 @@ class ChatGroqStandardIntegrationTests extends ChatModelIntegrationTests<
...
+    // Check raw output contains tool calls
+    if (!raw || typeof raw !== "object") {
+      throw new Error("Raw output should be an object");
+    }
+    
+    // For tool calling, raw should have tool_calls in additional_kwargs
+    const rawMessage = raw as any;
+    if (!rawMessage.additional_kwargs?.tool_calls || rawMessage.additional_kwargs.tool_calls.length === 0) {
+      throw new Error("Raw output should contain tool calls");
+    }
```
hntrl: not totally sure this is necessary? This should be handled by removing the this.skipTestMessage clause since standard tests does this on its own

janzheng: oh! got it i'll remove; wasn't sure about what the standard tests covered

janzheng: removed!

janzheng: also, thoughts on bumping up to kimi k2 for the test models?

hntrl: yeah good idea!

PR #8616, opened 2025-08-04
feat(agents): migrate createReactAgent to core LangChain project

christian-bromann:
As part of our v1 efforts we want to migrate core agent primitives to the core LangChain project. The main primitives ported over were:

- `createReactAgent`
- `withAgentName`
- `ToolNode`
- `RunnableCallable`
- `PreHookAnnotation`

I have ported Python tests from:

- [`test_react_agent.py`](https://github.com/langchain-ai/langgraph/blob/main/libs/prebuilt/tests/test_react_agent.py)
- [`test_tool_node.py`](https://github.com/langchain-ai/langgraph/blob/main/libs/prebuilt/tests/test_tool_node.py)

Coverage:

Screenshot 2025-08-05 at 6 07 04 PM


libs/langchain-agents/README.md
```diff
@@ -0,0 +1,425 @@
...
+    },
+  }),
+  // Validate responses
+  postModelHook: new RunnableLambda({
+    func: (state) => {
+      const lastMessage = state.messages[state.messages.length - 1];
+      console.log("Agent responded:", lastMessage.content);
+      return {};
+    },
+  }),
```
hntrl: think we just want these to be normal functions instead of RunnableLambda

libs/langchain-agents/README.md
```diff
@@ -0,0 +1,425 @@
...
+  AgentState.spec & StateSpec
+>;
+```
+
+A compiled LangGraph StateGraph that supports:
+
+- `.invoke(input, config?)`: Single execution
+- `.stream(input, config?)`: Streaming execution
+- `.astream(input, config?)`: Async streaming
+- All standard LangGraph patterns and configurations
```
hntrl: would rather just point people towards typedoc instead of inlining references

libs/langchain-agents/src/load/import_constants.ts
hntrl, line 1: not sure if we need to include import maps for createReactAgent (we can't serialize a graph runnable)

blocked by componentizing build steps so I'm fine to put this off

libs/langchain-agents/src/types.ts
```diff
@@ -0,0 +1,191 @@
...
+  >
+    ? Schema extends InteropZodType<infer V>
+      ? ReducerSchema extends InteropZodType<infer U>
+        ? BaseChannel<V, U>
+        : never
+      : never
+    : TShape[key] extends InteropZodType<infer V, infer U>
+    ? BaseChannel<V, U>
+    : never;
+};
```
hntrl: are these not exported from langgraph?

hntrl (commented):
thoughts on moving this to `@langchain/react-agent`? then this would live in something like `libs/agents/react-agent` (kind of like how we have for providers)

libs/langchain-agents/src/types.ts
christian-bromann: Good catch 👌

christian-bromann:
Before merging, migrate:

- https://github.com/langchain-ai/langgraphjs/pull/1519

christian-bromann:
Closing in favor of #8789

PR #8619, opened 2025-08-05
RFC(v1): standard outputs + new message types

hntrl:
## RFC: V1 Message Types

As apart of the effort for the first major release of LangChain.js, we're looking to introduce breaking changes to a number of our core primitives. One of these breaking changes is to introduce a new message abstraction that addresses some prolific blind spots in our current message structure, namely things like how:

- **Messages aren't type safe** -- Our current message types lack proper type safety, making it difficult to catch errors at compile time and leading to potential runtime issues.
- **Messages don't conform to a standard structure** -- One of the big value props of LangChain is that it gives you a standard way to interact with AI providers. If you want to use existing message history across providers (i.e. using messages generated by one provider in another), we currently get you about half way there. In order to implement provider specific features, we do some awkward things to make sure that provider specific metadata is attached to the message object. This means that even though we have one message type, we effectively have n number of message shapes.
- **Messages are represented as classes instead of plain objects** -- Using classes adds unnecessary complexity and makes serialization/deserialization more difficult. The fact that you have to import a message class to create a message object is awkward.

### Standard Outputs

The core concept behind our new message system is the introduction of standardized "content blocks" that serve as the source of truth for representing message contents. These content blocks provide a unified way to handle the diverse range of capabilities that AI providers support.

Content blocks are typed objects that represent specific kinds of content within a message. Each content block has a `type` field that identifies its purpose and structure. The standard content blocks include:

- **Text blocks** - Plain text content for basic conversational messages
- **Reasoning blocks** - Chain-of-thought or step-by-step reasoning content
- **Tool call blocks** - Structured representations of function/tool invocations
- **Tool result blocks** - Responses and outputs from executed tools
- **Multimodal blocks** - Support for images, audio, video, and other media types

The idea is that these represent common content types that are supported by a large number of providers. Rather than specifying different shapes of content blocks for each provider, we have a standard definition of what each content block should look like across providers.

### `$MessageStructure`

The need for a "message structure" is a result of the need to support more complex per-provider message types. If a provider uses a special type of content block, or has additional metadata associated with a message, or is expecting tool calls, we need to be able to represent that in a way that is type safe and easy to use.

`BaseMessage` is now a generic type that takes in a `MessageStructure` type. This type, among other things, is used to define the shape of the message. The idea is that for each provider, we can define a `MessageStructure` that is specific to that provider which contains the appropriate properties/ content blocks/ etc. for a given provider's message.

#### Tool Types

Tools are defined through the `tools` property, which maps tool names to their input/output type definitions using `$MessageToolDefinition`. This allows messages to reference and invoke specific tools with proper type checking:

```ts
type MyBaseMessage = BaseMessage<{
  tools: {
    multiply: {
      input: {
        a: number;
        b: number;
      };
      output: number;
    };
  };
}>;
```

```ts
// `tool` messages wont be accepted unless the message structure contains tool definitions
{
  type: "tool",
  toolCallId: "123",
  status: "success",
  // name can only be "multiply" since that is the name of the tool defined in the message structure
  name: "multiply",
  // content is required to be a number since the tool definition specifies that the output is a number
  content: 1,
}
// a new type of content block will also be added to other message types
{
  type: "ai",
  content: [
    {
      type: "tool_call",
      // name can only be "multiply" since that is the name of the tool defined in the message structure
      name: "multiply",
      // args is required to be an object with the same shape as the tool input
      args: { a: 1, b: 2 }
    }
  ]
}
```

#### Content Blocks

For each message type, we can define additional content blocks that are allowed to be used on that message type. This is useful to describe the capabilities of a given provider, or specify some special content blocks that are specific to a given provider.

```ts
type MyBaseMessage = BaseMessage<{
  contentBlocks: {
    // `ai` messages can have either a reasoning block or a multimodal image block, or a custom content block
    ai:
      | ContentBlock.Reasoning
      | ContentBlock.Multimodal.Image
      | MySpecialContentBlock;
    // `human` messages can have a multimodal video block
    human: ContentBlock.Multimodal.Video;
  };
}>;
```

You'll notice that we don't explicitly specify that text content blocks are allowed on messages. This is because we have a "standard" message structure that is used for all messages which defines that text content blocks are allowed on messages, and we "extend" that standard message structure by passing in our own message structure.

#### Custom Message Properties

For each message type, we can also pass in a "custom properties" object that will add additional properties to that message type. This is useful to attach provider specific metadata to a message.

```ts
type MyBaseMessage = BaseMessage<{
  properties: {
    ai: {
      responseMetadata: {};
    };
    human: {
      userMetadata: { userId: string };
    };
  };
}>;
```

```ts
const message: MyBaseMessage = {
  type: "ai",
  content: [{ type: "text", content: "Hello, world!" }],
  responseMetadata: {},
};
```

```ts
const message: MyBaseMessage = {
  type: "human",
  content: [{ type: "text", content: "Hello, world!" }],
  userMetadata: { userId: "123" },
};
```

#### Merging Message Structures

Because of the declarative nature of LangChain, we want to make it easy to extend the message structure in a way that carries message information throughout the chain. This is done by merging message structures together using $MergeMessageStructure.

```ts
// $StandardMessageStructure is the standard message structure that is used for all messages
// We can extend it by passing in our own message structure
// (this is done automatically from `BaseMessage`)
type MyMessageStructure = $MergeMessageStructure<
  $StandardMessageStructure,
  {
    contentBlocks: {
      ai: ContentBlock.Reasoning;
    };
  }
>;
```

This in effect combines the tools, content blocks, and properties of the two message structures into a single message structure that can be used in defining message types.

### Using new message types

Because we can give messages additional type information per provider, we can now carry over stronger message types for the user.

```ts
// langchain-openai/src/chat_models.ts

interface $OpenAIMessageStructure<TTools> {
  tools: TTools;
  contentBlocks: {
    ai: ContentBlock.Reasoning;
  }
  properties: {
    ai: {
      responseMetadata: {
        modelProvider: "openai";
        modelName: string;
        responsesId: string;
      }
    }
  }
}

// (this is illustrative, not actual code)
class ChatOpenAI<TTools> extends BaseChatModel {
  // declaring the message type here allows us to use it outside of the class
  declare Message: BaseMessage<$OpenAIMessageStructure<TTools>>;

  constructor(fields: ChatOpenAIFields<TTools>) {
    super(fields);
  }

  async invoke(messages: BaseMessage[]): Promise<this["Message"]> {
    ...
  }
}

  tools: {
    multiply: ...
  }
})

// I can either use the message type directly (message carries type information)
const message = await model.invoke({
  content: [
    { type: "reasoning", content: "1 + 1 = 2" }
  ]
})

// Or define it according to the model's message type
const message: typeof model.Message = {
  type: "ai",
  content: [
    { type: "reasoning", content: "1 + 1 = 2" }
  ],
  responseMetadata: {
    modelProvider: "openai",
    modelName: "gpt-4o",
    responsesId: "123"
  }
}
```

### What does the upgrade path look like?

With the move to a standard message structure based on type aliases instead of classes, there are two big things to consider:

- **We lose the ability to use constructors to define messages** -- What we're instructing people to do by moving away from classes is moving to a much more verbose way of defining messages.

```diff
- new HumanMessage("hey")
+ { type: "human", content: [{ type: "text", text: "hey" }] }

```

- **`toolCalls`, `invalidToolCalls`, and `reasoning` don't exist as a top-level property on messages** -- Because the source of truth for these items now live within content blocks, and the fact that we can't specify custom getters on generic objects, these fields will no longer be available from the top level.

```diff
// these could be augmented by utility `getToolCalls` and `getReasoning` functions
- message.toolCalls
+ message.content.filter((block) => block.type === "tool_call")
- message.invalidToolCalls
+ message.content.filter((block) => block.type === "invalid_tool_call")
- message.reasoning
+ message.content.find((block) => block.type === "reasoning")
```

#### How do we handle stored messages? (like in a checkpointer?)

The plan here is to keep v0 message classes around as a legacy artifact for now. When we encounter a v0 message object, we'll have a compatibility layer that will convert it to the new message structure.

---

### `TODO:`

- [ ] Add type tests
- [ ] Add compatibility layer for v0 message objects
- [ ] Add messages codemod


afirstenberg:
I am still trying to understand and digest all of this, and a few additional examples may help. But here is a first cut at my thoughts:

> Messages are represented as classes instead of plain objects

The advantage of messages as classes, however, is that the convenience methods can be built right into the class. This makes it easier on developers working with the library.

This is obvious with the example about creating a `HumanMessage` later:

```diff
- new HumanMessage("hey")
+ { type: "human", content: [{ type: "text", text: "hey" }] }
```

We end up exposing a LOT more boilerplate that now becomes essential, makes it a lot more difficult to read the code, and a lot more difficult to write the code with valid type checking.

> moving to a much more verbose way of defining messages

Not just of defining messages, but also getting the output from the `AIMessage`, it sounds like.  As you continue

> the source of truth for these items [tools] now live within content blocks, and the fact that we can't specify custom getters on generic objects

If I follow this correctly (and, I'll be honest - I'm not sure I am), it isn't just the tools. It is all content. So if I want just text content because that's what I'll be sending to my customer, I need to write something like

```typescript
const response: string = message.content
  .filter((block) => block.type === "text")
  .map((block) => block.text)
  .join("");
```

That isn't much of an improvement over what people need to do today to handle a `content` attribute that is either a string or an array when all developers want in some cases is... a string.

And we can't provide a convenience method because we can't provide methods.

> If a provider uses a special type of content block, or has additional metadata associated with a message, or is expecting tool calls, we need to be able to represent that in a way that is type safe and easy to use.

I think the need, however, is for a developer using the library to get the data _out_ in a platform independent way. As someone developing the platform module, this looks great (if complicated).

But if I'm a developer who wants to switch from OpenAI to Anthropic or Gemini or Gemma... it feels like I now need to care more about the underlying structure than I needed to before.

**Summary** (for now):

I like the idea, and I think the standard content blocks are good.

But I worry that we now make it more difficult for developers to _use_ those content blocks and we're exposing a lot more of the underlying structure in platform dependent ways.

If we don't, and if `BaseMessage` can provide some tools to help us do this - that isn't clear to me right now, and I apologize that I didn't get that.

libs/langchain-core/src/_standard/message.ts
```diff
@@ -0,0 +1,517 @@
...
+/**
+ * Represents a message from a tool in a conversation.
+ * Contains the tool's output, status, and metadata about the tool execution.
+ *
+ * @template TStructure - The message structure definition
+ */
+export type ToolMessage<
+  TStructure extends $MessageStructure = $StandardMessageStructure
+> = "tool" extends $MessageStructureTypes<TStructure>
+  ? BaseMessage<TStructure, "tool">
```
dqbd: This is a bit weird I think? I would expect this to work gracefully?

```typescript
type AnyMessage =
  | msg.AIMessage
  | msg.HumanMessage
  | msg.SystemMessage
  | msg.ToolMessage;

const msg: AnyMessage = {
  type: "tool" // <- not really working
}
```

libs/langchain-core/src/_standard/message.ts
```diff
@@ -0,0 +1,517 @@
...
+    TMessageType extends keyof TStructure["contentBlocks"]
+    ? // Apply the structure's defined properties to the inner message definition
+      $ApplyMessageProperties<
+        TStructure,
+        TMessageType,
+        {
+          type: TMessageType;
+          content: Array<
+            // If tools are undefined, use just the content block type
+            TStructure["tools"] extends undefined
```
dqbd: Can we use prettify here? Otherwise we're getting opaque types

hntrl:
I've made some changes to this messages rfc that diverge from the original proposal, so this is being continued in a new PR! (#8669)

PR #8626, opened 2025-08-06
chore(langchain): remove `./smith` export path

christian-bromann:
As part of the effort to clean up the `langchain` exports, this patch removes the `/smith` one. I will check in with the LangSmith team to ensure these aren't still use anywhere.


christian-bromann:
The LangSmith team confirmed that this is deprecated long ago and should be removed.

PR #8627, opened 2025-08-06
chore(textplitters): migrate from Jest to Vitest

christian-bromann:
Migrate `@langchain/textsplitters` to Vitest.


libs/langchain-textsplitters/vitest.config.ts
hntrl, line 1: maybe this config would be better? the one I put into core was adapted from langgraph

https://github.com/langchain-ai/langchainjs/pull/8622/
files#diff-1ba385f4d4a9b36a637b659d4328bd2c611a854a963b353666b449838a424ba0

christian-bromann: Using `env` is a much better approach, thanks! Given that the textsplitters package doesn't have integration test I kept the config simple but will adopt this for the agents package.

PR #8628, opened 2025-08-06
docs(agents): create extensive docs for various createReactAgent features

christian-bromann:
This patch aims to create examples for the following features characterizing `createReactAgent`:

1. Specify Custom System Prompt
2. Explicit Control Over Message Generation
3. Access External Context in Message Generation
4. Access Thread Level State in Message Generation
5. Access Long Term Memory in Message Generation
6. Update Thread Level Before Model Invocation
7. Access External Context in Tools
8. Access Thread Level Context in Tools
9. Access Long Term Memory in Tools
10. Update Thread Level Context in Tools
11. Update Long Term Memory in Tools
12. Update Tools Before Model Call
13. Update Model to Use Before Model Call
14. Post Model Hook

__Note:__ `langchain` dependency tests are failing due to the fact that we haven't published any version for `@langchain/agents`. We still may move the code into `langchain` directly.


examples/src/createReactAgent/accessExternalContext.ts
```diff
@@ -0,0 +1,116 @@
+/**
+ * Access External Context in Message Generation
```
hwchase17: put more concisely:

context can contain runtime parameters, or connections to dbs, etc

this can affect the system prompt

examples/src/createReactAgent/accessExternalContextInTools.ts
```diff
@@ -0,0 +1,162 @@
+/**
+ * Access External Context in Tools
+ *
+ * This enables tools to access external application state, user data, and environmental context when executing their functions, making tool behavior context-aware and personalized.
+ *
+ * Why this is important:
+ * - Contextual Tool Execution:
```
hwchase17: this example i think is a bit fuzzy/not the best

examples/src/createReactAgent/accessThreadLevelState.ts
```diff
@@ -0,0 +1,92 @@
+/**
+ * Access Thread Level State in Message Generation
+ *
+ * This enables the agent to access and utilize the current conversation history and thread-specific state
```
hwchase17: the key idea here is that the "state" of the agent may affect the system prompt/messages. eg the agent goes and fetches some data info, which is stored in state, which can then be injected into system prompt

examples/src/createReactAgent/controlOverMessageGeneration.ts
```diff
@@ -0,0 +1,82 @@
...
+ * - Message Structure Control:
+ *   Define exactly how system, user, and tool messages are formatted
+ * - Dynamic Content Injection:
+ *   Add timestamps, context, or real-time data to any message
+ * - Conversation Flow Management:
+ *   Control how chat history and tool outputs are presented
+ * - Custom Instructions:
+ *   Embed specific reasoning patterns or output formats directly in messages
+ *
+ * Example Scenario:
```
hwchase17: the key part is - why wouldnt you store this in the `messages` key in the state?

the answer should be situations where you want to transiently transofrm the messages list before passing to llm.

this example seems like you would always want that transformation, so its not transient?

a good example: wanting to append a "reminder" message to the end of the message list, to make the llm "remember" key rules and guidelines

examples/src/createReactAgent/accessExternalContextInTools.ts
```diff
@@ -0,0 +1,162 @@
...
+ *   Tools can adapt their behavior based on user identity, preferences, or current application state
+ * - Data Integration:
+ *   Seamlessly incorporates external data sources and APIs into tool functionality
+ * - Security and Personalization:
+ *   Ensures tools operate with appropriate permissions and user-specific context
+ *
+ * Example Scenario:
+ * You're building a travel planning assistant. When the weather tool is called, it not only gets weather data but
```
hwchase17: this is probably fine, but i think looking up user specific data (eg a users purchases, by passing through their user id) is way better

this also ties into auth around tools - super important!

examples/src/createReactAgent/accessLongTermMemory.ts
```diff
@@ -0,0 +1,79 @@
...
+ * persistent memory stores (like vector databases) when generating responses,
+ * enabling recall of past interactions and learned information.
+ *
+ * Why this is important:
+ * - Knowledge Persistence: Retains and recalls important information across sessions, creating a sense of continuity and learning
+ * - Contextual Enrichment: Enhances responses with relevant historical context and previously discussed topics
+ * - Personalized Experience: Builds long-term understanding of user preferences, needs, and interaction patterns
+ *
+ * Example Scenario:
+ * You're building a personal assistant for a busy executive. When they mention "the Johnson project",
```
hwchase17: i think importantly compared to example below, the agent isnt decidding what to look up, we are doing it programmatically. procedural memory (eg parts of the system prompt) may be better example here, rather than semantic search

examples/src/createReactAgent/accessThreadLevelState.ts
```diff
@@ -0,0 +1,92 @@
...
+ * Why this is important:
+ * - Conversational Continuity:
+ *   Maintains coherent dialogue by referencing previous exchanges and building upon established context
+ * - Adaptive Responses:
+ *   Modifies response style and content based on conversation flow and user interaction patterns
+ * - State-Aware Decision Making:
+ *   Makes informed choices about tools and actions based on what has already occurred in the conversation
+ *
+ * Example Scenario:
+ * You're building a coding tutor bot. If the user has been asking about Python basics for several messages,
```
hwchase17: mmmmm no this isnt greeat

we want an example where the agent in a previous tool call or something updated state (NOT messages key, some other key) and that is now pulled into prompt

examples/src/createReactAgent/accessThreadLevelStateInTools.ts
```diff
@@ -0,0 +1,335 @@
...
+ *
+ * Why this is important:
+ * - Conversation-Aware Tools:
+ *   Tools can reference previous interactions and maintain consistency with the ongoing conversation
+ * - Context-Dependent Actions:
+ *   Tool behavior can vary based on what has been discussed or accomplished earlier in the thread
+ * - Intelligent Workflow Management:
+ *   Enables tools to coordinate with each other and avoid redundant actions
+ *
+ * Example Scenario:
```
hwchase17, line 14: maybe better example: using a sub agent as a tool. you want to pass all message to sub agent which is inside tool

examples/src/createReactAgent/customSystemPrompts.ts
```diff
@@ -0,0 +1,76 @@
...
+ *   Ensures the agent consistently follows your specific guidelines and maintains
+ *   the desired tone and approach
+ * - Domain Expertise:
+ *   Allows you to inject specialized knowledge and constraints relevant to your
+ *   specific use case
+ * - Dynamic Adaptation:
+ *   Enable context-aware behavior changes based on conversation state, user type,
+ *   or external conditions
+ *
+ * **Example Scenario:**
```
hwchase17: eh, i think just formatting information about the user that is passed in in context or something is fine here. i dont actually think this example is great?

examples/src/createReactAgent/updateModelBeforeCall.ts
```diff
@@ -0,0 +1,557 @@
...
+ * This capability enables dynamic model selection based on context, allowing the agent to choose the most
+ * appropriate language model for each specific task or situation.
+ *
+ * Why this is important:
+ * - Cost Optimization: Uses smaller, cheaper models for simple tasks and reserves powerful models for complex reasoning
+ * - Performance Matching: Selects models with capabilities that match the specific requirements of each request
+ * - Specialized Routing: Directs different types of queries to models that excel in those particular domains
+ *
+ * Example Scenario:
+ * You're building a coding assistant that handles both simple syntax questions and complex algorithm design.
```
hwchase17, line 24: question: we can access state, long term memory, and context before we update the model to use, right?

examples/src/createReactAgent/accessThreadLevelStateInTools.ts
christian-bromann: Updated to:

> A supervisor agent has a `delegate_to_subagent` tool. When the user asks for a conversation-wide summary or action items, the supervisor calls the tool, which passes the full message history plus a task to the sub-agent. The sub-agent returns a polished result back to the supervisor.

examples/src/createReactAgent/updateModelBeforeCall.ts
christian-bromann: That is correct!

christian-bromann:
@hwchase17 thanks for the feedback, I've update the examples accordingly.

PR #8632, opened 2025-08-06
test(anthropic): migrate to Vitest

christian-bromann:
Migrating `@langchain/anthropic` to Vitest.


libs/providers/langchain-anthropic/package.json
```diff
@@ -53,13 +51,11 @@
     "eslint-plugin-jest": "^27.6.0",
```
hntrl:
```suggestion
```

libs/providers/langchain-anthropic/src/tests/chat_models.standard.int.test.ts
```diff
@@ -44,8 +43,4 @@ class ChatAnthropicStandardIntegrationTests extends ChatModelIntegrationTests<
 }
 
 const testClass = new ChatAnthropicStandardIntegrationTests();
-
-test("ChatAnthropicStandardIntegrationTests", async () => {
-  const testResults = await testClass.runTests();
-  expect(testResults).toBe(true);
-});
+testClass.runTests("ChatAnthropicStandardIntegrationTests");
```
hntrl, line 46: standard tests doesn't enclose test suites inside of a test block, so I'm pretty sure this will cause false positives

libs/providers/langchain-anthropic/src/tests/chat_models.standard.test.ts
```diff
@@ -32,8 +31,4 @@ class ChatAnthropicStandardUnitTests extends ChatModelUnitTests<
 }
 
 const testClass = new ChatAnthropicStandardUnitTests();
-
-test("ChatAnthropicStandardUnitTests", () => {
-  const testResults = testClass.runTests();
-  expect(testResults).toBe(true);
-});
+testClass.runTests("ChatAnthropicStandardUnitTests");
```
hntrl, line 34: same here

libs/providers/langchain-anthropic/vitest.config.ts
```diff
@@ -0,0 +1,40 @@
...
+    } satisfies UserConfigExport;
+  }
+
+  return {
+    test: {
+      ...common.test,
+      include: configDefaults.include,
+      typecheck: { enabled: true },
+    },
+  } satisfies UserConfigExport;
```
hntrl:
```suggestion
  const common = {
    test: {
      environment: "node",
      hideSkippedTests: true,
      // TODO: remove this once standard tests are updated to use vitest
      globals: true
      testTimeout: 30_000,
      maxWorkers: 0.5,
      exclude: ["**/*.int.test.ts", ...configDefaults.exclude],
      setupFiles: ["dotenv/config"],
    },
  };

  if (env.mode === "standard-unit") {
    return {
      test: {
        ...common.test,
        testTimeout: 100_000,
        exclude: configDefaults.exclude,
        include: ["**/*.standard.test.ts"],
        name: "standard-unit",
        environment: "node",
      },
    };
  }

  if (env.mode === "standard-int") {
    return {
      test: {
        ...common.test,
        testTimeout: 100_000,
        exclude: configDefaults.exclude,
        include: ["**/*.standard.int.test.ts"],
        name: "standard-int",
        environment: "node",
      },
    };
  }

  if (env.mode === "int") {
    return {
      test: {
        ...common.test,
        globals: false,
        testTimeout: 100_000,
        exclude: configDefaults.exclude,
        include: ["**/*.int.test.ts"],
        name: "int",
        environment: "node",
      },
    };
  }

  return {
    test: {
      ...common.test,
      environment: "node",
      include: configDefaults.include,
      typecheck: { enabled: true },
    },
  };
```

libs/providers/langchain-anthropic/package.json
```diff
@@ -20,12 +20,12 @@
...
-    "test:single": "NODE_OPTIONS=--experimental-vm-modules pnpm run jest --config jest.config.cjs --testTimeout 100000",
-    "test:int": "NODE_OPTIONS=--experimental-vm-modules jest --testPathPattern=\\.int\\.test.ts --testTimeout 100000 --maxWorkers=50%",
-    "test:standard:unit": "NODE_OPTIONS=--experimental-vm-modules jest --testPathPattern=\\.standard\\.test.ts --testTimeout 100000 --maxWorkers=50%",
-    "test:standard:int": "NODE_OPTIONS=--experimental-vm-modules jest --testPathPattern=\\.standard\\.int\\.test.ts --testTimeout 100000 --maxWorkers=50%",
+    "test": "vitest --run",
+    "test:watch": "vitest --watch",
+    "test:single": "vitest --run",
+    "test:int": "vitest --run --mode int",
+    "test:standard:unit": "vitest --run",
+    "test:standard:int": "vitest --run --mode int",
```
hntrl:
```suggestion
    "test": "vitest run",
    "test:watch": "vitest",
    "test:int": "vitest run --mode int",
    "test:standard:unit": "vitest run --mode standard-unit",
    "test:standard:int": "vitest run --mode standard-int",
```

libs/providers/langchain-anthropic/src/tests/chat_models.standard.int.test.ts
christian-bromann: > standard tests doesn't enclose test suites inside of a test block, so I'm pretty sure this will cause false positives

@hntrl the Vitest export of this package does see https://github.com/langchain-ai/langchainjs/blob/4fcd60595f19a59a846e237285d6fba1d8ebfc6a/libs/langchain-standard-tests/src/unit_tests/vitest.ts#L27-L48

The reason for this is that previous behavior would log an error but would throw away stack traces and diffs.

hntrl (approved):

PR #8645, opened 2025-08-07
feat(openai): custom tools

hntrl:


alexdlugo:
I see you added the verbosity param, but it's not actually being used. Is that a miss or is it implemented in a later PR?

PR #8649, opened 2025-08-08
feat(openai): add prompt cache key support

hntrl:


libs/langchain-openai/src/chat_models.ts
```diff
@@ -2621,6 +2639,7 @@ export class ChatOpenAICompletions<
             clientOptions
           );
         } else {
+          console.log("request", request);
```
nicogon, line 2642: 👀

hntrl: yeah this was a rookie mistake. Fixed in `@langchain/openai@0.6.7` 🤕

nicogon: Rookies use debugger, real programmers use console.log

PR #8654, opened 2025-08-08
openai/chat_models: remove console.log line

haraldschilly:
I'm pretty sure this console.log is a left over from development. This PR removes that line.


hntrl:
whoops! Thanks for the catch @haraldschilly

PR #8665, opened 2025-08-11
fix(weaviate) Hybrid improvements and tests

dudanogueira:
- allow hybrid to optionally receive the vector, using langchain configured vectorizer if no vectors passed
- pass HybridOptions for hybrid, allowing fine tuned configs
- allow running tests locally export WEAVIATE_URL=local
- adds test and fix metadata keys being returned

Fixes # (issue) No issue filled.
This will improve hybrid by allowing both vector and text to be passed. Also, it allows some hybrid options, such as `alpha`, `fusionType`, `maxVectorDistance` and `queryProperties`
`


hntrl (approved):
thanks for the PR @dudanogueira!

PR #8669, opened 2025-08-11
RFC(v1): new message structure

hntrl:
## RFC: V1 Messages

As apart of langchain.js v1, we're looking to introduce some breaking changes to our core message types to address some prolific blind spots in our current message structure, namely things like how:

- **Messages aren't type safe** -- Our current message types lack proper type safety with regards to tools and the actual content outputted by a chat model.
- **Messages don't conform to a standard structure** -- One of the big value props of LangChain is that it gives you a standard way to interact with AI providers. If you want to use existing message history across providers (i.e. using messages generated by one provider in another), we currently get you about half way there. In order to implement provider specific features, we do some awkward things to make sure that provider specific metadata is attached to the message object. This means that even though we have one message type, we effectively have n number of message shapes.
- **You have to represent messages as classes** -- This means that you can't specify a LangChain message without directly instantiating from a class. We currently don't give you an easy way to ducktype message objects in code to make sure that a generic object is usable as a message (this becomes especially awkward when trying to ensure type safety between backend + frontend)

### New Inheritance

The headline change with these new message objects is that we use a simpler inheritance structure:

* Currently, the common message objects (`AIMessage`, `HumanMessage`, etc.) were all classes that **extend** a `BaseMessage` class
* Now, we export new common message objects that **implement** a standard `Message` interface.

```ts
function useMessage(msg: Message) {}

// works because the object implements `Message`
useMessage({
  type: "ai",
  content: [{ type: "text", text: "hello" }]
});
// works because `AIMessage` implements `Message`
useMessage(new AIMessage("hello"));
```

We opted to keep the common message classes to minimize the number of codepaths that would change and keep the developer conveniences for working with messages, but the actual object that represents the "message data model" stays pure since it's just an interface.

### Standard Outputs

The core concept behind our new message system is the introduction of standardized "content blocks" that serve as the source of truth for representing message contents. Content blocks are typed objects that represent specific kinds of content within a message. Each content block has a `type` field that discriminates on what the block is describing, and as apart of the core package we export a number of content block types that are well understood to be transferable between providers.

The idea is that these represent common content types that are supported by a large number of providers. Rather than specifying different shapes of content blocks for each provider, we have a standard definition of what each content block should look like across providers.

**What does this mean?** This means the shape of `message.content` is changing. While this may seem like a subtle change, it has pretty sweeping implications for existing message handling and thus is a breaking change.

* `message.content` can no longer be a string. Text only messages must use text content blocks
  * with message classes, you can still use a string to create a message with a text content block
    * `new AIMessage("foo")` means message.content is `[{ type: "text", text: "foo" }]`
  * we still provide a `.text` getter on message classes as a shorthand to parse content
* `message.content` must conform to a standard set of content blocks
  * this may mean subtle differences in how existing messages are serialized
* Top level properties are being reworked to make `message.content` the single source of truth
  * This means attributes like `text` and `toolCalls` are now getters that parse and combine the data that lives inside of `message.content`.

### `$MessageStructure`

The need for a "message structure" is a result of the need to support more complex per-provider message types. If a provider uses a special type of content block, or has additional metadata associated with a message, or is expecting tool calls, we need to be able to represent that in a way that is type safe and easy to use.

`Message` is now a generic interface that takes in a `MessageStructure` as a type argument. This type, among other things, is used to define the shape of the message. The idea is that for each provider, we can define a `MessageStructure` that is specific to that provider which contains the appropriate properties/ content blocks/ etc. for a given provider's message.

#### Tool Types

Tools are defined through the `tools` property, which maps tool names to their input/output type definitions using `$MessageToolDefinition`. This allows messages to reference and invoke specific tools with proper type checking:

```ts
type MyMessage = Message<{
  tools: {
    multiply: {
      input: {
        a: number;
        b: number;
      };
      output: number;
    };
  };
}>;
```

```ts
// `tool` messages wont be accepted unless the message structure contains tool definitions
{
  type: "tool",
  toolCallId: "123",
  status: "success",
  // name can only be "multiply" since that is the name of the tool defined in the message structure
  name: "multiply",
  // content is required to be a number since the tool definition specifies that the output is a number
  content: 1,
}
// a new type of content block will also be added to other message types
{
  type: "ai",
  content: [
    {
      type: "tool_call",
      // name can only be "multiply" since that is the name of the tool defined in the message structure
      name: "multiply",
      // args is required to be an object with the same shape as the tool input
      args: { a: 1, b: 2 }
    }
  ]
}
```

#### Content Blocks

For each message type, we can define additional content blocks that are allowed to be used on that message type. This is useful to describe the capabilities of a given provider, or specify some special content blocks that are specific to a given provider.

```ts
type MyMessage = Message<{
  content: {
    // `ai` messages can have either a reasoning block or a multimodal image block, or a custom content block
    ai:
      | ContentBlock.Reasoning
      | ContentBlock.Multimodal.Image
      | MySpecialContentBlock;
    // `human` messages can have a multimodal video block
    human: ContentBlock.Multimodal.Video;
  };
}>;
```

You'll notice that we don't explicitly specify that text content blocks are allowed on messages. This is because we have a "standard" message structure that is used for all messages which defines that text content blocks are allowed on messages, and we "extend" that standard message structure by passing in our own message structure.

#### Merging Message Structures

Because of the declarative nature of LangChain, we want to make it easy to extend the message structure in a way that carries message information throughout the chain. This is done by merging message structures together using $MergeMessageStructure.

```ts
// $StandardMessageStructure is the standard message structure that is used for all messages
// We can extend it by passing in our own message structure
// (this is done automatically from `BaseMessage`)
type MyMessageStructure = $MergeMessageStructure<
  $StandardMessageStructure,
  {
    content: {
      ai: ContentBlock.Reasoning;
    };
  }
>;
```

This in effect combines the tools, content blocks, and properties of the two message structures into a single message structure that can be used in defining message types.

### Using new message types

Because we can give messages additional type information per provider, we can now carry over stronger message types for the user.

```ts
// langchain-openai/src/chat_models.ts

interface $OpenAIMessageStructure<TTools> {
  tools: TTools;
  content: {
    ai: ContentBlock.Reasoning;
  }
  properties: {
    ai: {
      responseMetadata: {
        modelProvider: "openai";
        modelName: string;
        responsesId: string;
      }
    }
  }
}

// (this is illustrative, not actual code)
class ChatOpenAI<TTools> extends BaseChatModel {
  // declaring the message type here allows us to use it outside of the class
  declare Message: Message<$OpenAIMessageStructure<TTools>>;

  constructor(fields: ChatOpenAIFields<TTools>) {
    super(fields);
  }

  async invoke(messages: BaseMessage[]): Promise<AIMessage<$OpenAIMessageStructure<TTools>> {
    ...
  }
}

  tools: {
    multiply: ...
  }
})

// I can either use the message type directly (message carries type information)
const message = await model.invoke({
  content: [
    { type: "reasoning", content: "1 + 1 = 2" }
  ]
})

// Or define it according to the model's message type
const message: typeof model.Message = {
  type: "ai",
  content: [
    { type: "reasoning", content: "1 + 1 = 2" }
  ],
  responseMetadata: {
    modelProvider: "openai",
    modelName: "gpt-4o",
    responsesId: "123"
  }
}
``

### CODE REVIEW

File: libs/langchain-core/src/_standard/message.ts [hunk 1]
```diff
@@ -0,0 +1,1119 @@
...
+ * // Access the combined text content
+ * console.log(aiMessageWithContent.text); // "Here's the answer: 42"
+ * ```
+ */
+export class AIMessage<
+  TStructure extends $MessageStructure = $StandardMessageStructure
+> implements Message<TStructure, "ai">, $BrandedMessage
+{
+  /** @internal */
+  [__BRANDED_MESSAGE]: never;
```
dqbd: _nit_: I think we should try naming this internal property differently so that it's either searchable or explainable at first glance, say if I do this:

```typescript
const system: SystemMessage = {
  type: "system",
  id: "123",
  content: [{ type: "text", text: "hello" }],
  text: "hello",
};
```

I'm getting the following error, which is a bit confusing:

```
Property '[__BRANDED_MESSAGE]' is missing in type '{ type: "system"; id: string; content: { type: "text"; text: string; }[]; text: string; }' but required in type 'SystemMessage<$StandardMessageStructure>'.
```

libs/langchain-core/src/_standard/message.ts
```diff
@@ -0,0 +1,1119 @@
...
+  /**
+   * Creates a system message with simple text content.
+   * @param text - The text content for the message
+   */
+  constructor(text: string);
+  /**
+   * Creates a system message with structured content blocks.
+   * @param content - Array of content blocks for the message
+   */
+  constructor(content: Array<$InferMessageContent<TStructure, "system">>);
```
dqbd: _nit_: Do we want to have this overload? B/c of that we're now suggesting every method from `Array` in IntelliSense

libs/langchain-core/src/_standard/message.ts
```diff
@@ -0,0 +1,1119 @@
...
+  TRole extends $MessageType = $MessageType
+> {
+  /** The message type/role, determines the content structure and available properties */
+  readonly type: TRole;
+  /** Unique identifier for this message */
+  id: string;
+  /** Optional name/identifier for the entity that created this message */
+  name?: string;
+  /** Array of content blocks that make up the message content, typed based on the structure and role */
+  content: Array<$InferMessageContent<TStructure, TRole>>;
```
dqbd: _nit_: should we apply Prettify util here to unwrap the generic types?

```typescript
type Prettify<T> = {
  [K in keyof T]: T[K];
} & {};
```

libs/langchain-core/src/_standard/message.ts
```diff
@@ -0,0 +1,1119 @@
...
+ *     {
+ *       type: "tool_call",
+ *       name: "search",
+ *       args: { query: "What is the capital of France?" }
+ *     }
+ *   ]
+ * };
+ * ```
+ */
+export interface Message<
```
dqbd: Can we massage the `Message` type more to do an union on `TRole`? ie: `Message = Message<TStructure, "human"> | Message<TStructure, "ai">`

Then I think this will work properly
```typescript

interface MyStructure extends $MessageStructure {
  content: {
    human: ContentBlock.Text | ContentBlock.Multimodal.Image;
    ai: ContentBlock.Text | ContentBlock.Reasoning;
  };
}


const message: Message<MyStructure> = new AIMessage(...)

if (message.type === "ai") {
  const parts = message.content.find(c => c.type === "reasoning")
  // Expected                        ^? c = ContentBlock.Text | ContentBlock.Reasoning
  // Actual                             c = ContentBlock.Text | ContentBlock.Reasoning | ContentBlock.Multimodal.Image
}
```

libs/langchain-core/src/_standard/message.ts
```diff
@@ -0,0 +1,1119 @@
...
+   * Excludes the "tool" message type since tool messages have a special structure.
+   * Each message type can specify what content block types it supports (text, images, etc).
+   */
+  readonly content?: Partial<{
+    [key in $MessageType]: ContentBlock;
+  }>;
+  /**
+   * Optional mapping of message types to arbitrary property objects.
+   * Allows attaching custom metadata or other information to specific message types.
+   * Properties are typed as Record<string, unknown> to allow flexible property definitions.
```
dqbd: _nit_: can we rephrase? I think we currently only allow modifying properties that are actually present in corresponding message types, eg. this won't work:

```typescript
type S = {
  properties: {
    ai: { random: string }
  }
}

const m = new AIMessage<S>({ random: "hello" })
```

libs/langchain-core/src/_standard/message.ts
hntrl: talked offline: no we cant. New info is this is also largely irrelevant since we have a constraint to make the `content` type the same for v1. The new message structure content types will be accessible by `.contentBlocks` instead

libs/langchain-core/src/_standard/message.ts
hntrl: don't think this is a perfect name either, but am going to opt for `__MESSAGE_SYMBOL` since its marginally better (if you have other ideas lmk)

christian-bromann:
🎉

PR #8671, opened 2025-08-11
feat(agents): support AbortSignal in createReactAgent

christian-bromann:
## Description

This PR adds support for `AbortSignal` in the `createReactAgent` function, allowing users to cancel agent execution and tool invocations. This is particularly useful for long-running agent operations where users may want to cancel execution based on external events or user interaction.

### Key Changes:

1. **Added `abortSignal` parameter** to `CreateReactAgentParams` - Users can now pass an `AbortSignal` when creating an agent
2. **Signal propagation to tools** - The abort signal is passed to the `ToolNode` and properly merged with any existing signals from LangGraph's Pregel class
3. **Signal merging with `AbortSignal.any()`** - When both agent-level and config-level signals are present, they are merged using `AbortSignal.any()` (with appropriate TypeScript declarations for this newer API)
4. **Improved error propagation** - Abort signal reasons are now properly propagated throughout the execution chain
5. **Comprehensive test coverage** - Added tests to verify abort functionality works correctly at both the agent and tool levels

### Usage Example:

```typescript
const abortController = new AbortController();

const agent = createReactAgent({
  llm: model,
  tools: [myTool],
  abortSignal: abortController.signal,
});

// Start agent execution
const executionPromise = agent.invoke({
  messages: [{ role: "user", content: "Do something long running" }],
});

// Cancel after some condition
setTimeout(() => {
  abortController.abort("User cancelled operation");
}, 5000);
```

### Additional Improvements:

- Fixed build script to not clean in watch mode (prevents IDE issues with missing types)
- Updated signal utilities in langchain-core to properly extract and propagate abort reasons
- Added TypeScript declarations for `AbortSignal.any()` which is not yet in TypeScript's lib

Fixes #(issue)

### DIFF
--- a/internal/build/index.ts
+++ b/internal/build/index.ts
@@ -46,10 +46,12 @@ async function buildProject(
   const sourcemap = !opts.skipSourcemap;
 
   /**
-   * don't clean if user passes `--skipClean` or if `--noEmit` is enabled
-   * as we don't want to clean previous builds if we're not emitting anything
+   * don't clean if we:
+   * - user passes `--skipClean` or
+   * - have watch mode enabled (it would confuse the IDE due to missing type for a short moment)
+   * - if `--noEmit` is enabled (we don't want to clean previous builds if we're not emitting anything)
    */
-  const clean = !opts.skipClean && !opts.noEmit;
+  const clean = !opts.skipClean && !watch && !opts.noEmit;
 
   /**
    * generate type declarations if not disabled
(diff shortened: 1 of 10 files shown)


hntrl:
I'd make the argument that an `AbortSignal` is better served as a call option then it is a "constructor" option.

christian-bromann:
> I'd make the argument that an `AbortSignal` is better served as a call option then it is a "constructor" option.

We should support both, a global abort signal that acts as default and one as `invoke` option that allows to overwrite the global one.

PR #8693, opened 2025-08-14
docs(langchain): Update configurableFields object key reference description

zMattew:
The referenced api_key and base_url aren't the right name of the key that will be used when executing initChatModel but apiKey and baseUrl. This change will give more clarity of key name to use when want to configure this fields.

Fixes # 
changed configurableFields name reference to what it will use in the langchainjs package


hntrl:
Thanks for the PR @zMattew!

PR #8697, opened 2025-08-15
Add RecallIO.AI memory integration

RecallIO:
Add RecallIO.AI as a memory Integration.
A key is needed from https://app.recallio.ai (free available).

Already available on Langchain (python).

### DIFF
--- a/docs/core_docs/docs/integrations/memory/recallio_memory.mdx
+++ b/docs/core_docs/docs/integrations/memory/recallio_memory.mdx
@@ -1,0 +1,27 @@
+---
+hide_table_of_contents: true
+---
+
+import CodeBlock from "@theme/CodeBlock";
+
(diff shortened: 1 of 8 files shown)


hntrl:
Hi @RecallIO 

The community package has become rather unwieldy, and we recommend against using it in production applications.
Is this something that you want people to be able to use in production? If so, I'd recommend repackaging your changes into a standalone NPM module. You can keep this in your own GitHub repo and publish it to NPM yourself without our involvement, but to make it discoverable you can still create a PR that adds a page about it in our integration docs. We'd be happy to support this, but the maintenance burden of adding another package to langchain-community is not something we're looking to take on at the minute.

RecallIO:
Understood. We will create a NPM and come back with another PR just for the doc. Thanks for the quick feedbacks

PR #8706, opened 2025-08-16
feat(community): Add GPT-5(Openai) support for forwarding verbosity parameter

clovisrodriguez:
- Adds a model-level verbosity option to ChatOpenAI and includes it as text.verbosity when using the OpenAI Responses API.
- Preserves any user-provided options.text; only augments it with verbosity if not already present.
- Continues forwarding reasoning to the Responses API;

### DIFF
--- a/libs/langchain-openai/src/chat_models.ts
+++ b/libs/langchain-openai/src/chat_models.ts
@@ -607,6 +607,11 @@ export abstract class BaseChatOpenAI<
   reasoning?: OpenAIClient.Reasoning;
 
   /**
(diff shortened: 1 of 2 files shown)


hntrl:
Thanks for the PR @clovisrodriguez!

I added a different implementation for this before coming across this PR in #8730. Going to roll with that one since it supports completions and responses (since both APIs support verbosity). Apologies again!

clovisrodriguez:
Yep no worries! thanks for looking into that

PR #8708, opened 2025-08-16
fix(langchain): make `langsmith` an optional dependency

christian-bromann:
We should not list `langsmith` as direct dependency for `langchain` given that it is mainly used within the hub module which doesn't seem to be relevant for most `langchain` package users.


jacoblee93:
Yeah I think should be fine since it's in core too.

PR #8709, opened 2025-08-16
fix(langchain): remove @langchain/openai as package dependency

christian-bromann:
A handful of modules in the `langchain` package were dependent on `@langchain/openai`. This patch cleans this up and removes its dependency from the `langchain` package. I think we don't want these primitives within the v1 package. If they are still useful to people , I would suggest to move them into the provider package instead.

It seems that there are [users out there using this](https://sourcegraph.com/search?q=context%3Aglobal+%22langchain%2Fchains%2Fopenai_functions%22&patternType=keyword&sm=0&df=%5B%22lang%22%2C%22TypeScript%22%2C%22lang%3Atypescript%22%5D&__cc=1). I will double check if this is still the recommended way to approach agents given we are investing into `createReactAgent`. Alternatively we can move this into the provider package instead.

### DIFF
--- a/libs/langchain/package.json
+++ b/libs/langchain/package.json
@@ -36,6 +36,7 @@
     "@langchain/cohere": "workspace:*",
     "@langchain/core": "workspace:*",
     "@langchain/eslint": "workspace:*",
+    "@langchain/openai": "workspace:*",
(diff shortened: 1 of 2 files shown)


christian-bromann:
duplicate of #8920

PR #8711, opened 2025-08-17
feat(core): Allow runnables to define custom stream chunk concat behavior

jacoblee93:
Useful for tracing but also potentially for longer `transform` pipelines

Before:

https://smith.langchain.com/public/4d6e0040-38bf-49ab-8c76-2493e62a2e02/r

After:

https://smith.langchain.com/public/9a027e20-315c-441a-b05d-e087a45f5757/r

### DIFF
--- a/langchain-core/src/runnables/base.ts
+++ b/langchain-core/src/runnables/base.ts
@@ -482,6 +482,7 @@ export abstract class Runnable<
     return outputs;
   }
 
+  /** @internal */
   _concatOutputChunks<O>(first: O, second: O): O {
     return concat(first, second);
   }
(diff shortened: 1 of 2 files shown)


langchain-core/src/runnables/base.ts
```diff
@@ -482,6 +482,10 @@ export abstract class Runnable<
     return outputs;
   }
 
+  _concatOutputChunks<O>(first: O, second: O): O {
+    return concat(first, second);
+  }
```
christian-bromann, line 488: Is there a need to bind this method to the `Runnable` class? This will trickle down and will be exposed to the user which will bloat e.g. auto-suggestion box etc. I wonder if we can implement this as utility on the module level rather than making it class method 🤔

jacoblee93: Could make it protected?

jacoblee93: Oh, no that causes issues since I'm delegating to e.g. `this.last._concatOutputChunks`

But will think of another way since yeah it's not ideal

jacoblee93: Leaving as `@internal` for now

PR #8712, opened 2025-08-17
fix(langchain): remove `js-tiktoken` dependency

christian-bromann:
The `langchain` package had a dependency to `js-tiktoken` for a basic type import. This is not necessary and we can clean this up.


hntrl:
cc @dqbd

dqbd (approved):

PR #8719, opened 2025-08-18
fix: add support for conversion of ref in array schema

jnjacobson:
This PR fixes the conversion of $ref items in array schemas and adds this case to the existing test.

I also added a test for the conversion of refs that are not in an array schema.

Example:

```yaml
openapi: 3.1.0
info:
  title: A fake spec for testing
  version: 0.0.1
paths:
  /foo:
    post:
      operationId: createFoo
      description: Create multiple foo
      parameters:
        - name: fooArray
          in: query
          schema:
            type: array
            items:
              $ref: '#/components/schemas/Foo'
      responses:
        '201':
          description: Created
          content:
            application/json:
              schema:
                properties:
                  success:
                    type: boolean
components:
  schemas:
    Foo:
      type: object
      properties:
        bar:
          type: string
```

Converting the `fooArray` schema would result in

```json
{
  "type": "array",
  "items": {
    "type": "string"
  }
}
```

After the fix it gets converted into:

```json
{
  "type": "array",
  "items": {
    "type": "object",
    "properties": {
      "bar": {
        "type": "string"
      }
    },
    "required": [],
    "additionalProperties": {}
  }
}
```

### DIFF
--- a/.changeset/puny-poems-sink.md
+++ b/.changeset/puny-poems-sink.md
@@ -1,0 +1,5 @@
+---
+"langchain": patch
+---
+
+add support for conversion of ref in array schema
--- a/langchain/src/chains/openai_functions/openapi.ts
+++ b/langchain/src/chains/openai_functions/openapi.ts
@@ -193,9 +193,11 @@ export function convertOpenAPISchemaToJSONSchema(
     );
   }
   if (schema.type === "array") {
+    const openAPIItems = spec.getSchema(schema.items) ?? {};
+
     return {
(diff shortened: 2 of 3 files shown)


jnjacobson:
Hey @hntrl, ist there something I need to change or is this ready to be merged?

hntrl:
@jnjacobson Nothing needed from you! If I remember right CI was failing right after I approved this so I was waiting for it to self correct. Will get this merged.

Thanks for the PR!

PR #8720, opened 2025-08-18
fix(community): postgres indexes getTime returning NaN due to missing alias

jeremylevy:
Hello,

When running the SQL query included in the `getTime()` function on PostgreSQL 13.18, I got the following result:

```csv
date_part
1755534822.917985
```

This causes the function to return `NaN` since the `extract` property does not exist.

When the `getTime` function returns `NaN`, both the `incremental` and `full` cleanup operations remove all added documents, rendering the indexing feature unusable:

```js
{ numAdded: 1, numDeleted: 1, numUpdated: 0, numSkipped: 0 }
```

This PR fixes the issue by always adding an alias using the `AS` SQL keyword.

### DIFF
--- a/.changeset/hot-bugs-clap.md
+++ b/.changeset/hot-bugs-clap.md
@@ -1,0 +1,5 @@
+---
+"@langchain/community": patch
+---
+
+postgres indexes getTime returning NaN due to missing alias
--- a/libs/langchain-community/src/indexes/postgres.ts
+++ b/libs/langchain-community/src/indexes/postgres.ts
@@ -69,7 +69,7 @@ export class PostgresRecordManager implements RecordManagerInterface {
 
   async getTime(): Promise<number> {
     const res = await this.pool.query(
-      "SELECT EXTRACT(EPOCH FROM CURRENT_TIMESTAMP)"
+      "SELECT EXTRACT(EPOCH FROM CURRENT_TIMESTAMP) AS extract"
     );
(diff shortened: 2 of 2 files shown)


hntrl:
@jeremylevy thanks for the PR!

Is this a regression between postgres versions? (i.e. was there once an additional extract property that came back from that query?) If so I think we'd want to have guards that cover both cases.

jeremylevy:
Hey @hntrl,

The query always returns only one property, but it was rewritten to `date_part` in older versions of PostgreSQL. See, for example:
https://stackoverflow.com/questions/38442340/difference-between-extractyear-from-timestamp-function-and-date-partyear-t

This behavior changed in PostgreSQL 14: http://postgresql.org/docs/release/14.0/

```
Change EXTRACT() to return type numeric instead of float8 (Peter Eisentraut)
This avoids loss-of-precision issues in some usages. The old behavior can still be obtained by using the old underlying function date_part().
Also, EXTRACT(date) now throws an error for units that are not part of the date data type.
```

When running the same query on PostgreSQL 14, it returns the expected `extract` property name:

```csv
      extract      
-------------------
 1755551254.164705
(1 row)
```

Adding an alias does not change this behavior:

```sql
SELECT EXTRACT(EPOCH FROM CURRENT_TIMESTAMP) AS extract;
```

```csv
      extract      
-------------------
 1755551254.164705
(1 row)
```

hntrl (approved):

PR #8722, opened 2025-08-18
feat: make listCollections & createCollection available for Chroma Vector Store

Johanneslueke:
I had a usecase where i needed to know the different available collections in the chromaDB. because the client has this method i would like to forward this to the vector store 🙂

### DIFF
--- a/libs/langchain-community/src/vectorstores/chroma.ts
+++ b/libs/langchain-community/src/vectorstores/chroma.ts
@@ -262,7 +262,7 @@ export class Chroma extends VectorStore {
         });
       }
 
-      return this.index.createCollection({ name, configuration, metadata, embeddingFunction);
+      return this.index.createCollection({ name, configuration, metadata, embeddingFunction });
   }
 
   /**


hntrl:
@Johanneslueke I saw in the linked PR you mentioned that this was motivated by the need to supply a different embeddings function. Is that the motivator behind this PR?

Johanneslueke:
My primary objective is to query available collections and create new ones. This Pull Request (PR) is similar to, but independent of, the one you referenced, but I also discovered that if ensureCollection is called no embedding is provided which causes error messages during its development."

hntrl:
In talking with the chroma folks (thanks @itaismith) I don't think this is something we want to enable for the langchain chroma integraion. The mental model is that a `Chroma` instance is 1:1 with a collection, so to enable a way to introspect other collections breaks the semantic boundary.

You can supply your own client to call these methods in the constructor if you still want to be able to do this

```ts
import { ChromaClient } from "chromadb";
import { Chroma } from "@langchain/community/vectorstores/chroma";

const index = new ChromaClient(...);
const store = new Chroma(embeddings, { index, ... });

index.createCollection(...);
index.listCollections(...);
```

Johanneslueke:
Is this documented somewhere and i just missed it? I was not aware of this 1:1 relationship. i do not mind it and if true makes absolut sense to have one store per collection. But it would be nice to make this clear in the documentation. 🙂

hntrl:
@Johanneslueke Yeah I think the reason this isn't abundantly clear already is because the original vector stores integrated within langchain don't have a conceptual definition of a "collection," so we didn't want to make it first class within our vector store abstraction. (it makes it a little awkward when your entrypoint into langchain vectorstores is a provider that has collections, so intuition is to consider them one in the same)

Agree that we can make this clearer in documentation (and would love a PR on this front!)

PR #8737, opened 2025-08-19
Fix for Anthropic Model Tool Call Chunk Generation Error

### LINKED ISSUE #8691: Unable to use anthropic's web search tool
### Checked other resources

### Example Code

```typescript
      model: 'claude-4-sonnet-20250514',
      maxTokens: 4096,
      temperature: 0,
    });

    const agent = createReactAgent({
      llm,
      tools: [
        {
          type: 'web_search_20250305',
          name: 'web_search',
          max_uses: 3,
        },
      ],
      name: 'Web Search Agent',
      prompt: system,
    });

      { messages: query },
      { streamMode: ['messages'], callbacks: [agentCallbacks] },
    );
```

 400 {"type":"error","error":{"type":"invalid_request_error","message":"messages.4.content.0: unexpected `tool_use_id` found in `tool_result` blocks: fallback-0. Each `tool_result` block must have a corresponding `tool_use` block in the previous message."}}

### Description

This is because when Claude uses web search, it returns a message like this
```
{
type: 'content_block_delta',
index: 1,
delta: { type: 'input_json_delta', partial_json: '"}' }
}
```

In langchain, this will result in a tool call: 

```
  } else if (
    data.type === "content_block_delta" &&
    data.delta.type === "input_json_delta"
  ) {
    return {
      chunk: new AIMessageChunk({
        content: fields.coerceContentToString
          ? ""
          : [
              {
                index: data.index,
                input: data.delta.partial_json,
                type: data.delta.type,
              },
            ],
        additional_kwargs: {},
        tool_call_chunks: [
          {
            index: data.index,
            args: data.delta.partial_json,
          },
        ],
      }),
    };
  }
```

```
const groupedToolCallChunk = fields.tool_call_chunks.reduce(
        (acc, chunk) => {
          // Assign a fallback ID if the chunk doesn't have one
          // This can happen with tools that have empty schemas
          const chunkId = chunk.id || `fallback-${chunk.index || 0}`;
          acc[chunkId] = acc[chunkId] ?? [];
          acc[chunkId].push(chunk);
          return acc;
        },
        {} as Record<string, ToolCallChunk[]>
      );

      const toolCalls: ToolCall[] = [];
      const invalidToolCalls: InvalidToolCall[] = [];
      for (const [id, chunks] of Object.entries(groupedToolCallChunk)) {
        let parsedArgs = {};
        const name = chunks[0]?.name ?? "";
        const joinedArgs = chunks.map((c) => c.args || "").join("");
        const argsStr = joinedArgs.length ? joinedArgs : "{}";
        // Use the original ID from the first chunk if it exists, otherwise use the grouped ID
        const originalId = chunks[0]?.id || id;
        try {
          parsedArgs = parsePartialJson(argsStr);
          if (
            parsedArgs === null ||
            typeof parsedArgs !== "object" ||
            Array.isArray(parsedArgs)
          ) {
            throw new Error("Malformed tool call chunk args.");
          }
          toolCalls.push({
            name,
            args: parsedArgs,
            id: originalId,
            type: "tool_call",
          });
```

Finally, langgraphql thinks that tool call exists:

```
const { messages } = state;
        const lastMessage = messages[messages.length - 1];

        // if there's no function call, we finish
        if (!isAIMessage(lastMessage) || !lastMessage.tool_calls?.length) {
          if (responseFormat != null) return "generate_structured_response";
          return END;
        }

        // there are function calls, we continue
        if (version === "v2") {
          return lastMessage.tool_calls.map(
            (toolCall) =>
              new Send("tools", { ...state, lg_tool_call: toolCall })
          );
        }

        return "tools";
```

This finally led to the error

### System Info

.

mikey247:
Fixes  #8691

🐛 Problem Description

LangChain’s Anthropic integration was incorrectly generating tool call chunks for all input_json_delta events during streaming, regardless of whether the JSON content represented an actual tool call.

This caused errors such as:

`400 {"type":"error","error":{"type":"invalid_request_error", "message":"messages.4.content.0: unexpected ‘tool_use_id‘ found in‘tool_result‘ blocks : fallback-0. Each ‘tool_result‘ block must have a corresponding ‘tool_use‘ block in the previous message."}}`

Root Cause

In makeMessageChunkFromAnthropicEvent, any streaming JSON content was treated as a tool call, resulting in tool call chunks with fallback IDs—even when the JSON was just regular response content.

// Original buggy logic
```
} else if (
  data.type === "content_block_delta" &&
  data.delta.type === "input_json_delta"
) {
  return {
    chunk: new AIMessageChunk({
      // ... content handling ...
      tool_call_chunks: [
        {
          index: data.index,
          args: data.delta.partial_json,
        },
      ],
    }),
  };
}
```

✅ Solution

The fix introduces a validation step:

JSON is parsed and only objects containing a name property (Anthropic’s marker for tool calls) are treated as tool calls.

Non-tool JSON is handled as regular content.

// Fixed code
```
} else if (
  data.type === "content_block_delta" &&
  data.delta.type === "input_json_delta"
) {
  let toolCallChunks = [];
  const additionalKwargs: Record<string, any> = {};

  try {
    const parsed = JSON.parse(data.delta.partial_json);
    if (parsed && typeof parsed === "object" && "name" in parsed) {
      toolCallChunks = [
        {
          index: data.index,
          args: data.delta.partial_json,
        },
      ];
    }
  } catch (err) {
    if (process.env.NODE_ENV !== "production") {
      console.warn(
        "Invalid partial JSON from Anthropic:",
        data.delta.partial_json
      );
    }
    additionalKwargs.raw_partial_json = data.delta.partial_json;
  }

  return {
    chunk: new AIMessageChunk({
      // ... content handling ...
      additional_kwargs: additionalKwargs,
      tool_call_chunks: toolCallChunks,
    }),
  };
}
```

✨ Key Improvements

- Precise tool call detection → Only JSON objects with a name property are treated as tool calls.
- Error resilience → Invalid JSON no longer produces invalid tool calls.
- Debugging support → Raw partial JSON is preserved in additional_kwargs when parsing fails.
- Backward compatibility → Legitimate tool calls continue to work without changes.

📈 Impact
This fix resolves the tool use ID mismatch error when using Anthropic models with LangChain agents—particularly in scenarios involving web search tools or other streaming JSON responses that are not tool calls.

Error Output
Screenshot 2025-08-19 at 08 35 13

Fixed Output
Screenshot 2025-08-19 at 20 32 26


mikey247:
@hntrl

christian-bromann (commented):
Thanks @mikey247 for raising the PR. Do you think this change can be accompanied with a test to ensure we don't regress here?

mikey247:
> Thanks @mikey247 for raising the PR. Do you think this change can be accompanied with a test to ensure we don't regress here?

Sorry I'm just seeing this, will look into it.

mikey247:
> Thanks @mikey247 for raising the PR. Do you think this change can be accompanied with a test to ensure we don't regress here?

Are there any guidelines to testing PRs?
@christian-bromann

hntrl:
@mikey247 Apologies for the delay in getting to this, we've been busy with our v1 efforts.

Reading through this PR, i think it pre supposes that `input_json_delta` chunks will always arrive from anthropic as aggregate segments from previous chunks, which unless something has changed, I don't believe is the case.

e.g. we're assuming that anthropic streams like this
```
event: input_json_delta
data: '{ "na'

event: input_json_delta
data: '{ "name": { "a":'

event: input_json_delta
data: '{ "name": { "a": 12 } }'
```

but in actuality it streams like this:
```
event: input_json_delta
data: '{ "na'

event: input_json_delta
data: 'me": { "a":'

event: input_json_delta
data: ' 12 } }'
```

Which means that any input_json_delta event that we encounter won't be asserted as a tool call chunk since it is never a complete json that has a `name` property.
(note that these aren't actual anthropic events, it's just illustrative of what `partial_json` looks like)

---

With that in mind, I think the solution is actually what @lzj960515 proposed as a workaround in the issue. When we go to serialize the messages, we look back to the bound tools, and if the tool call chunk is for a server tool, we don't add it as a `tool_use` field. The assumption is that we're always calling the same model that has the same bound which I think is a safe assertion to make. Otherwise we have to pin additional metadata to the tool message which I imagine is a breaking change. wdyt?

Totally get that it's not ideal to come back to this after a while. If you're busy shipping other stuff we're happy to take this on :)

christian-bromann:
Closing as the original issue is not reproducible anymore.

PR #8742, opened 2025-08-20
wip: v1 chat model

hntrl:
some highlights from this iteration of the `BaseChatModel` class:

* `BaseChatModel` now inherits from `Runnable` instead of `BaseLanguageModel`
  * this is to simplify the inheritance chain and use `Message` objects as the output of the "chat model runnable" instead of the legacy `Generation` object
* adds a new callback manager that more closely models the new chat model inheritance structure
  * (e.g. chat models call `handleChatModelX` instead of `handleLLMX`)
  * also uses `Message` instead of `Generation` objects in events
* adds an alternative implementation of `BaseCache` that uses `Message` objects instead of `Generation` objects
  * we technically are storing an array of `Message`'s to later enable replaying stream chunks
* instead of returning a `RunnableBinding` when calling `withStructuredOutput`, `bindTools`, or `withConfig`, we instead create a new instance of the current class by calling `this.constructor` and modifying an internal `defaultOptions` property that gets reused
  * for `withStructuredOutput` specifically, we're retaining runnable semantics by pinning the existing runnable chain to a `.outputParser` property that gets used in `invoke` and `_streamIterator` 
  * fixes a long standing issue where you couldn't call `withStructuredOutput` and `bindTools` together
* changes some of the abstract methods to be a little bit clearer for downstream implementations
  * `_generate` -> `generate`
  * `generate` -> `invoke`
  * `_streamResponseChunks` -> `streamResponseChunks`
* gets rid of the deprecated methods that existed on `BaseLanguageModel`


libs/langchain-core/src/callbacks/base.ts
```diff
@@ -115,12 +115,12 @@ abstract class BaseCallbackHandlerMethodsClass {
...
 
   /**
-   * Called at the start of a Chat Model run, with the prompt(s)
+   * Called at the start of a chat model run, with the messages
    * and the run ID.
    */
   handleChatModelStart?(
     llm: Serialized,
-    messages: BaseMessage[][],
+    messages: Message[],
```
hntrl, line 123: the inner workings of `BaseLanguageModel` lets you parse multiple prompts at once, whereas the new chat model class only takes one string of messages

libs/langchain-core/src/language_models/tests/v1.test.ts
```diff
@@ -0,0 +1,135 @@
...
+
+  withStructuredOutput<Output>(
+    schema: InteropZodType<Output> | JSONSchema,
+    config?: StructuredOutputMethodOptions<boolean>
+  ):
+    | SimpleChatModel<Runnable<Message, Output>>
+    | SimpleChatModel<Runnable<Message, { raw: Message; parsed: Output }>> {
+    return super.withStructuredOutput(schema, config) as
+      | SimpleChatModel<Runnable<Message, Output>>
+      | SimpleChatModel<Runnable<Message, { raw: Message; parsed: Output }>>;
```
hntrl, line 78: unfortunately I'm not sure if there's a clean solution to avoid this verbose typing

libs/langchain-core/src/language_models/tests/v1.test.ts
```diff
@@ -0,0 +1,135 @@
...
+      .bindTools([makeToolDefinition("tool2")])
+      .withConfig({
+        tool_choice: "any",
+      })
+      .withStructuredOutput({
+        type: "object",
+        properties: {},
+        required: [],
+      })
+      .bindTools([makeToolDefinition("tool3")]);
```
hntrl, line 133: would expect that this composition means:
* we keep tools 1-3
* `tool_choice` will be any
* we only use the second withStructuredOutput

libs/langchain-core/src/language_models/v1.ts
```diff
@@ -0,0 +1,674 @@
...
+    this.cache = iife(() => {
+      if (typeof params.cache === "object") {
+        return params.cache;
+      } else if (params.cache) {
+        return InMemoryCacheV1.global();
+      }
+      return undefined;
+    });
+    this.caller = new AsyncCaller(params ?? {});
+  }
```
hntrl, line 237: this constructor is largely just taking after `BaseLangChain`

libs/langchain-core/src/language_models/v1.ts
```diff
@@ -0,0 +1,674 @@
...
+}
+
+export type BaseChatModelInput =
+  | BasePromptValueInterface
+  | MessageLike[]
+  | Exclude<MessageLike, MessageTuple>;
+
+export abstract class BaseChatModelV1<
+    CallOptions extends BaseChatModelCallOptions = BaseChatModelCallOptions,
+    TOutputParser extends ChatModelOutputParser | undefined = undefined
```
hntrl, line 135: TOutputParser actually determines whats on the output end of the runnable -- if we want structured output out we use whats defined on the output of the passed in parser, if its undefined we use `Message`

hntrl: this has a naming conflict with non-chat model output parsers, so am open to ideas on what to name this if we stick w/ this

libs/langchain-core/src/language_models/v1.ts
```diff
@@ -0,0 +1,674 @@
...
+      )) {
+        aggregated = aggregated ? concat(aggregated, chunk) : chunk;
+        if (this.outputParser === undefined) {
+          yield chunk as InferChatModelOutput<TOutputParser>;
+        } else {
+          yield this.outputParser.invoke(
+            chunk,
+            callOptions
+          ) as InferChatModelOutput<TOutputParser>;
+        }
```
hntrl, line 357: unclear on what the semantics should be for parsing chunks
* run the output parser on each chunk (what this is currently doing)
* only run the output parser on the aggregated chunk

libs/langchain-core/src/language_models/v1.ts
```diff
@@ -0,0 +1,674 @@
...
+        return (toolCall as any).args as Output;
+      }
+    );
+
+    if (!config?.includeRaw) {
+      return instance.withOutputParser(
+        toolMessageParser.withConfig({
+          runName: "StructuredOutput",
+        })
+      );
```
hntrl, line 609: instead of returning a runnable, we return a reconstructed instance that has a different outputParser

libs/langchain-core/src/language_models/v1.ts
```diff
@@ -0,0 +1,674 @@
...
+}
+
+export function convertChatModelInput(input: BaseChatModelInput): Message[] {
+  const messages: Message[] = [];
+
+  // // BasePromptValueInterface
+  // if (isPromptValue(input)) {
+  //   // TODO: Implement
+  //   throw new Error("not implemented");
+  // }
```
hntrl, line 646: will readd when prompts accept new message types

PR #8747, opened 2025-08-20
Update import_constants.ts

Kennytian:
Two embedding models have been added

Fixes # (issue)


Kennytian:
@hntrl  Please review again, thanks

Kennytian:
@bracesproul @jacoblee93 Please review it, thanks. I need to use text-embedding-v4 😁

PR #8749, opened 2025-08-21
BM25Retriever: escape regex metacharacters in getTermFrequency to prevent crashes

### LINKED ISSUE #8746: BM25Retriever crashes on queries with markdown / regex characters (unescaped getTermFrequency regex)
### Checked other resources

### Example Code

```typescript
import { getTermFrequency } from "./node_modules/@langchain/community/dist/utils/@furkantoprak/bm25/BM25.js";

const term   = "**Version 1:**";
const corpus = "**Version 1:** What is the country of origin for the person in question?";

await getTermFrequency(term, corpus);   // 💥
```

```
SyntaxError: Invalid regular expression: /**Version 1:**/g: Nothing to repeat
    at new RegExp (<anonymous>)
    at getTermFrequency (node_modules/@langchain/community/dist/utils/@furkantoprak/bm25/BM25.js:13:34)
    ...
```

### Description

- I’m using `@langchain/community` BM25 utilities via `BM25Retriever`.
- For any term containing RegExp metacharacters (`*`, `+`, `?`, `|`, `[`, `]`, etc.), `getTermFrequency(term, corpus)` constructs an unsafe `RegExp(term, "g")`.
- This crashes with `SyntaxError: Invalid regular expression … Nothing to repeat`.

Root cause: the term is not escaped before supplying it to the `RegExp` constructor.

A safe implementation might do:
```typescript
const safeTermFrequency = (term: string, corpus: string) => {
    const escaped = term.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
    return ((corpus || "").match(new RegExp(escaped, "g")) || []).length;
}
```

**Additional context**
The crash shows up immediately in normal downstream usage because `BM25Retriever` is used by higher-level utilities such as `MultiQueryRetriever`.
When `queryCount > 1`, LangChain asks an LLM to paraphrase the user query; those paraphrases may contain Markdown or other non-escaped characters. For example, here’s an excerpt from an actual trace when MultiQueryRetriever was called:

```
<questions>
**Version 1:** What is the country of origin for the person in question?
**Version 2:** From which nation does this individual originate?
**Version 3:** What nationality does the person hold?
</questions>
```

Because each generated sub-query is fed straight into `getTermFrequency`, any line that includes `**`, `[`, `|`, `?`, etc. will throw the `Invalid regular expression … Nothing to repeat` error, and the whole retrieval pipeline fails.
So even if a user never calls `getTermFrequency` directly, any code path that involves `MultiQueryRetriever + BM25Retriever` (for hybrid search, ensemble retrieval, etc.) is vulnerable.

### System Info

langchain@0.3.30
@langchain/community@0.3.50

OS         : Windows 11
Node     : v22.16.0

erudzitis:
Fixes #8746 

**Problem:** 
unescaped RegExp(term) throws on `**`, `[`, `?`, etc., crashing BM25 based retrieval functionality

**Fix:** 
escape metacharacters via `term.replace(/[.*+?^${}()|[\\]\\\\]/g, "\\\\$&")` before constructing `RegExp`

### DIFF
--- a/libs/langchain-community/src/retrievers/tests/bm25.test.ts
+++ b/libs/langchain-community/src/retrievers/tests/bm25.test.ts
@@ -25,3 +25,24 @@ test("BM25Retriever", async () => {
     "The quick brown fox jumps over the lazy dog"
   );
 });
+
+test("getTermFrequency escapes regex metacharacters", async () => {
+  // Import directly from the inlined BM25 util to test the helper.
+  const { getTermFrequency } = await import(
+    "../../utils/@furkantoprak/bm25/BM25.js"
+  );
+
+  const corpus = "**Version 1:** What is the country of origin for the person in question?";
+  const term = "**Version 1:**";
+
+  // Should not throw and should find at least one match
+  const freq = getTermFrequency(term, corpus);
+  expect(freq).toBeGreaterThanOrEqual(1);
+
+  // Also test other metacharacters
+  const corpus2 = "Does this match (maybe)? [yes] *stars* +plus+";
+  expect(getTermFrequency("(maybe)?", corpus2)).toBeGreaterThanOrEqual(1);
+  expect(getTermFrequency("[yes]", corpus2)).toBeGreaterThanOrEqual(1);
+  expect(getTermFrequency("*stars*", corpus2)).toBeGreaterThanOrEqual(1);
+  expect(getTermFrequency("+plus+", corpus2)).toBeGreaterThanOrEqual(1);
+});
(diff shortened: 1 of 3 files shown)


libs/langchain-community/src/retrievers/tests/bm25.test.ts
```diff
@@ -25,3 +25,24 @@ test("BM25Retriever", async () => {
     "The quick brown fox jumps over the lazy dog"
   );
 });
+
+test("getTermFrequency escapes regex metacharacters", async () => {
+  // Import directly from the inlined BM25 util to test the helper.
+  const { getTermFrequency } = await import(
+    "../../utils/@furkantoprak/bm25/BM25.js"
+  );
```
christian-bromann: Why not have a:
```ts
import { getTermFrequency } from "../../utils/@furkantoprak/bm25/BM25.js";
```

at the top?

erudzitis: @christian-bromann I've made the changes, check again, thanks

PR #8751, opened 2025-08-22
support vastbase database as vectorstore

yayaxxww:
Add @langchain/vastbase Package


christian-bromann:
@yayaxxww thank you for raising the PR. We are currently developing on a v1 release and can't pay much attention to new integrations. Please allow us some time before we can actively review new provide packages. Thank you!

hntrl:
Jumping in here --

Is this something that you want people to be able to use in production? If so, I'd recommend repackaging your changes into a standalone NPM module. You can keep this in your own GitHub repo and publish it to NPM yourself without our involvement, but to make it discoverable you can still create a PR that adds a page about it in our integration docs.

PR #8768, opened 2025-08-23
[DRAFT] Refactor for consistent results when chaining bindTools and withConfig

patinthehat:
***DRAFT, DO NOT MERGE***

This pull request implements and tests order independence for the `bindTools` and `withConfig` methods on chat models, ensuring that applying these methods in any order yields the same result. The main changes involve refactoring how method operations are queued and applied within the `ConfigurableModel` class, and adding a new test to verify the correct behavior.

Currently, the calling of `bindTools` and `withConfig` on chat models depends on the order in which they are chained. This makes the API error-prone and unintuitive, since two logically equivalent sequences of calls can yield different results. For example, using `.withConfig(...).bindTools(...)` may behave differently than using `.bindTools(...).withConfig(...)`.

**Key changes:**

### Order independence for configuration and tool binding

* Refactored `ConfigurableModel` to store queued method operations as an ordered array (`QueuedMethodOperation[]`) instead of an object, ensuring that methods like `bindTools` and `withConfig` are applied in the correct order regardless of how they're chained. [[1]](diffhunk://#diff-df80dc5eb66c1678f818acaa0bea4d41ea5d495f822de6b714a6bceff9144f72R226-R231) [[2]](diffhunk://#diff-df80dc5eb66c1678f818acaa0bea4d41ea5d495f822de6b714a6bceff9144f72L239-R247) [[3]](diffhunk://#diff-df80dc5eb66c1678f818acaa0bea4d41ea5d495f822de6b714a6bceff9144f72L275-R282) [[4]](diffhunk://#diff-df80dc5eb66c1678f818acaa0bea4d41ea5d495f822de6b714a6bceff9144f72L313-R319)
* Updated the implementation of `bindTools` and `withStructuredOutput` to append their operations to the queue, returning a new model instance with the updated queue. [[1]](diffhunk://#diff-df80dc5eb66c1678f818acaa0bea4d41ea5d495f822de6b714a6bceff9144f72L335-R356) [[2]](diffhunk://#diff-df80dc5eb66c1678f818acaa0bea4d41ea5d495f822de6b714a6bceff9144f72L361-R373) [[3]](diffhunk://#diff-df80dc5eb66c1678f818acaa0bea4d41ea5d495f822de6b714a6bceff9144f72L429-R436)

### Testing

* Added a new test file, `bind_tools.test.ts`, which introduces a `FakeChatModelWithBindTools` and verifies that chaining `bindTools` and `withConfig` in any order produces identical results. This test demonstrates the correctness of the new queuing mechanism.

### Minor code cleanup

* Minor import and formatting adjustments in `chat_models.test.ts` for consistency.


patinthehat:
ping @jacoblee93 @bracesproul @nfcampos

jacoblee93:
I think this makes sense - CC @hntrl as I think we're working on making something like this universal behavior in v1!

patinthehat:
Is it possible to get one or more reviews, as well as an approval for the workflow runs please?
ping @hntrl @bracesproul @christian-bromann

langchain/src/chat_models/universal.ts
```diff
@@ -332,24 +333,27 @@ export class ConfigurableModel<
 
   async _generate(
     messages: BaseMessage[],
-    options?: this["ParsedCallOptions"],
+    options?: CallOptions,
```
hntrl, line 336: keep as the same
```suggestion
    options?: this["ParsedCallOptions"],
```

langchain-core/src/language_models/tests/chat_models.test.ts
```diff
@@ -1,14 +1,13 @@
 /* eslint-disable no-promise-executor-return */
 
-import { test, expect } from "@jest/globals";
-import { z } from "zod/v3";
+import { expect, test } from "@jest/globals";
+import { z, z as z4 } from "zod";
```
hntrl, line 4: we need to keep the explicit import from /v3 here
```suggestion
import { z } from "zod/v3";
```

langchain-core/src/language_models/tests/bind_tools.test.ts
hntrl, line 1: this reads as only touching net new code for the purposes of the test, so my vote would be to remove

hntrl (commented):
Thanks for the PR and apologies for the delay @patinthehat!

Left some comments to address before we merge.

PR #8772, opened 2025-08-23
internal(eslint): resolve eslint warnings and errors

christian-bromann:
The last patch fixes a lot of Eslint problems we had in our project. The most common ones are the following:

- removed unused error object in `try/catch` block
- removed unused eslint ignore comments
- enforce templating instead of string concatenation


dependency_range_tests/scripts/with_standard_tests/google-vertexai/test-with-lowest-deps.sh
```diff
@@ -35,4 +35,9 @@ pnpm install --no-frozen-lockfile
 # not try to build the package/its workspace dependencies.
 cd "$monorepo_vertexai_dir"
 
+# Install @langchain/google-gauth dependency explicitly
+# This is needed because the update script converts the workspace dependency
+# to use the minimum version, but in the test environment we need to install
+# it as a regular dependency since workspace packages aren't available
+pnpm add @langchain/google-gauth
 pnpm test
```
hntrl: I think this will lead to a bunch of false positives? We purposefully want to use the lowest version of the package for -gauth published on the registry

hntrl (approved):

dependency_range_tests/scripts/with_standard_tests/google-vertexai/test-with-lowest-deps.sh
christian-bromann: yep, will revert this as I improved environment tests in a different PR.

PR #8775, opened 2025-08-24
chore(azure-openai): remove deprecated package

christian-bromann:
All primitives of this package are deprecated. Let's remove the entire package.


hntrl:
Added the deprecation notice on https://npmjs.com/package/@langchain/azure-openai also

PR #8777, opened 2025-08-24
Test: Ensure chat model binding/config order does not affect results

chadgratts:
This PR adds and tests logic for chat model binding and configuration order in the core package. Specifically, it ensures that the order in which .withConfig() and .bindTools() are called on a chat model does not affect the resulting behavior or output.

**Key changes:**

- Added a new test: 
`chat_model_binding.test.ts`
This test verifies that binding tools before or after configuration produces the same result for chat models.

- Introduced a `FakeChatModelWithBindTools` class in the test to simulate tool binding and configuration.

**Motivation:**
This test guards against regressions where the order of configuration and tool binding could lead to inconsistent model behavior, improving reliability and maintainability.


hntrl:
Closing in favor of #8768

PR #8781, opened 2025-08-25
feat(google-common): add handling for well known content blocks

hntrl:
cc @afirstenberg

It's a little awkward to add openai format to google handling, but we have some upstream issues with how we format audio in some places so we need to add the case here. I don't expect this to stay long term since we're rectifying a lot of this with standard outputs.

### DIFF
--- a/libs/langchain-google-common/src/utils/gemini.ts
+++ b/libs/langchain-google-common/src/utils/gemini.ts
@@ -315,6 +315,58 @@ export function normalizeSpeechConfig(
   return ret;
 }
 
+// Compatibility layer for other well known content block types
+export function normalizeMessageContentComplex(
+  content: MessageContentComplex[]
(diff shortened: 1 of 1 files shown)


libs/langchain-google-common/src/utils/gemini.ts
```diff
@@ -315,6 +315,58 @@ export function normalizeSpeechConfig(
   return ret;
 }
 
+// Compatibility layer for other well known content block types
+export function normalizeMessageContentComplex(
```
christian-bromann, line 319: Is it worth having some extensive unit tests for this or will this be replaced with the new message standard in v1?

hntrl: we don't have a great way to test these input currently. All things to work on with consolidating packages 😁

PR #8783, opened 2025-08-25
feat(core): infer tool name from function

christian-bromann:
We can infer the tool name directly from the function allowing users to simplify a tool initiation to just:

```ts
const myTool = tool(function myTool() => {
  return "some result
})
```


hntrl (commented):
Also thought of doing this, but I don't think we can always assert that bundlers/transpilers will keep the function name intact. I'd rather that people explicitly specify it instead of sending them on a bundler goose chase

christian-bromann:
While we can't guarantee it, we have a runtime check in place that verifies that a tool name is provided. How about we mention this in the parameter documentation?

PR #8787, opened 2025-08-26
chore(openai): reorganize input parsing

hntrl:


libs/providers/langchain-openai/src/index.ts
```diff
@@ -6,7 +6,14 @@ export * from "./azure/llms.js";
...
 export * from "./utils/azure.js";
 export * from "./tools/index.js";
 export { convertPromptToOpenAI } from "./utils/prompts.js";
+
+// These methods are used in LangSmith, export is important here
+// TODO: put this conversion elsewhere
+export {
+  _convertMessagesToOpenAIParams,
+  messageToOpenAIRole,
+} from "./utils/message_inputs.js";
```
hntrl, line 19: cc @jacoblee93

this export used to be in `chat_models.ts` but it's being moved here. there won't be any noticeable difference until evals packages ([and friends](https://github.com/search?q=org%3Alangchain-ai%20_convertMessagesToOpenAIParams&type=code)) go to v1 `@langchain/openai`, but this shouldn't impact anything since it only gets imported from the index export (which we're reapplying here)

jacoblee93: Sounds good

hntrl:
work to unstick ci will continue in base branch

PR #8804, opened 2025-08-28
Bump @qdrant/js-client-rest version to 1.15.1

Kennytian:


hntrl:
Hey @Kennytian! You'll also need to run `yarn install` from the root to bump the lockfile

Kennytian:
> Hey @Kennytian! You'll also need to run `yarn install` from the root to bump the lockfile

Ok, thank you, i understand, I will do it lalter

PR #8814, opened 2025-08-29
feat: add personalAccessToken to jira loader (as in confluence loader)

raphaelblum:
### Description

This is my first PR to LangChain.js 🙂  
It adds support for authenticating the `JiraProjectLoader` with a `personalAccessToken`.  
When provided, it will be used as a Bearer token in the `Authorization` header, similar to the Confluence loader.  
This change is backwards compatible with existing authentication methods.


christian-bromann:
@raphaelblum thanks for raising this PR. Our team is currently focusing working on an alpha release and won't be able to merge this as quickly as it usually would. We hope we will be able to review and eventually merge this within the next 2 weeks or so. Thank you!

PR #8817, opened 2025-08-29
fix(langchain): setup message reducer when zod object is provided

christian-bromann:
This patch improves the state management when users using Zod in the following ways (compared to the prebuilt):

- a `message` property is not required to be set anymore, it's done automatically for the user, allowing them to focus on what they are interested in rather than having to do something like:
  ```ts
  const stateSchema = z.object({
    // custom user types ...
    messages: z.custom<BaseMessage>()
  })
  ```
- default values are now recognised


libs/langchain/src/agents/annotation.ts
```diff
@@ -83,3 +87,167 @@ export type ToAnnotationRoot<A extends AnyAnnotationRoot | InteropZodObject> =
...
+  if (
+    typeof stateSchema === "object" &&
+    "_def" in stateSchema &&
+    "shape" in stateSchema._def
+  ) {
+    const shape =
+      typeof stateSchema._def.shape === "function"
+        ? stateSchema._def.shape()
+        : stateSchema._def.shape;
+    const annotationFields: Record<string, any> = {};
```
hntrl: do we need to be handling zod 3 + 4 schemas here? if so you can reach for the [zod utils](https://github.com/langchain-ai/langchainjs/blob/main/langchain-core/src/utils/types/zod.ts) exported from core

hntrl (commented):
just to parrot this so I know I understand:

if we want to expand agent state and we encounter a zod schema, we do our best effort to cast it to an annotation? (based on well known zod types??)

libs/langchain/src/agents/annotation.ts
christian-bromann: Thank you! I was able to use a bunch of these.

christian-bromann:
> if we want to expand agent state and we encounter a zod schema, we do our best effort to cast it to an annotation? (based on well known zod types??)

That is correct, the prebuilt version allowed the same using a langgraph helper but it missed some devx niceties that I mentioned above.

PR #8819, opened 2025-08-29
feat(mongodb): auto append client metadata for mongodb

durran:
Instead of using the docs show users how to set metadata, this PR adds the ability to add the name to the metadata post client construction in the MongoDB vcetor search, memory, and chat history modules.

### DIFF
--- a/examples/src/memory/mongodb.ts
+++ b/examples/src/memory/mongodb.ts
@@ -4,9 +4,7 @@ import { ChatOpenAI } from "@langchain/openai";
 import { ConversationChain } from "langchain/chains";
 import { MongoDBChatMessageHistory } from "@langchain/mongodb";
 
-const client = new MongoClient(process.env.MONGODB_ATLAS_URI || "", {
-  driverInfo: { name: "langchainjs" },
-});
+const client = new MongoClient(process.env.MONGODB_ATLAS_URI || "");
 await client.connect();
 const collection = client.db("langchain").collection("memory");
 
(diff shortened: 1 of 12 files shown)


examples/src/memory/mongodb.ts
```diff
@@ -4,9 +4,7 @@ import { ChatOpenAI } from "@langchain/openai";
 import { ConversationChain } from "langchain/chains";
 import { MongoDBChatMessageHistory } from "@langchain/mongodb";
 
-const client = new MongoClient(process.env.MONGODB_ATLAS_URI || "", {
-  driverInfo: { name: "langchainjs" },
-});
```
hntrl, line 9: is this driverInfo key not relevant anymore?

hntrl: ah nvm I see it's a concern of .appendMetadata now

libs/langchain-mongodb/src/version.ts
```diff
@@ -0,0 +1 @@
+export const VERSION = '0.1.0';
```
hntrl: won't this be 0.1.1 once we release? wonder if there's a better way to pull this package version (e.g. just reading package.json)

Is the `mongodb` package we're pulling in here node only? If it is then I don't have trouble with using `require('package.json')`

durran: @hntrl I've updated to pull the version now from the package.json, however I've broken the bundling process since rollup is failing to import the JSON. I don't seem to see any rollup config but think I now would need to include @rollup/plugin-json for this and put a configuration somewhere. I guess my question then is this acceptable?

durran: Version has been removed.

PR #8828, opened 2025-09-01
fix(langchain): add more message exports

christian-bromann:
Found these being used quite extensively in the docs, so we should export them from the root.


christian-bromann:
> Might be worth double checking that exporting `type ContentBlock` doesn't wash the namespace types

Works as expected:

Screenshot 2025-09-01 at 12 53 50 PM

PR #8829, opened 2025-09-02
fix(langchain): port and export AgentRuntime

christian-bromann:
Let's have user import an `AgentRuntime` type that is similar to Langgraphs version but now can be imported with type comments directly from `langchain`.


hntrl (approved):
happy to merge, but thoughts on renaming `AgentRuntime` -> `AgentContext` in a later release?

PR #8841, opened 2025-09-02
feat(core)!: v1 messages

hntrl:
post alpha continuation of #8669


libs/langchain-core/src/messages/function.ts
```diff
@@ -40,25 +52,30 @@ export class FunctionMessage extends BaseMessage {
...
+export class FunctionMessageChunk<
+  TStructure extends $MessageStructure = $StandardMessageStructure
+> extends BaseMessageChunk<TStructure, "function"> {
   static lc_name() {
     return "FunctionMessageChunk";
   }
 
+  readonly type = "function" as const;
+
   _getType(): MessageType {
```
dqbd: _nit_: now that `type` is abstract, we don't need to define `_getType` as abstract anyway

PR #8844, opened 2025-09-04
ci: add codeql security config

sinedied:
Adds fine-tuned CodeQL workflow focused on security scanning, as discussed on Slack.
The aim for config changes is enhance security coverage while decrease the false positives to a minimum, I can share the results of this workflow on my fork if needed to give you a preview.

### DIFF
--- a/.github/codeql/codeql-config.yml
+++ b/.github/codeql/codeql-config.yml
@@ -1,0 +1,23 @@
+name: CodeQL config
+
+# Needed to customize the security-extended query suite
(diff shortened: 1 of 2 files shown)


sinedied:
Made a small tweak to still have it run on PRs but not error, as upload should only happen on main

sinedied:
Ah the error indicates that it conflict with the default CodeQL set up, you'll have to disable it to allow the config for this PR to run (makes sense to not have both!)

PR #8846, opened 2025-09-04
bump firebase-admin devDependencies

urakozz:
Bump firebase-admin dependency

### DIFF
--- a/examples/package.json
+++ b/examples/package.json
@@ -88,7 +88,7 @@
     "dotenv": "^16.0.3",
     "duck-duck-scrape": "^2.2.5",
     "exa-js": "^1.0.12",
-    "firebase-admin": "^12.0.0",
+    "firebase-admin": "^13.0.0",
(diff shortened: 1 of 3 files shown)


hntrl (commented):
Thanks for the PR @urakozz!

One thing -- I think we'll need to bump this in peerDependencies if we're upgrading the dep

https://github.com/langchain-ai/langchainjs/pull/8846/files#diff-252ebd2576253f53ca111235c69d74ae803b12621d671319d9edd5ceb0188dc5R312

urakozz:
@hntrl yes sure, it accidentally slipped out of my attention :) I see Aritable connector is failing on Ubuntu 18, but its logic seems to be far from `firebase-admin`, I'm not sure if it was my change that caused that 🤔. Also,  Locally on macOS and Ubuntu 24 tests passed

hntrl (approved):

PR #8847, opened 2025-09-04
fix: URL encode paths in GitHub document loader

dd-eg-user:
Fix issue where directories with special characters (like '%5Fmeta') would cause 404 errors because paths were not properly URL encoded when constructing GitHub API requests.

- Apply encodeURIComponent() to path parameter in fetchRepoFiles method
- Add unit test to verify proper encoding of paths with special characters


hntrl (approved):

hntrl:
not quite sure what happened here -- tried to rebase and github closed this. Continued on in #8860

PR #8848, opened 2025-09-04
fix(langchain): handle prompts created from runs

hntrl:
prompt manifests can sometimes contain direct chat models (instead of RunnableBinding) which we don't handle with node pull


langchain/src/util/misc.ts
jacoblee93, line 1: nit: does this need its own file? Would there be another place this makes sense?

PR #8853, opened 2025-09-05
fix(core): Fix deep nesting of runnables within traceables

jacoblee93:
Previously the callback manager would always prefer the implicit parent ids, which flattened deeply nested runnables within traceables.

### DIFF
--- a/langchain-core/src/tracers/tests/langsmith_interop.test.ts
+++ b/langchain-core/src/tracers/tests/langsmith_interop.test.ts
@@ -150,21 +150,6 @@ test.each(["true", "false"])(
     });
     expect(thirdCallParams).toMatchObject({
       end_time: expect.any(Number),
-      inputs: {
-        args: [
-          {
-            lc: 1,
-            type: "constructor",
-            id: ["langchain_core", "messages", "HumanMessage"],
-            kwargs: {
-              content: "Hello!",
-              additional_kwargs: {},
-              response_metadata: {},
-            },
-          },
-          "David",
-        ],
-      },
       outputs: { outputs: "Hello!David" },
       parent_run_id: firstCallParams.id,
       extra: expect.any(Object),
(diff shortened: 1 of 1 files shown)


langchain-core/src/callbacks/manager.ts
```diff
@@ -1256,9 +1256,14 @@ export class CallbackManager
...
-        callbackManager._parentRunId =
-          LangChainTracer.getTraceableRunTree()?.id ??
-          callbackManager._parentRunId;
+        const implicitRunTree = LangChainTracer.getTraceableRunTree();
+        if (implicitRunTree && callbackManager._parentRunId === undefined) {
+          callbackManager._parentRunId = implicitRunTree.id;
+          const tracerV2 = callbackManager.handlers.find(
+            (handler) => handler.name === "langchain_tracer"
+          ) as LangChainTracer;
+          tracerV2?.updateFromRunTree(implicitRunTree);
```
jacoblee93, line 1265: Previously this was always setting `callbackManager._parentRunId` to the implicit traceable value, so runs would get flattened

Before #8849, this only applied in the case where a `LangChainTracer` was instantiated automatically, which happens if it is not manually passed and environment variables are set. My fix in #8849 for cases where a tracer is passed manually caused this reparenting logic to apply even if the tracer were already present, exposing the issue

This change only applies the reparenting on the initial handoff from LangSmith to LangChain (if there's no already existing parent id on the callback manager and there is an implicit one from a wrapping LangSmith `traceable`). The reverse handoff is handled in `langsmith` `traceable`

langchain-core/src/tracers/tests/langsmith_interop.test.ts
```diff
@@ -150,21 +150,6 @@ test.each(["true", "false"])(
     });
     expect(thirdCallParams).toMatchObject({
       end_time: expect.any(Number),
-      inputs: {
```
jacoblee93, line 153: Removing `inputs` in these tests is a side-effect of bumping `langsmith` - we made an optimization to stop unnecessarily double-sending inputs in the second update call for a run when they were not changing.

dqbd (approved):

PR #8855, opened 2025-09-05
docs(langchain-azure-cosmosdb): add storage configuration details for Azure Cosmos DB integrations

sinedied:
While working with these integrations and hitting a few bumps, I though adding more details here would be useful.


hntrl:
Thanks @sinedied! Happy to merge these, but as a heads up the same change will need to happen in the v1 [docs repo](https://github.com/langchain-ai/docs) for it to stick (we've already migrated all of the existing integration docs)

hntrl (approved):

PR #8858, opened 2025-09-05
chore(core): swap default cache key algorithm

hntrl:
cc @manekinekko @sinedied

we'll be shipping this key change with v1


manekinekko:
Thanks for the heads up @hntrl

PR #8860, opened 2025-09-05
fix(community): URL encode paths in GitHub document loader

hntrl:
continuation of #8847


dd-eg-user:
Thanks for getting this merged @hntrl !

since there is no changeset, how does this get released into the wild?

hntrl:
@dd-eg-user We're just using changesets for tracking right now. The latest community version was released alongside core in [this PR](https://github.com/langchain-ai/langchainjs/pull/8862)

PR #8867, opened 2025-09-06
feat(langchain-community): add custom schema option for neon vector store

FlickDaWist:
This PR introduces custom schemas option for Neon Postgres vector store. Default to public schema.


libs/langchain-community/src/vectorstores/neon.ts
```diff
@@ -68,6 +72,12 @@ export class NeonPostgres extends VectorStore {
     this.metadataColumnName = config.columns?.metadataColumnName ?? "metadata";
   }
 
+  get computedTableName() {
+    return this.schemaName == null
```
christian-bromann: If `schemaName` has `public` as default value, should we really check for `null` here? This can technically only happen if user intentionally passes an empty string here right?

FlickDaWist: you're right, will change this to allow null instead like `libs/langchain-community/src/vectorstores/pgvector.ts` did.

libs/langchain-community/src/vectorstores/neon.ts
```diff
@@ -60,6 +63,7 @@ export class NeonPostgres extends VectorStore {
 
     this.neonConnectionString = config.connectionString;
     this.tableName = config.tableName ?? "vectorstore_documents";
+    this.schemaName = config.schemaName ?? null;
```
christian-bromann: No need to set `null` as default
```suggestion
    this.schemaName = config.schemaName;
```

libs/langchain-community/src/vectorstores/neon.ts
christian-bromann:
```suggestion
    return typeof this.schemaName !== "string"
```

docs/core_docs/docs/integrations/vectorstores/neon.mdx
```diff
@@ -47,6 +47,15 @@ const vectorStore = await NeonPostgres.initialize(embeddings, {
...
 ```
 
+### Optional
+You can put the vector store table in custom db schema by passing `schemaName` to the config options during initialization.
+```typescript
+const vectorStore = await NeonPostgres.initialize(embeddings, {
+  connectionString: NEON_POSTGRES_CONNECTION_STRING,
+  schemaName: "mycustomschema"
+});
+```
```
christian-bromann:
```suggestion
### Optional

You can put the vector store table in custom db schema by passing `schemaName` to the config options during initialization.

```typescript
const vectorStore = await NeonPostgres.initialize(embeddings, {
  connectionString: NEON_POSTGRES_CONNECTION_STRING,
  schemaName: "mycustomschema",
});
```
```

christian-bromann (commented):
lint fix

christian-bromann:
Will take over and land in https://github.com/langchain-ai/langchainjs/pull/8921

PR #8868, opened 2025-09-06
fix(core): make mustache prompt with nested object working correctly

taiichiro-nagase:
~This PR adds a test case that reveals a circumstance in which Mustache nested objects do not work correctly.~
~Since I don't have a detailed enough understanding to make code changes, I'll start by adding test cases.~
~This test case passes on @langchain/core@0.3.64 which I'm using for now.~

~This issue may be caused by the following PR: https://github.com/langchain-ai/langchainjs/pull/8580.~

This PR fixes an mustache template issue that nested objects won't work correctly.

### DIFF
--- a/.changeset/strong-worms-sneeze.md
+++ b/.changeset/strong-worms-sneeze.md
@@ -1,0 +1,5 @@
+---
+"@langchain/core": patch
+---
+
+make mustache prompt with nested object working correctly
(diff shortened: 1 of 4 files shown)


taiichiro-nagase:
@hntrl @christian-bromann thanks for the rapid review & merging!

PR #8870, opened 2025-09-06
Fixes incorrect runtime behavior of isAIMessage and isAIMessageChunk.

Ankur-Kumar222:
Previously, both guards only checked `_getType() === "ai"`, which caused them to return true for both `AIMessage` and `AIMessageChunk`. This made it impossible to distinguish between full messages and streaming chunks at runtime, leading to issues in streaming workflows.

This PR updates the guards to use `instanceof` checks, ensuring proper runtime discrimination between `AIMessage` and `AIMessageChunk`. This improves reliability when handling streaming output and prevents duplicate or incorrect processing of messages.

[Related Discussion](https://github.com/langchain-ai/langchainjs/discussions/8318)

### DIFF
--- a/langchain-core/src/messages/ai.ts
+++ b/langchain-core/src/messages/ai.ts
@@ -220,11 +220,11 @@ export class AIMessage extends BaseMessage {
 }
 
 export function isAIMessage(x: BaseMessage): x is AIMessage {
-  return x._getType() === "ai";
+  return x instanceof AIMessage;
 }
 
 export function isAIMessageChunk(x: BaseMessageChunk): x is AIMessageChunk {
-  return x._getType() === "ai";
+  return x instanceof AIMessageChunk;
 }
 
 export type AIMessageChunkFields = AIMessageFields & {


hntrl:
Hey @Ankur-Kumar222!

The `instanceof` operator is a tricky one when using in library code -- if two versions of `@langchain/core` were installed in a project, it means that the instanceof check for AIMessage/AIMessageChunk in package version A fails when comparing with AIMessage/AIMessageChunk in package version B, which has led to some hard-to-debug issues before. It's why we strictly disallow using `instanceof` in our lint rules for langchain.

If you're using typescript and are passing in an arbitrary values to `isAIMessage`/`isAIMessageChunk`, the type check will fail since we explicitly require that you pass in a `BaseMessage` or `BaseMessageChunk`.

To safely assert that a message/chunk is what it is at runtime, you currently need to do something like this:

```ts
if (
  isBaseMessage(obj) &&
  isAIMessage(obj)
) {
  // do thing
}

if (
  isBaseMessageChunk(obj) &&
  isAIMessageChunk(obj)
) {
  // do thing (but with a chunk)
}
```

This isn't the most ideal solution -- we'll be iterating on this apart of our v1 efforts (probably through a static `AIMessage.isInstance` method)

PR #8877, opened 2025-09-08
chore(langchain): adapt agent to messages changes

hntrl:


libs/langchain/src/agents/utils.ts
```diff
@@ -61,22 +57,19 @@ export type AgentNameMode = "inline";
...
     return message;
   }
 
   const { name } = message;
 
   if (typeof message.content === "string") {
     return new AIMessage({
-      ...(Object.keys(message.lc_kwargs ?? {}).length > 0
-        ? message.lc_kwargs
-        : message),
```
hntrl, line 79: `MessageLike` isn't spreadable, and lc_kwargs should always exist on the object after the `AIMessage.isInstance` guards

hntrl:
dependency tests are failing because latest `langchain-core` hasn't been published

PR #8879, opened 2025-09-08
fix(openai/v1): handle OpenRouter image generation responses

christian-bromann:
This PR adds support for OpenRouter's image generation API response format in the ChatOpenAI integration.

## Problem
When using OpenRouter's image generation models (e.g., `google/gemini-2.5-flash-image-preview`), the generated images were not being received in LangChain.js responses. OpenRouter returns images in a separate `choices[0].message.images` array, which differs from the standard OpenAI format.

## Solution
- Modified `_convertCompletionsMessageToBaseMessage` to detect and extract images from OpenRouter's response format
- Convert the images to LangChain's standard `Multimodal.Image` content blocks
- Preserve backward compatibility by also adding image URLs to `additional_kwargs`

## Testing
Added comprehensive unit tests covering:
- Single image responses
- Multiple image responses  
- Text-only responses (no images)

The fix has been tested with real OpenRouter API calls and confirms that image URLs are now properly received in both streaming and invoke responses.

Fixes #8791

PR #8880, opened 2025-09-08
fix(openai/0.x): handle OpenRouter image generation responses

christian-bromann:
backported from #8879


christian-bromann:
Updated PR in the same way as #8879

hntrl (approved):

PR #8900, opened 2025-09-09
chore: bump core versions for aws + ollama

hntrl:
pre-requisite for v1 alphas of these packages

### DIFF
--- a/libs/providers/langchain-aws/package.json
+++ b/libs/providers/langchain-aws/package.json
@@ -35,7 +35,7 @@
     "@aws-sdk/credential-provider-node": "^3.750.0"
   },
   "peerDependencies": {
-    "@langchain/core": ">=0.3.58 <0.4.0"
+    "@langchain/core": "^1.0.0-alpha.1"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


libs/providers/langchain-aws/package.json
```diff
@@ -35,7 +35,7 @@
     "@aws-sdk/credential-provider-node": "^3.750.0"
   },
   "peerDependencies": {
-    "@langchain/core": ">=0.3.58 <0.4.0"
+    "@langchain/core": "^1.0.0-alpha.1"
```
christian-bromann: In `langchain` I set this to `"@langchain/core": "^1.0.0-alpha.3 <2.0.0",` as peer deps have to define a range 🤔 if this does the same, let's go with that.

hntrl: Isn't `"^1.0.0-alpha.3 <2.0.0"` synonymous with `"^1.0.0-alpha.3"`? Since the caret in front is technically defining a range up to the next major release (e.g. v2)

hntrl: nvm that only matches alpha versions

Issue #8094, opened 2025-04-30
Missing Model Name in Together.AI Integration Defaults Without Warning


YasharF:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain.js documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain.js rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```
// Expect to get an error or a warning for not properly providing a model name

const intendedModel='meta-llama/Llama-4-Scout-17B-16E-Instruct'
const llm = new ChatTogetherAI({
      apiKey: process.env.TOGETHERAI_API_KEY,
      modelName: intendedModel,
    });
```

### Error Message and Stack Trace (if applicable)

[no error or warning]

### Description

When a Node.js developer fails to properly pass a model name to ChatTogetherAI(), it defaults to `mistralai/Mixtral-8x7B-Instruct-v0.` without any error or warning. Instead of silently falling back to a default model, the expected behavior should be either:

- Throwing an error to notify the developer that a model name was not properly set.
- Emitting a console warning to indicate that the default model is being used.

_**Relevant Code:**_
This default fallback behavior occurs in the following line: https://github.com/langchain-ai/langchainjs/blob/46439657b27a01baf3290e4ac2021fb22ef5f594/libs/langchain-community/src/chat_models/togetherai.ts#L459

_**Proposed Fix:**_
I can submit a PR to remove the default model fallback and instead throw an error when no model name is provided. This approach would be consistent with the handling of missing API keys, where an error is raised in the preceding lines.

Please let me know if this would be an acceptable fix or if there are preferred alternatives.

### System Info

$ node -v
v22.15.0

$ npm ls langchain
[project name]
├─┬ @langchain/community@0.3.41
│ └── langchain@0.3.24 deduped
└── langchain@0.3.24

OS: WSL Ubuntu 24.04.01 LTS

YasharF:
fix: https://github.com/langchain-ai/langchainjs/pull/8263

PR #8263, opened 2025-05-28
feat(togetherAI): warn for missing chat model name

YasharF:
When a Node.js developer fails to properly pass a model name to ChatTogetherAI(), it silently defaults to mistralai/Mixtral-8x7B-Instruct-v0. without any error or warning.  With this PR, the developer will see a warning in the console.  This PR also changed the default (fallback) to meta-llama/Llama-3.3-70B-Instruct-Turbo-Free which togetherAI is graciously hosting for free with more restricted rate limits.

Fixes https://github.com/langchain-ai/langchainjs/issues/8094

x: YasharF
linkedin.com/in/yasharf

PR #8780, opened 2025-08-25
chore(*): Remove deprecated `modelName` parameter

christian-bromann:
This PR removes the deprecated `modelName` parameter throughout the entire code base.

PR #8904, opened 2025-09-10
feat(community): warn if chat model name is missing

YasharF:
Added a console warning for all chat model provider integrations that
silently fall back to a default model when no model name is provided.
This makes the behavior explicit and consistent across providers.

Silent fallbacks can catch developers off guard. For example, a simple
typo in the parameter name previously wouldn't be caught until the
developer dug deep to identify the cause of unexpected model behavior
or a billing issue from the provider.

For `ChatTogetherAI()`, the default (fallback) model has also been
changed from `mistralai/Mixtral-8x7B-Instruct-v0` to
`meta-llama/Llama-3.3-70B-Instruct-Turbo-Free`, which TogetherAI
graciously hosts for free with more restrictive rate limits.

Fixes https://github.com/langchain-ai/langchainjs/issues/8094
Replaces: https://github.com/langchain-ai/langchainjs/pull/8263

x: YasharF3
linkedin.com/in/yasharf

### DIFF
--- a/libs/langchain-community/src/chat_models/alibaba_tongyi.ts
+++ b/libs/langchain-community/src/chat_models/alibaba_tongyi.ts
@@ -280,6 +280,12 @@ export class ChatAlibabaTongyi
     this.maxTokens = fields.maxTokens;
     this.repetitionPenalty = fields.repetitionPenalty;
     this.enableSearch = fields.enableSearch;
+
+    if (!fields?.model && !fields?.modelName) {
+      console.warn(
+        `Alibaba Tongyi model name not specified. Defaulting to "qwen-turbo".`
+      );
+    }
     this.modelName = fields?.model ?? fields.modelName ?? "qwen-turbo";
     this.model = this.modelName;
   }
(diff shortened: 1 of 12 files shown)


YasharF:
Converting to draft, because this PR is blocked on https://github.com/langchain-ai/langchainjs/pull/8780 and needs to be modified after that PR is merged.  I have subscribed to the other pull request to get a ping when it is merged to come back and fix this.

christian-bromann (commented):
I am not sure if we want to `console.log` this as it is to explicit and users don't have control over it. Do you think there is a chance we can better document this to resolve the confusion? Alternatively: require user to define a model?

YasharF:
I think we should document it regardless, but it wouldn't catch cases when someone has made an error when passing the model.  ie. using `modelName` instead of `model`, or a missed typo in `model`.  I do think that requiring it and throwing an Error is a good solution, but that would make it a breaking change.   Do you want me to go with that route?

hntrl:
I'm with Christian on this one. I think generally it would be better if we failed fast if we were missing model instead of polluting the console. That's a breaking change though which we can slate for later

YasharF:
Thank for the great work folks.  All good.  I'm actually planning to get off Together AI soon.

PR #8907, opened 2025-09-10
fix(langchain): add LangGraph re-exports for better compatibility

christian-bromann:
We've discovered that some primitives in LangGraph v0.x aren't compatible with Langchain v1 as it depends on LangGraph v1. To avoid friction let's allow users to import these primitives from `langchain` directly.

### DIFF
--- a/libs/langchain/src/index.ts
+++ b/libs/langchain/src/index.ts
@@ -45,10 +45,15 @@ export {
 } from "./agents/index.js";
 
 /**
- * LangChain Memory
- * Check in what we want to export here
+ * Re-export essential LangGraph primitives for state persistence, storage,
+ * and execution control to ensure version compatibility across the ecosystem
  */
-export { MemorySaver, InMemoryStore } from "@langchain/langgraph";
+export {
+  MemorySaver,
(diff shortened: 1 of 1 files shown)


hntrl (approved):

libs/langchain/src/index.ts
```diff
@@ -45,10 +45,15 @@ export {
...
- * LangChain Memory
- * Check in what we want to export here
+ * Re-export essential LangGraph primitives for state persistence, storage,
+ * and execution control to ensure version compatibility across the ecosystem
  */
-export { MemorySaver, InMemoryStore } from "@langchain/langgraph";
+export {
+  MemorySaver,
+  InMemoryStore,
+  Command,
```
dqbd, line 54: Might be worth re-exporting `Send` as well

christian-bromann: Are there scenarios where users use it in the `createAgent` context?

christian-bromann:
Docs PR: https://github.com/langchain-ai/docs/pull/481

christian-bromann:
Closing as discussed in docs PR.

PR #8918, opened 2025-09-10
fix(groq/v1): Support generic messages in messageToGroqRole

christian-bromann:
v1 for #8917


libs/providers/langchain-groq/src/chat_models.ts
```diff
@@ -292,8 +292,32 @@ interface TokenUsage {
...
+}
+
+/**
+ * Extract the role from a message.
+ * @param message - The message to extract the role from.
+ * @returns The role of the message.
+ */
 export function messageToGroqRole(message: BaseMessage): GroqRoleEnum {
-  const type = message._getType();
+  const type = message.getType();
```
hntrl:
```suggestion
  const type = message.type;
```
we want to get rid of these methods for v1

PR #8920, opened 2025-09-10
chore(langchain): remove dependency to openai provider package

christian-bromann:
Another attempt to remove `@langchain/openai` from `langchain` in a minimal invasive way.


libs/langchain-community/src/load/import_map.ts
hntrl, line 1: where did these come from?

hntrl (approved):
ooc what's blocking us from getting rid of the dep entirely? (instead of it staying as a peer dep)

christian-bromann:
> ooc what's blocking us from getting rid of the dep entirely? (instead of it staying as a peer dep)

Good call, removed it.

PR #8926, opened 2025-09-11
chore(community/v1): rm deprecated couchbase integration

hntrl:
@ejscribner You mentioned that `CouchbaseVectorStore` is deprecated. We can ship the deprecation for v0, but for v1 we can just get rid of it since there is a conscious upgrade path between v0 and v1. Lmk if you have other opinions


ejscribner:
@hntrl That sounds like the best path forward to me, thanks for your help on this!

PR #8930, opened 2025-09-11
rfc(core): message serialization changes 

dqbd:
## RFC: V1 Message Serialization

With LangChain V1 we're revisiting the default behaviour of `BaseMessage.toJSON()`, which in 0.x defaults to the `Serialized` schema. The `Serialized` schema, although useful when paired with custom reviver, is too verbose when used to send messages across the wire.

We're proposing the following changes:

- The new `toJSON()` return type now semantically matches how people interact with messages, using the same schema that LangGraph API uses when ser/de messages
- Introduce a new `toSerialized()` method in `Serializable` interface that returns the previous format for backwards compatibility.
- Add serialization helper in `@langchain/core/load` that invokes `toSerialized()` if present.

### DIFF
--- a/libs/langchain-core/src/load/index.ts
+++ b/libs/langchain-core/src/load/index.ts
@@ -244,7 +244,7 @@ export async function load<T>(
   return reviver.call({ ...mappings }, json) as Promise<T>;
 }
 
-export function stringify<T>(instance: T): string {
+export function dump<T>(instance: T): string {
   function replacer(this: unknown, key: string, value: unknown): unknown {
     if (typeof this !== "object" || this == null) return value;
 
--- a/libs/langchain-core/src/messages/tests/base_message.test.ts
+++ b/libs/langchain-core/src/messages/tests/base_message.test.ts
@@ -9,7 +9,7 @@ import {
   coerceMessageLikeToMessage,
   SystemMessage,
 } from "../index.js";
-import { load, stringify } from "../../load/index.js";
+import { load, dump } from "../../load/index.js";
 import { concat } from "../../utils/stream.js";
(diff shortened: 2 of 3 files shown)


libs/langchain-core/src/messages/base.ts
```diff
@@ -320,12 +331,18 @@ export abstract class BaseMessage<
...
         .kwargs as StoredMessageData,
     };
   }
 
+  toJSON(): Message<TStructure, TRole> {
+    const { type: _, ...data } = (this.toSerialized() as SerializedConstructor)
+      .kwargs as Message<TStructure, TRole>;
+    return { type: this.type, ...data };
+  }
+
```
hntrl, line 345: any reason we shouldn't just read off `this.content` and `this.response_metadata`?

libs/langchain-core/src/load/index.ts
```diff
@@ -243,3 +243,23 @@ export async function load<T>(
...
+      typeof origValue.toSerialized === "function"
+    ) {
+      return origValue.toSerialized();
+    }
+
+    return value;
+  }
+
+  return JSON.stringify(instance, replacer);
+}
```
hntrl: is a better name for this something like `serialize`?

libs/langchain-core/src/messages/base.ts
dqbd: Since we're already relying on Serialized, might as well make sure that we're properly retrieving the properties (in case of incorrect mutation).

libs/langchain-core/src/load/index.ts
dqbd: Might as well rename it to `dump` to match Python

PR #8786, opened 2025-08-26
feat: add Decodo integration documentation

domantas-jurkus-dcd:
In this PR:
- Adding documentation on how to use [@decodo/langchain-ts](https://www.npmjs.com/package/@decodo/langchain-ts). Got a tip from #8752 that integrations should be placed in company-owned repos.

Let me know if i missed anything.

PR #8943, opened 2025-09-12
feat(community): Enhance SAP HANA Vector integration with metadata columns, keyword filtering, internal embeddings, hanatranslator and support for vector column types

VarinThakur01:
## Description

This PR enhances the SAP HANA Cloud Vector Store integration in `langchainjs` with five major features that align its behavior with the existing Python implementation:

---

### 1. Dedicated Metadata Columns

Metadata fields can now be stored in dedicated table columns in addition to the JSON metadata field. This enables more efficient filtering and indexing, and mirrors the behavior of the Python implementation (`add_texts` in `hanavector.py`).

### 2. Keyword Search Filtering

Added support for a `$contains` operator in metadata filters, which constructs SQL full-text search conditions such as:

```sql
SCORE(? IN ("<column>" EXACT SEARCH MODE 'text')) > 0
```

When filtering on fields stored in the JSON metadata, a `WITH` clause is used to project the relevant key into its own column.

### 3. Internal Embedding Support

Introduced a `HanaInternalEmbeddings` class that uses SAP HANA's native `VECTOR_EMBEDDING` function to compute document and query embeddings directly in the database. This allows fully in-database similarity search with reduced external dependencies.  
Depending on whether internal or external embeddings are used, the query structure dynamically switches between injecting a precomputed vector (via `embedQuery()`) or calling the native HANA `VECTOR_EMBEDDING` function inline.

### 4. Support for Vector Column Datatypes

Introduced a new parameter `vectorColumnType` in the `HanaDBArgs` Interface to support the datatypes of `REAL_VECTOR` (32-bit floating point precision)  and `HALF_VECTOR` (16-bit floating point precision). This allows to specify the `vectorColumnType` while creating the db instance. 

### 5. HanaTranslator Class

The `HanaTranslator` Class is now added to support `SelfQueryRetrieval`.

---

## Tests & Documentation

- Integration tests and documentation have been updated to reflect the new features.
- This implementation brings the `langchainjs` SAP HANA Vector Store in alignment with the [external LangChain Integration Package - Python](https://github.com/SAP-samples/sap-hana-ml-samples/tree/main/Python-langchain).

---

**Issue**: No associated issue.  
**Dependencies**: No new dependencies were introduced in this PR.  
**Twitter handle**: [@sapopensource](https://twitter.com/sapopensource)

### DIFF
--- a/examples/src/indexes/vector_stores/hana_vector/basics.ts
+++ b/examples/src/indexes/vector_stores/hana_vector/basics.ts
@@ -29,7 +29,9 @@ await new Promise<void>((resolve, reject) => {
   });
 });
 const embeddings = new OpenAIEmbeddings();
+
 // define instance args
+// check the interface to see all possible options
 const args: HanaDBArgs = {
   connection: client,
   tableName: "testBasics",
--- a/examples/src/indexes/vector_stores/hana_vector/internalEmbeddings.ts
+++ b/examples/src/indexes/vector_stores/hana_vector/internalEmbeddings.ts
@@ -37,6 +37,7 @@ await new Promise<void>((resolve, reject) => {
 });
 
 // Define the arguments for the vector store instance.
+// Check the interface to see all possible options.
 const args: HanaDBArgs = {
   connection: client,
   tableName: "testInternalEmbeddings",


hntrl:
Hey @VarinThakur01! Thanks for the extensive PR!

This seems like a pretty invasive change that we should consider as breaking. It's not really tenable for us to release a minor version of the community package to introduce this, and a patch version would auto-upgrade people to this new implementation which also is out of the cards.

Generally when it comes to contributions to the community package, what we've been recommending to people is to make their own repo + npm package and release it independently of us, and we can add a backlink in our docs to your repo. If you're looking to get these changes introduced, we'd recommend hoisting out the hanavector community integration altogether and introducing them in the new package. If you need some inspiration, I think the recently merged [decodo](https://www.npmjs.com/package/@decodo/langchain-ts) + subsequent [docs PR](https://github.com/langchain-ai/langchainjs/pull/8786) is a good blueprint to follow.

Let me know if you need clarification or if we can support in anyway!

PR #8944, opened 2025-09-12
fix(@langchain/google-genai/v1): remove type field from string enums for Gemini compatibility

christian-bromann:
## Description

This PR fixes an issue where the Google Gemini API throws a 400 Bad Request error when using Zod enums in tool input schemas. The error occurs because Gemini's API doesn't accept the `type` field alongside `enum` fields for string enums, even though this is valid JSON Schema.

### The Problem

When using tools with enum parameters like:
```typescript
const schema = z.object({
  operation: z.enum(["add", "subtract", "multiply", "divide"])
});
```

The Zod to JSON Schema conversion produces:
```json
{
  "type": "string",
  "enum": ["add", "subtract", "multiply", "divide"]
}
```

But Gemini's API rejects this with: `"enum: only allowed for STRING type"`

### The Solution

This PR introduces a new `processEnumFieldsForGemini` function that:
- Removes the `type` field from objects that have both `enum` and `type: "string"`
- Recursively processes nested schemas (properties, items, oneOf, anyOf, allOf)
- Preserves the `type` field for non-string enums (e.g., numeric enums)

The fix is applied in both `schemaToGenerativeAIParameters` and `jsonSchemaToGeminiParameters` to ensure all schemas are properly transformed before being sent to Gemini.

### Testing

Added comprehensive unit and integration test cases.

Fixes #8883

PR #8945, opened 2025-09-12
fix(google-genai): remove type field from string enums for Gemini compatibility

christian-bromann:
Copied over from [#8944](https://github.com/langchain-ai/langchainjs/pull/8944)


hntrl:
assuming this was a v0 port over that never landed

PR #8948, opened 2025-09-12
fix(community): Add INSERT support to PrismaVectorStore for ParentDocumentRetriever compatibility (#8833)

### LINKED ISSUE #8833: Can not use ParentDocumentRetriever with Prisma
### Checked other resources

### Example Code

`node_modules\@langchain\community\dist\vectorstores\prisma.cjs`
Function: `addVectors`
Line: 159
Error code:
```
await this.db.$transaction(vectors.map((vector, idx) => this.db.$executeRaw(this.Prisma.sql `UPDATE ${tableNameRaw}
            SET ${vectorColumnRaw} = ${`[${vector.join(",")}]`}::vector
            WHERE ${idColumnRaw} = ${documents[idx].metadata[this.idColumn]}
          `)));
```

_No response_

### Why it's a bug:
When integrate with `ParentDocumentRetriever` follow this doc: https://js.langchain.com/docs/how_to/parent_document_retriever/
- `ParentDocumentRetriever` create child vector record automatically but prisma use UPDATE statement, it must be create or update
- It doesn't throw any error cause SQL statement is successful but no record is created

### System Info

Node 22
Windows/Mac
"@langchain/community": "^0.3.41",
"@langchain/core": "^0.3.49",

christian-bromann:
Previously, PrismaVectorStore only used UPDATE statements when adding vectors, which caused silent failures when used with ParentDocumentRetriever. The retriever creates new child documents that don't exist in the database, so UPDATE statements would succeed but not create any records.

Changes:
- Add new `addDocumentsWithVectors` method that uses INSERT statements to create records
- Modify `addDocuments` to use `addDocumentsWithVectors` instead of `addVectors`
- Maintain backward compatibility by keeping the original `addVectors` method unchanged
- Add tests to verify the new behavior and ensure no regression

This fix ensures PrismaVectorStore works correctly with ParentDocumentRetriever while maintaining compatibility with existing code.

Fixes #8833

### DIFF
--- a/.changeset/fifty-plants-drive.md
+++ b/.changeset/fifty-plants-drive.md
@@ -1,5 +1,0 @@
----
-"@langchain/mcp-adapters": minor
----
-
-feat(mcp-adapters): Add resource management methods and structured content support
--- a/.changeset/hungry-dolls-turn.md
+++ b/.changeset/hungry-dolls-turn.md
@@ -1,6 +1,0 @@
----
-"@langchain/core": patch
-"@langchain/openai": patch
----
-
-fix(langchain): Fix toJsonSchema mutating underlying zod schema


ivaaaan:
I've faced this issue. Basically, PrismaVectorStore is not usable if you're trying to store multiple chunks per document. Are there plans to merge this?

christian-bromann:
@hntrl is this good to go?

libs/langchain-community/src/vectorstores/prisma.ts
```diff
@@ -286,7 +286,7 @@ export class PrismaVectorStore<
    */
   async addDocuments(documents: Document<TModel>[]) {
     const texts = documents.map(({ pageContent }) => pageContent);
-    return this.addVectors(
+    return this.addDocumentsWithVectors(
       await this.embeddings.embedDocuments(texts),
       documents
```
hntrl: I think we want to put this new insert logic behind a flag? Especially when it comes to database interactions I want to avoid changing up write ops haphazardly unless it's in a breaking change, and we don't really have a want to minor bump the community package ATM

christian-bromann: @ivaaaan what are your thoughts here, how breaking would this change be?

ivaaaan: I guess makes sense. For my case it wouldn't be breaking change as PrismaVectorStore doesn't work with UPDATE, not sure about others.

christian-bromann:
@hntrl put the change behind a flag

PR #8950, opened 2025-09-12
chore(*): backport zod compat changes

hntrl:


hntrl:
> ooc: why do we use Zod v3 instead of v4?

For all the internal code paths that are using zod, we were originally using zod v3, and migrating all these paths to zod/v4 would make for a lot of busy work

hntrl:
Closing this -- we can't backport zod changes in this way since setting `zod` to be `3.25.76 || 4` violates the peer dependency requirement of `zod-to-json-schema` (only `3`) which causes failures in pnpm + yarn, and we can't remove `zod-to-json-schema` as a dependency since that is implicitly meeting the peer dep requirement for `zod-to-json-schema` in langgraph.

Will be fixed in v1

PR #8961, opened 2025-09-15
chore(core): inline zod-to-json-schema

hntrl:
## Problem
`zod-to-json-schema` has a dependency requirement on zod@3 (with no plan to upgrade), which depending on some package setups can lead to some awkward dependency errors with zod.

## Solution
Fixes the import issues by inlining the zod-to-json-schema package and changing code so it is always importing from `zod/v3`. Since we're permissive for either zod@3 or zod@4 in our package deps, it can cause issues since zod-to-json-schema will attempt to pull in zod/v4 by default if `zod@4` is installed (which it isn't compatible with).


hntrl:
> Instead of actually importing the code, can we just exclude it as `external` from the build? This would make it easier to maintain this as we wouldn't need to copy&paste code into the project.

unfortunately no. The reason we're inlining it is because we have to change the import paths for `zod` to always pull from `zod/v3`

PR #8972, opened 2025-09-15
feat(langchain): refine `reactAgent`

christian-bromann:
Further refinements and improvements to `createAgent` with middleware:

- updated `createAgent` symbol documentation
- improved type handling for `invoke` and `stream` invocation parameter
- acknowledge middleware context schema's that are optional or have default values as actually optional
- improvements to HITL middleware: support for multiple tool interrupts at the same time


libs/langchain/src/agents/middlewareAgent/middleware/hitl.ts
```diff
@@ -1,33 +1,118 @@
...
+  allowRespond: z.boolean().optional(),
+  description: z.string().optional(),
+});
+
+type ToolConfigSchema = z.input<typeof ToolConfigSchema>;
+type ToolCall = NonNullable<AIMessage["tool_calls"]>[number];
 
 /**
- * Interrupt request for tool approval
+ * Configuration that defines what actions are allowed for a human interrupt.
```
eyurtsev: Would be good to clarify the doc-string. Actions appear in two places and mean two different things:

1. There's the action requested by the agent / workflow etc
2. There's the action performed by the reviewer

It would be good to specify that this is the action performed by the request specifically otherwise users will think it's referring to the ActionRequest

libs/langchain/src/agents/middlewareAgent/middleware/hitl.ts
```diff
@@ -1,33 +1,118 @@
...
+   * Key-value pairs of arguments needed for the action (e.g., {"a": 1, "b": 2})
+   */
   args: Record<string, any>;
-  toolCallId: string;
+}
+
+/**
+ * Represents an interrupt triggered by the graph that requires human intervention.
+ *
+ * @param actionRequest - The specific action being requested from the human
```
eyurtsev: This action isn't requested form the human... It's an action that will be taken by the workflow / agent and is provided for review. (kind of like a proposed action)

libs/langchain/src/agents/middlewareAgent/middleware/hitl.ts
```diff
@@ -1,33 +1,118 @@
...
-export interface ToolApprovalRequest {
+export interface HumanInTheLoopConfig
+  extends Omit<ToolConfigSchema, "description"> {}
+
+/**
+ * Represents a request with a name and arguments.
+ */
+export interface ActionRequest {
+  /**
+   * The type or name of action being requested (e.g., "add_numbers")
```
eyurtsev: `send_email` might be a better as an example (more concrete)

w/ arguments being subject, body, cc, bcc etc

libs/langchain/src/agents/middlewareAgent/middleware/hitl.ts
```diff
@@ -1,33 +1,118 @@
...
   args: Record<string, any>;
-  toolCallId: string;
+}
+
+/**
+ * Represents an interrupt triggered by the graph that requires human intervention.
+ *
+ * @param actionRequest - The specific action being requested from the human
+ * @param config - Configuration defining what response types are allowed
+ * @param description - Optional detailed description of what input is needed
```
eyurtsev: We could check what agent inbox does, but I suspect that the description isn't for what input is needed, but generally to describe the result of performing the action. i.e., describe something non obvious about the action (e.g., send_email, after approval the email will be scheduled for sending but may not be sent for up to 5 mins)

---

I am not sure about this part, I think it depends a bit how the UI uses it, but this what I'd use the description for personally

libs/langchain/src/agents/middlewareAgent/middleware/hitl.ts
```diff
@@ -1,33 +1,118 @@
...
+     * If a tool doesn't have an entry, it's auto-approved by default.
+     *
+     * - `true` -> all actions allowed
+     * - `false` -> indicates that the tool is auto-approved.
+     * - `ToolConfig` -> indicates the specific actions allowed for this tool.
+     */
+    toolConfigs: z.record(z.union([z.boolean(), ToolConfigSchema])).default({}),
+    /**
+     * The prefix to use when constructing action requests.
+     * This is used to provide context about the tool call and the action being requested.
```
eyurtsev: Could we explain that it's only used for tools that do not a custom description. if the tools have a custom description (via ToolConfigSchema) the descriptionPrefix is not used for those tools

PR #8976, opened 2025-09-15
fix(mcp-adapters): use createAgent for examples and update tests

christian-bromann:
This patch looks at the `@langchain/mcp-adapter` package and:

- updates examples to use the new `createAgent` primitive from `langchain`
- updates test infrastructure


libs/langchain-core/src/messages/index.ts
```diff
@@ -15,6 +15,7 @@ export {
   type ToolMessageFields,
   ToolMessage,
   ToolMessageChunk,
+  type ToolCall,
```
hntrl: if we're adding this we can prob just

```ts
export * from "./tool.js";
```

.gitignore
```diff
@@ -52,4 +52,5 @@ credentials.json
 bun.lock
 coverage/
 
-examples/src/createAgent/*.png
\ No newline at end of file
+examples/src/createAgent/*.png
+libs/providers/langchain-mcp-adapters/examples/filesystem_test
```
hntrl: ?

PR #8978, opened 2025-09-16
fix: use `keyEncoder` instead of insecure cache key getter

hntrl:
https://forum.langchain.com/t/insecure-cache-algorithms/1514


christian-bromann (approved):
LGTM 👍

PR #8980, opened 2025-09-16
add new handleConnectionErrorsGracefully parameter to MultiServerMCPC…

gabedemesa:
WIP


hntrl (approved):
Made a couple of edits:
- changed it to be an enum flag instead of a boolean. Makes it a little less verbose that way
- Removed the connection behavior from loadMcpTools options, before it was a no-op
- added another failover test

hntrl:
Closing in favor of #9642

PR #8993, opened 2025-09-17
feat(langchain): add dynamic prompt built-in middleware

christian-bromann:
Adds a first-class dynamic prompt middleware to set the system prompt at invocation time.

- **New middleware**: `dynamicPrompt` for defining system prompts dynamically via middleware.
- **Exports**: Added to `langchain/middleware` export surface.
- **Tests**: Integration test validates provider receives the correct `SystemMessage`.
- **Docs**: Expanded Typedoc with typed context support and examples.

### Motivation
Previously, users could pass a callable `prompt` to agents. With the middleware system, dynamic prompts should be handled consistently via middleware. This improves ergonomics, composability, and clarity around when the prompt is computed.

### Usage
```ts
import { dynamicSystemPrompt } from "langchain/middleware";
import { z } from "zod";

const contextSchema = z.object({ region: z.string().optional() });

const middleware = dynamicSystemPrompt<z.infer<typeof contextSchema>>(
  (_state, runtime) =>
    `You are a helpful assistant. Region: ${runtime.context.region ?? "n/a"}`
);

const agent = createAgent({
  model: "anthropic:claude-3-5-sonnet",
  contextSchema,
  middleware: [middleware] as const,
});
```

### DIFF
--- a/examples/src/createAgent/middleware/hitl.ts
+++ b/examples/src/createAgent/middleware/hitl.ts
@@ -1,5 +1,5 @@
 import { createAgent, HumanMessage, tool, MemorySaver } from "langchain";
-import { humanInTheLoopMiddleware } from "langchain/middleware";
+import { humanInTheLoop } from "langchain/middleware";
 import { Command } from "@langchain/langgraph";
 import { z } from "zod";
 
@@ -51,7 +51,7 @@ const writeFileTool = tool(
 );
 
 // Configure HITL middleware
-const hitlMiddleware = humanInTheLoopMiddleware({
+const hitlMiddleware = humanInTheLoop({
   toolConfigs: {
     write_file: {
       requireApproval: true,
--- a/examples/src/createAgent/middleware/hitl.ts
+++ b/examples/src/createAgent/middleware/hitl.ts
@@ -1,5 +1,5 @@
 import { createAgent, HumanMessage, tool } from "langchain";
-import { humanInTheLoop } from "langchain/middleware";
+import { humanInTheLoopMiddleware } from "langchain/middleware";
 import { Command, MemorySaver } from "@langchain/langgraph";
 import { z } from "zod";
 
@@ -51,7 +51,7 @@ const writeFileTool = tool(
 );
 
 // Configure HITL middleware
-const hitlMiddleware = humanInTheLoop({
+const hitlMiddleware = humanInTheLoopMiddleware({
   toolConfigs: {
     write_file: {
       requireApproval: true,
--- a/examples/src/createAgent/middleware/hitl.ts
+++ b/examples/src/createAgent/middleware/hitl.ts
@@ -1,6 +1,6 @@
-import { createAgent, HumanMessage, tool, MemorySaver } from "langchain";
+import { createAgent, HumanMessage, tool } from "langchain";
 import { humanInTheLoop } from "langchain/middleware";
-import { Command } from "@langchain/langgraph";
+import { Command, MemorySaver } from "@langchain/langgraph";
 import { z } from "zod";
 
 const checkpointSaver = new MemorySaver();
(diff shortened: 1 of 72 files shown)


libs/langchain/src/agents/middlewareAgent/middleware/hitl.ts
```diff
@@ -172,9 +172,7 @@ const contextSchema = z.object({
  * @see {@link Command} for resuming interrupted execution
  * @public
  */
-export function humanInTheLoopMiddleware(
-  options: z.input<typeof contextSchema> = {}
-) {
+export function humanInTheLoop(options: z.input<typeof contextSchema> = {}) {
```
hntrl: radical idea: what if we called this something like `approvalMiddleware` instead of HITL

libs/langchain/src/agents/middlewareAgent/middleware/dynamicSystemPrompt.ts
```diff
@@ -0,0 +1,73 @@
...
+    name: "DynamicSystemPromptMiddleware",
+    prepareModelRequest: async (options, state, runtime) => {
+      const system = await fn(
+        state as AgentBuiltInState,
+        runtime as Runtime<TContextSchema>
+      );
+      const systemMessage =
+        typeof system === "string" ? new SystemMessage(system) : system;
+
+      return { ...options, systemMessage };
```
hntrl: I'm out of my depth with this, but does systemMessage get added to agent state? If there's multiple turns are we prepending a new system message to the beginning of the messages array each time?

christian-bromann: Changes in `modifyModelRequest` don't persist within state. The `dynamicSystemPrompt` middleware uses that hook to inject these system prompts dynamically.

libs/langchain/src/agents/middlewareAgent/middleware/hitl.ts
christian-bromann: Thoughts @sydney-runkle @eyurtsev ?

examples/src/createAgent/middleware/hitl.ts
```diff
@@ -1,5 +1,5 @@
 import { createAgent, HumanMessage, tool } from "langchain";
-import { humanInTheLoopMiddleware } from "langchain/middleware";
+import { humanInTheLoop } from "langchain/middleware";
```
eyurtsev: We should decide whether we keep the suffix everywhere or drop it.

I suspect some middleware will not be implemented in langchain/middleware but in integration packages. In this case, maybe the the Middleware suffix makes sense so it's easy to recognize things as agent middleware?

libs/langchain/src/agents/middlewareAgent/middleware/dynamicSystemPrompt.ts
```diff
@@ -0,0 +1,73 @@
...
+ *   middleware: [middleware],
+ * });
+ *
+ * await agent.invoke({ messages }, { context: { region: "EU" } });
+ * ```
+ *
+ * @example Returning a string instead of a SystemMessage
+ * ```ts
+ * const middleware = dynamicSystemPrompt((_state, runtime) =>
+ *   `You are a helpful assistant. Region: ${runtime.context.region ?? "n/a"}`
```
eyurtsev: Should we have this be the default example? system message is virtually always a string and isn't associated with ids that we care about?

libs/langchain/src/agents/middlewareAgent/middleware/hitl.ts
eyurtsev: I like HumanInTheLoopMiddleware (clear that it's associated w/ human in the loop functionality). no strong opinions just slight preference

libs/langchain/src/agents/middlewareAgent/middleware/dynamicSystemPrompt.ts
christian-bromann: Makes, sense, let's do string only

PR #8996, opened 2025-09-17
fix(langchain): remove MemorySaver re-export

christian-bromann:
We should not re-export LangGraph primitives from `langchain` canonical.


hntrl (approved):
there's prob a bunch of docs update we also have to do

christian-bromann:
> there's prob a bunch of docs update we also have to do

Part of https://github.com/langchain-ai/docs/pull/569

PR #9009, opened 2025-09-18
add embeddingColumnType and export indexes

urakozz:
Hello, I'm happy to contribute again.

Working in Google Cloud it turned out that implementation could be improved to support larger indexes like `gemini-embedding-001` with size of 3072. It looks like indexing of type `vector` is limited by 2000 dimensions and for bigger vector different type `halfvec` should be used.

In this PR I addressed 2 crucial points to achieve that:
- added a property for the column type to add a possibility to specify it as `halfvec`
- exported indexes.ts to make it more convenient to use them

### DIFF
--- a/libs/langchain-google-cloud-sql-pg/src/engine.ts
+++ b/libs/langchain-google-cloud-sql-pg/src/engine.ts
@@ -18,6 +18,7 @@ export interface VectorStoreTableArgs {
   schemaName?: string;
   contentColumn?: string;
   embeddingColumn?: string;
+  embeddingColumnType?: "vector" | "halfvec" | "bit" | "sparsevec";
   metadataColumns?: Column[];
   metadataJsonColumn?: string;
   idColumn?: string | Column;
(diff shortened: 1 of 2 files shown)


libs/langchain-google-cloud-sql-pg/src/engine.ts
```diff
@@ -18,6 +18,7 @@ export interface VectorStoreTableArgs {
   schemaName?: string;
   contentColumn?: string;
   embeddingColumn?: string;
+  embeddingColumnType?: "vector" | "halfvec" | "bit" | "sparsevec";
```
christian-bromann, line 21: Is there a way we can add a comment with a link to the spec? It would help know where these options come from.

christian-bromann (commented):
On tiny nit

urakozz:
@christian-bromann yes sure, done

christian-bromann (approved):
LGTM 👍

PR #9010, opened 2025-09-18
fix(core): change response metadata type mismatch

hntrl:
One-liner fix for a bug found when implementing standard content


libs/langchain-core/src/messages/tests/messages.test-d.ts
```diff
@@ -969,6 +970,12 @@ describe("$InferMessageProperty<TStructure, TRole, K>", () => {
 
 describe("$InferResponseMetadata<TStructure, TRole>", () => {
   // TODO(hntrl): implement
```
christian-bromann, line 972: I guess there is more to implement right?

hntrl: yus

PR #9011, opened 2025-09-18
chore(aws): adapt message changes for aws

hntrl:
Adapts some of the breaking/deprecated chat model changes for aws


libs/providers/langchain-aws/src/chat_models.ts
```diff
@@ -944,7 +944,7 @@ export class ChatBedrockConverse
             text: "",
             message: new AIMessageChunk({
               content: "",
-              response_metadata: chunk,
+              response_metadata: { ...chunk },
```
hntrl, line 947: there's some super strange typescript black magic happening here, but this is necessary to make it not complain

thank you typescript very cool

PR #9016, opened 2025-09-20
fix(langchain/agent): re-enable support for LangGraph

christian-bromann:
If someone tries to run a `createAgent` with LangGraph, it will fail due to the fact that `createAgent` wraps the LangGraph graph instance to provide better type security.

Rather then documenting that users can just do this:
```ts
const agent = createAgent(...)
export const graph = agent.graph
```

This patch makes the `createAgent` function return a proxy that directs access to graph properties to the right object. For the user this is totally transparent.

### DIFF
--- a/libs/langchain/src/agents/middlewareAgent/tests/reactAgent.test.ts
+++ b/libs/langchain/src/agents/middlewareAgent/tests/reactAgent.test.ts
@@ -1,3 +1,4 @@
+import { describe, it, expect } from "vitest";
 import { createAgent } from "../index.js";
 
 describe("ReactAgent", () => {


libs/langchain/src/agents/middlewareAgent/index.ts
```diff
@@ -368,7 +384,53 @@ export function createAgent<
...
+        target[
+          prop as keyof ReactAgent<
+            StructuredResponseFormat,
+            ContextSchema,
+            TMiddleware
+          >
+        ];
+      return typeof value === "function" ? value.bind(target) : value;
+    },
+  });
```
hntrl, line 433: do we foresee any collisions between `ReactAgent` properties and `CompiledStateGraph`?

hntrl (commented):
I'm partial to amending LGP to look for some kind of symbol that returns a graph instead of returning a proxy. I'm generally not a fan of using proxy objects since there's a lot of ways to shoot yourself in the foot

libs/langchain/src/agents/middlewareAgent/index.ts
christian-bromann: This is why have a `!agentMethodNames.has(prop as string)` check to ensure the methods we expose from `ReactAgent` will be called in favor of what `StateGraph` exposes.

hntrl: Hmm so in that hypothetical case `ReactAgent` would only partially implement state graph?

christian-bromann: If a function like `invoke` or `stream` is implemented in `ReactAgent` is will prefer that over the one implemented in state graph. With the proxy it will still reach into state graph for methods not implemented in `ReactAgent` which will guarantee the support for LGP.

christian-bromann:
Closing in favor of https://github.com/langchain-ai/langchainjs/pull/9117

PR #9017, opened 2025-09-20
fix(langchain/agents): remove llm prop from middleware agent

christian-bromann:
With introducing middleware we got rid of allowing users to provide an `llm` property. Instead they pass in a chat instance or a string to the `model` property. There are still some inconsistencies with Python which I will clear up in a separate PR.

### DIFF
--- a/libs/langchain/src/agents/middlewareAgent/ReactAgent.ts
+++ b/libs/langchain/src/agents/middlewareAgent/ReactAgent.ts
@@ -310,8 +310,6 @@ export class ReactAgent<
      */
     this.#graph = allNodeWorkflows.compile({
       checkpointer: this.options.checkpointer ?? this.options.checkpointSaver,
-      interruptBefore: this.options.interruptBefore,
-      interruptAfter: this.options.interruptAfter,
       store: this.options.store,
       name: this.options.name,
       description: this.options.description,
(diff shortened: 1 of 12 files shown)


libs/langchain/src/agents/middlewareAgent/ReactAgent.ts
```diff
@@ -96,30 +96,6 @@ export class ReactAgent<
     public options: CreateAgentParams<StructuredResponseFormat, ContextSchema>
   ) {
     this.#toolBehaviorVersion = options.version ?? this.#toolBehaviorVersion;
-
```
eyurtsev, line 99: Why is the name of the class ReactAgent?

libs/langchain/src/agents/middlewareAgent/ReactAgent.ts
```diff
@@ -335,8 +310,6 @@ export class ReactAgent<
      */
     this.#graph = allNodeWorkflows.compile({
       checkpointer: this.options.checkpointer ?? this.options.checkpointSaver,
-      interruptBefore: this.options.interruptBefore,
```
eyurtsev, line 338: we likely want to match js and python for now -- @sydney-runkle

eyurtsev:
Removing `llm` looks good to me

libs/langchain/src/agents/middlewareAgent/ReactAgent.ts
christian-bromann: I will sync with Sydney later on getting parity on interfaces

libs/langchain/src/agents/middlewareAgent/ReactAgent.ts
christian-bromann: no particular reason, this is not exposed directly to users.

PR #9020, opened 2025-09-20
Enhance Ollama thinking parameter. Separate thinking content from final answer content output.

JamesMustafa:
Fixes # (issue)


libs/langchain-ollama/src/chat_models.ts
```diff
@@ -56,6 +56,11 @@ export interface ChatOllamaCallOptions extends BaseChatModelCallOptions {
   tools?: BindToolsInput[];
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   format?: string | Record<string, any>;
+  /**
+   * Whether to enable thinking mode for this specific invocation.
+   * Can be a boolean (true/false) or a string intensity level ("high", "medium", "low").
+   */
+  think?: boolean | 'high' | 'medium' | 'low';
```
christian-bromann, line 63: It seems we are using this type 3 times across different places, should we create a type for this and reference instead?

libs/langchain-ollama/src/tests/chat_models_think.int.test.ts
```diff
@@ -52,3 +52,175 @@ test("test deep seek model with think=true (default)", async () => {
...
+    think: false,
+    maxRetries: 1,
+  });
+
+  // All should be properly instantiated
+  expect(ollamaHigh).toBeDefined();
+  expect(ollamaMedium).toBeDefined();
+  expect(ollamaLow).toBeDefined();
+  expect(ollamaTrue).toBeDefined();
+  expect(ollamaFalse).toBeDefined();
```
christian-bromann, line 93: I don't think this assertion provides any value

libs/langchain-ollama/src/tests/chat_models_think.int.test.ts
```diff
@@ -52,3 +52,175 @@ test("test deep seek model with think=true (default)", async () => {
...
+  
+  // Quick test that string values work in practice
+  const res = await ollamaHigh.invoke([
+    new HumanMessage({ content: "How many r in the word strawberry?" })
+  ]);
+  
+  expect(res).toBeDefined();
+  expect(typeof res.content).toBe("string");
+  expect(res.content.length).toBeGreaterThan(0);
```
christian-bromann, line 102: No reason to use the model if we just do type checks. In `v1` branch we have migrated to Vitest which allows to create type tests.

libs/langchain-ollama/src/tests/chat_models_think.int.test.ts
```diff
@@ -52,3 +52,175 @@ test("test deep seek model with think=true (default)", async () => {
...
+  expect(res.content).toBeDefined();
+  expect(res.additional_kwargs).toBeDefined();
+
+  // Validate basic functionality - response should exist and not contain thinking tags
+  expect(typeof res.content).toBe("string");
+  expect(res.content.length).toBeGreaterThan(0);
+  expect(res.content).not.toMatch(/<think>.*?<\/think>/is); // No thinking tags in content
+
+  // For string thinking levels, validate thinking content if present
+  if (res.additional_kwargs?.thinking_content) {
```
christian-bromann, line 133: Why the `if` statement? Shouldn't we expect this flag to be set? I am afraid if we introduce a bug were this becomes `false` for some reason , we stop doing all the assertions and may miss a bug.

libs/langchain-ollama/src/tests/chat_models_think.int.test.ts
JamesMustafa: Hi @christian-bromann and thanks a lot for your review and feedback! :))

Can you please clarify what would be the best approach here in your opinion? As I already do test model invocation in other test (`test string thinking parameter '$thinkLevel'`), I removed it from this one, as you suggested.

I read a bit about Vitest, and I guess we can do something like that here:
```js
test("test type safety for thinking parameter values", () => {
  expectTypeOf<ChatOllamaInput['think']>().toEqualTypeOf<boolean | ThinkingIntensity | undefined>()

  expectTypeOf<'high'>().toExtend<ThinkingIntensity>()
  expectTypeOf<'medium'>().toExtend<ThinkingIntensity>()
  expectTypeOf<'low'>().toExtend<ThinkingIntensity>()

  expectTypeOf<{ think: ThinkingIntensity }>().toExtend<Partial<ChatOllamaInput>>()
  expectTypeOf<{ think: boolean }>().toExtend<Partial<ChatOllamaInput>>()
});
```

However, Vitest is still not present inside the `package.json` file of `langchain-ollama` package. I see that it's already used in some of your internal libraries and also you plan to migrate to it, but what would be the best option currently? On one hand, I am not sure if adding `vitest` in the package.json would do any harm, on the other hand, if I try to do some workaround for type checking with Jest, this will create a technical debt for you in the future. What do you suggest?

libs/langchain-ollama/src/utils.ts
```diff
@@ -14,6 +14,9 @@ import type {
 } from "ollama";
 import { v4 as uuidv4 } from "uuid";
 
+// Track previous thinking content to calculate incremental changes
+let previousThinkingContent = "";
```
christian-bromann, line 18: I am not sure if we should use a global environment like this to store thinking tokens. What happens if someone works with multiple chat models and invokes them at the same time?

PR #9033, opened 2025-09-23
fix(langchain/createAgent): improve structured output handling

christian-bromann:
This patch updates the structured output behavior for `createAgent`:

- it now better detects models that support structured responses
- it automatically selects structured responses if user only provides a json or Zod schema
- it adds e2e tests verifying the behavior
- it generally improves the implementation, e.g. compute decision which strategy to use after the `modifyModelRequest` hook of the middleware

### DIFF
--- a/libs/langchain/src/agents/tests/reactAgent.int.test.ts
+++ b/libs/langchain/src/agents/tests/reactAgent.int.test.ts
@@ -2,7 +2,7 @@ import { describe, it, expect, vi, beforeEach } from "vitest";
 import { ChatAnthropic } from "@langchain/anthropic";
 import { ChatOpenAI } from "@langchain/openai";
 import { tool } from "@langchain/core/tools";
-import { HumanMessage, AIMessage, ToolCall } from "@langchain/core/messages";
+import { HumanMessage, AIMessage } from "@langchain/core/messages";
 import z from "zod/v3";
 
 import { createAgent, providerStrategy } from "../index.js";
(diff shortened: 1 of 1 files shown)


libs/langchain/src/agents/middlewareAgent/tests/reactAgent.int.test.ts
```diff
@@ -303,3 +304,92 @@ Please provide a clear, direct, and authoritative answer, as this information wi
     });
   });
 });
+
+describe("sturctured response format", () => {
```
hntrl:
```suggestion
describe("structured response format", () => {
```

libs/langchain/src/agents/responses.ts
```diff
@@ -426,3 +411,77 @@ export type JsonSchemaFormat = {
...
+const CHAT_MODELS_THAT_SUPPORT_JSON_SCHEMA_OUTPUT = ["ChatOpenAI", "ChatXAI"];
+const MODEL_NAMES_THAT_SUPPORT_JSON_SCHEMA_OUTPUT = [
+  "grok",
+  "gpt-5",
+  "gpt-4.1",
+  "gpt-4o",
+  "gpt-oss",
+  "o3-pro",
+  "o3-mini",
+];
```
hntrl, line 424: I'm assuming this list is incomplete? Post V1 I'd love if we did some exploration on how to make this native to the provider packages (maintaining this list separately/ any kind of "model capability matrix" feels like asking for trouble)

(no fix needed here, just a comment)

libs/langchain/src/agents/model.ts
```diff
@@ -0,0 +1,29 @@
...
+  _model: () => Promise<BaseChatModel>;
+}
+
+export function isBaseChatModel(
+  model: LanguageModelLike
+): model is BaseChatModel {
+  return (
+    "invoke" in model &&
+    typeof model.invoke === "function" &&
+    "_modelType" in model
```
hntrl: I think `_streamResponseChunks` is a better candidate for this check

libs/langchain/src/agents/responses.ts
christian-bromann: 💯 Let's create a post v1 list

PR #9062, opened 2025-09-28
docs: Update qa_chat_history tutorial to use non-deprecated method

jet-isnt-haha:
## Description

Updated the `prettyPrint` function in the qa_chat_history tutorial to use the non-deprecated `getType()` method instead of `_getType()`.

## Changes
- Changed `message._getType()` to `message.getType()` in the prettyPrint helper function
- The `_getType()` method signature is deprecated and should be replaced with `getType()`

## Type of Change
- [x] Documentation update


christian-bromann (approved):
LGTM 👍 

Thank you!

PR #9071, opened 2025-09-30
fix(langchain/createAgent): create new node for structured response

christian-bromann:
We have discovered that we can't use middleware to influence structured responses. This patch introduces a new structured response node that creates the structured response at the end of the graph instead of in the agent node.


examples/src/createAgent/middleware/structuredOutput.ts
```diff
@@ -0,0 +1,136 @@
...
+          new RemoveMessage({ id: lastMessage.id as string }),
+          new AIMessage({
+            content: JSON.stringify({
+              ...origContent,
+              email: "modified@example.com",
+            }),
+          }),
+        ],
+      };
+    }
```
christian-bromann, line 90: Seems very verbose "just" to update the structured response.

PR #9074, opened 2025-09-30
fix(langchain): only create tools node for client tools

pokey:
Fixes unreachable node error when using server-side tools (like Anthropic's
code_execution, web_search, bash) with createAgent.

Previously, the tools node was created whenever ANY tools were passed, but
only client tools (Runnables) need the tools node for local execution.
Server tools (plain objects) are executed by the provider and return results
in the message content, not as tool_calls.

Fixes # (issue)


pokey:
Done. I just copied to test over ~verbatim. Is there a better way to do it?

PR #9075, opened 2025-09-30
fix(vscode): remove experimental `useTsgo` setting

pokey:
This setting was causing VSCode typescript extension to hang on startup, so all autocomplete, hover, etc was broken


christian-bromann (approved):
LGTM 👍

PR #9086, opened 2025-10-01
fix(langchain/createAgent): don't modify messages in PromptCachingMiddleware, return new objects

christian-bromann:
Prompt caching is only possible for 5 messages at the same time. The current version of the middleware updates the message object in directly rather than creating a new object. This causes messages to stay cached over multiple hook iterations and eventually failing due to the limit of 5.

### DIFF
--- a/libs/langchain/src/agents/middlewareAgent/middleware/promptCaching.ts
+++ b/libs/langchain/src/agents/middlewareAgent/middleware/promptCaching.ts
@@ -172,7 +172,7 @@ export function anthropicPromptCachingMiddleware(
   return createMiddleware({
     name: "PromptCachingMiddleware",
     contextSchema,
-    modifyModelRequest: (options, state, runtime) => {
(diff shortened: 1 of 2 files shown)


hntrl (commented):
Do we know if the `cache_control` breakpoint only caches the content contained within a single message? Anthropic docs make it a little ambiguous..

> The system automatically checks for cache hits at all previous content block boundaries 

If we're only adding the cache_control param to the last message, and given that cache breakpoints are contained within a single message, then we aren't caching much of anything

christian-bromann:
> Do we know if the `cache_control` breakpoint only caches the content contained within a single message?

As far as I understood it caches the message flagged with the cache property and all previous ones.

> then we aren't caching much of anything

In Python we do the same here. The integration test also checks that we actually see cached tokens.

hntrl (approved):

PR #9097, opened 2025-10-02
fix(anthropic): add support for container upload block type in message formatting

pokey:
We currently silently drop all content blocks of type `"container_upload"`, which are used to make files accessible from Anthropic beta code execution containers. This PR adds support for them

See [Anthropic code execution docs](https://docs.claude.com/en/docs/agents-and-tools/tool-use/code-execution-tool#upload-and-analyze-files) for more info

- Needed by https://github.com/pokey/langchainjs/pull/1


libs/providers/langchain-anthropic/src/utils/message_inputs.ts
```diff
@@ -271,6 +272,15 @@ function* _formatContentBlocks(
...
       } as any;
+    } else if (contentPart.type === "container_upload") {
+      yield {
+        ...contentPart,
+        ...(cacheControl ? { cache_control: cacheControl } : {}),
+      } as AnthropicContainerUploadBlockParam;
+    } else {
+      console.warn(
+        `Skipping unsupported content block of type "${contentPart.type}".`
+      );
```
pokey: One big gotcha here for me was that we silently drop unknown blocks. I turned it into a warning, but my instinct would be to warn and then just pass them through untouched. Any thoughts?

libs/providers/langchain-anthropic/src/types.ts
```diff
@@ -48,6 +48,8 @@ export type AnthropicWebSearchResultBlockParam =
 // TODO(hntrl): beta blocks should be separated
 export type AnthropicSearchResultBlockParam =
   Anthropic.Beta.BetaSearchResultBlockParam;
+export type AnthropicContainerUploadBlockParam =
+  Anthropic.Beta.BetaContainerUploadBlockParam;
```
pokey, line 52: I followed the current protocol here and removed the `Beta` from the type name, but it's not obvious to me that's the best thing to do, as we're taking something that is beta from Anthropic's perspective and marking it as non-beta on our side

libs/providers/langchain-anthropic/src/utils/message_inputs.ts
christian-bromann: I would throw here because it means we not fully support Anthropic's API which causes such unknown errors.

christian-bromann (approved):
I am in favor for throwing an error but wouldn't consider this blocking.

Thanks 👍

libs/providers/langchain-anthropic/src/types.ts
hntrl: Agree this isn't totally ideal. There's a ton of enhancements like this that we can make. We're only really using this as an internal type alias so I'm not as concerned about making this _semantically accurate_ so to speak

libs/providers/langchain-anthropic/src/utils/message_inputs.ts
hntrl: One consideration is that message .content looks subtly different between providers, which means if I used openai generated messages in an anthropic call and had the content fall through we'd always get 400 errors. I also don't think it's great DX to flood the console with these warning messages if we know that some messages aren't compatible with anthropic to boot.

Generally when serializing provider requests, we take the approach of "best effort" parsing where we serialize messages for a request to the best of our ability (typically by having explicit conversion for messages that come out of the same provider + some other well known formats), but we don't explicitly fail since we can't interpret every possible message format (n^n complexity)

this is part of the motivation for standard content in v1!

hntrl (commented):
thank you for catching this! A few things:
 
* an integration test would be much appreciated 🙏 
* I detailed this a bit in the comments, but we should remove the warning

libs/providers/langchain-anthropic/src/utils/message_inputs.ts
pokey: I see. That makes sense, though today's behaviour of just dropping the messages is a bit surprising. I wonder if we should coerce it into a text block or something. Or have a flag to govern behaviour, because it would be useful during dev to fail fast but then be robust in prod

Anyways I'll remove the warning from this PR as it sounds like there's a bigger conversation to be had here. Shall I open a github discussion?

pokey: Ok I've replaced the warning with a comment explaining why we silently drop blocks ofa unexpected type, and added a test to capture that behaviour

pokey:
Integration test added and warning removed; should be good to merge

hntrl (approved):

pokey:
Any idea why CI is failing? Doesn't look related to my change but can't say for certain

hntrl:
yep I definitely broke it

https://github.com/langchain-ai/langchainjs/pull/9113

PR #9102, opened 2025-10-03
chore(*): remove deprecations

hntrl:
anything that was marked as @deprecated is being removed here


libs/langchain-core/src/utils/env.ts
```diff
@@ -63,13 +63,6 @@ export type RuntimeEnvironment = {
...
 let runtimeEnvironment: RuntimeEnvironment | undefined;
 
-/**
- * @deprecated Use getRuntimeEnvironmentSync instead
- */
-export async function getRuntimeEnvironment(): Promise<RuntimeEnvironment> {
-  return getRuntimeEnvironmentSync();
-}
-
 export function getRuntimeEnvironmentSync(): RuntimeEnvironment {
```
jacoblee93: Let's just rename this to `getRuntimeEnvironment`?

hntrl: shore

PR #9106, opened 2025-10-03
fix(CONTRIBUTING.md): update directory path for langchain-core build instructions

pokey:
It appears it was using the old location of langchain-core


hntrl (approved):
thanks for catching!

PR #9108, opened 2025-10-03
fix(anthropic): Add execution tool response block types

pokey:
Adds `bash_code_execution_tool_result` and `text_editor_code_execution_tool_result`

See https://docs.claude.com/en/docs/agents-and-tools/tool-use/code-execution-tool

- Needed by https://github.com/pokey/langchainjs/pull/1

PR #9109, opened 2025-10-03
feat(anthropic): Add support for container reuse for code execution

pokey:
Adds `container` parameter to `ChatAnthropicCallOptions` and passes it through to Anthropic client.

See https://docs.claude.com/en/docs/agents-and-tools/tool-use/code-execution-tool#container-reuse

### DIFF
--- a/libs/providers/langchain-anthropic/src/tests/chat_models.test.ts
+++ b/libs/providers/langchain-anthropic/src/tests/chat_models.test.ts
@@ -191,3 +191,41 @@ test("Can properly format anthropic messages when given two tool results", async
     system: undefined,
   });
 });
+
+test("invocationParams includes container when provided in call options", () => {
+  const model = new ChatAnthropic({
+    modelName: "claude-3-haiku-20240307",
+    temperature: 0,
+    anthropicApiKey: "testing",
+  });
+
+  const params = model.invocationParams({ container: "container_123" });
+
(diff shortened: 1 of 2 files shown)


pokey:
Do we want an integration test?

christian-bromann (approved):
LGTM 👍

christian-bromann:
> Do we want an integration test?

If you can come up with an easy one that validates the functionality and is stable, let's do it!

libs/providers/langchain-anthropic/src/tests/chat_models.test.ts
hntrl, line 1: am fine to keep this, but as commentary I'm not sure how much benefit we get from testing `invocationParams` directly (it's only part of the messages + options -> sdk payload "chain")

ideal state is the conversion logic of messages + options is collapsed into something like `getRequestPayload` and that's what we test (this is how python does it iirc). Not needed for this PR though.

ok I'll get off my soap box now

pokey:
I have a forthcoming PR that will add integration tests for this PR as well as #9108, because the end-to-end won't work without both changes and another forthcoming change.

Which argues that maybe I went a bit too fine-grained with my PRs here 😅. Hopefully made them nice and easy to review tho

Are we ok to merge this one without the integration test and I'll fast follow with the integration test?

PR #9110, opened 2025-10-03
feat(langchain/createAgent): Add support for callOptions to middleware

pokey:
This allows us to pass things like `headers` and `container` to the model call

### DIFF
--- a/libs/langchain/src/agents/middlewareAgent/types.ts
+++ b/libs/langchain/src/agents/middlewareAgent/types.ts
@@ -152,6 +152,11 @@ export interface ModelRequest {
    * The tools to make available for this step.
    */
   tools: string[];
+
+  /**
+   * Extra call options to pass to the model.
+   */
+  callOptions?: Record<string, unknown>;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


christian-bromann (approved):
LGTM 👍 

Let's sync with the Python team on the new property on `ModelRequest`

pokey:
As discussed with @eyurtsev, we're holding off on merging this one until we look into Python version

libs/langchain/src/agents/middlewareAgent/types.ts
```diff
@@ -152,6 +152,11 @@ export interface ModelRequest {
    * The tools to make available for this step.
    */
   tools: (ServerTool | ClientTool)[];
+
+  /**
+   * Extra call options to pass to the model.
+   */
+  callOptions?: Record<string, unknown>;
```
eyurtsev: cc @christian-bromann  / @pokey is this a standard name in js?

@sydney-runkle this is extra config that we'll be passing to chat model invoke. We can either use the same standard name if it makes sense or follow whatever is convention in each language. 


* extra_params
* invoke_kwargs

christian-bromann: to be more specific here I would suggest `modelCallOptions` as an alternative.

pokey:
As proposed in Slack, I've switched to just implementing `modelSettings` as we have in Python. This version works fine for my use case

PR #9114, opened 2025-10-04
fix(langchain-core): handle missing parent runs in tracer to prevent LangSmith 400 errors

### LINKED ISSUE #9077: Un-awaited LLM call triggers `invalid 'dotted_order'` during LangSmith trace
### Checked other resources

### Example Code

https://github.com/yuku-sandbox/langsmith-dotted-order-repro

```
Failed to Failed to send multipart request. Received status [400]: Bad Request. Message: {"error":"Bad request: invalid 'dotted_order': dotted_order must contain at least two parts for child runs for run_id:0c84e859-506a-44f3-9424-5e5fd10e0c75 trace_id:0c84e859-506a-44f3-9424-5e5fd10e0c75 dotted_order:20250930T162253554001Z0c84e859-506a-44f3-9424-5e5fd10e0c75 parent_run_id:a54cdf92-665b-41fd-b328-ac6291b94a7d"}

Context: trace=0c84e859-506a-44f3-9424-5e5fd10e0c75,id=0c84e859-506a-44f3-9424-5e5fd10e0c75
```

### Description

When an LLM is called without `await` inside LangGraph nodes, the parent run may have already disappeared from the memory by the time the LLM call is actually executed, which leads the invalid format of `dotted_order`.

### System Info

```
$  bun pm ls | grep langchain
[0.32ms] ".env"
├── @langchain/core@0.3.78
├── @langchain/langgraph@0.4.9
├── @langchain/openai@0.6.14
├── langchain@0.3.35

$ node -v
v22.14.0
```

yukukotani:
Fixes #9077

In some cases, the parent run is [cleaned up](https://github.com/yukukotani/langchainjs/blob/debe7c6419a3da783c45114e6cb15b173e881f02/langchain-core/src/tracers/base.ts#L202-L206) before its child run, typically when the child run is triggered as an async background call.

Ideally, the parent run should remain in memory until all child runs have completed. One possible approach could be to manage this with a WeakMap, but that would require significant changes.

As a temporary workaround, this fix isolates the child run to prevent the issue.

### DIFF
--- a/.changeset/tall-jeans-drop.md
+++ b/.changeset/tall-jeans-drop.md
@@ -1,0 +1,5 @@
+---
+"@langchain/core": patch
+---
+
+handle missing parent runs in tracer to prevent LangSmith 400 errors


christian-bromann:
Really appreciate you taking the time to open this PR, @yukukotani 🙏

jacoblee93:
Thanks @yukukotani! I think this is ok yeah - something like the below may help you as well:

```ts
import { withRunTree, getCurrentRunTree } from "langsmith/traceable";

RunnableLambda.from(async () => {
  const currentRunTree = getCurrentRunTree();
  void withRunTree(currentRunTree, async () => {
    // async work
  });
});
```

> One possible approach could be to manage this with a WeakMap, but that would require significant changes.

Yeah this is interesting too - CC @dqbd

christian-bromann:
@yukukotani I wasn't able to rebase/update your branch myself. Could you please do that? Happy to merge after.

yukukotani:
@hntrl Thank you for handling this!

PR #9119, opened 2025-10-04
fix(langchain/createAgent): tools in ModelRequest back to tool instance

christian-bromann:
We've decided to keep `tools` in `ModelRequest` a list of tool instances so users have access on more tool information. This also removes `tools` and `toolCalls` from runtime as it differs from our middleware designs.

### DIFF
--- a/libs/langchain/src/agents/middlewareAgent/types.ts
+++ b/libs/langchain/src/agents/middlewareAgent/types.ts
@@ -151,7 +151,7 @@ export interface ModelRequest {
   /**
    * The tools to make available for this step.
    */
-  tools: string[];
+  tools: (ServerTool | ClientTool)[];
 }
 
 /**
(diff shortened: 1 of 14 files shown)


libs/langchain/src/agents/middlewareAgent/types.ts
```diff
@@ -151,7 +151,7 @@ export interface ModelRequest {
   /**
    * The tools to make available for this step.
    */
-  tools: string[];
+  tools: (ServerTool | ClientTool)[];
```
eyurtsev, line 154: Does JS already define ServerTools?

christian-bromann: Not really it is very broad defined, e.g. as `Record<string, unknown>`

Issue #9100, opened 2025-10-02
Calling `withStructuredOutput` with schema modifies schema and removes `transform()` configurations


ro0sterjam:
### Checked other resources

- [x] This is a bug, not a usage question. For questions, please use the LangChain Forum (https://forum.langchain.com/).
- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain.js documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain.js rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```typescript
import { ChatPromptTemplate } from "@langchain/core/prompts";
import { ChatOpenAI } from "@langchain/openai";
import { z } from "zod/v4";

const schema = z.object({
  message: z.string().transform((x) => x.toUpperCase()),
});

const schemaUntouched = z.object({
  message: z.string().transform((x) => x.toUpperCase()),
});

const chatbot = new ChatOpenAI({ temperature: 0 });

const messages = ["Hello"];
const prompt = await ChatPromptTemplate.fromMessages(messages);

console.log(JSON.stringify(schema) === JSON.stringify(schemaUntouched)); // logs true

const chatbotWithStructuredOutput = chatbot.withStructuredOutput(schema, {
  includeRaw: true,
});

console.log(JSON.stringify(schema) === JSON.stringify(schemaUntouched)); // logs false

const runnable = prompt.pipe(chatbotWithStructuredOutput);
const result = await runnable.invoke({});

console.log(schema.parse(result.parsed)); // Logs { message: 'Hello' }
console.log(schemaUntouched.parse(result.parsed)); // Logs { message: 'HELLO' }
```

### Description

When I pass a schema to `withStructuredOutput`, it appears to modify the schema itself, in my particular case it is removing any `transform()` methods I have configured on the schema.

The example demonstrates that calling parse from the passed in schema does not perform the transform, but calling parsed on the untouched copy of the schema does.

Additionally you can see via the logs before and after calling `withStructuredOutput` that the schema is in fact mutated.

### System Info

Langchain 0.3.35
Node v24.5.0
OSX 15.6.1 (24G90)

ro0sterjam:
[...]

PR #9125, opened 2025-10-05
fix(langchain): Fix toJsonSchema mutating underlying zod schema

ro0sterjam:
In `interopZodTransformInputSchema`, need to make a shallow copy of the schema before replacing the properties rather than replacing in place to prevent mutation of underlying schema.

Doing so however breaks the OpenAI integration because it takes advantage of the underlying schema being mutated as part of the first call to `toJsonSchema`. Thus when it reaches `toJSONSchemaV4` from zod, it already assumes that the transforms have been stripped.

Since this PR fixes the underlying mutation, this assumption no longer holds, so I have to manually make sure that the downstream schema passed to zod's `toJSONSchemaV4` runs through the transform stripping logic.

Note: I was unable to determine how best to write a unit/integration test for OpenAI to ensure that it works with `transform` on schemas. But I was able to validate live.

Fixes https://github.com/langchain-ai/langchainjs/issues/9100

### DIFF
--- a/.changeset/hungry-dolls-turn.md
+++ b/.changeset/hungry-dolls-turn.md
@@ -1,0 +1,6 @@
+---
+"@langchain/core": patch
+"@langchain/openai": patch
+---
+
+fix(langchain): Fix toJsonSchema mutating underlying zod schema


langchain-core/src/utils/json_schema.ts
```diff
@@ -22,16 +20,8 @@ export { deepCompareStrict, Validator } from "@cfworker/json-schema";
...
   if (isZodSchemaV4(schema)) {
-    const inputSchema = interopZodTransformInputSchema(schema, true);
-    if (isZodObjectV4(inputSchema)) {
-      const strictSchema = interopZodObjectStrict(
-        inputSchema,
-        true
-      ) as ZodObjectV4;
-      return toJSONSchema(strictSchema);
-    } else {
-      return toJSONSchema(schema);
```
ro0sterjam, line 33: Am I correct to assume that `toJSONSchema(schema)` was a typo and that it should have been `toJSONSchema(inputSchema)`?

If not, please let me know, as this PR effectively makes that change.

langchain-core/src/utils/types/zod.ts
```diff
@@ -748,7 +750,9 @@ export function interopZodTransformInputSchema(
     if (recursive) {
       // Handle nested object schemas
       if (isZodObjectV4(outputSchema)) {
-        const outputShape: Mutable<z4.$ZodShape> = outputSchema._zod.def.shape;
+        const outputShape: Mutable<z4.$ZodShape> = {
+          ...outputSchema._zod.def.shape,
+        };
```
ro0sterjam, line 755: Primary fix here.

langchain-core/src/utils/types/zod.ts
```diff
@@ -781,3 +785,27 @@ export function interopZodTransformInputSchema(
...
+
+/**
+ * Sanitizes a Zod schema by transforming it to its input schema and then making it strict.
+ * Supports both Zod v3 and v4 schemas. If `recursive` is true, applies strictness recursively to all nested object schemas and arrays of object schemas.
+ *
+ * @param schema - The Zod schema instance (v3 or v4)
+ * @param {boolean} [recursive=false] - Whether to recursively process nested objects/arrays.
+ * @returns The sanitized Zod schema.
+ */
+export function interopZodSanitizeSchema(
```
ro0sterjam, line 797: Extracted so that it can be called independently [here](https://github.com/langchain-ai/langchainjs/pull/9125/files#diff-fddcb19e8c7c2718f4e2d6476956f4c3d5b3bc9c020ec5ca28301d04c0bab974R1246).

christian-bromann:
Really appreciate you taking the time to open this PR, @ro0sterjam 🙏

libs/langchain-core/src/utils/types/zod.ts
```diff
@@ -814,7 +814,7 @@ export function interopZodTransformInputSchema(
     if (recursive) {
       // Handle nested object schemas
       if (isZodObjectV4(outputSchema)) {
-        const outputShape: Mutable<z4.$ZodShape> = outputSchema._zod.def.shape;
+        const outputShape: Mutable<z4.$ZodShape> = {};
```
ro0sterjam, line 819: primary fix here

libs/providers/langchain-openai/src/utils/output.ts
```diff
@@ -104,7 +105,7 @@ export function interopZodResponseFormat(
           ...props,
           name,
           strict: true,
-          schema: toJSONSchemaV4(zodSchema, {
+          schema: toJsonSchema(zodSchema, {
```
ro0sterjam, line 108: Call the version of `toJsonSchema` that also strips off transforms, since we can no longer assume that that has already been done for us by this point.

libs/langchain-core/src/utils/json_schema.ts
```diff
@@ -16,22 +16,30 @@ import {
...
 /**
  * Converts a Zod schema or JSON schema to a JSON schema.
  * @param schema - The schema to convert.
+ * @param _params - The parameters to pass to the toJSONSchema function.
  * @returns The converted schema.
  */
-export function toJsonSchema(schema: InteropZodType | JSONSchema): JSONSchema {
+export function toJsonSchema(
+  schema: InteropZodType | JSONSchema,
+  _params?: ToJSONSchemaParams
```
christian-bromann: Since we are using `params` in the function, no need to have it start with `_`
```suggestion
  params?: ToJSONSchemaParams
```

ro0sterjam: Fair point. I only named it such as that's what zod's method def uses:

```
export declare function toJSONSchema(schema: schemas.$ZodType, _params?: ToJSONSchemaParams):
```

But I agree we should just use without the `_` here.

libs/langchain-core/src/utils/json_schema.ts
```diff
@@ -16,22 +16,30 @@ import {
...
     const inputSchema = interopZodTransformInputSchema(schema, true);
     if (isZodObjectV4(inputSchema)) {
       const strictSchema = interopZodObjectStrict(
         inputSchema,
         true
       ) as ZodObjectV4;
-      return toJSONSchema(strictSchema);
+      return toJSONSchema(strictSchema, params);
     } else {
-      return toJSONSchema(schema);
```
ro0sterjam, line 34: Incidentally, I'm wondering if the intended param here wasn't originally supposed to be `inputSchema` rather than `schema`?

PR #9126, opened 2025-10-05
fix(google-common): set function calling mode to auto when responseMimeType is application/json

athan37:
Fixes function calling mode to use 'auto' instead of 'any' when responseMimeType is set to 'application/json' for optimal structured output performance.

## Changes
- Modified `formatToolConfig` in `gemini.ts` to dynamically set function calling mode based on `responseMimeType`
- Enhanced `withStructuredOutput` in `chat_models.ts` to automatically set `responseMimeType: "application/json"`
- Added comprehensive unit tests covering both default and override scenarios

## Testing
- Added 2 unit tests verifying the fix works correctly
- Tests cover both scenarios: default behavior and user override
- All tests pass with proper mock data

## Impact
- Improves performance for structured output scenarios
- Prevents 400 errors from Google API when using `responseMimeType: "application/json"`
- Maintains backward compatibility


christian-bromann:
Really appreciate you taking the time to open this PR, @athan37 🙏

athan37:
> Part of this was a bit unintuitive when reviewing -- google rolled out a responseMimeType field on generation configs that strongly dictates the response on google's end. We depend on forced function calling here in order to render structured output, but us pinning an additional responseMimeType onto call options here would suggest that we have support for this structured output scheme (we don't). There seems to be a more esoteric bug here where the default responseMimeType isn't getting pinned to the request generationConfig when forcing function calls like this.
> 
> How are you calling this model? Are you setting `application/json` somewhere?

Thanks for raising this concern. You're right that there's a deeper issue here.

The 400 error occurs because Google requires `mode="AUTO"` when `responseMimeType = "application/json"`. However, there's a more critical production bug. 

Without `responseMimeType="application/json"`, Gemini frequently returns invalid or truncated JSON for large structured outputs, with failures typically beginning around 8k tokens and becoming severe as responses approach the 65k token limit. 

=> This makes `withStructuredOutput()` fundamentally unreliable in production environments. 

Setting `responseMimeType="application/json"` enables Google's native JSON mode, which guarantees well-formed JSON at scale and resolves these failures. 

* The correct design would be to use Google's native `responseSchema` for structured output, rather than injecting an "extraction" tool for schema definition. The issue is that `responseSchema` isn't wired up in LangChain yet.

* The longer-term fix would be to add `responseSchema` support to `GeminiGenerationConfig`, handle schema conversion, and update `withStructuredOutput()` to use that native Google approach.

I’m happy to expand the scope of this PR to incorporate Gemini’s native `responseSchema` support if that’s preferred.

PR #9133, opened 2025-10-08
fix(langchain/createAgent): merge middleware agent with original react agent

christian-bromann:
We are moving forward with the middleware concept of the agent and merging everything together to avoid confusions, having two different agents.


hntrl (approved):
nice

PR #9164, opened 2025-10-13
chore(langchain): move middleware primitives to core

christian-bromann:
First pass moving agent primitives into `@langchain/core` to enable provider packages to create their own middleware.


christian-bromann:
It works but not moving forward with it.

PR #9165, opened 2025-10-13
fix(mcp-adapters): preserve timeout from RunnableConfig in MCP tool calls

### LINKED ISSUE #9136: Long timeouts on MCP calls ignored
I'm using an MCP server tool that can take several minutes to complete. 

I tried passing the timeout as part of the call, as well as setting defaultToolTimeout on the MCP server definition.

Langchain is setting the timeout, but it is then intercepting the timeout value and setting up its own timeout signal but isn’t passing the timeout value to the MCP call. As a result, the underlying library times out the request after the default 60 seconds.

The issue seems to be that the "timeout" value is removed from the configuration passed to the MCP tool call:
- https://github.com/langchain-ai/langchainjs/blob/main/langchain-core/src/runnables/config.ts#L191

Simply removing this line fixes the issue, but I'm not sure what other side effects that change might have.

pawel-twardziak:
The timeout value from RunnableConfig was being lost when calling MCP tools because ensureConfig() converts it to an AbortSignal and removes the numeric timeout field. This prevented the MCP SDK from receiving explicit timeout values.

Changes:
  - Store original timeout in metadata.timeoutMs in ensureConfig()
  - Retrieve and pass numeric timeout to MCP SDK RequestOptions in _callTool()
  - Add integration test verifying per-call timeout is properly honored

This ensures MCP tools respect timeouts passed through RunnableConfig, allowing callers to override default timeouts on a per-call basis.

**Why deleting timeout is necessary** (as it is so far)

- **Idempotent normalization**: ensureConfig may be called multiple times down the stack. If timeout remains, each call would synthesize new timeout signals and combine them, changing the effective timeout unpredictably.
- **Single enforcement path**: downstream code relies on signal to enforce cancellation. Leaving timeout means two competing mechanisms (numeric timeout and signal) can be applied, sometimes with different semantics.
- **Propagation to children**: pickRunnableConfigKeys would keep forwarding timeout to nested runnables, causing repeated re-normalization and stacked timeouts.
- **Backward compatibility**: a lot of components and tests assume ensureConfig removes timeout post-normalization; changing that would be a breaking change.

If an integration needs the numeric value later (e.g., to pass to an SDK that expects timeout), preserve it out-of-band (e.g., via metadata) and still use signal for actual cancellation.

Fixes #9136

### DIFF
--- a/.changeset/stale-carrots-trade.md
+++ b/.changeset/stale-carrots-trade.md
@@ -1,0 +1,6 @@
+---
+"@langchain/core": patch
+"@langchain/mcp-adapters": patch
+---
+
+fix(mcp-adapters): preserve timeout from RunnableConfig in MCP tool calls


KurtzOmer:
Hey @pawel-twardziak , is this one going to be merged soon?
Also, is there a temp workaround to actually set timeout for tool calls?

pawel-twardziak:
hi @KurtzOmer, it is a good question for @hntrl  :)

libs/langchain-core/src/runnables/config.ts
```diff
@@ -176,7 +176,18 @@ export function ensureConfig<CallOptions extends RunnableConfig>(
...
+    // Preserve the numeric timeout for downstream consumers that need to pass
+    // an explicit timeout value to underlying SDKs in addition to an AbortSignal.
+    // We store it in metadata to avoid changing the public config shape.
+    if (!empty.metadata) {
+      empty.metadata = {};
+    }
+    // Do not overwrite if already set upstream.
+    if (empty.metadata.__lc_originalTimeoutMs === undefined) {
+      empty.metadata.__lc_originalTimeoutMs = originalTimeoutMs;
+    }
```
hntrl, line 190: could we just call this `timeoutMs`?

libs/langchain-mcp-adapters/src/tests/client.test.ts
hntrl, line 1: Would maybe add a smoke test that the timeout is honored (e.g. setting timeout to a lower value then sleepMsec also)

pawel-twardziak:
@hntrl sure, will do, thanks for your review!

libs/langchain-core/src/runnables/config.ts
pawel-twardziak: done

libs/langchain-mcp-adapters/src/tests/client.test.ts
pawel-twardziak: done

pawel-twardziak:
@hntrl changes made :)

christian-bromann:
@pawel-twardziak since you already did the research, mind adding a comment documenting why `delete empty.timeout;` is there?

pawel-twardziak:
@christian-bromann sure thing, but this requires me to retrace my train of thought over the code :) It's because I don't actually remember why it should stay there. What I remember is that I had tried to fix the issue by removing that `delete`, but I ended up in a situation where something else got messed up. I need to figure out again what that "something else" was.

pawel-twardziak:
@christian-bromann it just dawned on me that the comment about why `delete empty.timeout;` is actually in the description of this PR :) Documentation rules!! When human memory fails :)

KurtzOmer:
It got fixed indeed, thanks very much @pawel-twardziak , it was so annoying! 🙏

pawel-twardziak:
my pleasure @KurtzOmer :) great it works!

PR #9168, opened 2025-10-14
fix(anthropic/v1): better multimodal handling

hntrl:


libs/providers/langchain-anthropic/src/utils/standard.ts
```diff
@@ -75,12 +78,95 @@ function _formatStandardCitations(
   return Array.from(iterateAnnotations());
 }
 
+function _formatBase64Data(data: string | Uint8Array): string {
+  return typeof data === "string" ? data : Buffer.from(data).toString("base64");
```
christian-bromann: Let's not use Buffer here as it is a Node.js primitive, instead `btoa` will do

PR #9177, opened 2025-10-14
fix(anthropic): Add integration test for Anthropic multi-step code execution

pokey:
Adds an integration test that ensures we can preserve container across multiple Anthropic code execution tool calls and pass their code execution tool call messages back to them. This integration test exercises the following PRs:

- #9108
- #9109

### DIFF
--- a/libs/providers/langchain-anthropic/src/utils/message_inputs.ts
+++ b/libs/providers/langchain-anthropic/src/utils/message_inputs.ts
@@ -81,36 +81,6 @@ function _formatImage(imageUrl: string) {
   );
 }
 
-/**
- * Work around Anthropic API bug where bash_code_execution_output blocks
- * cause 500 errors when included in conversation history.
- * Filters out bash_code_execution_output blocks from bash_code_execution_tool_result blocks.
- *
- * @param contentPart The content block to potentially filter
- */
-function _filterBashCodeExecutionOutputs(contentPart: { type: string }): void {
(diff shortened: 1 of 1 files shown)


libs/providers/langchain-anthropic/src/tests/chat_models-code_execution.int.test.ts
```diff
@@ -86,26 +91,64 @@ describe("Files API with Code Execution", () => {
...
+    expect(secondResult).toBeInstanceOf(AIMessage);
+    const secondContent = Array.isArray(secondResult.content)
+      ? secondResult.content
+      : [];
+    expect(secondContent.length).toBeGreaterThan(0);
+
+    // Verify the same container was reused
+    const secondContainer = (secondResult as AIMessage).additional_kwargs
+      ?.container as { id: string; expires_at: string } | undefined;
+    expect(secondContainer?.id).toBe(container?.id);
```
christian-bromann, line 150: Can we verify that either `secondContainer` or `container` is defined?

pokey: That is actually checked [above](https://github.com/langchain-ai/langchainjs/pull/9177/files#diff-9734a90d8c7d72449a0d9a1329810acd2a175eb3f776192bc4e6af597bfca401R119), when `container` is first defined (line 119)

PR #9179, opened 2025-10-14
feat(@langchain/core): migrate subset of agent primitives to `@langchain/core`

christian-bromann:
This allows to create middleware without depending on `langchain` or `@langchain/langgraph`.


dependency_range_tests/scripts/langchain/node/update_resolutions_lowest.js
hntrl, line 1: I think this is just invalidating the point of dep tests. Fine to either disable for this pass, but I think the fix for the latest tests is to allow for pre-release versions if thats what's in dependencies (since we're just `pnpm add @langchain/core`'ing within the script)

libs/langchain-core/src/middleware/types.ts
```diff
@@ -190,15 +278,31 @@ export interface AgentMiddleware<
...
-      TFullContext
+      Runtime<
+        TContextSchema extends InteropZodObject
+          ? InferInteropZodOutput<TContextSchema>
+          : TContextSchema extends InteropZodDefault<any>
+          ? InferInteropZodOutput<TContextSchema>
+          : TContextSchema extends InteropZodOptional<any>
+          ? Partial<InferInteropZodOutput<TContextSchema>>
+          : never
+      >
```
hntrl, line 298: is it possible to dedupe the `TContextSchema extends InteropZodObject...` closures into a utility type?

libs/langchain/src/agents/middleware/tests/promptCaching.test.ts
```diff
@@ -13,10 +14,20 @@ import {
...
 
+/**
+ * Cast the promptCachingMiddleware to `any` to avoid the following error:
+ * > Type instantiation is excessively deep and possibly infinite.
+ *
+ * This is because the actual `promptCachingMiddleware` is implemented within
+ * the `@langchain/anthropic` package and tesed in here. To avoid circular dependencies,
+ * we have to run the unit and integration tests here as we need `createAgent`.
+ */
+const anthropicPromptCachingMiddleware = promptCachingMiddleware as any;
```
hntrl, line 29: this is surprising to me actually. Can get that rolldown is detecting circular dependencies, but typescript is throwing errors like this also? My intuition would be that there's no circular type trees since anthropic is only a dev dependency

libs/langchain/src/agents/middleware/tests/promptCaching.int.test.ts
hntrl, line 1: should these prompt caching tests just be living inside of `@langchain/anthropic`?

libs/providers/langchain-anthropic/package.json
```diff
@@ -30,7 +30,8 @@
     "format:check": "prettier --config .prettierrc --check \"src\""
   },
   "dependencies": {
-    "@anthropic-ai/sdk": "^0.65.0"
+    "@anthropic-ai/sdk": "^0.65.0",
+    "zod": "^3.25.76"
```
hntrl:
```suggestion
    "zod": "^3.25.76 || ^4"
```

dependency_range_tests/scripts/langchain/node/update_resolutions_lowest.js
christian-bromann: yes, ideally we need to be able to push dev releases from branches and then adjust the peer dep in `langchain`.

libs/langchain/src/agents/middleware/tests/promptCaching.test.ts
christian-bromann: For TypeScript it seems to not matter if it is a dev, optional or peer dependency.

libs/langchain/src/agents/middleware/tests/promptCaching.int.test.ts
christian-bromann: Ideally yes, but they use `createAgent` which lives in `langchain` so we are having the same problem. This brings me back of saying: let's collapse all into one langchain package.

libs/langchain-core/src/middleware/types.ts
christian-bromann: Will do in a separate PR.

christian-bromann:
Moving back to draft due to:
```
examples:build: src/createAgent/middleware/promptCaching.ts:13:16 - error TS2589: Type instantiation is excessively deep and possibly infinite.
examples:build: 
examples:build: 13   middleware: [promptCachingMiddleware({ ttl: "5m" }) as AgentMiddleware],
examples:build:                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
examples:build: 
```

PR #9194, opened 2025-10-14
feat(google/v1): V1 testing and support for google-common based packages

afirstenberg:
This ensures that the google-common based libraries will work with v1 libraries. Primary tasks were:
* Move from jest to vitest
* Dependencies on langchain-core v1.alpha.6
* Remove and restructure some calls that no longer worked under v1
* Remove legacy Gemini models

(Review of what I've done so far appreciated @hntrl @christian-bromann to make sure this is on the right track for v1.)

TODO:
* Make sure all the changes through langchain-google-common 0.2.18 are included (v1 seems to have been based on 0.2.16)

Desirable:
* Add `ContentBlock` support
* Add Nano-Banana support


libs/providers/langchain-google-common/package.json
hntrl: these scripts will need to be updated

libs/providers/langchain-google-gauth/src/tests/chat_models.int.test.ts
```diff
@@ -28,7 +28,9 @@ import { BlobStoreGoogleCloudStorage } from "../media.js";
 
 describe("GAuth Chat", () => {
   test("invoke", async () => {
-    const model = new ChatGoogle();
+    const model = new ChatGoogle({
+      modelName: "gemini-2.0-flash",
```
hntrl, line 32: not blocking, but we're preferring `model` for this field name

libs/providers/langchain-google-common/package.json
afirstenberg: Missed that one. Thank you for catching.
Updated these scripts.

hntrl:
Closing this in favor of other google efforts

PR #9208, opened 2025-10-16
fix(anthropic): support Haiku 4.5 topP parameter

bysxx:
## Description

This PR fixes a compatibility issue with Claude Haiku 4.5 models where the default `topP` parameter value of `-1` was causing a `BadRequestError` when making API calls.

## Issue

Claude Haiku 4.5 models throw the following error when used with the current configuration:
```
BadRequestError: 400 {"type":"error","error":{"type":"invalid_request_error","message":"top_p cannot be set to -1 for this model."}}
```

## Root Cause

The issue occurred because:
1. `topP` defaults to `-1` in the constructor
2. Only `opus-4-1` and `sonnet-4-5` models were configured to use `undefined` for `topP`
3. **Haiku 4.5 was missing** from this configuration, causing the API to receive `top_p: -1` which is not supported

According to [Anthropic's Claude 4.5 migration guide](https://docs.anthropic.com/claude/docs/claude-4-5-migration), Claude 4.5 models (including Haiku 4.5) require either `temperature` OR `top_p` to be used, but not both, and `top_p` cannot be set to `-1`.

## Changes

- Added `haiku-4-5` to the list of models that default `topP` to `undefined` in two locations:
  1. Constructor logic (lines 735-740)
  2. `invocationParams` thinking logic (line 854)
- Updated comments to reflect that Haiku 4.5 is now included in Claude 4 models

## Testing

- ✅ Haiku 4.5 models now work without the `top_p` error
- ✅ Existing Opus 4.1 and Sonnet 4.5 behavior remains unchanged  
- ✅ Other Claude models continue to work as expected
- ✅ No breaking changes to existing APIs

## Backward Compatibility

This change is fully backward compatible - it only adds support for the previously unsupported Haiku 4.5 model without modifying behavior for existing models.

### DIFF
--- a/libs/providers/langchain-anthropic/src/chat_models.ts
+++ b/libs/providers/langchain-anthropic/src/chat_models.ts
@@ -732,8 +732,8 @@ export class ChatAnthropicMessages<
 
     this.invocationKwargs = fields?.invocationKwargs ?? {};
 
-    if (this.model.includes("opus-4-1") || this.model.includes("sonnet-4-5")) {
-      // Default to `undefined` for `topP` for Opus 4.1 models
+    if (this.model.includes("opus-4-1") || this.model.includes("sonnet-4-5") || this.model.includes("haiku-4-5")) {
+      // Default to `undefined` for `topP` for Claude 4 models (Opus 4.1, Sonnet 4.5, Haiku 4.5)
       this.topP = fields?.topP === null ? undefined : fields?.topP;
     } else {
       this.topP = fields?.topP ?? this.topP;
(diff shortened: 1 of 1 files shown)


libs/providers/langchain-anthropic/src/chat_models.ts
```diff
@@ -732,8 +732,8 @@ export class ChatAnthropicMessages<
 
     this.invocationKwargs = fields?.invocationKwargs ?? {};
 
-    if (this.model.includes("opus-4-1") || this.model.includes("sonnet-4-5")) {
-      // Default to `undefined` for `topP` for Opus 4.1 models
+    if (this.model.includes("opus-4-1") || this.model.includes("sonnet-4-5") || this.model.includes("haiku-4-5")) {
```
bracesproul, line 735: maybe just check for `-4-5`? opus 4.5 will prob have the same issue

hntrl:
Thanks @bysxx! This PR is based off of the v1 version of langchain (we merged in our v1 changes yesterday!). I've migrated your changes over to v0 in #9211

We'll make some fixes to the default options so we don't have to do this weirdness in the future, so in lieu of that am going to close this PR!

PR #9212, opened 2025-10-16
fix(anthropic,openai): handle structured output with v1 output version

hntrl:


libs/langchain-core/src/language_models/chat_models.ts
```diff
@@ -508,7 +527,8 @@ export abstract class BaseChatModel<
             { ...parsedOptions, promptIndex: i },
             runManagers?.[i]
           );
-          if (this.outputVersion === "v1") {
+          console.log("outputVersion", outputVersion);
```
eric-langchain: consolelog

hntrl: lol ty. need to turn that lint rule back on

PR #9228, opened 2025-10-17
feat(core): Add `BaseMessage.toFormattedString()`

pokey:
This matches `pretty_print()` from Python, except it just returns the string. You can do `console.log(message.prettyPrint())` to get python's `pretty_print` behaviour


libs/langchain-core/src/messages/base.ts
```diff
@@ -378,6 +378,41 @@ export abstract class BaseMessage<
...
+  /**
+   * Protected method for derived classes to add type-specific details
+   * to the pretty print output. Returns an array of strings to be inserted
+   * between the title and content sections.
+   */
+  protected _prettyPrintDetails(): string[] {
+    return [];
+  }
+
+  prettyPrint(): string {
```
pokey: note that in python it directly prints instead of returning a string. Should we match that? If so I'd be inclined to have that be a simple wrapper of this func for testability

libs/langchain-core/src/messages/ai.ts
```diff
@@ -435,6 +451,22 @@ export class AIMessageChunk<
     };
   }
 
+  protected override _prettyPrintDetails(): string[] {
```
pokey: This duplication with `_prettyPrintDetails` above is awkward. Seems to be due to the inheritance / interface hierarchy of `AIMessageChunk`. @hntrl any thoughts on this one? I could obv factor into a function but was hoping to understand the situation a bit better before blindly doing that

hntrl: Yeah it is a little awkward, and I don't love the repeated code, but I think its fine to have two overrides (this is precedent with some other things in message inheritance as I understand it).

Could hoist this logic and have the methods call it directly (if that's not what you're talking about already)

libs/langchain-core/src/messages/base.ts
hntrl: gut says no. Wonder if we should call this something like `toFormattedString()` instead to match convention better

libs/langchain-core/src/messages/ai.ts
pokey: Huh where were you thinking we'd hoist it to?

libs/langchain-core/src/messages/base.ts
pokey: done

hntrl (approved):

libs/langchain-core/src/messages/tool.ts
```diff
@@ -124,6 +124,14 @@ export class ToolMessage<TStructure extends MessageStructure = MessageStructure>
...
     };
   }
+
+  protected override _toFormattedStringDetails(): string[] {
+    const lines: string[] = [];
+    if (this.name) {
+      lines.push(`Name: ${this.name}`);
+    }
+    return lines;
+  }
```
hntrl: later enhancement: tool call id, status, args, etc.

pokey: You get those from the original ai message so if you're printing messages in a loop that would end up being a bit redundant. But agreed useful when printing on its own

Also fwiw I believe this is matching what Python pretty print does

.changeset/metal-hats-camp.md
```diff
@@ -0,0 +1,5 @@
+---
+"@langchain/core": patch
```
pokey, line 2: I think according to semver this would be `minor`, because it is adding something to public api? Admittedly it is a small thing we're adding; not sure how strictly we're adhering to semver

hntrl: think you're the first of us to ask that Q post v1! Will confer and let you know

hntrl:
@pokey hope you don't mind, I took some liberties with the implementation to resolve CI issues. When I mentioned hoisting the pretty print function, this is what I had in mind (and maybe lets us introduce other message `toString` formats later down the line? don't know what that looks like but hey)

PR #9234, opened 2025-10-20
feat(langchain): tool retry middleware

pokey:
Adds `toolRetryMiddleware` to automatically retry failed tool calls with configurable exponential backoff, exception filtering, and error handling.

Port of https://github.com/langchain-ai/langchain/pull/33503

## Example

```typescript
import { createAgent, toolRetryMiddleware } from "langchain";

// Retry up to 3 times with exponential backoff
const retry = toolRetryMiddleware({
  maxRetries: 3,
  initialDelay: 1000,
  backoffFactor: 2.0,
});

const agent = createAgent({
  model: "openai:gpt-4o",
  tools: [searchTool, databaseTool],
  middleware: [retry],
});

// Tool failures are automatically retried
  messages: [{ role: "user", content: "Search for AI news" }],
});
```

For advanced usage with specific exception handling:

```typescript
class TimeoutError extends Error {}

function shouldRetry(error: Error): boolean {
  // Only retry on 5xx errors or timeouts
  if (error.constructor.name === "HTTPError") {
    const statusCode = (error as any).statusCode;
    return 500 <= statusCode && statusCode < 600;
  }
  return error.constructor === TimeoutError;
}

const retry = toolRetryMiddleware({
  maxRetries: 4,
  retryOn: shouldRetry,
  tools: ["search_database"], // Only apply to specific tools
});
```

### DIFF
--- a/libs/langchain/src/agents/middleware/tests/toolRetry.test.ts
+++ b/libs/langchain/src/agents/middleware/tests/toolRetry.test.ts
@@ -5,7 +5,7 @@
 import { describe, it, expect } from "vitest";
 import { HumanMessage, ToolMessage } from "@langchain/core/messages";
 import { tool } from "@langchain/core/tools";
-import { z } from "zod";
+import { z } from "zod/v3";
 import { MemorySaver } from "@langchain/langgraph-checkpoint";
 import { performance } from "node:perf_hooks";
 


libs/langchain/src/agents/middleware/toolRetry.ts
```diff
@@ -0,0 +1,358 @@
...
+   * Multiplier for exponential backoff. Each retry waits
+   * `initialDelay * (backoffFactor ** retryNumber)` milliseconds.
+   * Set to 0.0 for constant delay. Default is 2.0.
+   */
+  backoffFactor?: number;
+
+  /**
+   * Initial delay in milliseconds before first retry. Default is 1000 (1 second).
+   */
+  initialDelay?: number;
```
pokey: Just flagging that this is milliseconds, where python uses seconds. It is more common in js to use milliseconds, but I don't love the inconsistency with the Python version. Any opinions? We could name it `initialDelayMs` to make unit clearer if we decide to stick with milliseconds

hntrl: No super strong opinions, but would agree we should use ms.

libs/langchain/src/agents/middleware/toolRetry.ts
```diff
@@ -0,0 +1,377 @@
...
+  }
+  if (initialDelay < 0) {
+    throw new ValueError("initialDelay must be >= 0");
+  }
+  if (maxDelay < 0) {
+    throw new ValueError("maxDelay must be >= 0");
+  }
+  if (backoffFactor < 0) {
+    throw new ValueError("backoffFactor must be >= 0");
+  }
```
hntrl: If we're matching convention, maybe makes sense to defer this validation/ default logic to zod. https://github.com/langchain-ai/langchainjs/blob/bf158d29532e957c53d531a93bacf6af2f3df562/libs/langchain/src/agents/middleware/llmToolSelector.ts#L57

libs/langchain/src/agents/middleware/toolRetry.ts
```diff
@@ -0,0 +1,377 @@
...
+
+  // Extract tool names from BaseTool instances or strings
+  const toolFilter: string[] | undefined = tools?.map((tool) =>
+    typeof tool === "string" ? tool : (tool.name as string)
+  );
```
hntrl: might pay to be more strict here:
```suggestion
  const toolFilter = tools?.reduce((acc, tool) => {
    if (typeof tool === "string") return [...acc, tool];
    else if ("name" in tool && typeof tool.name === "string") return [...acc, tool];
    else return acc;
  }, [] as string[]);
```

libs/langchain/src/agents/middleware/toolRetry.ts
```diff
@@ -0,0 +1,377 @@
...
+      tool_call_id: toolCallId,
+      name: toolName,
+      status: "error",
+    });
+  };
+
+  return createMiddleware({
+    name: "toolRetryMiddleware",
+    wrapToolCall: async (request, handler) => {
+      const toolName = request.tool.name as string;
```
hntrl, line 329: very strange that these properties are unknown. No need to fix in this PR but this caught me off guard

libs/langchain/src/agents/middleware/tests/toolRetry.test.ts
```diff
@@ -0,0 +1,899 @@
...
+        middleware: [retry] as const,
+        checkpointer: new MemorySaver(),
+      });
+
+      const result = await agent.invoke(
+        { messages: [new HumanMessage("Use failing tool")] },
+        { configurable: { thread_id: "test" } }
+      );
+
+      const toolMessages = result.messages.filter(ToolMessage.isInstance);
```
hntrl, line 355: neat!

libs/langchain/src/agents/middleware/toolRetry.ts
pokey: So we'd just silent drop tools that don't look right? Would it not be better to fail fast?

libs/langchain/src/agents/middleware/toolRetry.ts
pokey: oh duh thanks 🤦

libs/langchain/src/agents/middleware/toolRetry.ts
pokey: Added `Ms` suffix to make it clearer

pokey:
Ok feedback addressed and left some responses

libs/langchain/src/agents/middleware/toolRetry.ts
hntrl: Idk if it super matters. If that's something we do want to guard against, I'd say we need to throw a better error instead of just whatever comes out of accessing tool.name on a weird looking object

pokey: Throwing a better error sounds good to me. Any objection?

PR #9235, opened 2025-10-20
feat(examples): Add supervisor example

pokey:
To accompany https://docs.langchain.com/oss/javascript/langchain/supervisor from https://github.com/langchain-ai/docs/pull/981

- Depends on https://github.com/langchain-ai/langchainjs/pull/9228

### DIFF
--- a/examples/src/createAgent/supervisor.ts
+++ b/examples/src/createAgent/supervisor.ts
@@ -36,7 +36,7 @@ import { z } from "zod";
 const createCalendarEvent = tool(
   async ({ title, startTime, endTime, attendees, location }) => {
     // Stub: In practice, this would call Google Calendar API, Outlook API, etc.
-    return `Event created: ${title} from ${startTime} to ${endTime} with ${attendees.length} attendees`;
+    return `Event created: ${title} from ${startTime} to ${endTime} with ${attendees.length} attendees at ${location}`;
   },
   {
     name: "create_calendar_event",
@@ -54,7 +54,12 @@ const createCalendarEvent = tool(
 const sendEmail = tool(
   async ({ to, subject, body, cc }) => {
     // Stub: In practice, this would call SendGrid, Gmail API, etc.
-    return `Email sent to ${to.join(", ")} - Subject: ${subject}`;
+    return [
+      `Email sent to ${to.join(", ")}`,
+      `- Subject: ${subject}`,
+      `- Body: ${body}`,
+      `- CC: ${cc.join(", ")}`,
+    ].join("\n");
   },
   {
     name: "send_email",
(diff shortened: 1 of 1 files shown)


examples/src/createAgent/supervisor.ts
```diff
@@ -0,0 +1,334 @@
...
+  {
+    name: "schedule_event",
+    description: `
+Schedule calendar events using natural language.
+
+Use this when the user wants to create, modify, or check calendar appointments.
+Handles date/time parsing, availability checking, and event creation.
+
+Input: Natural language scheduling request (e.g., 'meeting with design team next Tuesday at 2pm')
+    `.trim(),
```
hntrl, line 180: shower thought: we should have some kind of `prompt` template literal that dedents + trims

```
  description: prompt`
    Schedule calendar events using natural language.

    Use this when the user wants to create, modify, or check calendar appointments.
    Handles date/time parsing, availability checking, and event creation.

    Input: Natural language scheduling request (e.g., 'meeting with design team next Tuesday at 2pm')
  `,
```

hntrl (approved):

examples/src/createAgent/supervisor.ts
pokey: huh that's an interesting idea. I think we'd want interpolation to work like regular string interpolation, which _should_ be easy to do?

examples/src/createAgent/supervisor.ts
```diff
@@ -0,0 +1,374 @@
...
+// ============================================================================
+// Step 3: Wrap sub-agents as tools for the supervisor
+// ============================================================================
+
+const scheduleEvent = tool(
+  async ({ request }, config) => {
+    // Customize context received by sub-agent
+    // Access full thread messages from the config
+    const currentMessages =
+      getCurrentTaskInput<InternalAgentState>(config).messages;
```
pokey: As mentioned on docs PR, this is an error because it appears `InternalAgentState` is not exported from `langchain` so I didn't import it at the top

examples/src/createAgent/supervisor.ts
hntrl: yeah that should just work out of the box! This is moreso just me not wanting to do weird dedenting to properly format prompts in my code 😁

examples/src/createAgent/supervisor.ts
hntrl: Hmm maybe we should change it then

hntrl: updated

examples/src/createAgent/supervisor.ts
pokey: Right yeah agreed that would be cleaner

PR #8963, opened 2025-09-15
Feat/redis vectorstore custom schema

Dag7:
Support custom schema for redis vector store

PR #9240, opened 2025-10-20
feat(redis): Advanced pre-filter for document metadata

tishun:
First off - massive apologies to @Dag7 - we seem to both started to work on generally the same issue and we only realised that when we had to merge his changes in #8963.

### Goal

The purpose of this change is to align the implementation of langchainjs with that of langchain, specifically in regards to using Redis as a vector store. There are some major drifts in both solutions, some of which make them incompatible with each other (one app using langchainjs and one langchain might have trouble saving to the same Redis store).

### Solution

Some of these were addressed by @Dag7 already, but we wanted to provide a more complete implementation, including:
- API abstractions that make it easier to build up custom queries, possibly addressing problems such as #5010 
- do not expose driver specific (node-redis) models as part of the API of langchainjs (potentially allowing for driver change of required)
- using custom filters does not rely on providing a custom schema, instead the langchainjs driver would infer the schema from the provided metadata
- attempt to provide backwards compatibility with the old implementation, instead of providing a new one
- extend the integration tests with a lot of scenarios that were missing
- use UUIDs for generating keys, similar to langchain
- etc.

Please let us know if there is something we can do to improve this solution.

IMHO it would be very good to have it in the 1.0 release, otherwise we would have to change the contract again after releasing the solution provided by @Dag7

### DIFF
--- a/.changeset/fluent-redis-vector-store.md
+++ b/.changeset/fluent-redis-vector-store.md
@@ -1,0 +1,11 @@
+---
+"@langchain/redis": minor
+---
+
+feat(redis): add FluentRedisVectorStore with advanced pre-filtering for document metadata
+
+- Added `FluentRedisVectorStore` class with type-safe fluent filtering API
+- Added filter expression classes: `Tag`, `Num`, `Text`, `Geo`, `Timestamp`, `Custom`
+- Added `MetadataFieldSchema` array-based schema definition with automatic inference
+- Supports complex filter combinations with AND/OR logic and nesting
+- Existing `RedisVectorStore` preserved for backwards compatibility


Dag7:
Hey @tishun 👋
No worries at all 😅

My main goal with this change was to keep it fully backward compatible — for example, I intentionally avoided forcing UUIDs, since many existing users already rely on predictable or custom keys. Forcing UUIDs could break compatibility or make adoption slower for those users.

I really like your approach to making the implementation driver-agnostic — that direction makes a lot of sense. We’ll just need to make sure we add some form of adapter layer since different clients (like ioredis vs node-redis) behave differently in a few areas (pipelines, return types, etc.).

Regarding the schema, I decided to have it explicitly defined at index creation time. That way, the index and the schema are always aligned, and we can validate metadata on insert. If we infer the schema automatically from the first batch of documents, we might not capture all possible fields (since not all are required), which leads to a tricky situation:
- The index would be created based on an incomplete schema
- Later inserts could include new fields or types we will have to decide whether to re-index, ignore, or error out — all of which have trade-offs.

So defining the schema first felt like the safest and most predictable approach.

let me know what you think

tishun:
> My main goal with this change was to keep it fully backward compatible — for example, I intentionally avoided forcing UUIDs, since many existing users already rely on predictable or custom keys. Forcing UUIDs could break compatibility or make adoption slower for those users.

Agreed, backwards compatibility (and predictability) will be sacrificed if we choose to use UUIDs. We would gain - however - compatibility with the python implementation. To be fair they are not really incompatible and collisions are highly unlikely even now, but strictly speaking they will generate keys in two different ways. One could argue that this is also a benefit, allowing us to identify which driver added which vectors, so I am not at all adamant on this change. If you recommend that we revert to the old implementation I see no problem for us to do that (specifically in that regard).

> I really like your approach to making the implementation driver-agnostic — that direction makes a lot of sense. We’ll just need to make sure we add some form of adapter layer since different clients (like ioredis vs node-redis) behave differently in a few areas (pipelines, return types, etc.).

True, and this is also why I stopped short at a complete solution. The change becomes quite large; and the benefit right now is not entirely visible (Redis intends to support node-redis primarily and in terms of functionality, quality and performance it is also better). This part of my change was more "good-practise" and would allow (should this becomes necessary) to have a smaller impact if we migrate from one driver to another.

> So defining the schema first felt like the safest and most predictable approach.

Completely reasonable. Using a custom schema is definitely the more stable solution; and I assume most users would do that. Inferring the schema would be a generally lazier (but still valid - if used correctly) approach. I think they both serve different use cases:
- a simple approach where metadata is always the same for all documents and it could (safely) be inferred by the driver; and thus a much simpler usage is required
- an advanced use-case - perhaps the one most user would choose - where the metadata schema is defined by the user and the driver follows these definitions strictly

BTW I am not sure it was apparent from my description, but currently both modes are available:
- legacy filter or legacy metadata field (in the existing vector storage) results in legacy metadata handling
- missing custom schema results in a custom metadata schema being inferred from the first batch of documents
- existing custom metadata schema results in it being applied with priority

This is also - mostly - how the Python implementation of langchain works.

Does that make sense?

libs/providers/langchain-redis/package.json
```diff
@@ -27,6 +27,7 @@
     "format:check": "prettier --config .prettierrc --check \"src\""
   },
   "dependencies": {
+    "uuid": "^10.0.0",
```
vchomakov, line 31: Did you run `pnpm install` after this change? It would probably update the `pnpm-lock.yaml` that should also be added to the PR.

tishun: Good catch!

libs/providers/langchain-redis/src/vectorstores.ts
```diff
@@ -263,68 +208,39 @@ export class RedisVectorStore extends VectorStore {
...
 
     const multi = this.redisClient.multi();
 
     await Promise.all(
       vectors.map(async (vector, idx) => {
         const key =
-          keys && keys.length
-            ? keys[idx]
-            : `${this.keyPrefix}${idx + lastKeyCount}`;
+          keys && keys.length ? keys[idx] : `${this.keyPrefix}${uuidv4()}`;
```
hntrl: This jumped out at me. Without being too familiar with redis vectorstores (so with a big grain of salt), would this hash field change technically mean that new keys are being added to existing vectorstore instances?

hntrl (commented):
Thanks @tishun and apologies for the delay! Really like this API for filtering you're introducing!

One question about how metadata fields get populated -- the thing I'm trying to protect against is the underlying data model (how it's defined & accessed) changing in a minor version. There's already about 1.1k downloads of the 1.0 version of `@langchain/redis`, so I think we're past the point of bundling this in 1.0. I think we're fine to publish a 2.0 release, but I'll defer to you if (1) that's super necessary and (2) if there's anything else I'm missing that we should consider breaking in this regard. If that is the case we'll need to publish accompanying docs + a migration guide.

libs/providers/langchain-redis/src/vectorstores.ts
tishun: Hm, do you mean if there is a chance that we store the data in an existing key (in case of a conflict - a UUID that is repeated for two different documents) - given how UUIDs are generated it is very unlikely.

Additionally the existing solution also has a concurrency problem - in case two separate instances query the store to get the current ID they will attempt to store their data in the same key (next increment). This use case is much more possible.

That being said this is not something the team feels strongly for. We can revert to generating the IDs the old way (see my other comment to @Dag7)

Did this help with your question?

tishun:
> Thanks @tishun and apologies for the delay! Really like this API for filtering you're introducing!
> 
> One question about how metadata fields get populated -- the thing I'm trying to protect against is the underlying data model (how it's defined & accessed) changing in a minor version. There's already about 1.1k downloads of the 1.0 version of `@langchain/redis`, so I think we're past the point of bundling this in 1.0. I think we're fine to publish a 2.0 release, but I'll defer to you if (1) that's super necessary and (2) if there's anything else I'm missing that we should consider breaking in this regard. If that is the case we'll need to publish accompanying docs + a migration guide.

Hey @hntrl thanks for taking a look! Very valid questions, let me see if we can go over them.

Could you give me some general feeling about when 2.0 is expected? Ideally we would have wanted this to come ASAP as there is already a major use case that we want to enable.

On the API part there are three major changes:

* removing the `similaritySearchVectorWithScoreAndMetadata` -  I was not sure what the best approach is. We have - somewhat - deviated from the pattern in other providers by introducing the `similaritySearchVectorWithScoreAndMetadata` method and I was aiming to make sure we expose a similar contract. My assumption is that users of langchain want to benefit from the abstraction from the underlying implementation and be able to (relatively easily) switch between them. The more the contract of one provider deviates, the harder would it be to switch back and from it to another one.
* introducing the `FilterExpression` - non breaking, as it simply extends the contract, so we have no problem with it
* replacing the `CustomSchemaField[] ` with `MetadataFieldSchema` - could be implemented in a way similar to the `FilterExpression` if we want to preserve the existing pattern.

There is no problem for me to suggest another (temporary approach) that fully preserves the contract for 1.0 and we can leave this PR for 2.0 as a breaking change.

WDYT?

tishun:
# Redis Vector Store Improvements - Backwards compatible approach

After careful analysis of the filtering requirements and implementation challenges, we've adopted a **dual-class approach** that provides two distinct implementations optimized for different use cases:

### 1. **RedisVectorStore** (Basic/Legacy)
- **Purpose**: Existing solution is unchanged, providing full backwards compatibility
- **Filter Type**: simple tag-based, or custom TEXT/TAG/NUMERIC concatenated with AND
- **Metadata Storage**: dual (separate custom fields when schema used, single JSON when no custom schema is used)
- **Metadata Schema**: custom metadata schema supported, but no automatic inference

### 2. **FluentRedisVectorStore** (Advanced/New)
- **Purpose**: Proposed new approach, to be adopted when advanced pre-filtering is required
- **Filter Type**: AND, OR composable filters with TEXT,TAG,NUMERIC/GEO/TIMESTAMP support
- **Metadata Storage**: Individual indexed fields per metadata property only
- **Metadata Schema**: custom metadata schema supported, automatic schema inferred from data when importing documents

### Why Two Classes Instead of One?

1. **Incompatible Approaches** - The two filtering paradigms are fundamentally different:
   - Basic: Stores metadata as JSON, uses simple string matching
   - Advanced: Stores metadata as individual indexed fields, uses RediSearch expressions

2. **Clear Separation of Concerns**: Each class has a focused responsibility:
   - No complex conditional logic to handle both modes
   - Cleaner, more maintainable code
   - Better type safety and IDE support

3. **Backward Compatibility**: Existing users can continue using `RedisVectorStore` without changes and have clear steps to upgrade in the README.MD file

### Future steps

With the release of 2.x we can consider removing the `RedisVectorStore` in favor of the `FluentRedisVectorStore`

PR #9253, opened 2025-10-22
improve(pdf-loader): update error message to specify support for pdf-…

SeraphimKaito:
…parse v1 only

## description
* Update PDFLoader failure message to explicitly state that only pdf-parse v1 is supported and provide a concrete install command example (npm install pdf-parse@^1).
* Addresses user confusion when installing pdf-parse v2 by making the version requirement explicit.
* pdf-parse latest version is v2 and is incompatible with current implementation; previous message was vague and led users to install v2 inadvertently.

### DIFF
--- a/.changeset/orange-cars-mix.md
+++ b/.changeset/orange-cars-mix.md
@@ -1,0 +1,5 @@
+---
+"@langchain/community": patch
+---
+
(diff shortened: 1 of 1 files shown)


christian-bromann (approved):
LGTM 👍

PR #9259, opened 2025-10-24
feat(agents): Improve handling of `withLangGraph` annotations

pokey:
When passing `withLangGraph` as the `stateSchema` either directly to `createAgent` or to middleware, defaults are not properly detected

Note that I was not able to get the type checking to work properly here, so I'm leaving this as draft. See the tests and the type errors in CI

I would be tempted to have langgraph make types optional if a default is passed or if there is a reducer, which tries to apply a default, but maybe that wasn't done for a reason

### DIFF
--- a/libs/langchain/src/agents/nodes/tests/utils.test.ts
+++ b/libs/langchain/src/agents/nodes/tests/utils.test.ts
@@ -1,0 +1,65 @@
+import { describe, it, expect } from "vitest";
+import { z } from "zod/v3";
+import { withLangGraph } from "@langchain/langgraph/zod";
+import { initializeMiddlewareStates } from "../utils.js";
+import type { AgentMiddleware } from "../../middleware/types.js";
+
+describe("initializeMiddlewareStates", () => {
+  it("should handle withLangGraph defaults", async () => {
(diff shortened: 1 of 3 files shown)


christian-bromann (commented):
Thanks @pokey for the PR. The `createAgent` infers the state type based on all middleware schema outputs and recognizes optional or schemas with default values when you ise Zod v4. `withLangGraph` is primarily for Zod v3 users. If you define the state as following:

```ts
    const agent = createAgent({
      model: llm,
      stateSchema: z.object({
        fieldWithDefault: withLangGraph(z.string(), {
          reducer: {
            fn: (_left: string, right: string) => right,
          },
          default: () => "default-value",
        }),
      }),
    });
```

Then `fieldWithDefault` will be defined as `ReducedZodChannel<z.ZodString, InteropZodType<string>>` as still remain a required property. That said, I don't think we can make `withLangGraph` work in this context. This brings us down to 2 options: either remove the support of inferred state types and allow Zod v3 users to use `withLangGraph` or not support schemas with default value for Zod v3 users.

For now I would advocate to do latter one given that Zod v4 is widely adopted and most users that need to can upgrade. Happy to be convinced otherwise.

PR #9261, opened 2025-10-24
Add `providerToolDefinition` field to support built-in client tools

pokey:
# Why?

Today, the way we determine whether a tool is a server or client tool is based on its shape: it it has `{type: ..., name: ...}` then it is server, otherwise client. This heuristic breaks for Anthropic memory and text editor tools, which have that shape, but are client tools

# What?

This PR introduces the `providerToolDefinition` field on client tools, allowing you to define a client tool but override the way it presents itself to the provider, rather than automatically generating based on schema

### DIFF
--- a/libs/langchain-core/src/tools/index.ts
+++ b/libs/langchain-core/src/tools/index.ts
@@ -389,12 +389,15 @@ export class DynamicTool<
 
   func: DynamicToolInput<ToolOutputT>["func"];
 
+  providerToolDefinition?: Record<string, unknown>;
+
   constructor(fields: DynamicToolInput<ToolOutputT>) {
     super(fields);
     this.name = fields.name;
     this.description = fields.description;
     this.func = fields.func;
     this.returnDirect = fields.returnDirect ?? this.returnDirect;
+    this.providerToolDefinition = fields.providerToolDefinition;
   }
 
   /**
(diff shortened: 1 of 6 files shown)


pokey:
@christian-bromann yeah that all makes sense to me. I guess one question is how we'd handle the version of Anthropic tools that stores the file contents on the graph state. That seems to rely on langgraph / middleware types. Did we end up finding a way to do that without the circular deps issue?

christian-bromann:
@pokey would there still be a need for middleware if we abstract this as tool functions?

pokey:
Good question. There are 4 different middlewares in my PR, for combinations of (file system vs graph state X memory vs file tool). The one that leverages the most middleware functionality is memory tool with graph state. That uses the following middleware functionality:

- updating system prompt, though fwiw it is static, not dynamic
- adding a beta header to Anthropic request
- updating / reading graph state as that's where files are stored

If we can accomplish all of these things with just tools and no middleware, then I'm all for it!

See https://github.com/pokey/langchainjs/blob/pokey/featlangchainagents-add-anthropic-tools-middleware/libs/langchain/src/agents/middleware/anthropicTools/index.ts for the relevant code

christian-bromann:
@pokey thanks for raising the PR. I will go ahead and close this. We've decided to put provider specific logic into the `extras` field of the tool. We have the `invocationParams` automatically extend the request headers if a specific tool is used. Furthermore we've decided to hold off the middleware for JS for now.

PR #9288, opened 2025-10-29
chore(*): build system improvements

### LINKED ISSUE #9252: import { PromptTemplate } from "@langchain/core/prompts"; TS2307: Cannot find module @langchain/core/prompts or its corresponding type declarations. IDE is webstorm
### Checked other resources

### Example Code

import { PromptTemplate } from "@langchain/core/prompts";
TS2307: Cannot find module @langchain/core/prompts or its corresponding type declarations. IDE is webstorm

_No response_

### Description

@langchain/core

### System Info

yarn @langchain/core

### LINKED ISSUE #9268: Most public API types are not exported from @langchain/core entrypoints (e.g. ToolRunnableConfig, ToolWrapperParams, etc.)
### Checked other resources

### Example Code

// This is not possible:
import type { ToolRunnableConfig } from "@langchain/core/tools";

_No response_

### Description

Many types that are part of the public API surface of @langchain/core cannot be imported from public entrypoints. For example, types like `ToolRunnableConfig` and `ToolWrapperParams` are referenced in public APIs (e.g., tool parameters, config fields, etc.) but are not exported from any subpath allowed by the package `exports` field. This is a general issue across the package and not limited to tools.

#### Examples (not exhaustive)
```typescript
// Not possible via public API:
import type { ToolRunnableConfig } from "@langchain/core/tools"; // Not exported
import type { ToolWrapperParams } from "@langchain/core/tools"; // Not exported

// Only works with deep imports, which are NOT allowed by the exports field:
import type { ToolRunnableConfig } from "@langchain/core/dist/tools/types";
```

- ToolWrapperParams is only declared locally and is not exported at all.
- ToolRunnableConfig is referenced by public types like ToolParams, but not re-exported.
- This pattern exists for other config, wrapper, or utility types throughout @langchain/core as well.

### System Info

- "pnpm info langchain"
- platform: mac
- Node version: 20+
- pnpm: 8+
- @langchain/core: 1.0.1

hntrl:
* use tsdown cli/configuration scheme instead of internal options
* make tsdown plugins individually addressable
* adds a barrel-dts plugin to fix some compatibility issues with commonjs dependants
* changes build scripts in each package

Fixes #9252
Fixes #9268


libs/langchain-textsplitters/package.json
```diff
@@ -41,6 +42,7 @@
     "eslint": "^9.34.0",
     "prettier": "^2.8.3",
     "rollup": "^4.5.2",
+    "tsdown": "^0.15.12",
```
christian-bromann: Can we define this on the root level to avoid cases where we use different `tsdown` versions?

libs/providers/langchain-google-webauth/tsdown.config.ts
```diff
@@ -0,0 +1,4 @@
+import { defineConfig } from "tsdown";
+import { getBuildConfig } from "@langchain/build";
+
+export default defineConfig([getBuildConfig()]);
```
christian-bromann: Are brackets required?

```suggestion
export default defineConfig(getBuildConfig());
```

christian-bromann (commented):
As we are revisiting this: the idea of defining entry points via `src` property in `exports` of `package.json` I think made sense when we defined all build configs within `@langchain/build`. Now that we go back to workspace config files we could also define entry points there, e.g. as `getBuildConfig` parameter?

I don't see anything wrong with the current approach, except it is not standard conferment and vitest gets confused by this. What are your thoughts here?

libs/langchain-textsplitters/package.json
christian-bromann: Overall IMHO we should try to avoid defining dependencies for tooling within workspaces:

- makes it difficult to maintain
- version mismatch can cause issues

libs/providers/langchain-google-webauth/tsdown.config.ts
hntrl: no they are not, good catch!

hntrl:
> As we are revisiting this: the idea of defining entry points via src property in exports of package.json I think made sense when we defined all build configs within

Could you elaborate more on this? Do you mean we should be changing how the entry config option is populated? I saw that tsdown has an `exports` key that may be of interest:
https://tsdown.dev/reference/api/Interface.UserConfig#exports

christian-bromann:
Currently package entries are defines within `exports` of `package.json` via `input` key, e.g.:

```ts
"./callbacks/base": {
  "input": "./src/callbacks/base.ts",
  "require": {
    "types": "./dist/callbacks/base.d.cts",
    "default": "./dist/callbacks/base.cjs"
  },
  "import": {
    "types": "./dist/callbacks/base.d.ts",
    "default": "./dist/callbacks/base.js"
  }
},
```

This is not standard conferment and other tools may get thrown off by this. Using this exports function would be great, can we give this a shot?

hntrl:
closing in favor of #9416

PR #9290, opened 2025-10-30
chore: rm scripts

hntrl:
removing the langchain scripts package since we're not using it anymore


libs/langchain-classic/output_parsers/expression.d.ts
```diff
@@ -0,0 +1 @@
+export * from "../dist/output_parsers/expression";
```
christian-bromann: These should not be checked in right?

hntrl: ha no they should not. thanks for catching!

christian-bromann:
@hntrl pushed https://github.com/langchain-ai/langchainjs/pull/9293 removing two added files that I think are not needed.

PR #9292, opened 2025-10-30
fix(langchain/middleware): allow model strings in summarization middleware

hntrl:


libs/langchain/src/agents/middleware/tests/summarization.test.ts
```diff
@@ -335,4 +365,27 @@ describe("summarizationMiddleware", () => {
     expect(nonSystemMessages.length).toBeGreaterThanOrEqual(messagesToKeep);
     expect(nonSystemMessages.length).toBeLessThanOrEqual(messagesToKeep + 3); // Some buffer for safety
   });
+
+  it("can be created using a model string", async () => {
+    // Verify the mocked ChatAnthropic exists
+    const { ChatAnthropic } = await import("@langchain/anthropic");
```
christian-bromann: No need to import here, we can just import it at the top of the file.

libs/langchain/src/agents/middleware/tests/summarization.test.ts
```diff
@@ -335,4 +365,27 @@ describe("summarizationMiddleware", () => {
...
+      messagesToKeep: 2,
+    });
+
+    const agent = createAgent({
+      model,
+      middleware: [middleware],
+    });
+
+    const result = await agent.invoke({ messages: [] });
+    expect(result.messages.length).toBeGreaterThan(0);
```
christian-bromann: Can we add an assertion that the `ChatAnthropic` model was actually used?

PR #9306, opened 2025-11-03
fix(langchain): fix state in wrapToolCall and wrapModelCall

christian-bromann:
`wrapToolCall` and `wrapModelCall` do not properly retain the state of the middleware. This patch fixes these issues and adds test coverage to ensure we don't regress.

### DIFF
--- a/libs/langchain/src/agents/middleware/tests/modelCallLimit.test.ts
+++ b/libs/langchain/src/agents/middleware/tests/modelCallLimit.test.ts
@@ -112,7 +112,7 @@ describe("ModelCallLimitMiddleware", () => {
     }
   );
 
-  describe.each(["throw" /*, "end"*/] as const)(
+  describe.each(["throw", "end"] as const)(
     "thread limit with exit behavior %s",
     (exitBehavior) => {
       const checkpointer = new MemorySaver();
@@ -126,7 +126,7 @@ describe("ModelCallLimitMiddleware", () => {
         exitBehavior,
       });
 
-      it.only("should not throw if the thread limit is not exceeded", async () => {
+      it("should not throw if the thread limit is not exceeded", async () => {
         const model = new FakeToolCallingChatModel({
           responses: [toolCallMessage1],
         });
@@ -151,7 +151,6 @@ describe("ModelCallLimitMiddleware", () => {
             { messages: ["Hello, world!"] },
             config
           );
-          console.log("result", result);
           await expect(result.runModelCallCount).toBe(3);
           await expect(result.threadModelCallCount).toBe(3);
         } else {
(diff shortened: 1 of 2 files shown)


libs/langchain/src/agents/middleware/tests/modelCallLimit.test.ts
```diff
@@ -112,7 +112,7 @@ describe("ModelCallLimitMiddleware", () => {
     }
   );
 
-  describe.each(["throw", "end"] as const)(
+  describe.each(["throw" /*, "end"*/] as const)(
```
dqbd: Uncomment the test back

libs/langchain/src/agents/middleware/tests/modelCallLimit.test.ts
```diff
@@ -126,7 +126,7 @@ describe("ModelCallLimitMiddleware", () => {
         exitBehavior,
       });
 
-      it("should not throw if the thread limit is not exceeded", async () => {
+      it.only("should not throw if the thread limit is not exceeded", async () => {
```
dqbd: Remove `only`

libs/langchain/src/agents/middleware/tests/modelCallLimit.test.ts
```diff
@@ -151,6 +151,7 @@ describe("ModelCallLimitMiddleware", () => {
             { messages: ["Hello, world!"] },
             config
           );
+          console.log("result", result);
```
dqbd: Remove `console.log`

libs/langchain/src/agents/utils.ts
```diff
@@ -497,7 +501,10 @@ function chainToolCallHandlers(
  * @param middleware list of middleware passed to the agent
```
dqbd, line 501: _nit_: do we need to update the JSDoc as well?

libs/langchain/src/agents/ReactAgent.ts
```diff
@@ -129,6 +126,8 @@ export class ReactAgent<
 
   #agentNode: AgentNode<any, AnyAnnotationRoot>;
 
+  #stateManager = new StateManager();
```
dqbd, line 129: Does the `#stateManager` properly persist in checkpointer?

libs/langchain/src/agents/utils.ts
christian-bromann: good catch

libs/langchain/src/agents/ReactAgent.ts
christian-bromann: I don't think it needs to. It is just s small helper class to help merge the state between 4 nodes into a single agent state, e.g. if you change state in `beforeModel` it makes sure that `afterModel` gets these updates.

PR #9316, opened 2025-11-04
feat(@langchain/core): support of ToolRuntime

christian-bromann:
After trying to support state and context aware tools through `contextSchema` and `stateSchema` as tool arguments, here is a 2nd approach:

This patch attempts to add more overwrites to the `tool` function to allow users to define the second tool function argument to be a `ToolRuntime`. Note that specific ToolRuntime parameters are only applied (in ToolNode) when using the tool via `createAgent`.

### DIFF
--- a/.changeset/large-onions-attack.md
+++ b/.changeset/large-onions-attack.md
@@ -1,0 +1,6 @@
+---
+"@langchain/core": patch
+"langchain": patch
+---
+
+feat(@langchain/core): support of ToolRuntime


libs/langchain-core/src/tools/types.ts
```diff
@@ -425,3 +428,102 @@ export function isLangChainTool(tool?: unknown): tool is StructuredToolParams {
...
+ * );
+ * ```
+ *
+ * @template StateT - The type of the state schema (inferred from stateSchema)
+ * @template ContextT - The type of the context schema (inferred from contextSchema)
+ */
+export type ToolRuntime<
+  TState = unknown,
+  TContext = unknown
+> = RunnableConfig & {
```
hntrl, line 496: think we want this to extend `ToolRunnableConfig`? (since additional properties like `toolCall` get added within the tool invocation)

libs/langchain-core/src/tools/index.ts
```diff
@@ -635,6 +637,98 @@ export function tool<
 >(
   func: RunnableFunc<SchemaOutputT, ToolOutputT, ToolRunnableConfig>,
   fields: ToolWrapperParams<SchemaT>
+):
+  | DynamicStructuredTool<SchemaT, SchemaOutputT, SchemaInputT, ToolOutputT>
+  | DynamicTool<ToolOutputT>;
+
+// Overloads with ToolRuntime as CallOptions
+export function tool<
```
hntrl, line 645: what's the purpose of these overloads if `TState` + `TContext` is only used in the runtime arg? Would say something different if we were recommending the usage of this should be by passing in TState + TContext as generic args, but I don't think that's what we want either. E.g. this should be possible w/o overloads?

```ts
const myTool = tool((state, runtime: ToolRuntime<...>) => {}, {})
```

christian-bromann: This are unfortunately required because the runtime type is otherwise defined as:

```ts
  options:
    | CallOptions
    | Record<string, any>
    | (Record<string, any> & CallOptions)
) => RunOutput | Promise<RunOutput>;
```

which fails because of:

```
Type 'Record<string, any>' is missing the following properties from type '{ state: { userId: string; }; toolCallId: string; config: ToolRunnableConfig<Record<string, any>, any>; context: { db: { foo: string; }; }; store: BaseStore<string, unknown> | null; writer: ((chunk: unknown) => void) | null; }': state, toolCallId, config, context, and 2 more.ts(2769)
```

libs/langchain-core/src/tools/types.ts
christian-bromann: I couldn't quite use `ToolRunnableConfig` as it makes `context` an optional property which it isn't it every case, so I extended `ToolRuntime` with `toolCall` to match the same type.

Issue #9305, opened 2025-11-01
Milvus vector store: calculated on db side field is expected to be provided from client


ayanyev:
### Checked other resources

- [x] This is a bug, not a usage question. For questions, please use the LangChain Forum (https://forum.langchain.com/).
- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain.js documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain.js rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

no code

### Error Message and Stack Trace (if applicable)

{
  "errorMessage": "The field \"title_vector\" is not provided in documents[0].metadata.",
  "errorDetails": {},
  "n8nDetails": {
    "n8nVersion": "1.116.2 (Self Hosted)",
    "binaryDataMode": "default",
    "stackTrace": [
      "Error: The field \"title_vector\" is not provided in documents[0].metadata.",
      "    at /usr/local/lib/node_modules/n8n/node_modules/.pnpm/@langchain+community@0.3.50_a15f2f07b13189caf0f9b792748d9f32/node_modules/@langchain/community/dist/vectorstores/milvus.cjs:277:35",
      "    at Array.forEach (<anonymous>)",
      "    at Milvus.addVectors (/usr/local/lib/node_modules/n8n/node_modules/.pnpm/@langchain+community@0.3.50_a15f2f07b13189caf0f9b792748d9f32/node_modules/@langchain/community/dist/vectorstores/milvus.cjs:256:25)",
      "    at processTicksAndRejections (node:internal/process/task_queues:105:5)",
      "    at Milvus.addDocuments (/usr/local/lib/node_modules/n8n/node_modules/.pnpm/@langchain+community@0.3.50_a15f2f07b13189caf0f9b792748d9f32/node_modules/@langchain/community/dist/vectorstores/milvus.cjs:229:9)",
      "    at Function.fromDocuments (/usr/local/lib/node_modules/n8n/node_modules/.pnpm/@langchain+community@0.3.50_a15f2f07b13189caf0f9b792748d9f32/node_modules/@langchain/community/dist/vectorstores/milvus.cjs:567:9)",
[...]

PR #9326, opened 2025-11-05
fix(community/milvus): Milvus vector store client: ignore auto-calculated fields in collection schema during upsert payload validation

ayanyev:
Fixes [9305](https://github.com/langchain-ai/langchainjs/issues/9305)

**Current state:**
Milvus vector store client expects server-side auto-calculated field value to be explicitly provided in Document's metadata. If field is not present in metadata, error is thrown by the client. If dummy data is provided through metadata to satisfy client, error is thrown by server, which does not expect such field in payload.

**Target state:**
Milvus vector store client does not expect that such field will be sent to server as part of Document's metadata on any insert or update request.

PR removes such fields from the list of Milvus collection schema fields.

### DIFF
--- a/.changeset/tall-icons-warn.md
+++ b/.changeset/tall-icons-warn.md
@@ -1,0 +1,5 @@
+---
+"@langchain/community": patch
+---
+
+Milvus vector store client: ignore auto-calculated fields in collection schema during payload validation


ayanyev:
@christian-bromann could you please pay attention to this PR?

hntrl (approved):

ayanyev:
Thank you all!

PR #9327, opened 2025-11-05
feat: add web export to `langchain`

hntrl:
* adds a `langchain/web` export for compatibility in web/edge environments


christian-bromann:
Should we just remove the `setContextVariable`/`getContextVariable` export? We haven't documented it anywhere yet and we can later see how we can provide something that is more cross platform compatible.

christian-bromann:
Closing in favor of #9330

PR #9346, opened 2025-11-07
Fix structured outputparser toolcall

BeastxD7:
##  Fix: StructuredOutputParser now handles multiline JSON strings & OpenAI-style tool calls

This PR fixes a parsing issue where `StructuredOutputParser` failed when LLMs returned **literal newlines inside JSON string fields** — a common pattern in OpenAI-style function/tool calling.

This behavior appears in real-world usage, including models shown in the LangChain docs (e.g., the Ollama example using `withStructuredOutput`):  
https://github.com/langchain-ai/docs/blob/main/src/oss/javascript/integrations/chat/ollama.mdx#withstructuredoutput

###  What’s Improved
- Added `escapeNewlinesInsideStrings` to safely convert raw newlines to `\n` **only inside quoted JSON strings**, preventing `JSON.parse` errors.
- Updated `parse()` to clean and sanitize JSON before parsing.
- Fully supports `{ "arguments": { ... } }` tool-call style output with multiline values.
- Added new test: `structured-newlines.test.ts` validating multiline parsing.

###  Status
- ✅ All core tests pass  
- ✅ Lint clean  
- ✅ dpdm: no circular dependencies  

This fix significantly improves `StructuredOutputParser` reliability when handling real LLM output, especially function-calling 
models.

### Screenshot of Issue:
image

Social Handle: [X](https://x.com/shashankx02) ,[ Linkedin](https://www.linkedin.com/in/shashank-devadiga2/)

PR #9350, opened 2025-11-07
fix(ollama): support parsing structured output from tool calls

christian-bromann:
I originally reviewed this PR #9346. It turns out that our [docs example](https://github.com/langchain-ai/docs/blob/main/src/oss/javascript/integrations/chat/ollama.mdx#withstructuredoutput) throws and error. Even after applying the patch by the user it doesn't seem to work. It seems like the Ollama chat model does not support parsing structured output from tool calls. With this patch my example script started to work:

```ts
import { ChatOllama } from "@langchain/ollama";
import * as z from "zod";

// Define the model
const llmForWSO = new ChatOllama({
  model: "gpt-oss:120b",
  baseUrl: "https://ollama.com",
  headers: {
    Authorization: `Bearer ${process.env.OLLAMA_API_KEY}`,
  },
});

// Define the tool schema you'd like the model to use.
const schemaForWSO = z.object({
  location: z.string().describe("The city and state, e.g. San Francisco, CA"),
});

// Pass the schema to the withStructuredOutput method to bind it to the model.
const llmWithStructuredOutput = llmForWSO.withStructuredOutput(schemaForWSO, {
  name: "get_current_weather",
});

const resultFromWSO = await llmWithStructuredOutput.invoke(
  "What's the weather like today in San Francisco? Ensure you use the 'get_current_weather' tool."
);
console.log(resultFromWSO);
```

That said, given someone put this example up in the first place I wonder why it is failing now. I recognise that I am using a different model so maybe that could be it.

### DIFF
--- a/.changeset/unlucky-keys-poke.md
+++ b/.changeset/unlucky-keys-poke.md
@@ -1,0 +1,5 @@
+---
+"@langchain/ollama": patch
+---
+
+fix(ollama): support parsing structured output from tool calls


hntrl (approved):
Interesting. Yeah not sure if this ever worked. We may also look to support ollama native structured outputs at some point: #9353

PR #9351, opened 2025-11-07
feat(google-vertexai): support Gemini propertyOrdering in withStructu…

thomasrowe5:
…redOutput

Fixes # (issue)

Adds automatic property ordering inference for Gemini structured outputs when using Zod schemas.

### DIFF
--- a/libs/providers/langchain-google-common/src/chat_models.ts
+++ b/libs/providers/langchain-google-common/src/chat_models.ts
@@ -65,6 +65,19 @@ import {
   schemaToGeminiParameters,
 } from "./utils/zod_to_gemini_parameters.js";
 
+/**
+ * Helper to automatically infer Gemini property ordering from a Zod schema.
+ */
+function withPropertyOrdering<RunOutput extends Record<string, any>>(
+  schema: InteropZodType<RunOutput>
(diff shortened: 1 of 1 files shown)


christian-bromann:
@thomasrowe5 it seems that there are still a lot of un-relevant changes, can we revert them to only focus on the actual issue?

libs/providers/langchain-google-common/src/chat_models.ts
```diff
@@ -483,8 +428,15 @@ export abstract class ChatGoogleBase<AuthOptions>
...
     let tools: GeminiTool[];
+
     if (isInteropZodSchema(schema)) {
       const jsonSchema = schemaToGeminiParameters(schema);
+
+      // 🧩 Add inferred property ordering for Gemini structured outputs
+      if (!jsonSchema.propertyOrdering) {
+        jsonSchema.propertyOrdering = withPropertyOrdering(schema);
+      }
+
```
hntrl, line 439: If I'm reading this right, you have to specify `propertyOrdering` as a property on the schema object to enable this?

```ts
z.object({
  propertyOrdering: z.something()
});
```

PR #9386, opened 2025-11-13
fix(langchain): Prevent local file corruption when using `LocalFileStore` 

### LINKED ISSUE #9337: Concurrent `LocalFileStore` `mset` writes can corrupt chunked embeddings cache

Josh-Engle:
# Description

This PR addresses a race condition in `LocalFileStore` where concurrent writes to the same key could lead to data loss or corruption. Previously, multiple `mset` operations on the same key would execute in parallel, causing file writes to overwrite each other. Additionally, added guards against incomplete file writes due to unexpected process termination.

This is addressed through a number of mechanisms.
- Dedupe keys within the same `mset` call to ensure the last one is written.
- Use a promise queue to guarantee sequential execution across multiple `mset` calls.
- Write to tmp files and then rename to text files prevent partial writes in the case of unexpected termination.

Fixes #9337


libs/langchain-classic/src/storage/file_system.ts
```diff
@@ -194,6 +221,82 @@ export class LocalFileStore extends BaseStore<string, Uint8Array> {
...
+          await fs.unlink(tempPath).catch(() => {});
+        } else {
+          throw renameError;
+        }
+      }
+    } catch (error) {
+      await fs.unlink(tempPath).catch(() => {});
+      throw error;
+    }
+  }
```
hntrl, line 301: What's the purpose of the temp files if we're sequencing files using `withKeyLock`?

hntrl (commented):
Thanks @Josh-Engle! One Q about your approach:

libs/langchain-classic/src/storage/file_system.ts
Josh-Engle: This is an added guard to prevent partial file writes causing data corruption if the process terminates before the buffer is fully flushed for a particular file. 

We write to the temp file first as this might be interrupted. If it is interrupted, only the temp file is corrupted which is cleaned up. Once we have fully flushed the fs buffer and we know it's in a good state, we overwrite the previous file with the `rename` which cannot be interrupted as it is either successful or not.

hntrl: Is that a common occurrence? What is the rename operation fails?

Josh-Engle: It's hard to say if it's common but I can image relatively likely scenarios and the problem tends to scale up with larger file sizes. Larger files mean larger write times, increasing the chance of a failure mid-write. 

A good example might be:

1. A developer force terminates the process execution while `LocalFileStore` is halfway through writing a file to the file system.
2. The incomplete write operation results in malformed JSON as it is not terminated correctly.
3. Reading the `txt` file results in an invalid JSON exception.

The `rename` operation is considerably more reliable then the `writeFile` because it's a single operation. If we have already written data to the filesystem with `writeFile` it's very likely the `rename` will work correctly as it requires the same OS filesystem permissions. 

For windows I think it would only fail if there was a open file handle and on Linux it would fail if it doesn't have access to the directory. In both cases I think a failure is expected.

Josh-Engle: @hntrl Would you prefer I remove the atomic file writes to keep the PR focused?

hntrl: Yeah I think that's preferred. Unless we have a concrete case/ MRE that triggers this that we think is reasonable, I want to air on the side of not over optimizing too early

PR #9396, opened 2025-11-14
fix(langchain): rename exit behavior from throw to error

christian-bromann:
Minor interface adjustment to match with Python.


christian-bromann:
> are we considering this breaking?

Pushed a change to make it backwards compatible.

PR #9403, opened 2025-11-14
fix(langchain): improvements to toolEmulator middleware

christian-bromann:
This PR updates the `toolEmulatorMiddleware` to use the agent's model as the default emulator model instead of a hardcoded default. This change makes the middleware more flexible and reduces unnecessary model initialization when the agent model can be reused for emulation.

## Key Changes

**Before:**

- Middleware defaulted to `"anthropic:claude-sonnet-4-5-20250929"` when no model was provided
- Required explicit model configuration to use the agent model

**After:**

- Middleware defaults to using the agent model (captured via `wrapModelCall`)
- Only initializes a separate emulator model when explicitly provided
- Falls back to agent model if explicit model initialization fails

### DIFF
--- a/.changeset/tall-parrots-allow.md
+++ b/.changeset/tall-parrots-allow.md
@@ -1,0 +1,5 @@
+---
+"langchain": patch
+---
+
+align context editing with summarization interface
--- a/libs/langchain/src/agents/middleware/contextEditing.ts
+++ b/libs/langchain/src/agents/middleware/contextEditing.ts
@@ -8,7 +8,7 @@
  */
 
 import type { BaseMessage } from "@langchain/core/messages";
-import type { LanguageModelLike } from "@langchain/core/language_models/base";
+import type { BaseLanguageModel } from "@langchain/core/language_models/base";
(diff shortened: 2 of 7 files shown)


libs/langchain/src/agents/middleware/toolEmulator.ts
```diff
@@ -102,19 +101,32 @@ export function toolEmulatorMiddleware(
...
+    }
+    return agentModel as BaseChatModel;
   };
 
   return createMiddleware({
     name: "ToolEmulatorMiddleware",
+    wrapModelCall: async (request, handler) => {
+      agentModel = request.model as BaseChatModel;
+      return handler(request);
+    },
```
hntrl, line 129: feel like we should have some more native way to get agent model inside of middleware (and make that getter the default for all middleware that relies on models), but I'll live for now

christian-bromann: Yep, have another use case for this in the OpenAI content moderation middleware.

PR #9408, opened 2025-11-14
fix(openai): normalize user agent headers

sinedied:
Looking into our stats, we discovered that the user agent string were not always the same when received from OpenAI/Azure clients.

This PR fixes that and simplifies the headers management by factorizing the code into a single function.

### DIFF
--- a/.changeset/goofy-poets-search.md
+++ b/.changeset/goofy-poets-search.md
@@ -1,0 +1,5 @@
+---
+"@langchain/openai": patch
+---
+
+Fix missing and inconsistent user agent headers


libs/providers/langchain-openai/src/utils/azure.ts
```diff
@@ -158,3 +159,29 @@ export function normalizeHeaders(
...
+export function getFormattedEnv() {
+  let env = getEnv();
+  if (env === "node" || env === "deno") {
+    env = `(${env}/${process.version}; ${process.platform}; ${process.arch})`;
+  }
+  return env;
+}
+
+// Note: ideally version would be imported from package.json, but there's
+// currently no good way to do that for all supported environments (Node, Deno, Browser).
```
hntrl, line 172: ack

PR #9428, opened 2025-11-18
feat(langchain): use model profiles to for structured output in createAgent

christian-bromann:
Improve the `createAgent` to:

- use model profiles when to determine the default behavior for structured output strategy
- use the native `withStructuredOutput` function to handle getting structured ouput

I am currently blocked here as I can't bind tool when using `withStructuredOutput`.


hntrl:
Closing this as we need to wait until we can make a breaking change -- we want to depend on `withStructuredOutput` natively, but that requires us removing the use of RunnableBinding

PR #9431, opened 2025-11-18
fix(core): make sure `store` is provided to tools

dqbd:


hntrl:
How come this is needed? Would think that the only thing this function modifies is things pertinent to runnable invocation. All config keys should be accessible to tool functions (evidenced by how we call tools + how we pass it through in createAgent)

dqbd:
> How come this is needed? Would think that the only thing this function modifies is things pertinent to runnable invocation. All config keys should be accessible to tool functions (evidenced by how we call tools + how we pass it through in createAgent)

LangGraph.js does inject the `store` as top-level key of `RunnableConfig` / `Runtime`, omitting this will cause the tool to not see the `store`

PR #9436, opened 2025-11-18
fix(community): possible race condition on FileSystemChatMessageHistory

sinedied:
Fixes a small possible race condition on FileSystemChatMessageHistory, and removes unneeded extra check


christian-bromann:
> Fixes a small possible race condition on FileSystemChatMessageHistory, and removes unneeded extra check

I can see the `removes unneeded extra check` part but not the fix of a race condition 🤔

christian-bromann (changes requested):
Seems like the PR is not complete?

sinedied:
@christian-bromann It's surprising but complete :) I started initially by refactoring the code to use file descriptors, but it turned out that the first `fs.access()` check that causes the race condition wasn't even needed in the first place, so the fix was even simpler 😅

sinedied:
Bumping this, @christian-bromann / @hntrl could you review this one? Just confirming that removing this single line is the fix 🙂

PR #9440, opened 2025-11-18
fix: update polynomial regex to mitigate ReDoS vulnerability in SseStream

manekinekko:
This PR mitigates ReDoS vulnerability in SseStream in @langchain/google-common.

See:
[CWE-1333](https://cwe.mitre.org/data/definitions/1333.html)
[CWE-400](https://cwe.mitre.org/data/definitions/400.html)
[CWE-730](https://cwe.mitre.org/data/definitions/730.html)

### DIFF
--- a/libs/providers/langchain-google-common/src/utils/stream.ts
+++ b/libs/providers/langchain-google-common/src/utils/stream.ts
@@ -29,6 +29,40 @@ export interface AbstractStream {
   get streamDone(): boolean;
 }
 
+/**
+ * Parse an SSE event line into key-value pair.
+ * Format: "field: value" or "field:value" (with optional whitespace after colon)
+ *
+ * Uses string operations instead of regex to avoid ReDoS vulnerabilities.
+ * This is safer and more performant than regex-based parsing.
+ */
+function parseEventLine(line: string): { key: string; value: string } | null {
+  const colonIndex = line.indexOf(":");
+  if (colonIndex === -1 || colonIndex === 0) {
+    return null;
+  }
+
+  const key = line.substring(0, colonIndex).trim();
+  if (key.length === 0) {
+    return null;
+  }
+
+  // Skip colon and any whitespace after it
+  let valueStart = colonIndex + 1;
+  while (
+    valueStart < line.length &&
+    (line[valueStart] === " " ||
+      line[valueStart] === "\t" ||
+      line[valueStart] === "\r" ||
+      line[valueStart] === "\n")
+  ) {
+    valueStart++;
+  }
+
+  const value = line.substring(valueStart);
+  return { key, value };
+}
+
 export function complexValue(value: unknown): unknown {
   if (value === null || typeof value === "undefined") {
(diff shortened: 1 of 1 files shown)


libs/providers/langchain-google-common/src/utils/stream.ts
```diff
@@ -376,9 +376,35 @@ export class SseStream implements AbstractStream {
     }
     const ret: Record<string, string> = {};
 
+    // The old regex that was vulnerable to ReDoS: /^([^:]+): \s*(.+)\n*$/
+
+    // The new regex:
```
christian-bromann:
```suggestion
```

christian-bromann (changes requested):
Actually, looking at this a 2nd time: The `(.+)\n*$` part can cause backtracking: 
- `.+` greedily matches one or more characters
- `\n*` matches zero or more newlines
- `$` asserts end of string.
 
Since the code already splits by `\n` (line 379), each line won't contain newlines. If a line doesn't end with a newline, the engine may backtrack through all characters matched by `.+` trying to satisfy `\n*$`, leading to exponential time.

libs/providers/langchain-google-common/src/utils/stream.ts
```diff
@@ -376,9 +376,32 @@ export class SseStream implements AbstractStream {
     }
     const ret: Record<string, string> = {};
 
+    const regex = /^([^:]+):\s*([^\r\n]*?)(?:\r?\n)*$/;
```
christian-bromann:
```suggestion
    const regex = /^([^:]+):\s*(.+)$/;
```

libs/providers/langchain-google-common/src/utils/stream.ts
```diff
@@ -376,9 +376,32 @@ export class SseStream implements AbstractStream {
...
+      )
+      *$
+    /
+
+    This regex is not a ReDoS vulnerability because:
+    - It removes the ambiguity between what .+ and \n* can match
+    - It uses character classes that are more specific and don't overlap
+    - It uses non-greedy quantifiers where appropriate
+    - It makes the regex more deterministic in its matching behavior
+    */
```
christian-bromann:
```suggestion
      /*
       * Pattern breakdown:
       *   ^([^:]+)    - Matches and captures the field name (one or more non-colon characters from start)
       *   :\s*        - Matches colon followed by zero or more whitespace characters
       *   (.+)$       - Matches and captures the field value (one or more characters until end of string)
       *
       * This regex is safe from ReDoS (Regular Expression Denial of Service) because:
       * - Since lines are pre-split by /\n/, each line contains no newline characters
       * - There is no ambiguity or overlapping patterns (like .+ and \n*) that could cause exponential backtracking
       * - The greedy .+ quantifier is deterministic when anchored to $ (end of string)
       * - No nested quantifiers or optional groups that could create backtracking complexity
       */
```

manekinekko:
Thank you for reviewing this PR @christian-bromann. Your changes LGTM.

PR #9463, opened 2025-11-21
fix(core): update p-retry to fix memory leak

christian-bromann:
Users reported memory spikes during tool executions with retries that **never get released**, even after operations completed. Investigation revealed that multiple LangChain packages were using `p-retry@4.6.2`, which has a known memory leak where event listeners accumulate over time and are never cleaned up. With each retry attempt, more listeners are added but never removed, causing memory to continuously grow.

## Solution

Upgraded `p-retry` from version 4 to version 7.x across all LangChain packages. Version 7 fixes the memory leak with:

- ✅ Proper event listener cleanup (uses `{once: true}` and explicit removal)
- ✅ AbortSignal support for cancellation (v4 had none)
- ✅ Time-based retry limits via `maxRetryTime`
- ✅ Better network error detection

### Dependencies Updated

- `@langchain/core`: `p-retry@4` → `p-retry@^7.0.0`
- `@langchain/classic`: `p-retry@4` → `p-retry@^7.0.0`

### Code Changes

Updated `onFailedAttempt` callback usage to match p-retry v7 API:

**Before (v4):**

```typescript
onFailedAttempt: (error: any) => this.onFailedAttempt(error);
```

**After (v7):**

```typescript
onFailedAttempt: ({ error }: { error: any }) => this.onFailedAttempt(error);
```

## Breaking Changes Handled

p-retry v7 introduces breaking changes that were addressed:

1. **ESM-only**: ✅ Already compatible - LangChain packages use ESM
2. **Node.js 18+ requirement**: ✅ Already satisfied - packages require Node.js 20+
3. **`forever` option removed**: ✅ Not used in codebase
4. **`onFailedAttempt` callback signature**: ✅ Updated to use context object destructuring

### DIFF
--- a/environment_tests/test-exports-cjs/src/index.ts
+++ b/environment_tests/test-exports-cjs/src/index.ts
@@ -53,7 +53,7 @@ async function test(useAzure: boolean = false) {
   assert(typeof RunnableLambda === "function");
   let attemptCount = 0;
   const flakyRunnable = new RunnableLambda({
-    func: (_input) => {
+    func: () => {
       attemptCount += 1;
       if (attemptCount < 3) {
         throw new Error(`Attempt ${attemptCount} failed`);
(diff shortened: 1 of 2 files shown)


christian-bromann:
Double checked in LangGraph.js and found no issues.

PR #9464, opened 2025-11-21
test(internal): add environment tests for Node.js classic environments

christian-bromann:
After our build updates, this patch will introduce environment tests to ensure our packages can be run in Node.js classic environments.


hntrl (approved):
nice!

PR #9466, opened 2025-11-21
feat(langchain): Support conditional construction of agent middleware with AgentMiddlewareBuilder

carefreelife98:
## Summary
Agent creation often requires toggling multiple middleware components (todo list, human-in-the-loop, prompt caching, etc.) based on runtime conditions. Today every caller must import each factory and manually push them into arrays, which leads to duplicated logic and error-prone configuration handling.

## Proposal
1. Expose a centralized factory enum (`BuiltInAgentMiddleware`) plus `createBuiltInAgentMiddleware` so callers can request middleware by name.
2. Provide a lightweight `AgentMiddlewareBuilder` that accepts descriptors `{ name, config, when }` and returns the assembled array, mirroring the builder used in Mate-X.
3. Update docs/examples to illustrate building agent middleware stacks with conditional guards.
4. Add unit tests ensuring the builder respects `when` flags and that factory entries require the correct config shape.

## Motivation
- Improves ergonomics for teams composing agents with many optional behaviors.
- Avoids repeated imports and misconfiguration (e.g. missing retry configs).
- Gives us a single extension point when introducing new built-in middleware.

### DIFF
--- a/libs/langchain/src/agents/middleware/factory.ts
+++ b/libs/langchain/src/agents/middleware/factory.ts
@@ -1,0 +1,241 @@
+import { anthropicPromptCachingMiddleware } from "./provider/anthropic/promptCaching.js";
+import type { PromptCachingMiddlewareConfig } from "./provider/anthropic/promptCaching.js";
+import {
+  contextEditingMiddleware,
+  type ContextEditingMiddlewareConfig,
+} from "./contextEditing.js";
+import {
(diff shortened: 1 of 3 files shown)


hntrl (commented):
Could you clarify what the motivation for this is? Are you looking for a way to move around "sets" of middleware? Or supply builtin middleware based on string configurations?

PR #9473, opened 2025-11-21
feat(community): add GoogleCalendarDeleteTool  

gbrmaia:
This PR adds a new tool, GoogleCalendarDeleteTool, to the @langchain/community package, enabling users to delete events from their Google Calendar. This completes the basic CRUD operations (Create, Read, Delete) for the Google Calendar integration.

Features:

Delete by ID: Users can delete an event by providing its specific Event ID.

Natural Language Deletion: Users can request deletion using natural language queries (e.g., "delete the meeting with John at 3pm").

Smart Search: If an Event ID is not provided, the tool attempts to find the event based on the summary and time range. If a single event is found, it is deleted. If multiple events are found, the tool asks for clarification. If no events are found, it informs the user.

### DIFF
--- a/.changeset/polite-steaks-jump.md
+++ b/.changeset/polite-steaks-jump.md
@@ -1,0 +1,5 @@
+---
+"@langchain/community": patch
+---
+
+add GoogleCalendarDeleteTool  


christian-bromann (approved):
👍 

Thanks for the contribution!

PR #9484, opened 2025-11-24
feat(core): preserve thoughtSignature for Gemini 3+ tool calls

ZihaoZhou:
## Problem

  Gemini 3 models (e.g., `gemini-3-pro-preview`) return a `thoughtSignature` field with tool calls that must be included in subsequent `ToolMessage` requests for multi-turn conversations. Currently, this field is lost during:

  1. `tool_call_chunks` → `tool_calls` conversion in `AIMessageChunk` constructor
  2. Chunk concatenation via the `concat()` method

  This causes multi-turn tool calling to fail with Gemini 3.

## Solution

  Preserve `thoughtSignature` through both code paths in `libs/langchain-core/src/messages/ai.ts`.

## Testing

  - All existing message tests pass (166 tests across 6 files)
  - Verified with `gemini-3-pro-preview` model:
    - `thoughtSignature` preserved from `tool_call_chunks` to `tool_calls`
    - `thoughtSignature` preserved during chunk concatenation
    - Multi-turn tool calling works correctly with the signature

### DIFF
--- a/libs/langchain-core/src/messages/ai.ts
+++ b/libs/langchain-core/src/messages/ai.ts
@@ -331,12 +331,18 @@ export class AIMessageChunk<
           ) {
             throw new Error("Malformed tool call chunk args.");
           }
-          toolCalls.push({
+          const toolCall: ToolCall = {
             name,
             args: parsedArgs,
             id,
             type: "tool_call",
-          });
+          };
+          // Preserve thoughtSignature from chunks for Gemini 3+
+          if ((chunks[0] as ToolCallChunk & { thoughtSignature?: string })?.thoughtSignature) {
+            (toolCall as ToolCall & { thoughtSignature?: string }).thoughtSignature =
+              (chunks[0] as ToolCallChunk & { thoughtSignature?: string }).thoughtSignature;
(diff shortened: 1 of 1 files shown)


ZihaoZhou:
Closing in favor of #9441

  After further investigation, I discovered that #9441 already merged support for thoughtSignature in @langchain/google-genai using a different approach.

  #9441's approach:
  - Stores thoughtSignature in additional_kwargs as a map: {
  [_FUNCTION_CALL_THOUGHT_SIGNATURES_MAP_KEY]: { toolCallId: signature } }
  - Bypasses the need to modify @langchain/core

  This PR's approach:
  - Stores thoughtSignature directly on tool_calls objects
  - Requires changes to AIMessageChunk in @langchain/core

  Since the official solution uses additional_kwargs as a sideband transport mechanism, this PR is redundant. The @langchain/core changes here would only be needed if thoughtSignature was stored on tool_calls directly, which is not the pattern established by #9441.

  Closing to avoid confusion with the existing implementation.

nesbtesh:
Was this released?

PR #9510, opened 2025-11-27
fix(google-genai): add null safety for candidateContent.parts

shubham-021:
## Description:

Fixed a bug in @langchain/google-genai where using middleware with agents (createDeepAgent or createAgent) would randomly crash with:
`TypeError: undefined is not an object (evaluating 'candidateContent.parts')`

**Root Cause:**
in common.ts , functions mapGenerateContentResultToChatResult and convertResponseContentToChatGenerationChunk accessed candidateContent.parts without checking if candidateContent was defined.

**Changes Made:**
Added null safety for candidateContent.parts:
`const functionCalls = (candidateContent?.parts || []).reduce((acc, p) => {`

Also added a test to check if mapGenerateContentResultToChatResult handles candidate with undefined content
Only added test for one function since both had the same issue, doesn't make sense to test for same thing twice.

Fixes langchain-ai/deepagentsjs#70

### DIFF
--- a/.changeset/metal-rats-feel.md
+++ b/.changeset/metal-rats-feel.md
@@ -1,0 +1,5 @@
+---
+"@langchain/google-genai": patch
+---
+
+Fix crash when candidateContent.parts is undefined in middleware scenarios
--- a/libs/providers/langchain-google-genai/src/tests/chat_models.test.ts
+++ b/libs/providers/langchain-google-genai/src/tests/chat_models.test.ts
@@ -1,5 +1,9 @@
 import { test } from "@jest/globals";
-import type { HarmBlockThreshold, HarmCategory } from "@google/generative-ai";
+import type {
+  HarmBlockThreshold,
+  HarmCategory,
+  EnhancedGenerateContentResponse,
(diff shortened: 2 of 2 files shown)


libs/providers/langchain-google-genai/src/utils/common.ts
```diff
@@ -527,7 +527,7 @@ export function mapGenerateContentResultToChatResult(
   }
   const [candidate] = response.candidates;
   const { content: candidateContent, ...generationInfo } = candidate;
-  const functionCalls = candidateContent.parts.reduce((acc, p) => {
+  const functionCalls = (candidateContent?.parts || []).reduce((acc, p) => {
```
mariacretu-arch, line 530: Hello! This approach also works! Maybe we can check only once if `candidateContent` exists before calling reduce and log a suggestive message as warning if `candidateContent` is null.

mariacretu-arch (commented):
Just a suggestion from my side, otherwise LGTM!

hntrl:
Closing in favor of #9584

PR #9511, opened 2025-11-27
fix(core): allow parsing more partial JSON

dqbd:
Previous partial JSON parser did not handle all cases of partial JSON, causing incomplete tool calls to be erroneously marked as invalid.

### DIFF
--- a/libs/langchain-core/src/utils/tests/json.test.ts
+++ b/libs/langchain-core/src/utils/tests/json.test.ts
@@ -54,11 +54,12 @@ it("strings", () => {
   expectPartialJson(String.raw`"15\u00f8C`).toBe("15\u00f8C");
   expectPartialJson(String.raw`"15\u00f8C"`).toBe("15\u00f8C");
 
+  expectPartialJson(String.raw`"hello\\`).toBe("hello\\");
+  expectPartialJson(String.raw`"hello\\"`).toBe("hello\\");
+
   expectPartialJson(String.raw`"hello${"\\"}`).toBe("hello\\");
   expectPartialJson(String.raw`"hello\"`).toBe('hello"');
   expectPartialJson(String.raw`"hello\""`).toBe('hello"');
-
-  expectPartialJson(String.raw`"hello\\`).toBe("hello\\");
 
   expectPartialJson(String.raw`"\t\n\r\b\f\/`).toBe("\t\n\r\b\f/");
   expectPartialJson(String.raw`"\t\n\r\b\f\/"`).toBe("\t\n\r\b\f/");


libs/langchain-core/src/utils/json.ts
```diff
@@ -26,75 +26,297 @@ export function parseJsonMarkdown(s: string, parser = parsePartialJson) {
...
-  let escaped = false;
-
-  // Process each character in the string one at a time.
-  for (let char of s) {
-    if (isInsideString) {
-      if (char === '"' && !escaped) {
-        isInsideString = false;
-      } else if (char === "\n" && !escaped) {
-        char = "\\n"; // Replace the newline character with the escape sequence.
+  function parseString(): string {
```
dqbd, line 53: Might be good to try doing `JSON.parse` here?

libs/langchain-core/src/utils/tests/json.test.ts
```diff
@@ -0,0 +1,280 @@
...
+
+  expectPartialJson("[-").toEqual([-0]);
+
+  expectPartialJson("[").toEqual([]);
+  expectPartialJson("[n").toEqual([null]);
+  expectPartialJson("[null").toEqual([null]);
+  expectPartialJson("[null,").toEqual([null]);
+  expectPartialJson("[null,t").toEqual([null, true]);
+  expectPartialJson("[null,{").toEqual([null, {}]);
+  expectPartialJson('[null,{"a":1').toEqual([null, { a: 1 }]);
```
christian-bromann, line 37: not sure if these values are correct but worth covering it?
```suggestion
  expectPartialJson('[null,{"').toEqual([null, {}]);
  expectPartialJson('[null,{"a').toEqual([null, { a: undefined }]);
  expectPartialJson('[null,{"a"').toEqual([null, { a: undefined }]);
  expectPartialJson('[null,{"a":').toEqual([null, { a: undefined }]);
  expectPartialJson('[null,{"a":1').toEqual([null, { a: 1 }]);
```

christian-bromann (approved):
One suggestion on test coverage, otherwise LGTM 👍

libs/langchain-core/src/utils/json.ts
christian-bromann: Yeah, good call, trying JSON.parse would definitely have some positive performance impacts on larger complete objects.

hntrl: I think fine considering we're doing JSON.parse at the beginning of the method, and this is probably doing the same thing as JSON.parse anyways.

dqbd: Did experiment with string-healing `JSON.parse` but in the end we would need to also heal incomplete Unicode escape characters to make sure JSON.parse works correctly. Ended up keeping the current implementation.

PR #9518, opened 2025-11-30
feat(langchain-google-common): Add jsonSchema method support to withStructuredOutput in langchain-google-common and VertexAI

yukukotani:
Currently, `withStructuredOutput` in langchain-google-genai (Gemini API) supports `jsonSchema` method, which is using [native structured output](https://ai.google.dev/gemini-api/docs/structured-output).

While that, langchain-google-common (which is used by langchain-google-vertexai) supports only `functionCalling` method.

This PR adds `jsonSchema` method support to langchain-google-common. The implementation is mostly modeled after langchain-google-genai.

### DIFF
--- a/.changeset/hungry-bugs-lick.md
+++ b/.changeset/hungry-bugs-lick.md
@@ -1,0 +1,5 @@
+---
+"@langchain/google-common": minor
+---
+
+Add jsonSchema method support to withStructuredOutput in langchain-google-common and VertexAI


libs/providers/langchain-google-common/src/tests/chat_models.test.ts
```diff
@@ -1721,6 +1725,120 @@ describe("Mock ChatGoogle - Gemini", () => {
...
+
+    const schema = z.object({
+      testName: z.string().describe("The name of the test."),
+    });
+
+    const baseModel = new ChatGoogle({
+      authOptions,
+    });
+    // Not specifying method - should default to jsonSchema
+    const model = baseModel.withStructuredOutput(schema);
```
yukukotani, line 1785: Using `jsonSchema` method by default. This is the same behaviour as langchain-google-genai, but might be a breaking change for langchain-google-vertexai users. 

I would like to have an opinion of maintainers.

yukukotani: Of course we can keep `functionCalling` method as a default.

PR #9523, opened 2025-11-30
fix: Correctly identify ConfigurableModel for JSON schema support check

### LINKED ISSUE #9522: `isConfigurableModel` check fails for `ConfigurableModel` instances, breaking structured output detection for `createAgent`
### Checked other resources

### Example Code

```typescript
import { createAgent } from "langchain/agents";
import { z } from "zod";

const schema = z.object({
  answer: z.string(),
});

// When using a model string like "openai:gpt-5", createAgent uses initChatModel
// which returns a ConfigurableModel.
const agent = createAgent({
  model: "openai:gpt-5", 
  responseFormat: schema,
});

// Expected: agent uses ProviderStrategy (native structured output)
// Actual: agent uses ToolStrategy (tool calling) because isConfigurableModel fails
```

NA

### Description

When using `createAgent` with a model string (e.g. `"openai:gpt-5"`) and a Zod schema for `responseFormat`, the agent defaults to using tool calling (`ToolStrategy`) instead of native structured output (`ProviderStrategy`), even for models that support it.

### Root Cause
1. `createAgent` uses `initChatModel` internally, which returns a `ConfigurableModel` instance.
2. `AgentNode` calls `hasSupportForJsonSchemaOutput` to determine the strategy.
3. `hasSupportForJsonSchemaOutput` attempts to unwrap the configurable model using `isConfigurableModel`.
4. `isConfigurableModel` checks for the existence of a `_model` property/method.
5. However, `ConfigurableModel` (in `libs/langchain/src/chat_models/universal.ts`) implements `_getModelInstance`, not `_model`.

### Impact
`isConfigurableModel` returns `false`, causing `hasSupportForJsonSchemaOutput` to fail to inspect the underlying model configuration. This forces the agent to fall back to the default `ToolStrategy` (extract tool calls) instead of using the `ProviderStrategy` for supported models like `gpt-4o` or `gpt-5`.

### Proposed Fix
Update `isConfigurableModel` in `libs/langchain/src/agents/model.ts` to check for `_getModelInstance` instead of `_model`.

```typescript
export function isConfigurableModel(
  model: unknown
): model is ConfigurableModelInterface {
  return (
    typeof model === "object" &&
    model != null &&
    "_queuedMethodOperations" in model &&
    "_getModelInstance" in model && // Changed from _model
    typeof (model as { _getModelInstance: unknown })._getModelInstance ===
      "function"
  );
}
```

### System Info

Node: v25.2.1
pnpm: 10.14.0
Platform: Linux

LiteracyFanatic:
Previously, `isConfigurableModel` checked for a `_model` property, but `ConfigurableModel` instances returned by `initChatModel` use `_getModelInstance`. This caused `hasSupportForJsonSchemaOutput` to fail to unwrap the underlying model, leading to `createAgent` defaulting to `ToolStrategy` instead of `ProviderStrategy` even for supported models like GPT-4o/GPT-5 when initialized via string.

This change updates `isConfigurableModel` to check for `_getModelInstance` and adds tests.

Fixes #9522

### DIFF
--- a/libs/langchain/src/agents/utils.ts
+++ b/libs/langchain/src/agents/utils.ts
@@ -388,7 +388,11 @@ export async function bindTools(
   if (model) return model;
 
   if (isConfigurableModel(llm)) {
-    const model = _simpleBindTools(await llm._model(), toolClasses, options);
+    const model = _simpleBindTools(
+      await llm._getModelInstance(),
+      toolClasses,
+      options
+    );
     if (model) return model;
   }
 
--- a/.changeset/loud-hornets-battle.md
+++ b/.changeset/loud-hornets-battle.md
@@ -1,0 +1,5 @@
+---
+"langchain": patch
+---
+
+Fix detection of models with native support for structured output
--- a/libs/langchain/src/agents/model.ts
+++ b/libs/langchain/src/agents/model.ts
@@ -3,7 +3,7 @@ import type { BaseChatModel } from "@langchain/core/language_models/chat_models"
 
 export interface ConfigurableModelInterface {
   _queuedMethodOperations: Record<string, unknown>;
-  _model: () => Promise<BaseChatModel>;
+  _getModelInstance: () => Promise<BaseChatModel>;
 }
 
 export function isBaseChatModel(
@@ -23,7 +23,8 @@ export function isConfigurableModel(
     typeof model === "object" &&
     model != null &&
     "_queuedMethodOperations" in model &&
-    "_model" in model &&
-    typeof model._model === "function"
+    "_getModelInstance" in model &&
+    typeof (model as { _getModelInstance: unknown })._getModelInstance ===
+      "function"
   );
 }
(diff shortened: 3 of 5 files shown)


christian-bromann:
@LiteracyFanatic thanks for the contribution, can we have a integration test e.g. in `libs/langchain/src/agents/tests/responses.int.test.ts` to verify this behavior from end to end?

LiteracyFanatic:
@christian-bromann I added an integration test that asserts no `extract-*` tool calls are present.

christian-bromann:
@LiteracyFanatic mind taking a look at the failing tests?

libs/langchain/src/agents/utils.ts
```diff
@@ -388,7 +388,7 @@ export async function bindTools(
   if (model) return model;
 
   if (isConfigurableModel(llm)) {
-    const model = _simpleBindTools(await llm._model(), toolClasses, options);
+    const model = _simpleBindTools(await llm._getModelInstance(), toolClasses, options);
```
christian-bromann:
```suggestion
    const model = _simpleBindTools(
      await llm._getModelInstance(),
      toolClasses,
      options
    );
```

PR #9526, opened 2025-12-01
fix(google-genai): remove unsupported `propertyNames` and `patternProperties` from JSON schema for Gemini tool-calling

Nitinref:
## Fix: Gemini JSON Schema Error (propertyNames removed)

This PR fixes the issue where Gemini throws:

INVALID_ARGUMENT: Unexpected property: propertyNames

### What was the bug?
Zod → Gemini schema conversion was passing fields that Gemini does not support:
- propertyNames
- additionalProperties
- strict object metadata

Gemini rejects these and breaks tool calling + structured output.

### What we added
A new utility `cleanGeminiSchema.ts` that:
- Recursively removes unsupported JSON schema fields
- Ensures Gemini gets a clean schema
- Prevents the INVALID_ARGUMENT error

### Files changed
- chat_models.ts
- utils/common.ts
- cleanGeminiSchema.ts (new file)

### Manual Test (passed)
Running a tool with a Zod schema containing dynamic keys now works without errors.

### Summary
This PR ensures that all Gemini models accept LangChain-generated schemas safely and correctly.


hntrl (commented):
Could we add some kind of test suite to verify this works?

Nitinref:
@hntrl Added a focused unit test for the google-genai package
to verify that unsupported `propertyNames` and
`patternProperties` are removed from schemas.
The google-genai test suite passes locally.

IBJunior:
Hey @Nitinref  I ran into a very similar issue, but with a different property: exclusiveMinimum. This happens when a tool definition includes a min value constraint. For example:

```js
// Using zod
        start_index: z
          .number()
          .int()
          .min(0)
```
I noticed that this error specifically occurs when using tools defined with http://json-schema.org/draft-07/schema#

Because of that, it might be better to handle these fields in `cleanGeminiSchema`, rather than limiting it to just the two fields currently covered.

As a temporary workaround, I applied a local fix (using UNSUPPORTED_SCHEMA_KEYWORDS ) on my side, since the tools I’m using come from third-party MCP servers.

```js
const UNSUPPORTED_SCHEMA_KEYWORDS = [
  "$schema",
  "$id",
  "$ref",
  "$defs",
  "definitions",
  "exclusiveMinimum",
  "exclusiveMaximum",
  "multipleOf",
  "minLength",
  "maxLength",
  "pattern",
  "minItems",
  "maxItems",
  "uniqueItems",
  "minProperties",
  "maxProperties",
  "additionalProperties",
  "patternProperties",
  "allOf",
  "anyOf",
  "oneOf",
  "not",
  "if",
  "then",
  "else",
  "contentMediaType",
  "contentEncoding",
  "examples",
  "default",
  "const",
  "readOnly",
  "writeOnly",
  "deprecated",
  "title",
];
```

For the long term, it might also make sense to address this on the Google side, as this issue seems to surface primarily when using ChatGoogleGenerativeAI.

pedroferrarezzo:
Hi guys. Any update on this? I'm still facing this error with @langchain/google-genai@2.1.30 using gemini-2.5-flash

`[GoogleGenerativeAI Error]: Error fetching from https://generativelanguage.googleapis.com/v1beta/models/gemini-2.5-flash:streamGenerateContent?alt=sse: [400 Bad Request] Invalid JSON payload received. Unknown name "propertyNames" at 'tools[0].function_declarations[1].parameters.properties[1].value': Cannot find field. [{"@type":"type.googleapis.com/google.rpc.BadRequest","fieldViolations":[{"field":"tools[0].function_declarations[1].parameters.properties[1].value","description":"Invalid JSON payload received. Unknown name \"propertyNames\" at 'tools[0].function_declarations[1].parameters.properties[1].value': Cannot find field."}]}]`

percymcn:
Ran this against the shape that actually produces these keywords, and the fix as written doesn't reach it yet. Three things, all reproducible with no API key.

**1. The post-clean output still 400s, on the exact shape that emits `propertyNames`.**

Google validates the request payload *before* auth, so a dummy key still returns a real schema verdict. Measured today against `v1beta` on the function-declaration surface:

```
raw z.record()      -> Unknown name "propertyNames" at 'tools[0].function_declarations[0].parameters...'
after cleanGeminiSchema -> Unknown name "additionalProperties" at 'tools[0].function_declarations[0].parameters...'
control: {"type":"object","frobnicate":1} -> Unknown name "frobnicate"   (oracle is live and discriminating)
control: plain object                     -> "API key not valid"          (i.e. it got past payload validation)
```

`z.record(z.string(), z.string())` emits **both** `propertyNames` and `additionalProperties`. Deleting the first leaves the second, and the narrow `parameters` / `responseSchema` proto has no `additionalProperties` field either — so the user gets the same error class one keyword later. Repro:

```bash
curl -s -X POST "https://generativelanguage.googleapis.com/v1beta/models/gemini-2.0-flash:generateContent?key=DUMMY" \
  -H 'Content-Type: application/json' -d '{"contents":[{"parts":[{"text":"hi"}]}],"tools":[{"functionDeclarations":[{"name":"f","description":"d","parameters":{"type":"object","properties":{"m":{"type":"object","additionalProperties":{"type":"string"}}},"required":["m"]}}]}]}'
```

**2. The trigger is any `z.record()`, not an exotic key type.** On `zod@4.4.3`:

```
z.record(z.string(), z.string())          -> {"type":"object","propertyNames":{"type":"string"},"additionalProperties":{"type":"string"}}
z.record(z.string().regex(/^S_/), z.string()) -> propertyNames with the pattern
z.record(z.enum(["a","b"]), z.string())   -> propertyNames + "required":["a","b"]   <- note: required names keys `properties` never declares
```

Zod 4 emits `propertyNames` for *every* record, including the plain one. That's probably why this keeps getting reported against new versions (the two comments above are on 2.1.30). Note the third line separately: after the cleaner runs, that schema carries `required: ["a","b"]` with no `properties` at all.

**3. Both added tests pass whether or not the fix works on the reported case.** `removes propertyNames from schema` and `removes patternProperties from schema` are hand-written objects that carry `properties` and **no** `additionalProperties` — the opposite of what a record emits. They assert the two keys are gone, which is true, while the schema that motivated the PR still fails. A regression test built on `z.toJSONSchema(z.record(z.string(), z.string()))` would fail today and pass after a complete fix.

**On the shape of the fix:** a two-key denylist is open-ended here — `exclusiveMinimum` (the comment above), `$schema`, `const`, `uniqueItems`, `patternProperties`, `if`, `multipleOf` and others are all rejected by the same proto, so each one arrives as a new issue. It's worth deciding whether the map case is *strippable at all*: the narrow proto has no way to express "arbitrary keys of type V", so silently deleting `additionalProperties` turns a typed dictionary into an unconstrained object — the request then succeeds and the model is told nothing about the values. Surfacing that (or routing to a field that takes JSON Schema, where the client supports it) is more honest than a silent strip, and it's the difference between a 400 the user can act on and a 200 that quietly stopped enforcing something.

PR #9530, opened 2025-12-01
test(mongodb): add a 100-second hook timeout for mongodb tests

PavelSafronov:
No user facing changes here, just increasing the hook timeout of MongoDB test setup to accommodate our remote Atlas testing.


PavelSafronov:
> Thanks for the PR! Could we make this change inside of https://github.com/PavelSafronov/langchainjs/blob/mongodb-testing/libs/providers/langchain-mongodb/vitest.config.ts instead?

Thanks for pointing that out. Looks like all the other code is relying on the config files, so I removed all the custom logic from our package.json and moved it into the vitest config file.

PR #9534, opened 2025-12-02
fix(@langchain/core): update and bundle dependencies

christian-bromann:
A user reported that we still face an issue with `p-queue` not being loaded in CJS environment. This patch resolves this issue once and for all by marking some of these dependencies as `noExternal` which makes rolldown bundle them into `dist`. I've added a bunch more of them to the list and took the liberty to update their dependencies as well.

### DIFF
--- a/.changeset/eight-mangos-visit.md
+++ b/.changeset/eight-mangos-visit.md
@@ -2,4 +2,4 @@
 "@langchain/core": patch
 ---
 
-fix(@langchain/core): update and bundle `p-retry`, `p-queue`, `ansi-styles`, `camelcase` and `decamelize` dependencies
+fix(@langchain/core): update and bundle `p-retry`, `ansi-styles`, `camelcase` and `decamelize` dependencies
--- a/libs/langchain-core/eslint.config.ts
+++ b/libs/langchain-core/eslint.config.ts
@@ -6,13 +6,7 @@ import { langchainConfig, type ConfigArray } from "@langchain/eslint";
  * They only need to be devDependencies, not runtime dependencies.
  * Listed here as core-modules so import/no-extraneous-dependencies ignores them.
  */
-const bundledPackages = [
-  "p-retry",
-  "p-queue",
-  "ansi-styles",
-  "camelcase",
-  "decamelize",
-];
+const bundledPackages = ["p-retry", "ansi-styles", "camelcase", "decamelize"];
 
 const config: ConfigArray = [
   ...langchainConfig,
--- a/libs/langchain-core/package.json
+++ b/libs/langchain-core/package.json
@@ -49,7 +49,6 @@
     "dotenv": "^17.2.1",
     "dpdm": "^3.14.0",
     "eslint": "^9.34.0",
-    "eventemitter3": "^5.0.1",
     "ml-matrix": "^6.10.4",
     "p-retry": "^7.0.0",
     "prettier": "^2.8.3",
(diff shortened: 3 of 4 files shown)


libs/langchain-core/src/callbacks/tests/callbacks.test.ts
```diff
@@ -242,6 +242,10 @@ test("CallbackManager Chat Message Handling", async () => {
       await llmCb.handleLLMEnd({ generations: [] });
     })
   );
+
+  // In case background mode is on while running this test
+  await awaitAllCallbacks();
```
christian-bromann, line 247: This has been a race condition that was unnoticed before. `p-queue` v6 had synchronous task chaining:
```ts
_next() {
    this._pendingCount--;
    this._tryToStartAnother();  // Synchronously starts next task
    this.emit('next');
}

// In _tryToStartAnother():
job();  // Calls the job synchronously
```
while v9 uses microtask-deferred task chaining:
```ts
// In the add() method's finally block:
queueMicrotask(() => {
    this.#next();  // Deferred to next microtask!
});
```

internal/build/src/plugins/fix-eventemitter3.ts
```diff
@@ -0,0 +1,48 @@
...
+
+          if (fixed !== content) {
+            await fs.writeFile(eventemitter3Path, fixed, "utf-8");
+          }
+        } catch (error) {
+          // File might not exist if eventemitter3 wasn't bundled
+          // This is fine, we'll just skip the fix
+          if ((error as NodeJS.ErrnoException).code !== "ENOENT") {
+            throw error;
+          }
```
hntrl: with this, are we saying that this fix is only sometimes necessary?

internal/build/src/plugins/fix-eventemitter3.ts
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * `eventemitter3` is bundled with `p-queue`, which is bundled into `@langchain/core`.
+ * Unfortunately, rolldown doesn't  bundle `eventemitter3` correctly, so we need to fix it manually.
+ */
+export function fixEventEmitter3Plugin(): Plugin {
+  return {
+    name: "fix-eventemitter3",
+    buildEnd: {
+      async handler() {
+        console.log("HAH!H");
```
hntrl: HAH!H

internal/build/src/plugins/fix-eventemitter3.ts
hntrl, line 1: This feels brittle to me. If we were including `eventemitter3` in noExternal does this issue go away?

yukukotani:
@christian-bromann Thanks for tackling this! Did you intentionally exclude p-queue from `noExternal`? I still have the error below

```
Error [ERR_REQUIRE_ESM]: require() of ES Module /package/node_modules/@langchain/core/node_modules/p-queue/dist/index.js from /package/node_modules/@langchain/core/dist/singletons/callbacks.cjs not supported.
Instead change the require of index.js in /package/node_modules/@langchain/core/dist/singletons/callbacks.cjs to a dynamic import() which is available in all CommonJS modules.
```

descampsk:
The issue is still there after the 1.1.3 but a bit different : 
```
Must use import to load ES Module: /.../node_modules/.pnpm/p-queue@9.0.1/node_modules/p-queue/dist/index.js
```

christian-bromann:
Thanks folks for flagging and sorry for the inconvenience. We will publish a fix momentarily.

Issue #9528, opened 2025-12-01
createAgent with responseFormat not working with tools with optional parameters


MartijnLeplae:
### Checked other resources

- [x] This is a bug, not a usage question. For questions, please use the LangChain Forum (https://forum.langchain.com/).
- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain.js documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain.js rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```typescript
import * as z from "zod";
import { createAgent, tool, providerStrategy, toolStrategy } from "langchain";
import "dotenv/config";

const getWeather = tool(
  ({ city }) => `It's always sunny in ${city}!`,
  {
    name: "get_weather",
    description: "Get the weather for a given city",
    schema: z.object({
      city: z.string(),
      testParameter: z.string().optional(),
    }),
  }
);

const WeatherInfo = z.object({
  current_weather: z.string().describe("Short summary of the weather."),
});

async function main() {
  const agent = createAgent({
    model: "gpt-4o-mini",
    tools: [getWeather],
    responseFormat: providerStrategy(WeatherInfo),
  });

  const result = await agent.invoke({
    messages: [{ role: "user", content: "What's the weather in Tokyo?" }],
  });

  console.log(JSON.stringify(result, null, 2));
}

main().catch((err) => {
  console.error("Error running agent:", err);
});
```

### Error Message and Stack Trace (if applicable)

```
Error running agent: BadRequestError: 400 Invalid schema for function 'get_weather': In context=(), 'required' is required to be supplied and to be an array including every key in properties. Missing 'testParameter'.
   ) {
  status: 400,
  error: {
    message: "Invalid schema for function 'get_weather': In context=(), 'required' is required to be supplied and to be an array including every key in properties. Missing 'testParameter'.",
    type: 'invalid_request_error',
    param: 'tools[0].function.parameters',
    code: 'invalid_function_parameters'
  },
  code: 'invalid_function_parameters',
  param: 'tools[0].function.parameters',
  type: 'invalid_request_error',
}
```

### Description

I created an agent with tools and wanted to add `responseFormat`. However, I found that is does not work when the tool has optional parameters. 
The code above works when I either remove `responseFormat` or `testParameter` from the tool.

[...]

PR #9535, opened 2025-12-02
feat(agent): make responseFormat.strict configurable to support optional tool parameters

Nitinref:
## 🚀 What this PR does

This patch makes the `strict` behavior of `responseFormat` configurable when using `createAgent`.  
- Previously, `createAgent` always set `strict: true` internally — this caused agents using tools or optional parameter schemas to break with `invalid_function_parameters`.  
- Now, if the user passes `responseFormat.strict = false`, the agent honors it and does not enforce strict validation — enabling use of optional fields or more flexible tool parameter schemas.

## ✅ Why this matters

- Enables tool schemas with optional parameters under structured output or JSON schema response formats.  
- Resolves the issue https://github.com/langchain-ai/langchainjs/issues/9528 — previously there was no way to bypass strict validation in agent usage.  
- Maintains backward-compatibility: if `strict` is omitted, the default remains `true`.  

## 🔧 What changed

- In `AgentNode.ts`, `strict: true` is replaced with logic that reads `responseFormat.strict` if supplied, else defaults to `true`.  
- This means only when explicitly requested, strict mode is relaxed.  

## 🧪 Verification

Tested locally with real OpenAI model.  
Example usage that works with optional tool parameters:  
```ts
const agent = createAgent({
  model: new ChatOpenAI({ model: "gpt-4o-mini" }),
  tools: [myToolWithOptionalParams],
  responseFormat: {
    // your output schema...
    strict: false,
  },
});
```
Under this configuration, agent interacts successfully; no `invalid_function_parameters` error — the fix works as intended.

## ⚠️ Notes / Caveats

- `strict: false` should be used consciously — disabling strict validation may reduce safety guarantees (e.g. tool parameter validation).  
- Users who rely on strict schema validation should either omit `strict` or explicitly set it to `true`.  

---


Nitinref:
Hi! Please let me know if any changes are needed.  
Happy to update the PR. 🙂

MartijnLeplae:
Hi, thanks for your PR! I will make some time later to go through it fully. Some initial comments:
- Why did you remove the `getState` method? It seems unrelevant to this PR.
- It seems you included changes of another pull request you made. I would keep each one clearly scoped.
- I would personally not change the type of `responseFormat` here. Non-strict tool-calling should be possible while having strict outputFormat.
Perhaps changing the logic in `responses.ts` would be enough?
I was thinking changing the following lines to change: replace the `else if` with `if` such that the strictness of the toolcalling can be controlled with the `supportsStrictToolCalling` parameter. But perhaps some input from the initial author of that piece of code is needed so no unwanted side-effects are introduced.

https://github.com/langchain-ai/langchainjs/blob/34242933ade61304481d055605af06ec54c8f5e4/libs/providers/langchain-openai/src/chat_models/responses.ts#L73-L78

Nitinref:
Thanks a lot for the detailed review! Really appreciate the guidance.

1. Regarding the `getState` removal — you're right, that change was not intentional for this PR. I will revert it.

2. It seems I accidentally committed a part of another PR I was experimenting with — thanks for pointing that out. I will remove unrelated changes and keep this PR fully scoped to the `strict` behavior only.

3. Your suggestion makes sense: avoiding changes to the type of `responseFormat` and instead updating the logic inside `responses.ts` to allow non-strict tool-calling while keeping strict outputFormat seems cleaner.

I'll update the PR accordingly and push a revised version soon.

Nitinref:
Hi! I’ve pushed the updated changes, keeping the PR focused and applying the strict logic update in the responses API as suggested.

MartijnLeplae:
Small note in general: make sure to follow contributing guidelines. (linting, formatting, testing, ...).
Also, but that's personal for me, I do not really enjoy reading llm generated pr descriptions etc. Just write it yourself. (well that is if you are not a bot yourself, which does seem like it ;) )

Nitinref:
Thank you sir for the feedback i will follow the contributing guidelines from now.

libs/providers/langchain-openai/src/chat_models/responses.ts
```diff
@@ -90,32 +91,32 @@ export class ChatOpenAIResponses<
       include: options?.include,
       tools: options?.tools?.length
         ? this._reduceChatOpenAITools(options.tools, {
-            stream: this.streaming,
```
MartijnLeplae, line 93: Follow the steps here https://github.com/langchain-ai/langchainjs/blob/main/CONTRIBUTING.md#linting, to properly format/lint your code :)

libs/providers/langchain-openai/src/chat_models/responses.ts
```diff
@@ -73,7 +73,8 @@ export class ChatOpenAIResponses<
     let strict: boolean | undefined;
     if (options?.strict !== undefined) {
       strict = options.strict;
-    } else if (this.supportsStrictToolCalling !== undefined) {
+    }
+    if (strict === undefined && this.supportsStrictToolCalling !== undefined) {
```
MartijnLeplae, line 77: Did you try to run the tests? I would be interested to know if this is a breaking change

Nitinref:
Hey I ran pnpm test inside libs/providers/langchain-openai and all the tests passed fine on my side, I also checked the formatting with pnpm lint:eslint, and everything looks clean there too.
From what I’ve checked so far, this change shouldn’t cause any breaking issues.
Screenshot 2025-12-03 212116

christian-bromann (changes requested):
Thanks for the contribution @Nitinref! Two requests:

- can we resolve the formatting? `pnpm run format:check:fix` should do that for you.
- Can we have some unit tests for this?

Nitinref:
All requested changes have been addressed, the branch is updated with main,
and lint + tests are passing locally. Let me know if anything else is needed.

christian-bromann:
@MartijnLeplae any last thoughts?

MartijnLeplae:
Yes actually @christian-bromann. Am I right that the change does not actually do anything?
My initial idea was to change the `else if` to `if` to make the `strict` variable overwritable by `supportsStrictToolCalling`. 
The new code does do this, but it also added `strict === undefined` to the statement, which makes the behaviour equivalent to the current behaviour if I'm not mistaken.

Why did you add this extra check @Nitinref?

Nitinref:
Thanks for the clarification @MartijnLeplae 
You’re right — the extra strict === undefined check made the behavior equivalent to the previous implementation.
I misunderstood the intent initially.
The goal is:
if the user explicitly sets strict, always use that
otherwise, fall back to supportsStrictToolCalling
if neither is provided, default to false
I will update the PR to use the simplified logic:

const useStrict =
  options?.strict ?? options?.supportsStrictToolCalling ?? false;

And pass strict: useStrict to the OpenAI tool definition.
This will correctly allow overriding strict mode while preserving current behavior.
Let me push the update shortly.

MartijnLeplae:
I think your reasoning is not completely right @Nitinref 

Writing 
```typescript
const useStrict = options?.strict ?? options?.supportsStrictToolCalling ?? false;
```

Is not the same as saying overriding is possible. The `??` operator will only use `options?.supportsStrictToolCalling` if `options?.strict` is `undefined` or `null`. Making it a fallback, and not an override if it is given, which is the current behaviour.

Perhaps to clarify more, my initial issue is that the `createAgent` always sets strict to true in these lines: https://github.com/langchain-ai/langchainjs/blob/8a77e9af292b542d739b502ffbfbfd82dd12fcb4/libs/langchain/src/agents/nodes/AgentNode.ts#L839-L849
therefore options?.strict is never undefined or null when using agentNode, unless the agentNode code would be changed.

What do you think @christian-bromann ?

Nitinref:
@MartijnLeplae  since AgentNode always sets strict: true, options.strict never ends up undefined, so the fallback to supportsStrictToolCalling can't actually happen. I missed that connection earlier.I’ll update the AgentNode logic so user-provided strict truly overrides, and only fall back when it's not set. Will push the fix shortly. Thanks for catching this!

christian-bromann:
@Nitinref @MartijnLeplae I just realized that this conflicts [with a PR I put up last week](https://github.com/langchain-ai/langchainjs/pull/9578/files) that would allow to overwrite `strict` through the `providerStrategy` primitive. Any thoughts on what approach is better here?

Nitinref:
@christian-bromann Thanks for pointing that out!
It looks like both changes try to solve the same issue but at different layers.
My view is:
If the user explicitly sets strict, that should take priority.
Otherwise, falling back to the providerStrategy’s strict value makes sense.
I’m happy to adapt this PR to follow the direction of your approach if providerStrategy is meant to be the main place for overriding strict.Just let me know which direction you prefer!

christian-bromann:
> I’m happy to adapt this PR to follow the direction of your approach if providerStrategy is meant to be the main place for overriding strict

Let's do that, I am happy to close out my PR if we can incorporate that idea here as well.

Let's recap: by default we maintain a __strict__ type validation. User can disable that by setting it explicitly to `false` via:

```ts
// json schema / using default strategy
const agent = createAgent({
  model: new ChatOpenAI({ model: "gpt-4o-mini" }),
  tools: [myToolWithOptionalParams],
  responseFormat: {
    // your output schema...
    strict: false,
  },
});
```
```ts
// tool/provider strategy explicitly defined
const agent = createAgent({
  model: new ChatOpenAI({ model: "gpt-4o-mini" }),
  tools: [myToolWithOptionalParams],
  responseFormat: toolStrategy({ // or providerStrategy
    schema: z.object({ ... }), // or json schema
    strict: false
  }),
});
```

Right?

Nitinref:
@christian-bromann  Thanks for the clarification!
This makes sense — I’ll update the PR so the strict resolution follows:

- user-provided strict
- otherwise providerStrategy.strict
- otherwise default true

Once I push the update, we can fully replace the logic from your PR.
I’ll tag you for review when it’s ready

Nitinref:
I've updated the PR to move strict resolution entirely into AgentNode as suggested.
The logic now resolves strict in the correct priority order:
1. user strict override
2. providerStrategy.strict
3. default true
Let me know if you'd like any additional adjustments!

christian-bromann:
> 2\. providerStrategy.strict

How do you set `strict` via provider strategy, my idea in the PR was to allow something like:

```ts
  responseFormat: providerStrategy({
    schema: z.object({ ... })
    strict: false
  })
```

christian-bromann:
I can go ahead and merge this change once pipeline passes and then merge my PR after.

Nitinref:
@christian-bromann Thanks for the review!

Just to clarify — the strict resolution priority you described is already implemented in this PR inside `AgentNode.ts`:
This matches the desired behavior:
1. user override
2. providerStrategy.strict
3. default true
If there's any additional adjustment needed, I'd be happy to update it!
```ts
const resolvedStrict =
  preparedOptions?.modelSettings?.strict ??
  structuredResponseFormat?.strategy?.strict ??
  true;

**Comment by christian-bromann** [comment]
But it currently doesn't allow to define the strict value for a specific response format strategy, right?

**Comment by christian-bromann** [comment]
You've outlined this example:
```
const agent = createAgent({
  model: new ChatOpenAI({ model: "gpt-4o-mini" }),
  tools: [myToolWithOptionalParams],
  responseFormat: {
    // your output schema...
    strict: false,
  },
});
```

which I assume uses JSON schema. But what happens if I want to use a Zod schema?

**Comment by christian-bromann** [comment]
@Nitinref it seems like the current PR is not complete:
```
TypeCheckError: Property 'strict' does not exist on type 'ProviderStrategy<unknown>'.
 ❯ src/agents/nodes/AgentNode.ts:840:45
    838|       const resolvedStrict =
    839|         preparedOptions?.modelSettings?.strict ??
    840|         structuredResponseFormat?.strategy?.strict ??
       |                                             ^
    841|         true;
    842| 

```
?

christian-bromann:
Embedding this PR in #9578

Nitinref:
Thanks for the clarification!
Glad to see the strict handling landed cleanly via ProviderStrategy — makes sense to expose it at the API level. Happy to help test or document this if needed.

PR #9541, opened 2025-12-03
feat(anthropic): support computer use tool

christian-bromann:
This PR adds support for Anthropic's Computer Use tool to `@langchain/anthropic`, enabling Claude to interact with desktop environments through screenshot capture, mouse control, and keyboard input for autonomous desktop interaction.

### Changes

- **New `computer_20251124` function** - For Claude Opus 4.5
  - All actions including the new `zoom` action for viewing specific screen regions at full resolution
  - Supports `enableZoom` option to enable zoom functionality

- **New `computer_20250124` function** - For Claude 4 models and Claude 3.7
  - Enhanced actions: `screenshot`, `left_click`, `right_click`, `middle_click`, `double_click`, `triple_click`, `left_click_drag`, `left_mouse_down`, `left_mouse_up`, `scroll`, `type`, `key`, `mouse_move`, `hold_key`, `wait`

- **Added strongly-typed action interfaces** in `types.ts`:
  - `Computer20251124Action` - Union type with zoom action
  - `Computer20250124Action` - Union type for Claude 4/3.7
  - Individual action interfaces: `ComputerScreenshotAction`, `ComputerLeftClickAction`, `ComputerScrollAction`, `ComputerTypeAction`, `ComputerKeyAction`, `ComputerZoomAction`, etc.

- **Updated tools export** - Added `computer_20251124` and `computer_20250124` to the `tools` namespace

### Usage

```typescript
import { ChatAnthropic, tools } from "@langchain/anthropic";

  model: "claude-sonnet-4-5",
  clientOptions: {
    defaultHeaders: { "anthropic-beta": "computer-use-2025-01-24" },
  },
});

const computer = tools.computer_20250124({
  displayWidthPx: 1024,
  displayHeightPx: 768,
  displayNumber: 1,
  execute: async (action) => {
    if (action.action === "screenshot") {
      return captureScreenshot(); // Return base64-encoded image
    }
    if (action.action === "left_click") {
      await click(action.coordinate[0], action.coordinate[1]);
      return captureScreenshot();
    }
    // Handle other actions...
  },
});

const llmWithComputer = model.bindTools([computer]);
const response = await llmWithComputer.invoke(
  "Save a picture of a cat to my desktop."
);
```

### DIFF
--- a/.changeset/eight-cheetahs-dress.md
+++ b/.changeset/eight-cheetahs-dress.md
@@ -1,0 +1,5 @@
+---
+"@langchain/openai": minor
+---
+
+feat(openai): support web search tool


libs/providers/langchain-anthropic/src/tools/computer.ts
hntrl, line 1: Can you use a claude 3.7 computer tool with claude 4.5? If not I wonder if it makes sense to collapse it into one tool and then do tool type selection when we go to serialize the tool

christian-bromann: > Can you use a claude 3.7 computer tool with claude 4.5?

~Yep, you can.~ Actually you can't. You can run a modern model on an older tool but not vice verca.

christian-bromann: > If not I wonder if it makes sense to collapse it into one tool and then do tool type selection when we go to serialize the tool

Which tool name should we export then? I've been sticking with the version number in the name so we can just mimic what Anthropic exposes and don't have to explain users to which tool one will resolve when used by which model.

PR #9551, opened 2025-12-03
feat(anthropic): support for mcp toolset tool

christian-bromann:
Adds support for Anthropic's MCP (Model Context Protocol) connector feature, which enables Claude to connect to remote MCP servers directly from the Messages API without implementing a separate MCP client.

## Motivation

The MCP connector is a new Anthropic feature (beta: `mcp-client-2025-11-20`) that allows users to access tools from MCP servers through a simple API configuration. This eliminates the need to implement a separate MCP client while still benefiting from the MCP ecosystem of tools.

## Usage

```typescript
import { ChatAnthropic, tools } from "@langchain/anthropic";

  model: "claude-sonnet-4-5-20250929",
});

// Basic usage - enable all tools from an MCP server
const response = await llm.invoke("What tools do you have available?", {
  mcp_servers: [{
    type: "url",
    url: "https://example-server.modelcontextprotocol.io/sse",
    name: "example-mcp",
    authorization_token: "YOUR_TOKEN",
  }],
  tools: [
    tools.mcpToolset_20251120({ serverName: "example-mcp" }),
  ],
});

// Allowlist pattern - enable only specific tools
const response2 = await llm.invoke("Search for events", {
  mcp_servers: [{
    type: "url",
    url: "https://calendar.example.com/sse",
    name: "google-calendar-mcp",
  }],
  tools: [
    tools.mcpToolset_20251120({
      serverName: "google-calendar-mcp",
      defaultConfig: { enabled: false },
      configs: {
        search_events: { enabled: true },
        create_event: { enabled: true },
      },
    }),
  ],
});
```

### DIFF
--- a/.changeset/eight-cheetahs-dress.md
+++ b/.changeset/eight-cheetahs-dress.md
@@ -1,0 +1,5 @@
+---
+"@langchain/openai": minor
+---
+
+feat(openai): support web search tool


hntrl (commented):
Why not define this inside of the tool definition?

```ts
tools.mcpToolset_20251120({
  servers:     {
      type: "url",
      url: "https://example-server.modelcontextprotocol.io/sse",
      name: "example-mcp",
      authorization_token: "YOUR_TOKEN",
    },
}),
```

christian-bromann:
> Why not define this inside of the tool definition?

Because the tool defines configuration for a single MCP server. If we would extend that to allow configuring all server and their configuration, the tool would have to return an array of tools which seems awkward. Plus, `mcp_servers` seems like a reasonable property to add since we also have other tool properties on their like `container`.

PR #9561, opened 2025-12-04
fix: remove unused timeout

### LINKED ISSUE #9560: MCP tool invocations ignore custom timeout above 60 seconds
### Checked other resources

### Example Code

The following code:

```typescript
import { MultiServerMCPClient } from "@langchain/mcp-adapters";

const client = new MultiServerMCPClient({
	mcpServers: {
		remote: {
			url: "http://127.0.0.1:8000/mcp",
			transport: "http",
		},
	},
});

try {
	const tools = await client.getTools();
	if (tools.length > 0) {
		const firstTool = tools[0];
		if (!firstTool) {
			throw new Error("No tools available to invoke.");
		}

		const result = await firstTool?.invoke(
			{
				param1: "value1",
			},
			{ timeout: 100000 },
		);

		console.log("\nInvocation result:");
		console.log(JSON.stringify(result, null, 2));
	}
} catch (error) {
	console.error("Error:", error);
} finally {
	await client.close();
}

```

N/A

### Description

Specifying a timeout value greater than 60 seconds through the `RunnableConfig` option in the `invoke` function (as demonstrated in the example above) does not work as expected. Long-running MCP tool invocations consistently timeout after 60 seconds, even when a larger timeout is explicitly configured. This makes it impossible to successfully execute tools that require more than 60 seconds to complete.

The issue originates from this line in [config.ts](https://github.com/langchain-ai/langchainjs/blob/dc396c417a07dc1748099c5c71f47ff7e749796a/libs/langchain-core/src/runnables/config.ts#L188): https://github.com/langchain-ai/langchainjs/blob/dc396c417a07dc1748099c5c71f47ff7e749796a/libs/langchain-core/src/runnables/config.ts#L188

where the `timeout` parameter is deleted from the configuration object. For proper functionality, this timeout value should be passed through to the `options` parameter when calling the [MCP TypeScript SDK](https://github.com/modelcontextprotocol/typescript-sdk)'s `callTool` function. 

When the timeout is not provided, the MCP TypeScript SDK falls back to its default `DEFAULT_REQUEST_TIMEOUT_MSEC` constant of 60 seconds, which triggers the timeout behavior observed. The relevant SDK code can be found here: [protocol.ts#L1185](https://github.com/modelcontextprotocol/typescript-sdk/blob/724fe6eefeeb7f71328bc43c223fc9abd3df491e/src/shared/protocol.ts#L1185)
```typescript
// taken from https://github.com/modelcontextprotocol/typescript-sdk/blob/724fe6eefeeb7f71328bc43c223fc9abd3df491e/src/shared/protocol.ts
export const DEFAULT_REQUEST_TIMEOUT_MSEC = 60000;
...
const timeout = options?.timeout ?? DEFAULT_REQUEST_TIMEOUT_MSEC;
```

**Expected Behavior**: the timeout value from `RunnableConfig` should propagate to the MCP SDK's `callTool` options, allowing tools to run for the specified duration.

Fixed with #9561 

### System Info

Versions:
- langchain@1.1.3
- pnpm 10.18.3
- Node v24.11.1

FabrizioSandri:
Fixes #9560

### DIFF
--- a/libs/langchain-core/src/runnables/config.ts
+++ b/libs/langchain-core/src/runnables/config.ts
@@ -185,7 +185,6 @@ export function ensureConfig<CallOptions extends RunnableConfig>(
     } else {
       empty.signal = timeoutSignal;
     }
-    delete empty.timeout;
   }
   return empty as CallOptions;
 }


christian-bromann:
@FabrizioSandri thanks for the contribution. Can you provide more details why this delete statement was put there in the first place?

FabrizioSandri:
The `delete empty.timeout` statement exists because LangChain's core `RunnableConfig` handling creates an `AbortSignal` from the timeout value at a higher level in the execution chain. Specifically, the timeout is converted into an abort signal before being passed down to tool invocations.
https://github.com/langchain-ai/langchainjs/blob/dc396c417a07dc1748099c5c71f47ff7e749796a/libs/langchain-core/src/runnables/config.ts#L183

For standard `StructuredTool` implementations (non-MCP tools), this approach works correctly because the timeout value is consumed to create an `AbortSignal` at the config processing level.

However, MCP tools require different handling because they implement their own timeout mechanism at the MCP TypeScript SDK level. The SDK's `callTool` function expects an explicit `timeout` value in its options parameter and uses this pattern:

```typescript
const timeout = options?.timeout ?? DEFAULT_REQUEST_TIMEOUT_MSEC;
```

When the `timeout` field is deleted from the config before reaching the MCP tool invocation, the SDK falls back to `DEFAULT_REQUEST_TIMEOUT_MSEC` (60 seconds), regardless of what timeout the user specified in their config passed to `invoke`.

pawel-twardziak:
Hi @christian-bromann @FabrizioSandri there is already a PR on that. It provides more comprehensive solution.
https://github.com/langchain-ai/langchainjs/pull/9165
Removing the line that you removed @FabrizioSandri breaks something - I don't remember what it was, but unfortunately that's not the solution. I've tried that.

christian-bromann:
Thanks for letting us know. Will close this in favor of #9165 then.

PR #9562, opened 2025-12-04
fix(langchain): enable Anthropic native structured output with create agent

christian-bromann:
This patch enables native structured output for Anthropic models.


libs/langchain/src/load/import_constants.ts
hntrl, line 1: how did this happen 🙈

PR #9578, opened 2025-12-05
fix(langchain): allow to set strict tag manually in providerStrategy

christian-bromann:
ref https://github.com/langchain-ai/langchain/pull/34149

### DIFF
--- a/libs/langchain/src/agents/nodes/AgentNode.ts
+++ b/libs/langchain/src/agents/nodes/AgentNode.ts
@@ -835,13 +835,18 @@ export class AgentNode<
      * check if the user requests a native schema output
      */
     if (structuredResponseFormat?.type === "native") {
+      const resolvedStrict =
+        preparedOptions?.modelSettings?.strict ??
+        structuredResponseFormat?.strategy?.strict ??
+        true;
+
       const jsonSchemaParams = {
         name: structuredResponseFormat.strategy.schema?.name ?? "extract",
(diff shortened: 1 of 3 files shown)


hntrl (commented):
I'm generally not a fan when libraries bring positional args to APIs. Instead of having the second arg be strict, can we have a union for the first type that allows for either: a generic schema object, or an object that has `schema` + `strict` args? E.g.

```ts
providerStrategy(z.object(...))
// or
providerStrategy({ schema: z.object(...), strict: true })
```

(this is how we do it in `withStructuredOutput`)

christian-bromann:
@hntrl thanks for the feedback! Updated PR accordingly.

libs/langchain/src/agents/responses.ts
```diff
@@ -145,19 +145,33 @@ export class ProviderStrategy<T = unknown> {
   // @ts-expect-error - _schemaType is used only for type inference
   private _schemaType?: T;
 
-  private constructor(public readonly schema: Record<string, unknown>) {}
+  private constructor(
```
hntrl, line 163: nit: can we give a similar overload treatment to the constructor? Ik its internal but just for consistencies sake

```ts
new ProviderStrategy({ schema: ..., strict: true });
```

christian-bromann: added overwrites

PR #9579, opened 2025-12-06
feat(openai): add support for shell tool

christian-bromann:
This PR adds support for OpenAI's Shell tool, which allows `gpt-5.1` models to execute shell commands through the Responses API. Unlike the existing `localShell` tool (designed for Codex CLI), this tool supports executing multiple commands concurrently and uses a different interface optimized for the newer model.

### Usage

```typescript
import { ChatOpenAI, tools } from "@langchain/openai";
import { exec } from "node:child_process/promises";

const model = new ChatOpenAI({ model: "gpt-5.1" });

const shellTool = tools.shell({
  execute: async (action) => {
    const outputs = await Promise.all(
      action.commands.map(async (cmd) => {
        const { stdout, stderr } = await exec(cmd);
        return {
          stdout,
          stderr,
          outcome: { type: "exit" as const, exit_code: 0 },
        };
      })
    );
    return { output: outputs, maxOutputLength: action.max_output_length };
  },
});

const llmWithShell = model.bindTools([shellTool]);
const response = await llmWithShell.invoke("List all TypeScript files");
```

### Key Differences from `localShell`

| Feature | `localShell` | `shell` |
|---------|--------------|---------|
| Model | `codex-mini-latest` | `gpt-5.1` |
| Commands | Single command (argv array) | Multiple commands (string array) |
| Execution | Sequential | Concurrent support |
| Return format | String output | Structured `ShellResult` |

### DIFF
--- a/libs/providers/langchain-openai/README.md
+++ b/libs/providers/langchain-openai/README.md
@@ -508,6 +508,85 @@ const response = await llmWithShell.invoke(
 
 For more information, see [OpenAI's Local Shell Documentation](https://platform.openai.com/docs/guides/tools-local-shell).
 
+### Shell Tool
+
+The Shell tool allows models to run shell commands through your integration. Unlike Local Shell, this tool supports executing multiple commands concurrently and is designed for `gpt-5.1`.
+
+> **Security Warning**: Running arbitrary shell commands can be dangerous. Always sandbox execution or add strict allow/deny-lists before forwarding commands to the system shell.
+
+**Use cases**:
+
+- **Automating filesystem or process diagnostics** – e.g., "find the largest PDF under ~/Documents"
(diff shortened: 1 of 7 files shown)


libs/providers/langchain-openai/README.md
```diff
@@ -508,6 +508,87 @@ const response = await llmWithShell.invoke(
...
+  output: Array<{
+    stdout: string;
+    stderr: string;
+    outcome: { type: "exit"; exit_code: number } | { type: "timeout" };
+  }>;
+  maxOutputLength?: number | null; // Pass back from action if provided
+}
+```
+
+> **Note**: Only available through the Responses API with `gpt-5.1`. The `timeout_ms` from the model is only a hint—always enforce your own limits.
```
hntrl: think we can do away with this note -- we obscure the difference between responses + completions under the hood

PR #9584, opened 2025-12-07
fix(@langchain/google-genai): safe access around custom content parts

encodedz:
Description
Fix `parts` property access on converting response content into chunk when model returns content without a `parts` property such as:
```
 { role: 'model' } { finishReason: 'STOP', index: 0 }
```

Problem
Currently it's assumed that every content candidate has a `parts` property defined when accessed to check its `functionCalls` (unlike other places which has actually safe access to `parts`)

Solution
Wrapping with safe access

Changes
Updated 2 occurrences where we destruct the parts without safe access.

### DIFF
--- a/.changeset/good-pillows-reflect.md
+++ b/.changeset/good-pillows-reflect.md
@@ -1,0 +1,5 @@
+---
+"@langchain/google-genai": patch
+---
+
+safe access around custom content parts


libs/providers/langchain-google-genai/src/utils/common.ts
```diff
@@ -527,7 +527,7 @@ export function mapGenerateContentResultToChatResult(
   }
   const [candidate] = response.candidates;
   const { content: candidateContent, ...generationInfo } = candidate;
-  const functionCalls = candidateContent.parts.reduce((acc, p) => {
+  const functionCalls = candidateContent.parts?.reduce((acc, p) => {
```
wadamek65, line 530: @hntrl @encodedz 
```suggestion
  const functionCalls = candidateContent?.parts?.reduce((acc, p) => {
```

This is missing one more optional chain as the bug now evolved to:

```
Cannot read properties of undefined (reading 'parts')
```

Image

PR #9598, opened 2025-12-09
chore(google): test + converter enhancements

hntrl:


libs/providers/langchain-google/src/utils/handler.ts
hntrl, line 1: fine to just put `/* eslint-disable @typescript-eslint/no-explicit-any */` at the top of the file instead of an ignore clause at every step

libs/providers/langchain-google/src/converters/messages.ts
```diff
@@ -344,6 +344,116 @@ function convertStandardContentMessageToGeminiContent(
   return { role, parts };
 }
 
+function convertLegacyContentMessageToGeminiContent(
```
hntrl, line 467: jsdocs are important for these functions IMO -- I usually just tell cursor "write extensive jsdocs for this" and it gives me a pretty good output

hntrl: also would export

libs/providers/langchain-google/src/converters/messages.ts
```diff
@@ -696,3 +711,49 @@ export const convertGeminiCandidateToAIMessage: Converter<
     response_metadata,
   });
 };
+
+function addModalityCounts(
```
hntrl: nit: can we move this as a function inside of `convertGeminiGenerateContentResponseToUsageMetadata`?

libs/providers/langchain-google/src/utils/handler.ts
afirstenberg: Noted. Cleaned up

libs/providers/langchain-google/src/converters/messages.ts
afirstenberg: I've added a smaller jsdoc to match the one for `convertStandardContentMessageToGeminiContent()`, both of them referencing `convertMessagesToGeminiContents()` which is where they're primarily used. I mostly broke them out into their own functions because having 100 or so lines inside if/then conditions were unwieldy and I'm not sure anything else would use them, so not sure we need to export.

libs/providers/langchain-google/src/converters/messages.ts
afirstenberg: Yeah... I was going back and forth on this. Your nit tipped it the other way. {:

PR #9599, opened 2025-12-09
chore(ci): add ready label workflow

hntrl:


.github/workflows/pr-approval-label.yml
```diff
@@ -0,0 +1,27 @@
...
+
+permissions:
+  contents: read
+  pull-requests: write
+
+jobs:
+  add-ready-label:
+    runs-on: ubuntu-latest
+    # Only run when an org member approves
+    if: github.event.review.state == 'approved' && github.event.review.author_association == 'MEMBER'
```
christian-bromann, line 15: Maybe we should tweak this to say, when PR state is "open"

PR #9601, opened 2025-12-09
feat(zhipuai): support tool calling in ChatZhipuAI

### LINKED ISSUE #8211: How to implement bindtools and call agent in ChatZhipuAI

Orenoid:
## Summary
This PR fixes #8211.

## Changes
- Implemented the `bindTools` method for ChatZhipuAI, adding support for tool calling in both `invoke` and `stream`.
- Fixed two previously failing tests whose expected errors do not occur because ZhipuAI’s API actually handles these cases correctly.

### DIFF
--- a/.changeset/clean-poets-double.md
+++ b/.changeset/clean-poets-double.md
@@ -1,0 +1,5 @@
+---
+"@langchain/community": patch
+---
+
+feat(zhipuai): support tool calling in ChatZhipuAI


libs/langchain-community/src/chat_models/tests/chatzhipuai.int.test.ts
```diff
@@ -118,7 +160,7 @@ const testConfigs: TestConfig[] = [
     config: {
       description: "illegal input in streaming mode should throw an error",
       streaming: true,
-      temperature: 0,
+      temperature: -1,
```
Orenoid, line 163: for some reason, ZhipuAI's API works fine with the original parameters (temperature: 0), and the test expecting an error doesn't pass any more. changed it to an invalid one therefore.

libs/langchain-community/src/chat_models/tests/chatzhipuai.int.test.ts
```diff
@@ -35,6 +45,38 @@ test.skip("Test chat.stream work fine", async () => {
   expect(chunks.length).toBeGreaterThan(0);
 });
 
+test.skip("Test tool calling should work within chat.invoke", async () => {
```
christian-bromann:
```suggestion
test("Test tool calling should work within chat.invoke", async () => {
```

libs/langchain-community/src/chat_models/tests/chatzhipuai.int.test.ts
```diff
@@ -35,6 +45,38 @@ test.skip("Test chat.stream work fine", async () => {
...
+    new HumanMessage("How is the weather in New York"),
+  ]);
+  expect(response.tool_calls?.length).toBeGreaterThanOrEqual(1);
+  expect(
+    response.tool_calls &&
+      response.tool_calls.some((toolCall) => toolCall.name == "get_weather")
+  ).toBe(true);
+});
+
+test.skip("Test tool calling should work within chat.stream", async () => {
```
christian-bromann:
```suggestion
test("Test tool calling should work within chat.stream", async () => {
```

libs/langchain-community/src/chat_models/tests/chatzhipuai.int.test.ts
Orenoid: @christian-bromann Hi, thanks for the review. 
I skipped these two tests (which already passed) because the existing integration tests in the same file are skipped as well. I also noticed the same pattern in several other community test files, for example `chatllama_cpp.int.test.ts` and `minimax.int.test.ts`.
I’m not sure why this is the case—I couldn’t find any documentation or comments explaining it, so it may be a legacy issue.
That said, if we should unskip these tests, should I also update the existing ones (at least those in the same file)?

Orenoid: Done as suggested.

PR #9602, opened 2025-12-09
fix(openai/anthropic): extend tool schema for computer use

christian-bromann:
After merging the tools, I am building some example use cases and will fix issues I find along the way.

### DIFF
--- a/libs/providers/langchain-anthropic/src/tools/memory.ts
+++ b/libs/providers/langchain-anthropic/src/tools/memory.ts
@@ -41,9 +41,7 @@ import {
  *
  * @see https://docs.anthropic.com/en/docs/agents-and-tools/tool-use/memory-tool
  */
-export function memory_20250818(
-  options?: MemoryTool20250818Options
-): DynamicStructuredTool {
+export function memory_20250818(options?: MemoryTool20250818Options) {
   const memoryTool = tool(
     options?.execute as (
       input: unknown,


libs/providers/langchain-anthropic/src/tools/toolSearch.ts
```diff
@@ -55,12 +56,12 @@ interface ToolSearchOptions {
  */
 export function toolSearchRegex_20251119(
   options?: ToolSearchOptions
-): Anthropic.Beta.Messages.BetaToolSearchToolRegex20251119 {
+): ServerTool {
```
hntrl, line 59: Why are we casting to ServerTool here? Are we expecting to beef up that type somehow? My preference would be that we're just returning the anthropic type

libs/providers/langchain-anthropic/src/tools/types.ts
```diff
@@ -1,30 +1,80 @@
...
-export type Memory20250818StrReplaceCommand =
-  Anthropic.Beta.BetaMemoryTool20250818StrReplaceCommand;
-export type Memory20250818InsertCommand =
-  Anthropic.Beta.BetaMemoryTool20250818InsertCommand;
-export type Memory20250818DeleteCommand =
-  Anthropic.Beta.BetaMemoryTool20250818DeleteCommand;
-export type Memory20250818RenameCommand =
-  Anthropic.Beta.BetaMemoryTool20250818RenameCommand;
+
+// Zod schemas for memory commands
```
hntrl, line 10: Are these schema types exposed from anthropic at all? I wonder if it makes sense to bind an object shape that complies with these type to detect where we regress:

```
export const Memory20250818ViewComandSchema: z.ZodObject<SpecialAnthropicType> = z.object({ ... });

File: libs/providers/langchain-anthropic/src/tools/toolSearch.ts (hunk 1 above)
Comment (christian-bromann) [reply libs/providers/langchain-anthropic/src/tools/toolSearch.ts:59]: That was my intend before but it causes TypeScript to throw an error when you export the tool because of missing references. That's why we just cast it to `ServerTool`.

Comment (christian-bromann) [reply libs/providers/langchain-anthropic/src/tools/toolSearch.ts:59]: meaning of you do this:

```ts
export memory = tools.memory({ ... })
```

libs/providers/langchain-anthropic/src/tools/types.ts
christian-bromann: Unfortunately not.

PR #9603, opened 2025-12-09
fix(core): add comment why we remove the timeout comment

christian-bromann:
Adding comment that @pawel-twardziak was suppose to add.


pawel-twardziak:
Thanks @christian-bromann ! Sorry, it was late yesterday. I wanted to add the comment this morning, but I see it's already merged. Thanks again :)

hntrl (approved):

PR #9606, opened 2025-12-09
chore(classic): bin p-retry

patrykkopycinski:
Follow up for https://github.com/langchain-ai/langchainjs/pull/9575
Removes `p-retry` from `classic` lib

### DIFF
--- a/.changeset/thick-numbers-lay.md
+++ b/.changeset/thick-numbers-lay.md
@@ -1,0 +1,5 @@
+---
+"@langchain/community": patch
+---
+
+fix(community): IBM tests, add missing metadata properties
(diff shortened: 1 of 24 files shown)


libs/langchain-classic/src/util/hub.ts
```diff
@@ -1,5 +1,4 @@
-import pRetry from "p-retry";
-
+import pRetry from "@langchain/core/utils/p-retry";
```
christian-bromann: `@langchain/classic` doesn't have a dependency on `@langchain/core` , I think we should just do the same here too and duplicate what we've done in core.

patrykkopycinski: Thank you @christian-bromann, please let me know if it's what you meant and I haven't missed anything

christian-bromann:
@patrykkopycinski thanks for the contribution!

PR #9607, opened 2025-12-09
feat(ollama): add support for custom baseUrl

jonghwanhyeon:
This PR introduces support for the `OLLAMA_BASE_URL` environment variable across the `@langchain/ollama` provider, enabling global configuration of the Ollama server URL.

**Motivation**: This change is particularly useful for integration testing when the Ollama server is running on a host other than `localhost`

### DIFF
--- a/.changeset/thick-numbers-lay.md
+++ b/.changeset/thick-numbers-lay.md
@@ -1,0 +1,5 @@
+---
+"@langchain/community": patch
+---
+
+fix(community): IBM tests, add missing metadata properties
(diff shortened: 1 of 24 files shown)


jonghwanhyeon:
The motivation is that it’s hard to set `baseUrl` when performing integration tests because it assumes I’m running an Ollama server on my host. Since Ollama is widely used as a self-hosted solution, `baseUrl` is more variable than with other providers, where the `baseUrl` is usually fixed.

hntrl:
Gotcha, think that sounds reasonable. Thanks for the fix!

PR #9609, opened 2025-12-09
chore(standard-tests): fix CI

hntrl:


libs/langchain-standard-tests/src/vitest.ts
```diff
@@ -1,2 +1,2 @@
-export * from "./unit_tests/vitest.ts";
-export * from "./integration_tests/vitest.ts";
+export * from "./unit_tests/vitest.js";
+export * from "./integration_tests/vitest.js";
```
christian-bromann, line 2: This should remain `.ts` extensions since we don't compile the standard tests package to JS anymore.

libs/langchain-standard-tests/src/unit_tests/vitest.ts
```diff
@@ -1,12 +1,12 @@
 import { expect, describe, test } from "vitest";
 import { BaseChatModelCallOptions } from "@langchain/core/language_models/chat_models";
-import { BaseMessageChunk } from "@langchain/core/messages";
-import { BaseChatModelsTestsFields, RecordStringAny } from "../base.ts";
-import { ChatModelUnitTests as BaseChatModelUnitTests } from "./chat_models.ts";
+import { AIMessageChunk } from "@langchain/core/messages";
+import { BaseChatModelsTestsFields, RecordStringAny } from "../base.js";
+import { ChatModelUnitTests as BaseChatModelUnitTests } from "./chat_models.js";
```
christian-bromann, line 5: same here

PR #9631, opened 2025-12-12
feat(openai): Prefer responses API for 5.2 pro

jacoblee93:


hntrl (approved):

libs/providers/langchain-openai/src/utils/misc.ts
```diff
@@ -72,3 +72,7 @@ export function messageToOpenAIRole(
       throw new Error(`Unknown message type: ${type}`);
   }
 }
+
+export function _modelPrefersResponsesAPI(model: string): boolean {
```
christian-bromann, line 76: I personally don't think we need to underscore this function since we are not exporting it to users anyway.

PR #9637, opened 2025-12-12
fix(langchain): Prevent functions from being accidentally assignable to AgentMiddleware

christian-bromann:
Functions that return middleware instances can be accidentally passed to the `middleware` array without being called:

```typescript
const createMyMiddleware = () => createMiddleware({ name: "my-middleware", ... });

// Bug: passing function instead of calling it
createAgent({
  middleware: [createMyMiddleware],  // ❌ Should be createMyMiddleware()
});
```

This compiles without any TypeScript errors, leading to runtime failures when the agent tries to invoke middleware lifecycle hooks on a function object.

### Root Cause

In JavaScript/TypeScript (ES2022+), every function has a `name` property inherited from `Function.prototype`. Since `AgentMiddleware` only requires `name: string` and all other properties are optional, TypeScript's structural typing considers functions to be compatible with the interface:

```typescript
const fn = () => {};
fn.name;  // "fn" - functions have a name property!

// This is structurally compatible because:
// - fn.name is a string ✓
// - stateSchema is optional ✓
// - contextSchema is optional ✓
// - wrapModelCall is optional ✓
// - etc.
```

### Solution

Added a brand symbol to the `AgentMiddleware` interface that functions cannot satisfy:

```typescript
export const MIDDLEWARE_BRAND: unique symbol = Symbol("AgentMiddleware");

export interface AgentMiddleware<...> {
  readonly [MIDDLEWARE_BRAND]: true;  // Functions don't have this!
  name: string;
  // ... rest of interface
}
```

The `createMiddleware` function now includes this brand property, ensuring only properly created middleware instances are assignable to `AgentMiddleware`.

### DIFF
--- a/libs/langchain/src/agents/tests/utils.ts
+++ b/libs/langchain/src/agents/tests/utils.ts
@@ -33,6 +33,24 @@ import {
 import { LanguageModelLike } from "@langchain/core/language_models/base";
 import { z } from "zod/v3";
 
+/**
+ * Custom asymmetric matcher that matches any string value.
+ * Works with both Jest and Vitest's toEqual() assertions.
+ */
+class AnyString {
+  asymmetricMatch(other: unknown): boolean {
+    return typeof other === "string";
+  }
+
+  toString(): string {
+    return "Any<String>";
+  }
+
+  toAsymmetricMatcher(): string {
+    return "Any<String>";
+  }
+}
+
 export class _AnyIdAIMessage extends AIMessage {
   get lc_id() {
     return ["langchain_core", "messages", "AIMessage"];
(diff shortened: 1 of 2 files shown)


libs/langchain/src/agents/tests/utils.ts
hntrl, line 1: * is this meant to be in this PR?
* why is this necessary?

christian-bromann: It's a side issue that I've fixed while testing this. I am putting my scratchpad files in `./libs/langchain/src/` and when I import `createAgent` from `./index.ts` and run it with Bun, it throws an error. The problem is that we export primitives from `libs/langchain/src/agents/tests/utils.ts`, which means we can't have dev dependencies like `vitest` in there.

PR #9642, opened 2025-12-12
feat(mcp): add connection error handling modes

hntrl:
continuation of #8980


libs/langchain-mcp-adapters/src/client.ts
```diff
@@ -146,6 +146,11 @@ export class MultiServerMCPClient {
    */
   #config: ResolvedClientConfig;
 
+  /**
+   * Behavior when a server fails to connect
+   */
+  #onConnectionError: "throw" | "ignore";
```
christian-bromann: Should we allow to pass in a function?

PR #9654, opened 2025-12-13
fix(langchain): make embedded text from resources available to LLM

maxim-shilov-deel:
## Problem

When MCP tools return `EmbeddedResource` objects with embedded text content (e.g., `resource.text`), the text was not being extracted and made accessible to the LLM. This caused issues with tools like GitHub MCP's `get_file_contents`, which returns file content in `artifact[0].resource.text`, making the content inaccessible to the language model.

The previous implementation only checked `outputHandling` settings to determine where content should go, but didn't account for the fact that embedded text within resources should always be accessible to the LLM, regardless of where the resource structure itself is placed.

## Solution

Modified `_convertCallToolResult` to:
1. **Always extract embedded text** from `EmbeddedResource` objects that have a `text` property, promoting it to a standard text content block
2. **Respect `outputHandling` for resource structures** - the resource structure itself still follows `outputHandling` rules (can go to `content` or `artifact`)
3. **Simplify the logic** - removed duplicate filtering logic by building `contentBlocks` and `artifactBlocks` arrays in a single pass

This ensures that:
- Embedded text is always accessible to the LLM (fixes GitHub MCP and similar tools)
- Resource structures still follow user-configured `outputHandling` settings
- The fix is backward compatible - all existing tests pass without modification

## Changes

- **`libs/langchain-mcp-adapters/src/tools.ts`**:
  - Modified `_convertCallToolResult` to extract embedded text from `EmbeddedResource` objects
  - Fixed type guard in `_embeddedResourceToArtifact` to use `isResourceReference` helper
  - Simplified content/artifact separation logic

- **`libs/langchain-mcp-adapters/src/tests/tools.test.ts`**:
  - Added test for resources without text property
  - Added test for multiple embedded resources with text

## Testing

- All existing tests pass without modification (backward compatible)
- New tests verify:
  - Resources without text don't extract text
  - Multiple embedded resources correctly extract all texts

## Impact

This fix resolves issues with MCP tools that embed text content in resources, such as:
- GitHub MCP's `get_file_contents` tool
- Any other MCP tool that returns `EmbeddedResource` with `text` property

The fix is backward compatible and doesn't change behavior for tools that don't use embedded text in resources.

### DIFF
--- a/libs/langchain-mcp-adapters/src/tests/tools.test.ts
+++ b/libs/langchain-mcp-adapters/src/tests/tools.test.ts
@@ -737,5 +737,29 @@ describe("Simplified Tool Adapter Tests", () => {
       // Both resource structures should be in artifacts
       expect(toolMessageResult.artifact).toHaveLength(2);
     });
+
+    test("should handle resource reference correctly (regression: old approach checked resource.uri directly)", async () => {
+      mockClient.listTools.mockReturnValueOnce(
+        Promise.resolve({
+          tools: [{ name: "ref_tool", description: "", inputSchema: { type: "object", properties: {}, required: [] } }],
+        })
+      );
+      const tools = await loadMcpTools("mockServer", mockClient as Client);
+
+      // Resource reference: URI only, no blob/text
+      mockClient.callTool.mockReturnValue(
+        Promise.resolve({
+          content: [{ type: "resource", resource: { uri: "test://ref", mimeType: "text/plain" } }],
+        })
+      );
+      mockClient.readResource = vi.fn().mockResolvedValue({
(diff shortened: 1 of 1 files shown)


christian-bromann:
@maxim-shilov-deel thanks for the contribution! Mind updating the branch due to the conflict?

maxim-shilov-deel:
> @maxim-shilov-deel thanks for the contribution! Mind updating the branch due to the conflict?

I resolved conflicts, but now I think that the fix is incorrect. We can't copy embedded text to content. it will blow up LLM context.

For now I don't understand what would be the right fix. Do you have ideas @christian-bromann ?

PR #9655, opened 2025-12-13
Langchain agentstate

### LINKED ISSUE #9653: TypeDoc references non-exported AgentState / AgentBuiltInState in LangChain.js
### Checked other resources

### Example Code

```ts
import type { AgentState } from "langchain";
```

Image

Module '"langchain"' has no exported member 'AgentState'. (tsserver 2305)

### Description

In the LangChain.js documentation (TypeScript), under the **Short-Term Memory** section, there are multiple examples that show importing the `AgentState` type from `langchain`:

https://docs.langchain.com/oss/javascript/langchain/short-term-memory

```ts
import type { AgentState } from "langchain";
```

However, `AgentState` is not exported from the langchain JS package.

The referenced type appears to originate from the deprecated `@langchain/langgraph` interface:

https://reference.langchain.com/javascript/interfaces/_langchain_langgraph.prebuilt.AgentState.html

which is marked as “moved to langchain”, but no corresponding export exists.

### Related issues: non-exported `AgentBuiltInState`

Additionally, several public TypeScript examples and type signatures (middleware hooks) rely on a state parameter whose type is implicitly inferred by TypeScript.

In practice, this inferred type corresponds to the internal `AgentBuiltInState` shape (the same structure as deprecated `AgentState`). However, because `AgentBuiltInState` is not publicly exported, users cannot explicitly reference this type when annotating the state parameter, or writing reusable or generic helper functions that need to accept or return the agent state.

As a result, while inline usage works via inference, it is difficult to write well-typed reusable middleware or utilities without re-declaring internal types or falling back to any.

### Example:

```ts
import { createMiddleware, AIMessage } from "langchain";
// ❌ This import does not exist, but docs/types reference it
import type { AgentBuiltInState } from "langchain";

const createMessageLimitMiddleware = (maxMessages: number = 50) => {
  return createMiddleware({
    name: "MessageLimitMiddleware",
    beforeModel: (state: AgentBuiltInState) => {
      if (state.messages.length === maxMessages) {
        return {
          messages: [new AIMessage("Conversation limit reached.")],
          jumpTo: "end",
        };
      }
    },
    afterModel: (state) => {
      const lastMessage = state.messages[state.messages.length - 1];
      console.log(`Model returned: ${lastMessage.content}`);
    },
  });
};
```

Because `AgentBuiltInState` is not publicly exported, users cannot properly type middleware or state-handling logic without re-declaring internal types.

### Expected behavior

Public documentation and public type signatures should not reference non-exported symbols.

### Possible resolutions:

Export `AgentBuiltInState` (or `AgentState`) as a public alias, or Explicitly mark agent state types as internal and update docs accordingly

### System Info

LangChain Package Versions

```sh
@langchain/core: "^1.1.5"
@langchain/openai: "^1.2.0"
langchain: "^1.2.0"
```

Ayushsingla1:
Export AgentBuiltInState and define a public AgentState interface

Fixes #9653

This PR exposes the internal `AgentBuiltInState` type as a public export and defines a public `AgentState` interface in `langchain`.

This resolves a mismatch between the LangChain.js documentation and the actual package exports, where examples reference `AgentState` but no such public type exists.

### DIFF
--- a/libs/langchain/src/agents/types.ts
+++ b/libs/langchain/src/agents/types.ts
@@ -501,5 +501,5 @@ export interface AgentState<
   StructuredResponseType extends Record<string, any> = Record<string, any>
 > {
   messages: BaseMessage[];
-  structuredResponse: StructuredResponseType;
+  structuredResponse?: StructuredResponseType;
 }


christian-bromann:
Thank you @Ayushsingla1 for the contribution! We don't have an `AgentState` primitive yet. The examples you are referring to were outdated so I took a stab at updating them. Sorry for the confusion.

PR #9664, opened 2025-12-15
fix(langchain-openai): fix operator precedence in reasoning initialization

### LINKED ISSUE #9663: [v0.3] ChatOpenAI constructor overrides 'reasoning' option due to incorrect operator precedence
### Checked other resources

### Example Code

```js

// Attempting to use reasoning parameter for GPT-5 model
  model: "gpt-5.2",
  reasoning: {
    effort: "low",
    summary: "auto"
  },
  apiKey: "your-openai-api-key"
});

// Check if reasoning was set correctly
console.log(model.reasoning);
// Expected: { effort: "low", summary: "auto" }
// Actual: { effort: undefined }

// As a result, the reasoning parameter is not sent to OpenAI API
const response = await model.invoke("Hello wookingwoo!");
// console.log(response);

```

No error is thrown. The bug silently ignores the `reasoning` parameter, resulting in API calls without reasoning configuration.

### Problem

When initializing `ChatOpenAI` with a `reasoning` object (used for GPT-5 and o-series models), the parameter is silently ignored due to incorrect operator precedence in the null coalescing expression.

### What I'm trying to do

- I'm trying to use the `ChatOpenAI` class with OpenAI's new Responses API, which requires a `reasoning` parameter for GPT-5 models.
- I expect the `reasoning` object to be preserved and sent to the OpenAI API.
- Instead, the `reasoning` parameter is completely ignored, and API calls are made without reasoning configuration.

### Root Cause

The ChatOpenAI constructor incorrectly handles the `reasoning` field when it is provided in the options.

Due to an operator precedence issue with the ternary operator in `libs/langchain-openai/src/chat_models.ts`, providing `reasoning` causes the code to fall through to the branch that constructs the object from `reasoningEffort`. Since `reasoningEffort` is typically undefined in this case, the `reasoning` field ends up being overwritten with `{ effort: undefined }`.

**Code Snippet (Before Fix):**

```js
this.reasoning=
fields?.reasoning??fields?.reasoningEffort
? {effort:fields.reasoningEffort }
:undefined;

```

**Corrected Logic:**

```js
this.reasoning=
fields?.reasoning??
  (fields?.reasoningEffort
? {effort:fields.reasoningEffort }
:undefined);

```

### Branch

v0.3 (not main)

---

**Additional Context**:

This bug was discovered while implementing GPT-5 model support. The `reasoning` parameter is critical for OpenAI's new Responses API, which is the recommended way to use GPT-5 models according to OpenAI's documentation: https://platform.openai.com/docs/guides/reasoning

The bug exists in the v0.3 branch and affects all users trying to use reasoning models with the proper API parameters.

### System Info

- platform: Ubuntu 24.04.3 LTS
- Node version: v24.11.1
- npm version: 11.6.2
- @langchain/openai: 0.6.16

wookingwoo:
Fixes #9663 

## Description

This PR fixes a bug in the ChatOpenAI constructor where the `reasoning` parameter was incorrectly overridden due to an operator precedence issue.

## Problem

The original initialization logic was:

```jsx
this.reasoning =
    fields?.reasoning ?? fields?.reasoningEffort
        ? { effort: fields.reasoningEffort }
        : undefined;

```

When `fields.reasoning` is provided, the expression `fields?.reasoning ?? fields?.reasoningEffort` evaluates to the `reasoning` object (truthy). Consequently, the ternary operator executes the true branch: `{ effort: fields.reasoningEffort }`. 
However, since the user provided `reasoning`, `fields.reasoningEffort` is typically `undefined`. This results in the `reasoning` field being set to `{ effort: undefined }`, effectively discarding the user's input.

## Solution

I have added parentheses to correctly group the fallback logic. The code now only checks `reasoningEffort` if `reasoning` is null or undefined.

```jsx
this.reasoning=
fields?.reasoning??
  (fields?.reasoningEffort
? {effort:fields.reasoningEffort }
:undefined);

```

## Related Issues

- Fixes #9663
- Related to OpenAI Responses API documentation: https://platform.openai.com/docs/guides/reasoning

## Additional Context

This is a one-line fix in the v0.3 branch that restores critical functionality for GPT-5 and o-series models. The fix maintains full backward compatibility - existing code using `reasoningEffort` continues to work as expected.

- Contributor profile: https://www.linkedin.com/in/ronny-woo/

### DIFF
--- a/.changeset/neat-candles-share.md
+++ b/.changeset/neat-candles-share.md
@@ -1,0 +1,5 @@
+---
+"@langchain/openai": patch
+---
+
+fix operator precedence in reasoning initialization


wookingwoo:
> Can we add a test for this so we don't regress?

Great suggestion! I've added tests for this - see the "reasoning field initialization" test suite.

We can run just these tests with:
```bash
cd libs/langchain-openai
yarn test chat_models.test.ts -t "reasoning field initialization"
```

wookingwoo:
@christian-bromann Tests are added and the fix is ready.
I’d appreciate a final review/approval when you get a chance. Thanks!

PR #9665, opened 2025-12-15
feat(core): Make runnable transform trace in a single payload in LangChainTracer

jacoblee93:


libs/langchain-core/src/callbacks/base.ts
```diff
@@ -138,11 +138,12 @@ abstract class BaseCallbackHandlerMethodsClass {
     chain: Serialized,
     inputs: ChainValues,
     runId: string,
-    parentRunId?: string,
+    runType?: string,
```
jacoblee93, line 141: This is super unfortunate but only changes semantics that are misused in practice. Changing the actual implementation passed in the `CallbackManager` would be breaking.

PR #9668, opened 2025-12-16
fix: Cannot merge two undefined objects error

bracesproul:
Fixes # (issue)

### DIFF
--- a/libs/langchain-core/src/messages/base.ts
+++ b/libs/langchain-core/src/messages/base.ts
@@ -546,12 +546,7 @@ export function _mergeObj<T = any>(
     return (left + right) as T;
   } else if (Array.isArray(left) && Array.isArray(right)) {
     return _mergeLists(left, right) as T;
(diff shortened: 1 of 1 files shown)


libs/langchain-core/src/messages/base.ts
```diff
@@ -546,8 +546,16 @@ export function _mergeObj<T = any>(
...
   } else if (Array.isArray(left) && Array.isArray(right)) {
     return _mergeLists(left, right) as T;
-  } else if (typeof left === "object" && typeof right === "object") {
-    return _mergeDicts(left, right) as T;
+  } else if (
+    typeof left === "object" &&
+    typeof right === "object" &&
+    left !== null &&
+    right !== null
+  ) {
```
hntrl: _mergeDicts handles null inputs
```suggestion
  } else if (typeof left === "object" && typeof right === "object") {
```

PR #9225, opened 2025-10-17
fix(community): IBM WatsonX ChatWatsonx does not cancel requests on AbortSignal

noobgramming:
## Description

Fixes IBM WatsonX ChatWatsonx integration to properly handle AbortSignal by passing the signal parameter through to the underlying IBM SDK methods, enabling true server-side request cancellation.

## Issue

Fixes a bug where abort signals were accepted but not actually passed to the IBM SDK, causing:
- Requests to continue executing on WatsonX servers after abort
- Users being charged for tokens generated after abort call
- Delayed cancellation (waiting for completion instead of immediate termination)

This is the same bug pattern as:
- [#7257](https://github.com/langchain-ai/langchainjs/issues/7257) - Qianfan chat models not handling abort signal (labeled `auto:bug`)
- [#5001](https://github.com/langchain-ai/langchainjs/issues/5001) - OpenAI abort controller not working (labeled `auto:bug`)

Both issues involved signals being passed to AsyncCaller but not to the underlying provider SDK.

## Changes

Added `signal: options?.signal` to 4 SDK method calls:

**Non-streaming methods** (_generate):
```typescript
// textChat() - line 680
this.service.textChat({
  ...params,
  ...scopeId,
  messages: watsonxMessages,
  signal: options?.signal,  // Added
}, watsonxCallbacks)

// deploymentsTextChat() - line 671
this.service.deploymentsTextChat({
  ...scopeId,
  messages: watsonxMessages,
  signal: options?.signal,  // Added
}, watsonxCallbacks)
```

**Streaming methods** (_streamResponseChunks):
```typescript
this.service.textChatStream({
  ...params,
  ...scopeId,
  messages: watsonxMessages,
  returnObject: true,
  signal: options?.signal,  // Added
}, watsonxCallbacks)

this.service.deploymentsTextChatStream({
  ...scopeId,
  messages: watsonxMessages,
  returnObject: true,
  signal: options?.signal,  // Added
}, watsonxCallbacks)
```

Also removed ineffective manual abort checks that happened after requests completed.

## Technical Details

The IBM WatsonX SDK already supports AbortSignal via the `DefaultParams` interface:

```typescript
// From @ibm-cloud/watsonx-ai
interface DefaultParams {
  signal?: AbortSignal;
  headers?: OutgoingHttpHeaders;
}

export interface TextChatParams extends DefaultParams {
  modelId: string;
  messages: TextChatMessages[];
  // signal inherited from DefaultParams ✅
}
```

**Signal flow:**
[...]

PR #9671, opened 2025-12-16
feat[community]: support aborting for Model Gateway and WatsonxLLM for IBM

FilipZmijewski:
Continuation of: https://github.com/langchain-ai/langchainjs/pull/9225
The above PR introduced aborting via passing signal to SDK for ChatWatsonx, it is also needed for MG and WatsonxLLM which is presented in this PR.
Tests were added accordingly

### DIFF
--- a/libs/langchain-community/src/chat_models/ibm.ts
+++ b/libs/langchain-community/src/chat_models/ibm.ts
@@ -845,7 +845,7 @@ export class ChatWatsonx<
           ...scopeId,
           signal,
           stream,
-          returnObject: stream ? true : undefined, // this is redudant but with current tsconfig it is expected
+          ...(stream ? { returnObject: true } : {}),
           messages,
         });
       }


libs/langchain-community/src/chat_models/tests/ibm.int.test.ts
```diff
@@ -396,6 +394,7 @@ describe.each(parameters)(
         for await (const chunk of res) {
           chunks.push(chunk);
         }
+        console.log(chunks);
```
hntrl:
```suggestion
```

libs/langchain-community/src/chat_models/ibm.ts
```diff
@@ -843,7 +843,9 @@ export class ChatWatsonx<
         return this.gateway.chat.completion.create({
           ...params,
           ...scopeId,
+          signal,
           stream,
+          returnObject: stream ? true : undefined, // this is redudant but with current tsconfig it is expected
```
hntrl: Could you elaborate on this?

this is still technically a defined key, could we instead do something like this?
```suggestion
          ...(stream ? { returnObject: true } : {}),
```

FilipZmijewski: Alright, so actually this works like this as well. We need this despite the fact whenever `returnObject` is not passed return type should be the same as when passing `returnObject: true` see here:  
Screenshot 2025-12-16 at 18 16 25
Screenshot 2025-12-16 at 18 16 37

However regarding new changes in `tsconfig.json` in langchain I do not think TS catches this properly so passing `returnObejct` as true is necessary to see correct return type. Maybe this has something to do with our SDK ternary type implementation
```typescript
    create<S extends boolean | undefined = false, R extends boolean | undefined = undefined>(params: CreateChatCompletionsParams & {
        stream?: S;
        returnObject?: R;
    }): Promise<ChatCompletionResult<S, R>>;
```

hntrl (approved):

PR #9679, opened 2025-12-17
feat(openai): elevate OpenAI image generation outputs to proper image content blocks

christian-bromann:
This PR improves the handling of OpenAI's `image_generation_call` outputs from the Responses API by converting them into proper `ContentBlock.Multimodal.Image` content blocks, making it easier to work with generated images without needing custom extraction logic.

### Problem

Previously, when using OpenAI's image generation tool via the Responses API, the generated images were only accessible through `additional_kwargs.tool_outputs`, requiring users to write helper functions to extract and identify image outputs. I recognised this when working on the [AI SDK streaming PR](https://github.com/vercel/ai/pull/11238/files#diff-71954218f9cdc28090f64b9df9b1dc12ecad7abe39aab8ae0b225a18e30ffddaR624).

### Changes

**`libs/providers/langchain-openai/src/converters/responses.ts`**
- `convertResponsesMessageToAIMessage`: Adds image content blocks for `image_generation_call` items with available results
- `convertResponsesDeltaToChatGenerationChunk`: Adds streaming support for image generation outputs

**`libs/langchain-core/src/messages/block_translators/openai.ts`**
- `convertToV1FromResponses`: Converts `image_generation_call` tool outputs to standard image content blocks

### Backwards Compatibility

The original `image_generation_call` objects are **still preserved** in `additional_kwargs.tool_outputs` to maintain backwards compatibility and support multi-turn image editing (which requires the image generation call ID).

### DIFF
--- a/libs/providers/langchain-openai/src/tools/tests/imageGeneration.int.test.ts
+++ b/libs/providers/langchain-openai/src/tools/tests/imageGeneration.int.test.ts
@@ -37,7 +37,7 @@ describe("OpenAI Image Generation Tool Integration Tests", () => {
      */
     const expectedContent = response.content as [
       ContentBlock.Multimodal.Image,
-      ContentBlock.Text // expected empty text block
+      ContentBlock.Text, // expected empty text block
     ];
     expect(expectedContent.length).toBe(2);
     expect(expectedContent[0].type).toBe("image");


libs/providers/langchain-openai/src/tools/imageGeneration.ts
```diff
@@ -219,15 +219,15 @@ function convertInputImageMask(
...
-    partial_images: options?.partialImages,
-    quality: options?.quality,
-    size: options?.size,
+    model: options?.model ?? "gpt-image-1",
+    moderation: options?.moderation ?? "auto",
+    output_compression: options?.outputCompression ?? 100,
+    output_format: options?.outputFormat ?? "png",
+    partial_images: options?.partialImages ?? 0,
+    quality: options?.quality ?? "auto",
+    size: options?.size ?? "auto",
```
hntrl: do we want to have defaults for this?

libs/providers/langchain-openai/src/converters/responses.ts
hntrl, line 1: whats with the satisfies calls? and the assertions for annotations?

.changeset/shiny-roses-attend.md
hntrl, line 1: probably a patch no?

libs/providers/langchain-openai/src/converters/responses.ts
christian-bromann: This makes the types more strict and actually discovered a bug here. `part.annotations` are OpenAI Annotations which are assigned to LangChain annotations. Both annotation types don't match. I will push a change to add a `convertOpenAIAnnotationToLangChain`.

The origin of this issue is that that `ContentBlock.Text.annotations?: (BaseContentBlock | ContentBlock.Citation)[]`. The `BaseContentBlock` is very loosely typed which caused all OpenAI citations to match.

libs/providers/langchain-openai/src/tools/imageGeneration.ts
christian-bromann: The alternative is having OpenAI dictate the default which seems more dangerous to me then just setting them implicitely.

PR #9686, opened 2025-12-18
feat(core): usage_metadata in extra.metadata

hntrl:


libs/langchain-core/src/tracers/tracer_langchain.ts
```diff
@@ -43,6 +47,32 @@ export interface LangChainTracerFields extends BaseCallbackHandlerInput {
...
+  for (const generationBatch of generations) {
+    for (const generation of generationBatch) {
+      if ("message" in generation) {
+        const { message } = generation as { message: unknown };
+        if (
+          isBaseMessage(message) &&
+          isAIMessage(message) &&
+          message.usage_metadata !== undefined
+        ) {
+          return message.usage_metadata;
```
jacoblee93: Should we only do this in the same scenario as when we flatten?

Not that anyone should have multiple generations anymore but would be unexpected if we just took the first value

Proper thing would actually be to sum them I think?

hntrl: Yeah I like this more, will adjust. Going to avoid aggregating usage metadata since no one has multiple generations anymore and we have a bunch of different forms of usage metadata that I don't care to solve right now 😁

hntrl: actually nvm I remember we have `mergeUsageMetadata`

hntrl:
after: https://smith.langchain.com/public/31b28814-bd98-43ee-a3b2-38a448e7d1cb/r
before: https://smith.langchain.com/public/a8c6f3f6-0cb7-4327-b326-a44669fb9faa/r

PR #9688, opened 2025-12-18
chore(core): flatten generations for LangChainTracer

hntrl:


libs/langchain-core/src/tracers/tracer_langchain.ts
```diff
@@ -132,6 +132,14 @@ export class LangChainTracer
         metadata.usage_metadata = usageMetadata;
         run.extra.metadata = metadata;
       }
+
+      // Flatten outputs if there's only a single generation
+      if (
+        outputs.generations.length === 1 &&
+        outputs.generations[0].length === 1
+      ) {
+        run.outputs = outputs.generations[0][0];
```
jacoblee93, line 141: Nice that's all we need to do?

PR #9701, opened 2025-12-19
fix(langchain-aws): add support for Amazon Nova embedding models

### LINKED ISSUE #9694: BedrockEmbeddings with Amazon Nova Multimodal Embeddings v1 returns ValidationException "Malformed input request: required key [messages] not found"
### Checked other resources

### Example Code

Creating this ticket to fix the LangchainJS the way the Python version was fixed [here](https://github.com/langchain-ai/langchain-aws/issues/744).

We get the ValidationException "Malformed input request: required key [messages] not found"

### Description

Creating this ticket to fix the LangchainJS the way the Python version was fixed [here](https://github.com/langchain-ai/langchain-aws/issues/744).

The new Nova embedding model needs to be wired up [here](https://github.com/langchain-ai/langchainjs/blob/main/libs/providers/langchain-aws/src/embeddings.ts)

### System Info

MacOS
Node v20.19.4

christian-bromann:
This PR adds support for Amazon Nova embedding models (e.g., `amazon.nova-embed-text-v1`) in the `BedrockEmbeddings` class. Nova models require a different request payload format than the existing Titan models.

fixes #9694

### Problem

When using Nova embedding models with `BedrockEmbeddings`, users encounter the following error:

```
ValidationException: Malformed input request: required key [messages] not found
```

This occurs because Nova models expect a `messages` array format while the existing implementation only supports the `inputText` format used by Titan models.

### Solution

- Added detection for Nova embedding models via `isNovaEmbeddingModel()` helper function
- Updated `_embedText()` to dynamically select the correct request format:
  - **Nova models**: `{ messages: [{ role: "user", content: [{ text: "..." }] }] }`
  - **Titan models**: `{ inputText: "..." }`

### Related

- Mirrors the fix applied to the Python version: https://github.com/langchain-ai/langchain-aws/issues/744

### DIFF
--- a/.changeset/rich-students-float.md
+++ b/.changeset/rich-students-float.md
@@ -1,0 +1,5 @@
+---
+"@langchain/aws": patch
+---
+
+fix(langchain-aws): add support for Amazon Nova embedding models


libs/providers/langchain-aws/src/embeddings.ts
```diff
@@ -6,6 +6,16 @@ import {
...
 import { CredentialType } from "./types.js";
 
+/**
+ * Checks if the given model is an Amazon Nova embedding model.
+ * Nova models require a different request format using `messages` instead of `inputText`.
+ * @param model - The model ID string
+ * @returns true if the model is a Nova embedding model
+ */
+function isNovaEmbeddingModel(model: string): boolean {
+  return model.toLowerCase().includes("nova-embed");
```
kamilcglr, line 16: Hello, it seems this doesn’t work with the Nova 2 embedding model `amazon.nova-2-multimodal-embeddings-v1:0`.

If I’m understanding correctly, this model ID wouldn’t match the current TypeScript helper, which checks specifically for `"nova-embed"`:

```ts
function isNovaEmbeddingModel(model: string): boolean {
  return model.toLowerCase().includes("nova-embed");
}
```

In the Python implementation, the check is done more generically on `"nova"` and `"embed"` in the model ID:

```python
self._inferred_provider == "amazon" \
    and "nova" in self.model_id \
    and "embed" in self.model_id
```

I think a similar, more generic check here would also catch Nova 2 models, but please correct me if I’m missing something.

christian-bromann: You are right, great catch! Would you be open to raise a PR?

PR #9707, opened 2025-12-22
fix!(core/langchain): hardening for `load`

hntrl:
* changes default for `secretsFromEnv`
* wraps serialization of langchain-like objects
* clarifies security surface of `load`
* adds `maxDepth` parameter

### DIFF
--- a/libs/langchain-core/src/load/serializable.ts
+++ b/libs/langchain-core/src/load/serializable.ts
@@ -236,32 +236,24 @@ export abstract class Serializable implements SerializableInterface {
       }
     });
 
-    // Process kwargs with secret replacement and key mapping
-    const processedKwargs = mapKeys(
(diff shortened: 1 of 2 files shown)


dejoma:
Why did you change the default my brother!! 

```ts
  const chain = await hub.pull<
    Runnable<PromptConfig[K]['input'], PromptConfig[K]['output']>
  >(promptKey, {
    includeModel: true,
    modelClass: ChatAnthropic // fails because API key is not set from env anymore
  });
```

The wrapper isn't updated now

hntrl:
@dejoma Apologies for this! We saw this as a necessary break for a critical security vulnerability related to our serialization/de-serialization mechanics: https://github.com/langchain-ai/langchainjs/security/advisories/GHSA-r399-636x-v7f6

PR #9715, opened 2025-12-25
fix(langchain,classic): Add secretsFromEnv and secrets for prompt pulling

jacoblee93:
Breaking change necessary due to recent security changes around deserialization no longer loading secrets by default (see https://github.com/langchain-ai/langsmith-sdk/pull/2239)

### Before
langchain < 1.2.3 and (1.1.8 < langchain-core)

```ts
  process.env["OPENAI_API_KEY"] = "sk-..."

  // Secrets were automatically read from environment variables
  const prompt = await pull(
    "my-prompt-with-model", {
      includeModel: true
   });
```

### After 

**Option 1: Pass secrets explicitly (Recommended)**
```ts
  // ✅ Recommended: Pass secrets directly
  const prompt = await pull(
    "my-prompt-with-model", {
      includeModel: true,
      secretsMap: {"OPENAI_API_KEY": "sk-..."}
  });
```

**Option 2: Enable reading from environment (Only for trusted prompts)**
```ts
  process.env["OPENAI_API_KEY"] = "sk-..."

  // ⚠️ Only use for trusted prompts
  const prompt = await pull(
    "my-prompt-with-model", {
      includeModel: true,
      secretsFromEnv: true
  });
```

**Option 3: No secrets needed (No change required)**
```ts
  // No change needed if includeModel: false (default)
  const prompt = await pull("my-prompt");
```

## Migration Guide

- **If you pull prompts without models** (`includeModel: False` or not specified): No changes required
- **If you pull trusted prompts with models**: Add `secretsFromEnv: True` to restore previous behavior
- **If you pull untrusted prompts with models**: Use `secrets: {...}` to pass only the required secrets explicitly

### DIFF
--- a/libs/langchain/src/hub/node.ts
+++ b/libs/langchain/src/hub/node.ts
@@ -41,10 +41,6 @@ function isRunnableBinding(a: string[]): boolean {
  * @param options.includeModel Whether to also instantiate and attach a model instance to the prompt,
  *   if the prompt has associated model metadata. If set to true, invoking the resulting pulled prompt will
  *   also invoke the instantiated model. You must have the appropriate LangChain integration package installed.
- * @param options.modelClass If includeModel is true, the class of the model to instantiate. Required
- *   for non-OpenAI models. If you are running in Node or another environment that supports dynamic imports,
- *   you may instead import this function from "langchain/hub/node" and pass "includeModel: true" instead
- *   of specifying this parameter.
  * @param options.secrets A map of secrets to use when loading, e.g.
  *   {'OPENAI_API_KEY': 'sk-...'}`.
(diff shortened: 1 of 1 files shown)


libs/langchain/src/chat_models/universal.ts
```diff
@@ -161,10 +161,14 @@ export async function getChatModelByClassName(className: string) {
     if (
       "code" in err &&
       err.code?.toString().includes("ERR_MODULE_NOT_FOUND") &&
-      "message" in err
+      "message" in err &&
+      typeof err.message === "string"
```
jacoblee93, line 165: Ran into this while running integration tests

hntrl (approved):

PR #9726, opened 2026-01-02
fix(deepseek): parse <think> tags in content stream

murataslan1:
## Problem
DeepSeek R1 and compatible models sometimes embed reasoning content within `<think>...</think>` tags in the main content stream, instead of using the `reasoning_content` field. This causes raw tags to appear in the output.

## Solution
Modified `ChatDeepSeek` to override `_streamResponseChunks`. It now implements a stateful parser that:
1. Detects `<think>` tags in the stream.
2. Extracts content between tags and yields it as `additional_kwargs.reasoning_content`.
3. Strips the tags from the main `content` output.

## Tests
Added `libs/providers/langchain-deepseek/src/tests/chat_models_reasoning.test.ts` which simulates an API response containing `<think>` tags and verifies they are correctly parsed.

### DIFF
--- a/libs/providers/langchain-deepseek/src/tests/chat_models_reasoning.test.ts
+++ b/libs/providers/langchain-deepseek/src/tests/chat_models_reasoning.test.ts
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
 import { test, expect } from "vitest";
 import { ChatDeepSeek } from "../chat_models.js";
 import { AIMessageChunk } from "@langchain/core/messages";


christian-bromann:
@murataslan1 thanks for the PR, there are still some copilot comments that I think are worth to address. Let me know if I can help in any way.

murataslan1:
@christian-bromann The edge case tests are already added in commit 7d6c9c1:

- ✅ Empty think blocks
- ✅ Nested think tags  
- ✅ Malformed tags
- ✅ Multiple blocks with content before/after
- ✅ Unclosed tags (flush at end)
- ✅ Split tags across chunks

All 7 tests passing. Let me know if there's a specific case I missed!

christian-bromann (approved):
LGTM 👍

PR #9744, opened 2026-01-06
chore(internal): Turborepo improvements

christian-bromann:
This PR overhauls the Turborepo configuration to achieve significantly faster cached builds and reduce maintenance overhead by consolidating redundant configuration files.

### Performance Results

| Metric | Before | After | Improvement |
|--------|--------|-------|-------------|
| Cached build time | 44.5s | **1.96s** | **22.7x faster** |
| Cache hit rate | 82% (31/38) | **100% (38/38)** | +18% |
| Config files to maintain | 44 | **6** | 86% reduction |

#### Root `turbo.json` Improvements

- **Added `inputs` configuration** for all tasks - enables precise cache invalidation based on actual source file changes
- **Added `globalPassThroughEnv`** - properly handles `CI`, `NODE_ENV`, and `GITHUB_ACTIONS` environment variables
- **Added `pnpm-lock.yaml` as global dependency** - ensures dependency updates trigger rebuilds
- **Removed unnecessary transitive dependencies** - `lint` and `format` tasks no longer wait for `^lint`/`^format`, enabling parallel execution
- **Simplified test task dependencies** - tests now depend directly on `build:compile` instead of complex chains
- **Enabled test caching** - unit tests are now cacheable with proper input definitions

#### Consolidated Package Configurations

**Deleted 38 redundant package-level `turbo.json` files** that were:
- Incorrectly defining a `build` task instead of `build:compile`
- Duplicating root configuration without adding unique value

**Retained 5 package-level configs** that have legitimate overrides:
- `@langchain/core`, `langchain`, `@langchain/community`, `@langchain/classic` - exclude auto-generated files from inputs
- `create-langchain-integration/template` - template for new integrations

### Why This Matters

The previous configuration had several issues causing poor cache performance:

1. **Missing `inputs`** - Turbo couldn't determine which files affected each task, leading to conservative cache invalidation
2. **Task name mismatch** - Package configs defined `build` but root used `build:compile`, causing cache keys to diverge
3. **Overly sequential dependencies** - Tasks like `lint` waited for `^lint` (all dependency packages to lint first) when they could run in parallel

With proper `inputs` defined, Turbo can now:
- Hash only relevant source files per task
- Skip rebuilds when unrelated files change
- Achieve 100% cache hit rate on unchanged packages


hntrl (approved):
nice!

PR #9752, opened 2026-01-06
fix(environment_tests): enhance cloudflare integration tests

christian-bromann:
This PR adds comprehensive Cloudflare Workers compatibility tests for the `langchain` v1 package. During testing, we discovered that **LangChain v1 requires the `nodejs_compat` compatibility flag** to work in Cloudflare Workers.

### Key Finding: `nodejs_compat` Required

LangChain v1 cannot run in Cloudflare Workers without the `nodejs_compat` flag because:

| Module | Source | Used By |
|--------|--------|---------|
| `node:async_hooks` | `@langchain/langgraph` | Async context tracking |
| `node:fs/promises` | `langchain/storage/file_system` | File system storage |
| `node:path` | `langchain/storage/file_system` | Path utilities |

Users must add this to their `wrangler.toml`:

```toml
compatibility_date = "2024-09-23"
compatibility_flags = ["nodejs_compat"]
```

### Why This Wasn't Detected Before

The existing Cloudflare Workers tests only covered `@langchain/classic` and `@langchain/core`:

1. **Test coverage gap:** The `langchain` v1 package was listed as a dependency in the test package, but the worker entry point (`index.ts`) only imported from `@langchain/openai`, `@langchain/core`, and `@langchain/classic` — not from the main `langchain` package.

2. **New dependency chain:** LangChain v1 introduces `@langchain/langgraph` as a core dependency, which uses `node:async_hooks` for async context management. This dependency didn't exist in the classic package.

3. **entrypoints.js only tested classic:** The existing `entrypoints.js` file exclusively tested `@langchain/classic` exports, never importing from the v1 `langchain` package.

4. **No functional v1 tests:** There were no tests that actually instantiated v1 constructs like `tool()`, `createAgent`, or v1 message types in the CF Workers environment.

### DIFF
--- a/environment_tests/test-exports-cf/src/index.core.unit.test.ts
+++ b/environment_tests/test-exports-cf/src/index.core.unit.test.ts
@@ -1,0 +1,55 @@
+import { unstable_dev } from "wrangler";
+import type { UnstableDevWorker } from "wrangler";
+import { describe, expect, it, beforeAll, afterAll } from "vitest";
+
+describe("@langchain/core Worker (no nodejs_compat)", () => {
+  let worker: UnstableDevWorker;
+
+  beforeAll(async () => {
+    worker = await unstable_dev("src/index.core.ts", {
+      experimental: { disableExperimentalWarning: true },
+      config: "wrangler.core.toml",
+    });
+  }, 30000);
+
+  afterAll(async () => {
+    await worker.stop();
+  });
+
+  it("should pass all core unit tests", async () => {
+    const resp = await worker.fetch("/test");
+    expect(resp.status).toBe(200);
+
+    const data = await resp.json();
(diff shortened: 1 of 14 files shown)


environment_tests/test-exports-cf/src/index.core.unit.test.ts
hntrl, line 1: I think if we're importing incompatible modules into workers it'll fail at `wrangler build`. Think we should test against that since it actually kicks up a workerd task

libs/langchain-core/src/callbacks/dispatch/index.ts
hntrl, line 1: Think I'd be more comfortable with just telling people to use `nodejs_compat` flag. This was the way it was in their docs for the longest time, but I think they replaced us with the python version when they rolled out python support ☹️

PR #9758, opened 2026-01-07
feat(ollama): add think parameter support to Ollama LLM class

### LINKED ISSUE #8481: NITt: Add think Field to ChatOllama for Controlling Thought Process in Responses
### Checked other resources

### Example Code

The following code:

```typescript
```

_No response_

### Description

* I’m trying to use the langchain library and its Ollama integration to stream responses from Ollama with the think: false option so I can hide the model’s chain-of-thought (deepseek-r1:32b).
* I expect not to see the chain-of-thought wrapped in <think> … </think> tags inside the content field (the same format Ollama returns when think is false).
* Instead, when think is false the <think> wrapper still appears. 

### System Info

"yarn info langchain"

langchain:  0.3.29
langchain-core:  0.3.62

platform:  macOS 15.5 (Apple Silicon)
Node version:  v20.9.0
yarn version: 1.22.22

Gulianrdgd:
## Description

  Adds support for the `think` parameter to the `Ollama` LLM class, matching the existing implementation in `ChatOllama`. This enables reasoning/thinking mode for models like DeepSeek R1 when using the LLM interface.

  ## Changes

  - Added `think?: boolean` parameter to `OllamaInput` interface and `Ollama` class
  - Updated `invocationParams()` to pass the `think` flag to the Ollama API
  - Modified `_streamResponseChunks()` to prioritize `chunk.thinking` over `chunk.response` when think mode is enabled
  - Added integration tests (`llms_think.int.test.ts`) to verify thinking behavior

  ## Related Issues

Fixes #8481 but then not only for ChatOllama

  ## Testing
  - Added integration test that compares response lengths with `think=true`, `think=false`, and `think=undefined`
  - Test verifies that thinking mode produces significantly longer responses (reasoning + answer)
  - All existing tests pass

  ## Implementation Details

  The implementation follows the same pattern as `ChatOllama`:
  - When `think: true`, the model's thinking/reasoning process is included in the response
  - When `think: false`, only the final answer is returned
  - The `GenerateResponse` interface from the Ollama SDK includes a `thinking?: string` field that we utilize in streaming


christian-bromann:
@Gulianrdgd mind running the formatter?

Gulianrdgd:
> @Gulianrdgd mind running the formatter?

Oops, sorry. I pushed the formatted code now.

PR #9786, opened 2026-01-12
fix: resolve flaky tests and configuration issues

SkrOYC:
## Summary
This PR fixes 3 issues that were causing the test suite to fail or be flaky in certain environments (e.g. CI, devenv):

1. **@langchain/turbopuffer**: The test script now uses `vitest run --passWithNoTests`. This package only has integration tests, so running unit tests would previously fail with "No test files found".
2. **@langchain/model-profiles**: Fixed a broken import in `generator.test.ts`. It was importing `findMonorepoRoot` from `../path-utils.js` (which doesn't exist) instead of `../config.js`.
3. **@langchain/classic**: Fixed `AutoGPTPrompt` tests to be locale-independent. The test was failing on systems where the default locale didn't match the expected format (e.g. `p. m.` vs `PM`). I updated the code to explicitly use `en-US` locale for date formatting in the prompt.

## Test Plan
Ran `pnpm test:unit` locally.
- `turbopuffer`: Passes (skipped)
- `model-profiles`: Passes
- `classic`: Passes
- Full suite: 71 successful, 71 total

## Changeset
Included.


hntrl (approved):
Thanks for the fix!

PR #9788, opened 2026-01-12
fix(google-genai): handle tool_call content blocks in v1 output

SkrOYC:
## Summary
This PR fixes a runtime error in the Google GenAI provider when using `outputVersion: "v1"`.

### The Problem
When `outputVersion: "v1"` is enabled, LangChain's `AIMessage` constructor converts tool calls into `ContentBlock` objects of type `tool_call`. However, the Google GenAI converter only handled `tool_use` blocks, leading to an `Unknown content type tool_call` error when sending messages with tool calls back to the model (e.g., in multi-turn conversations).

### The Fix
Added a handler for `tool_call` blocks in `_convertLangChainContentToPart` that correctly maps `block.args` to the expected `functionCall.args` format.

## Test Plan
Added a new regression test in `libs/providers/langchain-google-genai/src/tests/tool_call_conversion.test.ts` that verifies:
1. `AIMessage` correctly creates `tool_call` blocks with `outputVersion: "v1"`.
2. `convertBaseMessagesToContent` successfully converts these blocks into Google's `functionCall` format.

## Changeset
Included.

### DIFF
--- a/libs/providers/langchain-google-genai/src/tests/tool_call_conversion.test.ts
+++ b/libs/providers/langchain-google-genai/src/tests/tool_call_conversion.test.ts
@@ -1,4 +1,3 @@
-
 import { test, expect } from "@jest/globals";
 import { AIMessage } from "@langchain/core/messages";
 import { convertBaseMessagesToContent } from "../utils/common.js";


libs/providers/langchain-google-genai/src/tests/tool_call_conversion.test.ts
```diff
@@ -0,0 +1,49 @@
...
+      {
+        id: "call_123",
+        name: "search_tool",
+        args: { query: "AppSheet documentation" },
+      },
+    ],
+    response_metadata: {
+      output_version: "v1",
+    },
+  });
```
hntrl, line 20: I don't believe this is accurately testing against standard content. I'd expect the appropriate test case is something like: 
```suggestion
  const aiMessage = new AIMessage({
    contentBlocks: [{
      id: "call_123",
      type: "tool_call",
      name: "search_tool",
      args: { query: "AppSheet documentation" },
    }],
  });
```

SkrOYC: You are right, working on a more representative test

SkrOYC:
@hntrl I just changed the test with the right signature and reused the openai Calculator example as a more generally-common tool

SkrOYC:
Hey @hntrl !
When do you think this could be fixed in the package?
I'm sorry if I'm sounding pushy but I need it for my SaaS 😅

hntrl (approved):
Apologies for the delay, thanks for the PR!

SkrOYC:
Thanks! I'll be waiting for the updated package to start using it!

PR #9793, opened 2026-01-13
(ChatOllama) Set additional_kwargs.reasoning_content when streamEvents (on_chat_model_stream) via createAgent

bao-tran-iohub:
## Problem
Reasoning content is available under `additional_kwargs` of AIMessageChunk when streaming directly from ChatOllama. However,  when streaming from inside ReactAgent with `streamEvents` , for `on_chat_model_stream` the reasoning content is still part of the main content.

## Research
Tracing the code, it seems if the generation chunk is not passed into `onLLMNewToken` callback then only the main content is streamed, not the complete chunk message.

Comparing `*_streamResponseChunks` of ChatOllama with ChatOpenAIResponses and ChatAnthropicMessages confirms that we need to pass the chunk to the `runManager.handleLLMNewToken` callback.

## Solution
Pass the new generation chunk to the `runManager.handleLLMNewToken` callback as input.

## Test
I tried implementing the ChatOllama class locally with the proposed change and the reasoning content is correctly parsed seperatedly from main content.

### DIFF
--- a/libs/providers/langchain-ollama/src/chat_models.ts
+++ b/libs/providers/langchain-ollama/src/chat_models.ts
@@ -748,11 +748,11 @@ export class ChatOllama
 
     let lastMetadata: Omit<OllamaChatResponse, "message"> | undefined;
 
-    for await (const chunk of stream) {
+    for await (const streamChunk of stream) {
       if (options.signal?.aborted) {
         this.client.abort();
       }
-      const { message: responseMessage, ...rest } = chunk;
+      const { message: responseMessage, ...rest } = streamChunk;
       usageMetadata.input_tokens += rest.prompt_eval_count ?? 0;
       usageMetadata.output_tokens += rest.eval_count ?? 0;
       usageMetadata.total_tokens =


bao-tran-iohub:
Hi @christian-bromann, 

I'm blocked by dependency issue, which prevents the import of createAgent from langchain into the test file. I did see ChatGroq having the same issue, so maybe this test of Agent streaming can be skipped, unless you have some other ways.

libs/providers/langchain-ollama/src/chat_models.ts
```diff
@@ -764,11 +764,19 @@ export class ChatOllama
         ? (responseMessage.thinking ?? responseMessage.content ?? "")
         : (responseMessage.content ?? "");
 
-      yield new ChatGenerationChunk({
+      const cg = new ChatGenerationChunk({
```
christian-bromann:
```suggestion
      const chunk = new ChatGenerationChunk({
```

libs/providers/langchain-ollama/src/chat_models.ts
```diff
@@ -764,11 +764,19 @@ export class ChatOllama
...
         text: token,
         message: convertOllamaMessagesToLangChain(responseMessage),
       });
-      await runManager?.handleLLMNewToken(token);
+      yield cg;
```
christian-bromann:
```suggestion
      yield chunk;
```

libs/providers/langchain-ollama/src/chat_models.ts
```diff
@@ -764,11 +764,19 @@ export class ChatOllama
...
+      await runManager?.handleLLMNewToken(
+        token,
+        undefined,
+        undefined,
+        undefined,
+        undefined,
+        { chunk: cg }
```
christian-bromann:
```suggestion
        { chunk }
```

PR #9812, opened 2026-01-15
fix(langchain): tag summarization model calls with lc_source metadata

christian-bromann:
Mirror python PR langchain-ai/langchain#34763 in JS: tag the summarization model invocation with `metadata.lc_source="summarization"` while inheriting the parent runnable/LangGraph config.

This enables downstream handlers (e.g. streaming/trace attribution) to correctly identify and tag the internal summarization call.


mdrxy:
FYI I made some minor corrections in python after this was opened/initially ported

PR #9827, opened 2026-01-20
refactor: optimize stream chunk aggregation and remove redundant sorting 🚀

sanjaiyan-dev:
- This PR optimizes the logic used to aggregate `ChatGenerationChunk` sequences from the stream. It replaces the `Record` based (Object) storage with a native `Array` and removes the manual sorting 
> This refactor improves performance by replacing the object-based storage and redundant $O(N  log N)$ manual sorting with a native array, achieving $O(N)$ linear efficiency. Since JavaScript engines automatically maintain the order of numeric indices, the previous sorting step was unnecessary; removing it eliminates significant CPU overhead and results in a cleaner, faster, and more memory-efficient implementation.

- This pull request utilizes `Promise.all` to run operations concurrently, resulting in reduced latency.

### DIFF
--- a/libs/providers/langchain-google-genai/src/chat_models.ts
+++ b/libs/providers/langchain-google-genai/src/chat_models.ts
@@ -871,7 +871,9 @@ export class ChatGoogleGenerativeAI
           finalChunks[index] = finalChunks[index].concat(chunk);
         }
       }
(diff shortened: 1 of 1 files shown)


hntrl (approved):
Thanks for the PR!

PR #9832, opened 2026-01-20
fix(jira-document-loader): Fix Jira Document Loader and tests to reflect updates in Jira API v3

phong-phuong:
**Fix: Update Jira loader and tests for Jira API v3 description format (ADF)**

**Summary**

Jira Cloud REST API v3 no longer returns issue.fields.description as a plain string. Instead, it is returned as an Atlassian Document Format (ADF) JSON structure. This PR updates the Jira document loader and its unit tests to correctly model and handle this behaviour.

**What changed**

- Updated Jira issue test fixtures to emit valid ADF for fields.description
- Fixed unit tests to assert against the textual rendering of the ADF description rather than the raw object
- Aligned type usage and expectations with Jira Cloud REST API v3 responses
- Ensured the existing description formatter continues to convert ADF to text correctly
- Add descriptionFormatter to allow user to customise the description formatting if the default string formatter is suitable for their needs (obeys open/closed principle)
- provides two description formatters, one for string (friendly for LLMs), one for JSON (preserves the original description object).
- fixed infinite loop in the fetchIssues function, now correctly handles pagination for the v3 JQL endpoint.

**Why this change is needed**

- Jira REST API v3 returns rich-text fields (including description) as structured ADF, not strings
- Existing tests relied on an outdated v2-style string description, causing false positives and incorrect assumptions
- This change ensures the loader and tests accurately reflect real Jira Cloud data and avoids silent data loss
- Without the change, the issue description when converted to string ends up as [object, object]
- Infinite loop when trying to fetch issues spanning more than one request

**Impact**

- No breaking changes to public APIs
- Improves correctness and forward compatibility with Jira Cloud
- Makes test fixtures representative of real-world Jira responses

**References**

Jira Cloud REST API v3 documentation
https://developer.atlassian.com/cloud/jira/platform/rest/v3/intro/

Atlassian Document Format (ADF) specification
https://developer.atlassian.com/cloud/jira/platform/apis/document/structure/


phong-phuong:
> Do you think we can add some unit tests for these changes or at least start creating some unit tests?

Added the unit tests:
- filterFn
- issueDescriptionFormatter
- pagination
- increased integration test timeout to 30 seconds for large jira projects

All jira unit and integration tests now pass. Please verify.

phong-phuong:
Awesome, too easy, I like that!

PR #9834, opened 2026-01-20
fix(core): throw error when bindTools and withStructuredOutput used together

### LINKED ISSUE #7757: Throw an error when both bindTools and withStructuredOutput are used, getting tools attached to an agent
### Checked other resources

- [x] I searched the LangChain documentation with the integrated search.
- [x] I am sure that this is a bug in LangChain rather than my code.

### Example Code

model.bindTools(tools);
model.withStructuredOutput(schema);
model.invoke([...]) // does not throw Error

_No response_

### Description

I was using the gemini model, and per documentation was using tools, and then was trying to use withStructuredOutput as well.

Things didnt work as expected (and looking at the traces I saw no functions being passed to upstream calls. After much debugging landed on the documentation that called out we can't use both of these together. Also this [discussion](https://github.com/langchain-ai/langchain/discussions/26573).

It would have been helpful if:
1) There was an error thrown on calling invoke with both bindTools and withStructuredOutput are used on the chat node. If the goal is to allow this to be used for some usecases, then a flag could be passed to invoke to allow.
2) There seems no way to see what tools are bound to a chat node after calling bindTools. I cannot do chatllm.tools or chatllm.getTools. Documentation also shows nothing to get tools attached to a chat node! This caused frustration when trying to debug the above issue. Can we expose such a method or attribute on a chat model so we can see what are the current tools attached to a model node.

### System Info

I'm on:

@langchain/core": "^0.3.39"
@langchain/google-genai": "^0.1.8"

felmonon:
## Summary

Addresses issue #7757 - adds validation to `withStructuredOutput()` to detect when tools have already been bound via `bindTools()` and throws a clear error message.

### Changes:
- Add optional `getBoundTools()` method to `BaseChatModel` for subclasses to implement, allowing inspection of bound tools
- Add validation in `withStructuredOutput()` that checks `getBoundTools()` and throws an informative error if tools are already bound
- Add `getBoundTools()` implementation to `FakeStreamingChatModel` for testing
- Add test cases verifying the new behavior

This prevents silent failures when users accidentally chain `bindTools()` and `withStructuredOutput()`, which are mutually exclusive since `withStructuredOutput` internally uses tool calling.

Fixes #7757

## Test plan

- [x] Added test case for error when tools already bound
- [x] Added test case verifying `getBoundTools()` returns empty array when no tools bound
- [x] Added test case verifying normal operation when no tools bound
- [x] All 21 existing tests pass

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/.changeset/fix-bindtools-structured-output.md
+++ b/.changeset/fix-bindtools-structured-output.md
@@ -1,0 +1,5 @@
+---
+"@langchain/core": patch
+---
+
+Add validation to throw an error when `withStructuredOutput` is called on a model that already has tools bound via `bindTools`. These methods are mutually exclusive because `withStructuredOutput` internally uses tool calling to generate structured output. Added optional `getBoundTools()` method to `BaseChatModel` that subclasses can implement to enable this validation.


christian-bromann:
First off, thanks for the contribution!

I don't think this solves the underlying issue. It is currently already not possible to use them together, e.g.

Screenshot 2026-02-05 at 12 05 10 PM

You will get an error `TypeError: bound.withStructuredOutput is not a function`.

felmonon:
Thanks for the review @christian-bromann!

You're right that chaining them (`model.bindTools(tools).withStructuredOutput(schema)`) already throws `TypeError: bound.withStructuredOutput is not a function`, so that case is covered.

I went back and re-read the original issue #7757 though, and the pattern the reporter was hitting is actually different:

```typescript
model.bindTools(tools);            // return value never saved
model.withStructuredOutput(schema); // same here
await model.invoke([...]);          // this runs the original model with nothing attached
```

Both methods return new objects instead of mutating the model, so if you don't capture the return value, `model.invoke()` just runs the bare model. No error, just silently wrong behavior. That's what tripped up the original reporter.

Looking at it again, my fix doesn't catch that pattern either since the original model never actually gets modified. It's really just a footgun from the immutable API design and not something we can detect at runtime.

Would you be open to a smaller PR that gives a better error message for the chaining case? Something more helpful than the generic `TypeError: not a function`. Or if you'd rather handle this through docs, totally fine too.

Either way, appreciate the clear feedback.

christian-bromann:
> Something more helpful than the generic `TypeError: not a function`

I think it can't be more clear than that, if you bindTools, you can't call `withStructuredOutput` on it. A more interesting problem to solve is how we can better enable it?

PR #9839, opened 2026-01-21
feat(langchain): add state schema support to agents

maahir30:
Adds full support for LangGraph's `StateSchema` API in `createAgent` and `middleware`, enabling users to define agent state using the modern schema-compliant validation approach with `ReducedValue` for custom reducers and `UntrackedValue` for transient fields.

### Type Definitions (`libs/langchain/src/agents/types.ts`)

Updated typing to accept StateSchema along with Zod objects.

This completes the type chain that was already prepared in:
- `AgentTypeConfig` interface
- `CreateAgentParams` type
- All `createAgent` function overloads
- `ToAnnotationRoot` and `InferSchemaInput` utilities
- `createAgentAnnotationConditional` runtime handling

### Tests

Added comprehensive tests verifying:
- StateSchema instances are accepted as `stateSchema`
- `ReducedValue` works correctly for custom merge logic
- Type inference works end-to-end

## Usage Example

```typescript
import { StateSchema, ReducedValue } from "@langchain/langgraph";
import { z } from "zod";

const AgentState = new StateSchema({
  userId: z.string(),
  count: z.number().default(0),
  history: new ReducedValue(
    z.array(z.string()).default(() => []),
    { 
      inputSchema: z.string(), 
      reducer: (current, next) => [...current, next] 
    }
  ),
});

const agent = createAgent({
  model: "openai:gpt-4o",
  tools: [searchTool],
  stateSchema: AgentState,
});

  messages: [{ role: "user", content: "Hello" }],
  userId: "user123",
  count: 5,
});
// result.userId, result.count, result.history are fully typed
```

### DIFF
--- a/.changeset/silly-impalas-kiss.md
+++ b/.changeset/silly-impalas-kiss.md
@@ -1,0 +1,15 @@
+---
+"langchain": major
+---
+
+## Add state schema support to agents 
+- Introduce `createAgentState` function replacing `createAgentAnnotationConditional`
+- Migrate from Zod-based schemas to LangGraph's native `StateSchema`, `ReducedValue`, and `UntrackedValue` primitives
+- Support both `StateSchema` and Zod v3/v4 objects as input schemas
+- Automatically merge state fields from user-provided schemas and middleware
+- Properly handle reducer metadata extraction from Zod v4 schemas via `schemaMetaRegistry`
+- Generate separate input/output schemas to avoid channel conflicts with reducers
+- Add comprehensive test coverage for state schema functionality
+
+
+


libs/langchain/src/agents/tests/state_schema.test.ts
```diff
@@ -0,0 +1,68 @@
...
+      stateSchema: AgentState,
+    });
+
+    expect(agent).toBeDefined();
+    expect(agent.options.stateSchema).toBe(AgentState);
+  });
+
+  test("should work with ReducedValue", () => {
+    const AgentState = new StateSchema({
+      history: new ReducedValue(
```
hntrl: think we should also have some kind of smoke test that this reduced value will actually get called by an agent. Can probably do that by forcing a tool call with a tool that returns a `Command`

hntrl: or through middleware states

hntrl: (should prob test both)

libs/langchain/src/agents/annotation.ts
```diff
@@ -39,7 +41,23 @@ export function createAgentAnnotationConditional<
   // Separate shape for input/output without reducer metadata (to avoid channel conflicts)
   const ioSchemaShape: Record<string, any> = {};
 
-  const applySchema = (schema: InteropZodObject) => {
+  const isStateSchema = (schema: unknown): schema is StateSchema<any> =>
+    StateSchema.isInstance(schema);
+
+  const applySchema = (schema: InteropZodObject | StateSchema<any>) => {
+    // Handle StateSchema
+    if (isStateSchema(schema)) {
```
hntrl: nit
```suggestion
  const applySchema = (schema: InteropZodObject | StateSchema<any>) => {
    // Handle StateSchema
    if (StateSchema.isInstance(schema)) {
```

libs/langchain/src/agents/types.ts
```diff
@@ -84,9 +84,14 @@ export interface AgentTypeConfig<
...
     | ResponseFormatUndefined,
-  TState extends AnyAnnotationRoot | InteropZodObject | undefined =
+  TState extends
     | AnyAnnotationRoot
     | InteropZodObject
+    | StateSchema<any>
+    | undefined =
+    | AnyAnnotationRoot
+    | InteropZodObject
+    | StateSchema<any>
```
hntrl: Can we collapse these into one type? I think you can import `StateDefinitionInit` as a type from langgraph

libs/langchain/src/agents/annotation.ts
```diff
@@ -39,7 +41,23 @@ export function createAgentAnnotationConditional<
...
+      // Extract field keys from StateSchema.fields
+      for (const key of Object.keys(schema.fields)) {
+        if (key.startsWith("_")) continue;
+        if (!(key in schemaShape)) {
+          schemaShape[key] = z.any();
+          ioSchemaShape[key] = z.any();
```
hntrl: I think as apart of this we want to refactor the state handling to not be so dependent on zod. There's things like this that we do that sort of just washes validations against state keys which is problematic. This might be optimistic, but we shouldn't need to import zod anywhere inside of `langchain/agents` if we're using StateSchema

libs/langchain/src/agents/middleware/types.ts
```diff
@@ -731,15 +743,22 @@ export type InferContextInput<
     ? ToAnnotationRoot<ContextSchema>["State"]
     : {};
 
-export type ToAnnotationRoot<A extends AnyAnnotationRoot | InteropZodObject> =
-  A extends AnyAnnotationRoot
+export type ToAnnotationRoot<
+  A extends AnyAnnotationRoot | InteropZodObject | StateSchema<any>,
+> =
+  A extends StateSchema<any>
     ? A
```
hntrl: Whats the need for this StateSchema conditional if we have a case for it in InferSchemaInput anyways?

libs/langchain/src/agents/annotation.ts
```diff
@@ -1,125 +1,164 @@
...
+  isInteropZodObject,
 } from "@langchain/core/utils/types";
-import type { BaseMessage } from "@langchain/core/messages";
+
+/**
+ * Type for jumpTo navigation targets
+ */
+type JumpToTarget = "model_request" | "tools" | "end" | undefined;
 
 export function createAgentAnnotationConditional<
```
hntrl: while we're here, can we rename this to something like `createAgentState`?

libs/langchain/src/agents/annotation.ts
```diff
@@ -1,125 +1,164 @@
...
-    | undefined = undefined,
+  StateDefinitionInit | undefined = undefined,
   TMiddleware extends readonly AgentMiddleware<any, any, any>[] = [],
 >(
   hasStructuredResponse = true,
   stateSchema: TStateSchema,
   middlewareList: TMiddleware = [] as unknown as TMiddleware
 ) {
+  const isStateSchema = (schema: unknown): schema is StateSchema<any> =>
+    StateSchema.isInstance(schema);
```
hntrl:
```suggestion
```

libs/langchain/src/agents/annotation.ts
```diff
@@ -1,125 +1,164 @@
...
-      .union([
-        z.literal("model_request"),
-        z.literal("tools"),
-        z.literal("end"),
-        z.undefined(),
-      ])
-      .optional(),
+  const stateFields: Record<string, any> = {
+    // jumpTo is used for internal navigation control
+    jumpTo: new UntrackedValue<JumpToTarget>(),
```
hntrl, line 37: @christian-bromann double checking that its fine that we're not checkpointing these? (I don't see any reason to)

libs/langchain/src/agents/annotation.ts
```diff
@@ -1,125 +1,164 @@
...
    */
-  if (stateSchema && ("shape" in stateSchema || isZodSchemaV4(stateSchema))) {
-    applySchema(stateSchema as InteropZodObject);
+  if (stateSchema) {
+    if (isStateSchema(stateSchema)) {
+      applySchema(stateSchema);
+    } else if (isInteropZodObject(stateSchema)) {
+      applySchema(stateSchema);
+    }
   }
```
hntrl: nit
```suggestion
  if (StateSchema.isInstance(stateSchema) || isInteropZodObject(stateSchema)) {
    applySchema(stateSchema);
  }
```

libs/langchain/src/agents/annotation.ts
```diff
@@ -1,125 +1,164 @@
...
+              outputFields[key] = fieldSchema;
+            } else {
+              stateFields[key] = new ReducedValue(fieldSchema, {
+                reducer: meta.reducer.fn,
+              });
+              // No inputSchema, use the value schema
+              inputFields[key] = fieldSchema;
+              outputFields[key] = fieldSchema;
+            }
+            continue;
```
hntrl, line 99: nice

libs/langchain/src/agents/tests/state_schema.test.ts
```diff
@@ -0,0 +1,570 @@
...
+          },
+        }),
+      });
+
+      const middleware = createMiddleware({
+        name: "HistoryMiddleware",
+        beforeModel: async () => {
+          return {
+            history: "from_beforeModel",
+          } as any;
```
hntrl: why do we need these as any assertions?

libs/langchain/src/agents/tests/state_schema.test.ts
```diff
@@ -0,0 +1,570 @@
...
+    });
+
+    it("should properly merge StateSchema with middleware stateSchema types", async () => {
+      const AgentState = new StateSchema({
+        agentField: z.string(),
+      });
+
+      const middleware = createMiddleware({
+        name: "TestMiddleware",
+        stateSchema: z.object({
```
hntrl: probably overkill, but can we add a test for both StateSchema in parent agent + middleware?

libs/langchain/src/agents/annotation.ts
```diff
@@ -1,125 +1,164 @@
...
   }
 
+  /**
+   * Add state schema properties from middleware.
+   * Supports both StateSchema and Zod v3/v4 objects.
+   */
   for (const middleware of middlewareList) {
     if (middleware.stateSchema) {
-      applySchema(middleware.stateSchema as InteropZodObject);
+      if (isStateSchema(middleware.stateSchema)) {
```
hntrl: same thing

libs/langchain/src/agents/middleware/types.ts
maahir30: good point, the conditional in InferSchemaInput should handle StateSchema so ill go ahead and remove it from here

libs/langchain/src/agents/tests/state_schema.test.ts
maahir30: I should have defined `stateSchema` for the middleware but I missed that. will fix it everywhere else too

PR #9864, opened 2026-01-26
fix(google-common): support nullable object types in zod to gemini schema conversion

yukukotani:
## Summary

Followup #8951

Fixed an issue where nullable/nullish object types (e.g., `z.object({...}).nullable()` or `.nullish()`) were incorrectly rejected as union types when converting Zod schemas to Gemini API schemas.

### Previous Behavior

Previously, all schemas containing `anyOf`/`oneOf` were unconditionally treated as errors. This caused simple nullable types to be rejected, such as:

```typescript
z.object({
  color: z.object({ name: z.string() }).nullable()
})
```

This Zod schema is converted to the following JSON Schema:

```json
{
  "properties": {
    "color": {
      "anyOf": [
        { "type": "object", "properties": { "name": { "type": "string" } } },
        { "type": "null" }
      ]
    }
  }
}
```

### New Behavior

This change now distinguishes between two cases:

1. **Simple nullable unions (`T | null`)**: 
   - A 2-element array where one element is `{type: "null"}`
   - Extract the non-null type and set `nullable: true`
   - Properly handles `z.object({...}).nullable()` and `.nullish()`

2. **Complex union types (e.g., discriminatedUnion)**:
   - All other union types
   - Continue to throw an error (not supported by Gemini API)

### Test Coverage

Added test cases for:

- `nullable object type`
- `optional object type`  
- `nullish object type`
- `required fields`
- Validation of `required` array for each pattern

## Related Issues

None

### DIFF
--- a/libs/providers/langchain-google-common/src/tests/zod_to_gemini_parameters.test.ts
+++ b/libs/providers/langchain-google-common/src/tests/zod_to_gemini_parameters.test.ts
@@ -1,7 +1,6 @@
 import { expect, test, describe } from "@jest/globals";
 import { z } from "zod";
 import { schemaToGeminiParameters } from "../utils/zod_to_gemini_parameters.js";
-import { extendInteropZodObject } from "@langchain/core/utils/types";
 
 describe("schemaToGeminiParameters - edge cases", () => {
   test("should throw error for discriminatedUnion", () => {


yukukotani:
@christian-bromann Could you please take a look at this since it's related to #8951?

PR #9867, opened 2026-01-26
fix: handle undefined output_index in tool call chunks

satotek:
- Set default index to 0 when event.output_index is undefined
- Ensures tool call chunks are properly merged even when output_index is missing
- Add test case for undefined output_index scenarios

Fixes # (issue)


satotek:
Closing as this was opened by mistake. A corrected PR will be opened shortly.

PR #9870, opened 2026-01-26
fix(langchain): StateSchema handling in AgentNode middleware

maahir30:
### Problem

When middleware uses `StateSchema` for its `stateSchema` property, `AgentNode` throws an error:

```
Schema must be an instance of z3.ZodObject or z4.$ZodObject
```

This happens because `interopZodObjectPartial()` only handles Zod objects, but `middleware.stateSchema` can also be a `StateSchema`.

### Solution

Added a new helper function `toPartialZodObject()` that properly handles all supported schema types:

- **ZodObject (v3/v4)**: Delegates to `interopZodObjectPartial()`
- **StateSchema**: Extracts fields (handling `ReducedValue`), converts to Zod object, and makes all fields optional
- updated E2E test to test this workflow

### DIFF
--- a/.changeset/seven-goats-train.md
+++ b/.changeset/seven-goats-train.md
@@ -1,0 +1,6 @@
+---
+"langchain": patch
+---
+
+fix(langchain): StateSchema handling in AgentNode middleware
+- Added `toPartialZodObject` helper that correctly handles both Zod objects and LangGraph's StateSchema when parsing middleware state in AgentNode .


libs/langchain/src/agents/nodes/utils.ts
```diff
@@ -145,6 +149,46 @@ export function derivePrivateState(
...
+  }
+
+  // Handle StateSchema: convert fields to Zod shape, then make partial
+  if (StateSchema.isInstance(schema)) {
+    // First extract Zod schemas from StateSchema fields
+    const shape: Record<string, any> = {};
+    for (const [key, field] of Object.entries(schema.fields)) {
+      if (ReducedValue.isInstance(field)) {
+        // For ReducedValue, use inputSchema if available, otherwise valueSchema
+        shape[key] = field.inputSchema || field.valueSchema;
```
hntrl: we should add a check that these are zod 4 schemas. If they are, we should pass them through. If not, we should just do the `z.any()` trick we were doing earlier

PR #9874, opened 2026-01-26
Bugfix: Jira failing to fetch issues after merge of  (#9832)

phong-phuong:
Apologies, the previous pull request (#9832) inadvertently caused fetchIssues() to fail in real usage after the code was modified to pass the integration tests. The change converted the Jira created date into a DateTime string to satisfy the integration test, but this broke real-life behaviour when querying Jira, because Jira expects its original date string format for filtering and pagination.

This PR:

- Restores the original Jira date string format when building JQL queries.
- Updates integration tests to match Jira’s real behaviour.
- Removes certain date filter tests that were relying on the previous, incorrect date conversion.

Impact:

- Fixes fetchIssues() for real Jira projects.
- Ensures integration tests are consistent with Jira’s behaviour.
- Prevents regression for future changes to date handling.

Testing:

- Verified fetchIssues() works against live Jira instances.
- Integration tests pass without modifying the Jira API query format.
- Removed or adjusted tests that were incompatible with real Jira date filtering.

### DIFF
--- a/.changeset/red-lies-think.md
+++ b/.changeset/red-lies-think.md
@@ -1,0 +1,5 @@
+---
+"@langchain/community": patch
+---
+
+fix jira failing to fetch issues


phong-phuong:
@hntrl rebased with no conflicts.

Issue #9877, opened 2026-01-27
Anthropic Multimodal File ContentBlock ignored.


aaronjwhiteside-appdirect:
### Checked other resources

- [x] This is a bug, not a usage question. For questions, please use the LangChain Forum (https://forum.langchain.com/).
- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain.js documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain.js rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```typescript
import { ChatAnthropic } from "@langchain/anthropic";
import { ContentBlock, HumanMessage, SystemMessage } from "@langchain/core/messages";
import dotenv from "dotenv";

dotenv.config({ path: "xxxxx/.env" });

const main = async () => {
  const model = new ChatAnthropic({
    model: "claude-sonnet-4-5",
    apiKey: process.env.ANTHROPIC_API_KEY!,
    clientOptions: {
      fetch: async (url: string | RequestInfo | URL, init?: RequestInit) => {
        console.log("request", init?.body);
        return fetch(url, init);
      },
    },
  });
  const messages = [
    new SystemMessage({
      content: "You are a helpful assistant that answers questions about the world.",
    }),
    new HumanMessage({
      contentBlocks: [
        { type: "text", text: "summary of this document" } satisfies ContentBlock.Text,
        {
          type: "file",
          url: "https://www.appropedia.org/w/images/c/ca/Writing_Sample.pdf",
          mimeType: "application/pdf",
          metadata: {
            fileId: "cmh43owcq0001rag7ce3flj36",
          },
        } satisfies ContentBlock.Multimodal.DataRecordUrl & ContentBlock.Multimodal.File,
        {
          type: "text",
          text: 'The user cannot see this text only you can, they have uploaded a file.\n<file id="cmh43owcq0001rag7ce3flj36" filename="Writing_Sample.pdf" mimeType="application/pdf" size="47104 bytes" url="https://www.appropedia.org/w/images/c/ca/Writing_Sample.pdf" />',
        } satisfies ContentBlock.Text,
      ],
    }),
  ];

  const response = await model.invoke(messages);
  console.log(response);
};

void main();
```

### Error Message and Stack Trace (if applicable)

This is the resulting request generated by LangChain. You'll note this does not contain any "document" content types.

```
[...]

PR #9881, opened 2026-01-27
fix(anthropic): handle standard file content blocks

marvikomo:
This PR fixes issue #9877, where the Anthropic provider was silently dropping files in the ContentBlock.Multimodal.File format (with type: "file" and properties like url or data). File ContentBlocks are now properly converted to Anthropic's document block format, enabling document analysis with Claude.


aaronjwhiteside-appdirect:
This doesn't appear to support `ContentBlock.Multimodal.DataRecordFileId` for Anthropic https://platform.claude.com/docs/en/build-with-claude/pdf-support#option-3-files-api

marvikomo:
> This doesn't appear to support `ContentBlock.Multimodal.DataRecordFileId` for Anthropic https://platform.claude.com/docs/en/build-with-claude/pdf-support#option-3-files-api

@aaronjwhiteside-appdirect I will add this

PR #9893, opened 2026-01-28
fix(google-common): handle nullable Zod types in Gemini schema conversion

shivgodhia:
## Description

`removeAdditionalProperties` in `zod_to_gemini_parameters.ts` throws on **all** `anyOf`/`oneOf` JSON Schema constructs. However, Zod's `.nullable()` on objects produces `anyOf: [{actualType}, {type: "null"}]`, which is a valid pattern that Gemini supports natively via its `nullable: true` field.

This means using `.nullable()` with `withStructuredOutput` on any Gemini model currently throws:

```
zod_to_gemini_parameters: Gemini cannot handle union types (discriminatedUnion, anyOf, oneOf).
Consider using a flat object structure with optional fields instead.
```

### The fix

Instead of throwing on all `anyOf`/`oneOf`, detect the nullable pattern — exactly two variants where one is `{type: "null"}` and the other is the real type. If matched, flatten it:

```
// Before (JSON Schema from Zod .nullable()):
{ anyOf: [{ type: "object", properties: {...} }, { type: "null" }] }

// After (Gemini-compatible):
{ type: "object", properties: {...}, nullable: true }
```

Genuine multi-type unions (e.g. `z.union([z.string(), z.number()])`) and discriminated unions still throw as before.

### Note

`adjustObjectType` already handles the `type: ["string", "null"]` form of nullable (which Zod produces for primitives like `z.string().nullable()`). This fix complements that by handling the `anyOf` form (which Zod produces for `z.object({...}).nullable()`).

## Tests

Added 3 test cases:
- Nullable object type via `anyOf` — verifies flattening to `{ type: "object", nullable: true }`
- Nested nullable objects — verifies recursive handling
- Real union types still throw — regression test

All 226 existing tests continue to pass.

### DIFF
--- a/libs/providers/langchain-google-common/src/utils/zod_to_gemini_parameters.ts
+++ b/libs/providers/langchain-google-common/src/utils/zod_to_gemini_parameters.ts
@@ -63,11 +63,11 @@ export function removeAdditionalProperties(
     // natively via the `nullable` field.
     if ("anyOf" in newObj || "oneOf" in newObj) {
       const variants = newObj.anyOf || newObj.oneOf;
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      const nullVariants = variants.filter((v: any) => v && v.type === "null");
+      const nullVariants = variants.filter(
+        (v: Record<string, unknown>) => v && v.type === "null"
+      );
       const nonNullVariants = variants.filter(
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        (v: any) => !v || v.type !== "null"
+        (v: Record<string, unknown>) => !v || v.type !== "null"
       );
       if (nullVariants.length === 1 && nonNullVariants.length === 1) {
         // Nullable type: flatten to { ...realType, nullable: true }
--- a/.changeset/fix-gemini-nullable-union.md
+++ b/.changeset/fix-gemini-nullable-union.md
@@ -1,0 +1,5 @@
+---
+"@langchain/google-common": patch
+---
+
+fix(google-common): handle nullable Zod types in Gemini schema conversion


libs/providers/langchain-google-common/src/utils/zod_to_gemini_parameters.ts
```diff
@@ -57,12 +57,33 @@ export function removeAdditionalProperties(
...
+    // Check for union types (anyOf, oneOf) which Gemini doesn't support.
+    // However, nullable types (e.g. Zod's .nullable()) produce
+    // anyOf: [{actualType}, {type: "null"}] which Gemini supports
+    // natively via the `nullable` field.
     if ("anyOf" in newObj || "oneOf" in newObj) {
-      throw new Error(
-        "zod_to_gemini_parameters: Gemini cannot handle union types (discriminatedUnion, anyOf, oneOf). " +
-          "Consider using a flat object structure with optional fields instead."
+      const variants = newObj.anyOf || newObj.oneOf;
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
shivgodhia: no this should not be done

libs/providers/langchain-google-common/src/utils/zod_to_gemini_parameters.ts
```diff
@@ -57,12 +57,33 @@ export function removeAdditionalProperties(
...
+      const nullVariants = variants.filter((v: any) => v && v.type === "null");
+      const nonNullVariants = variants.filter(
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
shivgodhia: no! don't do this. no typing as any

PR #9898, opened 2026-01-30
fix(openai): pass `service_tier` param to API when using `useResponsesApi`

### LINKED ISSUE #9897: service_tier parameter not passed to API when using Responses API (useResponsesApi: true)
### Checked other resources

### Example Code

```

  model: "gpt-4o",
  useResponsesApi: true,
  service_tier: "priority",  // This is ignored when useResponsesApi: true
});

const response = await model.invoke("Hello");

// Response metadata shows service_tier: "default" instead of "priority"
console.log(response.response_metadata.service_tier); // "default"


```

_No response_

### Expected behavior
When service_tier: "priority" is set on the ChatOpenAI constructor with useResponsesApi: true, the API request should include service_tier: "priority" and OpenAI should respond with service_tier: "priority" in the response metadata.

### Actual behavior
The service_tier parameter is silently ignored when using the Responses API path. The OpenAI response shows service_tier: "default".

### Root Cause
In libs/langchain-openai/src/chat_models/responses.ts, the ChatOpenAIResponses class builds request params but does not include service_tier:

```
// In ChatOpenAIResponses._getParams() or invocationParams()
{
  model: this.model,
  // ... other params ...
  ...this.modelKwargs  // service_tier is NOT explicitly included
}
```

Compare to completions.ts which does include it:
`service_tier: this.serviceTier,`

### System Info

@langchain/openai: 1.2.3
Node.js: v20.x

Muhammad-Kamran-Khan:
## Description

This PR fixes an issue where the `service_tier` parameter (e.g., "auto", "default") was being ignored when initializing `ChatOpenAI` with `useResponsesApi: true`.

Previously, the parameter was only handled in the Completions path, but missing from the new Responses API implementation. I have updated `ChatOpenAIResponses.invocationParams` to explicitly include `service_tier` in the request payload.

## Fixes
Fixes #9897

## Changes Made
- Updated `libs/providers/langchain-openai/src/chat_models/responses.ts` to pass `service_tier` to the OpenAI client.
- Added a regression test in `libs/providers/langchain-openai/src/chat_models/tests/chat_openai_responses.test.ts` to verify that the parameter is correctly forwarded.
- Added a changeset.


libs/providers/langchain-openai/src/chat_models/tests/chat_openai_responses.test.ts
hntrl, line 1: Can we fold this into the existing test file? Also think we want to add a smoke integration test just to make sure that the sdk will accept this param

Muhammad-Kamran-Khan: Thanks @hntrl for the feedback! I've refactored the tests as requested:

1. Removed the separate test file (`chat_openai_responses.test.ts`).
2. Moved the unit test into `libs/providers/langchain-openai/src/chat_models/tests/responses.test.ts`.
3. Added a smoke integration test in `libs/providers/langchain-openai/src/chat_models/tests/responses.int.test.ts` to verify the SDK accepts the parameter.

Ready for another look!

libs/providers/langchain-openai/src/chat_models/tests/responses.int.test.ts
```diff
@@ -1241,3 +1241,18 @@ it("won't modify structured output content if outputVersion is set", async () =>
...
+    useResponsesApi: true,
+    service_tier: "default",
+    apiKey: process.env.OPENAI_API_KEY || "sk-fake-key",
+  });
+  try {
+    const res = await model.invoke("Hello");
+    expect(res.content).toBeDefined();
+  } catch (e) {
+    console.log("Skipping invocation check due to missing creds/quota");
+  }
```
hntrl: this wont catch any failures, integration tests are only run presuming you have the proper credentials

```suggestion
  const res = await model.invoke("Hello");
  expect(res.content).toBeDefined();
```

libs/providers/langchain-openai/src/chat_models/tests/responses.int.test.ts
```diff
@@ -1241,3 +1241,14 @@ it("won't modify structured output content if outputVersion is set", async () =>
...
+test("ChatOpenAI with service_tier and useResponsesApi", async () => {
+  const model = new ChatOpenAI({
+    model: "gpt-4o",
+    useResponsesApi: true,
+    service_tier: "default",
+    apiKey: process.env.OPENAI_API_KEY || "sk-fake-key",
+  });
+  const res = await model.invoke("Hello");
+  expect(res.content).toBeDefined();
+});
```
christian-bromann, line 1254: Will this actually test anything? Does it make any difference whether `service_tier: "default",` is set or not? If the integration test doesn't provide much value I rather avoid having it as it adds time to the CI pipleine.

hntrl: I recommended this just to make sure that we don't get esoteric errors from passing service tier.

(also, we aren't running integration tests in the CI hot path (we would absolutely go bankrupt if that were the case))

PR #9923, opened 2026-02-05
feat(groq): add native JSON Schema structured output for gpt-oss models

eusoubrasileiro:
Add support for Groq's native JSON Schema structured output mode (strict: true) for gpt-oss models (openai/gpt-oss-20b, openai/gpt-oss-120b).

Changes:
- New groqStrictifySchema() utility to transform schemas for Groq's strict mode (additionalProperties: false, all props required, optional→nullable, removes anyOf/oneOf/not at root)
- New getGroqStructuredOutputMethod() to auto-select jsonSchema for gpt-oss models
- Enable structuredOutput: true in profiles for gpt-oss models
- Add jsonSchema case in withStructuredOutput() method

### DIFF
--- a/libs/providers/langchain-groq/src/tests/groq-schema.test.ts
+++ b/libs/providers/langchain-groq/src/tests/groq-schema.test.ts
@@ -171,10 +171,7 @@ describe("groqStrictifySchema", () => {
 
   it("should throw when root anyOf has no object variant", () => {
     const schema = {
-      anyOf: [
-        { type: "string" },
-        { type: "number" },
-      ],
+      anyOf: [{ type: "string" }, { type: "number" }],
     };
 
     expect(() => groqStrictifySchema(schema)).toThrow(


libs/providers/langchain-groq/src/utils/groq-schema.ts
```diff
@@ -0,0 +1,262 @@
...
+        // Replace root with the object variant
+        const cleaned = groqStrictifySchema(objectVariant, true);
+        return { ...cleaned, type: "object" };
+      }
+    }
+
+    // Ensure root is explicitly type: "object" and remove forbidden keywords
+    delete result.anyOf;
+    delete result.oneOf;
+    delete result.not;
```
christian-bromann: If the root has anyOf but none of its variants are objects (e.g., anyOf: [{ type: "string" }, { type: "number" }]), the code deletes anyOf without replacing it with anything meaningful. The resulting schema would have no type, no properties -- essentially empty and invalid. Consider throwing an error if no object variant is found, rather than silently producing a broken schema.

libs/providers/langchain-groq/src/utils/groq-schema.ts
```diff
@@ -0,0 +1,262 @@
...
+    return { ...schema, type: Array.from(new Set([...schema.type, "null"])) };
+  }
+
+  // Has properties but no type - it's an object
+  if (schema.properties) {
+    return { ...schema, type: ["object", "null"] };
+  }
+
+  // Last resort: assume string (safer than anyOf which breaks root)
+  return { ...schema, type: ["string", "null"] };
```
christian-bromann: If a schema has no type, no properties, no enum, and no anyOf, the function silently assumes it's a string. This could hide bugs -- a schema with a $ref but no type would get incorrectly converted. Consider either throwing or returning unchanged.

libs/providers/langchain-groq/src/chat_models.ts
```diff
@@ -1423,14 +1428,51 @@ export class ChatGroq extends BaseChatModel<
...
     let outputParser: BaseLLMOutputParser<RunOutput>;
     let llm: Runnable<BaseLanguageModelInput>;
 
-    if (method === "jsonMode") {
+    if (method === "jsonSchema") {
+      // Native JSON Schema mode for gpt-oss models
+      // Uses Groq's strict JSON Schema response format
+      const schemaName = name ?? "structured_output";
+      const description = getSchemaDescription(schema);
+      const jsonSchema = toJsonSchema(schema);
```
christian-bromann: When schema is already a plain JSON object (not Zod), toJsonSchema double-converts. While toJsonSchema likely handles this gracefully, it's worth being explicit. The OpenAI implementation handles this by passing the raw [schema directly](https://github.com/Muhammad-Kamran-Khan/langchainjs/blob/a3530be314966dab139cbddcf7a64eb303935c37/libs/providers/langchain-openai/src/chat_models/base.ts#L1046).

libs/providers/langchain-groq/src/utils/groq-schema.ts
```diff
@@ -0,0 +1,262 @@
...
+  return result;
+}
+
+/**
+ * Models that support native JSON Schema structured output.
+ */
+const MODELS_WITH_JSON_SCHEMA_SUPPORT = [
+  "openai/gpt-oss-20b",
+  "openai/gpt-oss-120b",
+];
```
christian-bromann: This will get potentially outdated, shall we do `model.startsWith("openai/gpt-oss")` instead?

eusoubrasileiro:
please take a look if I missed anything
thanks a lot!

libs/providers/langchain-groq/src/profiles.ts
```diff
@@ -169,7 +169,7 @@ const PROFILES: Record<string, ModelProfile> = {
     audioOutputs: false,
     videoOutputs: false,
     toolCalling: true,
-    structuredOutput: false,
+    structuredOutput: true, // Supports native JSON Schema structured output
```
christian-bromann:
```suggestion
    structuredOutput: true,
```

libs/providers/langchain-groq/src/profiles.ts
```diff
@@ -183,7 +183,7 @@ const PROFILES: Record<string, ModelProfile> = {
...
```
christian-bromann:
```suggestion
    structuredOutput: true,
```

christian-bromann (approved):
LGTM 👍

christian-bromann:
Thank you for the contribution!

Issue #9930, opened 2026-02-05
@langchain/aws BedrockEmbeddings ignores modelKwargs, has no way to adjust dimension of output vector


boyer-victor:
### Checked other resources

- [x] This is a bug, not a usage question. For questions, please use the LangChain Forum (https://forum.langchain.com/).
- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain.js documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain.js rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

The following code show's that the modelKwargs parameter found on the python client does not adjust dimension:

```typescript
import { BedrockEmbeddings } from '@langchain/aws';

const embeddings = new BedrockEmbeddings({
  credentials: {
    accessKeyId: "your-access-key-id",
    secretAccessKey: "your-secret-access-key",
  },
  model: 'amazon.titan-embed-text-v2:0',
  region: process.env["AWS_REGION"] || 'us-east-1',
});

// Embed a query and log the result
const res = await embeddings.embedQuery(
  "What would be a good company name for a company that makes colorful socks?"
);
console.log({ length: res.length });

const embeddings512 = new BedrockEmbeddings({
  credentials: {
    accessKeyId: "your-access-key-id",
    secretAccessKey: "your-secret-access-key",
  },
  model: 'amazon.titan-embed-text-v2:0',
  region: process.env["AWS_REGION"] || 'us-east-1',
 modelKwargs: {
    dimensions: 512
    }
});

const res512 = await embeddings512.embedQuery(
  "What would be a good company name for a company that makes colorful socks?"
);
console.log({ length: res512.length });

```

### Description

I'm using the langchain/aws, but the BedrockEmbeddings which uses a [BedrockEmbeddingParams](https://reference.langchain.com/javascript/interfaces/_langchain_aws.BedrockEmbeddingsParams.html) does not support modelKwargs or any configuration to adjust dimensions, normalization for titan models. It should either be provided in an object or as a top level optional parameter as is the case in the [python library](https://github.com/langchain-ai/langchain-aws/blob/b3d339c0f144ad0ee6a41614a6976c107f07069b/libs/aws/langchain_aws/embeddings/bedrock.py#L45).

### System Info

"bun info langchain"
bun info langchain
[...]

PR #9937, opened 2026-02-06
fix(aws): add modelKwargs and dimensions support to BedrockEmbeddings

akintunero:
BedrockEmbeddings ignored modelKwargs and had no way to adjust the output vector dimension, unlike the Python equivalent and other Bedrock integrations (LLM, Chat) in the same codebase.

Changes:
- Add modelKwargs param to BedrockEmbeddingsParams interface, spread into the InvokeModel request body
- Add dimensions as a first-class param (takes precedence over modelKwargs.dimensions)
- Rename response body variable to avoid shadowing
- Apply same fix to deprecated @langchain/community copy
- Add unit tests verifying request body construction

Fixes langchain-ai/langchainjs#9930

### DIFF
--- a/libs/langchain-community/src/embeddings/bedrock.ts
+++ b/libs/langchain-community/src/embeddings/bedrock.ts
@@ -27,6 +27,23 @@ export interface BedrockEmbeddingsParams extends EmbeddingsParams {
   region?: string;
 
   credentials?: CredentialType;
+
+  /**
+   * Additional keyword arguments to pass to the model as part of the
+   * InvokeModel request body. These are merged into the request payload,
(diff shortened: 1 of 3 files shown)


libs/langchain-community/src/embeddings/bedrock.ts
christian-bromann, line 82: I am not sure if we want to call it `modelKwargs` here, we want to move away from this Python'ish language, so maybe just `modelParameters`?

akintunero: Thanks @christian-bromann 

I agree completely – `modelParameters` is much cleaner.

christian-bromann:
Closing in favor of #9976

PR #9946, opened 2026-02-06
Fix CVE-2025-66031: override node-forge to >=1.3.2

jkennedyvz:
The `node-forge` package (< 1.3.2) has a high severity DoS vulnerability via unbounded ASN.1 recursion. Adding a pnpm override forces all transitive `node-forge` dependencies (from firebase-admin) to resolve to the patched version.


christian-bromann:
Thank you @jkennedyvz 🙏

PR #9973, opened 2026-02-09
fix(google-common): surface actual API error when gaxios throws for non-2xx responses

hntrl:
## Summary

Fixes an issue where `@langchain/google-vertexai` users see `"undefined"` as the error message when Google Gemini can't fetch `image_url` / `fileData` URLs (e.g. CDN thumbnails that 404 due to race conditions, robots.txt-blocked URLs, or timeouts).

The root cause: `GAuthClient._fetch` delegates to `google-auth-library`'s `GoogleAuth.fetch()`, which uses gaxios internally. For non-2xx responses, gaxios **throws** a `GaxiosError` before returning a `Response` object. This meant `_request()`'s well-formatted `!res.ok` error handler was never reached — the raw `GaxiosError` propagated with an empty/undefined `.message`.

### What changed

- **`libs/providers/langchain-google-common/src/auth.ts`**: Wrapped `this._fetch()` in a `try/catch` that intercepts errors thrown by the underlying fetch implementation (e.g. `GaxiosError`), extracts the HTTP status and response body from the error object, and re-throws with the same `"Google request failed with status code N: <body>"` format. Both the `catch` path and the existing `!res.ok` path now funnel through a shared `_throwRequestError()` helper to avoid duplication.

### Before → After

| | Before (bug) | After (fix) |
|---|---|---|
| **Error message** | `"undefined"` or `""` | `"Google request failed with status code 400: {"error":{"code":400,"message":"Cannot fetch content from the provided URL...","status":"INVALID_ARGUMENT"}}"` |
| **`.response.status`** | Available on `GaxiosError` but not surfaced | Preserved on the re-thrown error |
| **Debuggability** | User has no idea what went wrong | User sees exact URL that failed and why |


libs/providers/langchain-google-common/src/auth.ts
```diff
@@ -42,6 +42,32 @@ export abstract class GoogleAbstractedFetchClient
     }
   }
 
+  /* eslint-disable @typescript-eslint/no-explicit-any */
+  /**
+   * Build and throw a standardised Google request error.
+   * Both the `!res.ok` path (native fetch) and the catch path (gaxios)
+   * funnel through here so the caller always sees the same shape.
+   */
+  protected _throwRequestError(
```
christian-bromann, line 48:
```suggestion
  #throwRequestError(
```

libs/providers/langchain-google-common/src/auth.ts
```diff
@@ -42,6 +42,32 @@ export abstract class GoogleAbstractedFetchClient
     }
   }
 
+  /* eslint-disable @typescript-eslint/no-explicit-any */
```
christian-bromann:
```suggestion
```

libs/providers/langchain-google-common/src/auth.ts
```diff
@@ -42,6 +42,32 @@ export abstract class GoogleAbstractedFetchClient
...
+      opts: GoogleAbstractedClientOps;
+      fetchOptions?: Record<string, unknown>;
+    }
+  ): never {
+    const message = body
+      ? `Google request failed with status code ${status}: ${body}`
+      : `Google request failed with status code ${status}`;
+    const error = new Error(message);
+    (error as any).response = response;
+    (error as any).details = context;
```
christian-bromann, line 65: Should we create a custom error class for this?

libs/providers/langchain-google-common/src/auth.ts
```diff
@@ -42,6 +42,32 @@ export abstract class GoogleAbstractedFetchClient
...
+  /* eslint-disable @typescript-eslint/no-explicit-any */
+  /**
+   * Build and throw a standardised Google request error.
+   * Both the `!res.ok` path (native fetch) and the catch path (gaxios)
+   * funnel through here so the caller always sees the same shape.
+   */
+  protected _throwRequestError(
+    status: number,
+    body: string | undefined,
+    response: any,
```
christian-bromann:
```suggestion
    response: unknown,
```

libs/providers/langchain-google-common/src/auth.ts
```diff
@@ -42,6 +42,32 @@ export abstract class GoogleAbstractedFetchClient
...
+  ): never {
+    const message = body
+      ? `Google request failed with status code ${status}: ${body}`
+      : `Google request failed with status code ${status}`;
+    const error = new Error(message);
+    (error as any).response = response;
+    (error as any).details = context;
+    throw error;
+  }
+  /* eslint-enable @typescript-eslint/no-explicit-any */
```
christian-bromann:
```suggestion
```

libs/providers/langchain-google-common/src/auth.ts
hntrl: moved inline

libs/providers/langchain-google-common/src/auth.ts
hntrl: I'd rather match convention in this class here, I think we move to private members later if its important

libs/providers/langchain-google-common/src/auth.ts
hntrl: making the minimally invasive change here. this will be invalidated by new google package

PR #9976, opened 2026-02-09
fix(aws): add modelParameters and dimensions support to BedrockEmbeddings

akintunero:
- Add modelParameters as the preferred param for passing extra model options
- Remove modelKwargs entirely; use modelParameters as the sole extra-params container.
- Add top-level dimensions param with precedence over modelParameters.dimensions
- Support both Titan (inputText) and Nova (messages) request formats
- Include changeset for patch bump


libs/providers/langchain-aws/src/embeddings.ts
```diff
@@ -43,6 +43,30 @@ export interface BedrockEmbeddingsParams extends EmbeddingsParams {
...
+   *
+   * If `dimensions` is also provided as a top-level parameter, it will take
+   * precedence over a `dimensions` key set in `modelParameters`.
+   */
+  modelParameters?: Record<string, unknown>;
+
+  /**
+   * @deprecated Use `modelParameters` instead.
+   */
+  modelKwargs?: Record<string, unknown>;
```
christian-bromann: We never introduced this parameter, so there is no reason to deprecate it, just remove `modelKwargs`

akintunero (commented):
It has been removed

christian-bromann (approved):
Thanks for tackling this feature! The implementation looks solid and addresses the previous feedback well. LGTM 👍

PR #9986, opened 2026-02-10
Don't re-emit input state from middleware hooks that return void

brettshollenberger:
## Bug Description

When using `deepagentsjs`, the agent creates some number of todos, then sets number of todos to 0 (empty array), then returns the todos with updates. Any frontend observing the todos temporarily loses them. 

##  Root Cause

When a middleware lifecycle hook (e.g., `afterModel`) returns `void 0` to indicate "no changes", `MiddlewareNode.invokeMiddleware()` returns `{ ...state, jumpTo: void 0 }`, spreading the entire input state back as a node output. This causes every key from the middleware's stateSchema to be written back to its channel as a no-op update on every hook invocation.

For middleware like `todoListMiddleware` (which declares `todos: z.array(TodoSchema).default([])` in its stateSchema and has an `afterModel` hook), this means:

  **Step 1:** model_request
           LLM returns AIMessage { tool_calls: [write_todos([n items]), task(...), ...] }
           Channel state: todos = []  (write_todos hasn't run yet)

  **Step 2:** todoListMiddleware.after_model
           Reads todos channel → []
           Hook returns void 0
           invokeMiddleware returns { todos: [], jumpTo: void 0, ... }
           Writes todos: [] to channel  ← this is the spurious 0

  **Step 3:** Router sees tool_calls → routes to tools node

  **Step 4:** tools node
           write_todos executes → Command({ update: { todos: [n items] } })
           Writes todos: [n items] to channel  ← this is the real value

When I apply this as a patch locally, it resolves the issue. Is there any reason not to do this, to avoid spurious writes?

### DIFF
--- a/libs/langchain/src/agents/nodes/middleware.ts
+++ b/libs/langchain/src/agents/nodes/middleware.ts
@@ -18,7 +18,8 @@ class AgentRuntime {}
 
 type NodeOutput<TStateSchema extends Record<string, any>> =
   | TStateSchema
-  | Command<any, TStateSchema, string>;
+  | Command<any, TStateSchema, string>
+  | { jumpTo?: JumpToTarget };
 
 export interface MiddlewareNodeOptions {
   getState: () => Record<string, unknown>;


libs/langchain/src/agents/nodes/middleware.ts
```diff
@@ -119,10 +119,12 @@ export abstract class MiddlewareNode<
...
     /**
-     * If result is undefined, return current state
+     * If result is undefined, the hook made no state changes — return
+     * only the jumpTo sentinel so we don't re-emit every input key as
+     * a state update.
      */
     if (!result) {
-      return { ...state, jumpTo: undefined };
+      return { jumpTo: undefined };
     }
```
christian-bromann, line 129: Excellent fix! This change correctly prevents spurious state updates when middleware hooks return void. The comment clearly explains the intent and the implementation is minimal and correct.

The original behavior was problematic because spreading `...state` would emit updates for all keys in the middleware's stateSchema, even when the hook explicitly indicated no changes by returning `undefined`.

Returning only `{ jumpTo: undefined }` maintains the necessary control flow sentinel while eliminating unnecessary state channel writes.

christian-bromann:
just fyi: I've been testing a review agent, so my comment above sounds very AI'ish .. will tweak the wording. But definitely great PR!

christian-bromann:
@brettshollenberger it seems like our type interface requires `jumpTo` to be defined, we should make adjustments to allow `undefined` as well. See

```
⎯⎯⎯ Unhandled Source Error ⎯⎯⎯
TypeCheckError: Object literal may only specify known properties, and 'jumpTo' does not exist in type 'Command<any, TStateSchema, string>'.
 ❯ src/agents/nodes/middleware.ts:127:16
    125|      */
    126|     if (!result) {
    127|       return { jumpTo: undefined };
       |                ^
    128|     }
    129| 

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
```

brettshollenberger:
@christian-bromann gotcha, new change should fix that type error!

PR #9987, opened 2026-02-10
fix(google-common): prevent 400 error when AIMessage has empty text content

### LINKED ISSUE #9922: google-common: When Gemini is invoked with an AIMessage that contains empty string text content the resulting message causes a 400 response from Vertex API
### Checked other resources

### Example Code

```javascript
import { HumanMessage, AIMessage } from "@langchain/core/messages";

async function main() {
    model: "gemini-2.0-flash-001",
    temperature: 0.7,
    authOptions: {
      projectId: process.env.GOOGLE_CLOUD_PROJECT,
        apiKey: process.env.GOOGLE_VERTEX_API_KEY,
    },
  });

  const messages = [
    new HumanMessage("I think the capital of France is Paris, if that is the case respond with a blank message"),
    new AIMessage(""),
    new HumanMessage("What is a famous landmark there?"),
  ];

  console.log("Sending messages to ChatVertexAI...\n");

  console.log("Response:");
  console.log(response.content);
}

main().catch(console.error);
```

```json
{
  "error": {
    "code": 400,
    "message": "Unable to submit request because it must include at least one parts field, which describes the prompt input. Learn more: https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/gemini",
    "errors": [
      {
        "message": "Unable to submit request because it must include at least one parts field, which describes the prompt input. Learn more: https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/gemini",
        "domain": "global",
        "reason": "badRequest"
      }
    ],
    "status": "INVALID_ARGUMENT"
  }
}
```

### Description

Gemini regularly generates a blank message response (usually when making a tool call at the request of a user where no additional information is needed). Naively taking this LLM output and passing it back as part of the history results in this error.

 * Expected Behavior: A message with an empty string text part should be passed to Gemini and the LLM should generate a response
 * Actual Behavior: A 400 error from the Vertex API
 * Workaround: Replace `""` with `" "` in any messages being sent to the Vertext API

### System Info

Linux
Node v20.20.0
@langchain/core 1.1.19
@langchain/google-vertexai 2.1.15

mbarinov:
When Gemini produces a blank response (commonly during tool calls), passing that empty `AIMessage` back as conversation history causes the Vertex API to reject with a 400 error:

> "Unable to submit request because it must include at least one parts field"

**Root cause:** `messageContentText()` returns `null` for empty strings (because `""` is falsy in JS). The `null` gets filtered out in `messageContentToParts()`, producing a message with an empty `parts` array — which the Gemini API rejects.

**Fix:** Added a minimal safety fallback in `roleMessageToContent()` that ensures the `parts` array is never empty. When both `contentParts` and `toolParts` are empty, a `{ text: "" }` part is pushed as a valid placeholder. This approach was chosen over modifying `messageContentText` directly because that would shift part indices and break the position-based `thoughtSignature` alignment for messages with tool calls.

**Tests added:**
- `AIMessage` with empty string content produces a valid text part
- `AIMessage` with empty string in complex content array produces a valid text part
- `AIMessage` with non-empty text content still works correctly
- All 240 existing tests continue to pass

Fixes #9922


libs/providers/langchain-google-common/src/utils/gemini.ts
```diff
@@ -732,6 +732,14 @@ export function getGeminiAPI(config?: GeminiAPIConfig): GoogleAIAPI {
     }
     const parts: GeminiPart[] = [...contentParts, ...toolParts];
 
+    // Ensure at least one part exists to prevent Gemini API 400 errors
+    // when a message has empty text content and no tool calls.
+    // This commonly happens when an AI message with empty content
+    // (e.g. during a tool call) is passed back as conversation history.
+    if (parts.length === 0) {
```
christian-bromann, line 749: This is a good safety fallback! However, I'm curious about the edge case where a message has tool calls but also empty content. Could you add a test case that ensures a message with both empty text content AND tool calls doesn't get the empty text part added unnecessarily? 

For example:
```javascript
const messageWithToolCall = new AIMessage({
  content: "",
  tool_calls: [{ name: "search", args: { query: "test" } }]
});
```

This would help ensure the fix only applies when both `contentParts` and `toolParts` arrays are truly empty.

christian-bromann (changes requested):
Thanks for the contribution! This is a solid fix for the Gemini API 400 error when dealing with empty AIMessage content. The solution is well-targeted and the tests provide good coverage of the edge cases.

mbarinov:
Hi @christian-bromann,

Just following up on this PR. I’ve added the requested edge case test and addressed the previous feedback. Could you please take another look when you have time?

Thank you.

libs/providers/langchain-google-common/src/utils/gemini.ts
mbarinov: @christian-bromann  Following up on this PR.

PR #9990, opened 2026-02-10
feat(core,community): ssrf hardening

hntrl:
## Summary

* Add new @langchain/core/utils/ssrf module with URL validation utilities to protect against SSRF attacks (private IPs, cloud metadata endpoints, localhost)
* Harden RecursiveUrlLoader in @langchain/community by integrating shared SSRF utilities — replaces vulnerable startsWith URL matching with origin-based comparison and adds validateSafeUrl before all fetch operations
* Bump @langchain/core peer dependency in @langchain/community to >=1.1.21

New exports
* @langchain/core/utils/ssrf:
  * validateSafeUrl(url, options?) — async URL validation with DNS resolution (throws on unsafe URLs)
  * isSafeUrl(url, options?) — non-throwing boolean wrapper
  * isPrivateIp(ip) — checks RFC 1918, loopback, link-local ranges
  * isCloudMetadata(hostname, ip?) — detects cloud metadata endpoints (AWS, GCP, Azure)
  * isLocalhost(hostname, ip?) — detects localhost variations
  * isSameOrigin(url1, url2) — origin-based URL comparison

### DIFF
--- a/libs/langchain-core/src/utils/ssrf.ts
+++ b/libs/langchain-core/src/utils/ssrf.ts
@@ -32,7 +32,8 @@ const LOCALHOST_NAMES = ["localhost", "localhost.localdomain"];
 /**
  * IPv4 regex: four octets 0-255
  */
-const IPV4_REGEX = /^(?:(?:25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)\.){3}(?:25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)$/;
+const IPV4_REGEX =
+  /^(?:(?:25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)\.){3}(?:25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)$/;
 
 /**
  * Check if a string is a valid IPv4 address.
@@ -261,7 +262,6 @@ export function isLocalhost(hostname: string, ip?: string): boolean {
 
   return false;
 }
-
 
 /**
  * Validate that a URL is safe to connect to.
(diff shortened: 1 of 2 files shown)


libs/langchain-core/src/utils/ssrf.ts
```diff
@@ -0,0 +1,401 @@
...
+  }
+
+  const addr = parseIp(addrStr);
+  if (!addr) {
+    return null;
+  }
+
+  const prefixLen = parseInt(prefixStr, 10);
+  if (isNaN(prefixLen)) {
+    return null;
```
christian-bromann, line 144: Consider adding validation for the `prefixLen` parameter to ensure it's non-negative:
```suggestion
const prefixLen = parseInt(prefixStr, 10);
if (isNaN(prefixLen) || prefixLen < 0) {
  return null;
}
```

libs/langchain-community/src/document_loaders/web/recursive_url.ts
```diff
@@ -142,6 +143,7 @@ export class RecursiveUrlLoader
   private async getUrlAsDoc(url: string): Promise<Document | null> {
     let res;
     try {
+      validateSafeUrl(url, { allowHttp: true });
```
christian-bromann, line 146: The `validateSafeUrl` call is synchronous but placed before an async operation. Consider awaiting it for consistency with the other call on line 174:
```suggestion
      await validateSafeUrl(url, { allowHttp: true });
```

libs/langchain-community/src/document_loaders/tests/recursive_url.test.ts
```diff
@@ -0,0 +1,83 @@
+import { test, describe, expect } from "@jest/globals";
+import { RecursiveUrlLoader } from "../web/recursive_url.js";
```
christian-bromann, line 2: Consider adding an integration test that actually instantiates the loader and verifies the SSRF protection works end-to-end, rather than just testing the logic in isolation.

PR #10000, opened 2026-02-11
feat: @langchain/google

hntrl:
## Summary

Add new `@langchain/google` package -- a unified Google/Gemini integration that will replace `@langchain/google-genai`, `@langchain/google-common`, and the Vertex AI family of packages.

Huge huge huge shoutout to @afirstenberg for driving a lot of this feature work

### Why a new package?

The existing Google integration stack has accumulated significant technical debt:

- **`@langchain/google-genai`** depends on a long-deprecated SDK from Google (`@google/generative-ai`). That SDK has been superseded by `@google/genai`, but migrating to it would be a breaking change with limited upside -- we'd still be coupling to Google's SDK release cycle and its opinionated abstractions.
- **`@langchain/google-common` / Vertex AI packages** had to adopt an awkward multi-package split (`google-common`, `google-gauth`, `google-webauth`, `google-vertexai`, `google-vertexai-web`) to accommodate web and edge environments. This fragmentation makes it confusing for users to know which package to install, duplicates logic across packages, and creates a maintenance burden.
- **Hand-coded types in `google-common`** drifted from the actual Gemini API surface over time, leading to broken fields, missing properties, and constant churn to keep them in sync with Google's rapid iteration.

### Usage

The package exposes two entrypoints. The constructor takes the model name as the first positional argument, with an optional params object:

**`@langchain/google`** -- for web, edge, and serverless environments (Cloudflare Workers, Vercel Edge, Deno, browsers). Uses the Web Fetch API and authenticates via API key only.

```typescript

const model = new ChatGoogle("gemini-2.5-flash", {
  apiKey: "...", // or set GOOGLE_API_KEY env var
});
```

**`@langchain/google/node`** -- for Node.js environments. Extends the base entrypoint with `google-auth-library` support, enabling Vertex AI via Application Default Credentials (ADC), service account JSON, and other Node-native auth flows.

```typescript

// Google AI with API key (reads GOOGLE_API_KEY env var if not provided)
const aiStudio = new ChatGoogle("gemini-2.5-flash");

// Vertex AI with ADC (auto-detected)
const vertex = new ChatGoogle("gemini-2.5-flash", {
  vertexai: true,
  project: "my-gcp-project",
  location: "us-central1",
});
```

Both entrypoints export the same `ChatGoogle` class name, so switching between them is just an import path change.

### Versioning and stability

This package is being published at **0.1.0** to signal that it is an early release. While we have confidence in the core functionality (chat, streaming, tool calling, reasoning, multimodal), the API surface may evolve as we gather feedback and close the remaining feature gaps relative to the existing packages. We intend to promote to 1.0 once we've validated the API through user feedback and reached feature parity with the existing packages.

### Support policy for existing packages

The existing `@langchain/google-genai`, `@langchain/google-vertexai`, and related packages **will continue to receive maintenance updates** (bug fixes, security patches) for the foreseeable future. We are not deprecating them with this release. Once `@langchain/google` reaches feature parity, we will publish a migration guide and announce a deprecation timeline with adequate notice.

---

### New package: `libs/providers/langchain-google/`

| Area | Files | Description |
|------|-------|-------------|
| **Chat model** | `base.ts`, `node.ts`, `index.ts`, `types.ts` | `ChatGoogle` extending `BaseChatModel` with tool calling, streaming, structured output, and multimodal support |
| **API types** | `api-types.ts`, `typegen.jsonc` | Auto-generated `Gemini` namespace from the OpenAPI spec, plus module augmentations for utility types and spec gaps |
| **Converters** | `converters/messages.ts`, `converters/params.ts`, `converters/tools.ts` | Bidirectional conversion between LangChain messages/tools and Gemini wire format |
| **Clients** | `clients/index.ts`, `clients/node.ts` | HTTP client with platform-specific auth (Google AI API key vs Vertex AI OAuth/ADC) |
| **Utilities** | `utils/errors.ts`, `utils/gcp-auth.ts`, `utils/handler.ts`, `utils/misc.ts`, `utils/stream.ts` | Error handling, GCP auth helpers, SSE stream parsing |
| **Tests** | `tests/index.test.ts`, `tests/index.int.test.ts` | 40 unit tests (reasoning config across model families), integration tests |

### `@langchain/core` changes

- Added `ChatGoogleTranslator` block translator (`block_translators/google.ts`) and registered it under the `"google"` provider key.

### DIFF
--- a/libs/langchain-core/src/messages/block_translators/google.ts
+++ b/libs/langchain-core/src/messages/block_translators/google.ts
@@ -10,17 +10,19 @@ function convertToV1FromChatGoogleMessage(
     const content = iife(() => {
       if (typeof message.content === "string") {
         if (message.additional_kwargs.originalTextContentBlock) {
-          return [{
-            ...message.additional_kwargs.originalTextContentBlock,
-            type: "text",
-          }]
+          return [
+            {
+              ...message.additional_kwargs.originalTextContentBlock,
+              type: "text",
+            },
+          ];
         } else {
-          return [{ type: "text", text: message.content }]
+          return [{ type: "text", text: message.content }];
         }
       } else {
         return message.content;
       }
-    })
+    });
     for (const block of content) {
       const contentBlockBase: ContentBlock.Standard = iife(() => {
         if (_isContentBlock(block, "text") && _isString(block.text)) {
@@ -70,26 +72,27 @@ function convertToV1FromChatGoogleMessage(
           return { type: "non_standard", value: block };
         }
         return { type: "non_standard", value: block };
-      })
+      });
       const contentBlock: ContentBlock.Standard = iife(() => {
         if ("thought" in block && block.thought) {
-          const reasoning: string = contentBlockBase.type === "text" ? contentBlockBase.text : "";
+          const reasoning: string =
+            contentBlockBase.type === "text" ? contentBlockBase.text : "";
           return {
             type: "reasoning",
             reasoning,
             reasoningContentBlock: contentBlockBase,
-          }
+          };
         } else {
           return contentBlockBase;
         }
-      })
+      });
 
       const ret: ContentBlock.Standard = {
         thought: block.thought,
         thoughtSignature: block.thoughtSignature,
         partMetadata: block.partMetadata,
         ...contentBlock,
-      }
+      };
       for (const attribute in ret) {
         if (ret[attribute] === undefined) {
           delete ret[attribute];


libs/providers/langchain-google/src/clients/index.ts
```diff
@@ -0,0 +1,294 @@
...
+   *
+   * @param request - The HTTP request to execute. Headers will be modified
+   *                  in-place to include authentication information.
+   * @returns A Promise that resolves to the HTTP response
+   *
+   * @throws {AuthError} If service account authentication fails (e.g., invalid
+   *                     credentials, network error, or token generation failure)
+   *
+   * @example
+   * ```typescript
```
christian-bromann: Consider adding error handling for the case where both `params.apiKey` and `this.credentials` are undefined, but `fetch()` is called. Currently this would result in an unauthenticated request that might fail in a confusing way.

```suggestion
  fetch(request: Request): Promise<Response> {
    if (this.params.apiKey) {
      request.headers.set(GOOGLE_API_KEY_HEADER, this.params.apiKey);
    } else if (this.credentials) {
      request.headers.set(
        GCP_API_KEY_HEADER,
        `Bearer ${getGCPCredentialsAccessToken(this.credentials)}`
      );
    } else {
      throw new Error("No authentication method available. Please provide either apiKey or credentials.");
    }
    return fetch(request);
  }
```

libs/providers/langchain-google/src/utils/gcp-auth.ts
```diff
@@ -0,0 +1,266 @@
...
+ *   type: "service_account",
+ *   project_id: "my-project",
+ *   // ... other fields
+ * });
+ * ```
+ */
+export function normalizeGCPCredentials(
+  credentials: string | GCPCredentials
+): GCPCredentials {
+  return typeof credentials === "string"
```
christian-bromann: The type assertion to `string` is redundant here since `credentials` is already checked to be a string in the conditional.

```suggestion
  return typeof credentials === "string"
    ? Object.freeze(JSON.parse(credentials))
    : Object.isFrozen(credentials)
    ? credentials
    : Object.freeze(credentials);
```

libs/providers/langchain-google/src/converters/tools.ts
```diff
@@ -0,0 +1,501 @@
...
+    const typeIndex = nullIndex === 0 ? 1 : 0;
+    obj.type = obj.type[typeIndex];
+    obj.nullable = true;
+  } else if (len === 1 && nullIndex === 0) {
+    // This is nullable only without a type, which doesn't make sense for Gemini
+    throw new Error("Gemini cannot handle null type");
+  } else if (len === 1) {
+    // Although an array, it has only one value.
+    // So set it to the string to match what Gemini expects.
+    obj.type = obj?.type[0];
```
christian-bromann: Consider handling the edge case where `obj.type` is an empty array, which would currently cause an index out of bounds error in the `adjustObjectType` function.

```suggestion
  } else if (len === 1) {
    // Although an array, it has only one value.
    // So set it to the string to match what Gemini expects.
    obj.type = obj?.type[0];
  } else if (len === 0) {
    // Empty type array is invalid
    throw new Error("Gemini cannot handle empty type array");
  } else {
```

libs/providers/langchain-google/src/chat_models/base.ts
```diff
@@ -0,0 +1,858 @@
...
+  protected _location?: string;
+
+  protected _apiVersion?: string;
+
+  protected apiClient: ApiClient;
+
+  constructor(protected params: BaseChatGoogleParams) {
+    super(params);
+
+    if (!params.apiClient) {
```
christian-bromann: Consider providing a more descriptive error message that guides users on how to resolve the issue:

```suggestion
    if (!params.apiClient) {
      throw new ConfigurationError(
        "BaseChatGoogle requires an apiClient. This should be provided automatically by ChatGoogle constructors. If you're extending BaseChatGoogle directly, please provide an apiClient instance."
      );
    }
```

libs/providers/langchain-google/src/converters/params.ts
```diff
@@ -0,0 +1,392 @@
...
+  }
+
+  const thinkingLevelRaw =
+    fields?.reasoningEffort ??
+    fields?.thinkingLevel ??
+    convertReasoningTokensToReasoningEffort(
+      model,
+      fields?.maxReasoningTokens ?? fields?.thinkingBudget
+    );
+  let thinkingLevel = thinkingLevelRaw?.toUpperCase();
```
christian-bromann: The `toUpperCase()` call could potentially throw if `thinkingLevelRaw` is not a string. Add a type guard to ensure safety:

```suggestion
  let thinkingLevel = typeof thinkingLevelRaw === "string" ? thinkingLevelRaw.toUpperCase() : thinkingLevelRaw;
```

libs/providers/langchain-google/src/clients/index.ts
```diff
@@ -0,0 +1,294 @@
...
+    }
+    if (this.credentials) {
+      request.headers.set(
+        GCP_API_KEY_HEADER,
+        `Bearer ${getGCPCredentialsAccessToken(this.credentials)}`
+      );
+    }
+    return fetch(request);
+  }
+}
```
christian-bromann, line 303: The authentication priority logic here could cause issues. When both `apiKey` and `credentials` are provided, the function only checks `this.params.apiKey` but should also check the resolved `this.apiKey` field:

```suggestion
  fetch(request: Request): Promise<Response> {
    if (this.apiKey) {
      request.headers.set(GOOGLE_API_KEY_HEADER, this.apiKey);
    } else if (this.credentials) {
      request.headers.set(
        GCP_API_KEY_HEADER,
        `Bearer ${getGCPCredentialsAccessToken(this.credentials)}`
      );
    }
    return fetch(request);
  }
```

This ensures consistent behavior with the `hasApiKey()` method.

libs/providers/langchain-google/src/converters/tools.ts
```diff
@@ -0,0 +1,501 @@
...
+  } else {
+    // Anything else could be a union type, so reject it.
+    throw new Error("Gemini cannot handle union types");
```
christian-bromann: The union type rejection here might be too aggressive and could break valid schemas. Consider providing a more informative error message that explains which specific union types are problematic:

```suggestion
    // Anything else could be a union type, so reject it.
    throw new Error(`Gemini cannot handle union types. Found type array: [${obj.type.join(', ')}]. Consider using a single type with the 'nullable' property instead.`);
```

libs/providers/langchain-google/src/utils/errors.ts
```diff
@@ -0,0 +1,698 @@
...
+    }
+
+    isInstance(obj: unknown): obj is this {
+      return (
+        typeof obj === "object" &&
+        obj !== null &&
+        symbol in obj &&
+        obj[symbol as keyof typeof obj] === true
+      );
+    }
```
christian-bromann, line 24: The type guard implementation could be more robust. Consider strengthening the check:

```suggestion
    isInstance(obj: unknown): obj is this {
      return (
        typeof obj === "object" &&
        obj !== null &&
        symbol in obj &&
        obj[symbol as keyof typeof obj] === true &&
        obj instanceof Error
      );
    }
```

This ensures the object is actually an Error instance, not just any object with the symbol property.

libs/providers/langchain-google/src/chat_models/base.ts
```diff
@@ -0,0 +1,858 @@
...
+    );
+
+    await runManager?.handleCustomEvent(`google-response-${moduleName}`, {
+      url: response.url,
+      headers: response.headers,
+      status: response.status,
+      statusText: response.statusText,
+    });
+
+    if (!response.ok) {
```
christian-bromann, line 504: The streaming response handling could benefit from better error recovery. When parsing chunks fails, the current implementation might not provide enough context:

```suggestion
            } catch (parseError) {
              // If parsing fails, still yield a chunk with error context
              const errorMessage = parseError instanceof Error ? parseError.message : 'Unknown parsing error';
              yield new ChatGenerationChunk({
                message: new AIMessageChunk({ 
                  content: `[Parse Error: ${errorMessage}]`,
                  additional_kwargs: { parseError: true }
                }),
                text: `[Parse Error: ${errorMessage}]`,
              });
            }
```

libs/providers/langchain-google/src/converters/params.ts
```diff
@@ -0,0 +1,392 @@
...
+  function isSpeechConfig(
+    config: Gemini.SpeechConfig | SimplifiedSpeechConfig
+  ): config is Gemini.SpeechConfig {
+    return (
+      typeof config === "object" &&
+      (Object.hasOwn(config, "voiceConfig") ||
+        Object.hasOwn(config, "multiSpeakerVoiceConfig"))
+    );
+  }
+
```
christian-bromann, line 325: The speech configuration conversion logic should validate the voice names against known prebuilt voices to catch configuration errors early:

```suggestion
  // If this is already a SpeechConfig, validate it before returning
  if (isSpeechConfig(config)) {
    // Add validation for voice names if needed
    return config;
  }
```

Consider adding a list of valid prebuilt voice names for validation.

libs/providers/langchain-google/src/utils/gcp-auth.ts
```diff
@@ -0,0 +1,266 @@
...
+    const expires = lastModified
+      ? Math.floor(new Date(lastModified).valueOf() / 1000) + data.expires_in
+      : now + data.expires_in;
+
+    return { expires, token: data.access_token };
+  });
+
+  cache.set(cacheKey, { created: now, promise });
+  return await promise.then((data) => data.token);
+}
```
christian-bromann, line 266: The token caching mechanism could lead to race conditions when multiple requests are made simultaneously. Consider using a more robust caching strategy or adding proper locking mechanisms to prevent duplicate token requests.

libs/langchain-core/src/messages/block_translators/google.ts
```diff
@@ -0,0 +1,109 @@
...
+          _isObject(block.functionCall) &&
+          _isString(block.functionCall.name) &&
+          _isObject(block.functionCall.args)
+        ) {
+          return {
+            type: "tool_call",
+            id: message.id,
+            name: block.functionCall.name,
+            args: block.functionCall.args,
+          };
```
christian-bromann, line 55: The tool_call conversion assumes `message.id` is available, but this might not always be the case. Add a fallback:

```suggestion
          return {
            type: "tool_call",
            id: message.id || `tool_call_${Date.now()}_${Math.random().toString(36).substr(2, 9)}`,
            name: block.functionCall.name,
            args: block.functionCall.args,
          };
```

christian-bromann (changes requested):
Thanks for the contribution! This is an excellent and comprehensive implementation of a unified Google/Gemini integration. The code is well-structured with proper abstractions, thorough error handling, and extensive documentation. The approach of consolidating the fragmented package ecosystem is very valuable. A few edge case suggestions and improvements below.

libs/providers/langchain-google/src/converters/params.ts
hntrl: we can treat this like a model string from google -- we can't know these without an extra network request locally

libs/providers/langchain-google/src/utils/gcp-auth.ts
hntrl: thats fine

libs/langchain-core/src/messages/block_translators/google.ts
hntrl: we shouldn't do this

libs/providers/langchain-google/src/chat_models/base.ts
hntrl: mmm if we do this stream consumers cant catch errors

christian-bromann (approved):

PR #10010, opened 2026-02-12
fix(langchain): fix structured output on opus

christian-bromann:
I've discovered this bug when testing on Opus with deepagents.

### DIFF
--- a/libs/langchain/src/agents/nodes/AgentNode.ts
+++ b/libs/langchain/src/agents/nodes/AgentNode.ts
@@ -910,7 +910,7 @@ export class AgentNode<
 
       Object.assign(options, {
         /**
-         * OpenAI-style options (snake_case)
+         * OpenAI-style options
          * Used by ChatOpenAI, ChatXAI, and other OpenAI-compatible providers.
          */
         response_format: {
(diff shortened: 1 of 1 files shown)


libs/langchain/src/agents/nodes/AgentNode.ts
```diff
@@ -909,17 +909,36 @@ export class AgentNode<
...
+          format: {
+            type: "json_schema",
+            schema: structuredResponseFormat.strategy.schema,
+          },
         },
+        /**
+         * Enable the Anthropic structured outputs beta via both the `betas`
+         * array (used by invocationParams) and the `headers` fallback.
+         */
+        betas: ["structured-outputs-2025-11-13"],
```
hntrl: I don't think this beta is needed anymore

hntrl (approved):

libs/langchain/src/agents/nodes/AgentNode.ts
christian-bromann: Yep , seems like:

> Migrating from beta? The output_format parameter has moved to output_config.format, and beta headers are no longer required. The old beta header (structured-outputs-2025-11-13) and output_format parameter will continue working for a transition period. See code examples below for the updated API shape.

PR #10013, opened 2026-02-12
fix(anthropic): fix streaming with server tools producing invalid_too…

saakshigupta2002:
Fixes three related bugs that cause streaming with Anthropic server tools (`web_search`, `web_fetch`, `code_execution`) to produce `invalid_tool_calls` entries and silently drop content blocks.

When using `ChatAnthropic` with Anthropic's `web_search_20250305` server tool and streaming enabled, the streamed `AIMessageChunk` objects incorrectly contain `invalid_tool_calls`. This happens because the streaming event handler does not distinguish between client `tool_use` blocks and `server_tool_use` blocks when processing `input_json_delta` events.

### 1. `input_json_delta` unconditionally emits `tool_call_chunks` for `server_tool_use` blocks

**Problem:** In `message_outputs.ts`, the `input_json_delta` handler creates `tool_call_chunks` for every delta event regardless of whether the parent content block is a client `tool_use` or a `server_tool_use`. Server tool deltas have no `id` or `name`, so `collapseToolCallChunks()` in `@langchain/core` flags them as malformed and pushes them to `invalid_tool_calls`.

**Fix:** Introduced a `blockTypesByIndex` map (`Map<number, string>`) that tracks content block types as they arrive in `content_block_start` events. The `input_json_delta` handler now checks whether the corresponding block is a client `tool_use` before emitting `tool_call_chunks`. For `server_tool_use` blocks, an empty array is emitted instead.

**Files changed:**
- `libs/providers/langchain-anthropic/src/utils/message_outputs.ts` - Added `blockTypesByIndex` parameter, type tracking on `content_block_start`, conditional `tool_call_chunks` emission
- `libs/providers/langchain-anthropic/src/chat_models.ts` - Creates the `blockTypesByIndex` map before the streaming loop and passes it through

### 2. Missing content block types in streaming and round-trip whitelists

**Problem:** Several Anthropic server tool result types were missing from the whitelists in both `message_outputs.ts` (streaming handler) and `message_inputs.ts` (message serialization). This caused `web_fetch_tool_result`, `code_execution_tool_result`, `mcp_tool_result`, and `mcp_tool_use` blocks to be silently dropped during streaming and multi-turn conversations.

**Fix:** Added the missing types to both whitelists:
- `web_fetch_tool_result`
- `code_execution_tool_result`
- `bash_code_execution_tool_result`
- `text_editor_code_execution_tool_result`
- `mcp_tool_result`
- `mcp_tool_use`

**Files changed:**
- `libs/providers/langchain-anthropic/src/utils/message_outputs.ts` - Extended `content_block_start` whitelist
- `libs/providers/langchain-anthropic/src/utils/message_inputs.ts` - Extended `toolTypes` whitelist

### 3. Double-JSON encoding of `server_tool_use` args

**Problem:** In `block_translators/anthropic.ts`, the `web_search` and `code_execution` handlers for `ServerToolUseBlock` check `typeof block.input === "string"` and return the raw string as the field value. During streaming accumulation, `input` is a JSON string like `'{"query": "weather NYC"}'`. The first branch returns this entire JSON string as the `query` value, producing double-nested args like `{ query: '{"query": "weather NYC"}' }`.

**Fix:** Parse the JSON string before extracting the field value using `safeParseJson()`. Falls back to the raw string if parsing fails (for partial accumulation during active streaming).

**Files changed:**
- `libs/langchain-core/src/messages/block_translators/anthropic.ts` - Parse string `input` as JSON before extracting `query`/`code` fields

## Test Plan

- [x] Added unit test: `input_json_delta` for `server_tool_use` does NOT emit `tool_call_chunks`
- [x] Added unit test: `input_json_delta` for regular `tool_use` still emits `tool_call_chunks` (regression check)
- [x] Added unit test: Mixed `tool_use` and `server_tool_use` in same stream are handled correctly
- [x] Added unit test: `web_search_tool_result` `content_block_start` is handled
- [x] Added unit test: `web_fetch_tool_result` `content_block_start` is not dropped
- [x] Added unit test: `web_fetch_tool_result` round-trips through message formatting
- [x] Added unit test: String input for `web_search` `server_tool_use` is correctly parsed (no double-JSON)
- [x] Added unit test: String input for `code_execution` `server_tool_use` is correctly parsed (no double-JSON)
- [x] All 62 existing `chat_models.test.ts` tests pass
- [x] All 6 existing `standard_content.test.ts` tests pass
- [x] All existing block translator tests pass

## Backward Compatibility

The `blockTypesByIndex` parameter is optional with a default of `undefined`. When not provided, the handler defaults to the current behavior (always emitting `tool_call_chunks`), so existing callers are unaffected. The whitelist additions only add pass-through support for previously-dropped block types, with no behavior change for existing types.

## Reproduction

Before this fix, the following would produce `invalid_tool_calls`:

```typescript
import { ChatAnthropic } from "@langchain/anthropic";

  model: "claude-sonnet-4-5-20250514",
}).bindTools([{ type: "web_search_20250305", name: "web_search" }]);

const stream = await model.stream("What's the weather in Melbourne today?");

  // chunk.invalid_tool_calls would contain malformed entries
  // chunk.tool_call_chunks would have chunks from server_tool_use
  console.log(chunk.invalid_tool_calls); // was: [{...}], now: []
}
```

After this fix, `server_tool_use` streaming events are correctly handled as opaque server-side content blocks without interfering with the client tool call pipeline.


libs/providers/langchain-anthropic/src/chat_models.ts
```diff
@@ -1271,6 +1271,10 @@ export class ChatAnthropicMessages<
       signal: options.signal,
     });
 
+    // Track content block types by index so that input_json_delta events
+    // for server_tool_use blocks are not emitted as tool_call_chunks.
+    const blockTypesByIndex = new Map<number, string>();
```
christian-bromann, line 1278: It doesn't seem we use this anywhere after?

stheophil:
What is the state of this PR? What is missing and can we help out? We ran into the same issues and are now patching langchainjs locally, which is not optimal.

christian-bromann:
> What is the state of this PR?

It seems like the author didn't respond to a review here. I will go ahead and close this. The direction of the fix seems valid. Feel free to recreate it and I am happy to review.

PR #10044, opened 2026-02-13
fix(core): respect tracingEnabled with LangChainTracer

hntrl:
## Summary

Fixes a bug where `CallbackManager._configureSync` would not remove inherited `LangChainTracer` handlers when the current `RunTree` has `tracingEnabled: false`. This is the `@langchain/core` side of a two-part fix: https://github.com/langchain-ai/langsmith-sdk/pull/2427

Previously, even when a `traceable({ tracingEnabled: false })` wrapper disabled tracing, child LangChain Runnables could still produce traces because they inherited a `LangChainTracer` handler from a parent callback manager. Now, `_configureSync` explicitly strips inherited `LangChainTracer` handlers when the `RunTree` signals `tracingEnabled: false`.

## Changes

- **`libs/langchain-core/src/callbacks/manager.ts`** — When `RunTree.tracingEnabled === false`, `_configureSync` now removes any inherited `LangChainTracer` handlers from the callback manager. Also caches the `getTraceableRunTree()` call to avoid redundant lookups.
- **`libs/langchain-core/src/callbacks/tests/manager.test.ts`** — Three new tests:
  - Removes inherited `LangChainTracer` when `tracingEnabled=false`
  - Keeps inherited `LangChainTracer` when `tracingEnabled=true`
  - Removes tracer but keeps other custom handlers when `tracingEnabled=false`

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -425,7 +425,7 @@ importers:
         specifier: ^0.3.15
         version: 0.3.17
       rolldown:
-        specifier: 1.0.0-rc.4
+        specifier: ^1.0.0-rc.4
         version: 1.0.0-rc.4
       tsdown:
         specifier: ^0.20.3


hntrl:
EDIT: christian is a goat and fixed this

internal/build/src/index.ts
```diff
@@ -62,6 +62,11 @@ export function getBuildConfig(options?: Partial<BuildOptions>): BuildOptions {
     },
     sourcemap: true,
     unbundle: true,
+    outExtensions: ({ format }) => {
+      if (format === "es") return { js: ".js", dts: ".d.ts" };
+      if (format === "cjs") return { js: ".cjs", dts: ".d.cts" };
+      return undefined;
+    },
```
christian-bromann: Alternatively:
```
    // rolldown/tsdown can emit `.mjs` for ESM when `fixedExtension` is enabled.
    // We want stable `.js` ESM output for `"type": "module"` packages.
    fixedExtension: false,
```

internal/build/package.json
```diff
@@ -3,10 +3,10 @@
   "private": true,
   "version": "0.1.1",
   "type": "module",
-  "main": "./src/index.ts",
-  "types": "./src/index.ts",
+  "main": "./dist/index.js",
+  "types": "./dist/index.d.ts",
   "scripts": {
-    "build": "exit 0",
+    "build": "tsdown",
```
christian-bromann: We should not compile this code, `tsdown` should be able to load TypeScript files

PR #10072, opened 2026-02-17
fix(@langchain/google): emit on_chat_model_stream events for non-text content blocks

### LINKED ISSUE #10071: @langchain/google: non-text streaming chunks (executableCode, inlineData) missing from on_chat_model_stream events
### Checked other resources

- [x] The bug is not resolved by updating to the latest stable version of LangChain.js (and, if applicable, parsing).

### Example Code

```typescript

const withExec = model.bindTools([{ codeExecution: {} }]);

// Direct streaming works — all block types appear:
for await (const chunk of await withExec.stream([new HumanMessage("Calculate fibonacci numbers using code")])) {
  // chunk.content includes executableCode, codeExecutionResult, inlineData ✅
}

// But through streamEvents (or LangGraph graph.streamEvents), they're missing:
for await (const event of withExec.streamEvents([new HumanMessage("Calculate fibonacci numbers using code")], { version: "v2" })) {
  if (event.event === "on_chat_model_stream") {
    // executableCode/codeExecutionResult/inlineData chunks NEVER appear here ❌
    // They only appear in on_chat_model_end
  }
}
```

No error — the chunks are silently dropped. Non-text content blocks (`executableCode`, `codeExecutionResult`, `inlineData`) from Gemini native code execution never appear in `on_chat_model_stream` events when using `streamEvents()` or LangGraph's `graph.streamEvents()`.

### Description

When Gemini returns content blocks that contain no text (e.g., standalone `executableCode`, `codeExecutionResult`, or `inlineData` image blocks from native code execution), they are missing from `on_chat_model_stream` events.

**Root cause:** In `BaseChatGoogle._streamResponseChunks` ([base.ts:629](https://github.com/langchain-ai/langchainjs/blob/main/libs/providers/langchain-google/src/chat_models/base.ts#L629)), `handleLLMNewToken` is only called when `value.text` is truthy:

```typescript
yield value;  // ← chunk IS yielded (direct .stream() works)
if (value.text) {  // ← gate: falsy for non-text chunks
  await runManager?.handleLLMNewToken(value.text, ...);  // ← never called
}
```

`value.text` is produced by [`convertAIMessageToText`](https://github.com/langchain-ai/langchainjs/blob/main/libs/providers/langchain-google/src/converters/messages.ts#L1169) which filters to only `type === "text"` blocks. When a chunk contains only `executableCode`, `codeExecutionResult`, or `inlineData`, the text is `""` (falsy), so `handleLLMNewToken` is never called and no `on_chat_model_stream` event is emitted.

**Why direct `.stream()` works but `streamEvents()` doesn't:**
- Direct `.stream()` → `_streamIterator()` → `tapOutputIterable` emits events for ALL yielded chunks
- `streamEvents()` / LangGraph `graph.streamEvents()` → model called via `invoke` → `_generateUncached` → internal streaming via `_streamResponseChunks` → relies on `handleLLMNewToken` callback to emit `on_chat_model_stream` events

**Impact:** Anyone using Gemini native code execution (`{ codeExecution: {} }`) through LangGraph or `streamEvents` cannot receive real-time streaming of code execution results or generated images. These blocks only appear in `on_chat_model_end`.

### Suggested Fix

Change line 629 in `base.ts` to always call `handleLLMNewToken`:

```typescript
// Before:
if (value.text) {
  await runManager?.handleLLMNewToken(value.text, ...);
}

// After:
await runManager?.handleLLMNewToken(value.text ?? "", undefined, undefined, undefined, undefined, { chunk: value });
```

The full `ChatGenerationChunk` is already passed via `{ chunk: value }`, so downstream consumers (like `EventStreamCallbackHandler`) receive the complete `AIMessageChunk` with all content blocks regardless of the text argument.

I can submit a PR with this fix if this makes sense.

### System Info

```
@langchain/google@0.1.0
@langchain/core@0.3.44
@langchain/langgraph@0.2.60
Node.js v22.x
macOS
```

### Related Issues

- Python equivalent: https://github.com/langchain-ai/langchain/issues/33920 (code execution tool result blocks missing from streaming)
- Gemini tool streaming: https://github.com/langchain-ai/langchainjs/issues/8454

kyrisu:
## Summary

Fixes #10071

When Gemini returns content blocks without text (e.g., `executableCode`, `codeExecutionResult`, or `inlineData` from native code execution via `{ codeExecution: {} }`), `handleLLMNewToken` was not called because it was gated by `if (value.text)`. This caused these blocks to be silently missing from `on_chat_model_stream` events when using `streamEvents()` or LangGraph's `graph.streamEvents()`.

### Root cause

In `BaseChatGoogle._streamResponseChunks` ([base.ts:629](https://github.com/langchain-ai/langchainjs/blob/main/libs/providers/langchain-google/src/chat_models/base.ts#L629)):

```typescript
yield value;                    // ← chunk IS yielded (direct .stream() works)
if (value.text) {               // ← gate: "" is falsy for non-text chunks
  await runManager?.handleLLMNewToken(value.text, ..., { chunk: value });
}
```

`value.text` is produced by `convertAIMessageToText`, which filters to only `type === "text"` blocks. For chunks containing only `executableCode`, `codeExecutionResult`, or `inlineData`, the text is `""` (falsy), so `handleLLMNewToken` is never called.

**Why direct `.stream()` works but `streamEvents()` doesn't:**
- Direct `.stream()` → `_streamIterator()` → `tapOutputIterable` emits events for ALL yielded chunks
- `streamEvents()` / LangGraph `graph.streamEvents()` → model called via `invoke` → `_generateUncached` → internal streaming via `_streamResponseChunks` → relies on `handleLLMNewToken` to emit `on_chat_model_stream` events

### Fix

Remove the `if (value.text)` gate so `handleLLMNewToken` is always called with `value.text ?? ""`. The full `ChatGenerationChunk` is already passed via `{ chunk: value }`, so downstream consumers receive the complete `AIMessageChunk` with all content blocks regardless of the text argument.

## Test plan

- [x] Existing unit tests pass (`pnpm --filter @langchain/google test` — 40 tests)
- [x] Package builds successfully (`pnpm --filter @langchain/google build`)
- [ ] Manual verification: `executableCode`/`codeExecutionResult`/`inlineData` blocks appear in `on_chat_model_stream` events when using `graph.streamEvents()` with `{ codeExecution: {} }` bound

### DIFF
--- a/libs/providers/langchain-google/src/chat_models/tests/index.test.ts
+++ b/libs/providers/langchain-google/src/chat_models/tests/index.test.ts
@@ -70,7 +70,6 @@ class MockResponse implements Response {
 class MockStreamingResponse implements Response {
   private readonly bodyText: string;
 
-
   readonly headers: Headers = new Headers();
   readonly ok: boolean = true;
   readonly redirected: boolean = false;


christian-bromann (commented):
Thanks for the contribution @kyrisu , can we have a unit test for this?

kyrisu:
Added! I need to also add MockStreamingResponse class cause I couldn't find a way to mock stream in the existing tests. Let me know if I've missed an existing mock.

I've used the following script to capture the outputs from Gemini if you want to verify the chunk structure (needs GOOGLE_API_KEY to be set):
```javascript
const API_KEY = process.env.GOOGLE_API_KEY;
if (!API_KEY) {
  console.error("Set GOOGLE_API_KEY env var");
  process.exit(1);
}

const model = "gemini-2.5-flash";
const url = `https://generativelanguage.googleapis.com/v1beta/models/${model}:streamGenerateContent?alt=sse&key=${API_KEY}`;

const body = {
  contents: [
    {
      role: "user",
      parts: [{ text: "Use matplotlib to plot y=x^2 and show the chart as an image. Use code execution." }],
    },
  ],
  tools: [{ codeExecution: {} }],
};

const res = await fetch(url, {
  method: "POST",
  body: JSON.stringify(body),
});

if (!res.ok) {
  console.error(`HTTP ${res.status}: ${await res.text()}`);
  process.exit(1);
}

// Collect full response text first, then parse SSE events
const reader = res.body.getReader();
const decoder = new TextDecoder();
let fullText = "";

while (true) {
  const { done, value } = await reader.read();
  if (done) break;
  fullText += decoder.decode(value, { stream: true });
}
fullText += decoder.decode(); // flush

// Parse SSE events: each "data: {...}\n" line
const lines = fullText.split("\n");
let chunkIndex = 0;
const chunks = [];

for (const line of lines) {
  const trimmed = line.trim();
  if (!trimmed.startsWith("data: ")) continue;
  try {
    const json = JSON.parse(trimmed.slice(6));
    chunks.push(json);
    console.log(`\n=== SSE Chunk ${chunkIndex++} ===`);
    console.log(JSON.stringify(json, null, 2));
  } catch (e) {
    console.error(`Failed to parse: ${trimmed.slice(0, 100)}...`);
  }
}

console.log(`\n=== Total SSE chunks: ${chunkIndex} ===`);

// Also output a compact summary of part types per chunk
console.log("\n=== Part types per chunk ===");
for (let i = 0; i < chunks.length; i++) {
  const parts = chunks[i]?.candidates?.[0]?.content?.parts ?? [];
  const types = parts.map(p => {
    if (p.text !== undefined) return `text(${p.text.length} chars)`;
    if (p.executableCode) return "executableCode";
    if (p.codeExecutionResult) return "codeExecutionResult";
    if (p.thought) return "thought";
    return Object.keys(p).join(",");
  });
  const finish = chunks[i]?.candidates?.[0]?.finishReason;
  console.log(`  Chunk ${i}: [${types.join(", ")}]${finish ? ` finishReason=${finish}` : ""}`);
}
```

PR #10091, opened 2026-02-18
fix(langchain): implement cache serialization for multimodal messages

### LINKED ISSUE #10092: Cache key ignores multimodal content (image_url, document blocks) causing cache collisions
### Checked other resources

### Example Code

```typescript
import { RedisCache } from "@langchain/community/caches/ioredis";
import Redis from "ioredis";

const cache = new RedisCache(new Redis(), { ttl: 60 });
  model: "gemini-2.0-flash",
  cache,
});

const result1 = await model.invoke([
  new HumanMessage({
    content: [
      { type: "image_url", image_url: { url: "data:image/png;base64,<IMAGE_A>..." } },
      { type: "text", text: "Describe this image" },
    ],
  }),
]);

// Returns cached result from first call despite a completely different image
const result2 = await model.invoke([
  new HumanMessage({
    content: [
      { type: "image_url", image_url: { url: "data:image/png;base64,<IMAGE_B>..." } },
      { type: "text", text: "Describe this image" },
    ],
  }),
]);

// result2 === result1 is wrong, silently returns stale cached response
```

No error. Silent incorrect behavior. The second call returns the cached response from the first call.

### Description

When a `BaseChatModel` has a cache enabled, the cache key is derived by converting messages to a string. The conversion chain drops **all non-text content blocks**:

1. `BaseChatModel._generateCached` ([chat_models.ts L677-678](https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-core/src/language_models/chat_models.ts#L677-L678)) builds the cache key:
   ```typescript
   const prompt = BaseChatModel._convertInputToPromptValue(baseMessage).toString();
   ```

2. `ChatPromptValue.toString()` delegates to `getBufferString(this.messages)`.

3. `getBufferString` reads `m.text` for each message ([utils.ts L346](https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-core/src/messages/utils.ts#L346)):
   ```typescript
   const readableContent = m.text;
   ```

4. `BaseMessage.get text()` ([base.ts L297-309](https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-core/src/messages/base.ts#L297-L309)) only extracts `text`-type blocks and maps everything else to `""`:
   ```typescript
   get text(): string {
     if (typeof this.content === "string") return this.content;
     if (!Array.isArray(this.content)) return "";
     return this.content
       .map((c) => {
         if (typeof c === "string") return c;
         if (c.type === "text") return c.text;
         return ""; // image_url, document, etc. silently discarded
       })
       .join("");
   }
   ```

This means two invocations that share identical text content but have different `image_url` or `document` blocks produce **the same cache key**. The first response is cached and returned for every subsequent call, regardless of image/document content.

The same issue occurs on the `cache.update` path ([chat_models.ts L837-840](https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-core/src/language_models/chat_models.ts#L837-L840)), ensuring the stale entry is never overwritten.

This affects:
- **Any `BaseCache` implementation** (RedisCache, InMemoryCache, etc.)
- **Any chat model provider** (ChatGoogleGenerativeAI, ChatOpenAI, ChatAnthropic, etc.) when multimodal messages are used with caching

**Suggested fix:** The cache key derivation should include a representation of all content blocks, not just text. For example, a dedicated cache-key codepath could `JSON.stringify` array content instead of relying on the `.text` getter. The `getBufferString` / `BaseMessage.text` behavior itself is correct for display/memory purposes and should not change.

### System Info

@langchain/core: 1.1.24
@langchain/community: 1.1.15
@langchain/google-genai: 2.1.18
platform: darwin
node: v24.13.0

daukadolt:
### Fix: Cache key for `BaseChatModel` now includes all content blocks, preventing false hits on multimodal messages

Fixes #10092

When a `BaseChatModel` has a cache enabled, the cache key was derived via `ChatPromptValue.toString()` → `getBufferString()` → `BaseMessage.text`, which silently discards all non-text content blocks (`image_url`, `document`, etc.). This means two invocations with identical text but different images/documents produced the **same cache key**, causing the second call to silently return the first call's stale cached response.

### Changes

**`libs/langchain-core/src/language_models/chat_models.ts`**
- Added `_serializeCachePrompt()` - a private helper that `JSON.stringify`s the full message content (type, content, name, tool_calls) to produce a deterministic cache key that includes **all** content blocks.
- Replaced the two cache-key derivation sites (in `_generateCached` and `generate`) that used `_convertInputToPromptValue(...).toString()` with `_serializeCachePrompt(...)`.

**`libs/langchain-core/src/language_models/tests/chat_models.test.ts`**
- Updated existing cache test assertions to match the new key format.
- Added test: **different** multimodal content (same text, different images) must **not** collide in the cache.
- Added test: **identical** multimodal content must still produce cache hits.

### Notes
- `getBufferString()` and `BaseMessage.get text()` are intentionally unchanged - their text-only behavior is correct for display/memory purposes.
- Existing caches will simply see misses under the new key format and re-populate correctly; no migration needed.


libs/langchain-core/src/language_models/chat_models.ts
```diff
@@ -155,6 +155,33 @@ export type BaseChatModelCallOptions = BaseLanguageModelCallOptions & {
...
+ */
+function _serializeCachePrompt(messages: BaseMessage[]): string {
+  return JSON.stringify(
+    messages.map((m) => {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      const obj: Record<string, any> = {
+        type: m.type,
+        content: m.content,
+      };
+      if (m.name != null) obj.name = m.name;
```
hntrl, line 236: what's this for?

daukadolt: included for additional disambiguation,  as same multimodal content with different names might collide (+ for consistency, as `getBufferString` was using name there too)

daukadolt:
Hi @hntrl, thanks for the prompt review!

> for the basic values of messages (e.g. ones with just text content) the values for this new cache key getter and the existing _convertInputToPromptValue will stay the same (this is so existing caches don't break). This can be just a simple test we can add in chat_models.test.ts

Agreed. Added [`Test ChatModel cache key backward compat for plain text messages`](https://github.com/daukadolt/langchainjs/blob/9dc538be3f81aaad639bf78e4da83a963a59d54d/libs/langchain-core/src/language_models/tests/chat_models.test.ts#L353) test case for that.

> we don't have flaky cache misses when multimodal inputs are in base64. (I wonder if in those cases we should just take a sha digest of the base64 so it's easier on the cache key lookup)

Added [test case](https://github.com/daukadolt/langchainjs/commit/dd3473acbf3d4b74e31d7635d61462a3df1768e3) for that too. I took liberty of using `runCollector` there, let me know if that looks ok

daukadolt:
force pushing last commit as latest commits were not reflecting ¯\_(ツ)_/¯

daukadolt:
Hi @hntrl, just a gentle nudge on this. I'd love to get this reviewed before any merge conflicts arise with the main branch. Let me know if you need any other changes from my side!

libs/langchain-core/src/language_models/chat_models.ts
```diff
@@ -155,6 +157,89 @@ export type BaseChatModelCallOptions = BaseLanguageModelCallOptions & {
   outputVersion?: MessageOutputVersion;
 };
 
+const BASE64_DATA_URL_RE = /^data:[^;]+;base64,/;
```
hntrl: believe using regex here would be susceptible to ReDOS, can we use some other pattern match?

hntrl: also I think this pattern for base64 is only consistent with openai -- other providers just use plain jane base64 strings

daukadolt: Good catches, both addressed:

1. Replaced the regex with a simple `_isDataUrl()` helper that uses `startsWith("data:")` + `includes(";base64,")`, no backtracking possible.
2. Added a second branch that hashes `block.data` directly when `type` is "image", "audio", or "file" (for cases if base64 gets stored as plain string without data URL wrapper)

daukadolt:
Hey @hntrl! Addressed your feedback in the latest commit, would love a second look when you get a chance! Your point about the provider assumption was a real gap, ended up being a good catch.

PR #10098, opened 2026-02-19
fix(core): preserve response_metadata in streaming invoke path

yukukotani:
## Summary

Fixes an issue where `response_metadata` (containing `finish_reason`, `usage_metadata`, `safety_ratings`, etc.) was not properly propagated when using `createAgent(...).stream()` with a specified `streamMode` parameter.

## Root Cause

When a callback handler prefers streaming (indicated by `lc_prefer_streaming = true`), the `BaseChatModel._generateUncached` method takes a fast path that streams chunks directly via `_streamResponseChunks`. However, this path was missing the crucial step of merging `chunk.generationInfo` into `chunk.message.response_metadata` that exists in the standard path and the `_streamIterator` path.

This caused providers like Google VertexAI (which set critical metadata like `safety_ratings` and `finish_reason` via `generationInfo`) to have incomplete response metadata when `streamMode` is specified.

## Testing

All tests pass:
- `pnpm --filter @langchain/core test src/language_models/tests/chat_models.test.ts` ✅
- `pnpm --filter langchain test src/agents/tests/stream_response_metadata.test.ts` ✅

## Backward Compatibility

This change is fully backward compatible. It only fixes missing behavior by adding metadata propagation where it was absent.


hntrl (dismissed):
Nice catch! Thanks for the PR!

hntrl:
Tried rebasing myself but I was getting 403 errors against your remote. Would you mind rebasing and taking a look at the failing langchain test?

yukukotani:
@hntrl Thank you for the review! Rebased and updated the tests

yukukotani:
@hntrl rebased again

PR #10102, opened 2026-02-19
feat: implement aynchronous generator tool calling for streaming partial tool results

colifran:
### Description
This PR extends the tool API so tools can be implemented as async generator functions (async function*). Each yielded value is treated as a partial result and sent to the callback system via handleToolStream; the final return value of the generator is still used as the tool’s result (e.g. for ToolMessage content or artifact).

#### API / types
- Tool handler return type: Promise<ToolOutputT> | AsyncGenerator<unknown, ToolOutputT> in:
  - DynamicStructuredTool (object schema): DynamicStructuredToolInput.func, StructuredTool._call
  - DynamicTool (string schema): DynamicToolInput.func, DynamicTool._call
 
Existing promise-returning tools are unchanged.
 - Callbacks: handleToolStream(chunk, runId, ...) is invoked once per yielded value. Order for generator tools: handleToolStart → handleToolStream (per yield) → handleToolEnd.
- handleToolStart: An optional toolCallId parameter was added so handlers can receive the tool call id directly. The tools invocation path passes it when the tool is invoked with a ToolCall or config.toolCall.

#### Implementation
- Generator consumption: Centralized in consumeAsyncGenerator in libs/langchain-core/src/runnables/iter.ts. It runs the generator loop and always closes the generator in a finally block via generator.return(), so generator cleanup (e.g. finally blocks) runs on both normal completion and on error.
- Resilience: If handleToolStream throws, the tool catches the error, calls handleToolError, and continues consuming the generator so a single failing callback does not abort the tool run; the tool still resolves with the final return value.
- DynamicTool.call: Kept as a non-async method so it can return either a Promise or an AsyncGenerator directly; the base StructuredTool.call() then detects and consumes the generator.
- Relevant files: tools/types.ts, tools/index.ts (tool types and invocation/callback wiring), callbacks/base.ts and callbacks/manager.ts (handleToolStart + toolCallId), runnables/iter.ts (consumeAsyncGenerator).

### Testing
Unit tests cover:
- Async generator tools (object schema): each yield goes to handleToolStream, final return value is the tool result.
- DynamicTool (string schema) generator: same behavior for schema: z.string(); yields and final result asserted.
- Promise-returning tools unchanged; handleToolStream is not used.
- Generator with no yields (only return): same behavior as a normal tool.
- Callback order: handleToolStart → handleToolStream (per yield) → handleToolEnd.
- Generator throws after yielding: invoke rejects and handleToolError is called.
handleToolStream throws: error reported via handleToolError, consumption continues, invoke still resolves with the final result.
- Generator throws mid-stream: generator is closed in finally and the generator’s own finally runs (cleanup verified).
handleToolStart receives toolCallId when the tool is invoked with a ToolCall.

### DIFF
--- a/libs/langchain-core/src/callbacks/base.ts
+++ b/libs/langchain-core/src/callbacks/base.ts
@@ -214,7 +214,7 @@ abstract class BaseCallbackHandlerMethodsClass {
    * Called when a streaming tool yields a partial value. Tools that are async generators
    * invoke this once per yielded value.
    */
-  handleToolStream?(
+  handleToolEvent?(
     chunk: unknown,
     runId: string,
     parentRunId?: string,
--- a/libs/langchain-core/src/callbacks/manager.ts
+++ b/libs/langchain-core/src/callbacks/manager.ts
@@ -565,13 +565,13 @@ export class CallbackManagerForToolRun
     );
   }
 
-  async handleToolStream(chunk: unknown): Promise<void> {
+  async handleToolEvent(chunk: unknown): Promise<void> {
     await Promise.all(
       this.handlers.map((handler) =>
         consumeCallback(async () => {
           if (!handler.ignoreAgent) {
             try {
-              await handler.handleToolStream?.(
+              await handler.handleToolEvent?.(
                 chunk,
                 this.runId,
                 this._parentRunId,
--- a/libs/langchain-core/src/tools/tests/tools.test.ts
+++ b/libs/langchain-core/src/tools/tests/tools.test.ts
@@ -525,7 +525,7 @@ describe("Generator tools (async function*)", () => {
       {
         callbacks: [
           {
-            handleToolStream(chunk: unknown) {
+            handleToolEvent(chunk: unknown) {
               partials.push(chunk);
             },
           },
@@ -561,7 +561,7 @@ describe("Generator tools (async function*)", () => {
     const result = await stringSchemaTool.invoke("hello", {
       callbacks: [
         {
-          handleToolStream(chunk: unknown) {
+          handleToolEvent(chunk: unknown) {
             partials.push(chunk);
           },
         },
@@ -593,7 +593,7 @@ describe("Generator tools (async function*)", () => {
       {
         callbacks: [
           {
-            handleToolStream(chunk: unknown) {
+            handleToolEvent(chunk: unknown) {
               partials.push(chunk);
             },
           },
@@ -648,7 +648,7 @@ describe("Generator tools (async function*)", () => {
             },
           },
           {
-            handleToolStream() {
+            handleToolEvent() {
               events.push("stream");
             },
           },
@@ -760,7 +760,7 @@ describe("Generator tools (async function*)", () => {
       {
         callbacks: [
           {
-            handleToolStream: async () => {
+            handleToolEvent: async () => {
               throw new Error("Callback failed");
             },
             handleToolError(err: unknown) {
--- a/libs/langchain-core/src/tools/index.ts
+++ b/libs/langchain-core/src/tools/index.ts
@@ -307,7 +307,7 @@ export abstract class StructuredTool<
       result = isAsyncGenerator(raw)
         ? await consumeAsyncGenerator(raw, async (chunk) => {
             try {
-              await runManager?.handleToolStream(chunk);
+              await runManager?.handleToolEvent(chunk);
             } catch (streamError) {
               await runManager?.handleToolError(streamError);
             }


libs/langchain-core/src/callbacks/base.ts
```diff
@@ -208,6 +208,18 @@ abstract class BaseCallbackHandlerMethodsClass {
     tags?: string[]
   ): // eslint-disable-next-line @typescript-eslint/no-explicit-any
     Promise<any> | any;
+  
+  /**
+   * Called when a streaming tool yields a partial value. Tools that are async generators
+   * invoke this once per yielded value.
+   */
+  handleToolStream?(
```
hntrl: maybe would break it out into
* `handleToolOutput` for partial outputs
* `handleToolEvent` for custom events

libs/langchain-core/src/callbacks/base.ts
```diff
@@ -209,6 +209,18 @@ abstract class BaseCallbackHandlerMethodsClass {
...
+    chunk: unknown,
+    runId: string,
+    parentRunId?: string,
+    tags?: string[]
+  ): // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    Promise<any> | any;
```
christian-bromann, line 223: Any chance this can be typed `unknown`?
```suggestion
  ): Promise<unknown> | unknown;
```

libs/langchain-core/src/callbacks/manager.ts
```diff
@@ -565,6 +565,35 @@ export class CallbackManagerForToolRun
...
+                this._parentRunId,
+                this.tags
+              );
+            } catch (err) {
+              const logFunction = handler.raiseError
+                ? console.error
+                : console.warn;
+              logFunction(
+                `Error in handler ${handler.constructor.name}, handleToolStream: ${err}`
+              );
```
christian-bromann: We do we log the error here?

libs/langchain-core/src/tools/tests/tools.test.ts
```diff
@@ -496,3 +497,139 @@ describe("tool()", () => {
...
+  });
+
+  test("handleToolStart and handleToolEnd still fire with generator tools", async () => {
+    const events: string[] = [];
+
+    const testTool = tool(
+      async function* (input: { city: string }) {
+        yield { status: "searching", city: input.city };
+        yield { status: "found", temperature: 72 };
+        return `Weather in ${input.city}: 72F`;
```
christian-bromann, line 632: I am not sure if we want to allow yielding arbitrary data payloads. At the minimum we should require users to pass along some sort of output type so rendering the data is easy. Some libraries support concrete input and output schemas for tools, e.g.:

```ts
const myTool = tool({
  inputSchema: z.object({ query: z.string() }),

  outputSchema: z.discriminatedUnion("status", [
    z.object({
      status: z.literal("progress"),
      percent: z.number(),
    }),
    z.object({
      status: z.literal("done"),
      result: z.string(),
    }),
  ]),

  async function* execute({ query }) {
    yield { status: "progress", percent: 0 };

    const result = await longRunningTask(query);

    // ✅ final result must be yielded
    yield {
      status: "done",
      result,
    };
  },
});
```

Introducing `outputSchema` may be not scoped for this effort but we should talk about if this is the direction we should be heading.

libs/langchain-core/src/callbacks/base.ts
colifran: I was mostly pattern matching here. I noticed all the other handler definitions besides `handleText`, `handleAgentAction`, and `handleAgentEnd` use `Promise<any> | any`. Do you think this needs to be typed as `unknown`?

libs/langchain-core/src/callbacks/manager.ts
colifran: Mostly just pattern matching against how we're handling the error case for the other tool handlers. They all had a pattern of logging `Error in handler ${handler.constructor.name},<handler-name>: ${err}`. I assumed that was a pre-established pattern so following it would be a predictable output for this one.

libs/langchain-core/src/tools/tests/tools.test.ts
colifran: Can you help me understand this a little bit more? I would think since the developer owns the tool and the frontend, defining an `outputSchema` might be overkill outside of getting autocomplete and type checking. What gap does something like that fill? My thinking is the input schema helps ensure the LLM is calling the tool correctly but the output format is deterministic since the developer owns it.

colifran:
@darshjme-codes updated - thanks for your review!

libs/langchain-core/src/callbacks/base.ts
colifran: What is the difference between a custom event and a partial output? My thinking was the developer yields data in whatever format they want, e.g., `yield { progress: '...', data: '...', x: '...', y: '...', z: '...' }` and then they can handle that data from partial tool output on the frontend. How are you imagining separating a tool output and a tool event?

libs/langchain-core/src/tools/tests/tools.test.ts
christian-bromann: > outside of getting autocomplete and type checking.

I think this is a big devx factor we should consider. Introducing output schemas for tools is probably out of scope for this effort but having type safety backed in I would be strongly in favor for.

colifran: I’m on board with improving type safety and devx here, but I’d lean toward deferring output schemas (or a formal “stream chunk” type) to a follow-up. This PR doesn’t remove or change any existing behavior and customers keep the same experience with the added benefit of improved streaming of partial tool chunks. We can design the typed output story (and whether we want something like outputSchema or a different approach or both) once this is in.

libs/langchain-core/src/callbacks/manager.ts
hntrl: lets just remove this

libs/langchain-core/src/tools/tests/tools.test.ts
```diff
@@ -496,3 +497,280 @@ describe("tool()", () => {
     });
   });
 });
+
+describe("Generator tools (async function*)", () => {
+  test("Generator tool yields partial results and returns final result", async () => {
+    const partials: unknown[] = [];
+
+    const testTool = tool(
+      async function* (input: { city: string }) {
```
hntrl:
```suggestion
      async function* (input) {
```

libs/langchain-core/src/callbacks/manager.ts
colifran: Will do.

PR #10137, opened 2026-02-23
fix(langchain): use StateSchema value types in middleware

hntrl:
## Summary

Fix incorrect type propagation for `StateSchema` (notably `ReducedValue`) so middleware hooks receive the stored/value type rather than the update/input type.

This showed up when providing a `StateSchema` to middleware: hook state (and some exported inference helpers) would resolve to the update type, which is wrong for reduced channels (e.g. `ReducedValue<string[], string>` should expose `string[]` to hooks).

### Correct value vs update typing for StateSchema

- Adjusted schema normalization/inference to treat `StateSchema` as its own case and infer **value** types for runtime state passed through middleware hooks.
- Split internal helpers into:
  - `InferSchemaValue<...>`: the state/value type (what’s stored and observed)
  - `InferSchemaInput<...>`: the update type (what you provide as updates)

This fixes reduced channels where the value type differs from the update type.

### Align agent-level inference with runtime

- Updated agent state/context inference helpers to use `InferSchemaValue` so `InferAgentState`/`InferAgentContext` reflect the value types that actually exist at runtime (and that middleware sees).

### DIFF
--- a/libs/langchain/src/agents/middleware/types.ts
+++ b/libs/langchain/src/agents/middleware/types.ts
@@ -94,9 +94,7 @@ export interface MiddlewareTypeConfig<
  */
 export type DefaultMiddlewareTypeConfig = MiddlewareTypeConfig;
 
-export type NormalizedSchemaInput<
-  TSchema extends StateDefinitionInit | undefined | never = any,
-> = [TSchema] extends [never]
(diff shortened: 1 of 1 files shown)


hntrl:
also makes middleware returns consistent -- a lot of middleware impl's were just lazy casting to AgentMiddleware

PR #10149, opened 2026-02-25
fix(langchain): don't use unique symbol for branding middleware

christian-bromann:
This has caused some devx friction as it causes type issues when a transient langchain dependency has a different version.

### DIFF
--- a/.changeset/friendly-lobsters-attend.md
+++ b/.changeset/friendly-lobsters-attend.md
@@ -1,0 +1,5 @@
+---
+"langchain": patch
+---
+
+fix(langchain): don't use unique symbol for branding middleware


libs/langchain/src/agents/middleware/types.ts
```diff
@@ -352,11 +352,15 @@ export type AfterAgentHook<
...
+ * String-keyed brand for middleware instances.
+ * Uses a string literal instead of a unique symbol so that the AgentMiddleware
+ * interface is structurally compatible across different package installations
+ * (e.g., when deepagents pins langchain@1.2.26 but the consumer uses 1.2.27).
+ *
+ * This still prevents functions from being accidentally assignable to
+ * AgentMiddleware since functions don't have this property.
  */
-export const MIDDLEWARE_BRAND: unique symbol = Symbol("AgentMiddleware");
+export const MIDDLEWARE_BRAND = "~middlewareBrand" as const;
```
hntrl, line 363: would this work?
```suggestion
export const MIDDLEWARE_BRAND: symbol = Symbol("AgentMiddleware");
```
I like keeping things in symbol namespace

hntrl (commented):
a little nervous that we need to do this everywhere 🙈

hntrl:
closing in favor of #10243

PR #10160, opened 2026-02-25
fix(langchain): prevent llmToolSelectorMiddleware from leaking into message stream

### LINKED ISSUE #10042: JS LLM Tool Selector Middleware: internal structuredModel.invoke() is streamed to UI
### Checked other resources

### Example Code

```typescript
import { z } from "zod";
import { createAgent, llmToolSelectorMiddleware } from "langchain";
import { MemorySaver, MessagesZodState } from "@langchain/langgraph";
 
// --- Main agent model (intentional streaming for UI) ---
const mainModel = new ChatOpenAI({
  model: "gpt-4o-mini",
  streaming: true,
  temperature: 0,
});
 
// --- Internal tool selector model (should NOT stream) ---
const toolSelectorModel = new ChatOpenAI({
  model: "gpt-4o-mini",
  streaming: false, // <-- disabled
  callbacks: [],    // <-- none
  temperature: 0,
});

// --- LLM Tool Selector middleware  ---
const toolSelectorMiddleware = llmToolSelectorMiddleware({
  model: toolSelectorModel,        // <-- internal model
  alwaysInclude: ["transfer_to_date_expert"],
  systemPrompt:
      "Your goal is to select the most relevant tools for answering the user's query. If no tools are relevant then don't use any tools."
});

// --- Minimal agent state ---
const agentState = z.object({
  messages: MessagesZodState.shape.messages,
});

// --- Agent with only tool selector middleware ---
const agent = createAgent({
  model: mainModel,
  tools: [],
  middleware: [toolSelectorMiddleware],
  stateSchema: agentState,
  checkpointer: new MemorySaver(),
});

async function main() {
  console.log("Starting stream… Watch for a leaked summary BEFORE final answer.\n");

    {
      messages: [{ role: "user", content: "what is today?"}],
    },
    {
      streamMode: ["messages"], // <-- required to observe the bug
      configurable: { thread_id: "toolSelector-reproduction" },
    }
  );
 
  for await (const [mode, chunk] of stream) {
    if (mode !== "messages") continue;

    const [msg] = chunk;
    const role = (msg as any).role ?? "unknown";
    const name = (msg as any).name ?? "";
    const content =
      typeof msg.content === "string"
        ? msg.content
        : JSON.stringify(msg.content);

    console.log(`\n[STREAMED MESSAGE] role=${role} name=${name}`);
    console.log(content.slice(0, 200) + (content.length > 200 ? "..." : ""));

    // Bug behavior:
    //
    // 1. You will see an object with a tools key appear FIRST ({"tools":[]}.  In this example no tools were selected.
    //    (coming from structuredModel.invoke inside llToolSelectorMiddleware)
    //
    // 2. Then you will see the real agent reply
    //
    // The selected tools should NEVER appear in the UI because it is an internal
    // model call
  }

  console.log("\nStream complete.");
}
 
main().catch((err) => console.error("Repro error:", err));
```

There is no thrown exception.

The bug is incorrect behavior: **unwanted streamed middleware output.**

### What I am doing
Using the official JS llmToolSelectorMiddleware with an agent that streams events (stream() or stream_mode="messages").

### What I expect
Internal LLM Tool Selector calls should be:
- completely isolated
- not streamed to front end
- invisible to the user

### What happens instead
The internal line:
```typescript
await structuredModel?.invoke([
        { role: "system", content: selectionRequest.systemMessage },
        selectionRequest.lastUserMessage,
      ])
```
inside the middleware is **always streamed to the UI** as if it were a normal assistant response.

This causes:
- an extra message to appear in the UI

### Related Issues
I believe this behavior is similar to issue #9455 which related to the same behavior but in the Summarization Middleware

### System Info

**Platform:** macOs
**Package Manager:** npm (10.8.2)
 
langchain-js: 1.2.10
langgraph-js: 1.1.2
Node: 20.19.5

JadenKim-dev:
## Summary

`llmToolSelectorMiddleware` internally calls `structuredModel.invoke()` to select relevant tools. When `agent.stream()` is used with `streamMode: "messages"`, LangGraph injects a `StreamMessagesHandler` into `config.callbacks` and stores it in `AsyncLocalStorage`. Without an explicit config override, this handler is inherited by the internal invoke call, causing the tool selection response (`{"tools":["..."]}`) to appear as an assistant message in the UI stream.

### Root cause

`ensureConfig()` in `@langchain/core` merges the explicit config with the implicit config from `AsyncLocalStorage`. If `config.callbacks` is `undefined`, the `StreamMessagesHandler` from the parent context is inherited. Passing `callbacks: []` (an empty array) breaks this inheritance because a non-`undefined` value always overrides the implicit one.

LangSmith tracing is unaffected — it is injected via `LangChainTracer.getTraceableRunTree()` inside `_configureSync`, not through `config.callbacks`.

### Fix

Build an explicit config using `mergeConfigs`:

```typescript
const baseConfig: RunnableConfig = pickRunnableConfigKeys(request.runtime) ?? {};
const config = mergeConfigs(baseConfig, {
  metadata: { lc_source: "llmToolSelector" },
  callbacks: [],
});
```

- **`callbacks: []`** prevents `StreamMessagesHandler` from being inherited via `AsyncLocalStorage`
- **`pickRunnableConfigKeys`** inherits the parent config (tags, metadata, configurable, etc.) from `runtime`
- **`lc_source: "llmToolSelector"`** tags the call for observability, consistent with `summarizationMiddleware`

Fixes #10042

## Test plan

- [x] Unit tests: `yarn test src/agents/middleware/tests/llmToolSelector.test.ts`
- [x] Integration test verifying no stream leak: `OPENAI_API_KEY=... yarn vitest run --mode int src/agents/middleware/tests/llmToolSelector.int.test.ts`


christian-bromann (approved):
LGTM 👍

libs/langchain/src/agents/middleware/tests/llmToolSelector.int.test.ts
```diff
@@ -0,0 +1,79 @@
...
+
+describe("llmToolSelectorMiddleware – streaming isolation (integration)", () => {
+  it("does not leak tool-selector output into messages stream", async () => {
+    const middleware = llmToolSelectorMiddleware({
+      model: new ChatOpenAI({ model: "gpt-4o-mini", temperature: 0 }),
+      maxTools: 1,
+    });
+
+    const agent = createAgent({
+      model: new ChatOpenAI({ model: "gpt-4o-mini", temperature: 0 }),
```
christian-bromann: We shouldn't need a real model for this and instead just use a mock as documented in https://docs.langchain.com/oss/javascript/langchain/test/unit-testing

christian-bromann (changes requested):

JadenKim-dev:
@christian-bromann
Replaced the integration test with FakeListChatModel so it no longer needs a real API call.

christian-bromann (approved):
LGTM 👍 

Thanks for the update!

Issue #10103, opened 2026-02-19
google-genai: `Unknown content type thinking` — Multi-turn crashes when thinkingConfig is enabled with tool calling


ahmadhuzaifa:
## Description

When using `ChatGoogleGenerativeAI` with `thinkingConfig.includeThoughts: true` and tools bound, the first LLM call succeeds but the **second turn crashes** with `Unknown content type thinking`.

The issue is in `_convertLangChainContentToPart` (`packages/langchain-google-genai/src/utils/common.ts`). When Gemini returns thinking blocks, they're correctly parsed into LangChain content blocks with `type: "thinking"`. However, when the conversation history is sent back to Gemini on the next turn, `_convertLangChainContentToPart` has no case for `type: "thinking"` and throws.

This makes `thinkingConfig` **completely unusable** in any multi-turn agent loop with tools (e.g., LangGraph's `createReactAgent`).

## Reproduction

```typescript
import { createReactAgent } from "@langchain/langgraph/prebuilt";
import { ChatGoogleGenerativeAI } from "@langchain/google-genai";
import { tool } from "@langchain/core/tools";
import { z } from "zod";

const getWeather = tool(
  async ({ city }) => `Weather in ${city}: 72°F, sunny`,
  {
    name: "get_weather",
    description: "Get weather for a city",
    schema: z.object({ city: z.string() }),
  }
);

const model = new ChatGoogleGenerativeAI({
  model: "gemini-2.5-flash", // any thinking-capable model
  thinkingConfig: {
    includeThoughts: true,
    thinkingBudget: 8000,
  },
});

const agent = createReactAgent({ llm: model, tools: [getWeather] });

const result = await agent.invoke({
  messages: [{ role: "user", content: "What's the weather in Tokyo and SF? Compare them." }],
});
// ^ Crashes on second LLM call after tool results are returned
```

## Error

```
Error: Unknown content type thinking
    at _convertLangChainContentToPart (node_modules/@langchain/google-genai/src/utils/common.ts:343:13)
    at convertMessageContentToParts (node_modules/@langchain/google-genai/src/utils/common.ts:406:21)
    at convertBaseMessagesToContent (node_modules/@langchain/google-genai/src/utils/common.ts:473:21)
    at ChatGoogleGenerativeAI._streamResponseChunks (node_modules/@langchain/google-genai/src/chat_models.ts:915:20)
```

## Root Cause

[...]

Issue #9624, opened 2025-12-11
Support 'thought_signature' for Gemini Thinking models in ChatVertexAI tool calling


andantonyan:
### Checked other resources

- [x] This is a bug, not a usage question. For questions, please use the LangChain Forum (https://forum.langchain.com/).
- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain.js documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain.js rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```
import { Calculator } from '@langchain/community/tools/calculator';
import { ChatVertexAI } from '@langchain/google-vertexai-web';
import { createAgent } from 'langchain';

const account = {
  ...
};

export const model = new ChatVertexAI({
  model: 'gemini-3-pro-preview',
  location: 'global',
  temperature: 0,
  maxOutputTokens: 5000,
  maxReasoningTokens: 2000,
  streaming: true,
  authOptions: {
    credentials: account,
  },
});

const agent = createAgent({
  model: model,
  tools: [new Calculator()],
});

await agent.invoke({
  messages: [['human', 'Calculate 1 + 1 using calculator tool.']],
});

// Error: Google request failed with status code 400: [{
//   "error": {
//     "code": 400,
//     "message": "Unable to submit request because function call `default_api:calculator` in the 2. content block is missing a `thought_signature`. Learn more: https://docs.cloud.google.com/vertex-ai/generative-ai/docs/thought-signatures",
//     "status": "INVALID_ARGUMENT"
//   }
// }]
```

### Description

When using `gemini-3-pro-preview` with LangChain's Google integrations, tool execution fails with a 400 error. The error message indicates that the `thought_signature` is missing from the function call response.

Gemini 3.0 models introduced a strict requirement for "Thinking" models: when the model outputs a `functionCall`, it includes a hidden `thought_signature`. This signature **must** be echoed back to the API in the subsequent `functionResponse` (tool output) to maintain the reasoning chain.

It appears the current implementation in `ChatVertexAI` and `ChatGoogle` does not capture or forward this signature when invoking tools, causing the API to reject the tool output.

**Affected Packages**
I have confirmed this issue occurs in the following packages:
1. **`@langchain/google-vertexai-web`** (using `ChatVertexAI`)
[...]

PR #10179, opened 2026-02-27
fix(@langchain/google): preserve thought and thoughtSignature on text parts

kyrisu:
Gemini thinking models (e.g. `gemini-2.5-pro`, `gemini-3-pro-preview`) return text parts with `thought: true` and `thoughtSignature` fields when performing function calling. The legacy message converter in `convertLegacyContentMessageToGeminiContent` was converting text content blocks as `{ text: item.text }`, stripping both fields. (The `functionCall` branch already preserves extra fields via `...etc` spread, so `thoughtSignature` on function call parts was not affected.)

This causes a 400 error on the next API call because a model turn containing a `functionCall` part may include `thought` parts but **not** regular `text` parts. With `thought` stripped, the API sees `[text, text, functionCall]` instead of `[thought, thought, functionCall]` and rejects the request.

Gemini 3 models additionally require `thoughtSignature` to be echoed back on all parts, per [Google's documentation](https://ai.google.dev/gemini-api/docs/thought-signatures):
> "When using Gemini 3 models, you must pass back thought signatures during function calling, otherwise you will get a validation error (4xx status code)."

### Fix

Preserve `thought` and `thoughtSignature` when converting text content blocks back to Gemini parts in the legacy (v0) path.

### Tests

- Verifies `thought` and `thoughtSignature` are preserved on text parts in a thinking model tool call scenario
- Verifies regular text parts (without thinking fields) are not affected

### Note

The diff includes unrelated Prettier formatting changes (ternary operator indentation) on lines that were already non-compliant with the repo's `.prettierrc`.

**Scope**: This fix applies to `@langchain/google` (`ChatGoogle`). The same issue exists in `@langchain/google-genai` (`ChatGoogleGenerativeAI`) — see #10103.

Related:
- #9624 — thought signature support for `ChatVertexAI` tool calling
- #10103 — same root cause in `@langchain/google-genai` (`_convertLangChainContentToPart` missing `type: "thinking"` handler)
- #10072 — prior fix for streaming non-text content blocks (merged in 0.1.3)

### DIFF
--- a/libs/providers/langchain-google/src/converters/messages.ts
+++ b/libs/providers/langchain-google/src/converters/messages.ts
@@ -56,13 +56,13 @@ export const geminiContentBlockConverter: StandardContentBlockConverter<{
           typeof block.data === "string"
             ? block.data
             : typeof block.data === "object" && block.data !== null
-            ? // Convert Uint8Array to base64 string
-              btoa(
-                Array.from(block.data as Uint8Array)
-                  .map((byte) => String.fromCharCode(byte))
-                  .join("")
-              )
-            : String(block.data);
+              ? // Convert Uint8Array to base64 string
+                btoa(
+                  Array.from(block.data as Uint8Array)
+                    .map((byte) => String.fromCharCode(byte))
+                    .join("")
+                )
+              : String(block.data);
         return {
           inlineData: {
             mimeType: block.mime_type,
@@ -129,13 +129,13 @@ export const geminiContentBlockConverter: StandardContentBlockConverter<{
           typeof block.data === "string"
             ? block.data
             : typeof block.data === "object" && block.data !== null
-            ? // Convert Uint8Array to base64 string
-              btoa(
-                Array.from(block.data as Uint8Array)
-                  .map((byte) => String.fromCharCode(byte))
-                  .join("")
-              )
-            : String(block.data);
+              ? // Convert Uint8Array to base64 string
+                btoa(
+                  Array.from(block.data as Uint8Array)
+                    .map((byte) => String.fromCharCode(byte))
+                    .join("")
+                )
+              : String(block.data);
         return {
           inlineData: {
             mimeType: block.mime_type,
@@ -202,13 +202,13 @@ export const geminiContentBlockConverter: StandardContentBlockConverter<{
           typeof block.data === "string"
             ? block.data
             : typeof block.data === "object" && block.data !== null
-            ? // Convert Uint8Array to base64 string
-              btoa(
-                Array.from(block.data as Uint8Array)
-                  .map((byte) => String.fromCharCode(byte))
-                  .join("")
-              )
-            : String(block.data);
+              ? // Convert Uint8Array to base64 string
+                btoa(
+                  Array.from(block.data as Uint8Array)
+                    .map((byte) => String.fromCharCode(byte))
+                    .join("")
+                )
+              : String(block.data);
         return {
           inlineData: {
             mimeType: block.mime_type,
@@ -714,7 +714,15 @@ function convertLegacyContentMessageToGeminiContent(
         parts.push({ text: item });
       } else if (typeof item === "object" && item !== null) {
         if (isMessageContentText(item)) {
-          parts.push({ text: item.text });
+          // Gemini thinking models include `thought` and `thoughtSignature`
+          // on text parts. These must be forwarded back to the API so it
+          // can distinguish thought parts from regular text.
+          const { thought, thoughtSignature } = item as unknown as Gemini.Part;
+          parts.push({
+            text: item.text,
+            ...(thought !== undefined ? { thought } : {}),
+            ...(thoughtSignature !== undefined ? { thoughtSignature } : {}),
+          });
         } else if (isDataContentBlock(item)) {
           parts.push(
             convertToProviderContentBlock(item, geminiContentBlockConverter)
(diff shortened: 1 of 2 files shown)


afirstenberg:
Good catch! Thank you.

I don't see a problem with this, but some thoughts:
* I'm concerned that the integration tests didn't catch this, since they're specifically looking for it.
* I didn't think we were converting v0 content to Gemini Parts to send. I thought we were using the `contentBlocks` for that, which should have the information to reconstitute the thoughts and thought signatures. So I'm a little worried that something else regressed somewhere.
* We would also need to handle the case where the legacy content is just a string (about 6 lines up). At least for the thought signature. But this is why I thought we were using the `contentBlocks` to generate what to send to Gemini.

kyrisu:
Thanks for the review @afirstenberg

Regarding the v0 vs v1 path - `convertGeminiCandidateToAIMessage` stores multi-part responses (including thought parts) as a `ContentBlock[]` array, but passes it via the `content` constructor field, not `contentBlocks`. `BaseMessage` only sets `output_version: "v1"` when `contentBlocks` is used, so `output_version` is never set and `convertMessagesToGeminiContents` always defaults to the v0 legacy path. This isn't a regression - the package never sets `output_version: "v1"` on the messages it produces, so its own output always routes through the v0 path.

Even if we switched to v1, `convertStandardContentBlockToGeminiPart` only handles `text`, `image`, `audio`, and `video` - there's no `"reasoning"` case to map back to `{ text, thought: true, thoughtSignature }` for the Gemini API. So neither path currently handles thought round-tripping cleanly; the v0 path was just closer (only needed these two fields preserved).

Happy to work on a follow-up PR to migrate `@langchain/google` to the v1 content block path if that's the right direction - that would be a bigger scoped change though.

Regarding the string content case (line ~680): when a thinking model returns thought parts + functionCall, `convertGeminiCandidateToAIMessage` stores `content` as an array (the multi-part branch at line 1123), not a string. The string branch (line 1107) requires exactly one content part that isn't a thought, so thinking model responses always take the array path. That means `convertLegacyContentMessageToGeminiContent` hits the `Array.isArray(message.content)` branch, not the string branch. But worth adding a guard for defensiveness - I can include that here or in the v1 follow-up.

afirstenberg:
> Happy to work on a follow-up PR to migrate `@langchain/google` to the v1 content block path if that's the right direction - that would be a bigger scoped change though.

A quick note on this to be clear - moving to v1 is *not* the right direction. At least not at this point, I don't think, since that would make the `content` v1 (if memory serves), and we need it to maintain this bizarre mutant v0 format for backwards compatibility.

I'll look at the rest of what you wrote when I get a moment, but thank you for breaking it down clearly.

kyrisu:
Thanks. You can hold off on the review for a bit - the PR is still valid but I've found additional issues with how this provider passes back function calls in multi turn conversation. For example:

The Vertex AI REST API (aiplatform.googleapis.com/v1) does not expose id in FunctionResponse. Sending it produces:

```
400 Invalid JSON payload received. Unknown name "id" at 
'contents[2].parts[0].function_response': Cannot find field.
```

I will update this PR when I do some more tests and verify if the same issues happens with the Google GenAI.

paddyfink:
Hello, is there a workaround while the issue is being resolved? We cannot use gemini models right now

kyrisu:
Just a quick update - while working on the `thought`/`thoughtSignature` preservation fix (Bug 1), I discovered two more bugs that cause 400 errors in multi-turn tool-call conversations with thinking models:

1. `functionResponse.name` resolves to `"unknown"` (legacy path) - The converter looks up the parent `AIMessage.name` instead of finding the matching tool call's name. This means the function response is sent back with the wrong name.

2. `functionResponse.id` rejected by Vertex AI - We're sending `tool_call_id` as `functionResponse.id`, but Vertex AI doesn't recognize that field and returns `400 Unknown name "id"`. This affects both the legacy and v1 standard converter paths.

Both bugs only manifest when the first model call returns a tool call and you send the tool result back for a second call - the exact scenario thinking models hit frequently.

**What's being added:**
- Remove `id` from `functionResponse` objects (2 locations in `messages.ts`)
- Fix name resolution on the legacy path to use `message.name || matchedToolCall.name` instead of `parentAIMessage.name`
- Unit tests covering correct name and no-id for both converter paths
- A round-trip unit test covering all 3 fixes together
- An integration test for the full thinking + tool call conversation flow

I need this to work on both Gemini 2.5 Pro and Gemini 3 Pro but I'm hitting rate limits so often that it's hard to test :/ (I have failing integration tests that hit this). I needed to jump on a different task last week but will get back to it next week. Also happy for anyone to build from there if they have more time.

kyrisu:
Rebased on latest main. The two additional bugs I mentioned in my last update (wrong functionResponse.name and the id field rejected by Vertex AI) have already been fixed on main by #10300, #10400, #10292, and #10343 - so I've dropped those changes from this PR.                                                                     
                                                                                                                                                                                                                                                                                                                                             This PR is now focused solely on its original fix: preserving thought and thoughtSignature on text parts in the legacy converter. Squashed to a single commit.

murphy2mo:
Hello, is there a workaround while the issue is being resolved? We cannot use gemini models right now

paddyfink:
seems like it was fixed in https://github.com/langchain-ai/langchainjs/pull/10415

afirstenberg:
@paddyfink and @murphy2mo - Which library, exactly, are you using where you're seeing this issue and can you provide a code sample that illustrates it? I thought we had resolved this with the `@langchain/google` package a while ago.

PR #10185, opened 2026-02-27
feat: Add custom Vitest matchers for message and agent assertions

maahir30:
## Summary

Adds a set of custom test matchers (`toBeAIMessage`, `toContainToolCall`, `toHaveBeenInterrupted`, etc.) and utility helpers (`firstOfType`, `lastOfType`, `messagesOfType`) for testing LangChain applications.

The matchers live in `@langchain/core/testing/matchers` as pure, framework-agnostic functions — they follow the standard `expect.extend()` contract used by both Vitest and Jest, with no runtime dependency on either. TypeScript users get full autocomplete via a `declare module "vitest"` type augmentation that's compiled into the `.d.ts` output only (zero runtime cost, no vitest peer dep on core).

A convenience re-export at `langchain/testing` gives full-package users a shorter import path.

### Usage

```ts
// vitest.setup.ts (one-time, 3 lines)
import { expect } from "vitest";
import { langchainMatchers } from "langchain/testing";
expect.extend(langchainMatchers);
```

```ts
// any test file — full autocomplete, no extra imports needed
expect(result).toBeAIMessage();
expect(result).toContainToolCall({ name: "get_weather" });
expect(messages).toHaveToolMessages([{ content: "72°F" }]);
```

### What's included

**Matchers:** `toBeHumanMessage`, `toBeAIMessage`, `toBeSystemMessage`, `toBeToolMessage`, `toHaveToolCalls`, `toHaveToolCallCount`, `toContainToolCall`, `toHaveToolMessages`, `toHaveBeenInterrupted`, `toHaveStructuredResponse`

**Helpers:** `firstOfType(messages, AIMessage)`, `lastOfType(messages, HumanMessage)`, `messagesOfType(messages, ToolMessage)`

### Design decisions

- **Source of truth in `@langchain/core`** — every package in the ecosystem depends on core, so the matchers are available to all users (app developers, provider authors, core-only users) with zero extra installs.
- **No vitest/jest dependency on core** — the matcher functions are plain TypeScript. The `declare module "vitest"` block in the source compiles to a type declaration only, creating no runtime link to vitest.
- **Re-exported from `langchain/testing`** — most app developers install `langchain` and would naturally look there first.

### DIFF
--- a/.changeset/unlucky-onions-attend.md
+++ b/.changeset/unlucky-onions-attend.md
@@ -1,0 +1,8 @@
+---
+"@langchain/core": minor
+"langchain": minor
+---
+
+feat: add custom Vitest matchers for LangChain message and tool call assertions
+
+Adds a new `@langchain/core/testing/matchers` export containing custom Vitest matchers (`toBeHumanMessage`, `toBeAIMessage`, `toBeSystemMessage`, `toBeToolMessage`, `toHaveToolCalls`, `toHaveToolCallCount`, `toContainToolCall`, `toHaveToolMessages`, `toHaveBeenInterrupted`, `toHaveStructuredResponse`) that external users can register via `expect.extend(langchainMatchers)` in their Vitest setup files. Re-exported from `langchain` for convenience.


libs/langchain/vitest.setup.ts
```diff
@@ -0,0 +1,26 @@
...
+  toBeAIMessage,
+  toBeSystemMessage,
+  toBeToolMessage,
+  toHaveToolCalls,
+  toHaveToolCallCount,
+  toContainToolCall,
+  toHaveToolMessages,
+  toHaveBeenInterrupted,
+  toHaveStructuredResponse,
+});
```
hntrl: does everyone who want to use these matchers have to do this also?

internal/standard-tests/src/matchers.ts
hntrl, line 1: is there a way we can test that these matchers work appropriately?

libs/langchain/vitest.setup.ts
maahir30: got rid of this, its just one line now via `@langchain/standard-tests/vitest-setup`

maahir30: nope, its just three liens like this 


```ts
// vitest.setup.ts (one-time, 3 lines)
import { expect } from "vitest";
import { langchainMatchers } from "langchain/testing";
expect.extend(langchainMatchers);
```

libs/langchain-core/src/testing/matchers.ts
```diff
@@ -0,0 +1,522 @@
...
+    if (cls.isInstance(msg)) {
+      return msg;
+    }
+  }
+  return undefined;
+}
+
+/**
+ * Returns all messages in the array matching the given message class.
+ * Replaces the pattern: `messages.filter(X.isInstance) as X[]`
```
christian-bromann:
```ts
type foo: X[] = typeof messages.filter(X.isInstance)
```

Should actually work. What does NOT work is `messages.filter((m) => X.isInstance(m))`.

libs/langchain-core/src/testing/matchers.ts
```diff
@@ -0,0 +1,522 @@
...
+  toHaveToolCalls,
+  toHaveToolCallCount,
+  toContainToolCall,
+  toHaveToolMessages,
+  toHaveBeenInterrupted,
+  toHaveStructuredResponse,
+};
+
+export interface LangChainMatchers<R = unknown> {
+  toBeHumanMessage(expected?: string | { content?: string; id?: string }): R;
```
christian-bromann, line 432: Can we tell TypeScript that after the assertion the object should be typed properly?
```suggestion
  toBeHumanMessage(expected?: string | { content?: string; id?: string }): R is HumanMessage;
```

libs/langchain/package.json
```diff
@@ -204,6 +205,17 @@
...
+      "input": "./src/testing/index.ts",
+      "require": {
+        "types": "./dist/testing/index.d.cts",
+        "default": "./dist/testing/index.cjs"
+      },
+      "import": {
+        "types": "./dist/testing/index.d.ts",
+        "default": "./dist/testing/index.js"
+      }
+    },
```
christian-bromann: Let's not do `langchain/testing`. I don't think users optimize for bundle size when building agents, we should just do:

```ts
import { expect } from "vitest";
import { langchainMatchers } from "langchain";

expect.extend(langchainMatchers);
```

libs/langchain/src/testing/index.ts
```diff
@@ -0,0 +1,17 @@
...
+  toBeHumanMessage,
+  toBeAIMessage,
+  toBeSystemMessage,
+  toBeToolMessage,
+  toHaveToolCalls,
+  toHaveToolCallCount,
+  toContainToolCall,
+  toHaveToolMessages,
+  toHaveBeenInterrupted,
+  toHaveStructuredResponse,
```
christian-bromann: Is any of this useful for users?

libs/langchain-core/src/testing/matchers.ts
```diff
@@ -0,0 +1,522 @@
...
+      message: () =>
+        `${utils.matcherHint("toHaveToolCalls")}\n\n` +
+        `Expected ${isNot ? "not " : ""}${expected.length} tool call(s), received ${actual.length}`,
+      actual: actual.length,
+      expected: expected.length,
+    };
+  }
+
+  for (let i = 0; i < expected.length; i++) {
+    const match = Object.entries(expected[i]).every(([key, value]) =>
```
colifran: It looks like we're doing strict positional matching here. It feels like it might make more sense to unordered matching for tool calls since I don't think it is deterministic.

libs/langchain-core/src/testing/matchers.ts
```diff
@@ -0,0 +1,522 @@
...
+      pass: false,
+      message: () =>
+        `${utils.matcherHint("toHaveToolMessages")}\n\n` +
+        `Expected ${isNot ? "not " : ""}${expected.length} tool message(s), found ${toolMessages.length}`,
+      actual: toolMessages.length,
+      expected: expected.length,
+    };
+  }
+
+  for (let i = 0; i < expected.length; i++) {
```
colifran, line 292: Seeing this pattern a lot. We should consider moving this into a reusable function

libs/langchain-core/src/testing/matchers.ts
maahir30: i think its worth keeping these helpers. they are small, readable, and save users from having to know the `filter(X.isInstance)` trick. Can't hurt.

libs/langchain-core/src/testing/matchers.ts
maahir30: thats a good point. would increase flexibility for integration tests. will refactor to that!

libs/langchain-core/vitest.config.ts
```diff
@@ -15,7 +15,7 @@ export default defineConfig((env) => {
       testTimeout: 30_000,
       maxWorkers: 0.5,
       exclude: ["**/*.int.test.ts", ...configDefaults.exclude],
-      setupFiles: ["dotenv/config", "vitest.setup.js"],
+      setupFiles: ["dotenv/config", "vitest.setup.js", "./vitest.setup.matchers.ts"],
```
christian-bromann: Can we move this into `vitest.setup.js`?
```suggestion
      setupFiles: ["dotenv/config", "vitest.setup.js"],
```

libs/langchain/vitest.config.ts
```diff
@@ -20,7 +20,7 @@ export default defineConfig((env) => {
         "**/*.bench.test.ts",
         ...configDefaults.exclude,
       ],
-      setupFiles: ["dotenv/config"],
+      setupFiles: ["dotenv/config", "./vitest.setup.matchers.ts"],
```
christian-bromann: let's not do that as long as we don't use the matchers in any of the tests

PR #10189, opened 2026-02-28
feat(utils): add fakeModel() builder

maahir30:
### Summary

- Adds a `fakeModel()` builder to `@langchain/core/utils/testing` that replaces the need for `FakeToolCallingModel`, `FakeToolCallingChatModel`, and all their custom subclasses with a single fluent API.
- The existing two fake model classes have confusing overlapping APIs, inconsistent `bindTools` behavior, and require 30+ line custom subclasses for common patterns like error injection. The builder eliminates all of that.

### The problem

Tests currently choose between two fake model classes with no guidance on which to use:

- `FakeToolCallingModel` — scripts tool-call sequences per turn, derives content from input
- `FakeToolCallingChatModel` — returns pre-built `BaseMessage` responses verbatim

They have subtle behavioral differences (`bindTools` return types, auto-reset logic, JSON schema support detection) that leak into production code as special cases. Testing "fail N times then succeed" requires a custom 30-line subclass with `vi.fn()` wrapping and unsafe casts — this pattern is repeated 7 times in `modelRetry.test.ts` alone.

### The solution
Add a new `fakeModel()` test utility to `@langchain/core/utils/testing` that provides a builder-pattern API for creating fake chat models with pre-configured responses. This makes it straightforward to script multi-turn agent interactions in unit tests without needing real LLM calls.

- **`fakeModel()`** factory function exported from `@langchain/core/utils/testing`
- **`FakeBuiltModel`** class extending `BaseChatModel` with the following builder methods:
  - `.respond(message | error | factory)` — queue a `BaseMessage`, throw an `Error`, or compute a response dynamically from the input messages
  - `.respondWithTools(toolCalls[])` — queue an `AIMessage` with tool calls (auto-generates IDs if not provided)
  - `.alwaysThrow(error)` — make every invocation throw, overriding the queue
  - `.structuredResponse(value)` — configure the return value for `withStructuredOutput()`
- **Shared state** across `.bindTools()` — response queue, call index, and call recording are preserved
- **Call recording** via `.calls` and `.callCount` for assertions on what the model received
- Comprehensive test suite covering response ordering, mixed sequencing, queue exhaustion, error handling, `bindTools`, and structured output

### DIFF
--- a/libs/langchain-core/src/utils/testing/fake_model_builder.ts
+++ b/libs/langchain-core/src/utils/testing/fake_model_builder.ts
@@ -31,7 +31,10 @@ interface FakeModelCall {
 }
 
 function deriveContent(messages: BaseMessage[]): string {
-  return messages.map((m) => m.text).filter(Boolean).join("-");
+  return messages
+    .map((m) => m.text)
+    .filter(Boolean)
+    .join("-");
 }
 
 let idCounter = 0;


libs/langchain-core/src/utils/testing/fake_model_builder.ts
```diff
@@ -0,0 +1,346 @@
...
+    return "fake-model-builder";
+  }
+
+  _combineLLMOutput() {
+    return [];
+  }
+
+  bindTools(tools: (StructuredTool | ToolSpec)[]) {
+    const merged = [...this.tools, ...tools];
+    const wrapped = formatToolDicts(merged, this.config.toolStyleValue);
```
hntrl: do we need to have this format to tool dicts logic? I'd almost consider this a legacy artifact of the older fake model logic

libs/langchain-core/src/utils/testing/fake_model_builder.ts
```diff
@@ -0,0 +1,346 @@
...
+    this._turns.push(
+      toolCalls.map((tc) => ({
+        name: tc.name,
+        args: tc.args,
+        id: tc.id,
+        type: "tool_call" as const,
+      }))
+    );
+    return this;
+  }
```
hntrl: I think we regard turn differently (though it is kind of a loaded term). Turn generally means that a user will submit a message, then an agent exhausts its tool call queue. Every time a user sends a new message it will be classified as a new turn. I wonder if a more explicit api for this should look like:

```typescript
fakeModel()
  .respondWithToolCalls()
  .respondWithMessage()
  .respondWithError()
``` 
or maybe there's a better shorthand for this?

libs/langchain-core/src/utils/testing/fake_model_builder.ts
```diff
@@ -0,0 +1,346 @@
...
+ * Creates a builder for a fake chat model suitable for testing.
+ *
+ * Two modes — use one or the other, not both:
+ *
+ * `.turn(toolCalls[])` — script tool-call sequences per model invocation.
+ * An empty array means no tool calls (terminal response).
+ * Content is derived from the input messages automatically.
+ *
+ * `.respond(message)` — supply pre-built `BaseMessage` responses.
+ * The model returns them in order.
```
hntrl: i'm admittedly a little bit lost on what the benefit of having these separate are, surely these are analogous? e.g. specifying a `turn` is the same as `responding` with an AIMessage formatted in a certain way

libs/langchain-core/src/utils/testing/fake_model_builder.ts
```diff
@@ -0,0 +1,346 @@
...
+/**
+ * Creates a builder for a fake chat model suitable for testing.
+ *
+ * Two modes — use one or the other, not both:
+ *
+ * `.turn(toolCalls[])` — script tool-call sequences per model invocation.
+ * An empty array means no tool calls (terminal response).
+ * Content is derived from the input messages automatically.
+ *
+ * `.respond(message)` — supply pre-built `BaseMessage` responses.
```
hntrl: wonder if it makes sense to allow either a message instance OR a factory method on these methods, e.g.

```typescript
.respond((input) => new BaseMessage(...))
// or
.respond(new BaseMessage(...))
```

libs/langchain-core/src/utils/testing/tests/fake_model_builder.test.ts
```diff
@@ -0,0 +1,302 @@
...
+
+  description = "A dummy tool";
+
+  schema = z.object({ input: z.string() });
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  async _call(_input: any) {
+    return "dummy result";
+  }
+})();
```
hntrl: y not just use `tool`

libs/langchain-core/src/utils/testing/fake_model_builder.ts
hntrl: ordering of the responses are determined by what order you call .respond with

libs/langchain-core/src/utils/testing/fake_model_builder.ts
```diff
@@ -0,0 +1,346 @@
...
+    this._structuredResponseValue = value;
+    return this;
+  }
+
+  toolStyle(style: ToolStyle): this {
+    this._toolStyleValue = style;
+    return this;
+  }
+
+  build(): FakeBuiltModel {
```
hntrl: if we go the direction of a single path "no modes" do we need to build the fake model? E.g. couldn't I do something like this since things are now declarative?

```typescript
const model = fakeModel()
  .respondWithToolCalls()
  .respondWithMessage()
  .respondWithError()

// no need to build
model.invoke()
```

libs/langchain-core/src/utils/testing/fake_model_builder.ts
maahir30: oh yea true, this formatting isn't really used at all. its dead weight so I'll remove it.

libs/langchain-core/src/utils/testing/fake_model_builder.ts
```diff
@@ -0,0 +1,275 @@
...
+                return (p as { text: string }).text;
+              }
+              return "";
+            })
+            .join("-");
+        } else {
+          return JSON.stringify(part);
+        }
+      })
+      .join("-");
```
hntrl: I think you can just use `message.text`

PR #10197, opened 2026-02-28
fix(google-common): preserve all parallel tool calls in streaming mode

### LINKED ISSUE #10196: responseToChatGeneration drops parallel tool calls during streaming (only uses first part)
### Checked other resources

- [x] The bug is not resolved by updating to the latest stable version of LangChain.js (and related packages).

### Example Code

```typescript
import { ChatVertexAI } from '@langchain/google-vertexai';
import { z } from 'zod';

const searchTool = tool(
    async ({ query }) => `Results for: ${query}`,
    {
        name: 'search',
        description: 'Search the web for information',
        schema: z.object({ query: z.string() }),
    },
);

    model: 'gemini-2.5-flash',
    temperature: 0,
}).bindTools([searchTool]);

const query = 'Search for Mac Mini M4 specs and also search for Mac Mini M4 reviews. Use two separate search calls.';

// ✅ invoke() returns 2 tool_calls
const invokeResult = await model.invoke(query);
console.log('invoke tool_calls:', invokeResult.tool_calls?.length); // 2

// ❌ stream() returns only 1 tool_call
const stream = await model.stream(query);
let accumulated = null;
    accumulated = !accumulated ? chunk : accumulated.concat(chunk);
}
console.log('stream tool_calls:', accumulated?.tool_calls?.length); // 1 ← BUG
```

No error thrown — the second tool call is silently dropped.

### Description

When Gemini returns multiple parallel tool calls (e.g., two `functionCall` parts in a single response), `model.invoke()` correctly returns all tool calls, but `model.stream()` only returns the first one.

**Root cause** is in `@langchain/google-common` at [`libs/providers/langchain-google-common/src/utils/gemini.ts` line 1217](https://github.com/langchain-ai/langchainjs/blob/main/libs/providers/langchain-google-common/src/utils/gemini.ts#L1217):

```typescript
function responseToChatGeneration(response: GoogleLLMResponse): ChatGenerationChunk {
    return new ChatGenerationChunk({
        text: responseToString(response),
        message: partToMessageChunk(responseToParts(response)[0]), // ← BUG: only first part
        generationInfo: responseToGenerationInfo(response),
    });
}
```

`responseToParts(response)` returns all parts (which may contain multiple `functionCall` entries), but `responseToChatGeneration` only passes `[0]` (the first part) to `partToMessageChunk`.

Compare with the non-streaming path (`responseToBaseMessageFields`) which correctly uses **all** parts:

```typescript
function responseToBaseMessageFields(response: GoogleLLMResponse): BaseMessageFields {
    const parts = responseToParts(response);
    return partsToBaseMessageChunkFields(parts); // ← Correct: uses ALL parts
}
```

**Suggested fix**: Change `responseToChatGeneration` to use all parts:

```typescript
function responseToChatGeneration(response: GoogleLLMResponse): ChatGenerationChunk {
    const parts = responseToParts(response);
    const fields = partsToBaseMessageChunkFields(parts);
    const message = new AIMessageChunk(fields);
    return new ChatGenerationChunk({
        text: responseToString(response),
        message,
        generationInfo: responseToGenerationInfo(response),
    });
}
```

### Impact

This affects any LangGraph agent using `streamEvents()` or `stream()` with Gemini models. The `langgraph-api` dev server (used by LangGraph Studio / LangSmith) uses `graph.streamEvents()` internally, so parallel tool calls are silently dropped when running agents through the dev server.

This also means `_generate()` with `streaming: true` is affected, since it calls `_streamResponseChunks` which uses `responseToChatGeneration`.

### System Info

```
@langchain/google-common: 2.1.22
@langchain/google-vertexai: 2.1.22
@langchain/core: 1.1.29
@langchain/langgraph: 1.2.0
Node.js: v20.19.6
Model: gemini-2.5-flash
```

xutruth:
## Description

Fixes #10196

`responseToChatGeneration` in `@langchain/google-common` was only processing the first part from the Gemini streaming response, silently dropping any additional `functionCall` parts. This caused parallel tool calls to be lost when using `model.stream()`, `streamEvents()`, or any streaming-based execution (including `langgraph-api` dev server).

### Root Cause

In `libs/providers/langchain-google-common/src/utils/gemini.ts`, the `responseToChatGeneration` function (used during streaming) was:

```typescript
// BEFORE (bug): only first part processed
message: partToMessageChunk(responseToParts(response)[0])
```

While the non-streaming path (`responseToBaseMessageFields`) correctly processes **all** parts:

```typescript
// Non-streaming (correct): all parts processed
const parts = responseToParts(response);
return partsToBaseMessageChunkFields(parts);
```

### Fix

Changed `responseToChatGeneration` to process all response parts through `partsToBaseMessageChunkFields`, matching the non-streaming behavior.

### Reproduction

```typescript
const model = new ChatVertexAI({ model: 'gemini-2.5-flash' }).bindTools([searchTool]);
const query = 'Search for X and also search for Y';

// ✅ invoke: 2 tool_calls
const result = await model.invoke(query);

// ❌ stream (before fix): 1 tool_call
// ✅ stream (after fix): 2 tool_calls
const stream = await model.stream(query);
```

### Test

Added a unit test `responseToChatGeneration preserves all parallel tool calls` that verifies a response with two `functionCall` parts produces two `tool_calls` in the resulting `ChatGenerationChunk`.

### Impact

This affects:
- Any code using `model.stream()` with Gemini models that return parallel tool calls
- `streamEvents()` in LangGraph agents
- `langgraph-api` dev server (LangGraph Studio / LangSmith playground) which uses `graph.streamEvents()` internally

Made with [Cursor](https://cursor.com)


christian-bromann (approved):
LGTM 👍

libs/providers/langchain-google-common/src/utils/gemini.ts
```diff
@@ -1316,9 +1316,27 @@ export function getGeminiAPI(config?: GeminiAPIConfig): GoogleAIAPI {
   function responseToChatGeneration(
     response: GoogleLLMResponse
   ): ChatGenerationChunk {
+    const parts = responseToParts(response);
+    const fields = partsToBaseMessageChunkFields(parts);
+    const message =
+      typeof fields.content === "string"
+        ? new AIMessageChunk(fields)
+        : fields.content?.every((item) => item.type === "text")
+          ? new AIMessageChunk({
```
christian-bromann, line 1325: Could we add a short comment explaining the intent?

As written, the text-only branch looks like it may be unnecessarily reshaping the response by collapsing multiple text content blocks into a string. I think the intent is to preserve the existing Gemini streaming behavior from `partToMessageChunk`, where text-only chunks are exposed as legacy string content, while still allowing multi-part tool calls to be preserved.

A small comment and ideally a regression test for a multi-text-part streaming response would make that compatibility decision explicit.

christian-bromann (commented):
One minor request before we go ahead and merge.

PR #10204, opened 2026-03-01
feat(core): implement standard schema support for structured output

colifran:
### Summary
Adds Standard Schema support to @langchain/core, enabling providers to accept any Standard Schema compliant schema (e.g. Zod v4, ArkType, Valibot) in withStructuredOutput alongside the existing Zod v3 and plain JSON Schema inputs.

### Changes
- **@standard-schema/spec dependency** - Added to @langchain/core for the StandardSchemaV1 type definitions
- **SerializableSchema type and type guards (utils/standard_schema.ts)** - A schema type that implements both StandardSchemaV1 and StandardJSONSchemaV1, with isStandardSchema, isStandardJsonSchema, and isSerializableSchema guards
- **StandardSchemaOutputParser (output_parsers/standard_schema.ts)** - Output parser that validates parsed output using schema["~standard"].validate()
- **toJsonSchema update (utils/json_schema.ts)** - Handles StandardJSONSchemaV1 by calling schema["~standard"].jsonSchema.input({ target: "draft-07" })
- **JsonOutputKeyToolsParser update** - Accepts serializableSchema for Standard Schema validation of tool call results
- **Shared structured output helpers (language_models/structured_output.ts)** - createContentParser, createFunctionCallingParser, and assembleStructuredOutputPipeline to reduce duplication across providers
- **BaseChatModel.withStructuredOutput update** - Added SerializableSchema overloads and uses shared helpers
- **getSchemaDescription update** - Accepts SerializableSchema in its type signature

### Tests
- Unit tests for isStandardSchema, isStandardJsonSchema, isSerializableSchema
- Unit tests for StandardSchemaOutputParser (valid and invalid inputs)
- Unit tests for toJsonSchema with StandardJSONSchemaV1
- Unit tests for createContentParser and createFunctionCallingParser
- Unit tests for BaseChatModel.withStructuredOutput with Standard Schema

### DIFF
--- a/libs/langchain-core/src/language_models/base.ts
+++ b/libs/langchain-core/src/language_models/base.ts
@@ -27,7 +27,7 @@ import {
   InteropZodType,
 } from "../utils/types/zod.js";
 import { ModelProfile } from "./profile.js";
-import { SerializableSchema } from "../utils/standard_schema.js";
+import { type SerializableSchema } from "../utils/standard_schema.js";
 
 // https://www.npmjs.com/package/js-tiktoken
 
--- a/libs/langchain-core/src/language_models/chat_models.ts
+++ b/libs/langchain-core/src/language_models/chat_models.ts
@@ -63,7 +63,7 @@ import { getEnvironmentVariable } from "../utils/env.js";
 import { castStandardMessageContent, iife } from "./utils.js";
 import {
   isSerializableSchema,
-  SerializableSchema,
+  type SerializableSchema,
 } from "../utils/standard_schema.js";
 import { assembleStructuredOutputPipeline } from "./structured_output.js";
 
--- a/libs/langchain-core/src/language_models/structured_output.ts
+++ b/libs/langchain-core/src/language_models/structured_output.ts
@@ -1,6 +1,7 @@
 import { BaseMessage } from "../messages/index.js";
 import {
   BaseLLMOutputParser,
+  BaseOutputParser,
   JsonOutputParser,
   StructuredOutputParser,
 } from "../output_parsers/index.js";
@@ -35,7 +36,7 @@ export function createContentParser<
     | SerializableSchema<RunOutput>
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     | Record<string, any>
-): BaseLLMOutputParser<RunOutput> {
+): BaseOutputParser<RunOutput> {
   if (isInteropZodSchema(schema)) {
     return StructuredOutputParser.fromZodSchema(schema);
   }


libs/langchain-core/src/language_models/base.ts
```diff
@@ -27,6 +27,7 @@ import {
   InteropZodType,
 } from "../utils/types/zod.js";
 import { ModelProfile } from "./profile.js";
+import { SerializableSchema } from "../utils/standard_schema.js";
```
christian-bromann:
```suggestion
import { type SerializableSchema } from "../utils/standard_schema.js";
```

libs/langchain-core/src/language_models/chat_models.ts
```diff
@@ -63,6 +61,11 @@ import { callbackHandlerPrefersStreaming } from "../callbacks/base.js";
 import { toJsonSchema } from "../utils/json_schema.js";
 import { getEnvironmentVariable } from "../utils/env.js";
 import { castStandardMessageContent, iife } from "./utils.js";
+import {
+  isSerializableSchema,
+  SerializableSchema,
```
christian-bromann:
```suggestion
  type SerializableSchema,
```

libs/langchain-core/src/language_models/structured_output.ts
```diff
@@ -0,0 +1,131 @@
...
+  }
+  if (isSerializableSchema(schema)) {
+    return StandardSchemaOutputParser.fromSerializableSchema(schema);
+  }
+  return new JsonOutputParser<RunOutput>();
+}
+
+type FunctionCallingParserConstructor<
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  T extends Record<string, any> = Record<string, any>,
```
christian-bromann, line 51: if possible `unknown` > `any`
```suggestion
  T extends Record<string, unknown> = Record<string, unknown>,
```

christian-bromann (commented):
Minor nits

libs/langchain-core/src/language_models/base.ts
colifran: Updated!

libs/langchain-core/src/language_models/chat_models.ts
colifran: Updated!

libs/langchain-core/src/language_models/structured_output.ts
colifran: My concern is that `T extends Record<string, any> = Record<string, any>` is pretty well established for `RunOutput` throughout the code base for this (take a look at all the `withStructuredOutput` overloads). I think changing it here might be inconsistent.

christian-bromann: The ideal state should be `Record<string, unknown>` , so if we can we should use `Record<string, unknown>` as much as possible. If changing this requires a too big of a refactor, we can keep as is.

PR #10235, opened 2026-03-02
feat(@langchain/core and @langchain/google): Multimodal Embedding support

afirstenberg:
This is initial support for embeddings in Google, but it also includes backwards compatible changes for embeddings in core to support `ContentBlock`s.

Major core changes:
* Adding `EmbeddingsInterfaceExperimental` which defines the additional methods. The intent is that these methods eventually be part of `EmbeddingsInterface`
* Adding `EmbeddingsExperimental` which has a basic implementation which calls the traditional methods, with the goal that these eventually be part of `Embeddings`. This allows all existing embeddings classes to be forward compatible with the new interface.
* Adding a `MultimodalEmbeddings` class with default implementations for the traditional methods that call the new methods.

Google changes:
* An extension of `MultimodalEmbeddings` which works with both AI Studio embeddings and Vertex AI embeddings for both web and node cases.
* Initial tests for the above.

TODO:
* Additional, more robust, tests
  * Vertex multimodal embedding model
  * Other vertex models
  * Output dimensionality
  * Tests for the core interface/class?
* Documentation
* Automatic schema generation where possible
* Discuss location of some of the generated schema types. Currently, the `Gemini.Part` interface is used from the `chat_models`, which seems like an awkward dependency.
* Discuss the role of the experimental parts in core and best practice there.

### DIFF
--- a/libs/providers/langchain-google/src/embeddings/base.ts
+++ b/libs/providers/langchain-google/src/embeddings/base.ts
@@ -316,11 +316,12 @@ export abstract class BaseGoogleEmbeddings<TOutput = number[]>
         return prediction.textEmbedding ?? [];
       } else if ("imageEmbedding" in prediction) {
         return prediction.imageEmbedding ?? [];
-      } else if ("videoEmbedding" in prediction) {
+      } else if ("videoEmbeddings" in prediction) {
         const videoEmbeddings = (prediction as Vertex.VideoEmbeddings).videoEmbeddings;
         const videoEmbedding = videoEmbeddings?.[0];
         return videoEmbedding?.embedding ?? [];
       } else {
+        console.error(`Could not determine prediction type: ${Object.keys(prediction ?? {})}`);
         return [];
       }
     })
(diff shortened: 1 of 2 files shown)


libs/providers/langchain-google/src/embeddings/ai-typtes-aistudio.ts
hntrl, line 1: `ai-typtes-aistudio.ts`
"typtes"

libs/providers/langchain-google/src/embeddings/index.ts
```diff
@@ -0,0 +1,30 @@
...
+  BaseGoogleEmbeddingsParams,
+  getGoogleEmbeddingsParams,
+} from "./base.js";
+
+export interface GoogleEmbeddingsParams extends BaseGoogleEmbeddingsParams {
+  apiKey?: string;
+  authOptions?: never;
+}
+
+export class GoogleEmbeddings extends BaseGoogleEmbeddings {
```
hntrl, line 14: /** @experimental */

PR #10252, opened 2026-03-03
feat: implement standard schema support for structured output

colifran:
### Summary
Adds Standard Schema support to createAgent's responseFormat parameter. SerializableSchema is accepted alongside Zod and JSON schemas in toolStrategy, providerStrategy, and raw responseFormat inputs, with typed overloads that infer Record<string, unknown> for Standard Schema types.

### Changes
- Added SerializableSchema overloads to ToolStrategy.fromSchema() and ProviderStrategy.fromSchema()
- Added SerializableSchema / SerializableSchema[] overloads to toolStrategy() and providerStrategy() helper functions
- Updated transformResponseFormat() to detect and handle Standard Schema inputs (single and array)
- Updated providerStrategy() options-object disambiguation to check isSerializableSchema

### Tests
- New unit tests for toolStrategy and providerStrategy with SerializableSchema (single, array, end-to-end agent invocation)
- New unit tests for ToolStrategy.fromSchema and ProviderStrategy.fromSchema with SerializableSchema
- New type tests verifying structuredResponse infers Record<string, unknown> for Standard Schema via toolStrategy, providerStrategy, and providerStrategy with options

### DIFF
--- a/libs/langchain/src/agents/responses.ts
+++ b/libs/langchain/src/agents/responses.ts
@@ -327,7 +327,7 @@ export function transformResponseFormat(
   }
 
   /**
-   * If users provide an array, it should only contain raw schemas (Zod or JSON schema),
+   * If users provide an array, it should only contain raw schemas (Zod, Standard Schema or JSON schema),
    * not ToolStrategy or ProviderStrategy instances.
    */
   if (Array.isArray(responseFormat)) {


libs/langchain/src/agents/responses.ts
```diff
@@ -327,6 +343,15 @@ export function transformResponseFormat(
...
 
+    /**
+     * Check if all items are Standard Schema
+     */
+    if (responseFormat.every((item) => isSerializableSchema(item))) {
+      return responseFormat.map((item) =>
+        ToolStrategy.fromSchema(item as SerializableSchema, options)
+      );
+    }
+
```
hntrl, line 354: is this analogous to the responseFormat loop below? we're checking standard schema in both places

colifran: This is part of the "array" path and it calls out specifically:
```
If users provide an array, it should only contain raw schemas (Zod or JSON schema),
not ToolStrategy or ProviderStrategy instances.
```
So in this instance we construct the ToolStrategy off of the raw schema.

Further down we're outside of the "array" path and we construct a ToolStrategy or Provider directly from the responseFormat. It follows the same pattern as the Zod array check above it.

colifran: Actually I'll update the comment since it can now contain Standard Schema as well

PR #10254, opened 2026-03-03
fix(core): preserve thoughtSignature in array content during streaming with thinking models

### LINKED ISSUE #10105: @langchain/google - thought signature differences between AI Studio API and Vertex API
### Privileged issue

### Issue Content

Test file: libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
Test suite: Google Thinking
Test: thought signature - stream
(Other tests in this suite work fine, suggesting this is specific to streaming.)

Models tested:
* gemini-3-pro-preview
* gemini-3-flash-preview
* gemini-3.1-pro-preview

Test config (in the `expansionInfo` variable):
* AI Studio: `{useApiKey: true}`
* Vertex AI: `{node: true}`

For all of the models, this test passes with AI Studio and fails with Vertex AI.
The failure is on the line: `expect(hasThoughtSignature).toBe(true);`

In both cases, I see the `thoughtSignature` in the streamed response, so it isn't immediately obvious what the problem is.

@hntrl - Please assign this to me.

pawel-twardziak:
# Thought signature preserved when streaming from Vertex AI

When streaming responses from thinking models (e.g. Gemini 2.5 Pro/Flash) via Vertex AI, the `thoughtSignature` field is present in the raw streamed response but is lost during chunk concatenation. This causes the "thought signature - stream" integration test to fail with Vertex AI while passing with AI Studio.

## Root cause

- During streaming, `thoughtSignature` arrives in a metadata-only chunk and is stored in `additional_kwargs.originalTextContentBlock`.
- When thinking parts make `message.content` an array (instead of a string), `mergeContent(array, string)` creates a new `{ type: "text", text: string }` object and drops the original chunk’s `thoughtSignature`.
- `ChatGoogleTranslator` in `google.ts` only reads `originalTextContentBlock` when content is a string (lines 12–18). When content is an array (streaming + thinking), it passes the array through and the signature is lost.

## Fix

In the **else** branch of `convertToV1FromChatGoogleMessage` (array content path), check `originalTextContentBlock` for a `thoughtSignature`. If it exists and is not already present in the content array blocks, merge it into the last non-thinking text block.

**Guards:**

- Only runs when `originalTextContentBlock` has a `thoughtSignature`.
- `hasSignatureInContent` avoids adding it twice (covers non-streaming / AI Studio where the signature is already in the content array).
- The signature is merged into the last non-thinking text block, matching how the API sends it with the answer text.
- If no suitable block is found, falls through to the default `return message.content`.

## Tests

Added `google.test.ts` with 5 unit tests for `ChatGoogleTranslator`:

1. **String content with `originalTextContentBlock`** - existing behavior.
2. **Array content with `thoughtSignature` already in content** - no double-add.
3. **Array content with `thoughtSignature` only in `originalTextContentBlock`** - the bug fix.
4. **Array content with no `thoughtSignature`** - no-op.
5. **Array content with only thinking blocks** - edge case.

Fixes #10105

### DIFF
--- a/.changeset/shiny-timers-wave.md
+++ b/.changeset/shiny-timers-wave.md
@@ -1,0 +1,5 @@
+---
+"@langchain/core": patch
+---
+
+fix(core): preserve thoughtSignature in array content during streaming with thinking models


afirstenberg:
Excellent! Thank you.

Do we have a good integration test case for this that illustrates the problem that we can test for the fix?

christian-bromann:
@pawel-twardziak waiting to merge until we address @afirstenberg question.

pawel-twardziak:
hi @afirstenberg  and @christian-bromann 

sorry, the comment slipped my attention. 
@christian-bromann could you confirm the integration tests are being run against Vertex AI?

pawel-twardziak:
I've connected to Vartex AI and for this

image

I'm getting this;

Screenshot from 2026-03-06 13-19-18

afirstenberg:
LGTM
The code looks reasonable and I've run the integration tests and confirmed the resolve the issue.

@christian-bromann @hntrl - Let's get this in ASAP. I'll talk to Hunter if there are others we should make sure are in, and let's get this released.

@pawel-twardziak - My apologies. I think when I originally posted about integration tests, I didn't realize that the original ones were the ones failing. They didn't when we originally released. Thank you for the patch!

Godrules500:
Do we know when this will be moved into production and what the version will be?

PR #10255, opened 2026-03-03
Feature/mongodb voyage endpoint

RaschidJFR:
Expose `basePath` in VoyageEmbeddings constructor

Fixes the issue where `basePath` exists as a class property but cannot be configured during instantiation. This creates confusion as users might try to modify `basePath` after construction, which has no effect since `apiUrl` is only constructed once in the constructor.

Changes:
- Add `basePath` to the constructor parameter types

Fixes langchain-ai/langchainjs-community#20

### DIFF
--- a/.changeset/funky-hoops-fold.md
+++ b/.changeset/funky-hoops-fold.md
@@ -1,0 +1,5 @@
+---
+"@langchain/community": minor
+---
+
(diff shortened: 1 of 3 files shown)


RaschidJFR:
Closed in favour of langchain-ai/langchainjs-community#41

PR #10262, opened 2026-03-03
fix: Move fakeModel from utils/testing to testing namespace

maahir30:
## Summary

- Moved `fakeModel` (the queue-based fake chat model builder) from `@langchain/core/utils/testing` to the newer `@langchain/core/testing` namespace, alongside the vitest matchers.
- Added `./testing/fake_model_builder` as a new entrypoint in `@langchain/core` (package.json exports + tsdown build config).
- Re-exported `fakeModel` from the `langchain` package to match how `langchainMatchers` is already surfaced.


libs/langchain-core/tsdown.config.ts
```diff
@@ -73,6 +73,7 @@ export default getBuildConfig({
     "./src/utils/testing/index.ts",
     "./src/utils/tiktoken.ts",
     "./src/utils/types/index.ts",
+    "./src/testing/fake_model_builder.ts",
     "./src/testing/matchers.ts",
```
hntrl: do we want to have a separate entrypoint for this? can we just export matchers + fake model builder from an index barrel that exports both?

christian-bromann: 💯

PR #10263, opened 2026-03-04
fix(anthropic): properly transform reasoning tokens into content blocks

christian-bromann:
## Summary

- **Fix**: Anthropic's `thinking` / `redacted_thinking` content blocks are now properly converted to LangChain's standard `ContentBlock.Reasoning` format (`type: "reasoning"`) in both streaming and non-streaming response paths, instead of being passed through as raw Anthropic-native blocks (`type: "thinking"`).
- **Core type update**: Added `signature` and `redacted` fields to `ContentBlock.Reasoning` to support Anthropic's cryptographic signature (needed for round-tripping thinking blocks back to the API) and redacted thinking blocks.
- **Round-trip support**: Updated `_formatStandardContent` to correctly convert `redacted` reasoning blocks back to Anthropic's `redacted_thinking` format when sending messages to the API.

## Problem

When using Anthropic models with extended thinking enabled, the `AIMessageChunk.content` array contained Anthropic-native content blocks:

```json
{ "type": "thinking", "thinking": "Let me reason...", "signature": "sig_..." }
```

These did not match the standard `ContentBlock.Reasoning` interface, which expects:

```json
{ "type": "reasoning", "reasoning": "Let me reason...", "signature": "sig_..." }
```

This meant downstream code checking for `type === "reasoning"` would never match Anthropic thinking blocks.

### DIFF
--- a/.changeset/quiet-avocados-admire.md
+++ b/.changeset/quiet-avocados-admire.md
@@ -1,0 +1,5 @@
+---
+"@langchain/anthropic": patch
+---
+
+fix(anthropic): properly transform reasoning tokens into content blocks


libs/langchain-core/src/messages/content/index.ts
```diff
@@ -104,6 +104,18 @@ export declare namespace ContentBlock {
...
+     *
+     * Some providers (e.g. Anthropic) include a cryptographic signature
+     * that must be preserved when round-tripping thinking blocks back
+     * to the API.
+     */
+    signature?: string;
+    /**
+     * Whether the reasoning content was redacted by the provider.
+     */
+    redacted?: boolean;
```
hntrl, line 118: mmm I think we decided we intentionally didn't want to add this to the standard

libs/providers/langchain-anthropic/src/utils/message_outputs.ts
```diff
@@ -139,13 +146,31 @@ export function _makeMessageChunkFromAnthropicEvent(
...
+                index: data.index,
+                type: "reasoning" as const,
+                reasoning: contentBlock.thinking,
+              },
+            ] satisfies ContentBlock.Reasoning[],
+            response_metadata,
+          }),
+        };
+      }
+      if (contentBlock.type === "signature_delta") {
```
hntrl, line 163: this is a breaking change

PR #10292, opened 2026-03-07
fix(@langchain/google) Fixes for ID in Vertex function calls

### LINKED ISSUE #10266: bug(@langchain/google) Tests with functions on Vertex fail
### Privileged issue

### Issue Content

Two of the integration tests fail on Vertex AI for Gemini 3 models:
* function conversation
* function reply

with the error from Gemini:
```
Invalid JSON payload received. Unknown name "id" at 'contents[2].parts[0].function_response': Cannot find field.
```

afirstenberg:
Vertex does not accept IDs for function responses, considering it an invalid field. LangChain requires an ID to match responses to requests, so a previous fix added a synthetic ID for this purpose.
Changed this synthetic ID to match a known pattern so it could be removed when sending the results back to Gemini.

Added tests.
Integration tests pass with the following exception:
* gemini-3.1-flash-lite is generating an error about the function results. This appears to be an error on the Google side, and I'm researching further and will coordinate with Google.

Fixes #10266

### DIFF
--- a/libs/providers/langchain-google/src/converters/messages.ts
+++ b/libs/providers/langchain-google/src/converters/messages.ts
@@ -56,13 +56,13 @@ export const geminiContentBlockConverter: StandardContentBlockConverter<{
           typeof block.data === "string"
             ? block.data
             : typeof block.data === "object" && block.data !== null
-              ? // Convert Uint8Array to base64 string
-                btoa(
-                  Array.from(block.data as Uint8Array)
-                    .map((byte) => String.fromCharCode(byte))
-                    .join("")
-                )
-              : String(block.data);
+            ? // Convert Uint8Array to base64 string
+              btoa(
+                Array.from(block.data as Uint8Array)
+                  .map((byte) => String.fromCharCode(byte))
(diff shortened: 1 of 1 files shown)


libs/providers/langchain-google/src/converters/messages.ts
```diff
@@ -929,7 +931,7 @@ export const convertGeminiPartsToToolCalls: Converter<
       const functionCallPart = part as Gemini.Part.FunctionCall;
       toolCalls.push({
         type: "tool_call",
-        id: functionCallPart.functionCall.id ?? uuidv4().replace(/-/g, ""),
+        id: functionCallPart.functionCall.id ?? `lc-tool-call-${uuidv4().replace(/-/g, "")}`,
```
hntrl: and i'm guessing that second condition will only ever be met by vertex?

afirstenberg: The second condition should only be met if it uses a generated ID. But it is aimed at when it happens on Vertex, yes.

PR #10313, opened 2026-03-09
fix(langchain-google): parse JSON error bodies from non-JSON content types

### LINKED ISSUE #10306: [@langchain/google] RequestError.fromResponse discards error body from streaming endpoint
## Bug

When `ChatGoogle` with `platformType: 'gcp'` hits a 400 on the streaming endpoint (`?alt=sse`), the error message is a useless `"Request failed with status code 400"`. The actual Google API error (e.g. `"Unknown name \"const\""`) is discarded.

## Root cause

In [`libs/providers/langchain-google/src/utils/errors.ts`](https://github.com/langchain-ai/langchainjs/blob/main/libs/providers/langchain-google/src/utils/errors.ts), `RequestError.fromResponse()` checks `content-type` for `application/json` before parsing as JSON:

```ts
if (contentType.includes("application/json")) return await response.json();
return await response.text();
```

The streaming endpoint returns errors with a non-JSON content-type (likely `text/event-stream`), so it falls back to `.text()`. Then `errorBody?.error?.message` fails because `errorBody` is a string, not an object — it falls through to the generic `"Request failed with status code ${response.status}"`.

The same issue exists in `AuthError.fromResponse()`.

## Suggested fix

After the `.text()` fallback, attempt `JSON.parse`:

```ts
const text = await response.text();
try { return JSON.parse(text); } catch { return text; }
```

If the text is valid JSON (as Google's error responses are, regardless of content-type), it gets parsed and `.error.message` extraction works. Otherwise, the raw text is preserved.

## Reproduction

1. Define a tool with a Zod schema using `z.literal()` (produces `{"const": ...}` in JSON Schema, which Vertex AI rejects)
2. Call `ChatGoogle` with `platformType: 'gcp'` and that tool
3. Observe error: `"Request failed with status code 400"`
4. Apply the fix above
5. Observe error now shows: `"Invalid JSON payload received. Unknown name \"const\"..."`

## Environment

- `@langchain/google`: 0.1.5
- Platform: Vertex AI (GCP)
- Node.js 22

pawel-twardziak:
Fix: Parse JSON error bodies from non-JSON content types in Google provider
                                                                                                                                                                                                                                                                                                           
When GCP returns error responses for streaming requests, the response carries a text/event-stream content type even though the body is JSON. The previous implementation only attempted response.json() when the content type was application/json, falling back to response.text() otherwise. This caused error messages to be a generic "Request failed with status code 400" instead of surfacing the actual error detail from the JSON body.                                                                                                                                                             
                  
This PR:                                                                                                                                                                                                                                                                                                 
- Extracts a shared readErrorResponseBody helper used by both RequestError.fromResponse and AuthError.fromResponse
- The helper always attempts JSON.parse on text bodies when the content type isn't application/json, so JSON error payloads are correctly parsed regardless of content type                                                                                                                              
- Adds unit tests for RequestError and AuthError response parsing                                                                                                          
- Adds an integration-level mock test that reproduces the streaming error scenario end-to-end                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                           
Fixes #10306

### DIFF
--- a/.changeset/nice-berries-invent.md
+++ b/.changeset/nice-berries-invent.md
@@ -1,0 +1,5 @@
+---
+"@langchain/google": patch
+---
+
+fix(langchain-google): parse JSON error bodies from non-JSON content types


pawel-twardziak:
hi @hntrl - is this one legit? :)

christian-bromann (approved):
LGTM 👍

PR #10316, opened 2026-03-09
fix(google): Add withStructuredOutput overload signatures for type narrowing

maahir30:
## Summary

- Adds overload signatures to `BaseChatGoogle.withStructuredOutput()` so TypeScript narrows the return type based on `includeRaw`, matching the pattern already used by `@langchain/openai`, `@langchain/anthropic`, `@langchain/openrouter`, and other providers.
- Without overloads, the return type is always a union (`Runnable<I, RunOutput> | Runnable<I, { raw, parsed }>`), which causes `prompt.pipe(model.withStructuredOutput(schema))` to fail with a type error even when `includeRaw` is `false` or omitted.


hntrl (approved):
good catch, thanks

PR #10345, opened 2026-03-11
fix(google-genai): merge consecutive function-role messages in convertBaseMessagesToContent

### LINKED ISSUE #10342: Bug: @langchain/google-genai creates separate turns for consecutive ToolMessages, breaking Gemini parallel function calling
## Description

`convertBaseMessagesToContent` in `@langchain/google-genai` has an off-by-one bug that causes consecutive `ToolMessage`s to create separate "user" content entries instead of merging them into a single turn. The Gemini API requires all parallel function responses in a single user turn — separate turns cause the API to hang indefinitely.

## Bug Location

`libs/langchain-google-genai/src/utils/common.ts` in `convertBaseMessagesToContent`:

```js
const prevContent = acc.content[acc.content.length]; // BUG: always undefined (should be length - 1)
if (!acc.mergeWithPreviousContent && prevContent && prevContent.role === role)
  throw new Error("Google Generative AI requires alternate messages between authors");
```

`acc.content[acc.content.length]` is always `undefined` because arrays are 0-indexed. The duplicate-role check never fires, so consecutive ToolMessages silently create separate content entries with role "user".

## Expected Behavior

Consecutive ToolMessages (role "function" → "user") should be merged into a single content entry with multiple `functionResponse` parts, matching the format Gemini expects for parallel function call responses. This is analogous to how `@langchain/anthropic` merges consecutive tool_result messages in its `mergeMessages` function.

## Actual Behavior

Each ToolMessage creates its own content entry. When the model returns N parallel function calls and receives N separate "user" turns back, the Gemini API hangs indefinitely (no error, no response).

## Reproduction

1. Use `ChatGoogleGenerativeAI` with `gemini-3-flash-preview` and `bindTools()`
2. Send a prompt that triggers 3+ parallel tool calls
3. Return results as separate `ToolMessage` instances
4. Call `model.invoke(messages)` again — the API hangs

## Suggested Fix

```js
const prevContent = acc.content[acc.content.length - 1]; // fix off-by-one
if (!acc.mergeWithPreviousContent && prevContent && prevContent.role === role) {
  if (role === "function") {
    // Merge consecutive function responses into single turn
    const parts = convertMessageContentToParts(message, isMultimodalModel, messages.slice(0, index), model);
    prevContent.parts.push(...parts);
    return acc;
  }
  throw new Error("Google Generative AI requires alternate messages between authors");
}
```

## Environment

- `@langchain/google-genai`: 2.1.21
- `@langchain/core`: latest
- Model: `gemini-3-flash-preview`
- Node.js 20

mossein:
Fixes #10342

When multiple tool calls are executed in parallel, `convertBaseMessagesToContent` receives consecutive `ToolMessage`s that all map to the `function` role. The existing code throws "Google Generative AI requires alternate messages between authors" instead of merging them.

This patch:

1. Fixes an off-by-one bug (`acc.content[acc.content.length]` → `acc.content[acc.content.length - 1]`) that prevented the duplicate-role check from ever triggering.
2. When consecutive `function`-role messages are detected, merges their `functionResponse` parts into the previous content entry instead of throwing. Gemini expects all `functionResponse` parts for a given model turn to live in a single user turn.

### DIFF
--- a/.changeset/long-news-draw.md
+++ b/.changeset/long-news-draw.md
@@ -1,0 +1,5 @@
+---
+"@langchain/google-genai": patch
+---
+
+merge consecutive function-role messages in convertBaseMessagesToContent


mossein:
cc @hntrl @jacoblee93,  this fixes a bug that causes Gemini to hang when LangChain sends parallel tool results as separate user turns instead of merging them into one. Happy to adjust anything if needed.

pawel-twardziak:
hi @mossein 

great work! I think it would be good to use `actualRole` in the merge condition, e.g. `compute actualRole` before the `prevContent` check and use `prevContent.role === actualRole` and then `role === "function"` to merge

and add unit tests

mossein:
> hi Mohammad Jafari (@mossein)
> 
> great work! I think it would be good to use `actualRole` in the merge condition, e.g. `compute actualRole` before the `prevContent` check and use `prevContent.role === actualRole` and then `role === "function"` to merge
> 
> and add unit tests

Good call @pawel-twardziak. I moved the actualRole computation before the prevContent check so the comparison uses the mapped API role.
Also added unit tests covering parallel tool calls with 2 and 3 consecutive ToolMessages. Let me know if anything else needs adjusting.

pawel-twardziak:
LGTM, thanks @mossein :muscle:

mossein:
Hey @christian-bromann - this fixes the same class of bug as #10343 (which was merged) but for the `@langchain/google-genai` package. Would love a review when you get a chance!

PR #10402, opened 2026-03-12
fix(google): use dynamic ls_provider based on platform instead of hardcoded "google"

### LINKED ISSUE #10401: ChatGoogle reports ls_provider as "GoogleNode" instead of "google_vertexai" when using vertexai: true
### Checked other resources

### Example Code

 ```typescript
  import { StateGraph, START, END, Annotation } from
  "@langchain/langgraph";

    model: "gemini-2.5-flash",
    vertexai: true,
  });

  const graph = new StateGraph(Annotation.Root({}))
    .addNode("node", async () => {
      await model.invoke("Hello");
    })
    .addEdge(START, "node")
    .addEdge("node", END)
    .compile();

  for await (const event of graph.streamEvents({}, { version: "v2" })) {
    if (event.event === "on_chat_model_end") {
      console.log(event.metadata.ls_provider);   // "GoogleNode"  ←
  expected: "google_vertexai"
      console.log(event.metadata.ls_model_name); // ""            ←
  expected: "gemini-2.5-flash"
    }
  }

<img width="229" height="79" alt="Image" src="https://github.com/user-attachments/assets/0917e0dc-cacf-4afc-8bb9-ba595ed31c01" />

Where openai has a model name as well

<img width="221" height="70" alt="Image" src="https://github.com/user-attachments/assets/d389384b-fc9d-4dfb-87a3-0643897ce72c" />

### Description

When using `ChatGoogle` from `@langchain/google/node` with `vertexai:  
  true`, the `ls_provider` field in LangSmith stream events (via         
  `streamEvents`) is reported as `"GoogleNode"` instead of               
  `"google_vertexai"`. Additionally, `ls_model_name` is empty, making it 
  impossible to identify which model was used.                           
                                                                         
  This causes incorrect or missing provider/model identification in
  LangSmith traces, and breaks any cost-tracking logic that relies on
  these fields.

  In contrast, the old `ChatVertexAI` from `@langchain/google-vertexai`
  correctly reported `"google_vertexai"` and the model name. The
  `"GoogleNode"` value appears to derive from the module path
  (`@langchain/google/node`) rather than the runtime `vertexai: true`
  configuration.

  **Summary of missing fields:**

  | Field | Expected | Actual |
  |---|---|---|
  | `ls_provider` | `"google_vertexai"` | `"GoogleNode"` |
  | `ls_model_name` | e.g. `"gemini-2.5-flash"` | `""` (empty) |

  **Impact on cost tracking:** Because both fields are missing/incorrect,
   any downstream cost tracking that relies on `streamEvent.metadata`
  from `on_chat_model_end` events cannot determine which model was used.
  This forces consumers to implement hardcoded fallbacks, which produces
  incorrect cost calculations when using models other than the fallback
  (e.g. `gemini-2.5-pro` being billed as `gemini-2.5-flash`).

### System Info

 - `@langchain/google` version: `0.1.2`
  - `@langchain/langgraph` version: `1.1.5`                              
  - `langsmith` version: `0.5.6`                                         
  - Node.js version: `v22.21.1`

pawel-twardziak:
Fixes #10401

## Description

ChatGoogle from `@langchain/google` hardcodes `ls_provider: "google"` in `getLsParams()`, regardless of whether `vertexai: true` is set. This causes incorrect provider identification in LangSmith traces and breaks cost-tracking logic that relies on `ls_provider` and `ls_model_name` metadata.

The class already has a platform getter that returns `"gcp"` when `vertexai: true` - it just wasn’t being used in `getLsParams()`.

## Changes

- **base.ts:** Set `ls_provider` from a dynamic value based on `this.platform` instead of hardcoding `"google"`:
  - `"gcp"` → `"google_vertexai"`
  - `"gai"` → `"google_genai"`
- **index.test.ts:** Updated existing test expectations and added a test for `vertexai: true` to assert the correct provider string.

| Platform | Before | After |
|----------|--------|-------|
| Vertex AI (`vertexai: true`) | `"google"` | `"google_vertexai"` |
| Google AI (default) | `"google"` | `"google_genai"` |

### DIFF
--- a/.changeset/rich-guests-vanish.md
+++ b/.changeset/rich-guests-vanish.md
@@ -1,0 +1,5 @@
+---
+"@langchain/google": patch
+---
+
+fix(google): use dynamic ls_provider based on platform instead of hardcoded "google"


afirstenberg:
Is this the right approach, or would it be better to change LangSmith to handle "google"? (The pricing on the two platforms is supposed to be the same.)

I guess the question is - what is `ls_provider` supposed to represent?

pawel-twardziak:
Fair point. The fix belongs here though, not in LangSmith imho.

1. existing convention

The other Google packages already use distinct provider strings:
- @langchain/google-vertexai -> "google_vertexai"
- @langchain/google-genai -> "google_genai"
The fix aligns @langchain/google with that established pattern

2. ls_provider represents the API platform, not just the vendor

Vertex AI and Google AI (Generative AI) are  different API endpoints with different auth, quotas rate limits, and regions. Even if model pricing is the same today, the platform distinction matters for observability (e.g., "which API am I hitting?")

3. Fixing LangSmith would be a workaround

Making LangSmith map "google" -> something useful would require LangSmith to guess which platform you're using, which it can't do from just "google". The metadata should be  accurate at the source.

4. pricing isn't always identical

Vertex AI can have different pricing tiers depending on committed use  discounts, enterprise agreements, and region. Collapsing them breaks cost tracking for those users.

5. from a software design perspective

- the data should be accurate at the source
- it also follows the existing pattern
- your suggestion would create a coupling problem

If LangSmith had to map "google" -> the right platform, it would need knowledge about SDK internals (which package, which config). That's a layering violation - the tracing layer shouldn't know about SDK construction details.

**Right data at the right layer, consistent with existing conventions, no new coupling. That's why I consider my fix clean.**

if you have good counterarguments, I'll gladly listen :)

PR #10405, opened 2026-03-12
fix(langchain): add browser export

christian-bromann:
Instead of fixing our dependency to async_hooks , this may resolve the issue of importing messages via `langchain` in browser code.

### DIFF
--- a/libs/langchain/package.json
+++ b/libs/langchain/package.json
@@ -106,6 +106,7 @@
   "types": "./dist/index.d.cts",
   "exports": {
     ".": {
+      "browser": "./dist/browser.js",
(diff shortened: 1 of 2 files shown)


libs/langchain/package.json
```diff
@@ -116,6 +117,17 @@
         "default": "./dist/index.js"
       }
     },
+    "./browser": {
```
christian-bromann, line 120: This is auto-generated based on the new entry point I added to the tsdown config. We could make this configurable but this seems to just add complexity. Hence, I kept it as is.

hntrl (approved):

PR #10407, opened 2026-03-12
fix(google): lazy-load jose in CJS auth helpers

### LINKED ISSUE #10403: @langchain/google CJS entry points fail at runtime: require("jose") in gcp-auth.cjs breaks because jose v6 is ESM-only
### Checked other resources

### Example Code

```typescript
mkdir repro && cd repro
npm init -y
npm install @langchain/google @langchain/core
# package.json has no "type": "module", so .js = CJS
echo 'const { ChatGoogle } = require("@langchain/google/node");' > index.js
node index.js
# → ERR_REQUIRE_ESM
```

Error [ERR_REQUIRE_ESM]: require() of ES Module .../jose/dist/webapi/index.js
from .../dist/utils/gcp-auth.cjs not supported.
Instead change the require of index.js in .../@langchain/google/dist/utils/gcp-auth.cjs
to a dynamic import() which is available in all CommonJS modules.

### Description

**Description**
`dist/utils/gcp-auth.cjs` line 3 does:

`let jose = require("jose");`

But `jose@^6.1.3` (declared in `@langchain/google`'s dependencies) is ESM-only — it has no CJS exports. This crashes at runtime in any CJS context: Next.js Pages Router (webpack), Jest, NestJS, or any TypeScript project without `"type": "module"`.

This is the same class of issue as #8132 (which was about `@langchain/google-webauth` / `web-auth-library`). The new unified `@langchain/google@0.1.5` package inherited the problem via its own `gcp-auth.cjs` which uses `jose` directly.

**Suggested fix**
The functions that use `jose` (`getGCPPrivateKey`, `getGCPCustomToken`, `getGCPCredentialsAccessToken`) are already async, so replacing `require("jose")` with a lazy `await import("jose")` in the CJS output should be a straightforward fix.

**Workaround**
Use dynamic import on the consumer side to force ESM resolution:

`const { ChatGoogle } = await import("@langchain/google/node");`

### System Info

@langchain/google@0.1.5
@langchain/core@1.1.31
jose@6.1.3
Node.js v22.14.0

fahe1em1:
## Summary
- lazy-load `jose` inside the GCP auth helpers so the CJS bundle no longer eagerly `require()` an ESM-only dependency
- keep the ESM path unchanged while fixing CommonJS consumers of `@langchain/google/node`
- add focused regression tests that prove the helper module loads without importing `jose` until a signing helper is actually used

Fixes #10403.

## Why this is safe
This only changes when `jose` is loaded. The affected helpers were already async, so switching them to fetch `jose` lazily preserves behavior while avoiding the `ERR_REQUIRE_ESM` crash in CommonJS runtimes.

## Validation
- `pnpm --filter @langchain/google test -- src/utils/tests/gcp-auth.test.ts`
- `pnpm --filter @langchain/google build`
- `pnpm --filter @langchain/google lint`

### DIFF
--- a/libs/providers/langchain-google/src/utils/tests/gcp-auth.test.ts
+++ b/libs/providers/langchain-google/src/utils/tests/gcp-auth.test.ts
@@ -10,7 +10,9 @@ describe("gcp-auth", () => {
     const { normalizeGCPCredentials } = await import("../gcp-auth.js");
 
     expect(
-      normalizeGCPCredentials('{"type":"service_account","project_id":"p","private_key_id":"k","private_key":"pem","client_id":"c","client_email":"a@b.c","auth_uri":"https://auth","token_uri":"https://token","auth_provider_x509_cert_url":"https://cert","client_x509_cert_url":"https://client-cert"}')
+      normalizeGCPCredentials(
+        '{"type":"service_account","project_id":"p","private_key_id":"k","private_key":"pem","client_id":"c","client_email":"a@b.c","auth_uri":"https://auth","token_uri":"https://token","auth_provider_x509_cert_url":"https://cert","client_x509_cert_url":"https://client-cert"}'
+      )
     ).toMatchObject({
       project_id: "p",
       private_key_id: "k",


afirstenberg:
LGTM

pawel-twardziak:
great work @fahe1em1 !

PR #10408, opened 2026-03-12
fix(agents): continue after blocked tool calls

### LINKED ISSUE #10090: Empty response if tool call limit exceeded
### Checked other resources

### Example Code

```typescript
    const agent = createAgent({
      contextSchema: chatAgentContextSchema,
      model: generationModel,
      name: "Chat",
      systemPrompt: new SystemMessage(systemMessageBlocks),
      tools: [retrievalTool, scopedMaterialRetrievalTool],
      middleware: [
        summarizationMiddleware({
          model: generationModel,
          trigger: { tokens: MAX_MESSAGE_TOKENS },
        }),
        toolCallLimitMiddleware({
          runLimit: 3,
          toolName: retrievalTool.name,
        }),
        toolCallLimitMiddleware({
          runLimit: 3,
          toolName: scopedMaterialRetrievalTool.name,
        }),
      ],
    })
```

Looking at langfuse traces they end up with this as the last message in the response.

```json
{
  "lc": 1,
  "type": "constructor",
  "id": [
    "langchain_core",
    "messages",
    "ToolMessage"
  ],
  "kwargs": {
    "content": "Tool call limit exceeded. Do not call 'search_course_materials' again.",
    "tool_call_id": "call_zO0aoz9qGaGWCOyiFqAoGhBv",
    "name": "search_course_materials",
    "status": "error",
    "additional_kwargs": {},
    "response_metadata": {},
    "id": "fd17c226-295a-42d7-b9b9-ad0db920dee7"
  }
}
```

### Description

- When the tool call limit is exceeded the agent just stops instead of providing an answer to the user. There's no error, the last message in the stack (as seen above in the trace from langfuse) only has the last tool call.
- When the tool call limit is not exceeded the user gets a normal AI answer.

I would expect the LLM to still answer after the tool call limit was reached.

---

I tried to find the issue in our codebase together with Claude and it pointed this out. Maybe this helps.

> The bug is in LangChain's ReactAgent router (#createAfterModelRouter). When toolCallLimitMiddleware with "continue" behavior blocks all tool calls in an AI message, it injects error ToolMessages. The router then:
> 
> 1. Computes pendingToolCalls.length === 0 (all tool_call_ids have matching ToolMessages)
> 2. Tries step 5: pendingToolCalls.length === 0 && hasStructuredResponse — fails because hasStructuredResponse is false (no responseFormat configured)
> 3. Falls through to step 6: lastMessage is a ToolMessage (not AIMessage) → returns exitNode (END)
> 
> The agent exits without calling the model again to generate a text response. The legacy createReactAgent from @langchain/langgraph handles this correctly with if (lastMessage && isToolMessage(lastMessage)) return entrypoint, but the newer ReactAgent is missing this check.

### System Info

- All plattforms
- Node: v25.6.1
- Bun: 1.3.9

```
langchain@1.2.25 | MIT | deps: 5 | versions: 389

dependencies (5):
- @langchain/langgraph: ^1.1.2
- @langchain/langgraph-checkpoint: ^1.0.0
- langsmith: >=0.5.0 <1.0.0
- uuid: ^10.0.0
- zod: ^3.25.76 || ^4
```

```
@langchain/core@1.1.26 | MIT | deps: 10 | versions: 291
Core LangChain.js abstractions and schemas
https://github.com/langchain-ai/langchainjs/tree/main/langchain-core/

dependencies (10):
- @cfworker/json-schema: ^4.0.2
- ansi-styles: ^6.2.3
- camelcase: 6
- decamelize: 1.2.0
- js-tiktoken: ^1.0.12
- langsmith: >=0.5.0 <1.0.0
- mustache: ^4.2.0
- p-queue: ^6.6.2
- uuid: ^10.0.0
- zod: ^3.25.76 || ^4
```

fahe1em1:
## Summary
- route back to the model when after-model middleware injects only `ToolMessage`s and there are no tool calls left to execute
- keep the existing pending-tool-call routing intact so allowed tool calls still run before we re-enter the model
- update the tool call limit regression to assert the agent produces a final AI response after blocked tool calls

Fixes #10090

## Validation
- `pnpm exec vitest run src/agents/middleware/tests/toolCallLimit.test.ts`
- `pnpm exec eslint src/agents/ReactAgent.ts src/agents/middleware/tests/toolCallLimit.test.ts`

### DIFF
--- a/libs/langchain/src/agents/ReactAgent.ts
+++ b/libs/langchain/src/agents/ReactAgent.ts
@@ -935,6 +935,13 @@ export class ReactAgent<
         return AGENT_NODE_NAME;
       }
 
+      // afterModel middleware can inject ToolMessages (for example, when it blocks
+      // tool calls). Once every pending tool call has a corresponding ToolMessage,
+      // route back to the model so it can respond instead of exiting early.
+      if (ToolMessage.isInstance(lastMessage)) {
+        return AGENT_NODE_NAME;
+      }
+
       if (
         !AIMessage.isInstance(lastMessage) ||
         !lastMessage.tool_calls ||
--- a/libs/langchain/src/agents/middleware/tests/toolCallLimit.test.ts
+++ b/libs/langchain/src/agents/middleware/tests/toolCallLimit.test.ts
@@ -676,7 +676,7 @@ describe("toolCallLimitMiddleware", () => {
       }
     });
 
-    it("should run remaining tools until limit is exceeded", async () => {
+    it("should return to the model when remaining tool calls are blocked", async () => {
       const middleware = toolCallLimitMiddleware({
         threadLimit: 3,
         runLimit: 2,
@@ -697,7 +697,7 @@ describe("toolCallLimitMiddleware", () => {
             content: "",
             tool_calls: [{ id: "4", name: "search", args: { query: "test3" } }],
           }),
-          new AIMessage("Should not reach here"),
+          new AIMessage("Final response after tool limit"),
         ],
       });
 
@@ -712,8 +712,16 @@ describe("toolCallLimitMiddleware", () => {
       });
 
       const lastMessage = result.messages[result.messages.length - 1];
-      expect(lastMessage.content).toContain(
-        "Tool call limit exceeded. Do not make additional tool calls."
+      expect(AIMessage.isInstance(lastMessage)).toBe(true);
+      expect(lastMessage.content).toBe("Final response after tool limit");
+
+      const blockedToolMessages = result.messages.filter(
+        (message): message is ToolMessage =>
+          ToolMessage.isInstance(message) && message.status === "error"
+      );
+      expect(blockedToolMessages).toHaveLength(2);
+      expect(blockedToolMessages[0].content).toContain(
+        "Tool call limit exceeded"
       );
       expect(searchToolMock).toHaveBeenCalledTimes(2);
       expect(calculatorToolMock).toHaveBeenCalledTimes(0);


christian-bromann:
@fahe1em1 thanks for the patch. I would love to have an integration test for this to verify this behavior from e2e.

juno-junho:
We hit this middleware bug in production (our agent looped through re-plans instead of answering) and came to work on #10090, then found this PR already has the right fix. So instead of a competing PR, some verification and the integration test that was asked for in March.

First, the patch still lands: a820221fd cherry-picks cleanly onto today's main, the middleware suite passes with it (32/32), and your updated regression fails on current main without the ReactAgent change. Five months later the fix is still needed and still correct.

Here is an integration test in the style of the other `*.int.test.ts` files. One design note: with a naive "look up two cities" prompt, gpt-4o issues the tool calls in parallel, the turn becomes mixed allowed-plus-blocked, and the early-exit path this PR fixes never triggers. The two tools below have a data dependency (the second needs the id returned by the first), which forces sequential calls, so the second turn is fully blocked and the bug path is hit deterministically. Measured runs: fails on current main, passes with this branch's ReactAgent change (gpt-4o, temperature 0).

```ts
// libs/langchain/src/agents/middleware/tests/toolCallLimit.int.test.ts
import { describe, it, expect } from "vitest";
import { z } from "zod/v3";
import { AIMessage, ToolMessage } from "@langchain/core/messages";

import { toolCallLimitMiddleware } from "../toolCallLimit.js";
import { createAgent } from "../../index.js";

// The two tools have a data dependency (id from the first is required by the
// second), so the model has to call them sequentially. With threadLimit: 1 the
// second turn's only tool call is blocked, which is the state #10090 is about:
// every tool call in the turn blocked, nothing pending, model needs a re-entry.
const lookupCityId = tool(async ({ city }: { city: string }) => `${city.toLowerCase()}-001`, {
  name: "lookup_city_id",
  description: "Look up the internal city id for a city name. Required before get_weather.",
  schema: z.object({ city: z.string() }),
});

const getWeather = tool(async ({ cityId }: { cityId: string }) => `Weather for ${cityId}: sunny, 24C`, {
  name: "get_weather",
  description: "Get the current weather for a city id returned by lookup_city_id.",
  schema: z.object({ cityId: z.string() }),
});

describe("toolCallLimitMiddleware integration", () => {
  it("returns a final AI response after a fully blocked tool turn with exitBehavior continue", async () => {
    const agent = createAgent({
      model: new ChatOpenAI({ model: "gpt-4o", temperature: 0 }),
      tools: [lookupCityId, getWeather],
      middleware: [
        toolCallLimitMiddleware({ threadLimit: 1, exitBehavior: "continue" }),
      ],
    });

      messages: [
        {
          role: "user",
          content:
            "What is the weather in Paris? Use lookup_city_id first, then get_weather with the returned id.",
        },
      ],
    });

    const lastMessage = result.messages[result.messages.length - 1];
    expect(AIMessage.isInstance(lastMessage)).toBe(true);
    expect(lastMessage.content.length).toBeGreaterThan(0);

    const blockedToolMessages = result.messages.filter(
      (message): message is ToolMessage =>
        ToolMessage.isInstance(message) && message.status === "error"
    );
    expect(blockedToolMessages.length).toBeGreaterThanOrEqual(1);
    expect(blockedToolMessages[0].content).toContain("Tool call limit exceeded");
  }, 60000);
});
```

Take it as-is or reshape it, whichever helps this land. Posting it here so the credit for the fix stays with this PR.

PR #10415, opened 2026-03-14
fix(genai): round-trip thinking content blocks in multi-turn convos

### LINKED ISSUE #10103: google-genai: `Unknown content type thinking` — Multi-turn crashes when thinkingConfig is enabled with tool calling

pawel-twardziak:
## Problem

When `thinkingConfig.includeThoughts: true` is enabled with tool-calling, multi-turn conversations crash on the second LLM call with Unknown content type thinking (https://github.com/langchain-ai/langchainjs/issues/10103).

Root cause: `_convertLangChainContentToPart` in @langchain/google-genai handles the incoming direction (Gemini { text, thought: true } → LangChain { type: "thinking" }), but has no case for the outgoing direction. Sending those blocks back as conversation history throws.

### @langchain/google-genai

- Added a "thinking" case in `_convertLangChainContentToPart` that converts LangChain { type: "thinking", thinking, signature? } blocks back to Gemini { text, thought: true, thoughtSignature? } parts - the exact inverse of the incoming conversion.
- Added round-trip unit tests covering: thinking with/without signature, thinking-only content, and a full Gemini→LangChain→Gemini round-trip.

### @langchain/google

- Fixed two incorrect test assertions in `src/converters/tests/messages.test.ts`:
  - "merges consecutive ToolMessages" test incorrectly assumed an AIMessage with tool_calls produces no model content block (it does - with functionCall parts).
  - "Multiple tool calls name resolution (v1 path)" test expected 2 separate function contents, but the converter's same-role merging logic correctly combines them into 1.

Fixes #10103


pawel-twardziak:
maybe we should add some more integration tests?

PR #10416, opened 2026-03-14
fix(@langchain/google) fix Vertex AI tool calling (missing thought signature, and response parts not equal to the number of function call parts)

andruhon:
Two issues break tool-calling conversations with `@langchain/google`:

1. **Duplicate `functionCall` parts** - The legacy message converter emits each `functionCall` twice: once from `message.content` and once from `message.tool_calls`. Gemini sees 2N function calls but only N function responses and rejects the request with _"number of function response parts is equal to the number of function call parts"_.

2. **Missing `thoughtSignature`** - Gemini 3.x thinking models require `thoughtSignature` on `functionCall` parts. The converter drops it when round-tripping through `tool_calls`.

Both issues affect the legacy (v0) content path, which is the path all current consumers hit (see _Note on v0/v1_ below).

This PR initially had different diff, but the scope has changed when #10400 was merged.

## Fix (`converters/messages.ts`)

- **Legacy path**: skip `functionCall` blocks during `content` iteration - they're already emitted by the `tool_calls` loop
- **Both paths**: preserve `thoughtSignature` from `tool_calls` on emitted `functionCall` parts

### New unit tests (`converters/tests/messages.test.ts`)
- Duplicate functionCall prevention (legacy path)
- `thoughtSignature` preservation (legacy path)
- `thoughtSignature` preservation (standard path)

### New integration test (`index.int.test.ts`)
- `function - stream tools - parallel tool calling` - streams parallel tool calls, executes tools, feeds responses back, verifies final answer. Exercises the exact broken code path.

### Integration test filtering
- `GOOGLE_TEST_MODELS` - comma-separated model filter (e.g. `gemini-3.1-flash-lite-preview`)
- `GOOGLE_TEST_AUTH` - `adc` or `apikey` to limit auth combinations

## Note on v0/v1

All current consumers hit the legacy (v0) message conversion path. The standard (v1) path requires `LC_OUTPUT_VERSION=v1` opt-in, but even then it's ineffective when a callback handler prefers streaming (the default for `createAgent`) due to a missing `castStandardMessageContent` call in `BaseChatModel._generateUncached()`. Filed as #10476.

## Verified
- Downstream app tested with Gemini API v1 and v1beta1


andruhon:
Tested 3 variations (streaming):
gemini-2.5-pro with v1
gemini-3.1-pro-preview with v1
gemini-3.1-pro-preview with v1beta1

Additionally tested without streaming and it also worked.
Tested on gemini-3.1-flash-image-preview as well, it was able to call a tool

afirstenberg:
Do we have any integration tests that demonstrate the problems?

andruhon:
> Do we have any integration tests that demonstrate the problems?

This is a really good point. I'll check how can I expand libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts to cover this. Thank you!

andruhon:
I added `function - stream tools - multiple tools simultaneously` which is consistently failing without the change.

Apart from that I split test in three files: setup, index and tools and added some filters allowing to run specific model and specific auth type, for example
`GOOGLE_TEST_MODELS=gemini-3.1-pro-preview GOOGLE_TEST_AUTH=adc`

afirstenberg:
Any particular reason for the split, @andruhon ?
The additional filters sound good - I'll take a closer look at them.

andruhon:
> Any particular reason for the split, @andruhon ?
> 
> The additional filters sound good - I'll take a closer look at them.

It looks a bit too large with different sorts of types of tests jammed together. Do you think split not worth it? Should not be a big deal to undo it leaving only new test and filters.

afirstenberg:
@andruhon 
> It looks a bit too large with different sorts of types of tests jammed together. Do you think split not worth it? Should not be a big deal to undo it leaving only new test and filters.

Mostly selfish reasons - the way I run a final integration test is to run everything in the Core group. Having two Core groups just doubles my work. And for any new model or change _all_ of these need to be run. I currently only have different `describe.each` group for different _types_ of models.

Perhaps put them inside a simple `describe` group nested in the "Google Core" describe group? This way it is easier to run just those if necessary. (Or perhaps leave them as is and I'll adapt.)

andruhon:
@afirstenberg I merged the test back into one file. I needed to massage everything a bit more to make integration tests pass with all older models (gemini-2.5-*) and with newer models (gemini-3.1-*). I had to add some changes to common, adding thought signature to the part. 

"Works on my machine ©"

image

libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
```diff
@@ -254,32 +271,23 @@ function propSum(o: Record<string, number>): number {
 }
 
 const weatherTool = tool(
-  (_) => ({
-    temp: 21,
-  }),
+  (_) => ({ temp: 21, unit: "celsius" }),
```
afirstenberg: Was there a reason for this change (and the other similar changes)? It seems somewhat gratuitous.

libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
```diff
@@ -370,10 +378,10 @@ describe.each(coreModelInfo)(
 
       expect(AIMessage.isInstance(result)).to.equal(true);
 
-      expect(result.content as string).toMatch(/(1 + 1 (equals|is|=) )?2.? ?/);
+      expect(result.text as string).toMatch(/(1 + 1 (equals|is|=) )?2.? ?/);
```
afirstenberg: I think `result.content` is intentional here, although usually I would agree that it should be evaluating `result.text` (and elsewhere it is).

libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
```diff
@@ -370,10 +378,10 @@ describe.each(coreModelInfo)(
...
 
       expect(Array.isArray(result.contentBlocks)).to.equal(true);
-      expect(result.contentBlocks.length).to.equal(1);
+      expect(result.contentBlocks.length).toBeGreaterThanOrEqual(1);
```
afirstenberg: Have we seen cases where there is more than one content block? This is evaluating invoke, not stream. I wouldn't expect there to be more than one in this case. (And I don't think I've seen one.)

libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
```diff
@@ -893,8 +960,8 @@ describe.each(coreModelInfo)(
       const aiMessage = res as AIMessageChunk;
       expect(aiMessage.content).toBeDefined();
 
-      expect(typeof aiMessage.content).toBe("string");
-      const text = aiMessage.content as string;
+      expect(aiMessage.text).toBeDefined();
+      const text = aiMessage.text as string;
```
afirstenberg: I think this one should be left as `content` since it is evaluating the legacy way of doing things, for backwards compatibility...

libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
```diff
@@ -935,8 +1002,8 @@ describe.each(coreModelInfo)(
...
```
afirstenberg: ...but I agree that it makes sense that this should be `text`. Same for other legacy/not methods.

libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
```diff
@@ -1321,9 +1388,9 @@ describe.each(thinkingModelInfo)(
       });
       const result = await llm.invoke("Why is the sky blue?");
       const reasoningSteps = result.contentBlocks.filter(
-        (b) => b.type === "reasoning"
+        (b) => b.type === "reasoning" || ("thought" in b && b.thought)
```
afirstenberg: Under what conditions would we have "thought" set in a content block that isn't a reasoning block?

libs/providers/langchain-google/src/converters/messages.ts
```diff
@@ -392,13 +392,15 @@ function convertStandardContentMessageToGeminiContent(
...
   let role: Gemini.Role;
+  let isToolResponseMessage = false;
   if (HumanMessage.isInstance(message)) {
     role = "user";
   } else if (AIMessage.isInstance(message)) {
     role = "model";
   } else if (ToolMessage.isInstance(message)) {
-    // Tool messages in Gemini are represented as function responses
-    role = "function";
+    // Current Gemini/Vertex docs place functionResponse parts in a user turn.
```
afirstenberg: See #10400

libs/providers/langchain-google/src/converters/messages.ts
```diff
@@ -1148,7 +1250,8 @@ export const convertGeminiCandidateToAIMessage: Converter<
   } else if (
     contentParts.length === 1 &&
     "text" in contentParts[0] &&
-    !("thought" in contentParts[0])
+    !("thought" in contentParts[0]) &&
+    !parts.some(p => "thoughtSignature" in p)
```
afirstenberg: I think I need to look at this more closely, but why exclude turning a single text part that has a thought signature into a string?
I need to dig through things more carefully, but I was pretty sure we included this for backwards compatibility and added the thought signature as metadata on the `AIMessage` itself.

afirstenberg (changes requested):
Initial thoughts and questions. I still need to look more closely at the changes for parallel tool calls.

libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
andruhon: I recall with these new models weird things were happening with `content`, such as it not always being a string. I'll experiment more and will try to make changes in int tests as small as possible.

libs/providers/langchain-google/src/converters/messages.ts
andruhon: The 10400 should make many changes in my PR unnecessary. Should I rebase on top of 10400 and target it in this PR so that we have a nice clean diff?

afirstenberg: I expected @hntrl to have merged #10400 this weekend, but it didn't happen, so possibly wait until it gets merged into main. (It's the same solution, so I'm not exactly bothered by it. More just warning that it would generate a conflict when we tried to merge later.)

andruhon: @afirstenberg your branch has conflicts with main :(

libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
andruhon: reverted weather test tool changes

libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
andruhon: Since legacy way of doing does not work with Gemini 3, maybe we skip legacy tests on new models? For Gemini 2.5 - this is still text, but with Gemini 3 it is ContentBlock.Text with the thought signature prop.

afirstenberg: The legacy way _should_ be working. It is supposed to for backwards compatibility.

It was in my tests. If it is not, then something else is up.

andruhon:
The conditions I was trying to address no longer exist, the parallel tool calling now falls apart with `Please ensure that the number of function response parts is equal to the number of function call parts of the function call turn.`. Which is different. I updated to master, re-created filtering and expanded the integration test which fails on parallel tool calling.

andruhon:
@afirstenberg after merging in your PR I significantly reduced the size of my PR. I have to admit this is a symptomatic treatment, I experimented a little bit and it appearsd integration tests and my downstream app always go the legacy path `convertLegacyContentMessageToGeminiContent`:

castStandardMessageContent is NOT being called because it's gated behind an outputVersion parameter that defaults to "v0". It's only called when outputVersion === "v1" is explicitly opted into — either via a constructor param or environment variable on the           
  BaseChatModel.                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                             
  The flow:                                                                                                                                                                                                                                                                  
  1. BaseChatModel._generateUncached() checks outputVersion (defaults to "v0")                                                                                                                                                                                               
  2. Only if "v1" → calls castStandardMessageContent() which sets response_metadata.output_version = "v1"                                                                                                                                                                    
  3. The Google provider never explicitly sets output_version on its own AIMessages                                                                                                                                                                                          
  4. So when messages come back through the converter, they always have no output_version → defaults to "v0" → legacy path

**I'm not sure is this a bug, or feature.**, but looking in [libs/langchain-core/src/language_models/chat_models.ts](https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-core/src/language_models/chat_models.ts) it appears to be intentional.

andruhon:
The v0/v1 issue is probably a bug in in langchain-core's BaseChatModel _generateUncached(). The streaming aggregation branch is missing the castStandardMessageContent call that the non-streaming branch has. This makes messages.ts disrespect the output version even when LC_OUTPUT_VERSION environment variable is defined.

I will later create a separate issue for this.

afirstenberg:
I'll be honest - I'm having trouble keeping track of all the open issues and which might fix which.

The "number of function response parts is equal to the number of function call parts" isn't related to parallel calls. Something got introduced that is adding a part to the model part of the second request that wasn't in the model's response. See #10474 where I try to break it down.

andruhon:
> I'll be honest - I'm having trouble keeping track of all the open issues and which might fix which.
> 
> The "number of function response parts is equal to the number of function call parts" isn't related to parallel calls. Something got introduced that is adding a part to the model part of the second request that wasn't in the model's response. See #10474 where I try to break it down.

This PR addresses parallel call **and thought signature**, so yes it covers #10474. The parallel call producing wrong number of parts generally happens in v0 path and to my understanding should not happen in v1, however due to other issue I mentioned for Vertex AI the v1 is technically a dead code.

I created a separate issue for v1 vs v0 https://github.com/langchain-ai/langchainjs/issues/10476

andruhon:
Most of the issues were resolved since the original PR was created, leaving only integration tests in the scope. I created a separate test with integration test only #10498

PR #10417, opened 2026-03-14
fix(google): strip media from functionResponse, preserve as sibling parts

### LINKED ISSUE #10297: Images in ToolMessage content lost when converted to Gemini functionResponse
### Description

When a `ToolMessage` contains multimodal content (text + images via `image_url` blocks), the `@langchain/google` converter stuffs the `inlineData` (base64 images) **inside** the `functionResponse.response.result` as serialized JSON. Gemini's `functionResponse` only accepts JSON — it cannot interpret `inlineData` nested inside it. The images are effectively invisible to the model.

### Environment

- `@langchain/google`: 0.1.5
- `@langchain/core`: 1.1.31
- Model: `gemini-3-flash-preview` (also affects `gemini-2.5-flash`)

### Reproduction

```typescript
import { ToolMessage, AIMessage, HumanMessage } from '@langchain/core/messages'

const messages = [
  new HumanMessage('What is in the attachment?'),
  new AIMessage({
    content: '',
    tool_calls: [{ id: 'call_123', name: 'read_file', args: { path: 'image.png' } }]
  }),
  new ToolMessage({
    content: [
      { type: 'text', text: 'Image: screenshot.png (133KB)' },
      {
        type: 'image_url',
        image_url: { url: 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUg==' }
      }
    ],
    tool_call_id: 'call_123',
    name: 'read_file'
  })
]
```

### Expected Behavior

The converted Gemini content should have `inlineData` as a **sibling part** alongside `functionResponse`, so Gemini can visually interpret the image:

```json
{
  "role": "function",
  "parts": [
    {
      "functionResponse": {
        "name": "read_file",
        "response": { "result": "Image: screenshot.png (133KB)" }
      }
    },
    {
      "inlineData": {
        "mimeType": "image/png",
        "data": "iVBORw0KGgoAAAANSUhEUg=="
      }
    }
  ]
}
```

### Actual Behavior

`inlineData` is serialized as JSON inside `functionResponse.response.result`:

```json
{
  "role": "function",
  "parts": [
    {
      "functionResponse": {
        "name": "read_file",
        "response": {
          "result": "[{\"type\":\"text\",\"text\":\"Image: screenshot.png (133KB)\"},{\"inlineData\":{\"mimeType\":\"image/png\",\"data\":\"iVBORw0KGgoAAAANSUhEUg==\"}}]"
        }
      }
    }
  ]
}
```

The model receives the base64 string as text, not as visual data. It either hallucinates the image content or fails to interpret it.

### Root Cause

In `src/converters/messages.ts`, both `convertStandardContentMessageToGeminiContent` and `convertLegacyContentMessageToGeminiContent` convert all ToolMessage content parts into a single `functionResponse`. When the content is multimodal (contains `image_url` blocks), the `inlineData` parts get serialized into the `response.result` JSON string rather than being extracted as sibling Gemini content parts.

### Suggested Fix

When processing a `ToolMessage`, extract `inlineData` and `fileData` parts before building the `functionResponse`, then add them as sibling parts:

```typescript
if (ToolMessage.isInstance(message) && message.tool_call_id) {
  const mediaParts = parts.filter((p) => "inlineData" in p || "fileData" in p);
  const textParts = parts.filter((p) => !("inlineData" in p) && !("fileData" in p));
  
  const responseContent = typeof message.content === "string"
    ? message.content
    : textParts.map((p) => p.text || "").filter(Boolean).join("\n") || JSON.stringify(message.content);
  
  parts.length = 0;
  parts.push(
    {
      functionResponse: {
        id: message.tool_call_id,
        name: message.name || "unknown",
        response: { result: responseContent }
      }
    },
    ...mediaParts
  );
}
```

This keeps `functionResponse` as pure JSON (text only) and places `inlineData`/`fileData` as sibling parts where Gemini can interpret them visually.

### Impact

Any agent that returns images from tool calls (e.g., reading files, fetching screenshots, downloading attachments) will have the images silently dropped when using Gemini models through `@langchain/google`. The model receives only the base64 string as text and either hallucinates the image description or fails.

### Workaround

We are currently using a `pnpm patch` on `@langchain/google` to extract media parts from `ToolMessage` content before building the `functionResponse`.

pawel-twardziak:
When a ToolMessage contains multimodal content (text + images via image_url blocks), the converter serializes inlineData/fileData inside functionResponse.response.result as JSON text. Gemini cannot interpret images nested inside a JSON string - they become invisible to the model.

This PR fixes both the legacy and v1 converter paths to:

1. Exclude media items (image_url, media, image, audio, video, file) from functionResponse.response.result, keeping it text-only
2. Preserve media as sibling parts (inlineData/fileData) alongside functionResponse, where Gemini can interpret them natively
3. Filter out redundant text parts from tool responses - their content is already in functionResponse.response.result

Also adds @langchain/google to the CI unit test matrix - it was missing from the PACKAGES list in unit-tests-integrations.yml, so its tests were never run in CI.

#10415 - this PR must be merged first - it has some tests fixed that previously couldn't even get run in CI/CD - UPDATE: this is not valid any longer

Fixes #10297


afirstenberg:
@pawel-twardziak - Thank you for this!

This addresses approach 2 from #10297, right?
Do we know if this works on Gemini 2.5?

And do we have an integration test case that illustrates the issue?

pawel-twardziak:
hi, thanks for reviewing, @afirstenberg ! 

as far as I understood - this actually implements approach 1 (sibling parts) from #10297 - inlineData/fileData are extracted as separate Part objects alongside functionResponse. This is the approach @Toze confirmed works across model versions in their local patch.

Regarding Gemini 2.5: yes, the sibling parts approach works on Gemini 2.5 (and 3). Since Content.parts is repeated Part, placing functionResponse and inlineData in separate Part objects is structurally valid per the protobuf definition.

For approach 2 (FunctionResponse.parts for Gemini 3+), that could be a follow-up PR - per your earlier suggestion of supporting "both" with a model version check. Would you like me to tackle that as well? 

Regarding integration tests - yes, we can add one. The existing suite in index.int.test.ts already has both tool conversation tests (e.g. "function conversation") and image tests (using blue-square.png). A test for this fix would combine them:

 1. Define a tool (e.g. screenshot_tool)                                                                                                                                                                                         
 2. Build a message history with a ToolMessage containing [{type: "text", text: "screenshot"}, {type: "image_url", image_url: {url: "data:image/png;base64,..."}}] using the existing blue-square.png                            
 3. Invoke the model and assert the response mentions "blue" - proving the model visually interpreted the image rather than seeing serialized base64 text

pawel-twardziak:
the branch name is wrong :man_shrugging:

afirstenberg:
@pawel-twardziak 

> as far as I understood - this actually implements approach 1

Ah, ok. I hadn't had a chance to look closely at everything yet.

> For approach 2 (FunctionResponse.parts for Gemini 3+), that could be a follow-up PR - per your earlier suggestion of supporting "both" with a model version check. Would you like me to tackle that as well?

That would be fantastic - thank you! A model check plus a way to override would be great.

I think you're right that doing it as a second PR would be best. Let's get this in there, but plan to have an update coming.

(This is also why I'd like to see an integration test.)

> Regarding integration tests - yes, we can add one. 

Very much appreciated. I didn't think it would be hard to do, but wanted to make sure I understood the problem before we fixed it. Also helps to make sure we dont' break things in the future.

If you can add the test, I can go run it in the pre-patch environment to give it a try.

Thanks again!

pawel-twardziak:
Alright, thanks for your input @afirstenberg 

the plan is:
- within this PR I'll add some integration tests
- the approach 2 will be addressed in the next follow-up pr

pawel-twardziak:
hi @afirstenberg I just sent over the tests - the PR should be ready for your review

pawel-twardziak:
Blocked by #10415 - UPDATE: not valid any longer

afirstenberg:
I'm confused why this is blocked by #10415. Just for the tests? (Can we just move the tests to this one?)

libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
```diff
@@ -526,16 +526,34 @@ describe.each(coreModelInfo)(
...
+        } else {
+          expect(result2.content).toMatch(/21/);
+        }
+      } else {
+        expect(result2.content).toBeDefined();
+      }
     });
 
     test("function reply", async () => {
+      // Thinking models require thought_signature on functionCall parts,
```
afirstenberg: I'm not sure I understand this addition. Have we seen issues along these lines with this test? (I haven't before, I don't think.) And is this not a "real model invocation"?

libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
```diff
@@ -665,6 +827,10 @@ describe.each(coreModelInfo)(
       if (model.startsWith("gemini-2.0-flash")) {
         return;
       }
+      // Not available on Gemini 3 models
```
afirstenberg: There is backwards compatibility which maps it to the Google Search tool. So while Google doesn't support it - we do.
If this test fails - then something is broken with the backwards compatibility.

libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
```diff
@@ -1282,11 +1448,17 @@ describe.each(thinkingModelInfo)(
...
       expect(result.text as string).toMatch(/(1 + 1 (equals|is|=) )?2.? ?/);
-      // With includeThoughts: true, response may have multiple parts (reasoning + text)
+      // Thought signatures are best-effort per Google docs; only assert when
+      // the API actually returned thinking content.
       const hasThoughtSignature = result.contentBlocks.some(
         (b) => "thoughtSignature" in b
       );
-      expect(hasThoughtSignature).toBe(true);
+      const hasThinkingContent = result.contentBlocks.some(
+        (b) => b.type === "reasoning" || ("thought" in b && b.thought)
```
afirstenberg: Under what conditions would "thought" be set but we not have a reasoning content block?

libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
```diff
@@ -1282,11 +1448,17 @@ describe.each(thinkingModelInfo)(
       const result = await llm.invoke("What is 1 + 1?");
 
       expect(result.text as string).toMatch(/(1 + 1 (equals|is|=) )?2.? ?/);
-      // With includeThoughts: true, response may have multiple parts (reasoning + text)
+      // Thought signatures are best-effort per Google docs; only assert when
```
afirstenberg: Citation? (It seems like we run the risk of a false failure if we only test for thought signatures when we have thought signatures.)

libs/providers/langchain-google/src/converters/messages.ts
```diff
@@ -397,8 +401,9 @@ function convertStandardContentMessageToGeminiContent(
   } else if (AIMessage.isInstance(message)) {
     role = "model";
   } else if (ToolMessage.isInstance(message)) {
-    // Tool messages in Gemini are represented as function responses
-    role = "function";
+    // Tool messages in Gemini are represented as function responses.
+    // Recent Gemini API versions require 'user' or 'model' role for all contents.
+    role = "user";
```
afirstenberg, line 405: See #10400

libs/providers/langchain-google/src/converters/messages.ts
```diff
@@ -849,7 +973,7 @@ export const convertMessagesToGeminiContents: Converter<
     if (content) {
       const prev = contents[contents.length - 1];
       if (prev && prev.role === content.role) {
-        prev.parts.push(...content.parts);
+        prev.parts?.push(...(content.parts ?? ([] as Gemini.Part[])));
```
afirstenberg: I would say changing the `if` condition to make sure parts is non-null is a bit cleaner and easier to understand. But minor nit.

libs/providers/langchain-google/src/converters/messages.ts
```diff
@@ -467,11 +553,20 @@ function convertStandardContentMessageToGeminiContent(
     );
     parts.push({
       functionResponse: {
-        id: message.tool_call_id,
```
afirstenberg: There are cases where there is an ID that needs to be returned. This should not just be removed.
See #10292

libs/providers/langchain-google/src/converters/messages.ts
pawel-twardziak: We can't change the "if" condition (or can we?) since it contains only key conditions, not logic details. It shouldn't jump to the "else" just because the detailed conditions weren't met.

libs/providers/langchain-google/src/converters/messages.ts
pawel-twardziak: dozens of integration tests fail if I don't remove it. Will check it again

pawel-twardziak:
@afirstenberg thanks for you detailed review! Really appreciated :muscle: 
fyi, all the integration tests pass in this PR's current state - I've checked it on local a few times

libs/providers/langchain-google/src/converters/messages.ts
pawel-twardziak: ok, will do

libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
pawel-twardziak: will check it again

libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
pawel-twardziak: none of the tests fail - checked on local

pawel-twardziak:
> I'm confused why this is blocked by #10415. Just for the tests? (Can we just move the tests to this one?)

that PR was first fixing the "function -> user" change - but I don't see that change there now, strange... I must have mixed something up. Will add the change here in this PR

libs/providers/langchain-google/src/converters/messages.ts
pawel-twardziak: > RequestError: Invalid JSON payload received. Unknown name "id" at 'contents[2].parts[1].function_response': Cannot find field.

pawel-twardziak: there is no `id` for Vertex AI https://docs.cloud.google.com/vertex-ai/generative-ai/docs/reference/rest/v1/Content#FunctionResponse

and for Google AI it's optional and kinda useless https://ai.google.dev/api/caching?authuser=1&hl=pl#FunctionResponse

we have `GooglePlatformType = "gai" | "gcp"` so we could conditionally include `id`. But to do so, we'd need to thread platformType through the entire converter call chain - a fairly invasive change for a field that:                                                                             
- doesn't exist on Vertex AI at all                                                                                                                                                                                               
- gets rejected by Gemini 3 preview models even on Google AI

The pragmatic call for now: just remove id. It's not usable on either platform in practice.

libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
pawel-twardziak: Yes, this issue:
> RequestError: Function call is missing a thought_signature in functionCall parts. This is required for tools to work correctly, and missing thought_signature may lead to degraded model performance. Additional data, function call `default_api:test` , position 2. Please refer to https://ai.google.dev/gemini-api/docs/thought-signatures for more details.                                                                            

This is a real model invocation

libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
pawel-twardziak: yes, it fails when I remove it:
> AssertionError: expected { tokenUsage: { …(5) }, …(11) } to have property "url_context_metadata"

libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
pawel-twardziak: right, simplified

libs/providers/langchain-google/src/converters/messages.ts
afirstenberg: We had use cases where ID was sent and, since it wasn't returned, it was rejected. I, honestly, have no idea what the situations were.

This is handled in #10292.

libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
afirstenberg: My mistake - I thought this was the search tool, not URL Context tool.
And... this used to pass.
https://ai.google.dev/gemini-api/docs/url-context says it is supported.

If this is failing - something else is going on.

libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
pawel-twardziak: I'm not sure I follow this comment. Could you rephrase it please?

libs/providers/langchain-google/src/converters/messages.ts
pawel-twardziak: so then we might need the conditional way of handling it

libs/providers/langchain-google/src/converters/messages.ts
afirstenberg: Good point - my mistake.
I still think it looks messy (perhaps break it into two lines?), but, again, minor nit.
(You can resolve this. No clue why it isn't letting me.)

libs/providers/langchain-google/src/converters/messages.ts
afirstenberg: Which #10292 does.

pawel-twardziak: imho we should skip it for now since all tests are good without this `id` param - unless we know the exact case when it's required

libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
pawel-twardziak: do the limitations have something to do with it https://ai.google.dev/gemini-api/docs/url-context?hl=en#limitations ?

pawel-twardziak: `gemini-3-flash-preview` IS in the supported models list. So it should work but doesn't return url_context_metadata.

That's likely a preview model quirk - it's documented as supported but the response doesn't include the metadata consistently. I added a comment and a warning about it.

libs/providers/langchain-google/src/converters/messages.ts
afirstenberg: We do. It's in #10292 🙂

PR #10419, opened 2026-03-14
feat(langchain): add proactive warnings to toolCallLimitMiddleware

pawel-twardziak:
## FEATURE PROPOSAL

Refers to langchain-ai/langchain#35766
JS port of langchain-ai/langchain#35894

## What changed

`toolCallLimitMiddleware` was purely reactive - it only enforced limits in `afterModel` after the LLM had already generated tool calls. The LLM had no upfront awareness, leading to wasted calls and abrupt enforcement.

This PR adds three new config options (`proactive`, `warningThreshold`, `warningBuilder`) and implements a `wrapModelCall` hook that injects an ephemeral `HumanMessage` warning when the remaining tool call budget is at or below the threshold. The warning reaches the LLM but never persists in conversation state.

**Key design decision:** uses `wrapModelCall` (not `beforeModel`) because `beforeModel` returns persistent state updates via the messages reducer - repeated warnings would accumulate in history. `wrapModelCall` keeps warnings ephemeral by modifying `request.messages` before `handler()`, following the same pattern as `contextEditingMiddleware`.

## New options

| Option             | Type                              | Default | Description                                      |
|--------------------|-----------------------------------|--------|--------------------------------------------------|
| proactive         | boolean                           | false  | Enable proactive warnings via wrapModelCall      |
| warningThreshold  | number                            | 5      | Remaining call count at which to start warning  |
| warningBuilder    | (remaining, toolName) => string   | built-in | Custom warning message builder                 |

## warningBuilder for custom messages

```typescript
const limiter = toolCallLimitMiddleware({
  runLimit: 10,
  proactive: true,
  warningBuilder: (remaining, toolName) => {
    const desc = toolName ? ` for '${toolName}'` : "";
    if (remaining <= 0) return `[Budget exhausted${desc}. Stop calling tools.]`;
    return `[${remaining} call(s) left${desc}. Plan carefully.]`;
  },
});
```

Backward-compatible: `proactive` defaults to `false`, so existing behavior is unchanged with zero overhead (no `wrapModelCall` hook is set).

## How I verified

- 11 new tests (3 initialization + 7 unit + 1 integration via `createAgent`) - all passing (44 total in file)
- Full middleware test suite: 260 passed, 0 failures across 15 test files
- `lint:fix` and `format`

## Test plan

- proactive defaults to false - no wrapModelCall hook set
- wrapModelCall hook exists when proactive: true
- Negative warningThreshold throws
- Warning injected when remaining <= threshold
- No warning when remaining > threshold
- Warning at exact threshold boundary
- Exhausted message when remaining <= 0
- Uses minimum of thread/run remaining
- Tool-specific warnings include tool name
- warningThreshold: 0 only warns at exhaustion
- Custom warningBuilder called with correct args
- Warning reaches model but does NOT persist in result.messages


pawel-twardziak:
Damn! It was intended to be a draft :D and I can't make it a draft now... seems like a github issue

PR #10424, opened 2026-03-15
fix(core): preserve multimodal content in getBufferString as placeholders

### LINKED ISSUE #9811: 🐛 Regression in v1.2.4: Multimodal `input_audio` in `HumanMessage` is flattened into text
### Checked other resources

### Reproduction

``` ts
const userInput = new HumanMessage({
  content: [
    { type: 'text', text: 'a' },
    {
      type: 'input_audio',
      input_audio: {
        data: castAudioContent.data, // base64
        format: 'wav',
      },
    },
    { type: 'text', text: 'b' },
    {
      type: 'input_audio',
      input_audio: {
        data: castAudioContent.data, // base64
        format: 'wav',
      },
    },
  ],
});
```

### Behavior before v1.2.4

Observed in `handleLLMStart`:

``` ts
human: [
  { type: 'text', text: 'a' },
  {
    type: 'input_audio',
    input_audio: {
      data: 'base64...',
      format: 'wav',
    },
  },
  { type: 'text', text: 'b' },
  {
    type: 'input_audio',
    input_audio: {
      data: 'base64...',
      format: 'wav',
    },
  },
]
```

The multimodal structure is preserved and audio input works as expected.

------------------------------------------------------------------------

### Behavior since v1.2.4

Observed in `handleLLMStart`:

``` ts
human: "ab"
```

All `input_audio` segments are dropped, the content is flattened into
plain text, and audio input no longer works.

### Summary

Starting from **LangChain v1.2.4**, multimodal `HumanMessage.content`
(for example, mixing `text` and `input_audio`) is flattened into a plain
text string when observed in `handleLLMStart`.

As a result, `input_audio` segments are dropped entirely and audio input
no longer works.

This behavior is different from versions **before v1.2.4**, where the
original structured content array was preserved.

### Expected Behavior

-   `HumanMessage.content` should preserve its original multimodal
    structure in `handleLLMStart`
-   Alternatively, a new hook or documented mechanism should be provided
    to access raw multimodal prompts
-   If this change is intentional, an explicit migration strategy should
    be documented

------------------------------------------------------------------------

### Impact

-   Multimodal input using `HumanMessage.content[]` is broken in
    versions `>= v1.2.4`
-   `handleLLMStart` can no longer be used to inspect or intercept
    multimodal prompts
-   This appears to be a breaking change, but no clear migration path or
    changelog entry was found

------------------------------------------------------------------------

### Additional Notes

This behavior suggests a prompt normalization or serialization step
introduced in v1.2.4 that concatenates text segments and ignores
non-text segments such as `input_audio`.

If this is an intentional design change, clarification and documentation
would be appreciated.

### Environment

-   LangChain version: `>= v1.2.4`
-   Runtime: Node.js
-   Use case: Audio / multimodal input via `HumanMessage`

pawel-twardziak:
Since v1.2.4 (commit `c28d24a87`), `getBufferString()` uses `m.text` which only extracts `type: "text"` content blocks. All other content types - images, audio, video, files - are silently dropped. This breaks multimodal pipelines where callback handlers using `handleLLMStart` receive flattened text with no trace of non-text content.

The original change was intentional (avoid token inflation from `JSON.stringify(m.content, null, 2)` which leaked base64 data and metadata), but the fix was too aggressive.

This PR adds a local helper `_contentBlockToString()` in `utils.ts` that renders each content block type compactly - text as-is, multimodal types as short placeholders like `[image]`, `[audio]`, `[video]`, `[file]`. This preserves content existence without inflating tokens with base64 data.

**Why not revert to `JSON.stringify`?** Base64 data would massively inflate output.
**Why not modify `m.text`?** It has a clear contract ("get text only") used by many callers.

### Changes

- Added `_contentBlockToString()` helper covering all official `ContentBlock` types: `text`, `text-plain`, `image`, `image_url`, `audio`, `input_audio`, `video`, `file`, `reasoning`, tool-related blocks, and `non_standard`
- Replaced `m.text` usage in `getBufferString` with direct `m.content` rendering via the new helper
- Fixed existing masked test that asserted image-only content produced `"Human: "` (now correctly expects `"Human: [image][image]"`)
- Added comprehensive test `"getBufferString preserves non-text content block placeholders"` covering all multimodal types, reasoning exclusion, and mixed content

Fixes #9811


christian-bromann (approved):
👍

pawel-twardziak:
@christian-bromann any idea why `pnpm build --filter langchain --filter @langchain/core --filter @langchain/anthropic --filter @langchain/community --filter @langchain/openai` failed? It succeeded on local

PR #10430, opened 2026-03-16
feat(core): support for headless tools

christian-bromann:
Name is tbd .. but here is the original to get tools working in the browser.


libs/langchain/package.json
```diff
@@ -24,7 +24,8 @@
     "hub.d.cts",
     "hub.cjs",
     "hub.d.ts",
-    "hub.js"
+    "hub.js",
+    "tools"
```
christian-bromann:
```suggestion
    "hub.js"
```

.changeset/weak-kings-notice.md
```diff
@@ -0,0 +1,6 @@
+---
+"@langchain/core": patch
+"langchain": patch
```
christian-bromann:
```suggestion
"langchain": minor
```

.changeset/weak-kings-notice.md
```diff
@@ -0,0 +1,6 @@
...
+---
+
+feat(core): support for browser tools
```
christian-bromann:
```suggestion
feat(langchain): support for browser tools
```

libs/langchain/src/tools/headless.ts
```diff
@@ -0,0 +1,222 @@
...
+ * Pass to `useStream({ tools: [...] })` on the client side.
+ */
+export type HeadlessToolImplementation<
+  SchemaT extends InteropZodObject = InteropZodObject,
+  OutputT = unknown,
+  NameT extends string = string,
+> = {
+  tool: HeadlessTool<SchemaT, NameT>;
+  execute: (args: InferInteropZodOutput<SchemaT>) => Promise<OutputT>;
+};
```
hntrl, line 51: why does this need to be its own type?

libs/langchain/src/tools/headless.ts
```diff
@@ -0,0 +1,222 @@
...
+
+/**
+ * A headless tool that always interrupts agent execution on the server.
+ *
+ * The implementation is provided separately on the client via
+ * `useStream({ tools: [...] })` using `.implement()`.
+ */
+export type HeadlessTool<
```
hntrl, line 59: I still think we should call this `ToolDefinition` 🙈

hntrl (commented):
does it also make sense to export these from the top barrel? `src/index.ts`

libs/langchain/src/tools/headless.ts
christian-bromann: Already put it on the v2 wish list. Happy to clean this up when the time is right 😉

libs/langchain/src/tools/headless.ts
christian-bromann: It's consumed by `useStream` and represents a tool definition with implementation. This is NOT a traditional `DynamicStructuredTool` as it represents the pairing between a tool definition and its implementation.

christian-bromann:
> does it also make sense to export these from the top barrel? `src/index.ts`

We [do](https://github.com/langchain-ai/langchainjs/blob/87f2e6ef0b1af224b6f9fa1a7907b5d7f8eb9d4d/libs/langchain/src/index.ts#L25-L33).

PR #10438, opened 2026-03-17
fix(google): tool_choice none with empty tools, response metadata, structured output format

### LINKED ISSUE #10432: bug(@langchain/google) `tool_choice` not transformed and sent out in the body of `generateContent` if `tools` array is empty.
### Checked other resources

### Example Code

When I do:

```typescript
const llm = new ChatGoogle({
        model: modelName,
        apiKey,
        thinkingConfig: {
            thinkingLevel: options.triageThinkingLevel,
            includeThoughts: options.includeLLMThoughts,
        },
        callbacks: sentryCallback,
    });

    return llm.bindTools([], { tool_choice: 'none'});
```

which I do because sometimes Gemini models hallucinate tool calls which aren't defined on it, based on the system instructions and the requirements of the request, the actual config isn't transformed and sent with the outgoing `generateContent` request.

Missing:
```json
"toolConfig": {
      "functionCallingConfig": {
        "mode": "NONE"
      }
    },
```

`gemini-3-flash-preview` can hallucinate a `functionCall` without this setting even if it's not in the `functionDeclarations`. Now, this might be a provider bug, but at least until that's fixed on Google's side, it would be nice if the explicit config got sent as it is defined on the langchain model. 

This might be an issue with the specific behavior of `bindTools` being a no-op and maybe would be fixed if https://github.com/langchain-ai/langchainjs/issues/10431 is addressed? I think it should work with empty `bindTools` too however, if possible.

_No response_

### Description

In example.

### System Info

**Version:** `0.1.6`
**Runtime:** `bun` `1.3.10`

pawel-twardziak:
### 1. `tool_choice: "none"` with empty tools (`converters/tools.ts`)

`convertToolChoiceToGeminiConfig` had a guard `if (!hasTools || toolChoice === undefined) return undefined` that discarded `tool_choice: "none"` whenever tools were absent. This meant `bindTools([], { tool_choice: 'none' })` silently sent no `toolConfig` to the API, allowing models like `gemini-3-flash-preview` to hallucinate function calls.

The fix splits the guard so `"none"` bypasses the `!hasTools` check - it is the only mode that is semantically meaningful without tool declarations. All other modes (`"auto"`, `"any"`, etc.) still return `undefined` when no tools are present.

### 2. `response_metadata.model` missing (`chat_models/base.ts`)

Neither the invoke path nor the streaming path included the model name in `response_metadata`. Other providers (OpenAI, Anthropic, and now `@langchain/google-genai`) expose `response_metadata.model`. The fix adds `model: data.modelVersion` (invoke) and `model: chunk.modelVersion` (stream) from the API response.

### 3. `ls_structured_output_format` missing (`chat_models/base.ts`)

`withStructuredOutput` did not set the `ls_structured_output_format` config key, so LangSmith could not track which structured output strategy or schema was used. The fix adds `ls_structured_output_format: { kwargs: { method }, schema: toJsonSchema(schema) }` to both the `"jsonSchema"` and `"functionCalling"` paths, matching what `@langchain/google-genai` already does.

### Tests

- New unit test for `convertToolChoiceToGeminiConfig("none", false)` - verifies NONE config is returned even without tools.
- Updated existing test description to clarify it covers `hasTools=false` with non-none choices.

Fixes #10432


afirstenberg:
Can we get an integration test to demonstrate the problem and ensure it is fixed?

pawel-twardziak:
@afirstenberg yes, but can we first merge that [PR](https://github.com/langchain-ai/langchainjs/pull/10491)? With int tests fixed generally

afirstenberg:
> Allen Firstenberg (@afirstenberg) yes, but can we first merge that [PR](https://github.com/langchain-ai/langchainjs/pull/10491)? With int tests fixed generally

For the record - I don't have the power to merge. I'm an outside maintainer just like you. {:
Just one that was the original contributor, so I have a good (but not perfect) sense of the code.

But that's why I was urging we have fairly isolated integration test fixes and that we have integration tests that identify the problem a PR is meant to fix. Several of those look like they _shouldn't_ be changed, since there were other bugs that the integration tests helped identify that are in the queue to be fixed.

pawel-twardziak:
> > Allen Firstenberg (Allen Firstenberg (@afirstenberg)) yes, but can we first merge that [PR](https://github.com/langchain-ai/langchainjs/pull/10491)? With int tests fixed generally
> 
> For the record - I don't have the power to merge. I'm an outside maintainer just like you. {: Just one that was the original contributor, so I have a good (but not perfect) sense of the code.
> 
> But that's why I was urging we have fairly isolated integration test fixes and that we have integration tests that identify the problem a PR is meant to fix. Several of those look like they _shouldn't_ be changed, since there were other bugs that the integration tests helped identify that are in the queue to be fixed.

Alright, that's understandable. But then, from my perspective, there is no point in adding new int tests while many of the existing fail. I'd rather fix the existing ones first, which I've done in that [PR](https://github.com/langchain-ai/langchainjs/pull/10491)

PR #10441, opened 2026-03-17
fix(google-common): pass structured objects directly in FunctionResponse

### LINKED ISSUE #10439: bug(@langchain/google) ToolMessage serializes content into JSON string, even though Gemini API accepts structured objects as function results
### Checked other resources

### Example Code

I am not sure if this is a provider specific limitation or a general Langchain API trade-off but, doing this:
```typescript
return new ToolMessage({
    // TYPE COERCION: ToolMessage.content types don't include object arrays,
    // but LangChain and Gemini both accept structured JSON objects at runtime.
    content: toolResult as unknown as [] | string,
    name: toolCall.name,
    tool_call_id: toolCall.id ?? "",
});
```
results in this
```json
{
        "role": "function",
        "parts": [
          {
            "functionResponse": {
              "id": "crwsmb08",
              "name": "get_video_captions",
              "response": {
                "result": "[{\"url\":\"https://www.youtube.com/watch?v=redacted\",\"error\":\"All 5 caption URLs failed\"}]"
              }
            }
          }
        ]
      }
```
the LLM can understand it of course (how well is hard to determine), but the [upstream docs](https://ai.google.dev/api/caching#FunctionResponse) clearly state that
> response
> object ([Struct](https://protobuf.dev/reference/protobuf/google.protobuf/#struct) format)
> Required. The function response in JSON object format. Callers can use any keys of their choice that fit the function's syntax to return the function output, e.g. "output", "result", etc. In particular, if the function call failed to execute, the response can have an "error" key to return error details to the model.

Any arbitrary JSON object can be passed as is and doesn't need serialization.

I suppose the first problematic point to address would be the necessary type cast, that `content:` on `ToolMessage` accepts arbitrary object despite the type only accepting
```
(property) content?: string | ({
    [key: string]: unknown;
    readonly type: string;
    id?: string | undefined;
} | {
    [key: string]: unknown;
    readonly type: "text";
    text: string;
    index?: string | number | undefined;
    annotations?: (BaseContentBlock | ContentBlock.Citation)[] | undefined;
    id?: string;
})[] | undefined
``` 
(which requires the `type` or `text` property)
And I'm guessing the V1 message construction format with `contentBlocks` is even more strict.

I know this escape hatch is already being used for serializing and deserializing `functionCall` blocks with `toJSON()` (rather than just emitting `tool_calls`). I suppose I am wondering whether we can extend it to not auto-serialize non-standard ToolMessage `content:` ... and whether we should.

_No response_

### Description

In the example

### System Info

**Version:** `0.1.6`
**Runtime:** `bun` `1.3.10`

jackjin1997:
## Summary

- Fix `toolMessageToContent` in `@langchain/google-common` to pass structured objects directly to the Gemini API's `FunctionResponse.response` field instead of double-serializing them
- The Gemini API accepts any JSON object as the response (Struct format), but the previous code always wrapped parsed content in `{ content: parsed }`, which added an unnecessary nesting layer
- Now correctly handles: JSON string content (parsed and passed directly), plain string content (wrapped in `{ result }` ), JSON arrays, and ContentBlock arrays

## Why

The [Gemini API docs](https://ai.google.dev/api/caching#FunctionResponse) specify that `FunctionResponse.response` accepts any arbitrary JSON object directly. The previous implementation unnecessarily stringified and re-parsed content, and wrapped results in `{ content: ... }` which doesn't match the API's expected format. This caused issues when tools return structured JSON — the model received stringified JSON instead of the actual object structure.

## Test plan

- [x] Added unit tests for JSON string content → parsed object in response
- [x] Added unit tests for plain string content → `{ result: string }` 
- [x] Added unit tests for JSON array content → `{ result: array }`
- [x] All 43 existing + new tests pass

Fixes #10439

> This PR was developed with assistance from an AI coding agent.

### DIFF
--- a/.changeset/fix-google-tool-message-structured.md
+++ b/.changeset/fix-google-tool-message-structured.md
@@ -1,0 +1,5 @@
+---
+"@langchain/google-common": patch
+---
+
+Pass structured JSON objects directly in `FunctionResponse.response` instead of wrapping them in `{ content: ... }`, matching the Gemini API's expected format.
--- a/libs/providers/langchain-google-common/src/tests/utils.test.ts
+++ b/libs/providers/langchain-google-common/src/tests/utils.test.ts
@@ -1092,3 +1092,81 @@ describe("gemini empty text content handling", () => {
     expect(textPart.text).toBe("I am doing well");
   });
 });
+
+describe("gemini tool message formatting", () => {
+  test("passes JSON object content directly in FunctionResponse", async () => {
+    const api = getGeminiAPI();
+    const messages = [
+      new HumanMessage("What is the weather?"),
+      new AIMessage({
+        content: "",
+        tool_calls: [
+          { id: "call_1", name: "get_weather", args: { city: "SF" } },
+        ],
+      }),
+      new ToolMessage({
+        tool_call_id: "call_1",
+        content: JSON.stringify({ temperature: 72, unit: "F" }),
+      }),
+    ];
+
+    const formatted = (await api.formatData(messages, {})) as GeminiRequest;
+    const functionContent = formatted.contents?.find(
+      (c) => c.role === "function"
+    );
+    const part = functionContent?.parts[0] as any;
+    expect(part.functionResponse.response.temperature).toBe(72);
+    expect(part.functionResponse.response.unit).toBe("F");
+  });
+
+  test("wraps plain string content in result field", async () => {
+    const api = getGeminiAPI();
+    const messages = [
+      new HumanMessage("What is the weather?"),
+      new AIMessage({
+        content: "",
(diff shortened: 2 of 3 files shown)


jackjin1997:
Quick ping on this one — fixes Google ToolMessage content being serialized as a raw string instead of structured parts. Scoped to a single file in the Google provider converter. CI green, tests included. @bracesproul

jackjin1997:
Closing this — it's been sitting without review for a couple of months, and there are other open PRs (#10453, #10458) for the same issue. Stepping back to avoid duplicate effort; happy to pick it back up if a maintainer wants this particular approach.

PR #10446, opened 2026-03-17
fix(agents): propagate store and configurable to ToolNode middleware runtime

hntrl:
## Summary

Propagates `store` and `configurable` from the LangGraph config to the middleware runtime in `ToolNode`, fixing an issue where `wrapToolCall` middleware could not access the LangGraph store or configurable values.

### `langchain`

When `ToolNode` builds the runtime object passed to middleware's `wrapToolCall`, it was missing `store` and `configurable` from the `LangGraphRunnableConfig`. This meant middleware that needed to read from or write to the LangGraph store during tool execution had no way to access it.

This adds both `store` and `configurable` to the runtime object alongside the existing `context`, `writer`, `interrupt`, and `signal` fields.

Two tests are added to verify:
- `store` and `configurable` are available in `wrapToolCall` middleware runtime
- `store` is available in `wrapModelCall` middleware runtime

### DIFF
--- a/.changeset/config-propagation.md
+++ b/.changeset/config-propagation.md
@@ -1,0 +1,5 @@
+---
+"langchain": patch
+---
+
+fix(agents): propagate store and configurable to ToolNode middleware runtime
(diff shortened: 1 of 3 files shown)


christian-bromann (approved):
👍

PR #10450, opened 2026-03-18
fix(pgvector): prevent connection pool leak when external pool provided

Carlos-Zen:
## Description

Fixes langchain-ai/langchainjs-community#21

When a user provides an external pool via `config.pool`, calling `end()` should not destroy that pool since the caller is responsible for its lifecycle. Previously, `end()` would always destroy the pool, causing connection pool leaks for users managing their own pools.

## Changes

- Track whether pool was created internally with `_poolCreatedInternally` flag
- Only destroy pool in `end()` if it was created internally by PGVectorStore
- Remove unused `_initializeClient()` method and `client` property (they were allocating a connection from the pool that was never released)

## Use Case

Users who want to share a connection pool across multiple PGVectorStore instances or with other parts of their application:

```typescript
// Create a shared pool
const pool = new pg.Pool({ connectionString: "..." });

// Create multiple vector stores sharing the same pool
const store1 = await PGVectorStore.initialize(embeddings, { pool });
const store2 = await PGVectorStore.initialize(embeddings, { pool });

// When done with stores, call end() - it won't destroy the pool
await store1.end();
await store2.end();

// Pool is still usable
const result = await pool.query("SELECT 1");

// Only destroy the pool when you're done with it
await pool.end();
```

## Breaking Changes

None. The behavior is backward compatible:
- Users who don't provide a pool (most common case) will have the pool destroyed when `end()` is called, same as before
- Users who provide an external pool will now have correct behavior (pool not destroyed)

## Testing

The existing tests should pass. The change is backward compatible.

### DIFF
--- a/libs/community/langchain-community/src/vectorstores/pgvector.ts
+++ b/libs/community/langchain-community/src/vectorstores/pgvector.ts
@@ -311,7 +311,12 @@ export class PGVectorStore extends VectorStore {
 
   pool: Pool;
 
-  client?: PoolClient;
+  /**
+   * Tracks whether the pool was created internally by PGVectorStore.
+   * Used to determine whether end() should destroy the pool.
+   * If the pool was provided externally via config.pool, we should not destroy it.
+   */
+  private _poolCreatedInternally: boolean;
 
   chunkSize = 500;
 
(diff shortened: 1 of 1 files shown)


libs/community/langchain-community/src/vectorstores/pgvector.ts
```diff
@@ -432,6 +437,8 @@ export class PGVectorStore extends VectorStore {
         "You must provide either a `postgresConnectionOptions` object or a `pool` instance."
       );
     }
+    // Track if pool was created internally so we know whether to destroy it in end()
+    this._poolCreatedInternally = !config.pool;
     const pool = config.pool ?? new pg.Pool(config.postgresConnectionOptions);
```
christian-bromann, line 442: If `config.pool` is `undefined we create a pool connection at 442 and would miss to close it properly in 1130

christian-bromann (changes requested):

christian-bromann:
Closing as we moved the community package into a separate repository. We recommend to host the community code within your own project at this point.

PR #10451, opened 2026-03-18
fix(pdf-loader): support both pdf-parse v1 and v2

Carlos-Zen:
## Description

Fixes langchain-ai/langchainjs-community#22

PDFLoader was hardcoded to use pdf-parse v1's internal import path (`pdf-parse/lib/pdf.js/v1.10.100/build/pdf.js`), which is not exported in pdf-parse v2. This caused `ERR_PACKAGE_PATH_NOT_EXPORTED` errors when users installed pdf-parse v2 as declared in package.json.

## Changes

- Update `PDFLoaderImports()` to detect pdf-parse version at runtime
- Try pdf-parse v2 main export first (`PDFParse` class)
- Fall back to pdf-parse v1 internal path if v2 not available
- Add `parseWithV2()` method to handle v2 API
- Update `package.json` to accept both v1 and v2 (`^1.0.0 || ^2.0.0`)

### pdf-parse v1 API
- Uses `getDocument()` from internal pdf.js path
- Returns `{ getDocument, version }`

### pdf-parse v2 API
- Uses `PDFParse` class from main export
- `new PDFParse({ data: Uint8Array })` → `getText()`, `getInfo()`
- Requires calling `destroy()` to clean up resources

## Backward Compatibility

✅ Users with pdf-parse v1: Works as before (fallback path)
✅ Users with pdf-parse v2: Now works correctly (previously broken)

## Testing

- The fix has been tested with both v1 and v2 API signatures
- Error messages updated to guide users to install either v1 or v2


EurFelux:
Really need this. It cause every project using `PDFLoader` of `@langchain/community` cannot use pdf-parse@v2 meanwhile since it's declared peer dependency.

christian-bromann:
Moved to https://github.com/langchain-ai/langchainjs-community/pull/30

PR #10452, opened 2026-03-18
fix(google-genai): merge consecutive ToolMessages for parallel function calling

### LINKED ISSUE #10342: Bug: @langchain/google-genai creates separate turns for consecutive ToolMessages, breaking Gemini parallel function calling

Carlos-Zen:
## Description

Fixes #10342

## Problem

`convertBaseMessagesToContent` in `@langchain/google-genai` had an off-by-one bug:

```js
const prevContent = acc.content[acc.content.length]; // BUG: always undefined
```

This should be `acc.content[acc.content.length - 1]` because arrays are 0-indexed.

The bug prevented detection of consecutive `ToolMessage`s with the same role, causing each `ToolMessage` to create its own "user" content entry. The Gemini API requires all parallel function responses in a single user turn — separate turns cause the API to hang indefinitely.

## Solution

1. Fixed the off-by-one bug (`acc.content.length` → `acc.content.length - 1`)
2. Added handling to merge consecutive `ToolMessage`s (role "function") into a single content entry with multiple `functionResponse` parts

## Changes

- Fixed `prevContent` calculation to correctly access the last element
- Added logic to merge consecutive function/tool messages instead of throwing an error
- Consecutive ToolMessages now produce a single "user" turn with all `functionResponse` parts

## Testing

The fix handles the reproduction case described in the issue:
1. Model returns N parallel tool calls
2. Returns results as separate `ToolMessage` instances
3. Messages are now correctly merged into a single "user" turn
4. API no longer hangs

## Backward Compatibility

✅ No breaking changes - this only fixes the broken behavior

### DIFF
--- a/.changeset/popular-lizards-speak.md
+++ b/.changeset/popular-lizards-speak.md
@@ -1,0 +1,5 @@
+---
+"@langchain/google-genai": patch
+---
+
+fix(google-genai): merge consecutive ToolMessages for parallel function calling


christian-bromann:
@Carlos-Zen mind checking the unit test failures?

PR #10460, opened 2026-03-18
fix(google): add missing parts filter for ToolMessage in standard converter

### LINKED ISSUE #10459: bug: standard converter sends duplicate content for ToolMessage to Gemini API
### Checked other resources

### Example Code

```typescript
// When using the standard content converter with Gemini
const toolMessage = new ToolMessage({
  content: "result text",
  tool_call_id: "call_123",
});
// Converts to parts containing BOTH:
// 1. text part: { text: "result text" }
// 2. functionResponse part: { functionResponse: { response: { result: "result text" } } }
```

No error - the content is silently duplicated in the API request.

### Description

In `libs/providers/langchain-google/src/converters/messages.ts`, the function `convertStandardContentMessageToGeminiContent()` processes ToolMessages by:

1. Converting `contentBlocks` to text parts (L430-439)
2. Adding a `functionResponse` part (L454-476)

The result sends **both** text parts and functionResponse parts to the Gemini API, but for ToolMessages only the functionResponse part should be sent (the text content is already included in `functionResponse.response.result`).

The legacy converter in the same file correctly filters at L778-780:
```typescript
parts = parts.filter((part) => "functionResponse" in part);
```

This filter is missing in the standard converter path.

### Impact

- ToolMessage content is sent twice to the Gemini API (as text and as functionResponse)
- May confuse the model or waste tokens
- Inconsistency between standard and legacy converter behavior

### System Info

`@langchain/google`

jackjin1997:
## Summary

Fixes #10459

- Add `parts.filter()` to keep only `functionResponse` parts for ToolMessages in the standard content converter
- This matches the existing behavior in the legacy converter (same file, L780)

Without this fix, ToolMessages send both text parts and functionResponse parts to the Gemini API, duplicating the content.

## AI Disclosure

This bug was identified through code review with AI assistance. The fix applies the same filter pattern already used in the legacy converter path.

### DIFF
--- a/.changeset/fix-google-toolmessage-parts-filter.md
+++ b/.changeset/fix-google-toolmessage-parts-filter.md
@@ -1,0 +1,5 @@
+---
+"@langchain/google": patch
+---
+
+Filter non-`functionResponse` parts from ToolMessage conversion in the standard converter, matching the legacy converter behavior.
--- a/libs/providers/langchain-google/src/converters/messages.ts
+++ b/libs/providers/langchain-google/src/converters/messages.ts
@@ -421,7 +421,7 @@ function convertStandardContentMessageToGeminiContent(
     role = "user";
   }
 
-  const parts: Gemini.Part[] = [];
+  let parts: Gemini.Part[] = [];
 
   // Process standard content blocks
   const contentBlocks = Array.isArray(message.contentBlocks)
@@ -473,6 +473,10 @@ function convertStandardContentMessageToGeminiContent(
         response: { result: responseContent },
       },
     });
+
+    // For tool messages, only keep functionResponse parts since the text content
+    // is already included in the functionResponse.response.result
+    parts = parts.filter((part) => "functionResponse" in part);
   }
 
   // Only return content if we have parts
--- a/libs/providers/langchain-google/src/converters/messages.ts
+++ b/libs/providers/langchain-google/src/converters/messages.ts
@@ -474,8 +474,6 @@ function convertStandardContentMessageToGeminiContent(
       },
     });
 
-    // For tool messages, only keep functionResponse parts since the text content
-    // is already included in the functionResponse.response.result
     parts = parts.filter((part) => "functionResponse" in part);
   }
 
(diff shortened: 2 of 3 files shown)


afirstenberg:
Do we have an integration test that illustrates the problem?

jackjin1997:
Good point — updated the test so all messages carry `response_metadata: { output_version: "v1" }`, which routes them through the standard converter instead of the legacy path (the legacy path already has this filter at L780).

Without the filter, the standard converter produces two parts for a ToolMessage: a `{ text: "..." }` from `contentBlocks` processing and a `{ functionResponse: ... }` from the explicit handler. Verified locally that removing the filter causes the test to fail with `expected length 1, received 2`.

jackjin1997:
@afirstenberg Following up — I updated the test to route through the standard converter (v1 path) as you suggested, with response_metadata carrying output_version: "v1". The test now properly validates the parts filter on the standard path. Any further concerns?

jackjin1997:
Closing — afirstenberg asked for an integration test against a real Google API on Mar 21, and I don't have a Google API key I can use to record a meaningful cassette here. If anyone with API access wants to pick this up, the unit test in the diff and the linked issue should be enough to reproduce. Happy to revisit if the integration-test infra changes.

elierotenberg:
You don't need an API key. `GoogleRequestRecorder` can detect the problem without an API key.

```typescript
import { AIMessage, ToolMessage } from "@langchain/core/messages";
import { ChatGoogle, GoogleRequestRecorder } from "@langchain/google";

const recorder = new GoogleRequestRecorder();
  apiKey: "unused",
  model: "gemini-2.5-flash",
  callbacks: [recorder],
});

const messages = [
  new AIMessage({
    content: [{ type: "tool_call", id: "call_1", name: "search", args: {} }],
    response_metadata: { output_version: "v1" },
  }),
  new ToolMessage({
    content: "TOOL RESULT TEXT",
    tool_call_id: "call_1",
    response_metadata: { output_version: "v1" },
  }),
];

// The request body is built and recorded before the (failing) network call.
await model.invoke(messages).catch(() => {});

console.log(JSON.stringify(recorder.request.body.contents, null, 2));
```

PR #10462, opened 2026-03-18
fix(google): preserve thought/thoughtSignature in multi-turn conversations

### LINKED ISSUE #10461: bug(@langchain/google): thinking blocks lose thought/thoughtSignature in multi-turn conversations
### Checked other resources

### Example Code

```typescript

// First turn works fine

]);
```

### Description

```typescript
case "text":
```

```typescript
}
```

### Impact

### System Info

`@langchain/google` (latest)

jackjin1997:
## Summary

Fixes #10461

- Preserve `thought` and `thoughtSignature` properties on text content blocks when converting back to Gemini Parts
- Applies to both the standard converter (`convertStandardContentBlockToGeminiPart`) and legacy converter paths
- Adds tests for thinking block round-trip in multi-turn conversations

This is the same bug that was fixed in `@langchain/google-genai` via #10415, applied to `@langchain/google`.

## AI Disclosure

This bug was identified through code review with AI assistance, by comparing the recent fix in `@langchain/google-genai` (#10415) with the `@langchain/google` converter code.

### DIFF
--- a/libs/providers/langchain-google/src/converters/messages.ts
+++ b/libs/providers/langchain-google/src/converters/messages.ts
@@ -56,13 +56,13 @@ export const geminiContentBlockConverter: StandardContentBlockConverter<{
           typeof block.data === "string"
             ? block.data
             : typeof block.data === "object" && block.data !== null
-            ? // Convert Uint8Array to base64 string
-              btoa(
-                Array.from(block.data as Uint8Array)
-                  .map((byte) => String.fromCharCode(byte))
-                  .join("")
-              )
-            : String(block.data);
+              ? // Convert Uint8Array to base64 string
+                btoa(
+                  Array.from(block.data as Uint8Array)
+                    .map((byte) => String.fromCharCode(byte))
+                    .join("")
+                )
+              : String(block.data);
         return {
           inlineData: {
             mimeType: block.mime_type,
@@ -129,13 +129,13 @@ export const geminiContentBlockConverter: StandardContentBlockConverter<{
           typeof block.data === "string"
             ? block.data
             : typeof block.data === "object" && block.data !== null
-            ? // Convert Uint8Array to base64 string
-              btoa(
-                Array.from(block.data as Uint8Array)
-                  .map((byte) => String.fromCharCode(byte))
-                  .join("")
-              )
-            : String(block.data);
+              ? // Convert Uint8Array to base64 string
+                btoa(
+                  Array.from(block.data as Uint8Array)
+                    .map((byte) => String.fromCharCode(byte))
+                    .join("")
+                )
+              : String(block.data);
         return {
           inlineData: {
             mimeType: block.mime_type,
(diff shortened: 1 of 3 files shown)


pawel-twardziak:
Thanks for this follow-up @jackjin1997 👌

christian-bromann (dismissed):
👍

afirstenberg:
Do we have an integration test that demonstrates the problem?

libs/providers/langchain-google/src/converters/messages.ts
```diff
@@ -358,12 +358,22 @@ function convertStandardVideoContentBlockToGeminiPart(
...
+  return part as Gemini.Part;
+}
+
 function convertStandardContentBlockToGeminiPart(
   block: ContentBlock.Standard
 ): Gemini.Part | null {
   switch (block.type) {
     case "text":
-      return { text: block.text };
+      return textBlockToGeminiPart(block);
```
afirstenberg: This shouldn't be necessary. (And, if it is, this isn't the fix.)

As the name and signature suggests, `convertStandardContentBlockToGeminiPart()` is only called with `ContentBlock.Standard`, which should include the reasoning/thought block as a separate type.

This function is called from within an iterator over `message.contentBlocks`. The text block returned from that should never have thought set to true.

afirstenberg (commented):
Unless we have a test case that shows why this is necessary - I don't think this patch is needed. (And if we do have a tests case - I suspect there is a different underlying issue.)

libs/providers/langchain-google/src/converters/messages.ts
afirstenberg: @christian-bromann - can you remove your approval, please, until we review further?

jackjin1997: Emmm  let me see..

afirstenberg: Thanks for taking another look, @jackjin1997 !

jackjin1997:
@afirstenberg You were right — I dug into the block translator and confirmed that thinking parts always go through `thoughtPartToMessageContent`, producing `{type: "reasoning", reasoning: "..."}` content blocks. Text blocks never carry `thought: true`.

I've rewritten the PR with the correct approach:

- **Removed** the `thought`/`thoughtSignature` preservation on `case "text"` (that was wrong)
- **Added** `case "reasoning"` to the standard converter — without this, reasoning blocks hit `default: return null` and get silently dropped in multi-turn
- **Added** reasoning block handling in the legacy converter path too (same issue — would fall through to `parts.push(item as Gemini.Part)` which sends malformed data to the API)
- Tests now use `{type: "reasoning", reasoning: "..."}` which matches the actual data flow

Thanks for pushing back on the original approach — the architecture understanding was off.

jackjin1997:
@afirstenberg picking this back up — you were right that the fix path I took here doesn't match how thinking blocks actually flow through the block translator (text blocks never carry `thought: true`, thinking parts emit `type: "reasoning"` instead). Should I rewrite this PR around the integration-test approach you suggested (recorder mode against a real Gemini endpoint that demonstrates the multi-turn replay regression), or close it and let someone start fresh? Happy to do either — just want to make sure I don't waste your review time twice on the same misframe.

jackjin1997:
Closing this — the original approach (text-block thought attribute) was misaligned with how Google's block translator actually flows thinking blocks through (reasoning type, never text+thought=true). The rewrite path I proposed on 5/26 needs an integration test against the real API to be meaningful, which I can't easily set up. Happy to revisit if anyone takes it up with a proper repro.

PR #10464, opened 2026-03-18
fix(google-cloud-sql-pg): re-throw non-string errors in PostgresLoader

### LINKED ISSUE #10463: bug: PostgresLoader silently swallows non-string errors
## Bug Description

In `libs/providers/langchain-google-cloud-sql-pg/src/loader.ts`, there are two catch blocks (in `initialize()` and `lazyLoad()`) that only handle string-type errors and silently ignore Error objects:

```typescript
} catch (error) {
  if (typeof error === "string") {
    throw Error(error);
  }
  // Non-string errors silently ignored!
}
```

## Impact

When a database query throws an Error object (which is the standard case in JavaScript), it is silently swallowed. This causes:

1. In `initialize()`: `result` stays `undefined`, then crashes on `result.fields.map()` with an unhelpful `TypeError: Cannot read properties of undefined`
2. In `lazyLoad()`: errors are silently ignored, generator yields nothing with no indication of failure

## Expected Behavior

Non-string errors should be re-thrown so users get meaningful error messages about database connection failures, query syntax errors, etc.

## Fix

Add `throw error` after the string check in both catch blocks. I have a fix ready and will submit a PR shortly.

jackjin1997:
## Description

Fixes https://github.com/langchain-ai/langchainjs/issues/10463

`PostgresLoader` has two catch blocks (in `initialize()` and `lazyLoad()`) that only handle string-type errors. Non-string Error objects (the standard case in JS) are silently swallowed, causing confusing downstream failures.

## Changes

- Added `throw error` for non-string errors in both catch blocks in `loader.ts`
- Added unit tests verifying error propagation for both string and non-string errors

## Why this is the right fix

JavaScript errors are almost always Error objects, not strings. The current code only handles the rare string case and silently ignores the common case, which:
1. Makes `initialize()` crash with unhelpful `TypeError` on `result.fields.map()`
2. Makes `lazyLoad()` silently yield nothing on database errors

---
*This PR was developed with AI assistance.*

### DIFF
--- a/.changeset/smooth-rings-wonder.md
+++ b/.changeset/smooth-rings-wonder.md
@@ -1,0 +1,5 @@
+---
+"@langchain/google-cloud-sql-pg": patch
+---
+
+fix(google-cloud-sql-pg): re-throw non-string errors in PostgresLoader
--- a/libs/providers/langchain-google-cloud-sql-pg/src/loader.ts
+++ b/libs/providers/langchain-google-cloud-sql-pg/src/loader.ts
@@ -188,8 +188,9 @@ export class PostgresLoader extends BaseDocumentLoader {
       result = await engine.pool.raw(queryStmt);
     } catch (error) {
       if (typeof error === "string") {
-        throw Error(error);
+        throw new Error(error);
       }
+      throw error;
     }
     const columnNames = result.fields.map(
       (field: { name: string }) => field.name
@@ -276,8 +277,9 @@ export class PostgresLoader extends BaseDocumentLoader {
       }
     } catch (error) {
       if (typeof error === "string") {
-        throw Error(error);
+        throw new Error(error);
       }
+      throw error;
     }
   }
 }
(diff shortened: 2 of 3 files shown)


christian-bromann (dismissed):
👍

jackjin1997:
Rebased this onto latest `main` — CI is green again and the fix itself is unchanged: both `PostgresLoader` catch blocks now re-throw non-string `Error` objects instead of silently swallowing them. You'd 👍'd this earlier before the review got dismissed, so it should be an easy second look. Mind taking another pass when you get a chance @christian-bromann?

christian-bromann (approved):
LGTM 👍

jackjin1997:
Closing this one — the `@langchain/google-cloud-sql-pg` package was deprecated and removed from the monorepo in #11060, so there's no longer a `PostgresLoader` here for this fix to target. Thanks for the earlier approval @christian-bromann. If the loader lives on in a separate repo and still swallows non-string errors, I'm happy to port the fix over there.

PR #10478, opened 2026-03-20
fix: apply castStandardMessageContent in streaming aggregation branch

guoyangzhen:
## Problem

`BaseChatModel._generateUncached()` has three code paths for handling model output, but only two of them apply `castStandardMessageContent` when `outputVersion === "v1"`:

| Code Path | Applies castStandardMessageContent? |
|-----------|-------------------------------------|
| `_streamIterator` | ✅ Yes, per chunk |
| `_generateUncached` non-streaming | ✅ Yes, per generation |
| `_generateUncached` streaming aggregation | ❌ **No** |

The third path is taken when a callback handler prefers streaming (e.g. LangGraph's `createAgent` registers streaming-preferred handlers). Without `castStandardMessageContent`, `output_version` is never set in `response_metadata` even when `LC_OUTPUT_VERSION=v1` is configured.

**Impact:** Downstream provider converters that branch on `response_metadata.output_version` (e.g. `@langchain/google`, `@langchain/openai`, `@langchain/anthropic`) always receive messages without `output_version` set and fall back to the legacy (v0) conversion path. The `LC_OUTPUT_VERSION=v1` environment variable has no effect for any agent built with `createAgent`.

## Fix

Add `castStandardMessageContent` call after chunk aggregation in the streaming branch, matching the non-streaming branch behavior:

```typescript
if (outputVersion === "v1") {
  aggregated.message = castStandardMessageContent(aggregated.message);
}
```

This is a one-line fix in `libs/langchain-core/src/language_models/chat_models.ts` that ensures all three code paths handle `outputVersion` consistently.

### DIFF
--- a/.changeset/fix-fix-cast-standard-co.md.md
+++ b/.changeset/fix-fix-cast-standard-co.md.md
@@ -1,0 +1,5 @@
+---
+"@langchain/core": patch
+---
+
+fix(core): apply castStandardMessageContent in streaming aggregation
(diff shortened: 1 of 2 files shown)


christian-bromann:
Duplicate of https://github.com/langchain-ai/langchainjs/pull/10792

PR #10480, opened 2026-03-20
fix(google): prevent duplicate functionCall parts, normalize thought to reasoning blocks, fix thoughtSignature in streaming

xutruth:
## Summary

Fixes three related bugs in `@langchain/google`'s message converter and streaming path that caused errors with Gemini thinking models (gemini-3-flash, gemini-2.5-pro, etc.):

### Bug 1: Duplicate functionCall parts
When `convertGeminiCandidateToAIMessage` processes parallel tool calls, it stores them in **both** `AIMessage.content` (as `{ type: "functionCall", ... }` blocks) and `AIMessage.tool_calls`. The converter then serialized from both sources, doubling the functionCall count and causing:

```
RequestError: Please ensure that the number of function response parts
is equal to the number of function call parts
```

**Fix**: Added a deduplication guard — if content already has functionCall parts, skip serializing from `tool_calls`.

### Bug 2: thoughtSignature leaked into content blocks and tool_calls
The converter exposed Gemini's `thoughtSignature` (a cryptographic token for thinking context) directly on `AIMessage.content` blocks and `tool_calls` objects. This is non-standard for LangChain's types and caused:

1. **LangSmith tool replay**: `tool_call() got an unexpected keyword argument 'thoughtSignature'`
2. **LangSmith model replay**: `"Thought signature is not valid"` (stale/conflicting signatures)
3. **Serialization issues**: Non-standard fields on ToolCall objects

**Fix**: Store `thoughtSignature` in `additional_kwargs.__gemini_function_call_thought_signatures__` (keyed by tool call ID), matching the approach used by the previous `@langchain/google-genai` package. The converter reads from `additional_kwargs` and re-attaches `thoughtSignature` to the correct `functionCall` parts when sending back to the Gemini API.

### Bug 3: Mismatched tool_call IDs in streaming path (thought_signature lost)
In `_streamResponseChunks`, `convertGeminiPartsToToolCalls(parts)` was called a **second time** independently from `convertGeminiCandidateToAIMessage`, generating fresh UUIDs that didn't match the keys used to store `thoughtSignatures` in `additional_kwargs.__gemini_function_call_thought_signatures__`. When the converter later tried to re-attach signatures using the mismatched IDs, it failed:

```
RequestError: Unable to submit request because function call `lookup_assets`
in the 2. content block is missing a `thought_signature`.
```

This was especially insidious because `@langchain/core`'s `_generateUncached` forces the streaming path whenever a streaming callback handler (e.g. LangSmith tracer, LangGraph CLI) is active — even when `streaming: false` is set on the model.

**Fix**: Reuse `message.tool_calls` from the already-converted `AIMessage` instead of calling `convertGeminiPartsToToolCalls` again, and forward `__gemini_function_call_thought_signatures__` through `additional_kwargs` in the `AIMessageChunk`.

Per [Google's documentation](https://ai.google.dev/gemini-api/docs/thought-signatures): *"Strict validation is enforced for all function calls within the current turn."*

### `converters/messages.ts`
- **`convertGeminiPartsToToolCalls`**: Removed `thoughtSignature` from tool call output
- **`convertGeminiPartToContentBlock`**: Removed `thoughtSignature` from content block output
- **`convertGeminiCandidateToAIMessage`**: Collects `thoughtSignature` from Gemini parts and stores in `additional_kwargs.__gemini_function_call_thought_signatures__`
- **`convertLegacyContentMessageToGeminiContent`**: Re-attaches `thoughtSignature` from `additional_kwargs`; added dedup guard for functionCall parts
- **`convertStandardContentMessageToGeminiContent`**: Same as above for v1 path

### `chat_models/base.ts`
- **`_streamResponseChunks`**: Reuse `message.tool_calls` from `convertGeminiCandidateToAIMessage` instead of calling `convertGeminiPartsToToolCalls` again; forward `__gemini_function_call_thought_signatures__` in `additional_kwargs`

### Unit tests (`converters/tests/messages.test.ts`)
- [x] AIMessage with functionCall in content AND tool_calls → exactly N parts (not 2N)
- [x] AIMessage with empty content and tool_calls → functionCall parts generated normally
- [x] `thoughtSignature` stored in `additional_kwargs`, not on content blocks or tool_calls
- [x] `thoughtSignature` re-attached from `additional_kwargs` in tool_calls path
- [x] `thoughtSignature` re-attached from `additional_kwargs` in content block path

### Integration tests (`chat_models/tests/index.int.test.ts`)
- [x] Parallel tool calls multi-turn conversation (core models) — verifies no duplicate functionCall API error
- [x] `thoughtSignature` stored in `additional_kwargs`, not leaked into `tool_calls` (thinking models)
- [x] Multi-turn tool conversation with thinking model via invoke — verifies thoughtSignature round-trip
- [x] Multi-turn tool conversation with thinking model via streaming — verifies streaming path fix

### Manual verification
- [x] Verified with live Gemini API: parallel tool calls with gemini-2.5-flash and gemini-3-flash-preview
- [x] Verified LangSmith replay works correctly after fix
- [x] Verified streaming path (streamEvents) with gemini-3-flash-preview thinking model
- [x] Verified LangGraph CLI (`npx @langchain/langgraph-cli dev`) scenario

### DIFF
--- a/.changeset/fix-duplicate-functioncall.md
+++ b/.changeset/fix-duplicate-functioncall.md
@@ -2,4 +2,4 @@
 "@langchain/google": patch
 ---
 
-fix(google): prevent duplicate functionCall parts and store thoughtSignature in additional_kwargs for proper serialization
+fix(google): prevent duplicate functionCall parts, store thoughtSignature in additional_kwargs, and fix streaming tool_call ID mismatch
--- a/libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
+++ b/libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
@@ -535,6 +535,39 @@ describe.each(coreModelInfo)(
       expect(result2.content).toMatch(/21/);
     });
 
+    test("function conversation with parallel tool calls", async () => {
+      const tools = [weatherTool, calculatorTool];
+      const llm = newChatGoogle().bindTools(tools);
+
+      // Turn 1: request that triggers parallel tool calls
+      const history: BaseMessage[] = [
+        new HumanMessage(
+          "What is the weather in New York? Also calculate 1836 + 7262."
+        ),
+      ];
+      const result1 = await llm.invoke(history);
+      history.push(result1);
+
+      // Should have at least 1 tool call (model may or may not parallelize)
+      expect(result1.tool_calls.length).toBeGreaterThanOrEqual(1);
+
+      // Execute all tool calls and add results
+      for (const tc of result1.tool_calls) {
+        const matchingTool = tools.find((t) => t.name === tc.name);
+        if (matchingTool) {
+          const toolMessage = await matchingTool.invoke(tc);
+          history.push(toolMessage);
+        }
+      }
+
+      // Turn 2: model responds with tool results — previously failed with
+      // "function response parts != function call parts" due to duplicate
+      // functionCall parts in the converter
+      const result2 = await llm.invoke(history);
+      expect(result2.content).toBeDefined();
+      expect((result2.content as string).length).toBeGreaterThan(0);
+    });
+
     test("function reply", async () => {
       const tools: Gemini.Tool[] = [
         {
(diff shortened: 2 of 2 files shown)


afirstenberg:
Do we have an integration test that demonstrates the problem and that this resolves it?

libs/providers/langchain-google/src/converters/messages.ts
```diff
@@ -997,7 +1064,15 @@ export const convertGeminiPartToContentBlock: Converter<
   ContentBlock
 > = (part: Gemini.Part): ContentBlock => {
   const block: ContentBlock = iife(() => {
-    if ("text" in part && typeof part.text === "string") {
+    // Gemini's native thinking parts have { text, thought: true }. Convert them
```
afirstenberg: We can't do this conversion at this point.
`convertGeminiPartToContentBlock` is used when creating the `content` field for an `AIMessage`.
Since the content is "v0" and we're maintaining compatibility with legacy formats that didn't do this conversion, it needs to remain as is.
Getting `message.contentBlocks` returns the correct reasoning type.

See the integration tests for examples.

(Also, as an aside, non-text thinking parts are also valid.)

xutruth:
Thanks for the review @afirstenberg!

**Integration tests added.** I've pushed two new commits:

1. **Reverted the thought-to-reasoning conversion** in `convertGeminiPartToContentBlock` — you're right, the v0 `content` field should preserve the original Gemini format, and the v1 `contentBlocks` getter (via `ChatGoogleTranslator`) already handles the conversion correctly.

2. **Added integration tests** covering the remaining three fixes:
   - `function conversation with parallel tool calls` (core models) — verifies multi-turn with parallel tool calls doesn't produce duplicate functionCall parts
   - `thoughtSignature stored in additional_kwargs, not on tool_calls` (thinking models) — verifies signatures are stored cleanly
   - `multi-turn tool conversation succeeds with thinking model` (invoke path)
   - `multi-turn tool conversation succeeds via streaming with thinking model` (streaming path) — verifies the `_streamResponseChunks` fix

**Note on the LangSmith display behavior:** We originally added the `convertGeminiPartToContentBlock` change because LangSmith serializes `message.content` (v0) when reporting traces, and Gemini's `{ type: "text", thought: true }` blocks appear as regular text in the LangSmith UI instead of being displayed as reasoning. The old `@langchain/google-genai` package (which put `type: "reasoning"` directly in content) didn't have this issue. However, I understand the v0/v1 split is architecturally correct — the fix should come from the LangSmith side (reading `contentBlocks` or recognizing the `thought` flag).

libs/providers/langchain-google/src/converters/messages.ts
xutruth: You're right — I've reverted this change. The v0 `content` field should preserve the original Gemini format, and `message.contentBlocks` (via `ChatGoogleTranslator`) already handles the `thought → reasoning` translation correctly.

Good point about non-text thinking parts as well — the current translator correctly wraps those via `reasoningContentBlock`.

The original motivation was a LangSmith display issue: since LangSmith serializes `message.content` (v0), thinking parts show up as plain text rather than reasoning blocks. But that's better addressed on the LangSmith/tracing side rather than breaking the v0/v1 contract here.

PR #10481, opened 2026-03-20
fix(openai): detect DeepSeek context overflow errors as ContextOverfl…

hnustwjj:
## Summary                                                                                                                                                                         
                                                                                                                                                                                     
  `@langchain/openai`'s `_isOpenAIContextOverflowError()` fails to recognize DeepSeek API context length errors, preventing the `ContextOverflowError` conversion. This breaks       
  downstream safety nets like the `deepagents` summarization middleware, which relies on catching `ContextOverflowError` to trigger emergency summarization.                         
                                                                                                                                                                                     
  This PR adds `"maximum context length"` to the detection patterns in `_isOpenAIContextOverflowError`.                                                                              
                                                                                                                                                                                     
  ## Root Cause Analysis                                                                                                                                                             
                                                                                                                                                                                   
  ### The Error Flow                                                                                                                                                                 
   
  ChatDeepSeek (extends ChatOpenAICompletions)                                                                                                                                       
    → OpenAI SDK sends request to DeepSeek API                                                                                                                                     
    → API returns HTTP 400                                                                                                                                                           
    → catch(e) → wrapOpenAIClientError(e)                                                                                                                                            
      → _isOpenAIContextOverflowError(e)  // ← returns false for DeepSeek errors                                                                                                     
      → error passes through unchanged (no ContextOverflowError wrapping)                                                                                                            
                                                                                                                                                                                     
  ### Why Detection Fails                                                                                                                                                            
                                                                                                                                                                                     
  `_isOpenAIContextOverflowError` (`libs/providers/langchain-openai/src/utils/client.ts`) checks three patterns:                                                                     
                                                                                                                                                                                     
  | Check | Pattern | Matches DeepSeek? |                                                                                                                                            
  |---|---|---|                                                                                                                                                                    
  | `String(e).includes(...)` | `"context_length_exceeded"` | **No** — DeepSeek does not use OpenAI's error code |                                                                   
  | `e.message.includes(...)` | `"Input tokens exceed the configured limit"` | **No** |                                                                                              
  | `e.message.includes(...)` | `"exceeds the context window"` | **No** |                                                                                                            
                                                                                                                                                                                     
  DeepSeek's actual error message:                                                                                                                                                   
                                                                                                                                                                                   
  This model's maximum context length is 131072 tokens.                                                                                                                              
  However, you requested 131079 tokens (131079 in the messages, 0 in the completion).                                                                                              
  Please reduce the length of the messages or completion.                                                                                                                            
                                                                                                                                                                                     
  DeepSeek uses the **legacy OpenAI error format** (`"maximum context length is X tokens"`), while `@langchain/openai` only matches the **current OpenAI error patterns**            
  (`context_length_exceeded`, `exceeds the context window`). The two are incompatible.                                                                                               
                                                                                                                                                                                   
  ### Downstream Impact                                                                                                                                                              
                                                                                                                                                                                   
  The `deepagents` SDK's `createSummarizationMiddleware` has a safety net that catches `ContextOverflowError` when token estimation undercounts (common with CJK text where          
  character-based estimation is inaccurate):
                                                                                                                                                                                     
  ```typescript                                             
  // deepagents/dist/index.js — SummarizationMiddleware.wrapModelCall
  if (!shouldSummarize(truncatedMessages, totalTokens, maxInputTokens)) {                                                                                                            
    try {                                                                                                                                                                            
      return await handler({ ...request, messages: truncatedMessages });                                                                                                             
    } catch (err) {                                                                                                                                                                  
      if (!isContextOverflow(err)) throw err;  // ← re-throws because error is not ContextOverflowError
      // ... emergency summarization (never reached)                                                                                                                                 
    }                                                                                                                                                                                
  }                                                                                                                                                                                  
                                                                                                                                                                                     
 // isContextOverflow() walks the error's cause chain looking for ContextOverflowError instances. Since wrapOpenAIClientError() never converts the DeepSeek error, the chain contains  
 // only a plain BadRequestError → isContextOverflow() returns false → the error is re-thrown → unrecoverable 400 crash.
                                                                                                                                                                                     
//  Why Token Estimation Misses                               

//  Even though ChatDeepSeek correctly reports profile.maxInputTokens = 128000, which should trigger summarization at 85% (~108K tokens), the deepagents SDK uses character-based token
//   estimation (~4 chars/token). For CJK text, each character can be 1–2 tokens, causing the estimator to significantly undercount. The ContextOverflowError catch is designed as the
//  safety net for exactly this scenario — but it only works if the error is properly typed.                                                                                           
                                                            
 // Fix

 // Add "maximum context length" to _isOpenAIContextOverflowError:                                                                                                                     
   
  function _isOpenAIContextOverflowError(e: object): boolean {                                                                                                                       
    if (String(e).includes("context_length_exceeded")) return true;
    if (                                                                                                                                                                             
      "message" in e &&
      typeof e.message === "string" &&                                                                                                                                               
      (e.message.includes("Input tokens exceed the configured limit") ||
        e.message.includes("exceeds the context window") ||                                                                                                                          
        e.message.includes("maximum context length"))  // ← DeepSeek & legacy OpenAI format                                                                                          
    ) return true;                                                                                                                                                                   
    return false;                                                                                                                                                                    
  }                                                                                                                                                                                  
                                                            
 // This pattern is also used by other OpenAI-compatible providers (e.g., Azure OpenAI legacy, local inference servers) that adopt the same error format.                              
  ```

### DIFF
--- a/.changeset/deepseek-context-overflow.md
+++ b/.changeset/deepseek-context-overflow.md
@@ -1,0 +1,7 @@
+---
+"@langchain/openai": patch
+---
+
+fix(openai): detect DeepSeek context overflow errors as `ContextOverflowError`
+
+DeepSeek returns `maximum context length` in 400 error messages when the context limit is exceeded. These are now recognized by `wrapOpenAIClientError`, so downstream code (e.g. summarization middleware fallback) can handle them correctly.


hnustwjj:
My changes shouldn't affect this...would you like to help me see what's going on?  : ) @christian-bromann

christian-bromann:
This seems to be an issue with our environment test and the recent Next.js update we've done. I will go ahead and merge this and fix the root issue with a separate PR.

christian-bromann:
Thank you!

PR #10482, opened 2026-03-20
feature(ChatGoogle): allow custom headers

Nico385412:
Hello folks,

This PR will allow User to pass custom headers like the ones required for paygo

https://docs.cloud.google.com/vertex-ai/generative-ai/docs/priority-paygo#use_paygo_priority_pt-drest

### DIFF
--- a/libs/providers/langchain-google/src/chat_models/base.ts
+++ b/libs/providers/langchain-google/src/chat_models/base.ts
@@ -111,7 +111,6 @@ export interface BaseChatGoogleParams
 
   /**
    * Default headers to include with every request to the API.
-   *
    */
   additionalHeaders?: Record<string, string> | undefined;
 
@@ -442,8 +441,8 @@ export abstract class BaseChatGoogle<
       new Request(url, {
         method: "POST",
         headers: {
-          "Content-Type": "application/json",
           ...this._additionalHeaders,
+          "Content-Type": "application/json",
         },
         body: JSON.stringify(body),
       })
@@ -534,8 +533,8 @@ export abstract class BaseChatGoogle<
       new Request(url, {
         method: "POST",
         headers: {
-          "Content-Type": "application/json",
           ...this._additionalHeaders,
+          "Content-Type": "application/json",
         },
         body: JSON.stringify(body),
         signal: options.signal,


pawel-twardziak:
Hi @Nico385412 

good catch!  Thanks!

libs/providers/langchain-google/src/chat_models/base.ts
```diff
@@ -525,6 +535,7 @@ export abstract class BaseChatGoogle<
         method: "POST",
         headers: {
           "Content-Type": "application/json",
```
pawel-twardziak: Custom headers can override `Content-Type` (minor footgun).
Maybe spread custom headers first, or to use `request.headers.set()` after construction so `Content-Type` is always correct?

```ts
// Safer alternative
const request = new Request(url, {
  method: "POST",
  headers: {
    ...this._additionalHeaders,
    "Content-Type": "application/json", // always wins
  },
  body: JSON.stringify(body),
});
```

libs/providers/langchain-google/src/chat_models/base.ts
```diff
@@ -434,6 +443,7 @@ export abstract class BaseChatGoogle<
...
```
pawel-twardziak: Custom headers can override `Content-Type` (minor footgun).
Maybe spread custom headers first, or to use `request.headers.set()` after construction so `Content-Type` is always correct?

```ts
// Safer alternative
const request = new Request(url, {
  method: "POST",
  headers: {
    ...this._additionalHeaders,
    "Content-Type": "application/json", // always wins
  },
  body: JSON.stringify(body),
});
```

libs/providers/langchain-google/src/chat_models/base.ts
```diff
@@ -183,6 +191,7 @@ export abstract class BaseChatGoogle<
     this.model = params.model;
     this._platform = convertParamsToPlatformType(params);
     this._endpoint = params.endpoint;
+    this._additionalHeaders = params.additionalHeaders;
```
pawel-twardziak, line 193: `additionalHeaders` is only on the constructor params (`BaseChatGoogleParams`), not on `BaseChatGoogleCallOptions`. This means headers are fixed at instantiation time and can't vary per `.invoke()` / `.stream()` call. For the PayGo use case this is fine (the priority header is typically static), but it limits flexibility for use cases like per-request tracing headers.

libs/providers/langchain-google/src/chat_models/base.ts
```diff
@@ -109,6 +109,12 @@ export interface BaseChatGoogleParams
    **/
   endpoint?: string;
 
+  /**
+   * Default headers to include with every request to the API.
+   *
```
pawel-twardziak: unnecessary empty comment line :)

libs/providers/langchain-google/src/chat_models/base.ts
Nico385412: true, let me fix that

libs/providers/langchain-google/src/chat_models/base.ts
Nico385412: fixed !

libs/providers/langchain-google/src/chat_models/base.ts
Nico385412: True,
As i understand, per-request headers ask more work (have to change some logic in core part) 
(Sorry i'm a newbie on langchain, still discovering the project x) )

pawel-twardziak:
hi @Nico385412 
I've expanded on your PR there #10491

PR #10490, opened 2026-03-21
fix(@langchain/google): deduplicate functionCall parts and preserve thoughtSignature in round-trip

### LINKED ISSUE #10474: bug(@langchain/google) False duplicate `functionCall` sent in response
### Privileged issue

### Issue Content

The integration tests surrounding function calls are still failing.
Error messages include:
* "Please ensure that the number of function response parts is equal to the number of function call parts of the function call turn."
* "Unable to submit request because function call `default_api:test` in the 2. content block is missing a `thought_signature`."
* "Function call is missing a thought_signature in functionCall parts."
* In the function conversation test: AssertionError: expected '' to match /21/

I think this is related to a few things:
* Google just changed the rules about ID and function calls and the "function reply" test may no longer be valid. This tests streaming replies, so we probably should fix this. (Tho how we do so without having a valid thought signature, I'm not sure.)
* In the *second* request, the history sent to Gemini includes an extra "functionCall" part that is missing the thought signature and request id, but is otherwise a duplicate of the first. Both are duplicates of the "functionCall" part that Gemini sent in its response. This duplicate appears to be something that we generated... somewhere.

I'm not sure where that duplicate crept in yet, but #10416 sounds like it identifies the same bug, at least, and appears to have a solution.

Library: @langchain/google 0.1.7

A concrete illustration from a run of "function conversation" test, Vertex, gemini-3-flash-preview (but similar results for others). These are abbreviated to show relevant issues:

Request 1. Requests something that the defined tool can answer.
```
  "contents": [
   {
    "role": "user",
    "parts": [
     {
      "text": "What is the weather in New York?"
     }
    ]
   }
```

Response 1. Gemini issues the tool call.
```
    "content": {
     "role": "model",
     "parts": [
      {
       "functionCall": {
        "name": "get_weather",
        "args": {
         "location": "New York"
        }
       },
       "thoughtSignature": "CsQBAY89a1+phLX27hWX...ta5bIK5jXtjfPXaVGQ=="
      }
     ]
    },
```

So far so good.
Request 2. The issue shows up here in `contents[1]` where it has mysteriously gotten a second part that was not in Response 1.
```
  "contents": [
   {
    "role": "user",
    "parts": [
     {
      "text": "What is the weather in New York?"
     }
    ]
   },
   {
    "role": "model",
    "parts": [
     {
      "thoughtSignature": "CsQBAY89a1+phLX27hWX...ta5bIK5jXtjfPXaVGQ==",
      "functionCall": {
       "name": "get_weather",
       "args": {
        "location": "New York"
       }
      }
     },
     {
      "functionCall": {
       "name": "get_weather",
       "args": {
        "location": "New York"
       }
      }
     }
    ]
   },
   {
    "role": "user",
    "parts": [
     {
      "functionResponse": {
       "name": "get_weather",
       "response": {
        "result": "{\"temp\":21}"
       }
      }
     }
    ]
   }
  ]
 }
```

maahir30:
When Gemini returns a `functionCall` response, the converter stores it in two places on the `AIMessage`: the `content` array (as a typed content block with full metadata like `thoughtSignature`) and `tool_calls` (as a standard tool call). When converting that message back to Gemini format for the next request, both sources were iterated independently, producing a **duplicate `functionCall` part** — where the second copy was missing `thoughtSignature` and `id`.

This PR fixes the round-trip by:

- **Deduplicating** in the legacy converter: if the `content` array already emitted `functionCall` parts (which preserve all Gemini metadata), skip re-emitting from `tool_calls`.
- **Enriching** the `tool_calls` → `functionCall` conversion (both legacy and v1 paths) to carry `thoughtSignature` and native `id` through, for cases where `tool_calls` is the only source (e.g. manually constructed messages or the v1 path).
- **Fixing the "function reply" integration test** to use a real round-trip instead of a synthetic history that lacks a valid `thoughtSignature`.

closes https://github.com/langchain-ai/langchainjs/issues/10474

### DIFF
--- a/libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
+++ b/libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
@@ -536,52 +536,27 @@ describe.each(coreModelInfo)(
     });
 
     test("function reply", async () => {
-      const tools: Gemini.Tool[] = [
-        {
-          functionDeclarations: [
-            {
-              name: "test",
-              description:
-                "Run a test with a specific name and get if it passed or failed",
-              parameters: {
-                type: "object",
-                properties: {
-                  testName: {
-                    type: "string",
-                    description: "The name of the test that should be run.",
-                  },
-                },
-                required: ["testName"],
-              },
-            },
-          ],
-        },
-      ];
+      const tools = [weatherTool];
       const llm = newChatGoogle().bindTools(tools);
-      const toolResult = {
-        testPassed: true,
-      };
-      const messages: BaseMessage[] = [
-        new HumanMessage("Run a test on the cobalt project."),
-        new AIMessage({
-          tool_calls: [
-            {
-              type: "tool_call",
-              id: "lc-tool-call-test-id",
-              name: "test",
-              args: {
-                testName: "cobalt",
-              },
-            },
-          ],
-        }),
-        new ToolMessage(JSON.stringify(toolResult), "lc-tool-call-test-id"),
+      const history: BaseMessage[] = [
+        new HumanMessage("What is the weather in New York?"),
       ];
-      const res = await llm.stream(messages);
+
+      const result1 = await llm.invoke(history);
+      history.push(result1);
+
+      const toolCalls = result1.tool_calls!;
+      expect(toolCalls.length).toBeGreaterThanOrEqual(1);
+
+      const toolMessage = await weatherTool.invoke(toolCalls[0]);
+      history.push(toolMessage);
+
+      const res = await llm.stream(history);
       const resArray: BaseMessageChunk[] = [];
       for await (const chunk of res) {
         resArray.push(chunk);
       }
+      expect(resArray.length).toBeGreaterThanOrEqual(1);
     });
 
     test("function - force tool", async () => {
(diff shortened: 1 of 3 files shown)


libs/providers/langchain-google/src/converters/messages.ts
```diff
@@ -736,15 +742,33 @@ function convertLegacyContentMessageToGeminiContent(
     }
   }
 
-  // Convert AIMessage tool_calls to functionCall parts
+  // Convert AIMessage tool_calls to functionCall parts, but skip when the
+  // content array already emitted functionCall blocks (avoids duplicates).
```
afirstenberg, line 746: This is better fixed at the root cause. See #10493

libs/providers/langchain-google/src/converters/messages.ts
```diff
@@ -441,11 +441,17 @@ function convertStandardContentMessageToGeminiContent(
   // Convert AIMessage tool_calls to functionCall parts
```
afirstenberg, line 441: Not necessary since these are included above. This block only caused problems. See #10493

afirstenberg (commented):
This and #10493 are solving the same issue, I think.

I know I'm biased, but I think #10493 is the more correct solution - making sure the duplicates don't appear in the first place.

libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
```diff
@@ -536,52 +536,27 @@ describe.each(coreModelInfo)(
     });
 
     test("function reply", async () => {
```
afirstenberg, line 538: At a glance, changing this test seems good to me. But then I'm not sure how it is different than the previous test - "function conversation".

Using the synthetic history was intentional at the time - but doesn't make sense anymore, and I suspect this entire test can be removed.

hntrl:
Closing in favor of #10493

PR #10498, opened 2026-03-22
test(google) improve VertexAI integration tests

andruhon:
This PR includes following changes:
- New `function - stream tools - parallel tool calling` integration test
- Filters for `@langchain/google` integration tests

The integration tests covers parallel tool calling with `@langchain/google` which had a number of different issues recently.
Apart from new integration test it introduces filters to test only specific model or specific authentication method, for example `GOOGLE_TEST_MODELS=gemini-3.1-flash-lite-preview GOOGLE_TEST_AUTH=adc pnpm --filter @langchain/google` will only run tests against 3.1 flash authenticated with ADC, not the entire matrix.

This PR is a continuation of work started in #10416, most of the issues were resolved since the original PR was created, leaving only integration tests in the scope.


andruhon:
@afirstenberg please have a look, the scope of this one is reduced to integration test only when compared to original #10416

PR #10503, opened 2026-03-22
fix(google): restore tool_calls in legacy path with deduplication

### LINKED ISSUE #10502: bug (@langchain/google) REGRESSION from `0.1.7` to `0.1.8`, `AIMessage` with only `tool_calls` gets dropped from history sent to the Gemini API
### Checked other resources

### Example Code

Same code as in https://github.com/langchain-ai/langchainjs/issues/10500:

```typescript
/**
 *
 *
 *
 */

        try {
                body = JSON.parse(rawBody);
            } else {
                body = rawBody;
            }
        } catch {
            body = "(parse error)";
        }

            status: 500,
        });
    }

} as typeof globalThis.fetch;

// Test harness

const TOOL_NAME = "my_tool";

    const llm = new ChatGoogle({
        apiKey: "FAKE KEY",
    });

    const history = [
        new AIMessage({
            tool_calls: [
                {
                    id: TOOL_CALL_ID,
                    name: TOOL_NAME,
                    args: { input: "test" },
                },
            ],
        }),
        new ToolMessage({
            tool_call_id: TOOL_CALL_ID,
            name: TOOL_NAME,
        }),
    ];

    try {
        await llm.invoke(history);
    } catch {
    }
}

]);

    status: "ok",
    value: 42,
    items: ["foo", "bar"],
});
```

Example code output:
```
{
  "generationConfig": {
    "candidateCount": 1
  },
  "contents": [
    {
      "role": "user",
      "parts": [
        {
        },
        {
          "functionResponse": {
            "id": "tool-call-123",
            "name": "my_tool",
            "response": {
            }
          }
        }
      ]
    }
  ]
}

{
  "generationConfig": {
    "candidateCount": 1
  },
  "contents": [
    {
      "role": "user",
      "parts": [
        {
        }
      ]
    }
  ]
}
```

### Description

In `@langchain/google` version `0.1.7`, the tool calls, "functionCall" blocks are included (see https://github.com/langchain-ai/langchainjs/issues/10500), in `0.1.8`, they are dropped.

If you add / change `contents` to be non-empty, the "text" chunk still gets emitted, but still no tool / function calls. 

### System Info

**Version:** `0.1.8`
**Runtime:** `bun` `1.3.11`

pawel-twardziak:
PR #10493 fixed duplicate `functionCall` parts for round-tripped AIMessages (#10474) by removing the `tool_calls` → `functionCall` conversion from the legacy (v0) content path entirely. However, this was overly broad - it re-broke manually constructed `AIMessage` objects that only have `tool_calls` (no `functionCall` items in `content`), causing them to be silently dropped from the Gemini API request.

This PR restores the `tool_calls` → `functionCall` conversion in the legacy path, but guards it with a check: if the content array already produced `functionCall` parts (the round-trip case), the conversion is skipped to avoid duplicates. This handles both scenarios correctly:

- **Round-tripped AIMessages** (content has `functionCall` items + `tool_calls` populated): content is authoritative, `tool_calls` conversion skipped - no duplicates.
- **Manually constructed AIMessages** (content is empty, only `tool_calls`): `tool_calls` converted to `functionCall` parts - message no longer dropped.

### Tests

| Test | Status without fix | Status with fix | Added by |
|---|---|---|---|
| "AIMessage with tool_calls produces model turn with functionCall parts (legacy path)" | **FAIL** | PASS | #10397 (existing) |
| "merges consecutive ToolMessages into a single function turn for parallel tool calls (legacy path)" | **FAIL** | PASS | #10397 (existing) |
| "round-tripped AIMessage with functionCall in content does not produce duplicate functionCall parts (legacy path)" | n/a | PASS | this PR (new) |

The new test covers the deduplication invariant (round-tripped AIMessage with `functionCall` in content must not produce duplicate parts). The two existing tests from #10397 that were broken by #10493 now pass again.

### `thoughtSignature` fallback for Gemini 3.0+

Gemini 3.0+ requires a valid `thoughtSignature` on every `functionCall` part ([docs](https://ai.google.dev/gemini-api/docs/thought-signatures#function-calling)). For manually constructed or cross-provider tool calls (which lack a signature), this PR adds Google's documented dummy value `"context_engineering_is_the_way_to_go"` as a fallback to skip validation.

| Path | `thoughtSignature` handling | Status |
|---|---|---|
| **Legacy content loop** (`...etc` spread) | Preserved from content block if present | Already correct |
| **V1 `tool_calls`** | Uses `toolCall.thoughtSignature` if present, else dummy value | Fixed in this PR |
| **Legacy `tool_calls`** | Same as above | Fixed in this PR |

#### Why `"context_engineering_is_the_way_to_go"`?

This is one of two official dummy values provided by Google in their [thought signatures documentation](https://ai.google.dev/gemini-api/docs/thought-signatures#function-calling) (the other being `"skip_thought_signature_validator"`). The docs state:

> *"While injecting custom function call blocks into the request is strongly discouraged, in cases where it can't be avoided, e.g. providing information to the model on function calls and responses that were executed deterministically by the client, or transferring a trace from a different model that does not include thought signatures, you can set the following dummy signatures [...] to skip validation."*

Without this fallback, any `functionCall` part generated from `tool_calls` that lacks a real `thoughtSignature` (manually constructed messages, cross-provider history from OpenAI/Anthropic) gets a 400 error on Gemini 3.0+. The dummy value tells Gemini's validator to skip the check - which is exactly the escape hatch Google designed for this scenario.

#### `thoughtSignature` tests

| Test | Path | Verifies |
|---|---|---|
| "manually constructed tool_calls get dummy thoughtSignature" | legacy | Dummy fallback applied when no signature present |
| "manually constructed tool_calls get dummy thoughtSignature" | v1 | Same, v1 path |
| "real thoughtSignature on tool_calls is preserved, not overwritten" | legacy | Round-tripped signatures not clobbered by dummy |
| "real thoughtSignature on tool_calls is preserved, not overwritten" | v1 | Same, v1 path |

Fixes #10502

### DIFF
--- a/.changeset/fluffy-mails-complain.md
+++ b/.changeset/fluffy-mails-complain.md
@@ -1,0 +1,5 @@
+---
+"@langchain/openai": patch
+---
+
+fix(openai): guard JSON.parse in streaming json_schema when text is empty
--- a/libs/providers/langchain-openai/src/converters/responses.ts
+++ b/libs/providers/langchain-openai/src/converters/responses.ts
@@ -750,7 +750,7 @@ export const convertResponsesDeltaToChatGenerationChunk: Converter<
 
     usage_metadata = convertResponsesUsageToUsageMetadata(event.response.usage);
 
-    if (event.response.text?.format?.type === "json_schema") {
+    if (event.response.text?.format?.type === "json_schema" && msg.text) {
       additional_kwargs.parsed ??= JSON.parse(msg.text);
     }
     for (const [key, value] of Object.entries(event.response)) {
--- a/libs/providers/langchain-openai/src/converters/tests/responses.test.ts
+++ b/libs/providers/langchain-openai/src/converters/tests/responses.test.ts
@@ -2085,3 +2085,126 @@ describe("tool_search support", () => {
     });
   });
 });
+
+describe("convertResponsesDeltaToChatGenerationChunk - json_schema with tool calls only (#10505)", () => {
+  it("should not throw when response.completed has json_schema format but empty text (function call only)", () => {
+    // When the model responds with only a function_call (e.g. tool_choice: required),
+    // msg.text is "" but event.response.text.format.type is still "json_schema".
+    // Previously JSON.parse("") threw SyntaxError.
+    const event = {
+      type: "response.completed",
+      response: {
+        id: "resp_test",
+        model: "gpt-4o",
+        object: "response",
+        created_at: 1700000000,
+        status: "completed",
+        incomplete_details: null,
+        metadata: {},
+        user: null,
+        service_tier: null,
+        output: [
+          {
+            type: "function_call",
+            id: "fc_001",
+            call_id: "call_abc",
+            name: "lookup",
+            arguments: '{"query":"What is LangChain?"}',
+          },
+        ],
+        text: {
+          format: {
+            type: "json_schema",
+            name: "response",
+            strict: true,
+            schema: {
+              type: "object",
+              additionalProperties: false,
+              required: ["answer"],
+              properties: { answer: { type: "string" } },
+            },
+          },
+        },
+        usage: {
+          input_tokens: 50,
+          output_tokens: 20,
+          total_tokens: 70,
+          input_tokens_details: { cached_tokens: 0 },
+          output_tokens_details: { reasoning_tokens: 0 },
+        },
+      },
+    };
+
+    const result = convertResponsesDeltaToChatGenerationChunk(event as any);
+    expect(result).not.toBeNull();
+
+    const message = result!.message as AIMessageChunk;
+    // No parsed content since the model only returned a tool call
+    expect(message.additional_kwargs.parsed).toBeUndefined();
+    // Usage metadata should still be populated
+    expect(result!.message.usage_metadata).toBeDefined();
+    expect(result!.message.usage_metadata!.input_tokens).toBe(50);
+  });
+
+  it("should parse text correctly when response.completed has json_schema format with actual text", () => {
+    // Normal case: json_schema format with actual text content
+    const event = {
+      type: "response.completed",
+      response: {
+        id: "resp_test2",
+        model: "gpt-4o",
+        object: "response",
+        created_at: 1700000000,
+        status: "completed",
+        incomplete_details: null,
+        metadata: {},
+        user: null,
+        service_tier: null,
+        output: [
+          {
+            type: "message",
+            id: "msg_001",
+            role: "assistant",
+            status: "completed",
+            content: [
+              {
+                type: "output_text",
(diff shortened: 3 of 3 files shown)


libs/providers/langchain-google/src/converters/messages.ts
```diff
@@ -438,14 +438,19 @@ function convertStandardContentMessageToGeminiContent(
...
+  // See: https://ai.google.dev/gemini-api/docs/thought-signatures#function-calling
   if (AIMessage.isInstance(message) && message.tool_calls?.length) {
     for (const toolCall of message.tool_calls) {
       parts.push({
         functionCall: {
           name: toolCall.name,
           args: toolCall.args ?? {},
         },
+        thoughtSignature:
+          toolCall.thoughtSignature ?? "context_engineering_is_the_way_to_go",
```
afirstenberg, line 453: Nit: I'd rather go with "skip_thought_signature_validator" since it is more obvious what is going on. (Yes, even with the documentation.) But whatever.

libs/providers/langchain-google/src/converters/messages.ts
```diff
@@ -736,6 +741,28 @@ function convertLegacyContentMessageToGeminiContent(
...
+  // Preserve thoughtSignature if present (round-tripped Gemini messages);
+  // otherwise use Google's documented dummy value to skip validation on Gemini 3.0+.
+  // See: https://ai.google.dev/gemini-api/docs/thought-signatures#function-calling
+  if (AIMessage.isInstance(message) && message.tool_calls?.length) {
+    const hasExistingFunctionCalls = parts.some((p) => "functionCall" in p);
+    if (!hasExistingFunctionCalls) {
+      for (const toolCall of message.tool_calls) {
+        parts.push({
+          functionCall: {
+            name: toolCall.name,
```
afirstenberg, line 756: I think we also need to add the id if we have it, yes?

afirstenberg:
@pawel-twardziak 
Good catch on needing to handle manual/saved AIMessages! And I think the approach is a solid one. 

Do we have integration tests that illustrate the problem? I would have thought that the "function reply" test, that uses just the `tool_calls` attribute would have caused an error - but it passes.

libs/providers/langchain-google/src/converters/messages.ts
afirstenberg: (Then again, we don't include the id in the code where you added the thought signature above. Which seems odd. So perhaps I'm thinking of something else.)

afirstenberg:
Looking at the documentation (https://ai.google.dev/gemini-api/docs/thought-signatures#function-calling) says the signature will only be on the **first** function call and that it shouldn't be on subsequent ones. (Tho it doesn't forbid it.)

Does this change how we approach adding the fake signature?

PR #10504, opened 2026-03-22
fix(core): allow object content in isMessage() type guard

### LINKED ISSUE #10500: bug(@langchain/google): `ToolMessage` with arbitrary object as `contents:` dropped from history sent to API call
### Checked other resources

### Example Code

Run this example code:
```typescript
/**
 *
 *
 *
 */

        try {
                body = JSON.parse(rawBody);
            } else {
                body = rawBody;
            }
        } catch {
            body = "(parse error)";
        }

            status: 500,
        });
    }

} as typeof globalThis.fetch;

// Test harness

const TOOL_NAME = "my_tool";

    const llm = new ChatGoogle({
        apiKey: "FAKE KEY",
    });

    const history = [
        new AIMessage({
            tool_calls: [
                {
                    id: TOOL_CALL_ID,
                    name: TOOL_NAME,
                    args: { input: "test" },
                },
            ],
        }),
        new ToolMessage({
            tool_call_id: TOOL_CALL_ID,
            name: TOOL_NAME,
        }),
    ];

    try {
        await llm.invoke(history);
    } catch {
    }
}

]);

    status: "ok",
    value: 42,
    items: ["foo", "bar"],
});
```

Output of the script above:
```

{
  "generationConfig": {
    "candidateCount": 1
  },
  "contents": [
    {
      "role": "user",
      "parts": [
        {
        }
      ]
    },
    {
      "role": "model",
      "parts": [
        {
          "functionCall": {
            "name": "my_tool",
            "args": {
              "input": "test"
            }
          }
        }
      ]
    },
    {
      "role": "user",
      "parts": [
        {
          "functionResponse": {
            "id": "tool-call-123",
            "name": "my_tool",
            "response": {
            }
          }
        }
      ]
    }
  ]
}

{
  "generationConfig": {
    "candidateCount": 1
  },
  "contents": [
    {
      "role": "user",
      "parts": [
        {
        }
      ]
    },
    {
      "role": "model",
      "parts": [
        {
          "functionCall": {
            "name": "my_tool",
            "args": {
              "input": "test"
            }
          }
        }
      ]
    }
  ]
}
```

### Description

The array version invocation gets serialized into a string, even though it doesn't have to be (as discussed here: https://github.com/langchain-ai/langchainjs/issues/10439)
and the object version gets dropped from history altogether.

Either they should both serialize, or neither serialize (and pass as is). I would even say throwing an Error would not be a bad pattern, but that's a wider library-level consideration.

### System Info

**Version:** `0.1.7`
**Runtime:** `bun` `1.3.11`

pawel-twardziak:
`ToolMessage` with a plain object as `content` (e.g. `{ status: "ok", value: 42 }`) is silently dropped from the history sent to provider APIs. The message is constructed correctly, but `isMessage()` - the type guard used by `BaseMessage.isInstance()` and all subclass `.isInstance()` methods - rejects it because it only accepts `string` or `Array` content. This causes provider converters (Google, OpenAI, Anthropic) to skip the message entirely since `ToolMessage.isInstance()` returns `false`.

### Root cause

`isMessage()` in `libs/langchain-core/src/messages/message.ts` checks:

```typescript
typeof message.content === "string" || Array.isArray(message.content)
```

A plain object fails both checks → `isMessage()` returns `false` → `ToolMessage.isInstance()` returns `false` → the ToolMessage handler in the converter is never entered → zero parts produced → message dropped.

### Fix

Added `message.content.constructor === Object` as a third accepted content type in `isMessage()`. The `constructor === Object` check ensures only plain objects pass - `Set`, `Map`, `Date`, generators, and other object subclasses are still rejected.

### Impact on provider libraries

This change is **more permissive, not less** - no existing valid message will suddenly fail `isMessage()`. Provider converters already handle non-string content independently by checking `typeof content === "string"` and falling back to `JSON.stringify` for everything else. The ToolMessage handler in each provider (Google, OpenAI, Anthropic) already has this pattern - it just was never reached because `isMessage()` blocked it upstream.

### Tests

Updated the existing `isMessage` test suite:

| Test | Before | After |
|---|---|---|
| "returns true when 'content' is a plain object (e.g., tool output)" | n/a | **PASS** (new) |
| "returns false when 'content' is a function" | n/a | **PASS** (new, split from old combined test) |
| "returns false for iterable but non-array 'content' (e.g., Set/generator)" | PASS | PASS (unchanged - Set/generator still rejected) |

Fixes #10500

### DIFF
--- a/.changeset/giant-moles-shave.md
+++ b/.changeset/giant-moles-shave.md
@@ -1,0 +1,5 @@
+---
+"langchain": patch
+---
+
+Align Zod Importation For 'libs/langchain/src/agents/nodes/utils.ts'
--- a/libs/langchain/src/agents/nodes/utils.ts
+++ b/libs/langchain/src/agents/nodes/utils.ts
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
-import { z } from "zod/v4";
+import { z } from "zod";
 import { type BaseMessage } from "@langchain/core/messages";
 import {
   interopSafeParseAsync,


christian-bromann:
As per conversation, this is not a valid use case. `ToolMessages` should not have objects as `content` value.

PR #10519, opened 2026-03-25
fix: implement tool choice required on anthropic

baptistejamin:
Hello there!

When using createAgent with middleware and setting toolChoice: "required" in the wrapModelCall handler, the Anthropic provider throws:

```
Tool 'required' not found in provided tools
```

This happens because handleToolChoice() in libs/providers/langchain-anthropic/src/utils/tools.ts handles "any", "auto", and "none" explicitly, but any other string falls through to a handler that assumes it's a tool name:

### DIFF
--- a/.changeset/silent-schools-rule.md
+++ b/.changeset/silent-schools-rule.md
@@ -1,0 +1,5 @@
+---
+"@langchain/anthropic": patch
+---
+
+fix: implement tool choice required on anthropic


christian-bromann:
@baptistejamin thanks for the contribution. Can we have a unit test for this?

baptistejamin:
Unit tests added :)

PR #10544, opened 2026-03-28
fix(google): handle ToolMessage with non-standard content types

### LINKED ISSUE #10500: bug(@langchain/google): `ToolMessage` with arbitrary object as `contents:` dropped from history sent to API call

alex-levkivskiy:
## Problem

`ToolMessage` with plain object `content` (e.g. `{ status: "ok", value: 42 }`) is silently
dropped from Gemini API history. Array content gets incorrectly `JSON.stringify()`-ed into
a string instead of extracting text blocks.

Fixes #10500

## Root Cause

`ToolMessage.isInstance()` returns `false` when `content` is a plain object
(not `string | MessageContentComplex[]`). The converter doesn't recognize the message
as a ToolMessage → skips it → zero parts produced → message vanishes from history.

Additionally, the legacy converter path did not convert `AIMessage.tool_calls` into
`functionCall` parts, so AIMessages with `content: ""` + `tool_calls` also disappeared.

## Fix (provider-level, 3 files)

**`@langchain/google` — `converters/messages.ts`:**
- `isToolMessageLike()` — falls back to `message.type === "tool"` when `isInstance()` fails
- `toolMessageContentToResponseResult()` — normalizes content: string→JSON.parse, array→extract text blocks, object→pass-through
- Added `AIMessage.tool_calls` → `functionCall` parts conversion in legacy path
- Replaced 4× `ToolMessage.isInstance()` with `isToolMessageLike()`

**`@langchain/google-common` — `utils/gemini.ts`:**
- `Array.isArray()` guard in `toolMessageToContent()` prevents `.reduce()` crash on plain objects

**Tests (3 new):**
- Object content → preserved as object in `functionResponse.response.result`
- Array content → text blocks extracted and concatenated
- JSON string content → parsed into object

## Test Results
- `@langchain/google`: 115/115 passed ✅
- `@langchain/google-common`: 277/277 passed ✅

### DIFF
--- a/libs/providers/langchain-google/src/converters/messages.ts
+++ b/libs/providers/langchain-google/src/converters/messages.ts
@@ -508,9 +508,7 @@ function convertStandardContentMessageToGeminiContent(
 
   // Handle tool messages as function responses
   if (isToolMessageLike(message)) {
-    const responseContent = toolMessageContentToResponseResult(
-      message.content
-    );
+    const responseContent = toolMessageContentToResponseResult(message.content);
     // Find the matching tool call in a preceding AIMessage to get the function name
     const aiMsg = messages
       .filter(AIMessage.isInstance)
@@ -805,9 +803,7 @@ function convertLegacyContentMessageToGeminiContent(
 
   // Handle tool messages as function responses
   if (isToolMessageLike(message)) {
-    const responseContent = toolMessageContentToResponseResult(
-      message.content
-    );
+    const responseContent = toolMessageContentToResponseResult(message.content);
     // Find the matching tool call in a preceding AIMessage to get the function name
     const aiMsg = messages
       .filter(AIMessage.isInstance)
--- a/libs/providers/langchain-google/src/converters/tests/messages.test.ts
+++ b/libs/providers/langchain-google/src/converters/tests/messages.test.ts
@@ -756,7 +756,11 @@ describe("convertMessagesToGeminiContents", () => {
         ],
       }),
       new ToolMessage({
-        content: { status: "ok", value: 42, items: ["foo", "bar"] } as unknown as string,
+        content: {
+          status: "ok",
+          value: 42,
+          items: ["foo", "bar"],
+        } as unknown as string,
         tool_call_id: "tool-call-obj",
         name: "my_tool",
       }),


afirstenberg:
Thanks for this!

Two things up front:
* I'd like to see this split into two PRs - one for `@langchain/google-common` and one for `@langchain/google`. They are being maintained separately at this point and I don't want the review of one to block the other.
* I'd really like to see an integration test (particularly on `@langchain/google`) that demonstrates the problem pre-fix and proves the fix works.

alex-levkivskiy:
> Thanks for this!
> 
> Two things up front:
> 
> * I'd like to see this split into two PRs - one for `@langchain/google-common` and one for `@langchain/google`. They are being maintained separately at this point and I don't want the review of one to block the other.
> * I'd really like to see an integration test (particularly on `@langchain/google`) that demonstrates the problem pre-fix and proves the fix works.

Done! Split into two PRs and added an integration test:

- **`@langchain/google-common`**: #10547 — `Array.isArray` guard in `toolMessageToContent()`
- **`@langchain/google`**: #10548 — `isToolMessageLike()` fallback, content normalization, legacy AIMessage fix, unit tests + integration test

The integration test (`function reply with object tool content` in `index.int.test.ts`) sends a ToolMessage with plain object content through the full pipeline, then verifies the `functionResponse` is present in the request body with the object preserved.

Closing this PR in favor of the two above.

christian-bromann:
> Closing this PR in favor of the two above.

Thank you!

PR #10550, opened 2026-03-30
fix(google): align ChatGoogle mediaResolution with Gemini scalar type

### LINKED ISSUE #10546: Invalid `ChatGoogle.mediaResolution` global param type
### Checked other resources

### Example Code

```typescript

const model1 = new ChatGoogle({
  // Expected type: object with "level" property
  mediaResolution: { level: "MEDIA_RESOLUTION_HIGH" },
});
// Runtime error from Google API:
//   Invalid value at 'generation_config' (media_resolution), Starting an object on a scalar field

const model2 = new ChatGoogle({
  // @ts-expect-error — wrong type, but works at runtime
  mediaResolution: "MEDIA_RESOLUTION_HIGH",
});
```

When passing the value that satisfies the current type (`{ level: "MEDIA_RESOLUTION_HIGH" }`), the Google API returns:

```
Invalid value at 'generation_config' (media_resolution), Starting an object on a scalar field
```

### Description

When creating an instance of `GoogleChat`, the params object expect `mediaResolution` to be an object with a `level` property instead of a string.

This is invalid, and the Gemini API will accordingly reject the request.

Conversely, passing the appropriate string as `mediaResolution` fails to typecheck but works at runtime.

In [`types.ts`](https://github.com/langchain-ai/langchainjs/blob/50d5f32fd30cabebf058b1c13255c1daadde6107/libs/providers/langchain-google/src/chat_models/types.ts#L143):

```typescript
mediaResolution?: Prettify<Gemini.MediaResolution>;
```

`Gemini.MediaResolution` resolves in [`api-types.ts`](https://github.com/langchain-ai/langchainjs/blob/50d5f32fd30cabebf058b1c13255c1daadde6107/libs/providers/langchain-google/src/chat_models/api-types.ts#L47) to:

```typescript
export interface MediaResolution {
  level?: "MEDIA_RESOLUTION_UNSPECIFIED" | "MEDIA_RESOLUTION_LOW" | "MEDIA_RESOLUTION_MEDIUM" | "MEDIA_RESOLUTION_HIGH" | "MEDIA_RESOLUTION_ULTRA_HIGH" | (string & {});
}
```

In [`base.ts`](https://github.com/langchain-ai/langchainjs/blob/50d5f32fd30cabebf058b1c13255c1daadde6107/libs/providers/langchain-google/src/chat_models/base.ts#L368), the value is passed into `generationConfig` without unwrapping:

```typescript
...(fields.mediaResolution
  ? { mediaResolution: fields.mediaResolution }
  : {}),
```

However, according to [Gemini docs](https://ai.google.dev/gemini-api/docs/media-resolution#javascript_1), this setting should be a string from the [`MediaResolution` enum](https://github.com/googleapis/js-genai/blob/585c17747d903d599f8f24d204ac245661831698/src/types.ts#L682).

### System Info

```
$ pnpm info langchain

langchain@1.2.37 | MIT | deps: 5 | versions: 412

dist
.tarball: https://registry.npmjs.org/langchain/-/langchain-1.2.37.tgz
.shasum: 58a7c4ad408cd478b606f0d7f6ef802c00bf6fa9
.integrity: sha512-zjd0TpMfX8V440XJVavRmzLSvojVv5WZ7M9LWNtgYo0MDFTiWlCdnFNdHo1sZlNoI+8aSdLsVl/WXKtucyhwEA==
.unpackedSize: 2.8 MB

dependencies:
@langchain/langgraph-checkpoint: ^1.0.0 langsmith: >=0.5.0 <1.0.0               zod: ^3.25.76 || ^4                     
@langchain/langgraph: ^1.1.2            uuid: ^11.1.0                           

maintainers:
- davidduong <david@duong.cz>
- hntrl <hunter@hntrl.io>

dist-tags:
latest: 1.2.37                            next: 1.0.0-alpha.9                       tag-for-publishing-older-releases: 0.2.20 
dev: 1.3.0-dev-1773962633795              alpha: 1.0.0-alpha.9                      

published 5 days ago by GitHub Actions <npm-oidc-no-reply@github.com>

$ node --version
v24.14.0

$ pnpm --version
10.33.0

### PR DESCRIPTION
**Pull request by muhammadosama984**
## Summary
- Align `ChatGoogleFields.mediaResolution` with the Gemini `generationConfig.mediaResolution` scalar type (`GeminiBase.GenerationConfig[\"mediaResolution\"]`) instead of an object-shaped type.
- Add regression tests to verify `mediaResolution` serializes correctly from both constructor fields and per-invoke call options.
- Ensure this fixes the type/runtime mismatch reported in the issue.

Fixes #10546

## Test plan
- [x] `pnpm --filter @langchain/google test src/chat_models/tests/index.test.ts`
- [x] `pnpm --filter @langchain/google lint`
- [x] `pnpm test` (repo-wide currently fails on unrelated pre-existing tests in this environment)

### DIFF
--- a/.changeset/google-media-resolution-detail.md
+++ b/.changeset/google-media-resolution-detail.md
@@ -1,0 +1,5 @@
+---
+"@langchain/google": patch
+---
+
+fix(google): align `mediaResolution` with Gemini scalar values and support `detail` alias mapping (`low`/`high`/`auto`) for media prompts.

### CODE REVIEW

**Comment by afirstenberg** [comment]
Thanks, @muhammadosama984!

A few notes:
* I believe OpenAI uses "detail" for this. It might be good if we added a "detail" parameter as well that gets mapped into "mediaResolution".
* I see a regression test, but not an integration test. Please add an integration test.

Otherwise, I think this is good. It certainly took me a few tries to read through and realize that the `GenerationConfig` media resolution and the `Part` media resolution were two different creatures.

Because why make something like that consistent?

**Comment by muhammadosama984** [comment]
@afirstenberg Thanks a lot for taking a look and for the clear feedback!

Implemented as requested:
- Added an OpenAI-compatible `detail` parameter and mapped it to `generationConfig.mediaResolution` (with `mediaResolution` taking precedence when both are provided).
- Added integration coverage for the new `detail -> mediaResolution` behavior, along with regression tests.

Can you take a look?

File: libs/providers/langchain-google/src/chat_models/base.ts [hunk 1]
```diff
@@ -87,6 +87,21 @@ export function getPlatformType(
   }
 }
 
+function mapDetailToMediaResolution(
+  detail?: ChatGoogleFields["detail"]
+): Gemini.GenerationConfig["mediaResolution"] | undefined {
+  switch (detail) {
+    case "auto":
+      return "MEDIA_RESOLUTION_UNSPECIFIED";
```
afirstenberg: This isn't usually a valid value. It is just a side effect of the auto-generation of the REST interface based on gRPC.

"auto" should probably map to the default value by leaving it undefined.

libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
```diff
@@ -473,6 +473,14 @@ describe.each(coreModelInfo)(
       expect(contentBlock.text).toMatch(/(1 + 1 (equals|is|=) )?2.? ?/);
     });
 
+    test("detail maps to generationConfig.mediaResolution", async () => {
```
afirstenberg: More importantly, I'd like to see the integration test for `mediaResolution`. Would be good if it was included with the other image tests since this doesn't really mean anything for non-media prompts.

Should probably check the results to at least make sure there are image tokens that were processed. (We can't really test how many, since that will vary too much based on model, but may be worth making sure that high > low or something. Or may be more trouble than it is worth.)

libs/providers/langchain-google/src/chat_models/base.ts
```diff
@@ -306,6 +321,8 @@ export abstract class BaseChatGoogle<
 
   override invocationParams(options: this["ParsedCallOptions"]) {
     const fields = combineGoogleChatModelFields(this.params, options);
+    const mediaResolution =
+      fields.mediaResolution ?? mapDetailToMediaResolution(fields.detail);
```
afirstenberg, line 323: Looks good!

muhammadosama984:
@afirstenberg Thank you for the thoughtful review and guidance.

I have addressed all requested items:
- Updated `detail: "auto"` to use default Gemini behavior by omitting `generationConfig.mediaResolution`.
- Added/updated regression coverage to verify the `auto` behavior and preserve `mediaResolution` precedence when both options are provided.
- Reworked integration coverage to validate media-resolution behavior on image prompts (including request serialization and image-token usage assertions).

I appreciate your feedback and would be grateful for a re-review when you have a moment.

afirstenberg:
- [X] Integration test passes
- [X] Code looks good.

- [ ] You should add a changeset.

Otherwise, LGTM
(cc: @hntrl @christian-bromann)

muhammadosama984:
@afirstenberg changeset added in `80259da3c`.

Added file: `.changeset/google-media-resolution-detail.md` (`"@langchain/google": patch`) to cover this fix.

muhammadosama984:
@christian-bromann @hntrl @afirstenberg 
Just checking in on this PR. Let me know if any changes are needed — happy to update it. Thanks!

christian-bromann (approved):
LGTM 👍

muhammadosama984:
@christian-bromann @afirstenberg @hntrl  there were some formatting checks failing, I have fixed those can you please re-review. 
Thanks!!

afirstenberg:
Still LGTM

christian-bromann:
Thank you for the patch @muhammadosama984 and the review @afirstenberg 🙏

PR #10559, opened 2026-03-31
feat(openrouter): add `appCategories` and update attribution defaults

mdrxy:
Port attribution header changes from the Python `langchain-openrouter` package. The default `siteUrl` and `siteName` are updated to match the canonical values (`https://docs.langchain.com` and `LangChain`), and a new `appCategories` field enables marketplace attribution via the `X-OpenRouter-Categories` header. `buildHeaders()` now conditionally includes each attribution header only when set, rather than always emitting them.

### DIFF
--- a/libs/providers/langchain-openrouter/src/chat_models/tests/index.test.ts
+++ b/libs/providers/langchain-openrouter/src/chat_models/tests/index.test.ts
@@ -124,7 +124,6 @@ describe("ChatOpenRouter constructor", () => {
   });
 });
 
-// ─── attribution headers ─────────────────────────────────────────────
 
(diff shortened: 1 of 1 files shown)


libs/providers/langchain-openrouter/src/chat_models/tests/index.test.ts
```diff
@@ -111,6 +124,79 @@ describe("ChatOpenRouter constructor", () => {
   });
 });
 
+// ─── attribution headers ─────────────────────────────────────────────
```
hntrl:
```suggestion
```

PR #10587, opened 2026-04-01
fix(community): expose basePath in VoyageEmbeddings constructor params

tysoncung:
## Problem

The `VoyageEmbeddings` class has a `basePath` property that defaults to `https://api.voyageai.com/v1`, but it is not exposed in `VoyageEmbeddingsParams`. This means:

1. `basePath` cannot be set via the constructor
2. Setting `basePath` after instantiation has no effect because `apiUrl` is already computed from the default

```typescript
const embeddings = new VoyageEmbeddings({
  apiKey: "key",
  basePath: "https://custom-endpoint.com/v1", // TS error + ignored
});
```

## Solution

- Add `basePath` to `VoyageEmbeddingsParams` interface
- Apply `basePath` from constructor fields before computing `apiUrl`

## Files Changed

- `libs/community/langchain-community/src/embeddings/voyage.ts`

Closes langchain-ai/langchainjs-community#20

### DIFF
--- a/libs/community/langchain-community/src/embeddings/voyage.ts
+++ b/libs/community/langchain-community/src/embeddings/voyage.ts
@@ -16,6 +16,12 @@ export interface VoyageEmbeddingsParams extends EmbeddingsParams {
   batchSize?: number;
 
   /**
+   * Base URL for the Voyage AI API.
+   * @default "https://api.voyageai.com/v1"
+   */
+  basePath?: string;
+
+  /**
    * Input type for the embeddings request.
    */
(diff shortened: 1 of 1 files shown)


hntrl (commented):
Thank you for your contribution! As announced in https://github.com/langchain-ai/langchainjs-community/issues/61, the @langchain/community package is being sunset, and we are closing the open PRs as part of that wind-down.

For context and next steps (including how to create a standalone integration package), please see https://github.com/langchain-ai/langchainjs-community/issues/61.

PR #10598, opened 2026-04-01
fix(google): fail fast for Claude models on Vertex AI

### LINKED ISSUE #10596: docs: (@langchain/google) No usage example for Claude with VertexAI
### Checklist

- [x] I included a link to the documentation page I am referring to (if applicable).

### Issue with current documentation:

The @langchain/google integration indicates that it's possible to use claude models with the integration, but there's no example of how this is done. Inputting a claude model version from vertex ai into the model field results in an error.

https://docs.langchain.com/oss/javascript/integrations/providers/google#third-party-models-on-vertex-ai

Additionally, there doesn't seem to be a way to change the publisher from google to anthropic
https://github.com/langchain-ai/langchainjs/blob/12f8550a0257938a79f14eccfb90863ff38b95f6/libs/providers/langchain-google/src/chat_models/base.ts#L244-L246

### Idea or request for content:

_No response_

muhammadosama984:
## Summary

This PR addresses confusing Claude-on-Vertex behavior in `@langchain/google` by failing fast with actionable guidance.

Users currently see docs mention Claude on Vertex AI and may try `ChatGoogle` directly. Internally, `ChatGoogle` is Gemini-oriented (`generateContent` flow), so Claude model usage fails in a non-obvious way. This change makes the behavior explicit.

## Changes

- Add Claude model detection guard for Vertex (`platformType: "gcp"`) in `BaseChatGoogle`.
- Throw a clear `ConfigurationError` with migration guidance to:
  - `@langchain/anthropic`
  - `@anthropic-ai/vertex-sdk` via `ChatAnthropic.createClient`
- Add unit tests:
  - Claude + Vertex throws actionable error
  - Gemini + Vertex behavior remains unchanged (URL regression check)
- Add integration test:
  - deterministic fail-fast assertion for Claude + Vertex in int mode
- Add changeset:
  - patch release note for `@langchain/google`

## Testing

- `pnpm --filter @langchain/google test src/chat_models/tests/index.test.ts`
- `pnpm vitest run --mode int src/chat_models/tests/index.int.test.ts -t "throws actionable error before request for ChatGoogle"`

## Notes

- Running full `pnpm --filter @langchain/google test` in this environment also surfaced two existing converter tests failing in `src/converters/tests/messages.test.ts` (legacy-path assertions), unrelated to this change.

Fixes #10596

### DIFF
--- a/.changeset/chatgoogle-claude-vertex-guard.md
+++ b/.changeset/chatgoogle-claude-vertex-guard.md
@@ -1,0 +1,7 @@
+---
+"@langchain/google": patch
+---
+
+fix(@langchain/google): fail fast for Claude models on Vertex with actionable guidance
+
+`ChatGoogle` now throws a clear `ConfigurationError` when a Claude model is used on Vertex AI, and points users to `@langchain/anthropic` with the Vertex SDK path. Added unit and integration tests to keep this behavior deterministic.
--- a/libs/providers/langchain-google/src/chat_models/base.ts
+++ b/libs/providers/langchain-google/src/chat_models/base.ts
@@ -87,6 +87,10 @@ export function getPlatformType(
   }
 }
 
+function isClaudeModelName(modelName: string): boolean {
+  return /(^|[/:_-])claude([/:_-]|$)/i.test(modelName);
+}
+
 export interface BaseChatGoogleParams
   extends BaseChatModelParams,
     ChatGoogleFields {
@@ -245,6 +249,18 @@ export abstract class BaseChatGoogle<
     return "google";
   }
 
+  protected assertModelSupportedForVertex(): void {
+    if (this.platform !== "gcp") {
+      return;
+    }
+
+    if (isClaudeModelName(this.model)) {
+      throw new ConfigurationError(
+        "Claude models on Vertex AI are not supported by ChatGoogle in @langchain/google. Use @langchain/anthropic with @anthropic-ai/vertex-sdk (via ChatAnthropic createClient) for Claude on Vertex AI."
+      );
+    }
+  }
+
   protected get urlMethod(): string {
     return this.streaming ? "streamGenerateContent?alt=sse" : "generateContent";
   }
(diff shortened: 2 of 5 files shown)


muhammadosama984:
@afirstenberg Hi! 👋 This PR was created to address the issue you raised in #10596 — it adds a fail-fast guard with clear migration guidance when trying to use Claude models via `ChatGoogle` on Vertex AI, since `ChatGoogle` is Gemini-oriented under the hood.

Just wanted to check in with you: does this approach align with what you had in mind? If you feel this resolves the concern, I would love to get your thoughts before moving forward with the merge. Happy to make any adjustments if you'd like!

afirstenberg:
(Also, for the record, #10596 isn't my issue. @MarkKoester raised itl.)

guidorietbroek:
Why not accepting other providers? What will be the recommended way to use chatvertex ai with models not from google? Langchain is deprecating the old vertex ai version, so nothing is left?

guidorietbroek:
To clarify my question: the new `@langchain/google package` (which the  docs say replaces @langchain/google-vertexai) has publisher hardcoded to "google" in ChatGoogle (base.ts#L244-L246), so Claude models on Vertex don't work there. 

And `@langchain/google-vertexai / ChatVertexAI`, the only current path that does route claude-* models to publishers/anthropic is marked as deprecated in the ChatVertexAI docs. 

What is the supported path for Claude-on-Vertex going forward? Is the plan to add non-Google publisher support to @langchain/google, keep @langchain/google-vertexai alive indefinitely for this use case, or something else? 

@jacoblee93 
@christian-bromann

afirstenberg:
@guidorietbroek rightfully asked:
> What is the supported path for Claude-on-Vertex going forward? Is the plan to add non-Google publisher support to @langchain/google, keep @langchain/google-vertexai alive indefinitely for this use case, or something else?

The long term plan is to support everything in the Vertex AI Model Garden, including Anthropic models, through `@langchain/google`. At that point, `@langchain/google` should be a drop-in replacement for the older Google libraries.

The short term plan is to get `@langchain/google` fully stable with Gemini and then tackle other models and interfaces.

I hope that provides sufficient clarity. I wish I had a better idea of a time-frame, but this should give you an idea what the plan is.

guidorietbroek:
Thanks @afirstenberg , this will helps us in deciding to do for the moment.

PR #10657, opened 2026-04-03
feat(mongodb): add mongodb automated embedding support

PavelSafronov:
This PR adds MongoDB Automated Embedding support.

Issue: https://jira.mongodb.org/browse/NODE-7503
[This MongoDB Automated Embedding tutorial](https://www.mongodb.com/docs/atlas/atlas-vector-search/crud-embeddings/create-embeddings-automatic/?interface=driver&language=nodejs) explains how auto embedding works and how it can be used by the MongoDB Node Driver. This same approach was used to allow `MongoDBAtlasVectorSearch` to utilize Automated Embedding. Automated Embedding is supported in MongoDB 8.2 and higher.

### DIFF
--- a/libs/providers/langchain-mongodb/src/tests/utils.ts
+++ b/libs/providers/langchain-mongodb/src/tests/utils.ts
@@ -1,6 +1,7 @@
 import { Collection } from "mongodb";
 import { Readable } from "stream";
 import { setInterval } from "timers/promises";
+import type { MongoDBAtlasVectorSearch } from "../vectorstores.js";
 
 export function isUsingLocalAtlas() {
   // oxlint-disable-next-line no-process-env
@@ -23,7 +24,36 @@ export async function waitForIndexToBeQueryable(
 ): Promise<void> {
   return Readable.from(setInterval(5000, undefined, { ref: false }))
     .map(() => collection.listSearchIndexes().toArray())
-    .find((indexes: { name: string; queryable?: boolean }[]) =>
-      indexes.some((index) => index.name === indexName && index.queryable)
-    );
+    .find((indexes: { name: string; status: string; }[]) => {
+      const found = indexes.some((index) => index.name === indexName && index.status === "READY");
+      return found;
+    });
+}
+
+/**
+ * Wait for documents to be indexed and searchable after insertion.
+ * Polls the vector store with a test query until search succeeds,
+ * indicating that auto-embeddings are complete.
+ */
+export async function waitForDocumentsIndexed(
+  vectorStore: MongoDBAtlasVectorSearch,
+  testQuery: string,
+  maxWaitTime = 60000
+): Promise<void> {
+  const startTime = Date.now();
+
+  while (Date.now() - startTime < maxWaitTime) {
+    const result = await vectorStore.similaritySearch(testQuery, 1);
+    if (result.length > 0) {
+      return; // Search succeeded, documents are indexed
+    } else {
+      // Documents not yet indexed, retry after a short delay
+      await new Promise(resolve => setTimeout(resolve, 5000));
+    }
+  }
+
+  throw new Error(
+    `Documents not indexed after ${maxWaitTime}ms for query: "${testQuery}". ` +
+    `Auto-embedding may have failed or the timeout is too short.`
+  );
 }
--- a/libs/providers/langchain-mongodb/src/tests/utils.ts
+++ b/libs/providers/langchain-mongodb/src/tests/utils.ts
@@ -24,8 +24,10 @@ export async function waitForIndexToBeQueryable(
 ): Promise<void> {
   return Readable.from(setInterval(5000, undefined, { ref: false }))
     .map(() => collection.listSearchIndexes().toArray())
-    .find((indexes: { name: string; status: string; }[]) => {
-      const found = indexes.some((index) => index.name === indexName && index.status === "READY");
+    .find((indexes: { name: string; status: string }[]) => {
+      const found = indexes.some(
+        (index) => index.name === indexName && index.status === "READY"
+      );
       return found;
     });
 }
@@ -48,12 +50,12 @@ export async function waitForDocumentsIndexed(
       return; // Search succeeded, documents are indexed
     } else {
       // Documents not yet indexed, retry after a short delay
-      await new Promise(resolve => setTimeout(resolve, 5000));
+      await new Promise((resolve) => setTimeout(resolve, 5000));
     }
   }
 
   throw new Error(
     `Documents not indexed after ${maxWaitTime}ms for query: "${testQuery}". ` +
-    `Auto-embedding may have failed or the timeout is too short.`
+      `Auto-embedding may have failed or the timeout is too short.`
   );
 }
--- a/libs/providers/langchain-mongodb/src/tests/vectorstores_auto_embed.int.test.ts
+++ b/libs/providers/langchain-mongodb/src/tests/vectorstores_auto_embed.int.test.ts
@@ -141,28 +141,50 @@ describe.skipIf(skipAutoEmbedTests)("Auto Embedding Tests with data", () => {
     expect(results).toMatchObject([
       { pageContent: "What is a sandwich?", metadata: { c: 1 } },
     ]);
+  });
 
-    // // we can pre filter the search
-    // const preFilter = {
-    //   e: { $lte: 1 },
-    // };
+  test("MongoDBAtlasVectorSearch similaritySearchWithScore", async () => {
+    const vectorStore = new MongoDBAtlasVectorSearch({ collection });
 
-    // const filteredResults = await vectorStore.similaritySearch(
-    //   "That fence is purple",
-    //   1,
-    //   preFilter
-    // );
+    expect(vectorStore).toBeDefined();
 
-    // expect(filteredResults).toEqual([]);
+    // check if the database is empty
+    await collection.deleteMany({});
 
-    // const retriever = vectorStore.asRetriever({
-    //   filter: {
-    //     preFilter,
-    //   },
-    // });
+    await vectorStore.addDocuments([
+      {
+        pageContent: "Dogs are tough.",
+        metadata: { a: 1, created_at: new Date().toISOString() },
+      },
+      {
+        pageContent: "Cats have fluff.",
+        metadata: { b: 1, created_at: new Date().toISOString() },
+      },
+      {
+        pageContent: "What is a sandwich?",
+        metadata: { c: 1, created_at: new Date().toISOString() },
+      },
+      {
+        pageContent: "That fence is purple.",
+        metadata: { d: 1, e: 2, created_at: new Date().toISOString() },
+      },
+    ]);
 
-    // const docs = await retriever.getRelevantDocuments("That fence is purple");
-    // expect(docs).toEqual([]);
+    await waitForDocumentsIndexed(vectorStore, "Sandwich");
+
+    const results = await vectorStore.similaritySearchWithScore(
+      "Sandwich",
+      1
+    );
+
+    expect(results.length).toEqual(1);
+    expect(results[0]).toHaveLength(2);
+    expect(results[0][0]).toMatchObject({
+      pageContent: "What is a sandwich?",
+      metadata: { c: 1 },
+    });
+    expect(typeof results[0][1]).toBe("number");
+    expect(results[0][1]).toBeGreaterThan(0);
   });
 
   // Maximal Marginal Relevance is not yet supported with auto-embedding.
--- a/libs/providers/langchain-mongodb/src/tests/vectorstores_auto_embed.int.test.ts
+++ b/libs/providers/langchain-mongodb/src/tests/vectorstores_auto_embed.int.test.ts
@@ -172,10 +172,7 @@ describe.skipIf(skipAutoEmbedTests)("Auto Embedding Tests with data", () => {
 
     await waitForDocumentsIndexed(vectorStore, "Sandwich");
 
-    const results = await vectorStore.similaritySearchWithScore(
-      "Sandwich",
-      1
-    );
+    const results = await vectorStore.similaritySearchWithScore("Sandwich", 1);
 
     expect(results.length).toEqual(1);
     expect(results[0]).toHaveLength(2);
--- a/libs/providers/langchain-mongodb/src/tests/vectorstores_auto_embed.int.test.ts
+++ b/libs/providers/langchain-mongodb/src/tests/vectorstores_auto_embed.int.test.ts
@@ -1,0 +1,650 @@
+import {
+  afterAll,
+  beforeAll,
+  beforeEach,
+  describe,
+  expect,
+  vi,
+  test,
+} from "vitest";
+import { Collection, MongoClient } from "mongodb";
+import { Document } from "@langchain/core/documents";
+import { Document as BSONDocument } from "bson";
+import semver from "semver";
+
+import { MongoDBAtlasVectorSearch } from "../vectorstores.js";
+import {
+  uri,
+  waitForIndexToBeQueryable,
+  waitForDocumentsIndexed,
+} from "./utils.js";
+
+/**
+ * The following json can be used to create an index in atlas for Cohere embeddings.
+ * Use "langchain.test" for the namespace and "default" for the index name.
+
+{
+  "mappings": {
+    "fields": {
+      "e": { "type": "number" },
+      "embedding": {
+        "dimensions": 1536,
+        "similarity": "euclidean",
+        "type": "knnVector"
+      }
+    }
+  }
+}
+*/
+
+const tempClient = new MongoClient(uri());
+await tempClient.connect();
+const admin = tempClient.db().admin();
+const serverStatus = await admin.serverStatus();
+const version = serverStatus.version || "unknown";
+const runAutoEmbedTests = semver.gte(version, "8.2.0");
+const skipAutoEmbedTests = !runAutoEmbedTests;
+await tempClient.close();
+
+describe.skipIf(skipAutoEmbedTests)("Auto Embedding Tests with data", () => {
+  const defaultEmbeddingModel = "voyage-4";
+
+  let client: MongoClient;
+  let collection: Collection;
+  beforeAll(async () => {
+    client = new MongoClient(uri(), { monitorCommands: true });
+    await client.connect();
+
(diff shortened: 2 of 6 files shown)


libs/providers/langchain-mongodb/src/vectorstores.ts
```diff
@@ -64,21 +83,73 @@ export class MongoDBAtlasVectorSearch extends VectorStore {
...
+    args: MongoDBAtlasVectorSearchLibArgs
+  );
   constructor(
     embeddings: EmbeddingsInterface,
     args: MongoDBAtlasVectorSearchLibArgs
+  );
+  constructor(
+    embeddingModelName: string,
+    args: MongoDBAtlasVectorSearchLibArgs
+  );
```
RaschidJFR: Should the user be allowed to pass `MongoDBAtlasVectorSearchLibArgs` when using autoembeddings? wouldn't  `textKey` and `embeddingKey` be taken care of by the server?

libs/providers/langchain-mongodb/src/vectorstores.ts
```diff
@@ -64,21 +83,73 @@ export class MongoDBAtlasVectorSearch extends VectorStore {
...
+      useAutoEmbedding = true;
+    } else if (args !== undefined) {
+      // (embeddings, args) - traditional format where embeddings are provided by the user
+      embeddings = embeddingsModelNameOrArgs as EmbeddingsInterface;
+      libArgs = args;
+    } else {
+      // (args) - new format for auto-embedding mode, where the server handles embeddings and no model name is specified (defaults to 'voyage-4')
+      libArgs = embeddingsModelNameOrArgs as MongoDBAtlasVectorSearchLibArgs;
+      embeddings = new AutoEmbeddingStub();
+      embeddingModelName = 'voyage-4'; // default embedding model for auto-embedding mode
```
tadjik1: Let's not default to this model if none is provided, this parameter is optional for queryText and must be omitted for queryVector. Whilst it might look harmful (we only use queryText for now), the user will experience increased costs if the desire was to use -light model during index time and assumption was that query will use the same class of models.
```suggestion
      
```

libs/providers/langchain-mongodb/src/vectorstores.ts
tadjik1: Good catch, exposing `embeddingKey` is misleading here, as it will be handled internally by the database. `textKey` is still relevant, though (this is where `.pageContent` is stored).

libs/providers/langchain-mongodb/src/vectorstores.ts
```diff
@@ -64,21 +83,73 @@ export class MongoDBAtlasVectorSearch extends VectorStore {
...
   _vectorstoreType(): string {
     return "mongodb_atlas";
   }
 
+  /**
+   * Constructor with function overloads for backward compatibility.
+   * (embeddings, args) - traditional format where embeddings are provided by the user
+   * (args) - new format for auto-embedding mode, where the server handles embeddings and no model name is specified (defaults to 'voyage-4')
+   * (embeddingModelName, args) - new format for auto-embedding mode with a specified model name
+   */
```
tadjik1, line 95: I'd avoid "traditional", "new", "legacy", "classic" - those imply one is replacing the other, which isn't the case. Both are valid modes that will coexist long-term. We should rather use tech/product terms: [auto-embeddings vs. manual](https://www.mongodb.com/docs/atlas/atlas-vector-search/crud-embeddings/create-embeddings-automatic/?interface=driver&language=nodejs) (or client-side).

libs/providers/langchain-mongodb/src/vectorstores.ts
```diff
@@ -64,21 +83,73 @@ export class MongoDBAtlasVectorSearch extends VectorStore {
...
   constructor(
     embeddings: EmbeddingsInterface,
     args: MongoDBAtlasVectorSearchLibArgs
+  );
+  constructor(
+    embeddingModelName: string,
+    args: MongoDBAtlasVectorSearchLibArgs
+  );
+  constructor(
+    embeddingsModelNameOrArgs: EmbeddingsInterface | string | MongoDBAtlasVectorSearchLibArgs,
```
tadjik1: Did we consider simplifying this approach, instead of having 3 overloads limit it to 2, i.e. before the change:
```
constructor(
  args: MongoDBAtlasVectorSearchLibArgs
)
```
after the change:
```
// auto-embed, defaults to voyage-4
constructor(
  args: MongoDBAtlasVectorSearchLibArgs
);
// manual
constructor(
  args: MongoDBAtlasVectorSearchLibArgs
);
// auto-embed with model name - can be removed?
constructor(
  embeddingModelName: string, 
  args: MongoDBAtlasVectorSearchLibArgs
);
```

We can probably omit 3rd overload (where model is provided with the first argument) and instead extend `MongoDBAtlasVectorSearchLibArgs` further by adding optional `.model` or `.embeddingModelName` property there, what do you think?
This would also simplify `fromTexts/fromDocuments` which currently need the same three-way dispatch logic.

libs/providers/langchain-mongodb/src/vectorstores.ts
```diff
@@ -64,21 +83,73 @@ export class MongoDBAtlasVectorSearch extends VectorStore {
...
+      if (args === undefined) {
+        throw new Error(
+          "When using the new constructor convention (embeddingModelName, args), both parameters must be provided."
+        );
+      }
+      embeddingModelName = embeddingsModelNameOrArgs;
+      embeddings = new AutoEmbeddingStub();
+      libArgs = args;
+      useAutoEmbedding = true;
+    } else if (args !== undefined) {
```
tadjik1: Probably duck-type here would be safer, maybe validate by picking well-known property:
```
if ("collection" in embeddingsModelNameOrArgs) {
  // auto embeddings
} else { // optionally check existence of `.embedDocuments`
  // manual embeddings
}
``

File: libs/providers/langchain-mongodb/src/vectorstores.ts [hunk 6]
```diff
@@ -299,25 +540,86 @@ export class MongoDBAtlasVectorSearch extends VectorStore {
       });
       docs.push(newDoc);
     }
-    return this.fromDocuments(docs, embeddings, dbConfig);
+
+    // Detect which calling convention is being used
+    if (dbConfig !== undefined) {
+      // `fromTexts(texts, metadatas, embeddings, dbConfig)` (traditional)
```
tadjik1: Looking at the jsdoc for this method it seems that the auto-embedding way would also use dbConfig, making code down below is unreachable for this scenario:
```
* - `fromTexts(texts, metadatas, embeddingModelName, dbConfig)` (auto-embedding mode with specified model name)
```

libs/providers/langchain-mongodb/src/vectorstores.ts
```diff
@@ -299,25 +540,86 @@ export class MongoDBAtlasVectorSearch extends VectorStore {
...
+    dbConfig?: MongoDBAtlasVectorSearchLibArgs & { ids?: string[] }
   ): Promise<MongoDBAtlasVectorSearch> {
-    const instance = new this(embeddings, dbConfig);
-    await instance.addDocuments(docs, { ids: dbConfig.ids });
+    let embeddings: EmbeddingsInterface;
+    let finalDbConfig: MongoDBAtlasVectorSearchLibArgs & { ids?: string[] };
+
+    let instance;
+    // Detect which calling convention is being used
+    if (dbConfig !== undefined) {
```
tadjik1, line 573: Same code as above, this check makes auto-embedding scenario with `else if (typeof embeddingsModelNameOrDbConfig === "string") {` unreachable.

libs/providers/langchain-mongodb/src/vectorstores.ts
```diff
@@ -200,25 +311,138 @@ export class MongoDBAtlasVectorSearch extends VectorStore {
...
+   * @param filter - Optional filter based on `FilterType`.
+   * @param _callbacks - Optional callbacks for monitoring search progress
+   * @returns A promise resolving to an array of tuples, each containing a
+   *          document and its similarity score.
+   */
+  async similaritySearchWithScore(
+    query: string,
+    k = 4,
+    filter: this["FilterType"] | undefined = undefined,
+    _callbacks: Callbacks | undefined = undefined // implement passing to embedQuery later
```
tadjik1: Can you elaborate please why is this parameter ignored for now, and create follow-up ticket?

libs/providers/langchain-mongodb/src/vectorstores.ts
```diff
@@ -200,25 +311,138 @@ export class MongoDBAtlasVectorSearch extends VectorStore {
...
+    const resultsWithScore = await this.similaritySearchWithScore(query, k ?? 4, filter);
+    return resultsWithScore.map(([doc]) => doc);
+  }
+
+  private async textBasedSearchWithScore(
+    query: string,
+    k: number,
+    filter?: MongoDBAtlasFilter
+  ): Promise<[Document, number][]> {
+    if (this.useAutoEmbedding && !this.embeddingModelName) {
```
tadjik1: This guard seems unreachable as we always set `embeddingModelName`, but nevertheless it looks unnecessary, as `model` parameter is optional for queryText (see my comment above for more information). I would recommend removing it.

libs/providers/langchain-mongodb/src/tests/utils.ts
```diff
@@ -23,7 +24,36 @@ export async function waitForIndexToBeQueryable(
 ): Promise<void> {
   return Readable.from(setInterval(5000, undefined, { ref: false }))
     .map(() => collection.listSearchIndexes().toArray())
-    .find((indexes: { name: string; queryable?: boolean }[]) =>
-      indexes.some((index) => index.name === indexName && index.queryable)
-    );
+    .find((indexes: { name: string; status: string; }[]) => {
+      const found = indexes.some((index) => index.name === indexName && index.status === "READY");
```
tadjik1: @PavelSafronov can you elaborate please on this change?

libs/providers/langchain-mongodb/src/tests/vectorstores_auto_embed.int.test.ts
```diff
@@ -0,0 +1,639 @@
...
+    await waitForDocumentsIndexed(vectorStore, "Sandwich");
+
+    const results: Document[] = await vectorStore.similaritySearch("Sandwich", 1);
+
+    expect(results.length).toEqual(1);
+    expect(results).toMatchObject([
+      { pageContent: "What is a sandwich?", metadata: { c: 1 } },
+    ]);
+
+    // // we can pre filter the search
```
tadjik1: We should either implement or remove this block.

libs/providers/langchain-mongodb/src/tests/vectorstores_auto_embed.int.test.ts
```diff
@@ -0,0 +1,639 @@
...
+            model: defaultEmbeddingModel,
+          },
+        ],
+      },
+    });
+
+    await waitForIndexToBeQueryable(collection, "default");
+  });
+
+  const shouldClear = true;
```
tadjik1: Not sure how it's handled here, but I guess it's not needed since it's defined as `const`?

libs/providers/langchain-mongodb/src/tests/vectorstores_auto_embed.int.test.ts
```diff
@@ -0,0 +1,639 @@
...
+          .find({}, { projection: { text: 1, _id: 0 } })
+          .toArray();
+
+        expect(results).toEqual([{ text: "updated 1" }, { text: "updated 2" }]);
+      });
+    });
+  });
+});
+
+describe("Auto Embedding Tests without data", () => {
```
tadjik1, line 627: I assume this also should be guarded by 8.2+?

libs/providers/langchain-mongodb/src/tests/vectorstores_auto_embed.int.test.ts
```diff
@@ -0,0 +1,639 @@
...
+  test("MongoDBStore sets client metadata", () => {
+    const spy = vi.spyOn(client!, "appendMetadata");
+    new MongoDBAtlasVectorSearch(defaultEmbeddingModel, {
+      collection,
+    });
+    expect(spy).toHaveBeenCalledWith({ name: "langchainjs_vector" });
+    vi.clearAllMocks();
+  });
+
+  test("MongoDBAtlasVectorSearch with external ids", async () => {
```
tadjik1, line 106: Can you please add similar test for `similaritySearchWithScore`?

tadjik1 (commented):
Thanks @PavelSafronov , the general approach looks good - the separation between auto-embedding and manual embedding modes is clean, and the integration test coverage is also good.
I've spotted a few areas that might need attention, please let me know what you think.

libs/providers/langchain-mongodb/src/vectorstores.ts
tadjik1: Good catch, exposing `embeddingKey` will be misleading here, as it will be handled internally by the database. `textKey` is still relevant, though (this is where `.pageContent` is stored).
Maybe we can introduce another interface w/o `embeddingKey`, something like:

```
interface MongoDBAutoEmbedSearchLibArgs extends AsyncCallerParams {
  readonly collection: Collection<MongoDBDocument>;
  readonly indexName?: string;
  readonly textKey?: string;
  readonly primaryKey?: string;
  readonly embeddingModelName?: string;  // see my comment below for this property
}
```

PavelSafronov: Don't think we need to add embeddingModelName anywhere. It is being removed entirely, it's not necessary when using auto-embed.

libs/providers/langchain-mongodb/src/vectorstores.ts
PavelSafronov: embeddingModelName is being removed entirely, it's not necessary when using auto-embed.

libs/providers/langchain-mongodb/src/vectorstores.ts
PavelSafronov: Excellent point, removing "traditional" terms.

libs/providers/langchain-mongodb/src/vectorstores.ts
PavelSafronov: embeddingModelName is being removed entirely, it's not necessary when using auto-embed, so the extra overload is being dropped.

libs/providers/langchain-mongodb/src/vectorstores.ts
PavelSafronov: embeddingModelName is being removed entirely, it's not necessary when using auto-embed.
With the removal of the string, this complex check is reduced to a null check.

libs/providers/langchain-mongodb/src/tests/utils.ts
PavelSafronov: [Our auto-embed instructions](https://www.mongodb.com/docs/atlas/atlas-vector-search/crud-embeddings/create-embeddings-automatic/?interface=driver&language=nodejs) suggest using `index.queryable` to figure out if an index is ready, but that's not true. We should instead be checking the `status` field on the index.

libs/providers/langchain-mongodb/src/tests/vectorstores_auto_embed.int.test.ts
PavelSafronov: Removing.

libs/providers/langchain-mongodb/src/tests/vectorstores_auto_embed.int.test.ts
PavelSafronov: These are client-side tests, so it's irrelevant if the server has 8.2 or not.

libs/providers/langchain-mongodb/src/vectorstores.ts
PavelSafronov: Updated, this should be resolved, since the modelName was removed as a possible overload.

libs/providers/langchain-mongodb/src/vectorstores.ts
PavelSafronov: Should be resolved now.

libs/providers/langchain-mongodb/src/vectorstores.ts
PavelSafronov: modelName is removed, this should be resolved.

libs/providers/langchain-mongodb/src/vectorstores.ts
PavelSafronov: Leftover from testing, removing.

libs/providers/langchain-mongodb/src/tests/vectorstores_auto_embed.int.test.ts
PavelSafronov: Removing.

libs/providers/langchain-mongodb/src/tests/vectorstores_auto_embed.int.test.ts
PavelSafronov: Test added.

libs/providers/langchain-mongodb/src/vectorstores.ts
```diff
@@ -64,21 +82,52 @@ export class MongoDBAtlasVectorSearch extends VectorStore {
...
+    super(embeddings, libArgs);
+    this.collection = libArgs.collection;
+    this.indexName = libArgs.indexName ?? "default";
+    this.textKey = libArgs.textKey ?? "text";
+    this.embeddingKey = libArgs.embeddingKey ?? "embedding";
+    this.primaryKey = libArgs.primaryKey ?? "_id";
+    this.caller = new AsyncCaller(libArgs);
+    this.useAutoEmbedding = useAutoEmbedding;
     this.collection.db.client.appendMetadata({
       name: "langchainjs_vector",
```
tadjik1, line 130: @RaschidJFR is it enough to send generic name, or `_auto|_manual` would be better? So we can differentiate auto embedding from manual ones.

RaschidJFR: We don't need to make that distinction. It's enough to identify the wrapping library (langchain vector) regardless of the mode (auto vs manual).

hntrl (dismissed):

hntrl (approved):

PR #10676, opened 2026-04-08
fix(google-common): auto-set includeServerSideToolInvocations when mixing built-in and function tools

### LINKED ISSUE #10675: Support includeServerSideToolInvocations flag for Google
### Example Code

  ```typescript
  import { ChatGoogle } from '@langchain/google';
  import { z } from 'zod';

  const myTool = tool(
    async ({ query }) => `Result for ${query}`,
    {
      name: 'my_tool',
      description: 'A custom tool',
      schema: z.object({ query: z.string() }),
    },
  );

  const model = new ChatGoogle({ model: 'gemini-2.5-flash' });

  // This fails with 400 Bad Request
  const response = await model.bindTools([myTool, { googleSearch: {} }]).invoke(
    'Search for the latest news about AI',
  );
  ```

  ```
  [GoogleGenerativeAI Error]: Error fetching from https://generativelanguage.googleapis.com/v1beta/models/gemini-2.5-flash:streamGenerateContent?alt=sse: [400 Bad Request] Please enable
  tool_config.include_server_side_tool_invocations to use Built-in tools with Function calling.
```

### Description

When using `ChatGoogle` from `@langchain/google` with both a built-in/server-side tool (e.g. `{ googleSearch: {} }`) and function calling tools, the Google Generative AI API returns a 400
  error requiring `tool_config.include_server_side_tool_invocations` to be set to `true`.

As per the docs (https://ai.google.dev/gemini-api/docs/tool-combination) we should be able to pass `includeServerSideToolInvocations: true` to the tool config however it isnt possible to explicitly set this nor is it applied when a mix of built-in and custom tools is processed.

### System Info

  - `@langchain/google`: 0.1.7
  - Node.js: v22

### PR DESCRIPTION
**Pull request by jackjin1997**
## Summary

Fixes #10675

When a request to the Gemini API includes both server-side built-in tools (e.g. `googleSearch`, `codeExecution`) and function declarations, the API requires `tool_config.include_server_side_tool_invocations: true` and otherwise returns:

```
[400 Bad Request] Please enable tool_config.include_server_side_tool_invocations
to use Built-in tools with Function calling.
```

Today `formatToolConfig` only produces a `toolConfig` block when `tool_choice` is set, so this flag never reaches the API even when mixed tools are bound, which makes `bindTools([myTool, { googleSearch: {} }])` unusable.

This PR detects the mixed-tools case in `formatData` and sets the flag automatically — users don't need to opt in. The detection runs after `formatTools`, so it sees the actual tool buckets (`functionDeclarations` vs everything else) regardless of how the user originally passed them.

Also widens `GeminiRequest.toolConfig.functionCallingConfig` to be optional since the new flag can now appear without a function-calling-mode override.

Tests added to `chat_models.test.ts` (mock client, asserts on captured request body):
- `mixing googleSearch with function tool sets includeServerSideToolInvocations`
- `function tool only does not set includeServerSideToolInvocations`

Existing GoogleSearch / GoogleSearchRetrieval tests still pass.

## AI Disclosure

This bug was identified and fixed with AI assistance.

### CODE REVIEW

File: libs/providers/langchain-google-common/src/utils/gemini.ts [hunk 1]
```diff
@@ -1998,8 +2021,13 @@ export function getGeminiAPI(config?: GeminiAPIConfig): GoogleAIAPI {
     if (tools && tools.length) {
       ret.tools = tools;
     }
-    if (toolConfig) {
-      ret.toolConfig = toolConfig;
+    const needsServerSideToolInvocations =
```
navgarcha: Would it make sense to move this into `formatToolConfig` directly and handle the logic there?

jackjin1997: Good call — moved it into `formatToolConfig` so all toolConfig construction lives in one place. Also moved the `mixesBuiltinAndFunctionTools` helper right above it for locality. Updated in c36b877.

tt-kkaiser:
@navgarcha Any updates?

navgarcha:
@tt-kkaiser im unable to review - we'll need someone from the langchain team to take a look 🙏

jackjin1997:
Following up on @navgarcha's note — would someone from the langchain team be able to take a look here? @christian-bromann / @hntrl, you've both shepherded most of the recent Google provider fixes, tagging in case this is on your radar.

Quick recap: auto-sets `tool_config.include_server_side_tool_invocations` in `formatData` when mixed built-in + function tools are bound, which the Gemini API otherwise rejects with a 400. Scoped to `google-common`, unit test added, CI is green. Happy to address any review comments.

christian-bromann:
Thanks for the contribution and sorry for the delay.

@afirstenberg would love to get your thoughts on this

akim-bow:
@afirstenberg @christian-bromann Hello guys, any news here?
Would be good to have this PR released

jackjin1997:
Thanks for the +1 @akim-bow — same pattern shows up whenever folks mix in any built-in tool, not just the one in the repro. CI's been green since April; happy to address anything if there's a specific concern. @afirstenberg @christian-bromann anything I can tweak to move this along?

AliSarraf:
Big +1 to merging this as soon as possible - I'd like us to have this PR released so we don't have to fork this repo just for the change. Thanks!

hntrl (approved):
thanks for the fix!

akim-bow:
Fix was landed in `@langchain/google-common`.
`@langchain/google` still doesn't support mixing built-in and function tools as initially reported in the issue #10675.

PR #10679, opened 2026-04-09
fix(core): make RemoveMessage type-compatible across MessageStructure…

hnustwjj:
## Summary

  - Remove unnecessary `TStructure` generic from `RemoveMessage` and `RemoveMessageFields`
  - Add `{ type: "remove"; id: string }` to the `BaseMessageLike` union type

  ## Problem

  When using `RemoveMessage` with `@langchain/langgraph-sdk`'s `stream.submit()`, TypeScript rejects it:

  Type 'RemoveMessage<MessageStructure>' is not assignable to type 'Message'.
    Types of property 'content' are incompatible.

  The serialized form `{ type: "remove", id: "..." }` is also rejected:

  Type '{ type: "remove"; id: string; }' is not assignable to type 'RemoveMessage'.
    Property 'content' is missing in type '{ type: "remove"; id: string; }'.

  Both forms work at runtime — `coerceMessageLikeToMessage` handles the serialized format (utils.ts:272), and `add_messages` reducer processes
  `RemoveMessage` correctly. The issue is purely in the type system.

  ## Root Cause

  1. `RemoveMessage` carries a `TStructure` generic that parameterizes the `content` type via `BaseMessage<TStructure, "remove">`. Since `RemoveMessage`
  always sets `content: []` in its constructor, this generic is unnecessary and causes type mismatches between different `MessageStructure` instantiations.

  2. `BaseMessageLike` doesn't include the serialized remove format `{ type: "remove"; id: string }`, even though `coerceMessageLikeToMessage` handles it
  at runtime.

  ## Fix

  - **`modifier.ts`**: Remove `TStructure` generic from `RemoveMessage` and `RemoveMessageFields`. Content is always `[]`, so the structure parameter only
  introduced spurious incompatibilities.
  - **`base.ts`**: Add `{ type: "remove"; id: string }` as an explicit union member to `BaseMessageLike`.

  ## Minimal Reproduction

  ```typescript
  import { RemoveMessage } from "@langchain/core/messages";

  // This should work but fails with current types:
  const messages: BaseMessageLike[] = [
    { type: "remove", id: "msg-123" },       // ❌ TS error
    new RemoveMessage({ id: "msg-456" }),     // ❌ TS error when passed to langgraph-sdk
  ];

  // Workaround before this fix: `as any`

  After this fix, both forms are accepted by TypeScript without as any.
  ```


hntrl (dismissed):
Thanks for the fix!

hnustwjj:
> Thanks for the fix!

I will fix the ci tomorrow morning（utc+8）

hnustwjj:
done.

hnustwjj:
merge？

christian-bromann (approved):

hnustwjj:
@christian-bromann oops, It shows that the merge has been blocked, but it doesn't seem to be caused by this PR.

PR #10692, opened 2026-04-13
fix(core): prevent merging reasoning and text blocks on index collision (#10691)

### LINKED ISSUE #10691: AIMessageChunk.concat() merges reasoning and text content blocks with the same index into a single malformed block, losing the text response
### Checked other resources

### Example Code

1. Use `@langchain/openai` with the Responses API and a reasoning model (e.g. `gpt-5.2` with `reasoning: { effort: "medium", summary: "auto" }`)
2. Stream a response that includes both a reasoning summary and a text output
3. Inspect the final accumulated `AIMessage.content` array after all chunks are concatenated

### Expected

```json
[
  { "type": "reasoning", "reasoning": "The user wants...", "index": 0 },
  { "type": "text", "text": "Here are the results...", "index": 0 }
]
```

### Actual

```json
[
  { "type": "reasoning", "reasoning": "The user wants...", "text": "Here are the results...", "index": 0 }
]
```

The text response is trapped inside the reasoning block. Any downstream code that checks `block.type === 'text'` to extract the response will miss it entirely.

When the malformed message is persisted to a LangGraph checkpoint and then replayed as conversation history to OpenAI's Responses API, the API rejects it:

```
Error: 400 litellm.BadRequestError: AzureException BadRequestError - {
  "error": {
    "message": "Item 'rs_0f7cf321675f593e0069dd2d635e4c8197a81c09dbb591f65a' of type 'reasoning' was provided without its required following item.",
    "type": "invalid_request_error",
    "param": "input",
    "code": null
  }
}
```

No stack trace — this is an API-level 400 rejection, not a runtime exception.

### Description

When streaming from OpenAI's Responses API with `reasoning.summary: "auto"` (or `"detailed"`), the final accumulated `AIMessage` can contain a single content block with `type: "reasoning"` that has **both** a `reasoning` field and a `text` field. The `text` field contains the model's actual response, which should have been a separate `{ type: "text" }` block.

This happens because `_findMergeTarget` in `@langchain/core` matches content blocks by `index` value **without checking `type`**. When a reasoning summary delta arrives with `index: 0` (from `event.summary_index`) and a text delta arrives with `index: 0` (from `event.content_index`), they get merged into a single block.

### 1. `@langchain/openai` assigns independent index namespaces to the same field

In `libs/langchain-openai/src/converters/responses.ts` (dist: `dist/converters/responses.js`), text deltas use `event.content_index` and reasoning deltas use `event.summary_index`:

**Text delta** (line ~451 of dist):
```javascript
if (event.type === "response.output_text.delta") content.push({
    type: "text",
    text: event.delta,
    index: event.content_index   // from output_text event
});
```

**Reasoning delta** (line ~572 of dist):
```javascript
if (event.delta) content.push({
    type: "reasoning",
    reasoning: event.delta,
    index: event.summary_index   // from reasoning_summary_text event
});
```

Both `content_index` and `summary_index` can be `0` — they come from different OpenAI output items (a `reasoning` item and a `message` item), each with their own internal indexing. But they're both assigned to the `index` field on the content block, making them look the same to the merge logic.

### 2. `_findMergeTarget` matches by `index` without checking `type`

In `libs/langchain-core/src/messages/base.ts` (dist: `dist/messages/base.js`), the merge target finder only checks `index` (and optionally `id`):

```javascript
function _findMergeTarget(merged, item) {
    const itemHasIndex = hasMergeableIndex(item);
    const itemHasId = hasMergeableId(item);
    if (!itemHasIndex && !itemHasId) return -1;
    return merged.findIndex((leftItem) => {
        const leftHasIndex = hasMergeableIndex(leftItem);
        const leftHasId = hasMergeableId(leftItem);
        if (itemHasIndex && leftHasIndex) {
            if (!(leftItem.index === item.index)) return false;
            // NO CHECK: leftItem.type === item.type
            if (leftHasId && itemHasId) return leftItem.id === item.id;
            return true;
        }
        // ...
    });
}
```

When a `{ type: "text", index: 0 }` block arrives and a `{ type: "reasoning", index: 0 }` block already exists in `merged`, `_findMergeTarget` returns the reasoning block's position because `index` matches.

### 3. `_mergeDicts` preserves the left block's `type` and merges string fields

```javascript
// In _mergeDicts:
else if (typeof merged[key] === "string")
    if (key === "type") continue;           // <-- keeps left type ("reasoning")
    // ...
    else merged[key] += value;              // <-- appends text: "" + "Here are the results..."
```

The `type` field is explicitly skipped (preserves `"reasoning"` from the left block). The `text` field from the incoming text block gets string-concatenated onto the reasoning block (which initially had no `text` field, so it's set via the `merged[key] == null` path).

Result: `{ type: "reasoning", reasoning: "...", text: "...", index: 0 }`.

## Impact

- The model's actual text response is lost to any code that reads content blocks by `type`
- Checkpoint persistence stores the malformed block, so page reloads lose the response
- This is a regression from PR #10681, which added `index` to reasoning blocks to fix fragmentation — but introduced index collisions with text blocks

## Related issues

- **#10684** — Reasoning blocks fragment during streaming (fixed by #10681, which introduced this regression)
- **#10664** — Consecutive text blocks not merged after non-text blocks
- **#10421** — `contentBlocks` drops reasoning blocks entirely (still open)
- **#9879** — Text blocks mapped to `input_text` instead of `output_text`

## Suggested fix

`_findMergeTarget` should require `type` to match when both blocks have a `type` field:

```typescript
if (itemHasIndex && leftHasIndex) {
    if (leftItem.index !== item.index) return false;
    // Don't merge blocks of different types even if index matches
    if ('type' in leftItem && 'type' in item && leftItem.type !== item.type) return false;
    if (leftHasId && itemHasId) return leftItem.id === item.id;
    return true;
}
```

Alternatively, `@langchain/openai` could namespace the indices (e.g. `reasoning:0` vs `text:0`) to prevent collisions without changing core merge logic.

## Workaround

In the checkpoint-to-UIMessage conversion layer, check for the malformed block shape and extract both fields:

```typescript
if (block.type === 'reasoning') {
    if (block.reasoning) parts.push({ type: 'reasoning', text: block.reasoning });
    if (block.text)      parts.push({ type: 'text', text: block.text });
}
```

### System Info

- **Node.js**: v22.x
- **OS**: macOS (darwin 25.3.0)
- **`@langchain/core`**: 1.1.39
- **`@langchain/openai`**: 1.4.4
- **`langchain`**: 1.x (monorepo `langchain` wrapper)
- **Model**: `gpt-5.2` via OpenAI Responses API with `reasoning: { effort: "medium", summary: "auto" }`

muhammadosama984:
## Summary

Fixes a bug where `AIMessageChunk.concat()` / `_mergeLists` merged **different** content block types when they shared the same `index`, producing a single malformed block (e.g. `type: "reasoning"` with both `reasoning` and `text`). This showed up when streaming the OpenAI Responses API with reasoning summaries: `summary_index` and `content_index` are separate namespaces but were both mapped to `index` on blocks ([issue description](https://github.com/langchain-ai/langchainjs/issues/10691)).

## Changes

- **`_findMergeTarget`** (`libs/langchain-core/src/messages/base.ts`): After index matches, require **`contentBlockTypesAllowMerge`**: both blocks must have compatible `type` values before merging.
- **Exception**: Preserve the existing Anthropic streaming behavior where `text_block` and `text_block_delta` intentionally share an index but represent one logical text stream.
- **Test**: Regression test for `reasoning` + `text` at `index: 0` staying as two blocks.

## Testing

- `pnpm --filter @langchain/core test src/messages/tests/base_message.test.ts`

Fixes #10691

### DIFF
--- a/libs/langchain-core/src/messages/base.ts
+++ b/libs/langchain-core/src/messages/base.ts
@@ -566,11 +566,45 @@ function hasMergeableId(value: unknown): value is { id: string | number } {
 }
 
 /**
+ * Whether two content blocks with the same merge key may be merged when their
+ * `type` fields differ. Most blocks must match on `type`; exceptions cover
+ * provider-specific streaming pairs (e.g. Anthropic text_block + text_block_delta).
+ */
+function contentBlockTypesAllowMerge(leftItem: unknown, item: unknown): boolean {
+  if (
+    typeof leftItem !== "object" ||
+    leftItem === null ||
+    typeof item !== "object" ||
+    item === null
+  ) {
+    return true;
+  }
+  if (!("type" in leftItem) || !("type" in item)) {
+    return true;
+  }
+  const leftType = (leftItem as { type: unknown }).type;
+  const rightType = (item as { type: unknown }).type;
+  if (leftType === rightType) {
+    return true;
+  }
+  if (
+    (leftType === "text_block" && rightType === "text_block_delta") ||
+    (leftType === "text_block_delta" && rightType === "text_block")
+  ) {
+    return true;
+  }
+  return false;
+}
+
+/**
  * Find the index of an existing item in `merged` that should be merged with
  * `item`, based on index and/or id matching.
  *
  * Matching priority:
- * 1. Both have index → match on index (+ id when both present)
+ * 1. Both have index → match on index and compatible content block type (+ id when both present).
+ *    Indices from different provider namespaces (e.g. summary_index vs content_index) can
+ *    collide; incompatible `type` values prevent merging (e.g. reasoning vs text). Streaming
+ *    pairs like text_block + text_block_delta still merge.
  * 2. Neither has index, both have id → match on id alone
  * 3. Otherwise → no match (item should be appended)
  */
@@ -591,6 +625,9 @@ function _findMergeTarget<Content extends ContentBlock>(
       // Both have index: match on index, with id as tiebreaker
       const indicesMatch = leftItem.index === item.index;
       if (!indicesMatch) return false;
+      if (!contentBlockTypesAllowMerge(leftItem, item)) {
+        return false;
+      }
       if (leftHasId && itemHasId) return leftItem.id === item.id;
       return true; // indices match, one or both missing id
     }
--- a/libs/langchain-core/src/messages/tests/base_message.test.ts
+++ b/libs/langchain-core/src/messages/tests/base_message.test.ts
@@ -178,6 +178,42 @@ test("_mergeLists merges blocks by numeric index", () => {
   ]);
 });
 
+test("AIMessageChunk.concat does not merge reasoning and text blocks that share the same index (issue #10691)", () => {
+  const chunk1 = new AIMessageChunk({
+    content: [
+      {
+        type: "reasoning",
+        reasoning: "The user wants",
+        index: 0,
+      },
+    ],
+  });
+
+  const chunk2 = new AIMessageChunk({
+    content: [
+      {
+        type: "text",
+        text: "Here are the results",
+        index: 0,
+      },
+    ],
+  });
+
+  const merged = chunk1.concat(chunk2);
+  expect(merged.content).toEqual([
+    {
+      type: "reasoning",
+      reasoning: "The user wants",
+      index: 0,
+    },
+    {
+      type: "text",
+      text: "Here are the results",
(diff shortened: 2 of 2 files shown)


hntrl:
Fixing in #10790

PR #10693, opened 2026-04-13
fix(agents): derive middleware hook state from invocation state

hntrl:
## Summary

This updates `createAgent` middleware state handling to derive middleware hook state from the current graph invocation state, rather than cross-node cached middleware state

### `langchain`

- Removed middleware node options that merged state from `StateManager` into hook input state.
- Updated `BeforeAgentNode`, `BeforeModelNode`, `AfterModelNode`, and `AfterAgentNode` to use invocation state directly.
- Updated `ReactAgent` wiring to stop constructing middleware state-getter tuples and pass middleware directly to `AgentNode`.
- Updated `AgentNode` `wrapModelCallHookMiddleware` typing/normalization to support both:
  - `AgentMiddleware[]`
  - legacy `[AgentMiddleware, getterFn]` entries used by existing tests.

### Tests and repro artifact

- Added middleware regressions for thread isolation:
  - `should not leak beforeAgent middleware state across thread_id`
  - `should not leak wrapModelCall middleware state across thread_id`


christian-bromann (dismissed):
LGTM 👍

PR #10694, opened 2026-04-13
refactor(core): decouple tracer defaults from langsmith callback naming

hntrl:
## Summary

Decouples LangSmith-specific defaults from base callback metadata flow while preserving tracing behavior and backward compatibility.

### `@langchain/core` callbacks

- Added provider-agnostic callback-manager options for tracer defaults:
  - `tracerInheritableMetadata`
  - `tracerInheritableTags`
- Applied tracer defaults only to `LangChainTracer` handlers (via `instanceof LangChainTracer`), leaving non-tracer handlers unchanged.

### `@langchain/core` runnable config

- Stopped copying all primitive `configurable` values into general run `metadata` during `ensureConfig`.
- Preserved only `configurable.model -> metadata.model` mirroring behavior.
- Added `_getTracingInheritableMetadataFromConfig` for deriving tracer-only inheritable metadata from `configurable` values.
- Wired callback-manager configuration to pass tracer-only inheritable metadata through provider-agnostic option names.

### `@langchain/core` LangChain tracer

- Extended `LangChainTracer` constructor fields to support tracing defaults (`metadata`, `tags`).
- Added `copyWithTracingConfig(...)` to clone tracer instances with merged default metadata/tags while preserving shared run state maps.
- Added patching of missing run metadata/tags at persistence boundaries so explicit run metadata wins and tracer defaults fill only missing keys.

### Tests

- Added and updated tests across:
  - `runnables/config` metadata derivation and precedence
  - callback manager tracer-only default application
  - LangChain tracer default metadata/tag patching and clone behavior

### DIFF
--- a/libs/langchain-core/src/callbacks/manager.ts
+++ b/libs/langchain-core/src/callbacks/manager.ts
@@ -33,6 +33,8 @@ type BaseCallbackManagerMethods = {
 export interface CallbackManagerOptions {
   verbose?: boolean;
   tracing?: boolean;
+  tracerInheritableMetadata?: Record<string, unknown>;
+  tracerInheritableTags?: string[];
 }
 
 export type Callbacks =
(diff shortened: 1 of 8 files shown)


libs/langchain-core/src/callbacks/manager.ts
```diff
@@ -1342,6 +1344,31 @@ export class CallbackManager
...
     }
+    const tracerInheritableMetadata = options?.tracerInheritableMetadata;
+    const tracerInheritableTags = options?.tracerInheritableTags;
+
+    if (
+      callbackManager &&
+      (tracerInheritableMetadata || tracerInheritableTags)
+    ) {
+      callbackManager.handlers = callbackManager.handlers.map((handler) =>
+        handler instanceof LangChainTracer
```
jacoblee93, line 1355: nit: typeguard in case of silly bundler things

libs/langchain-core/src/runnables/config.ts
```diff
@@ -6,13 +6,43 @@ export const DEFAULT_RECURSION_LIMIT = 25;
...
+export function _getTracingInheritableMetadataFromConfig(
+  config: RunnableConfig
+): Record<string, unknown> | undefined {
+  const configurable = config.configurable ?? {};
+  const metadata = config.metadata ?? {};
+  const langSmithMetadata: Record<string, unknown> = {};
+  for (const [key, value] of Object.entries(configurable)) {
+    if (
+      !key.startsWith("__") &&
+      !Object.prototype.hasOwnProperty.call(metadata, key) &&
```
jacoblee93, line 21: What's this?

jacoblee93 (dismissed):

libs/langchain-core/src/callbacks/manager.ts
hntrl: solved with hasInstance overload

libs/langchain-core/src/runnables/config.ts
hntrl: Basically a safer way to assert that a property lives in an object. have noticed some weird behavior in our projects with `in` operator or just calling .hasOwnProperty on the object itself

PR #10698, opened 2026-04-14
feat(core): Support tracing only metadata and populate it with model invocation params

jacoblee93:
Fixes # (issue)


hntrl (commented):
just as regression check could you add a test using streamEvents or something to validate this doesn't make it to stream paths?

jacoblee93:
Done thank you!

Issue #10655, opened 2026-04-02
feat(@langchain/google) Support Flex and Priority tiers


afirstenberg:
### Privileged issue

- [x] I am a LangChain maintainer, or was asked directly by a LangChain maintainer to create an issue here.

### Issue Content

https://blog.google/innovation-and-ai/technology/developers-tools/introducing-flex-and-priority-inference/

Possibly just the AI Studio API, but this allows cost savings or priority access for a fee.

Mostly configuration parameter changes (so likely in the discovery doc), but it does have a header in the reply with the service tier actually used.

Is there a standard way to pass this along so LangSmith can handle the billing estimate accordingly?

(@hntrl @christian-bromann)

afirstenberg:
See also #10482 and #10491 for how this needs to be done on Vertex

saalihou:
For anyone looking for a patch solution while this is getting implemented, just extend the ChatGoogle class like below and use this class instead:

```js
/**
 * Extends ChatGoogle to inject the `service_tier` parameter into the request body.
 * LangChain doesn't support this yet, but it's a top-level field in the Gemini API
 * alongside generationConfig/contents. Since _generate spreads invocationParams()
 * into the body, we can sneak it in here.
 * Values: 'standard' (default), 'priority' (low-latency, ~75-100% premium), 'flex' (50% discount, sheddable)
 */
class ChatGoogleWithServiceTier extends ChatGoogle {
  private serviceTier?: 'standard' | 'priority' | 'flex';

  constructor(fields: ConstructorParameters<typeof ChatGoogle>[0] & { serviceTier?: 'standard' | 'priority' | 'flex' }) {
    const { serviceTier, ...rest } = fields;
    super(rest);
    this.serviceTier = serviceTier;
  }

  invocationParams(options: Parameters<ChatGoogle['invocationParams']>[0]): ReturnType<ChatGoogle['invocationParams']> {
    const params = super.invocationParams(options);
    if (this.serviceTier && this.serviceTier !== 'standard') {
      return { ...params, serviceTier: this.serviceTier } as ReturnType<ChatGoogle['invocationParams']>;
    }
    return params;
  }
}
```

PR #10704, opened 2026-04-15
feat(google): Service tier and custom headers

afirstenberg:
Adds support for:
* "flex" and "priority" service tiers for both AI Studio API and Vertex API
* reporting what service tier was used via `response_metadata.serviceTier`
* customHeaders (chosen for backwards compatibility with the `@langchain/google-genai` library) which may be set against the model creation or per-request (based on work by @Nico385412 and @pawel-twardziak)

Tests:
* Integration tests for setting and retrieving the service tier
* Integration test for setting the service tier via a custom header
* Unit tests for setting the custom headers (based on work by @Nico385412 and @pawel-twardziak) 

Internal changes:
* A new way to retrieve and process the OpenAPI document, since there were manual edits that were necessary to get it to work.
*  Determining the headers is now done in `BaseChatGoogle.getHeaders()` so it is consistent between generate and stream.

TODO:
* Documentation

Fixes:
* #10655 

Supplants:
* #10482
* #10491


afirstenberg:
@Nico385412 and @pawel-twardziak - can you please review and confirm this should address the issues you were working on?

* I incorporated your unit test verbatim, only changing the name of the attribute
* I used the same approach you did for per-model and per-request handling, tho I consolidated the processing of headers in a new method

Are there any other specific headers you're aware of we should be testing with? (I know there are headers for Vertex that control Provisioned Throughput or not, but I don't have a project that has provisioning, so I can't test it.)

Nico385412:
LGTM @afirstenberg 

I don't know other headers for vertex, but maybe they will add more headers in their documentation in the future. 
It should work like a charm, thanks again for handling this use case

Nico385412:
Hey ! any news about on this pull request ?

christian-bromann (dismissed):
LGTM 👍

christian-bromann (approved):
👍 LGTM , thanks @afirstenberg

PR #10706, opened 2026-04-15
fix(langchain): set name on todoListMiddleware ToolMessages

### LINKED ISSUE #10705: `todoListMiddleware` produces ToolMessages without `name`, breaking Gemini (400 InvalidArgument)
### Description
`todoListMiddleware` emits `ToolMessage`s for `write_todos` without setting `name`. Google Gemini requires `name` on function responses and rejects the request with HTTP 400. Because the ToolMessage persists in history, every subsequent turn in the session keeps failing.

Anthropic and OpenAI accept the missing field, so the bug only surfaces on Gemini.

### Affected code
Two call sites in [libs/langchain/src/agents/middleware/todoListMiddleware.ts](libs/langchain/src/agents/middleware/todoListMiddleware.ts):

1. Successful `write_todos` result (≈ line 312).
2. Parallel-call error branch in `afterModel` (≈ line 382).

Every other ToolMessage producer in the package (`ToolNode`, `hitl`, `toolCallLimit`, `toolRetry`, `toolEmulator`, `contextEditing`, `pii`) already sets `name`.

### Reproduction
1. Build `createAgent` with `middleware: [todoListMiddleware()]` and a Gemini chat model.
2. Prompt the agent to trigger `write_todos`.
3. Send any follow-up — the Gemini call returns 400, and every later turn fails too.

### Proposed fix
Set `name: "write_todos"` at both ToolMessage constructions, with regression tests asserting `name` is set on both the success and parallel-rejection paths.

JadenKim-dev:
## Summary
- `todoListMiddleware` built `ToolMessage`s without a `name` field, which Gemini rejects with a 400. Once such a message enters history, every later turn in the session keeps failing.
- Set `name: "write_todos"` at both ToolMessage construction sites in todoListMiddleware.ts, aligning with every other ToolMessage producer in the package
- Added regression coverage on both the success and parallel-rejection paths.

Fixes #10705

## Test plan
- [x] `pnpm vitest run src/agents/middleware/tests/todoList.test.ts` (7 passed)


christian-bromann (approved):
LGTM 👍 

Thank you!

PR #10719, opened 2026-04-16
feat(tavily): add exactMatch parameter to TavilySearch

yashb1708:
## Summary

- Adds `exactMatch` parameter to `TavilySearch` tool, allowing users to restrict results to those containing the exact quoted phrase(s) in their query
- Adds `exact_match` to `TavilySearchParamsBase` type and passes it through `convertCamelToSnakeCase` to the API
- Instantiation-only parameter (not overridable by LLM at runtime, consistent with other params like `includeUsage`, `maxResults`, etc.)

## Example

```ts
import { TavilySearch } from "@langchain/tavily";

const tool = new TavilySearch({ exactMatch: true });
const result = await tool.invoke({ query: '"Dario Amodei" "Anthropic"' });
```

## Test plan

- [x] Verified `exactMatch: true` sends `exact_match: true` on the wire (snake_case, via `convertCamelToSnakeCase`)
- [x] Verified `exactMatch: undefined` (default) omits the field entirely
- [x] Verified `exactMatch: false` sends `exact_match: false` (not dropped)
- [x] Verified API returns different results with exact_match enabled (quoted phrases preserved in query echo)
- [x] `pnpm build` passes clean
- [x] Changeset included (`minor` bump for `@langchain/tavily`)


libs/providers/langchain-tavily/src/utils.ts
```diff
@@ -758,6 +764,7 @@ export class TavilySearchAPIWrapper extends BaseTavilyAPIWrapper {
     const headers = {
       Authorization: `Bearer ${this.tavilyApiKey}`,
       "Content-Type": "application/json",
+      "X-Client-Source": TAVILY_CLIENT_SOURCE,
```
christian-bromann, line 767: This change seems to be unrelated to the exact_match feature, let's strip it out.

libs/providers/langchain-tavily/src/utils.ts
```diff
@@ -82,6 +83,11 @@ export type TavilySearchParamsBase = {
    * @default false
    */
   include_usage?: boolean;
+
+  /**
+   * Only return results containing the exact phrase(s) in quotes in your query.
+   */
+  exact_match?: boolean;
```
christian-bromann, line 90: Let's add a unit test that we properly pass this through

libs/providers/langchain-tavily/src/utils.ts
```diff
@@ -82,6 +83,11 @@ export type TavilySearchParamsBase = {
    * @default false
    */
   include_usage?: boolean;
+
+  /**
+   * Only return results containing the exact phrase(s) in quotes in your query.
```
christian-bromann, line 88: we should document the default param

libs/providers/langchain-tavily/src/tavily-search.ts
```diff
@@ -166,6 +166,11 @@ export type TavilySearchAPIRetrieverFields = ToolParams & {
    * @default false
    */
   autoParameters?: boolean;
+
+  /**
+   * Only return results containing the exact phrase(s) in quotes in your query.
```
christian-bromann, line 171: let's document the default param

christian-bromann (changes requested):

PR #10733, opened 2026-04-20
fix(core): Update inheritance behavior for tracer metadata for special keys

jacoblee93:


libs/langchain-core/src/tracers/tracer_langchain.ts
```diff
@@ -48,6 +48,21 @@ export interface LangChainTracerFields extends BaseCallbackHandlerInput {
...
+ * the run already has a value for them. This lets nested contexts
+ * (e.g. a subagent invoked from inside a parent agent) override a
+ * LangSmith-only tracing metadata value that was set by an ancestor.
+ *
+ * Keep this list very small: every key here loses the default
+ * "first wins" protection and is always clobbered by the nearest
+ * enclosing tracer config. Only keys that are strictly for LangSmith
+ * tracing bookkeeping should be added.
+ */
+export const LANGSMITH_INHERITABLE_METADATA_KEYS = new Set<string>([
```
hntrl: nit:

```suggestion
export const OVERRIDABLE_LANGSMITH_INHERITABLE_METADATA_KEYS = new Set<string>([
```

or something like that

hntrl (dismissed):

hntrl (approved):

PR #10735, opened 2026-04-21
feat(anthropic): add automatic prompt caching via top-level cache_control

chenzimin:
Brings `@langchain/anthropic` to parity with the Python package's [automatic prompt caching](https://docs.langchain.com/oss/python/integrations/chat/anthropic#automatic-caching) support. Passing `cache_control` as an invocation option now forwards it directly to the Anthropic API as a top-level `cache_control` parameter — the API automatically applies the cache breakpoint to the last cacheable block in the request and advances it as the conversation grows. This removes the need to place `cache_control` markers on individual content blocks manually and is the recommended pattern for multi-turn conversations.

Mirrors the Python equivalent in langchain-ai/langchain#35967.

Docs: https://platform.claude.com/docs/en/build-with-claude/prompt-caching#automatic-caching

```ts
const model = new ChatAnthropic({ model: "claude-sonnet-4-5-20250929" });

await model.invoke(messages, {
  cache_control: { type: "ephemeral" }, // or { type: "ephemeral", ttl: "1h" }
});
```


libs/providers/langchain-anthropic/src/utils/message_inputs.ts
```diff
@@ -551,80 +551,6 @@ export function _convertMessagesToAnthropicPayload(
...
-    messages[lastMessageIndex] = {
-      ...lastMessage,
-      content,
-    };
-    return { ...payload, messages };
-  }
-
-  return payload;
-}
-
```
hntrl: I don't have a ton of confidence that automatic prompt caching works in the same fashion as appending cache control params to the last block, so we need to treat removing this as a breaking change.

PR #10697, opened 2026-04-14
feat(langchain): adopt new streaming primitives

christian-bromann:
This patch adds support for tool and middleware level streaming for `ReactAgent` using the new streaming primitives from Langgraph.

PR #10772, opened 2026-04-24
feat(core): new event streaming

christian-bromann:
## Summary

This PR stacks on [#10678](https://github.com/langchain-ai/langchainjs/pull/10678) and consolidates the new chat model streaming protocol with the agent streaming work from [#10697](https://github.com/langchain-ai/langchainjs/pull/10697).

The first commit cherry-picks #10697 so the `ReactAgent` streaming changes can be developed and tested against the new chat model streaming primitives. The remaining commits align the stream event protocol, update the core stream aggregation behavior, refresh the Anthropic adapter, and add/adjust test coverage.

### Included From #10697

Cherry-picked the agent streaming work from #10697, including support for the new LangGraph streaming primitives in `ReactAgent`, native `toolCalls` and `middleware` stream projections, stream transformer typing, package exports, and related tests.

This PR does not restate all of #10697 in detail; it carries that work here so the chat model stream protocol changes can be validated end-to-end with agent streaming.

### Chat Model Stream Protocol Consolidation

- Standardized stream events on the `event` discriminator instead of `type`.
- Renamed message finish response metadata from `responseMetadata` to `metadata`.
- Simplified `content-block-delta` events so they carry incremental `content: ContentBlock` updates directly, instead of a separate delta union.
- Widened stream usage inputs to allow partial usage snapshots from providers, while normalizing usage before exposing final `AIMessage` output.
- Made `message-finish.reason` optional for providers that do not always emit a stop reason.

### Core Stream Behavior

- Updated `ChatModelStream` and its typed substreams to consume the consolidated event shape.
- Preserved append semantics for text, reasoning, and tool call chunk args.
- Kept stable tool call `id` / `name` values when later chunks omit them.
- Improved reasoning stream handling for reasoning/thinking blocks and finish boundaries.
- Changed the usage substream to resolve to `undefined` when no usage was reported, instead of manufacturing an empty zero-token snapshot.

### Compatibility Bridge And Anthropic Adapter

- Updated the legacy chunk-to-event bridge to emit the consolidated event format.
- Updated Anthropic stream conversion to emit content block updates directly.
- Preserved provider passthrough events for Anthropic-specific stream data.
- Kept tool call finalization behavior, including JSON argument parsing and invalid tool call handling.

### Tests And Environment Coverage

- Updated core stream protocol tests, stream assembly tests, and compatibility bridge tests.
- Updated Anthropic stream event tests for the consolidated protocol.
- Added CJS/export environment coverage needed for the new agent streaming exports.
- Added a changeset covering the protocol consolidation.


libs/langchain-core/src/language_models/event.ts
```diff
@@ -117,13 +78,13 @@ export type FinishReason = "stop" | "length" | "tool_use" | "content_filter";
...
  */
 export interface MessageFinishEvent {
-  type: "message-finish";
+  event: "message-finish";
   /** Why the model stopped generating. */
-  reason: FinishReason;
+  reason?: FinishReason;
   /** Final usage snapshot. */
-  usage?: UsageMetadata;
+  usage?: UsageMetadataLike;
```
hntrl: uber uber nit: can we just use `Partial<UsageMetadata>` instead of having the indirection

libs/langchain-core/src/language_models/event.ts
```diff
@@ -117,13 +78,13 @@ export type FinishReason = "stop" | "length" | "tool_use" | "content_filter";
...
   /** Provider-specific response metadata (model name, response ID, headers, etc.). */
-  responseMetadata?: Record<string, unknown>;
+  metadata?: Record<string, unknown>;
```
hntrl: am assuming we're gonna funnel this into responseMetadata when reconstructing messages?

libs/langchain/src/agents/ReactAgent.ts
```diff
@@ -1301,6 +1317,100 @@ export class ReactAgent<
...
+    state: InvokeStateParameter<Types>,
+    config?: InvokeConfiguration<
+      InferContextInput<
+        Types["Context"] extends AnyAnnotationRoot | InteropZodObject
+          ? Types["Context"]
+          : AnyAnnotationRoot
+      > &
+        InferMiddlewareContextInputs<Types["Middleware"]>
+    > & {
+      transformers?: ReadonlyArray<() => StreamTransformer<any>>;
```
hntrl: if we figured out how to type behavior like this at some point, 🔥🔥

hntrl (commented):
For right now I can speak to the public api surface -- I'm fine largely trusting the design decisions behind transformers (though I'm excited to see the readout as someone largely observing from arms length)
* Middleware isn't a protocol event right, it's just a projection of node events that we get from inside of a graph?
* on chat model events
  * I think we need to keep the `delta` semantics. Right now we extrapolate by just using PartialContentBlock in the delta type, but I think we'd benefit by making these explicit if I'm just thinking about it from a "completions first" perspective
    * Like its not abundantly clear to me that we'd do special handling for named `text` + `reasoning` fields (assuming that these come across as token chunks) vs. this third delta type
    * A lot of this is predicated on us having the same idea of how delta merging works which I know was very in flux 
  * Hopefully we can do away with `tool_call_chunk` and `server_tool_call_chunk` if we stick to the three delta types and just have tool call chunks be handled by a merge function that just goes one layer deep when we get that third delta type?

christian-bromann:
> Middleware isn't a protocol event right, it's just a projection of node events that we get from inside of a graph?

Correct, both `toolCall` and `middleware` stream transformer are just in-process stream projections.

christian-bromann:
> * I think we need to keep the `delta` semantics.

Kept delta semantics, see https://github.com/langchain-ai/agent-protocol/pull/78

hntrl (dismissed):
can we split up the langchain-core/langchain changes so we don't have to rely on a dev release?

hntrl (approved):

PR #10773, opened 2026-04-25
fix(core): correct model mapping for gpt-4o in getModelNameForTiktoken

abhilov23:
- gpt-4o is incorrectly matched by gpt-4 prefix condition
- This causes wrong tokenizer selection

### Problem
`gpt-4o` models are incorrectly matched by the `gpt-4-` prefix condition in `getModelNameForTiktoken`.  
This results in them being mapped to `"gpt-4"` instead of `"gpt-4o"`, leading to incorrect tokenizer selection.

### Solution
Reordered the conditions to check for `gpt-4o` before the more general `gpt-4-` prefix.

### Before
- `gpt-4o` → mapped to `"gpt-4"`

### After
- `gpt-4o` → correctly mapped to `"gpt-4o"`

### Notes
This change is backward-compatible and does not affect other model mappings.

- This is a small internal fix and does not require a version bump.


christian-bromann:
Thanks for the PR!

Reordering the `gpt-4o` and `gpt-4-` checks doesn’t change behavior, since `gpt-4o` never matches `startsWith("gpt-4-")` (the next character after `gpt-4` is `o`, not `-`).

PR #10776, opened 2026-04-27
fix(deps): remediate uuid@10 vulnerability

hntrl:
## Summary

Remediates the published `uuid@10` vulnerability by removing direct `uuid` usage from LangChain packages and routing internal UUID generation through `@langchain/core`.

## Changes

- Added vendored UUID helpers under `@langchain/core/src/utils/uuid` and included them in core build exports.
- Updated internal UUID call sites in core, langchain, and provider packages to import from the core UUID utility.
- Removed direct `uuid` dependencies/devDependencies from affected published packages (and bumped `examples` to `uuid@^14`).
- Added a patch changeset for all impacted published packages.


jwasnoggin:
Could the same thing be done with `@langchain/community` as well? It still has the `uuid` vulnerability.

PR #10777, opened 2026-04-27
feat(anthropic): support strict tool calling for custom tools

### LINKED ISSUE #10731: feat(anthropic): support strict tool calling for custom tools (parity with @langchain/openai)
### Summary

`@langchain/openai` exposes strict tool validation via `bindTools(tools, { strict: true })` and the instance-level `supportsStrictToolCalling` field. `@langchain/anthropic` only exposes `strict` for the built-in `webSearch_20250305` tool and has no first-class way to set it on custom LangChain tools, even though the Anthropic API supports `strict: true` on every tool definition since the introduction of [grammar-constrained sampling](https://platform.claude.com/docs/en/agents-and-tools/tool-use/strict-tool-use).

This issue proposes symmetric `strict` support on the Anthropic provider so users can opt into grammar-constrained sampling for their custom tools without the `extras.providerToolDefinition` escape hatch.

### What works today

- **OpenAI**: `model.bindTools([weatherTool], { strict: true })` propagates `strict: true` to the wire ([code](https://github.com/langchain-ai/langchainjs/blob/main/libs/providers/langchain-openai/src/chat_models/base.ts#L662-L693)). Also `new ChatOpenAI({ supportsStrictToolCalling: true })` enables it instance-wide.
- **Anthropic webSearch tool**: `tools.webSearch_20250305({ strict: true })` is honored ([code](https://github.com/langchain-ai/langchainjs/blob/main/libs/providers/langchain-anthropic/src/tools/webSearch.ts#L40-L49)).

### What does not work today

- **Anthropic custom tools**: there is no `strict` parameter on `bindTools`, no instance field, and `AnthropicToolExtrasSchema` does not include `strict`, so passing `extras: { strict: true }` to a `tool(...)` definition is silently dropped.
- The only way to ship `strict: true` for a custom Anthropic tool today is to bypass the LangChain mapping entirely via `extras.providerToolDefinition`, which forces users to construct the raw Anthropic tool object themselves.

### Proposed change

Bring Anthropic to parity with the OpenAI provider:

1. Add `strict?: boolean` to `ChatAnthropicCallOptions` so callers can use `bindTools(tools, { strict: true })`.
2. Add a `supportsStrictToolCalling?: boolean` (or similarly named) field on `ChatAnthropic` so callers can enable it at instance construction.
3. Update `formatStructuredToolToAnthropic` to set `strict: true` on each LangChain tool when the call-options or instance flag is set.
4. Add `strict: z.boolean().optional()` to `AnthropicToolExtrasSchema` so per-tool overrides via `extras: { strict: true }` work too (mirrors the per-tool override style available for `cache_control`, `defer_loading`, etc.).

### Use case

Users building agentic workflows want grammar-constrained tool inputs across providers without provider-specific shims. With this change, the same code works on both:

\`\`\`ts
const llm = createOpenAIOrAnthropic(...)
const llmWithTools = llm.bindTools([myTool], { strict: true })
\`\`\`

### Test cases

- \`bindTools([tool], { strict: true })\` → resulting Anthropic tool def has \`strict: true\`.
- \`bindTools([tool], { strict: false })\` → strict not sent (matches Anthropic API default).
- \`bindTools([tool])\` (no flag) → strict not sent.
- \`new ChatAnthropic({ supportsStrictToolCalling: true })\` + \`bindTools([tool])\` → strict: true.
- Per-call kwarg overrides instance default.
- \`tool(handler, { name, description, schema, extras: { strict: true } })\` → that single tool ships strict.
- \`AnthropicToolExtrasSchema\` parses \`{ strict: true }\` without dropping the field.

### Related

- Anthropic strict tool docs: https://platform.claude.com/docs/en/agents-and-tools/tool-use/strict-tool-use
- pydantic-ai is moving the same direction: https://github.com/pydantic/pydantic-ai/issues/3541
- OpenAI strict reference implementation: \`libs/providers/langchain-openai/src/chat_models/base.ts:662\`

Will follow up with a PR.

jonaslalin:
## Summary

Adds `strict` tool calling support to `@langchain/anthropic`. Mirrors the user-facing shape of `@langchain/openai` where it makes sense, but scoped to call options + per-tool defaults (no constructor capability flag). Setting `strict: true` opts a tool definition into Anthropic's [grammar-constrained sampling](https://platform.claude.com/docs/en/agents-and-tools/tool-use/strict-tool-use), guaranteeing tool inputs match the provided JSON Schema.

### Public API additions

- `ChatAnthropicCallOptions.strict?: boolean`. Pass via:
  - `model.bindTools(tools, { strict: true })`
  - `model.withConfig({ strict: true })` (or chained: `model.bindTools(tools).withConfig({ strict: true })`)
  - `model.invoke(input, { strict: true })`
  - `model.withStructuredOutput(schema, { method: "functionCalling", strict: true })`
- `AnthropicToolExtrasSchema.strict?: boolean`. Per-tool default for LangChain tools, alongside `cache_control` / `defer_loading` / `input_examples` / `allowed_callers`.

### Precedence

1. Per-call `strict` (any of the call options above).
2. Per-tool `strict`: `extras.strict` (LangChain shape), `function.strict` (OpenAI shape), or native `strict` (Anthropic shape).

If neither is set, `strict` is omitted from the request and Anthropic's default (non-strict) behavior applies.

### Safety guard for `withStructuredOutput`

Anthropic's `strict` is a property of a tool definition, so it only applies under `method: "functionCalling"`. The `"jsonSchema"` path uses Anthropic's [native structured outputs](https://docs.claude.com/en/docs/build-with-claude/structured-outputs) and does not accept `strict`. `withStructuredOutput` throws when `strict` is combined with `"jsonSchema"` or `"jsonMode"` instead of silently dropping it.

### Documentation

Added a "Strict Tool Use" section to `libs/providers/langchain-anthropic/README.md` covering per-call, bound-model default, per-tool, and `withStructuredOutput` usage, plus precedence and the `jsonSchema`/`jsonMode` guard.

Fixes #10731.

### Unit tests (`chat_models.test.ts`, mocked fetch)

The mocked fetch returns a valid Anthropic `Messages` response with a `tool_use` block, so `.invoke()` succeeds and tests assert directly on the captured request body.

**Single-source coverage**
- [x] `applies strict from .bindTools call args`
- [x] `applies strict from .withConfig call options`
- [x] `applies strict from .bindTools(...).withConfig({ strict })`
- [x] `applies strict from per-call invoke options`
- [x] `applies strict from .withStructuredOutput config`
- [x] `per-tool extras.strict applies to that tool only` (LangChain shape)
- [x] `per-tool function.strict applies on OpenAI-shaped tools`
- [x] `per-tool native strict applies on Anthropic-shaped tools`

**Baseline / negative**
- [x] `omits strict when not provided anywhere`
- [x] `omits strict when not passed in .withStructuredOutput`

**Precedence (per-call beats per-tool, across every tool shape)**
- [x] `per-call invoke strict overrides .bindTools strict`
- [x] `per-call strict overrides per-tool extras.strict` (LangChain shape)
- [x] `per-call strict overrides OpenAI-shaped tool's function.strict`
- [x] `per-call strict overrides Anthropic-shaped tool's own strict`

**Error cases**
- [x] `withStructuredOutput throws when strict is set with method = jsonSchema` (parameterized)
- [x] `withStructuredOutput throws when strict is set with method = jsonMode` (parameterized)

### Integration tests (`chat_models-strict.int.test.ts`, live API against `claude-haiku-4-5-20251001`)

- [x] `strict via .bindTools(tools, { strict: true })`
- [x] `strict via .bindTools(...).withConfig({ strict: true })`
- [x] `strict via .withConfig({ tools, strict: true })`
- [x] `strict via .invoke(input, { strict: true })`
- [x] `strict via tool() extras.strict`
- [x] `per-call strict=false overrides tool() extras.strict=true`
- [x] `strict via .withStructuredOutput(schema, { strict: true })` (round-trips through Zod)
- [x] `strict tool call composes with strict structured output` (exercises both strict surfaces end-to-end)

### Quality gates

- [x] 184/184 unit tests in \`@langchain/anthropic\` pass, no type errors
- [x] 8/8 integration tests pass against \`claude-haiku-4-5-20251001\`
- [x] \`pnpm lint\` clean
- [x] \`pnpm format:check\` clean

Made with [Cursor](https://cursor.com)

### DIFF
--- a/libs/providers/langchain-anthropic/src/chat_models.ts
+++ b/libs/providers/langchain-anthropic/src/chat_models.ts
@@ -183,6 +183,19 @@ export interface ChatAnthropicCallOptions
    * @see https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching
    */
   cache_control?: AnthropicCacheControl;
+
+  /**
+   * If `true`, model output is guaranteed to exactly match the JSON Schema
+   * provided in the tool definition. If `true`, the input schema will also be
+   * validated according to
+   * https://platform.claude.com/docs/en/agents-and-tools/tool-use/strict-tool-use.
+   *
+   * If `false`, input schema will not be validated and model output will not
+   * be validated.
+   *
+   * If `undefined`, `strict` argument will not be passed to the model.
+   */
+  strict?: boolean;
 }
 
 function _toolsInParams(
@@ -1096,10 +1109,12 @@ export class ChatAnthropicMessages<
    * Formats LangChain StructuredTools to AnthropicTools.
    *
    * @param {ChatAnthropicCallOptions["tools"]} tools The tools to format
+   * @param fields Optional `strict` flag applied to every formatted custom tool.
    * @returns {AnthropicTool[] | undefined} The formatted tools, or undefined if none are passed.
    */
   formatStructuredToolToAnthropic(
-    tools: ChatAnthropicCallOptions["tools"]
+    tools: ChatAnthropicCallOptions["tools"],
+    fields?: { strict?: boolean }
   ): Anthropic.Messages.ToolUnion[] | undefined {
     if (!tools) {
       return undefined;
@@ -1113,24 +1128,44 @@ export class ChatAnthropicMessages<
         return tool;
       }
       if (isAnthropicTool(tool)) {
+        if (fields?.strict !== undefined) {
+          return {
+            ...tool,
+            strict: fields.strict,
+          };
+        }
         return tool;
       }
       if (isOpenAITool(tool)) {
+        // LangChain's OpenAI tool type doesn't surface the optional `strict`
+        // flag that the OpenAI API itself supports on function definitions,
+        // so we read it through a runtime check rather than a type assertion.
+        const functionStrict =
+          "strict" in tool.function && typeof tool.function.strict === "boolean"
+            ? tool.function.strict
+            : undefined;
+        const strict = fields?.strict ?? functionStrict;
         return {
           name: tool.function.name,
           description: tool.function.description,
           input_schema: tool.function
             .parameters as Anthropic.Messages.Tool.InputSchema,
+          ...(strict !== undefined ? { strict } : {}),
         };
       }
       if (isLangChainTool(tool)) {
+        const { strict: extrasStrict, ...restExtras } = tool.extras
+          ? AnthropicToolExtrasSchema.parse(tool.extras)
+          : {};
+        const strict = fields?.strict ?? extrasStrict;
         return {
           name: tool.name,
           description: tool.description,
           input_schema: (isInteropZodSchema(tool.schema)
             ? toJsonSchema(tool.schema)
             : tool.schema) as Anthropic.Messages.Tool.InputSchema,
-          ...(tool.extras ? AnthropicToolExtrasSchema.parse(tool.extras) : {}),
+          ...restExtras,
+          ...(strict !== undefined ? { strict } : {}),
         };
       }
       throw new Error(
@@ -1148,7 +1183,9 @@ export class ChatAnthropicMessages<
     kwargs?: Partial<CallOptions>
   ): Runnable<BaseLanguageModelInput, AIMessageChunk, CallOptions> {
     return this.withConfig({
-      tools: this.formatStructuredToolToAnthropic(tools),
+      tools: this.formatStructuredToolToAnthropic(tools, {
+        strict: kwargs?.strict,
+      }),
       ...kwargs,
     } as Partial<CallOptions>);
   }
@@ -1205,7 +1242,9 @@ export class ChatAnthropicMessages<
       stop_sequences: options?.stop ?? this.stopSequences,
       stream: this.streaming,
       max_tokens: this.maxTokens,
-      tools: this.formatStructuredToolToAnthropic(options?.tools),
+      tools: this.formatStructuredToolToAnthropic(options?.tools, {
+        strict: options?.strict,
+      }),
       tool_choice,
       thinking: this.thinking,
       context_management: this.contextManagement,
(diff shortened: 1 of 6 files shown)


hntrl (commented):
Is there a better heuristic we can use instead of a constructor field? E.g. telling the user to add `supportsStrictToolCalling` feels a little awkward. If strict tool calling is only available for function calling, we should do a validation check inside of `withStructuredOutput` instead of at the top level

jonaslalin:
> Is there a better heuristic we can use instead of a constructor field? E.g. telling the user to add `supportsStrictToolCalling` feels a little awkward. If strict tool calling is only available for function calling, we should do a validation check inside of `withStructuredOutput` instead of at the top level

@hntrl  Thanks for the look. Did a careful side-by-side with @langchain/openai:

The constructor field mirrors ChatOpenAI.supportsStrictToolCalling (same name, JSDoc, precedence rule, call sites). I'm propagating an established pattern. A smarter auto-detection heuristic would benefit from landing in both providers together.
The validation check is already inside withStructuredOutput and mirrors @langchain/openai's base.ts:1045-1049.
Comparing line-by-line surfaced a real bug: withStructuredOutput's baked-in strict was being overwritten by invocationParams' instance fallback. Refactored to plumb strict via withConfig like OpenAI; added a regression test.
Also dropped a small _resolveStrict helper in favor of inline ??, removed a few redundant comments. Net: structurally and behaviorally identical to @langchain/openai

libs/providers/langchain-anthropic/src/chat_models.ts
```diff
@@ -401,6 +414,12 @@ export interface AnthropicInput {
    * See https://docs.claude.com/en/api/beta-headers for available beta features.
    */
   betas?: AnthropicBeta[];
+
+  /**
+   * Whether the model supports the `strict` argument when passing in tools.
+   * If `undefined` the `strict` argument will not be passed to Anthropic.
+   */
+  supportsStrictToolCalling?: boolean;
```
hntrl: This is a parameter that OpenAI has, but I'm not convinced that its great. We don't strictly need to match parity here.

Would the call option be sufficient enough for your case? At any rate I'd think we should rename this field since its working more like a default then a capability flag

hntrl (commented):
Thanks for giving it another once over. Could we also add some integration tests for this?

libs/providers/langchain-anthropic/src/chat_models.ts
jonaslalin: removed

jonaslalin:
> Thanks for giving it another once over. Could we also add some integration tests for this?

@hntrl integration tests added, constructor arg dropped. please review again

hntrl (approved):
Thanks for the fix!

PR #10783, opened 2026-04-29
fix(aws): ignore empty text blocks in converse conversion

### LINKED ISSUE #10782: @langchain/aws crashes on empty AIMessage text blocks in Bedrock Converse conversion
## Bug

`@langchain/aws` crashes in Bedrock Converse message conversion when an `AIMessage` contains a strictly empty text content block.

The crash comes from `convertAIMessageToConverseMessage` in:

`libs/providers/langchain-aws/src/utils/message_inputs.ts`

Current branch:

```ts
if (block.type === "text" && block.text !== "") {
  // ...
} else {
  throw new Error(
    `Unsupported content block type: ${block.type} with content of ...`
  );
}
```

Because of `&& block.text !== ""`, a block like `{ type: "text", text: "" }` falls into the final `else` and throws.

## Reproduction

Minimal repro shape:

```ts
import { convertToConverseMessages } from "../utils/message_inputs.js";

convertToConverseMessages([
  new AIMessage({
    content: [{ type: "text", text: "" }],
  }),
]);
```

Observed error:

```txt
Unsupported content block type: text with content of { "text": "" }
```

## Why this matters

This content shape can happen in tool-use flows where a model returns a tool call without accompanying prose. In those cases, an empty text block should be treated as a no-op rather than as an unsupported block type.

## Expected behavior

`convertToConverseMessages` should not throw for `{ type: "text", text: "" }`. Empty text blocks should be ignored (consistent with current handling intent for blank/whitespace text behavior).

## Proposed fix

Change:

```ts
if (block.type === "text" && block.text !== "") {
```

to:

```ts
if (block.type === "text") {
```

This keeps existing behavior for non-empty and whitespace text while making strict-empty text a no-op instead of an exception.

tinque:
## Summary

- fix `convertAIMessageToConverseMessage` to treat strictly empty text blocks (`{ type: "text", text: "" }`) as text blocks instead of falling into the unsupported block branch
- add a regression case in `chat_models.test.ts` covering an `AIMessage` with an empty text block and tool call content
- keep existing behavior for non-empty and whitespace text blocks unchanged

## Root cause

`convertAIMessageToConverseMessage` currently checks `block.type === "text" && block.text !== ""`.

That excludes strictly empty text blocks from the text branch, so they fall through to the final `else` and throw:

`Unsupported content block type: text with content of { "text": "" }`

## Verification

- `pnpm exec vitest run src/tests/chat_models.test.ts -t "ai message with strictly empty text block"` (fails before fix, passes after)
- `pnpm test`
- `pnpm lint`
- `pnpm format:check`

Fixes #10782


hntrl (dismissed):

hntrl (dismissed):

tinque:
Added a follow-up commit for the edge case raised on #10782.

Commit: e46fad98a (`fix(aws): skip empty assistant converse messages`)

What changed:
- Added a regression case for an `AIMessage` containing only `{ type: "text", text: "" }`.
- Filtered empty assistant Converse messages after conversion so the empty text block is a true no-op instead of producing `{ role: "assistant", content: [] }`.

Local verification:
- `pnpm --filter @langchain/aws test src/tests/chat_models.test.ts -t "ai message with only strictly empty text block"`
- `pnpm --filter @langchain/aws test src/tests/chat_models.test.ts`
- `pnpm --filter @langchain/aws build`

callmeGilles:
Hey @hntrl, thanks for your help. Any chance for you or your team to have another quick look please? We would really appreciate it.

tinque:
@hntrl any chance for you or your team to have another quick look please? We would really appreciate it.

PR #10785, opened 2026-04-30
feat(weaviate): create from json

dudanogueira:
## Summary

This PR upgrades the `@langchain/weaviate` integration to `weaviate-client` v3.13.0 and adds support for creating collections from a raw JSON schema using the new `createFromJson` API.

It will allow more advanced usage, as you can now define the advanced collections options, such as dynamic index, quantization/compression, multi tenant, different property tokenization, etc.

It will also allow integrations such as n8n to pass a json schema along.

[You can use this online tool to design your collection schema](https://weaviate.github.io/weaviate-add-collection/)

### `@langchain/weaviate`

- **New `jsonSchema` field** — `WeaviateLibArgs` now accepts a `jsonSchema?: Record<string, unknown>` property that maps directly to the raw Weaviate JSON schema format (as returned by the REST API or `client.collections.exportToJson()`). Internally it calls `client.collections.createFromJson()`, introduced in weaviate-client v3.13.0.
- **Backward compatible** — the existing `schema?: CollectionConfigCreate` field is fully preserved and continues to call `client.collections.create()`. When both fields are provided, `jsonSchema` takes precedence.
- **`crypto.randomUUID()`** — replaced the broken `@langchain/core/utils/uuid` import with the native `crypto.randomUUID()` (safe since Node ≥ 20 is already required by this package).
- **weaviate-client bumped to `^3.13.0`** — updated `package.json` and `pnpm-lock.yaml` (only the weaviate-client dependency and its transitive deps were changed in the lockfile).
- **Unit tests** — added four new tests covering all `initialize()` creation paths: `jsonSchema` only, `schema` only, `jsonSchema` takes precedence over `schema`, and skip creation when the collection already exists.
- Accept both `filter` and `filters` parameter for Hybrid Searches.

### `@langchain/core`

- Exports `./utils/uuid` as a public subpath in `package.json` and registers it in the import map, making UUID utilities available as `@langchain/core/utils/uuid`.

## Usage

```typescript
// New: create from a raw JSON schema (e.g. exported from Weaviate or the REST API)
const jsonSchema = {
  "class": "MyJsonCollection",
  "properties": [
    {
      "name": "text",
      "dataType": [
        "text"
      ],
      "indexFilterable": true,
      "indexSearchable": true,
      "tokenization": "word",
      "indexRangeFilters": false
    },
    {
      "name": "my_custom_id",
      "dataType": [
        "text"
      ],
      "indexFilterable": true,
      "indexSearchable": true,
      "tokenization": "field",
      "indexRangeFilters": false
    }
  ],
  "vectorConfig": {
    "default": {
      "vectorizer": {
        "text2vec-openai": {
          "properties": [
            "text"
          ]
        }
      },
      "vectorIndexType": "dynamic",
      "vectorIndexConfig": {
        "hnsw": {
          "rq": {
            "enabled": true
          }
        }
      }
    }
  },
  "multiTenancyConfig": {
    "enabled": true,
    "autoTenantCreation": true,
    "autoTenantActivation": true
  }
};

const store = await WeaviateStore.initialize(new OpenAIEmbeddings(), {
  client,
  jsonSchema,
  textKey: "text",
  metadataKeys: ["my_custom_id"],
});

// Existing: still works as before
const store = await WeaviateStore.initialize(embeddings, {
  client,
  schema: {
    name: "MyCollection",
    properties: [...],
  },
});
```

## Test plan

- [x] All existing unit tests pass (`pnpm test` in `libs/providers/langchain-weaviate`)
- [x] Four new unit tests added for `jsonSchema` / `schema` / precedence / skip-if-exists paths
- [x] TypeScript compiles with no errors in `vectorstores.ts`
- [ ] Integration tests against a live Weaviate instance (requires `WEAVIATE_URL` + credentials)

### DIFF
--- a/libs/providers/langchain-weaviate/src/vectorstores.ts
+++ b/libs/providers/langchain-weaviate/src/vectorstores.ts
@@ -306,7 +306,9 @@ export class WeaviateStore extends VectorStore {
    */
   async hybridSearch(
     query: string,
-    options?: HybridOptions<undefined, undefined, undefined>
+    options?: HybridOptions<undefined, undefined, undefined> & {
+      filter?: FilterValue;
+    }
   ): Promise<Document[]> {
     const collection = this.client.collections.get(this.indexName);
     let query_vector: number[] | undefined;
@@ -314,8 +316,10 @@ export class WeaviateStore extends VectorStore {
       query_vector = await this.embeddings.embedQuery(query);
     }
 
+    const { filter, ...restOptions } = options ?? {};
     const options_with_vector = {
-      ...options,
+      ...restOptions,
+      filters: restOptions.filters ?? filter,
       vector: options?.vector || query_vector,
       returnMetadata: [
         "score",
@@ -364,23 +368,26 @@ export class WeaviateStore extends VectorStore {
   async generate(
     query: string,
     generate: GenerateOptions<undefined, GenerativeConfigRuntime | undefined>,
-    options?: BaseHybridOptions<undefined, undefined, undefined>
+    options?: BaseHybridOptions<undefined, undefined, undefined> & {
+      filter?: FilterValue;
+    }
   ): Promise<WeaviateDocument[]> {
     const collection = this.client.collections.get(this.indexName);
+    const { filter, ...restOptions } = options ?? {};
+    const hybridOptions = {
+      ...restOptions,
+      filters: restOptions.filters ?? filter,
+    };
     let result;
     if (this.tenant) {
       result = await collection
         .withTenant(this.tenant)
-        .generate.hybrid(
-          query,
-          { ...(generate || {}) },
-          { ...(options || {}) }
-        );
+        .generate.hybrid(query, { ...(generate || {}) }, hybridOptions);
     } else {
       result = await collection.generate.hybrid(
         query,
         { ...(generate || {}) },
-        { ...(options || {}) }
+        hybridOptions
       );
     }
     const documents = [];
--- a/libs/providers/langchain-weaviate/src/vectorstores.ts
+++ b/libs/providers/langchain-weaviate/src/vectorstores.ts
@@ -12,6 +12,7 @@ import {
   type WeaviateField,
   BaseHybridOptions,
   MetadataKeys,
+  type WeaviateClass,
 } from "weaviate-client";
 import {
   type MaxMarginalRelevanceSearchOptions,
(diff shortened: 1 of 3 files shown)


libs/providers/langchain-weaviate/src/vectorstores.ts
```diff
@@ -155,7 +163,10 @@ export class WeaviateStore extends VectorStore {
       weaviateStore.indexName
     );
     if (!collection) {
-      if (weaviateStore.schema) {
+      if (weaviateStore.jsonSchema) {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        await weaviateStore.client.collections.createFromJson(weaviateStore.jsonSchema as any);
```
hntrl: what type does createFromJson accept? Can we just use that as the same type from within the args type?

libs/providers/langchain-weaviate/src/vectorstores.ts
```diff
@@ -120,11 +123,16 @@ export class WeaviateStore extends VectorStore {
     super(embeddings, args);
 
     this.client = args.client;
-    this.indexName = args.indexName || args.schema?.name || "";
+    this.indexName =
+      args.indexName ||
+      args.schema?.name ||
+      (args.jsonSchema?.["class"] as string | undefined) ||
```
hntrl: is `class` a standard json schema property?

hntrl (commented):
If you're looking to go the extra mile, you might consider accepting `StandardSerializableSchema` (a type exported from core) and depend on the tooling around that to orchestrate the json schema parameter. This lets you bring any json schema type that is compliant with a standard spec for validation

libs/providers/langchain-weaviate/src/vectorstores.ts
dudanogueira: Good catch — fixed. `createFromJson` accepts `WeaviateClass` (from `weaviate-client/openapi/types`, re-exported through `weaviate-client`'s `collections` index). I switched `WeaviateLibArgs.jsonSchema` to `WeaviateClass` directly, which dropped both the `Record<string, unknown>` and the `as any` cast.

libs/providers/langchain-weaviate/src/vectorstores.ts
dudanogueira: Not a standard JSON Schema property — `class` here is Weaviate's legacy REST field name for the collection identifier (renamed conceptually to "collection name" but the wire field is still `class` for backward compatibility, see the `WeaviateClass` type in `weaviate-client/openapi/types`). Now that the field is typed as `WeaviateClass` the access is type-safe, and I left a comment in the constructor flagging this so it isn't confused with the JSON Schema `class` keyword.

dudanogueira:
@hntrl Thanks for the pointer to `StandardSerializableSchema` / `SerializableSchema`! I looked at it and I don't think it lines up well with this surface — `SerializableSchema` is a runtime-validation + JSON-Schema-emitting type (Zod/Valibot/etc., shaped around `withStructuredOutput` and parsing LLM output). The payload Weaviate's `createFromJson` consumes is a different artifact: it's Weaviate's own collection-definition document (`WeaviateClass`) with fields like `vectorConfig`, `multiTenancyConfig`, `invertedIndexConfig`, property tokenization, quantization, etc. — none of which map onto a Standard Schema spec.

The intent of this field is specifically to let users round-trip a schema produced by Weaviate's REST API or `client.collections.exportToJson()` (and from the schema-builder tool linked in the PR description), which is why I'd rather keep it tightly typed as `WeaviateClass` and let the existing `schema: CollectionConfigCreate` path cover the programmatic-builder case. Happy to revisit if I'm misreading the suggestion.

christian-bromann (dismissed):
This makes sense to me 👍 

@hntrl any objections?

christian-bromann (approved):

hntrl (approved):

PR #10824, opened 2026-05-04
feat(mcp-adapters): parallelize server initialization in MultiServerMCPClient

ssabariram:
## Summary

- Replace sequential `for...of` loop in `initializeConnections()` with `Promise.allSettled()` so all MCP servers connect concurrently
- Significantly reduces startup time when multiple servers are configured (e.g., 5 servers × 2s each: ~10s → ~2s)
- Add `#initPromise` deduplication guard to prevent duplicate work when `initializeConnections()` is called concurrently from multiple callers (e.g., `getTools()` and `getClient()` at the same time)

## Details

**Error handling modes preserved:**
- `"throw"` — all servers are attempted in parallel; if any fail, throws `MCPClientError`
- `"ignore"` — failed servers tracked in `#failedServers`, successful tools returned
- Custom function handler — called independently for each failure

**Behavioral note:** In `"throw"` mode, the previous implementation stopped at the first failure (sequential). Now all servers are attempted before throwing. This is strictly better — you get all successful connections plus visibility into all failures — while still satisfying the contract "throws if any server fails."

## Test plan

- [x] New test: parallel timing proof (3 servers × 50ms completes in ~52ms, not ~150ms)
- [x] New test: mixed success/failure in parallel with ignore mode
- [x] New test: custom error handler called for each failure independently
- [x] New test: throw mode throws when any server fails in parallel
- [x] New test: concurrent `initializeConnections()` calls deduplicate
- [x] All 35 existing tests in `client.basic.test.ts` still pass
- [x] Full package test suite passes (153 tests pass, 1 pre-existing type error in `hooks.test.ts` unrelated to this change)


.changeset/parallel-mcp-initialization.md
```diff
@@ -0,0 +1,5 @@
+---
+"@langchain/mcp-adapters": minor
```
hntrl:
```suggestion
"@langchain/mcp-adapters": patch
```

hntrl (dismissed):

hntrl (dismissed):

ssabariram:
@hntrl I have fixed the formatting issues which reverted your approval, can you please look into it

PR #10828, opened 2026-05-05
fix(core): workaround CJS module export bug for uuid utility

spotlight21c:
## Description
This PR fixes an issue introduced in `@langchain/core` (`v1.1.44`) where the `uuid` utility functions were incorrectly exported as module objects instead of functions in the CommonJS build. 

To work around this bundler interop issue, this PR refactors `libs/langchain-core/src/utils/uuid/index.ts` to explicitly import each function and assign them to exported variables.

## Context
When downstream packages (like `@langchain/google`) try to call `uuid.v4()`, it currently throws a `TypeError: is not a function` in CommonJS environments (such as NestJS).


christian-bromann:
Duplicate of #10791

PR #10829, opened 2026-05-05
fix(agents): resolve generic type inference for createAgent state parameters

spotlight21c:
## Description
This PR fixes a TypeScript inference bug in `createAgent` where the `invoke` parameter type fails to resolve correctly when options like `middleware` or `stateSchema` are omitted and the agent is assigned implicitly. 

## Reproduction
In an implicit assignment, TypeScript throws an error on `.invoke()`:

```typescript
import { createAgent, tool } from "langchain";

const agent = createAgent({
  model: "openai:gpt-4",
  tools: [/* tools */],
  systemPrompt: '...',
});

// ❌ TypeScript Error:
// Argument of type '{ messages: { role: string; content: string; }[]; }' is not assignable to parameter of type 'never'.
await agent.invoke({
  messages: [{ role: 'user', content: "message" }],
});
```

## The Issue
When `Types["State"]` or `Types["Middleware"]` are inferred as broad default bounds (e.g., `readonly AgentMiddleware[]`), TypeScript delays the evaluation of distributive conditional types (like `T extends readonly [] ? ...`). This deferred evaluation results in an unresolved cross-type for `.invoke()`, causing valid inputs (e.g., `{ messages: [...] }`) to be incorrectly flagged as type errors.

## The Fix
Disabled the distributive behavior in the internal middleware and schema type helpers by wrapping the conditional parameters in tuples (e.g., `[T] extends [readonly []] ? ...`). This forces eager evaluation of the types, fully resolving the `InvokeStateParameter` even when explicit bounds are omitted.

### DIFF
--- a/libs/langchain/src/agents/middleware/types.ts
+++ b/libs/langchain/src/agents/middleware/types.ts
@@ -670,9 +670,9 @@ export type InferMiddlewareInputState<T extends AgentMiddleware> =
  * Helper type to infer merged state from an array of middleware (just the middleware states)
  */
 export type InferMiddlewareStates<T extends readonly AgentMiddleware[]> =
-  T extends readonly []
+  [T] extends [readonly []]
     ? {}
-    : T extends readonly [infer First, ...infer Rest]
+    : [T] extends [readonly [infer First, ...infer Rest]]
       ? First extends AgentMiddleware
         ? Rest extends readonly AgentMiddleware[]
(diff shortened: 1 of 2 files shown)


christian-bromann:
Please provide a reproduction case of the type issue you are seeing, I couldn't reproduce it. Will go ahead and close this but happy to re-open if we can get this reproduced.

PR #10835, opened 2026-05-05
fix(core): deno re-declaration in LangSmith is causing unit test failures

colifran:
### Summary

Unit tests are failing due to the following:
```
⎯⎯⎯⎯⎯⎯ Unhandled Errors ⎯⎯⎯⎯⎯⎯

Vitest caught 2 unhandled errors during the test run.
This might cause false positive tests. Resolve unhandled errors to make sure your tests are not affected.

⎯⎯⎯ Unhandled Source Error ⎯⎯⎯
TypeCheckError: Cannot redeclare block-scoped variable 'Deno'.
 ❯ src/utils/env.ts:3:9
      1| // Inlined from https://github.com/flexdinesh/browser-or-node
      2| declare global {
      3|   const Deno:
       |         ^
      4|     | {
      5|         version: {


⎯⎯⎯ Unhandled Source Error ⎯⎯⎯
TypeCheckError: Property 'env' does not exist on type '{ version { deno string; }; }'.
 ❯ src/utils/env.ts:86:20
     84|       return process.env?.[name];
     85|     } else if (isDeno()) {
     86|       return Deno?.env.get(name);
       |                    ^
     87|     } else {
     88|       return undefined;

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
```

### Root Cause
The langsmith 0.5.21 → 0.6.0 bump introduced a conflicting declare global { const Deno } declaration.

@langchain/core and langsmith both declare const Deno in a declare global block. Since const is block-scoped, TypeScript treats the second declaration as a redeclaration error. LangSmith's declaration only includes version (no env), and the conflict causes TypeScript to resolve to the narrower type.

This adds @ts-expect-error directives to suppress both errors. The conflict is environment-dependent and it only surfaces when LangSmith types are in the typecheck scope (e.g. in CI).

### DIFF
--- a/libs/langchain-core/src/utils/env.ts
+++ b/libs/langchain-core/src/utils/env.ts
@@ -1,6 +1,6 @@
 // Inlined from https://github.com/flexdinesh/browser-or-node
 declare global {
-  // @ts-ignore Conflicts with langsmith's ambient Deno declaration
+  // @ts-expect-error Conflicts with langsmith's ambient Deno declaration
   const Deno:
     | {
         version: {
@@ -84,7 +84,7 @@ export function getEnvironmentVariable(name: string): string | undefined {
       // oxlint-disable-next-line no-process-env
       return process.env?.[name];
     } else if (isDeno()) {
-      // @ts-ignore Langsmith's Deno declaration is missing `env`
+      // @ts-expect-error Langsmith's Deno declaration is missing `env`
       return Deno?.env.get(name);
     } else {
       return undefined;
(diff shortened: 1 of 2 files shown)


libs/langchain-core/src/utils/env.ts
```diff
@@ -1,16 +1,9 @@
...
-        };
-      }
-    | undefined;
-}
+type Deno = {
+  version?: { deno: string };
+  env?: { get: (name: string) => string | undefined };
+};
+
+const global = globalThis as typeof globalThis & { Deno?: Deno };
```
hntrl: instead of type widening like this, I think I'd rather just have @ts-expect-error comment explaining why we have it like this

colifran: Updated.

hntrl (approved):

Issue #10218, opened 2026-03-02
Structured Output Missing from Agent Response


upwindchange:
### Checked other resources

- [x] This is a bug, not a usage question. For questions, please use the LangChain Forum (https://forum.langchain.com/).
- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain.js documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain.js rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

## Impact

- **Severity**: Very High
- **Scope**: All agents using `toolStrategy` with Zod schemas
- **Workaround**: None - structured output is completely inaccessible
- **User Impact**: Runtime crashes when accessing `response.structuredResponse.fieldName`

### Example Code

```ts
const evaluatorAgent = createAgent({
	model: complexLangchainModel(),
	responseFormat: toolStrategy(
		z.object({
			is_task_successful: z
				.boolean()
				.describe(
					"boolean: true if task is successful, false if task failed",
				),
			result_explanation: z
				.string()
				.describe("text explanation of the task result"),
		}),
		{ handleError: true },
	),
	systemPrompt: evaluatorPrompt,
	middleware: retryMiddleware,
});

// Invoke EVALUATOR (structured output)
log.info("[action-executor] Invoking evaluator agent for subtask", i, ":", currentSubtask.label);
const evaluationResponse = await evaluatorAgent.invoke({
	messages: [evaluationMessage],
});

// DIAGNOSTIC LOGGING: Check the response structure
log.info("[action-executor] Evaluator response keys:", Object.keys(evaluationResponse));
log.info("[action-executor] Has structuredResponse?", "structuredResponse" in evaluationResponse);

if ("structuredResponse" in evaluationResponse) {
	log.info("[action-executor] structuredResponse value:", JSON.stringify(evaluationResponse.structuredResponse, null, 2));
	log.info("[action-executor] structuredResponse type:", typeof evaluationResponse.structuredResponse);
} else {
	log.error("[action-executor] structuredResponse is MISSING from response!");
}

// Log messages to check for retries
log.info("[action-executor] Number of messages in response:", evaluationResponse.messages?.length);
if (evaluationResponse.messages && evaluationResponse.messages.length > 0) {
[...]

PR #10879, opened 2026-05-09
fix(langchain): throw on terminal providerStrategy parse failure

### LINKED ISSUE #10878: createAgent with responseFormat silently returns undefined structuredResponse when LLM output doesn't match schema
### Checked other resources

### Example Code

// Schema with required fields
const responseSchema = {
  type: "object",
  properties: {
    assistantMeta: {
      type: "object",
      properties: {
        cartStatus: { type: "string", enum: ["loaded", "empty"] },  // REQUIRED
        currentSearchKey: { type: "string" },
        queryContext: { type: "string" },
      },
      required: ["cartStatus", "currentSearchKey", "queryContext"],
      additionalProperties: false,
    },
    nextStep: { 
      type: "string", 
      enum: ["search", "proceed", "clarify"] 
    },
  },
  required: ["assistantMeta", "nextStep"],
  additionalProperties: false,
};

const model = new ChatOpenAI({ model: "gpt-4o" });

const agent = createAgent({
  llm: model,
  tools: [],
  responseFormat: responseSchema,
});

// Invoke with a prompt that causes LLM to omit required field "cartStatus"
const response = await agent.invoke({
  messages: [{ role: "user", content: "echo back: {\"assistantMeta\":{\"currentSearchKey\":\"test\"},\"nextStep\":\"search\"}" }],
});

console.log("structuredResponse:", response.structuredResponse);
// Expected: Error or warning about missing required field "cartStatus"
// Actual: undefined (SILENT FAILURE!)

console.log("text:", response.messages?.at(-1)?.content);
// Contains valid JSON, but structuredResponse is undefined!

No error is thrown. That's the bug.

When LLM returns JSON that doesn't match the schema (e.g., missing required field "cartStatus"):
- response.structuredResponse = undefined
- No error thrown
- No warning logged
- Agent returns successfully with no indication of failure

Debug logs from our investigation:
response object keys: messages, variables, metadata, error, status, structuredResponse
response.structuredResponse: undefined
response.messages[-1].content: {"assistantMeta":{"currentSearchKey":"test"},\"nextStep":"search"}

Note: The message content contains valid JSON from the LLM, but "structuredResponse" 
is undefined because the JSON is missing the required "cartStatus" field. 
No error or warning is provided.

### Problem

When using `createAgent()` with `responseFormat` for structured output, if the LLM returns valid JSON that doesn't fully match the schema (e.g., missing a required field), the `structuredResponse` in the agent's response remains `undefined` with **no error, no warning, and no indication of failure**.

### What I'm doing

Using `createAgent` with a JSON schema that has required fields:
- Schema requires `assistantMeta.cartStatus` (required field)
- LLM returns JSON missing this field
- Using OpenAI's structured output mode via `responseFormat`

### What I expect

When the LLM output doesn't match the schema, I expect one of:
1. An error thrown explaining schema validation failed
2. A warning logged indicating which field(s) are invalid
3. An error field in the response (e.g., `response.structuredResponseError`)
4. At minimum: clear documentation about this silent failure behavior

### What actually happens

- `response.structuredResponse` = `undefined`
- No error thrown
- No warning logged
- `response.messages[-1].content` contains the raw JSON (which IS valid JSON, just missing required fields)
- Agent returns success status
- Downstream code that relies on `structuredResponse` silently fails

### Impact

This causes silent failures in production:
- Code expecting `structuredResponse.assistantMeta.cartStatus` gets `undefined`
- No debugging information available
- Developers have no way to know that schema validation failed
- The message content has the JSON, but `structuredResponse` is empty

### Note

I also raised a similar issue for `createReactAgent` in `langgraphjs` ([issue #2338](https://github.com/langchain-ai/langgraphjs/issues/2338)). This is the same bug pattern but in `createAgent` from the `langchain` package.

### Workaround

We implemented a fallback to parse from message content when `structuredResponse` is empty:

if (structuredOutputConfig.enable) {
  if (response.structuredResponse && Object.keys(response.structuredResponse).length > 0) {
    result.structuredOutput = response.structuredResponse;
  } else if (text) {
    try {
      result.structuredOutput = JSON.parse(text);
    } catch { /* ignore */ }
  }
}

### System Info

Package versions:
- langchain: 1.3.3
- @langchain/core: 1.1.39
- @langchain/openai: 1.4.1

Environment:
- Platform: Windows 10
- Node.js: v22.x
- Package Manager: npm

vignesh-gep:
Closes #10878. Related to #10218 / #10242 (which fixed a different facet of the same silent-failure umbrella in the `toolStrategy` + `wrapModelCall` path).

## Problem

When `createAgent` is configured with `responseFormat` resolving to a `providerStrategy` — either passed explicitly via `providerStrategy(schema)` or auto-promoted from a bare Zod / JSON schema for any model whose `profile.structuredOutput` is `true` (e.g. `gpt-4o`, recent Anthropic / Google models) — and the LLM produces a **terminal** response (no `tool_calls`) whose text either:

- isn't valid JSON, or
- is valid JSON but doesn't satisfy the schema (e.g. missing a required field),

…the agent **silently** resolves with `structuredResponse: undefined` and no thrown error. Downstream code then crashes with `TypeError: Cannot read properties of undefined`.

### Reproducer (also in #10878)

```ts
const agent = createAgent({
  model: new ChatOpenAI({ model: "gpt-4o" }),
  tools: [],
  responseFormat: z.object({
    assistantMeta: z.object({
      cartStatus: z.enum(["loaded", "empty"]), // required
    }),
  }),
});

const res = await agent.invoke({
  messages: [{ role: "user", content: 'echo: {"assistantMeta":{}}' }],
});

res.structuredResponse; // undefined — no error, no warning
```

### Why #10242 doesn't cover this

PR #10242 surgically fixed the case where a `wrapModelCall` middleware was substituting a structured-output retry `Command` with `lastAiMessage` and losing the retry intent. That fix only applies to the `toolStrategy` path where retries are configured. The bug above is in a different branch: `providerStrategy`, terminal response, no retry — `ProviderStrategy.parse` in `responses.ts` simply returns `undefined` on a failed JSON parse / failed schema validation, and the failure isn't propagated.

## Fix

**File:** `libs/langchain/src/agents/nodes/AgentNode.ts` — inside the `baseHandler` closure built by `AgentNode.#invokeModel`.

After `ProviderStrategy.parse()` returns `undefined`, distinguish the two reasons that can happen:

- The model is mid-conversation (it returned `tool_calls` and we'll come back to `baseHandler` after the tools run) → return the AIMessage unchanged so the agent loop continues, exactly as before.
- The model produced a **terminal** response (no `tool_calls`) and parsing failed → throw a `StructuredOutputParsingError` (the existing class from `errors.ts`) so the failure surfaces at the exact node where it occurred, with a clear stack trace, instead of being silently dropped.

```ts
if (structuredResponseFormat?.type === "native") {
  const structuredResponse =
    structuredResponseFormat.strategy.parse(response);
  if (structuredResponse) {
    return { structuredResponse, messages: [response] };
  }

  // NEW: only the no-tool-calls branch is added; the tool-calls flow is unchanged.
  if (!response.tool_calls || response.tool_calls.length === 0) {
    const schemaTitle =
      typeof structuredResponseFormat.strategy.schema?.title === "string"
        ? structuredResponseFormat.strategy.schema.title
        : "providerStrategy";
    throw new StructuredOutputParsingError(schemaTitle, [
      "Model output did not satisfy the provided response schema.",
    ]);
  }

  return response;
}
```

The error message is intentionally short and specific — it names the schema title (or falls back to `providerStrategy`) and states that the model output did not satisfy the schema. It does **not** include the raw model output, mirroring the conservative approach taken in PR [langchain-ai/langgraphjs#2339](https://github.com/langchain-ai/langgraphjs/pull/2339) for the deprecated `createReactAgent` path.

### Why this layer

- Throwing at the parse boundary surfaces the failure at the exact node where it occurred, with a clear stack trace.
- The alternative — adding a retry / `handleError` path here — would duplicate the existing `toolStrategy` retry machinery and broaden the public API surface (`providerStrategy()` currently has no `handleError` option). That's a separate, larger change and is out of scope for this fix.
- A user who wants retry behavior on parse failure can already opt in via `toolStrategy(schema, { handleError: true })`.

Total diff in `AgentNode.ts`: **+24 / −1** (1 widened import, 1 inserted guard block) — same surgical scale as #10242.

## Test

Added 3 regression tests in `libs/langchain/src/agents/tests/responses.test.ts` under `describe("providerStrategy") > describe("error handling on parse failure")`:

1. Terminal response that isn't valid JSON → expects `StructuredOutputParsingError`.
2. Terminal response that is valid JSON but doesn't satisfy the schema (the exact shape from #10878 — valid JSON missing a required field) → expects `StructuredOutputParsingError`.
3. **Bare** Zod `responseFormat` (auto-promoted to `providerStrategy` because `FakeToolCallingChatModel.profile.structuredOutput === true`) that fails to parse → expects an error matching `/structured output/i`.

All three tests use the existing `FakeToolCallingChatModel` test util (no new test infrastructure) to deterministically produce a terminal AI message whose content fails the schema. **All three fail on `main`** with `AssertionError: promise resolved "{ messages: [ …(2) ] }" instead of rejecting`, and **pass with this fix**.

Local validation:

- `pnpm --filter langchain test` — **53 test files, 781 tests, all passed**, no type errors (includes the 3 new ones above).
- `pnpm lint` — 0 warnings, 0 errors across 1906 files.
- `pnpm format:check` — all files correctly formatted.

## Out of scope / not changed

- Public API of `createAgent`, `responseFormat`, `providerStrategy`, `toolStrategy` — unchanged.
- The `toolStrategy` retry machinery — unchanged; #10242 covers that path.
- Provider profile metadata / `model.profile.structuredOutput`-based strategy selection — unchanged; #10489 covers strategy selection.

## Changeset

Included: `.changeset/fix-createagent-providerstrategy-silent-failure.md` — `langchain` patch bump. The changeset bot has already detected it on the PR.

### DIFF
--- a/libs/langchain/src/agents/tests/responses.test.ts
+++ b/libs/langchain/src/agents/tests/responses.test.ts
@@ -16,7 +16,6 @@ import {
   ProviderStrategy,
   ToolStrategy,
 } from "../responses.js";
-import { StructuredOutputParsingError } from "../errors.js";
 
 function makeSerializableSchema(
   jsonSchema: Record<string, unknown> = {
--- a/libs/langchain/src/agents/tests/responses.test.ts
+++ b/libs/langchain/src/agents/tests/responses.test.ts
@@ -533,6 +533,8 @@ describe("structured output handling", () => {
     });
 
     describe("error handling on parse failure", () => {
+      const errorMessage = /did not satisfy the provided response/;
+
       it("should throw when a terminal response cannot be parsed as JSON", async () => {
         const model = fakeModel().respond(
           new AIMessage({ content: "I cannot answer that question." })
@@ -549,7 +551,7 @@ describe("structured output handling", () => {
 
         await expect(
           agent.invoke({ messages: [{ role: "user", content: "hi" }] })
-        ).rejects.toThrow(StructuredOutputParsingError);
+        ).rejects.toThrow(errorMessage);
       });
 
       it("should throw when a terminal response is valid JSON but does not satisfy the schema", async () => {
@@ -568,7 +570,7 @@ describe("structured output handling", () => {
 
         await expect(
           agent.invoke({ messages: [{ role: "user", content: "hi" }] })
-        ).rejects.toThrow(StructuredOutputParsingError);
+        ).rejects.toThrow(errorMessage);
       });
 
       it("should throw when a bare Zod responseFormat auto-promoted to providerStrategy fails to parse", async () => {
@@ -587,7 +589,7 @@ describe("structured output handling", () => {
 
         await expect(
           agent.invoke({ messages: [{ role: "user", content: "hi" }] })
-        ).rejects.toThrow(/structured output/i);
+        ).rejects.toThrow(errorMessage);
       });
     });
 
--- a/libs/langchain/src/agents/tests/responses.test.ts
+++ b/libs/langchain/src/agents/tests/responses.test.ts
@@ -14,6 +14,7 @@ import {
   ProviderStrategy,
   ToolStrategy,
 } from "../responses.js";
+import { StructuredOutputParsingError } from "../errors.js";
 
 function makeSerializableSchema(
   jsonSchema: Record<string, unknown> = {
@@ -526,6 +527,67 @@ describe("structured output handling", () => {
             messages: [{ role: "user", content: "hi" }],
           })
         ).resolves.not.toThrowError();
+      });
+    });
+
+    describe("error handling on parse failure", () => {
+      it("should throw when a terminal response cannot be parsed as JSON", async () => {
+        const model = new FakeToolCallingChatModel({
+          responses: [
+            new AIMessage({ content: "I cannot answer that question." }),
+          ],
+        });
+        const agent = createAgent({
+          model,
+          tools: [],
+          responseFormat: providerStrategy(
+            z.object({
+              temperature: z.number(),
+            })
+          ),
+        });
+
+        await expect(
+          agent.invoke({ messages: [{ role: "user", content: "hi" }] })
+        ).rejects.toThrow(StructuredOutputParsingError);
+      });
+
+      it("should throw when a terminal response is valid JSON but does not satisfy the schema", async () => {
+        const model = new FakeToolCallingChatModel({
+          responses: [new AIMessage({ content: '{"foo":"bar"}' })],
+        });
+        const agent = createAgent({
+          model,
+          tools: [],
+          responseFormat: providerStrategy(
+            z.object({
+              temperature: z.number(),
+            })
+          ),
+        });
+
+        await expect(
+          agent.invoke({ messages: [{ role: "user", content: "hi" }] })
+        ).rejects.toThrow(StructuredOutputParsingError);
+      });
+
+      it("should throw when a bare Zod responseFormat auto-promoted to providerStrategy fails to parse", async () => {
+        const model = new FakeToolCallingChatModel({
+          responses: [
+            new AIMessage({ content: "I cannot answer that question." }),
+          ],
+        });
+        const agent = createAgent({
+          model,
+          tools: [],
+          responseFormat: z.object({
+            temperature: z.number(),
+          }),
+        });
+
+        await expect(
+          agent.invoke({ messages: [{ role: "user", content: "hi" }] })
+        ).rejects.toThrow(/structured output/i);
       });
     });
 
--- a/libs/langchain/src/agents/tests/responses.test.ts
+++ b/libs/langchain/src/agents/tests/responses.test.ts
@@ -7,6 +7,8 @@ import { ChatAnthropic } from "@langchain/anthropic";
 import { AIMessage } from "@langchain/core/messages";
 import type { SerializableSchema } from "@langchain/core/utils/standard_schema";
 
+import { fakeModel } from "@langchain/core/testing";
+
 import { createAgent, toolStrategy, providerStrategy } from "../index.js";
 import { FakeToolCallingModel, FakeToolCallingChatModel } from "./utils.js";
 import {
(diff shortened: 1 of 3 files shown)


libs/langchain/src/agents/tests/responses.test.ts
```diff
@@ -529,6 +530,67 @@ describe("structured output handling", () => {
       });
     });
 
+    describe("error handling on parse failure", () => {
+      it("should throw when a terminal response cannot be parsed as JSON", async () => {
+        const model = new FakeToolCallingChatModel({
```
christian-bromann: Can we use our new primitive for new tests here [`fakeModel`](https://docs.langchain.com/oss/javascript/langchain/test/unit-testing#mock-chat-model-with-fakemodel)?

christian-bromann (changes requested):

libs/langchain/src/agents/tests/responses.test.ts
vignesh-gep: Thanks for the pointer! Migrated tests 1 and 2 to fakeModel(). Kept test 3 on FakeToolCallingChatModel because it specifically exercises the bare-Zod → providerStrategy auto-promotion path, which is gated by hasSupportForJsonSchemaOutput(model) reading model.profile.structuredOutput === true. fakeModel inherits the default empty profile: {} from BaseLanguageModel and doesn't expose a way to override it, so the auto-promotion would silently fall through to toolStrategy and the test would no longer cover the intended path. Happy to migrate that one too if there's an idiomatic way to set profile.structuredOutput on fakeModel that I missed - and would be glad to open a separate small PR adding e.g. fakeModel().withProfile({ structuredOutput: true }) to langchain-core/testing if you think that'd be useful.

christian-bromann (dismissed):
LGTM 👍 

Thank you!

vignesh-gep:
Hi @christian-bromann  quick heads up, the more explicit assertion commit left the StructuredOutputParsingError import unused, which was failing typecheck in CI. Pushed a one-line fix (a6c7b62) to remove it. Could you re-enable auto-merge/review when you get a moment? Thanks!

christian-bromann (approved):

PR #10884, opened 2026-05-12
fix(perplexity): Add useResponsesApi flag to ChatPerplexity

jliounis:
## Summary

Adds a `useResponsesApi` option to `ChatPerplexity` so callers can route requests through Perplexity's **Agent API** (Responses-compatible) instead of Chat Completions. The flag mirrors `ChatOpenAI`'s `useResponsesApi` field in shape, placement, and behavior.

Perplexity exposes two HTTP surfaces on the same client:

| Surface | Endpoint | SDK method |
| --- | --- | --- |
| Chat Completions | `POST /chat/completions` | `client.chat.completions.create()` |
| Agent API | `POST /v1/agent` (alias: `POST /v1/responses`) | `client.responses.create()` |

The Agent API supports built-in tools (`web_search`, `fetch_url`, `finance_search`, `people_search`), `instructions`, `input`, `previous_response_id`, and `include` — none of which exist on Chat Completions. Today `ChatPerplexity` only talks to Chat Completions; this PR makes it able to talk to both, gated by a single flag.

## What changes

- New `useResponsesApi?: boolean` on `PerplexityChatInput`.
- New call-options fields on `PerplexityChatCallOptions` for `tools`, `previousResponseId`, `instructions`, `input`, and `include` so they can be passed via `bind` / call options.
- `_generate` and `_streamResponseChunks` now branch on a private `_useResponsesApi(payload)` and call `client.responses.create()` when routing to the Agent API.
- Module-level helpers `_isBuiltinTool` / `_useResponsesApi`, and exported `convertResponsesToChatResult` / `convertResponsesEventToChunk` to map Agent API responses and streaming events into `ChatResult` / `ChatGenerationChunk`.
- Private `_toResponsesPayload(payload)` that translates the Chat-Completions-shaped payload into the Responses shape (`messages` → `input`; Perplexity-only knobs go under `extra_body`). Per the [OpenAI-compatibility docs](https://docs.perplexity.ai/docs/agent-api/openai-compatibility), the Agent API alias accepts `max_tokens`, so it's passed through unchanged.

### Auto-detection rules (when `useResponsesApi` is unset)

- Route to Responses when the payload contains any **built-in Perplexity tool** (any `tools` entry whose `type` is not `"function"`).
- Route to Responses when the payload contains a **Responses-only field** (`previous_response_id`, `instructions`, `input`, `include`).
- Otherwise route to Chat Completions (today's default).

`useResponsesApi: true` / `useResponsesApi: false` override the auto-detection.

### Backwards compatibility

Additive only. Existing `new ChatPerplexity({ model: "sonar", ... })` users see no behavior change — the auto-detection only flips to Responses when the request explicitly uses an Agent API feature.

## Test plan

- [x] `pnpm --filter @langchain/perplexity test` — 77 passed, 1 skipped (10 new tests added).
- [x] `pnpm --filter @langchain/perplexity build` clean (CJS + ESM, attw + publint pass).
- [x] `pnpm lint libs/providers/langchain-perplexity` — 0 warnings, 0 errors.
- [x] `pnpm format:check` — clean.

New unit tests in `src/tests/chat_models_responses.test.ts`:

1. Auto-detect: `tools: [{ type: "web_search" }]` → Responses.
2. Auto-detect: `previousResponseId` via call options → Responses.
3. Auto-detect: plain text → Chat Completions.
4. Explicit `useResponsesApi: true` → Responses (even for plain text).
5. Explicit `useResponsesApi: false` → Chat Completions (even with built-in tools).
6. Response → `AIMessage` conversion (content, `usage_metadata`, `response_metadata`).
7. `function_call` output item → `aiMessage.tool_calls[0]`.
8. Streaming: async iterator of `response.output_text.delta` events produces matching content chunks; `response.completed` carries `usage_metadata`.
9. `response.error` event raises.
10. Unhandled event types return `null` (no chunk emitted).

Net new feature; no upstream issue.

### DIFF
--- a/.changeset/rotten-apes-fetch.md
+++ b/.changeset/rotten-apes-fetch.md
@@ -1,0 +1,5 @@
+---
+"@langchain/perplexity": patch
+---
+
+Add useResponsesApi flag to ChatPerplexity
--- a/libs/providers/langchain-perplexity/src/chat_models.ts
+++ b/libs/providers/langchain-perplexity/src/chat_models.ts
@@ -6,7 +6,9 @@ import {
   ChatMessageChunk,
   HumanMessageChunk,
   SystemMessageChunk,
+  UsageMetadata,
 } from "@langchain/core/messages";
+import { ToolCall } from "@langchain/core/messages/tool";
 import { concat } from "@langchain/core/utils/stream";
 import {
   BaseChatModel,
@@ -148,6 +150,18 @@ export interface PerplexityChatInput extends BaseChatModelParams {
 
   /** Configuration for web search behaviour. */
   webSearchOptions?: WebSearchOptions;
+
+  /**
+   * Whether to use the Perplexity Agent API (Responses-compatible) instead of Chat Completions.
+   *
+   * If `undefined` (default), inferred from the request payload: `true` when the request
+   * uses a built-in Perplexity tool (`web_search`, `fetch_url`, `finance_search`, `people_search`)
+   * or any Responses-only field (`previousResponseId`, `instructions`, `input`, `include`).
+   * `false` otherwise.
+   *
+   * Maps to `client.responses.create()` → `POST /v1/agent` (alias: `/v1/responses`).
+   */
+  useResponsesApi?: boolean;
 }
 
 export interface PerplexityChatCallOptions extends BaseChatModelCallOptions {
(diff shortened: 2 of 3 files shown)


jliounis:
Docs counterpart staged as draft: [langchain-ai/docs#3965](https://github.com/langchain-ai/docs/pull/3965). Will mark ready-for-review once this PR merges and the next `@langchain/perplexity` release ships.

hntrl (dismissed):

hntrl (approved):

PR #10886, opened 2026-05-12
fix(openai): expose Azure deployment name in tracing and invocation p…

### LINKED ISSUE #10874: AzureChatOpenAI: deployment name missing from callback/tracing run data (ls_model_name defaults to gpt-3.5-turbo)
## Description

When using `AzureChatOpenAI` from `@langchain/openai`, the Azure deployment name is not exposed in the LangChain callback/run data. `run.extra.metadata.ls_model_name` and `run.extra.invocation_params.model` both show `"gpt-3.5-turbo"` (the hardcoded default) instead of the actual deployment name.

This breaks all downstream observability/tracing tools (OpenTelemetry instrumentors, LangSmith, etc.) that rely on run metadata to identify the model.

## Reproduction

```typescript
import { AzureChatOpenAI } from "@langchain/openai";

const model = new AzureChatOpenAI({
  azureOpenAIApiDeploymentName: "gpt-4o-mini",
  azureOpenAIApiInstanceName: "my-instance",
  azureOpenAIApiKey: "...",
  azureOpenAIApiVersion: "2024-08-01-preview",
});
```

In a LangChain BaseTracer callback, `run.extra` for an LLM run shows:
```
run.extra.metadata.ls_model_name → "gpt-3.5-turbo"  (wrong)
run.extra.invocation_params.model → "gpt-3.5-turbo"  (wrong)
run.extra.invocation_params.azureOpenAIApiDeploymentName → undefined
```

The actual deployment name `"gpt-4o-mini"` is stored on `this.azureOpenAIApiDeploymentName` but never flows into the callback run data.

## Root Cause

`BaseChatOpenAI` sets `model = "gpt-3.5-turbo"` as the default (line 35 of `base.ts`). When using `AzureChatOpenAI` with only `azureOpenAIApiDeploymentName` (and no explicit `model`), `this.model` stays as the default.

- `getLsParams()` sets `ls_model_name: this.model` → `"gpt-3.5-turbo"`
- `invocationParams()` returns `{ model: this.model, ... }` → `"gpt-3.5-turbo"`
- Neither method is overridden by `AzureChatOpenAI`

## Python parity

The Python `AzureChatOpenAI` already fixes this in two ways:

1. `_identifying_params` includes `azure_deployment: self.deployment_name`
2. `_get_ls_params()` falls back to `self.deployment_name` when `self.model_name` is None

```python
# Python langchain_openai/chat_models/azure.py
@property
def _identifying_params(self):
    return {"azure_deployment": self.deployment_name, **super()._identifying_params}
```

See also: https://github.com/langchain-ai/langchain/issues/24838 (Python equivalent)

## Suggested Fix

Override `getLsParams()` in `AzureChatOpenAI` to fall back to the deployment name:

```typescript
getLsParams(options) {
  const params = super.getLsParams(options);
  if (this.azureOpenAIApiDeploymentName && 
      (!params.ls_model_name || params.ls_model_name === "gpt-3.5-turbo")) {
    params.ls_model_name = this.azureOpenAIApiDeploymentName;
  }
  return params;
}
```

And/or override `invocationParams()` to include the deployment name:

```typescript
invocationParams(options) {
  return {
    ...super.invocationParams(options),
    azureOpenAIApiDeploymentName: this.azureOpenAIApiDeploymentName,
  };
}
```

## Impact

All callback-based instrumentation that reads model identity from LangChain run data is affected, including:
- OpenTelemetry distros (microsoft/opentelemetry-distro-javascript)
- LangSmith tracing
- Any custom BaseTracer implementations

## Versions
- `@langchain/openai`: ^1.4.4
- `langchain`: ^1.2.31
- Node.js: 24.x

obielin:
This Fixes #10874

When `AzureChatOpenAI` is constructed with only `azureOpenAIApiDeploymentName` and no explicit `model`, `this.model` falls back to the `BaseChatOpenAI` default of `"gpt-3.5-turbo"`. This default then leaks into:

- `run.extra.metadata.ls_model_name` via `getLsParams()`
- `run.extra.invocation_params.model` via `invocationParams()`

…breaking every downstream tracer that reads model identity from LangChain run data (LangSmith, OpenTelemetry distros, custom `BaseTracer`s).

## Changes

- Overrides `getLsParams()` in `AzureChatOpenAI` to fall back to `azureOpenAIApiDeploymentName` when the model is unset or still at the `"gpt-3.5-turbo"` default. An explicitly-set `model` is preserved.
- Overrides `invocationParams()` to also include `azureOpenAIApiDeploymentName`, so callback/run metadata carries the deployment name regardless of whether `model` was set.

This brings parity with the Python implementation: langchain-ai/langchain#24838.

## Tests

I added three unit tests in `libs/providers/langchain-openai/src/azure/chat_models/tests/index.test.ts`:

1. `getLsParams` falls back to the deployment name when no explicit `model` is set.
2. `getLsParams` preserves an explicitly set `model` (no clobbering).
3. `invocationParams` exposes `azureOpenAIApiDeploymentName`.

All 247 tests in `@langchain/openai` pass locally. `pnpm format` and `pnpm --filter @langchain/openai lint` are clean.


obielin:
Hi @jkennedyvz, this PR is ready, but the merge is blocked because the CodeQL/code-scanning workflow is waiting to run on commits 3f31aa8 / 75980e2. It needs a maintainer to approve running workflows on a fork PR. Could you approve the run when you have a moment? The branch is up to date with main, and the other checks pass. It's a small fix for #10874 (Azure deployment name leaking a wrong model into tracing metadata), and a merged Microsoft OpenTelemetry PR (microsoft/opentelemetry-distro-javascript#127) is currently carrying a workaround pending this fix. Happy to adjust anything. Thanks.

JacksonWeber:
Thanks for putting this together! While validating the fix against real LangChain callback payloads, I found that updating only the outer `AzureChatOpenAI` class does not cover all execution paths.

`AzureChatOpenAI` delegates requests to `AzureChatOpenAICompletions` or `AzureChatOpenAIResponses`. Calls wrapped with `.withConfig()` or `.bindTools()` execute through those delegate classes, so they continue to publish the inherited `gpt-3.5-turbo` default in both `ls_model_name` and `invocation_params.model`.

I opened #11376 to apply the correction in both delegate classes. It covers bare, configured, and tool-bound calls across the Chat Completions and Responses APIs, while preserving explicitly configured models.

I’m happy to coordinate on whether those changes should be incorporated here or whether #11376 should supersede this PR. The important part is ensuring the delegate execution paths are included so the callback/tracing issue is fully resolved.

obielin:
Hi @JacksonWeber, thanks for digging into this, good catch on the delegate classes, I hadn't traced the .withConfig()/.bindTools() paths through to AzureChatOpenAICompletions/AzureChatOpenAIResponses.

I'd lean toward incorporating your delegate class fix into this PR rather than shipping two separate PRs for the same underlying issue (#10874) ,  that keeps one changeset, one test suite and avoids reviewers having to reconcile two overlapping fixes. If you're open to it, I can pull your changes from #11376 into this branch (with credit in the commit) and extend my existing tests to cover the bare/configured/tool-bound cases you mentioned. Or if you'd rather drive it from #11376 directly, happy to close this one out and review yours instead.

Let me know which you'd prefer  either way works for me as long as we land one coherent fix. 

Thank you

JacksonWeber:
> Hi Jackson Weber (@JacksonWeber), thanks for digging into this, good catch on the delegate classes, I hadn't traced the .withConfig()/.bindTools() paths through to AzureChatOpenAICompletions/AzureChatOpenAIResponses.
> 
> I'd lean toward incorporating your delegate class fix into this PR rather than shipping two separate PRs for the same underlying issue (#10874) , that keeps one changeset, one test suite and avoids reviewers having to reconcile two overlapping fixes. If you're open to it, I can pull your changes from #11376 into this branch (with credit in the commit) and extend my existing tests to cover the bare/configured/tool-bound cases you mentioned. Or if you'd rather drive it from #11376 directly, happy to close this one out and review yours instead.
> 
> Let me know which you'd prefer either way works for me as long as we land one coherent fix.
> 
> Thank you

That'd be great, I'd be happy to have my PR rolled into yours, just let me know if you have any questions when incorporating my PR and I can take a look/test after that's done. Thanks again for working on this!

obielin:
Thanks @JacksonWeber — just pushed the incorporation as `eede464f2`.

Your delegate-level fix is in via shared helpers in `common.ts`, applied in both `AzureChatOpenAICompletions` and `AzureChatOpenAIResponses`, with you credited as co-author. Two adjustments on top of #11376:

- Fixed a typecheck error in the helper — `params` was typed `ReturnType<BaseChatOpenAI["getLsParams"]>`, but `BaseChatOpenAI` is generic and needs a type argument. It now uses `LangSmithParams`, the actual return type.
- Dropped my outer `invocationParams` override. It surfaced the deployment by adding an `azureOpenAIApiDeploymentName` key, which leaked into the request body sent to Azure. The deployment is now surfaced by correcting the existing `model` key instead, so the body's `model` carries the deployment name rather than the `gpt-3.5-turbo` default. Routing is unaffected — the Azure client is constructed with `deployment`, so the target comes from the URL and not from the body. There's a test asserting both halves of that.

Tests now cover the bare / `.withConfig()` / `.bindTools()` matrix across both the Chat Completions and Responses APIs. Worth noting the `withConfig` path is a sharper version of what you spotted: it rebuilds the model as a plain `ChatOpenAI`, so the Azure subclass is dropped entirely and only the delegates survive — which is exactly why correcting the outer class couldn't hold.

Ready for you to take a look and test whenever you have a moment. Thanks again for catching this.

libs/providers/langchain-openai/src/azure/chat_models/completions.ts
```diff
@@ -82,6 +99,7 @@ export class AzureChatOpenAICompletions<
   ) {
     const fields = getAzureChatOpenAIParams(deploymentOrFields, fieldsArg);
     super(fields);
+    this.modelWasExplicitlyConfigured = fields?.model != null;
```
JacksonWeber: Could this also check `fields?.modelName != null`?  BaseChatOpenAI resolves `model ?? modelName`, so an explicitly configured `modelName: "gpt-3.5-turbo"` is currently overwritten by the deployment fallback. The same applies to both delegate classes; a regression test would be helpful.

obielin: Good catch — confirmed and fixed in `d42f24f37`.

You're right that `BaseChatOpenAI` resolves `model ?? modelName` (`base.ts:545`), so deriving the guard from `model` alone left `modelName` looking unset. I reproduced it before changing anything: with `modelName: "gpt-3.5-turbo"` and a deployment, `chat.model` correctly resolved to `gpt-3.5-turbo` but both `ls_model_name` and `invocation_params.model` came back as the deployment name.

Worth noting the blast radius is narrower than it first looks — the fallback also requires `=== DEFAULT_OPENAI_MODEL`, so `modelName: "gpt-4o"` was already safe. It only bites when someone explicitly configures the literal default, which is exactly the case a naive check can't distinguish from "nobody set anything."

Rather than duplicate the condition, I pulled it into a shared helper in `common.ts`:

```ts
export function wasModelExplicitlyConfigured(fields?: AzureChatOpenAIFields) {
  return fields?.model != null || fields?.modelName != null;
}
```

Both `AzureChatOpenAICompletions` and `AzureChatOpenAIResponses` call it now. The two paths having drifted is what let this through in the first place, so a single definition felt worth the indirection.

On the regression test: the existing "preserves an explicit model" case is now parameterized over `model` × `modelName` × `{"explicit-model", "gpt-3.5-turbo"}` × both APIs, plus a new six-case block asserting an explicit `modelName` survives in real callbacks across bare / `.withConfig()` / `.bindTools()` on both protocols. I checked they aren't vacuous — with the source fix stashed, exactly the eight `gpt-3.5-turbo` cases fail and the rest pass.

Full `@langchain/openai` suite is at 331 passing (up from 319), no new type errors, format and lint clean.

Ready for another look whenever you have a moment — thanks for staying on this one.

Issue #10505, opened 2026-03-23
bug(@langchain/openai): SyntaxError when streaming with json_schema structured output and function tools via Responses API


juhosuominen:
### Checked other resources

- [x] This is a bug, not a usage question. For questions, please use the LangChain Forum (https://forum.langchain.com/).
- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain.js documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain.js rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

A minimal reproduction is available at: https://github.com/juhosuominen/langchain-openai-tool-structured-output-issue

```ts
import { ChatOpenAI } from "@langchain/openai";
import { HumanMessage } from "@langchain/core/messages";
import { tool } from "@langchain/core/tools";
import { z } from "zod";

const lookupTool = tool(
  async ({ query }) => `Mock result for: ${query}`,
  {
    name: "lookup",
    description: "Look up factual information",
    schema: z.object({ query: z.string() }),
  },
);

async function main() {
  const model = new ChatOpenAI({
    model: "gpt-5.2",
    useResponsesApi: true,
    streaming: true,
  });

  // Bind with json_schema structured output and a function tool.
  // Force the model to call the tool so the bug triggers reliably.
  const bound = model.bindTools([lookupTool], {
    response_format: {
      type: "json_schema",
      json_schema: {
        name: "response",
        strict: true,
        schema: {
          type: "object",
          additionalProperties: false,
          required: ["answer"],
          properties: {
            answer: { type: "string" },
          },
        },
      },
    },
    tool_choice: "required",
  });

  console.log("Streaming with json_schema format + function tool (tool_choice: required)...\n");

  try {
    const stream = await bound.stream([
      new HumanMessage("Look up what LangChain is."),
    ]);

    for await (const chunk of stream) {
      const tc = chunk.tool_call_chunks;
      if (tc?.length) process.stdout.write(`[tool_call_chunk] `);
      else process.stdout.write(".");
    }

    console.log("\n\nStream completed without error (bug not triggered).");
  } catch (error) {
    if (
      error instanceof SyntaxError &&
      error.message.includes("Unexpected end of JSON input")
    ) {
      console.error("\n\n=== BUG REPRODUCED ===");
[...]

PR #10895, opened 2026-05-15
fix(openai): guard bare JSON.parse in Responses API converter against trailing characters

### LINKED ISSUE #10894: bug(@langchain/openai): Bare JSON.parse in Responses API converter crashes on structured output with trailing characters
### Checked other resources

### Example Code

```ts
import { z } from "zod";

  model: "gpt-5-mini",
  useResponsesApi: true,
});

const structuredModel = model.withStructuredOutput(
  z.object({
    status: z.string(),
    plan: z.object({ steps: z.array(z.object({ tool: z.string(), input: z.string() })) }),
  })
);

// Intermittently crashes with SyntaxError when gpt-5-mini appends
// trailing characters after the valid JSON object
const result = await structuredModel.invoke("Search for USB-C connectors");
```

```
SyntaxError: Unexpected non-whitespace character after JSON at position 570
    at JSON.parse (<anonymous>)
    at convertResponsesDeltaToChatGenerationChunk (responses.ts:752)
    at ChatOpenAIResponses._streamResponseChunks (responses.ts:144)
```

### Description

**Related to #10505 / PR #10507, but a different failure mode.**

#10505 fixed the case where `msg.text` is **empty** (function-call-only responses). This issue covers the case where `msg.text` contains **valid JSON followed by trailing non-whitespace characters**, causing `JSON.parse` to throw.

#### Root cause

In `convertResponsesDeltaToChatGenerationChunk` ([responses.ts line 752](https://github.com/langchain-ai/langchainjs/blob/main/libs/providers/langchain-openai/src/converters/responses.ts#L752)):

```ts
if (event.response.text?.format?.type === "json_schema" && msg.text) {
  additional_kwargs.parsed ??= JSON.parse(msg.text);  // <-- bare JSON.parse
}
```

When `gpt-5-mini` returns structured output with `response_format: { type: "json_schema" }`, it intermittently appends trailing characters after the valid JSON object (e.g., extra whitespace, partial tokens, or control characters). `JSON.parse` throws `SyntaxError: Unexpected non-whitespace character after JSON at position N`.

#### Impact

We observed this in production over 7 days (100+ LangSmith traces):
- **93% of all inner agent errors** were this JSON parse crash (not actual timeouts)
- **14 out of 15** of these errors succeeded when retried — the failure is transient
- Primarily affects `gpt-5-mini` with `service_tier: "auto"` on the Responses API
- The error propagates as an unhandled exception, killing the entire streaming response

#### Suggested fix

Wrap the `JSON.parse` in a try/catch so it degrades gracefully:

```ts
if (event.response.text?.format?.type === "json_schema" && msg.text) {
  try {
    additional_kwargs.parsed ??= JSON.parse(msg.text);
  } catch {
    // Model returned text that isn't valid JSON — leave parsed unset.
    // Callers that need structured output can fall back to their own parsing.
  }
}
```

This preserves backward compatibility — `additional_kwargs.parsed` is simply `undefined` when parsing fails, which is the same state as when no json_schema format is used. Callers already handle this case.

### System Info

```
@langchain/openai: 1.4.4 (also confirmed on 1.4.3 and 1.4.1)
@langchain/core: 1.1.39
Node.js: v22
Platform: Linux
```

### Copy paste this code to reproduce

```ts
// This is intermittent and depends on model behavior.
// To reproduce reliably, you need to make many calls with
// gpt-5-mini + json_schema response_format via the Responses API.
// ~5-10% of calls will produce trailing characters.
```

BertBR:
## Summary

`convertResponsesDeltaToChatGenerationChunk` in `libs/providers/langchain-openai/src/converters/responses.ts` calls `JSON.parse(msg.text)` directly when `response.text.format.type === "json_schema"`. Some models (observed with `gpt-5-mini` on `service_tier: "auto"`) intermittently emit **trailing non-whitespace characters** (extra tokens, control characters) after a valid JSON object on the Responses API. `JSON.parse` accepts trailing whitespace but rejects trailing non-whitespace, so the bare parse throws `SyntaxError` and kills the entire streaming response mid-flight.

This PR wraps the call in a `try`/`catch`:

```ts
if (event.response.text?.format?.type === "json_schema" && msg.text) {
  try {
    additional_kwargs.parsed ??= JSON.parse(msg.text);
  } catch {
    // graceful degradation; leave parsed undefined, caller can fall back
    // to msg.text or retry. See #10894.
  }
}
```

On failure `additional_kwargs.parsed` is left `undefined`, the rest of the message converts normally, the stream completes, and `usage_metadata` still flows through.

The downstream `withStructuredOutput(zodSchema)` pipeline already gives the right semantics from there:

- The `RunnableLambda` at `base.ts:1089` falls through to `altParser.invoke(content)` when `additional_kwargs.parsed` is absent.
- For Zod schemas, `createContentParser` returns `StructuredOutputParser`, whose `.parse()` already wraps `JSON.parse(...)` in try/catch and throws a typed `OutputParserException` (`libs/langchain-core/src/output_parsers/structured.ts:126-131`).
- `assembleStructuredOutputPipeline` already routes `includeRaw: true` through `withFallbacks`, so a typed throw becomes `parsed: null`. `includeRaw: false` passes the typed throw to the caller.

Net effect for the user reporting #10894: streams no longer die mid-flight; `includeRaw: true` degrades to `{ raw, parsed: null }`; `includeRaw: false` throws a typed `OutputParserException` the caller can catch and retry.

Closes #10894.

## Test plan

Converter-level unit tests in `libs/providers/langchain-openai/src/converters/tests/responses.test.ts`:

1. **Trailing non-whitespace case** — `msg.text = '{"status":"ok","plan":{"steps":[]}}x'` (valid JSON + trailing `x`). The conversion does not throw, `additional_kwargs.parsed` is `undefined`, `usage_metadata` is populated.
2. **Well-formed regression guard** — `msg.text` contains valid JSON with no trailing characters. The conversion parses cleanly into `additional_kwargs.parsed`.

End-to-end `withStructuredOutput(zodSchema)` streaming tests in `libs/providers/langchain-openai/src/chat_models/tests/index.test.ts`:

3. **`includeRaw: true` + malformed structured output** — returns `{ raw, parsed: null }`.
4. **`includeRaw: false` (default) + malformed structured output** — throws `OutputParserException` matching `/Failed to parse/i` (post-stream, not the mid-stream `SyntaxError` that originally killed the call).
5. **Well-formed structured output regression guard** — `withStructuredOutput(zodSchema)` returns the parsed object.

```
pnpm vitest run src/converters/tests/responses.test.ts src/chat_models/tests/index.test.ts
→ PASS (126) FAIL (0)
pnpm build → clean
```

## Notes

- Follows the same shape as the prior fix for #10505 (empty `msg.text` case, also in this file).
- Changeset under `.changeset/openai-responses-bare-json-parse-10894.md` as `patch` on `@langchain/openai`.
- Review feedback from @christian-bromann addressed in follow-up commit `4b35a2f` (NUL byte in test literal swapped for explicit non-whitespace `x`; E2E coverage added for both `includeRaw` modes).

### DIFF
--- a/libs/providers/langchain-openai/src/converters/responses.ts
+++ b/libs/providers/langchain-openai/src/converters/responses.ts
@@ -762,7 +762,15 @@ export const convertResponsesDeltaToChatGenerationChunk: Converter<
     usage_metadata = convertResponsesUsageToUsageMetadata(event.response.usage);
 
     if (event.response.text?.format?.type === "json_schema" && msg.text) {
-      additional_kwargs.parsed ??= JSON.parse(msg.text);
+      try {
+        additional_kwargs.parsed ??= JSON.parse(msg.text);
+      } catch {
+        // Some models (observed with gpt-5-mini on service_tier: "auto")
+        // intermittently emit trailing characters after a valid JSON object
+        // in the Responses API path. Throw-on-parse kills the entire stream
+        // even though msg.text is otherwise usable. Leave parsed undefined
+        // and let the caller fall back to msg.text or retry. See #10894.
+      }
     }
     for (const [key, value] of Object.entries(event.response)) {
       if (key === "id") continue;
--- a/libs/providers/langchain-openai/src/converters/tests/responses.test.ts
+++ b/libs/providers/langchain-openai/src/converters/tests/responses.test.ts
@@ -2224,10 +2224,12 @@ describe("convertResponsesDeltaToChatGenerationChunk - json_schema with tool cal
   });
 });
 
-describe("convertResponsesDeltaToChatGenerationChunk - json_schema with trailing characters (#10894)", () => {
-  it("should not throw when response text contains valid JSON followed by trailing characters", () => {
+describe("convertResponsesDeltaToChatGenerationChunk - json_schema with trailing non-whitespace characters (#10894)", () => {
+  it("should not throw when response text contains valid JSON followed by a trailing non-whitespace character", () => {
     // gpt-5-mini on service_tier: "auto" intermittently emits trailing
-    // characters after a valid JSON object on the Responses API. Previously
+    // characters after a valid JSON object on the Responses API. JSON.parse
+    // accepts trailing whitespace, so the failure mode is specifically
+    // trailing non-whitespace (extra tokens, control characters). Previously
     // the bare JSON.parse threw SyntaxError and killed the stream. The
     // conversion should now degrade gracefully: additional_kwargs.parsed is
     // left undefined, the rest of the message converts normally, and the
@@ -2253,7 +2255,7 @@ describe("convertResponsesDeltaToChatGenerationChunk - json_schema with trailing
             content: [
               {
                 type: "output_text",
-                text: '{"status":"ok","plan":{"steps":[]}} ',
+                text: '{"status":"ok","plan":{"steps":[]}}x',
                 annotations: [],
               },
             ],
--- a/libs/providers/langchain-openai/src/converters/tests/responses.test.ts
+++ b/libs/providers/langchain-openai/src/converters/tests/responses.test.ts
@@ -2224,6 +2224,149 @@ describe("convertResponsesDeltaToChatGenerationChunk - json_schema with tool cal
   });
 });
 
+describe("convertResponsesDeltaToChatGenerationChunk - json_schema with trailing characters (#10894)", () => {
+  it("should not throw when response text contains valid JSON followed by trailing characters", () => {
+    // gpt-5-mini on service_tier: "auto" intermittently emits trailing
+    // characters after a valid JSON object on the Responses API. Previously
+    // the bare JSON.parse threw SyntaxError and killed the stream. The
+    // conversion should now degrade gracefully: additional_kwargs.parsed is
+    // left undefined, the rest of the message converts normally, and the
+    // caller can fall back to msg.text or retry.
+    const event = {
+      type: "response.completed",
+      response: {
+        id: "resp_test_10894",
+        model: "gpt-5-mini",
+        object: "response",
+        created_at: 1700000000,
+        status: "completed",
+        incomplete_details: null,
+        metadata: {},
+        user: null,
+        service_tier: "auto",
+        output: [
+          {
+            type: "message",
+            id: "msg_001",
+            role: "assistant",
+            status: "completed",
+            content: [
+              {
+                type: "output_text",
+                text: '{"status":"ok","plan":{"steps":[]}} ',
+                annotations: [],
+              },
+            ],
+          },
+        ],
+        text: {
+          format: {
+            type: "json_schema",
+            name: "response",
+            strict: true,
+            schema: {
+              type: "object",
+              additionalProperties: false,
+              required: ["status", "plan"],
+              properties: {
+                status: { type: "string" },
+                plan: { type: "object" },
+              },
+            },
+          },
+        },
+        usage: {
+          input_tokens: 30,
+          output_tokens: 12,
+          total_tokens: 42,
+          input_tokens_details: { cached_tokens: 0 },
+          output_tokens_details: { reasoning_tokens: 0 },
+        },
+      },
+    };
+
+    expect(() =>
+      convertResponsesDeltaToChatGenerationChunk(event as any)
+    ).not.toThrow();
+
+    const result = convertResponsesDeltaToChatGenerationChunk(event as any);
+    expect(result).not.toBeNull();
+
+    const message = result!.message as AIMessageChunk;
+    // Parsing failed, so parsed should not be populated.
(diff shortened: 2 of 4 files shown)


libs/providers/langchain-openai/src/converters/responses.ts
```diff
@@ -762,7 +762,15 @@ export const convertResponsesDeltaToChatGenerationChunk: Converter<
     usage_metadata = convertResponsesUsageToUsageMetadata(event.response.usage);
 
     if (event.response.text?.format?.type === "json_schema" && msg.text) {
-      additional_kwargs.parsed ??= JSON.parse(msg.text);
+      try {
+        additional_kwargs.parsed ??= JSON.parse(msg.text);
+      } catch {
```
christian-bromann, line 767: I don’t think this fully fixes the linked `withStructuredOutput()` repro. After this catch, `additional_kwargs.parsed` is absent, so `ChatOpenAI.withStructuredOutput()` falls through to its fallback content parser. For the same malformed streamed text, that parser can throw again, just later in the structured-output pipeline.

Could we add an end-to-end structured-output streaming test for the `withStructuredOutput(zodSchema)` path and either make the fallback parser behavior graceful too, or set an explicit parsed failure sentinel that preserves the existing includeRaw/non-includeRaw semantics?

libs/providers/langchain-openai/src/converters/tests/responses.test.ts
```diff
@@ -2224,6 +2224,149 @@ describe("convertResponsesDeltaToChatGenerationChunk - json_schema with tool cal
...
+        output: [
+          {
+            type: "message",
+            id: "msg_001",
+            role: "assistant",
+            status: "completed",
+            content: [
+              {
+                type: "output_text",
+                text: '{"status":"ok","plan":{"steps":[]}} ',
```
christian-bromann: Nit: this test uses a hidden NUL/control character, while the PR description says “trailing space”. `JSON.parse` accepts trailing whitespace, so the bug is really about trailing non-whitespace characters. Could we make the suffix explicit, e.g. `\u0000` or `x`, and update the wording accordingly?

christian-bromann (changes requested):
2 comments

libs/providers/langchain-openai/src/converters/tests/responses.test.ts
BertBR: Right call. Fixed the literal to `'{"status":"ok","plan":{"steps":[]}}x'` (trailing `x`, non-whitespace). Updated the describe and `it` titles plus the changeset wording to say "trailing non-whitespace characters". Confirmed on disk the original had a stray NUL byte (`\x00`); replaced at byte level so the test now actually exercises the SyntaxError path.

libs/providers/langchain-openai/src/converters/responses.ts
BertBR: Looked into this. The chunk-level fix prevents the SyntaxError from killing the stream mid-flight, which was the original bug surface. For the downstream `withStructuredOutput(zodSchema)` path the existing pipeline already gives the semantics you described:

- The `RunnableLambda` at `base.ts:1089` falls through to `altParser.invoke(content)` when `additional_kwargs.parsed` is absent (which is the case after our catch).
- For Zod schemas, `createContentParser` returns `StructuredOutputParser`, whose `.parse()` already wraps `JSON.parse(...)` in a try/catch and throws a typed `OutputParserException` (`libs/langchain-core/src/output_parsers/structured.ts:126-131`).
- `assembleStructuredOutputPipeline` already routes `includeRaw: true` through `withFallbacks`, so a typed throw becomes `parsed: null`. `includeRaw: false` passes the throw to the caller.

Added end-to-end coverage in `libs/providers/langchain-openai/src/chat_models/tests/index.test.ts` under a new describe block scoped to #10894. Three tests:

1. `includeRaw: true` plus malformed structured output returns `{ raw, parsed: null }`.
2. `includeRaw: false` (default) plus malformed structured output throws `OutputParserException` matching `/Failed to parse/i`.
3. Well-formed structured output returns the parsed object (regression guard).

All 126 tests in `responses.test.ts` and `chat_models/tests/index.test.ts` pass; `pnpm build` is clean.

If you would rather have an explicit sentinel field on `additional_kwargs` instead of relying on the post-stream throw chain, happy to switch. The current shape preserves the existing semantics and does not add new public surface area.

christian-bromann: Thanks, that makes sense. The added end-to-end coverage clarifies the intended semantics: `includeRaw: true` degrades to `parsed: null`, while the default path still surfaces a typed parser failure after the stream completes.

I’m good with keeping the current shape and not adding a sentinel field, since it preserves the existing structured-output behavior and avoids new public surface area.

christian-bromann (approved):
LGTM 👍

PR #10923, opened 2026-05-19
fix(core): allow input_json_delta aggregation for anthropic tool_use …

### LINKED ISSUE #10907: @langchain/core 1.1.46 breaks withStructuredOutput streaming for Anthropic models
### Description

`@langchain/core` 1.1.46 introduced a regression that breaks `withStructuredOutput(...).stream()` for all Anthropic models (with and without thinking enabled). The final parsed result is an empty string, causing `AnthropicToolsOutputParser` to throw:

```
Error: Failed to parse. Text: """". Error: "Unexpected end of JSON input"
Troubleshooting URL: https://docs.langchain.com/oss/javascript/langchain/errors/OUTPUT_PARSING_FAILURE/
```

### Root Cause

PR #10790 ("keep different content block types separate when merging chunks") added `hasMismatchedMergeableType` / `getMergeableTypeBase` to `_findMergeTarget` in `messages/base.js`. This logic strips `_delta` suffixes to compare content block types at matching indices.

The problem is a naming asymmetry in Anthropic's streaming protocol:
- **Start event** emits: `{ index: N, type: "tool_use", input: "" }`
- **Delta events** emit: `{ index: N, type: "input_json_delta", input: "<partial_json>" }`

The new normalization logic does:
- `getMergeableTypeBase("tool_use")` → `"tool_use"`
- `getMergeableTypeBase("input_json_delta")` → `"input_json"` (strips `_delta`)

Since `"tool_use" !== "input_json"`, the merge is rejected. Delta chunks are appended as separate array entries instead of merging their `input` field into the `tool_use` block.

The final `message.content` array ends up with:
```js
{ type: "tool_use", input: "" }              // input stayed empty
{ type: "input_json_delta", input: "{" }     // appended, not merged
{ type: "input_json_delta", input: "\"key" } // appended, not merged
// ...
```

`AnthropicToolsOutputParser.parseResult()` then finds `block.input: ""` on the `tool_use` block and `JSON.parse("")` throws.

### Reproduction

```typescript
import { ChatAnthropic } from '@langchain/anthropic';
import { z } from 'zod';

const schema = z.object({ answer: z.string() });
const model = new ChatAnthropic({ model: 'claude-sonnet-4-20250514' });
const structured = model.withStructuredOutput(schema);

const stream = await structured.stream('What is 2 + 2? Answer in one word.');
let result;
  result = chunk;
}
console.log(result); // throws OUTPUT_PARSING_FAILURE
```

### Versions

- `@langchain/core`: 1.1.46 (broken), 1.1.45 (works)
- `@langchain/anthropic`: 1.3.26
- Node.js: 22.x

### Bisection Results

| Version | `withStructuredOutput` stream |
|---------|-------------------------------|
| 1.1.39–1.1.45 | PASS |
| 1.1.46 | FAIL |

### Suggested Fix

`getMergeableTypeBase` needs to handle the Anthropic naming convention where tool use deltas use `input_json_delta` (not `tool_use_delta`). Options:
1. Add a mapping: `input_json_delta` → `tool_use`
2. Only apply the type-mismatch check for types that follow the `<base>_delta` convention
3. Skip the check entirely when one type is `tool_use` and the other contains `_delta`

noishey:
### Description
Fixes a critical regression introduced in `v1.1.46` that breaks `withStructuredOutput().stream()` for Anthropic models. 

### Root Cause
Anthropic's streaming protocol uses an asymmetric naming convention during tool execution:
1. Initial frame emits: `type: "tool_use"`
2. Subsequent delta frames emit: `type: "input_json_delta"`

`getMergeableTypeBase()` normalizes `input_json_delta` to `input_json`. Because `"input_json" !== "tool_use"`, the strict nominal type check introduced in PR #10790 flags this as a type mismatch and rejects the merge, throwing an `OUTPUT_PARSING_FAILURE`.

### Solution
Explicitly map `input_json_delta` and `input_json` to return `"tool_use"` within `getMergeableTypeBase()`. This bridges the structural protocol gap, satisfies the type invariance check, and allows streamed JSON fragments to aggregate seamlessly.

### Related Issues
Closes #10907

---
**Author:** [@noishey](https://twitter.com/butn0tshy)


christian-bromann:
Thanks for the PR! Do you think we can add unit tests, e.g. in the Anthropic provider package verifying this works?

neuralmanacle:
Absolutely! I can definitely add some unit tests for this. I'll update the PR as soon as they are ready!

christian-bromann (approved):
Thank you! 👍 

LGTM

libs/providers/langchain-anthropic/src/tests/chat_models.test.ts
```diff
@@ -1285,6 +1285,52 @@ describe("Streaming tool call consolidation (input_json_delta handling)", () =>
       input: { prompt: "hello" },
     });
   });
+
+  test("should successfully aggregate tool_use and input_json_delta chunks via .concat()", async () => {
```
christian-bromann: It seems like this test doesn't use any primitives exported in the Anthropic package, so can we move it into the core workspace?

christian-bromann (changes requested):
one minor nit

libs/providers/langchain-anthropic/src/tests/chat_models.test.ts
neuralmanacle: Sure.

christian-bromann: Can we move this test into `@langchain/core`?

neuralmanacle: Sure.

libs/providers/langchain-anthropic/src/tests/chat_models.test.ts
```diff
@@ -1111,179 +1111,99 @@ describe("Tool search beta auto-append", () => {
 
     expect(
       paramsWithoutToolSearch.betas === undefined ||
-        !paramsWithoutToolSearch.betas.includes("advanced-tool-use-2025-11-20")
+      !paramsWithoutToolSearch.betas.includes("advanced-tool-use-2025-11-20")
     ).toBe(true);
   });
 });
+test("AIMessage with input_json_delta blocks falls back to index matching when tool_calls is empty", async () => {
```
christian-bromann: Can we revert all changes in here?

neuralmanacle: Sure.

PR #10932, opened 2026-05-20
fix(deepseek): make <think> tag parsing opt-in via enableThinkTagParsing

aierns:
## Summary

`<think>` tag parsing in `_streamResponseChunks` (added in #9726)
is now **disabled by default** and opt-in via `enableThinkTagParsing: true`.

The official DeepSeek API now returns reasoning content in
the dedicated `reasoning_content` field, which `ChatOpenAICompletions` already
handles. The `<think>` tag parser was only needed for older/third-party
endpoints, but when enabled by default it incorrectly strips `<think>` tags
from normal content produced by non-reasoning models like `deepseek-chat` or `deepseek-v4-flash`/`deepseek-v4-pro` with thinking disabled.

## Changes

- Added `enableThinkTagParsing?: boolean` option to `ChatDeepSeekInput`
  (default `false`)
- Added pass-through path in `_streamResponseChunks` that skips `<think>`
  tag parsing when the option is disabled
- Updated existing tests to pass `enableThinkTagParsing: true`
- Added two new tests: default behavior (no parsing) and opt-in behavior
- Added changeset

## Test plan

- [x] `pnpm --filter @langchain/deepseek test` — 18/18 pass
- [x] `pnpm --filter @langchain/deepseek build`
- [x] `pnpm format`

image

**Thanks for reviewing.**

### DIFF
--- a/.changeset/fix-deepseek-think-tag-parsing.md
+++ b/.changeset/fix-deepseek-think-tag-parsing.md
@@ -1,5 +1,5 @@
 ---
-"@langchain/deepseek": patch
+"@langchain/deepseek": minor
 ---
 
 fix(deepseek): disable <think> tag parsing by default, add enableThinkTagParsing option
--- a/.changeset/fix-deepseek-think-tag-parsing.md
+++ b/.changeset/fix-deepseek-think-tag-parsing.md
@@ -1,0 +1,15 @@
+---
+"@langchain/deepseek": patch
+---
+
+fix(deepseek): disable <think> tag parsing by default, add enableThinkTagParsing option
+
+`<think>...</think>` tag parsing in `_streamResponseChunks` (added in #9726) is
+now disabled by default and opt-in via `enableThinkTagParsing: true`. The
+official DeepSeek API (`api.deepseek.com`) returns reasoning content in the
+dedicated `reasoning_content` field (already handled by `ChatOpenAICompletions`). When this parsing is enabled by default, it
+incorrectly strips `<think>` tags from normal content produced by non-reasoning
+models like `deepseek-chat` (`deepseek-v4-flash` with thinking disabled), or deepseek-v4-flash/deepseek-v4-pro with thinking disabled, moving that content into `reasoning_content` and leaving the response content corrupted.
+
+When using third-party proxies or older API endpoints that embed reasoning in
+`<think>` tags, set `enableThinkTagParsing: true`.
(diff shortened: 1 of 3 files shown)


Sajith15:
The fix is correct and the approach is right - making tag parsing opt-in is the right call since `api.deepseek.com` already surfaces reasoning via the `reasoning_content` field, so the `<think>` parser is redundant there and actively harmful for non-reasoning models. A few things worth addressing before merge:

**1. Semver should be `minor`, not `patch`**
This is a behavioral breaking change for anyone running DeepSeek-R1 through a local/proxy endpoint (ollama, vllm, LM Studio, etc.) that embeds reasoning in `<think>` tags. After upgrading, their `reasoning_content` will silently stop being populated and raw `<think>` tags will appear in content instead. No error, no warning - just broken behavior until they find the changelog and add `enableThinkTagParsing: true`. That warrants `minor` semver at minimum.

**2. Redundant class-level initializer**
```ts
// current
private enableThinkTagParsing: boolean = false;  // <- redundant
// constructor also does:
this.enableThinkTagParsing = fields.enableThinkTagParsing ?? false;
```
Since the constructor always assigns the value, the `= false` on the field declaration is noise. Either remove the class-level `= false` and keep the constructor assignment, or vice versa.

**3. Missing test: native `reasoning_content` passthrough when flag is off**
The new tests cover `<think>` tags with the flag on/off, but there's no test for the primary production case: a mock API response that returns `reasoning_content` natively in the response object with `enableThinkTagParsing: false`. This is the case that matters most for official API users - worth confirming it still populates `additional_kwargs.reasoning_content` correctly.

**4. Consider a `console.warn` for discoverability**
Since the failure mode is silent (raw `<think>` tags in content, empty `reasoning_content`), affected users may spend a while debugging. A one-time warn like `"DeepSeek: <think> tags detected in stream but enableThinkTagParsing is false. Set enableThinkTagParsing: true if using a proxy or local endpoint."` would help without being obtrusive.

**5. Non-streaming path**
Is there a non-streaming equivalent to `_streamResponseChunks` that also had `<think>` tag handling? The diff only touches the streaming path - worth a quick check that `_generate` / `invoke` don't also need an update (or confirming they were never affected).

Overall: logic is sound, tests cover the core cases well, good changeset explanation. The semver point is the most important to address.

aierns:
Thanks for the detailed review, @Sajith15 .

I’ve pushed an update addressing your feedback:

* Changed the changeset from `patch` to `minor`.
* Removed the redundant class-level default initializer and kept the constructor assignment.
* Added a test to confirm native `reasoning_content` still works when `enableThinkTagParsing` is false.
* Added a one-time `console.warn` when streamed `<think>` tags are detected while parsing is disabled.
* Confirmed the non-streaming path was not affected; native `reasoning_content` is preserved for `invoke` as well.

Please take another look when you get a chance.

PR #10944, opened 2026-05-22
feat(mongodb): add $rerank support to MongoDBAtlasVectorSearch

caseyclements:
# feat(mongodb): add `$rerank` support to `MongoDBAtlasVectorSearch`

This PR adds MongoDB Atlas `$rerank` aggregation pipeline support to `MongoDBAtlasVectorSearch`, fixes flaky `MongoDBAtlasSemanticCache` integration tests, and updates the auto-embedding test infrastructure to use the `mongodb/mongodb-atlas-local:preview` Docker image.

---

Adds an optional `rerankOptions` constructor parameter that, when set, appends a `$rerank` stage after `$vectorSearch` to reorder results by semantic relevance using a MongoDB Atlas reranking model.

**New interface:**

```typescript
export interface MongoDBRerankOptions {
  model: string;               // e.g. "rerank-2.5", "rerank-2.5-lite", "rerank-2", "rerank-2-lite"
  path?: string | string[];    // field(s) to rerank on; defaults to textKey ("text")
}
```

**Usage:**

```typescript
const vectorStore = new MongoDBAtlasVectorSearch(embeddings, {
  collection,
  rerankOptions: {
    model: "rerank-2.5",
    path: "text",           // single field; omit to use textKey default
    // path: ["text", "title"],  // or multiple fields
  },
});
const results = await vectorStore.similaritySearchWithScore("query", 4);
// results[0][1] === rerankScore, results[0][0].metadata.relevanceScore === rerankScore
```

**Behaviour:**
- Works with both manual embeddings and auto-embedding mode.
- `$rerank` is inserted between `$vectorSearch` and the post-filter/limit stages.
- `numDocsToRerank` equals `k` — the reranker sees the same candidate set that is returned.
- The rerank score is returned as the numeric score in the `[Document, number]` tuple and also copied to `document.metadata.relevanceScore`.

**Test infrastructure:**
- New `vectorstores_rerank.int.test.ts` with four manual-embedding rerank tests and one combined auto-embed + rerank test.
- Rerank tests are skipped when `isUsingLocalAtlas()` is true (`$rerank` requires a real Atlas cluster, not the local Docker image).
- The combined auto-embed + rerank test additionally requires `VOYAGE_API_KEY`.
- Removed the earlier rerank tests that were incorrectly placed in `vectorstores_auto_embed.int.test.ts`.

> **Note:** `$rerank` is a preview feature and requires a MongoDB Atlas cluster enrolled in the preview program.

---

## fix(mongodb): fix flaky `MongoDBAtlasSemanticCache` integration tests

`waitForIndexToBeQueryable` returns as soon as the search index status is `READY`, but newly inserted documents are not immediately searchable even on a ready index. Tests that inserted a document and then immediately called `cache.lookup()` would intermittently return `null`.

Replaced all post-insert `waitForIndexToBeQueryable` calls with a new `waitForCacheIndexed` helper that polls `cache.lookup()` directly until the inserted entry is returned (up to 30 s).

---

## test(mongodb): update auto-embedding tests to use `mongodb/mongodb-atlas-local:preview`

The `mongodb/mongodb-atlas-local:latest` image does not support the `autoEmbed` index type. Updated `setup.ts` to use `mongodb/mongodb-atlas-local:preview`, pass through `VOYAGE_API_KEY` and `EMBEDDING_PROVIDER_ENDPOINT` environment variables, set `MONGODB_ATLAS_LOCAL_PREVIEW=true`, and increase the container startup timeout from 30 s to 120 s to allow the `mongot` process time to initialise.

### DIFF
--- a/libs/providers/langchain-mongodb/CHANGELOG.md
+++ b/libs/providers/langchain-mongodb/CHANGELOG.md
@@ -1,13 +1,5 @@
 # @langchain/mongodb
 
-## 1.3.0
-
-### Minor Changes
-
-- feat(mongodb): add `$rerank` aggregation pipeline support to `MongoDBAtlasVectorSearch`
-
-  Pass `rerankOptions: { model: string, path?: string | string[] }` to the constructor to enable MongoDB Atlas reranking. When set, `similaritySearchWithScore` appends a `$rerank` stage after `$vectorSearch`, returning results ordered by semantic relevance score. The rerank score is exposed as the numeric score in the returned tuple and as `metadata.relevanceScore` on each document. Works with both manual embeddings and auto-embedding mode.
-
 
 ### Patch Changes
 
--- a/libs/providers/langchain-mongodb/CHANGELOG.md
+++ b/libs/providers/langchain-mongodb/CHANGELOG.md
@@ -1,6 +1,5 @@
 # @langchain/mongodb
 
-
 ### Patch Changes
 
 - [#10872](https://github.com/langchain-ai/langchainjs/pull/10872) [`a640079`](https://github.com/langchain-ai/langchainjs/commit/a64007997a4940f51bba3c1c83dae89d1ccfb692) Thanks [@hntrl](https://github.com/hntrl)! - chore(deps): remove redundant @types/uuid declarations
--- a/libs/providers/langchain-mongodb/CHANGELOG.md
+++ b/libs/providers/langchain-mongodb/CHANGELOG.md
@@ -1,6 +1,13 @@
 # @langchain/mongodb
 
-## 1.2.1
+## 1.3.0
+
+### Minor Changes
+
+- feat(mongodb): add `$rerank` aggregation pipeline support to `MongoDBAtlasVectorSearch`
+
+  Pass `rerankOptions: { model: string, path?: string | string[] }` to the constructor to enable MongoDB Atlas reranking. When set, `similaritySearchWithScore` appends a `$rerank` stage after `$vectorSearch`, returning results ordered by semantic relevance score. The rerank score is exposed as the numeric score in the returned tuple and as `metadata.relevanceScore` on each document. Works with both manual embeddings and auto-embedding mode.
+
 
 ### Patch Changes
 
--- a/libs/providers/langchain-mongodb/src/tests/cache.int.test.ts
+++ b/libs/providers/langchain-mongodb/src/tests/cache.int.test.ts
@@ -4,6 +4,21 @@ import { Generation } from "@langchain/core/outputs";
 import { MongoDBCache, MongoDBAtlasSemanticCache } from "../cache.js";
 import { uri, waitForIndexToBeQueryable } from "./utils.js";
 
+async function waitForCacheIndexed(
+  cache: MongoDBAtlasSemanticCache,
+  prompt: string,
+  llmKey: string,
+  maxWaitMs = 30_000
+): Promise<void> {
+  const start = Date.now();
+  while (Date.now() - start < maxWaitMs) {
+    const result = await cache.lookup(prompt, llmKey);
+    if (result !== null) return;
+    await new Promise((resolve) => setTimeout(resolve, 2_000));
+  }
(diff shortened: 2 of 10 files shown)


caseyclements:
[Link to passing patch in our internal CI](https://spruce.corp.mongodb.com/version/6a21be881b02b700072d0958/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

```bash
evergreen patch -p ai-ml-pipeline-testing --param REPO_ORG=caseyclements --param REPO_NAME="langchainjs" --param REPO_BRANCH="mongodb-reranker" 
```

libs/providers/langchain-mongodb/src/tests/cache.int.test.ts
```diff
@@ -110,7 +123,7 @@ test("MongoDBAtlasSemanticCache: similar prompts share cache", async () => {
   const gen = [{ text: "Paris" }];
 
   await cache.update(prompt1, llmKey, gen);
-  await waitForIndexToBeQueryable(semanticCollection, "default");
+  await waitForCacheIndexed(cache, prompt1, llmKey);
```
hntrl, line 126: I know we're just modifying this line to, but would this be a good opportunity to just nix this line? I think its kind of a weird condition that we're waiting for an entire index to build in the hot path of creation. I'd think what we want to do is fail when we go to pull from the cache.

libs/providers/langchain-mongodb/src/tests/vectorstores_rerank.int.test.ts
```diff
@@ -0,0 +1,262 @@
...
+    documents: Document[],
+    options?: { ids?: string[] }
+  ): Promise<string[]> {
+    const ids = await super.addDocuments(documents, options);
+    const queryEmbedding = await this.embeddings.embedQuery("sandwich");
+    for (;;) {
+      const results = await this.similaritySearchVectorWithScore(queryEmbedding, documents.length);
+      if (results.length === documents.length) return ids;
+      await setTimeout(1000);
+    }
```
hntrl, line 55: It seems kinda weird that we have to patch over this behavior? I know these things will be eventually consistent but to me this reads as we're testing different vector store behavior from the actual vector store impl

libs/providers/langchain-mongodb/CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
...
 
-## 1.2.1
+## 1.3.0
+
+### Minor Changes
+
+- feat(mongodb): add `$rerank` aggregation pipeline support to `MongoDBAtlasVectorSearch`
+
+  Pass `rerankOptions: { model: string, path?: string | string[] }` to the constructor to enable MongoDB Atlas reranking. When set, `similaritySearchWithScore` appends a `$rerank` stage after `$vectorSearch`, returning results ordered by semantic relevance score. The rerank score is exposed as the numeric score in the returned tuple and as `metadata.relevanceScore` on each document. Works with both manual embeddings and auto-embedding mode.
+
```
hntrl:
```suggestion
```

Changesets will add this auto-magically

libs/providers/langchain-mongodb/src/tests/vectorstores_rerank.int.test.ts
caseyclements: Same as above. Put in place to stop tests from being flaky.
In a python version, I've added an a wait arg so that uses who know their setup can given it a well-chosen time.

libs/providers/langchain-mongodb/CHANGELOG.md
caseyclements: Oh strange. That came up as a workflow, or maybe a pre-commit. Reverted

hntrl (dismissed):

libs/providers/langchain-mongodb/src/tests/cache.int.test.ts
caseyclements: Indexes have eventually consistent, but take variable amounts of time. We commonly use these try/wait loops in testing to keep runs green. Do you have something else in mind?

hntrl (approved):

PR #10953, opened 2026-05-26
fix(aws): handle citation deltas in stream

### LINKED ISSUE #10942: [@langchain/aws] ChatBedrockConverse stream throws on PDF citation deltas (works on invoke)
## Summary

`ChatBedrockConverse#stream` (and `#streamEvents`) throws `Unsupported content block type(s)` when the input contains a PDF document block with `citations: { enabled: true }`. The non streaming `invoke()` path handles the same input correctly.

This breaks vision based PDF reading on any streaming chat surface. Without citations enabled, Bedrock falls back to its server side text extraction, which is unreliable on scans, handwriting, and photo PDFs. With citations enabled, the stream parser crashes on the first citation delta.

## Versions

- `@langchain/aws`: 1.3.9 (latest published)
- Verified the same bug on 1.3.4
- Models tested: `us.anthropic.claude-sonnet-4-6`, `us.anthropic.claude-opus-4-6-v1` (bug is model agnostic, it lives in the stream parser)
- Node 22, macOS 25

## Where it breaks

`@langchain/aws@1.3.9` `dist/utils/message_outputs.js` lines 91 to 114, `handleConverseStreamContentBlockDelta`:

```js
function handleConverseStreamContentBlockDelta(contentBlockDelta) {
  if (!contentBlockDelta.delta) throw new Error("No delta found in content block.");
  if (typeof contentBlockDelta.delta.text === "string") { ... }
  else if (contentBlockDelta.delta.toolUse) { ... }
  else if (contentBlockDelta.delta.reasoningContent) { ... }
  else throw new Error(`Unsupported content block type(s): ${JSON.stringify(contentBlockDelta.delta, null, 2)}`);
}
```

There is no branch for citation deltas. Bedrock Converse emits these as part of the streaming response whenever a document block has citations enabled.

The non streaming converter at the top of the same file already handles citations (lines 43 to 45 in 1.3.9):

```js
else if ("citationsContent" in c) content.push({
  type: "citations_content",
  citationsContent: c.citationsContent
});
```

So the library already knows about the type, the streaming branch just was not updated alongside it.

## Reproduction

```ts
import { ChatBedrockConverse } from '@langchain/aws'
import { HumanMessage } from '@langchain/core/messages'
import fs from 'node:fs/promises'

const model = new ChatBedrockConverse({
  model: 'us.anthropic.claude-sonnet-4-6',
  region: 'us-east-1',
  temperature: 0,
})

const bytes = await fs.readFile('./scanned.pdf')

  content: [
    { type: 'text', text: 'What does this document say?' },
    {
      type: 'document',
      document: {
        name: 'scanned',
        format: 'pdf',
        source: { bytes },
        citations: { enabled: true },
      },
    },
  ],
})

const stream = await model.stream([message])
  console.log(chunk)
}
```

This throws on the first citation delta with `Unsupported content block type(s): { "citation": { ... } }`.

The exact same input via `model.invoke([message])` succeeds and returns a message containing `citations_content` blocks.

## Impact

Users wanting vision quality reads of PDFs on a streaming chat agent are forced into one of two workarounds:

1. Disable citations on the stream path and accept Bedrock's text extraction fallback, which loses the entire reason for using citations in the first place.
2. Route every attachment turn through a separate non streaming tool call, then have the agent re emit the result as a normal streamed reply. Works, but adds an avoidable hop.

For context, we ended up doing option 2 in production. Main chat agent streams Sonnet without citations, and a dedicated `read_attachment` tool fires Opus via `invoke()` with citations enabled. It would be nice to collapse this back to one path.

## Suggested fix

Add a citation delta branch to `handleConverseStreamContentBlockDelta` mirroring the non streaming converter. Rough shape:

```js
else if (contentBlockDelta.delta.citation) {
  return new ChatGenerationChunk({
    text: "",
    message: new AIMessageChunk({
      content: [{
        type: "citations_content",
        citationsContent: { citations: [contentBlockDelta.delta.citation] },
      }],
    }),
  });
}
```

Happy to open a PR if a maintainer can confirm the desired output shape for streamed citation chunks (single citation per chunk vs accumulated, where to attach on `AIMessageChunk`, etc).

akintunero:
Fix streaming citation deltas for ChatBedrockConverse by adding citation handling in the stream delta converter. This mirrors the non‑streaming citations_content shape and prevents Unsupported content block type errors when PDFs with citations are streamed. Added a unit test for citation deltas.

Fixes #10942

### DIFF
--- a/libs/langchain-aws/src/common.ts
+++ b/libs/langchain-aws/src/common.ts
@@ -789,6 +789,23 @@ export function handleConverseStreamContentBlockDelta(
         ],
       }),
     });
+  } else if (
+    "citation" in contentBlockDelta.delta &&
+    contentBlockDelta.delta.citation
+  ) {
+    return new ChatGenerationChunk({
+      text: "",
+      message: new AIMessageChunk({
+        content: [
+          {
+            type: "citations_content",
+            citationsContent: {
+              citations: [contentBlockDelta.delta.citation],
+            },
+          },
+        ],
+      }),
+    });
   } else {
     throw new Error(
       `Unsupported content block type(s): ${JSON.stringify(
--- a/libs/langchain-aws/src/tests/chat_models.test.ts
+++ b/libs/langchain-aws/src/tests/chat_models.test.ts
@@ -11,6 +11,7 @@ import {
   ConversationRole as BedrockConversationRole,
   type Message as BedrockMessage,
   type SystemContentBlock as BedrockSystemContentBlock,
+  type ContentBlockDeltaEvent,
 } from "@aws-sdk/client-bedrock-runtime";
 import { z } from "zod";
 import { describe, expect, test } from "@jest/globals";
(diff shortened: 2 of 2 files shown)


lequocbinh04:
same problem ._. waiting for the merge

PR #10973, opened 2026-06-01
fix: harden LocalFileStore path boundary

dfgvaetyj3456356-hash:
## Description

This PR hardens `LocalFileStore` path containment in both `langchain` and `@langchain/classic`.

The existing guard resolved the requested key path and checked `fullPath.startsWith(commonPath)`. That rejects many traversal attempts, but string-prefix containment can still accept a sibling directory whose absolute path shares the root path as a prefix, such as `<root>` and `<root>-sibling`.

The fix uses `path.relative(commonPath, fullPath)` and rejects paths that resolve outside the configured root (`..`) or become absolute. This preserves valid nested keys while avoiding prefix-based boundary confusion.

## Tests

```text
pnpm install --frozen-lockfile
pnpm --filter @langchain/core build
pnpm --filter langchain test src/storage/tests/file_system.test.ts
pnpm --filter @langchain/classic test src/storage/tests/file_system.test.ts
git diff --check
```

Both focused storage test suites pass locally: 9 tests in `langchain` and 9 tests in `@langchain/classic`.


dfgvaetyj3456356-hash:
Follow-up: I found the same boundary class in the older @langchain/classic NodeFileStore used by the file tools. Pushed fd9e0991e to resolve file paths against the configured basePath and reject absolute/parent/sibling-prefix escapes before readFile/writeFile. Added focused NodeFileStore coverage for normal read/write, direct traversal, and sibling-prefix traversal. Validation: pnpm exec oxfmt --write on the touched files, pnpm exec oxlint on the touched files, pnpm --filter @langchain/classic test src/storage/tests/file_system.test.ts src/stores/file/tests/node.test.ts, and git diff --check all pass.

PR #11001, opened 2026-06-02
fix(openai): wrap responses stream errors

### LINKED ISSUE #10979: Responses API: streaming errors bypass wrapOpenAIClientError in _streamResponseChunks
### Description

When using `ChatOpenAI` with `useResponsesApi: true` and `streaming: true`, errors thrown mid-stream (SSE error events) bypass `wrapOpenAIClientError()`. This means `ContextOverflowError` (and other error transformations) are never applied for streaming Responses API calls.

### Root Cause

In `responses.ts`, `_streamResponseChunks()` iterates the stream without a try/catch:

```typescript
// responses.ts — _streamResponseChunks()
const streamIterable = await this.completionWithRetry(
  { ...params, stream: true },
  options
);

for await (const data of streamIterable) {   // ← errors thrown here escape unwrapped
  const chunk = convertResponsesDeltaToChatGenerationChunk(data);
  if (chunk == null) continue;
  yield chunk;
}
```

`completionWithRetry()` wraps the `responses.create()` call, but for streaming that only covers the initial HTTP request — not the async iteration. When an SSE error event fires mid-stream, the OpenAI SDK throws an `APIError` from within the async iterator, and it propagates directly to the caller without passing through `wrapOpenAIClientError()`.

### How this manifests with `previous_response_id`

When using `previous_response_id`, the full context is reconstructed server-side after the stream opens (HTTP 200). If the reconstructed context exceeds the model's limit, the error arrives as an SSE error event — not an HTTP 400 before the stream.

The OpenAI SDK's `streaming.ts` handles this:
```typescript
// openai SDK — streaming.ts
if (data && data.error) {
  throw new APIError(undefined, data.error, undefined, response.headers);
}
```

This produces a plain `APIError` (not `BadRequestError`) with `status: undefined` and `code: "context_length_exceeded"`. Because the error is thrown during iteration (not during `.create()`), it never reaches `wrapOpenAIClientError()` → `_isOpenAIContextOverflowError()` → `ContextOverflowError`.

### Impact

Any error handling that depends on `ContextOverflowError` (e.g., automatic context compaction) silently fails when using the Responses API with streaming and `previous_response_id`. The error surfaces as a raw `APIError` instead of the expected `ContextOverflowError`.

Other error transformations in `wrapOpenAIClientError()` are similarly bypassed:
- Timeout → `TimeoutError`
- Abort → `AbortError`
- 401 → `MODEL_AUTHENTICATION`
- 429 → `MODEL_RATE_LIMIT`
- 404 → `MODEL_NOT_FOUND`

### Reproduction

```typescript

  model: "gpt-4o",
  useResponsesApi: true,
  streaming: true,
});

try {
  // Use a previous_response_id that would cause context overflow
  // when combined with new messages
  const result = await model.invoke(messages, {
    previous_response_id: "<id-of-large-prior-conversation>",
  });
} catch (e) {
  // Expected: e instanceof ContextOverflowError
  // Actual:   e instanceof APIError (raw, unwrapped)
  console.log(e.constructor.name); // "APIError", not "ContextOverflowError"
  console.log(e.code);             // "context_length_exceeded"
  console.log(e.status);           // undefined (not 400)
}
```

### Suggested Fix

Wrap the stream iteration in `_streamResponseChunks()` with a try/catch that applies `wrapOpenAIClientError`:

```typescript
async *_streamResponseChunks(messages, options, runManager) {
  const streamIterable = await this.completionWithRetry(
    { ...params, stream: true },
    options
  );

  try {
    for await (const data of streamIterable) {
      if (options.signal?.aborted) return;
      const chunk = convertResponsesDeltaToChatGenerationChunk(data);
      if (chunk == null) continue;
      yield chunk;
      await runManager?.handleLLMNewToken(/* ... */);
    }
  } catch (e) {
    throw wrapOpenAIClientError(e);
  }
}
```

The same pattern should also be applied in `_generate()` where it calls `_streamResponseChunks()` via `for await`.

### Environment

- `@langchain/openai`: 0.5.x+ (any version with Responses API support)
- `openai` SDK: 4.x+
- Node.js: 20+

### Additional Context

`_isOpenAIContextOverflowError()` in `utils/client.ts` already handles both "context_length_exceeded" code and "exceeds the context window" message — the detection logic is correct. The issue is purely that this detection is never reached for streaming errors because the wrapping function is not called on the streaming iteration path.

Herrtian:
Fixes #10979

Wraps errors thrown while iterating Responses API streams with the existing OpenAI client error wrapper, so mid-stream errors are converted consistently.

Tests:
- `corepack.cmd pnpm --dir libs\providers\langchain-openai exec vitest run src/chat_models/tests/responses.test.ts`
- `corepack.cmd pnpm --dir libs\providers\langchain-openai exec oxfmt --check src/chat_models/responses.ts src/chat_models/tests/responses.test.ts`

### DIFF
--- a/.changeset/responses-stream-errors.md
+++ b/.changeset/responses-stream-errors.md
@@ -1,0 +1,5 @@
+---
+"@langchain/openai": patch
+---
+
+Wrap Responses API stream iteration errors with existing OpenAI client error handling.
--- a/libs/providers/langchain-openai/src/chat_models/responses.ts
+++ b/libs/providers/langchain-openai/src/chat_models/responses.ts
@@ -248,24 +248,29 @@ export class ChatOpenAIResponses<
       options
     );
 
-    for await (const data of streamIterable) {
-      if (options.signal?.aborted) {
-        return;
+    try {
+      for await (const data of streamIterable) {
+        if (options.signal?.aborted) {
+          return;
+        }
+        const chunk = convertResponsesDeltaToChatGenerationChunk(data);
+        if (chunk == null) continue;
+        yield chunk;
+        await runManager?.handleLLMNewToken(
+          chunk.text || "",
+          {
+            prompt: options.promptIndex ?? 0,
+            completion: 0,
+          },
+          undefined,
+          undefined,
+          undefined,
+          { chunk }
+        );
       }
-      const chunk = convertResponsesDeltaToChatGenerationChunk(data);
-      if (chunk == null) continue;
-      yield chunk;
-      await runManager?.handleLLMNewToken(
-        chunk.text || "",
-        {
-          prompt: options.promptIndex ?? 0,
-          completion: 0,
-        },
-        undefined,
-        undefined,
-        undefined,
-        { chunk }
-      );
+    } catch (e) {
+      const error = wrapOpenAIClientError(e);
+      throw error;
     }
   }
 
--- a/libs/providers/langchain-openai/src/chat_models/tests/responses.test.ts
+++ b/libs/providers/langchain-openai/src/chat_models/tests/responses.test.ts
@@ -87,7 +87,7 @@ describe("streaming errors", () => {
     };
 
     async function* streamWithError() {
-      throw originalError;
+      yield await Promise.reject(originalError);
     }
 
     model.completionWithRetry = vi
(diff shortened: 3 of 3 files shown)


christian-bromann (dismissed):
LGTM 👍 

Thanks for the contribution!

christian-bromann (approved):

PR #11015, opened 2026-06-02
fix(langchain): prevent null jumpTo writes in middleware

hntrl:
## Summary

Addresses the upstream agent middleware bug reported in `langchain-ai/deepagentsjs#480` where concurrent writes to the `jumpTo` channel could trigger:

`InvalidUpdateError: Invalid update for channel "jumpTo" with values [null,null]`

This change makes middleware node outputs sparse and only emits `jumpTo` when it is explicitly set to a valid string target, preventing nullish `jumpTo` writes during structured-output retry flows with multiple `afterModel` middleware hooks.

### `langchain` / agents middleware node

- Updated `MiddlewareNode.invokeMiddleware` to avoid emitting `jumpTo` for no-op middleware results.
- Changed control-action and normal middleware update paths to strip `jumpTo` from intermediate updates and only include it when present as a valid string target.
- Kept existing jump target validation behavior (`canJumpTo`) for explicit jump targets.
- Preserved terminate control-action behavior while ensuring sparse state updates.

### `langchain` / agent middleware tests

- Added regression coverage for structured-output retry with two no-op `afterModel` middlewares to ensure no concurrent `jumpTo` update error.
- Added a routing regression ensuring explicit `afterModel` `jumpTo` still works with another no-op `afterModel` middleware.
- Added a direct middleware node test confirming `undefined` `afterModel` results produce an empty update with no `jumpTo` key.

### DIFF
--- a/.changeset/mean-radios-hope.md
+++ b/.changeset/mean-radios-hope.md
@@ -1,0 +1,7 @@
+---
+"langchain": patch
+---
+
+fix(agents): avoid nullish jumpTo updates from middleware no-op returns
+
+Middleware hook outputs are now treated as sparse updates so no-op hooks do not emit `jumpTo` writes. This prevents concurrent graph update errors in structured output retry flows with multiple `afterModel` middleware hooks while preserving explicit jump routing behavior.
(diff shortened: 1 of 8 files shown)


javier-quincke-sytrex:
any updates on this bug resolution? I'm having the same problem

PR #11026, opened 2026-06-06
feat(langchain): add providerToolSearchMiddleware

aolsenjazz:
## Description

Adds `providerToolSearchMiddleware` - provider-side tool search for agents. Enables API consumers to opt tools into tool deferral + discovery by providing tool instances/names to the middleware's `searchableTools` arg.

## Changes

**`langchain` agent middleware:**
- Add `providerToolSearchMiddleware` at `src/agents/middleware/providerToolSearch.ts` Opt-in deferral via `searchableTools`
- Inject the provider-native search tool per model call
- Gate on provider only (Anthropic/OpenAI); defer model-version eligibility to the provider's own API

**`langchain` tests:**
- Add `src/agents/middleware/tests/providerToolSearch.test.ts` covering all `wrapModelCall` code paths
- Add `src/agents/middleware/tests/providerToolSearch.int.test.ts` with happy-path tests for OpenAI/Anthropic as well as check re. OpenAI models < 5.4

## Callouts
- Omitted changeset as it looks largely unused in this repo
- We only guard against improper provider selection. This will still fail at the provider API level if an unsupported model is chosen (see OpenAI integration test)
- `createMockModel` in the unit test suite is ugly, but followed precedent established by `llmToolSelector` unit tests
- Opted for `searchableTools` as it's more both more inline with the concepts in use here and more future proof than the initially proposes `include`/`includeTools`. Also open to a `deferredTools` nomenclature.


libs/langchain/src/agents/middleware/providerToolSearch.ts
```diff
@@ -0,0 +1,201 @@
...
+          );
+        }
+      }
+
+      const provider = getModelProvider(request.model);
+      if (!supportsProviderToolSearch(provider)) {
+        throw new Error(
+          `providerToolSearchMiddleware requires a provider with server-side tool search, but got ${provider}`
+        );
+      }
```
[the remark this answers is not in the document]
aolsenjazz: This is the intent. If an API consumer tries to use this middleware with an incompatible model provider, it should fail fast.

libs/langchain/src/agents/middleware/providerToolSearch.ts
```diff
@@ -0,0 +1,201 @@
...
+      // Warn if we try to defer a tool that is not bound to the model
+      if (deferNames.size > 0) {
+        const available = tools.filter(isLangChainTool).map((t) => t.name);
+        const unknown = [...deferNames].filter(
+          (name) => !available.includes(name)
+        );
+        if (unknown.length > 0) {
+          console.warn(
+            `ProviderToolSearchMiddleware: searchableTools references tool(s) not found: ${unknown.join(", ")}. They will not be deferred`
+          );
```
hntrl: we should fail fast here IMO

libs/langchain/src/agents/middleware/tests/providerToolSearch.int.test.ts
```diff
@@ -0,0 +1,100 @@
...
+      ],
+    });
+
+    const result = await agent.invoke({
+      messages: [
+        new HumanMessage("What is the delivery status of order A1234?"),
+      ],
+    });
+
+    expect(calledTool(result.messages, "lookup_order_status")).toBe(true);
```
hntrl, line 78: do we want to have some kind of condition check that the tool search tool was actually called?

.changeset/chilly-spies-ring.md
```diff
@@ -0,0 +1,7 @@
+---
+"langchain": minor
```
hntrl:
```suggestion
"langchain": patch
```

libs/langchain/src/agents/middleware/tests/providerToolSearch.int.test.ts
aolsenjazz: Didn't think necessary - but will add it in

libs/langchain/src/agents/middleware/providerToolSearch.ts
aolsenjazz: Addressed in 2cdba104756ce6bc1570bdf333b654d52ff04a0f

libs/langchain/src/agents/middleware/tests/providerToolSearch.int.test.ts
aolsenjazz: Addressed in ad2818e842735a394efd3c08f416873c99fd0ce0

The function to detect tool usage is kind of ugly since it's provider-specific, but imo it's fine as a util function for an int test module

hntrl (approved):
looks good!

PR #11029, opened 2026-06-06
fix(agents): prevent unhandled rejection on HITL interrupts in createToolCallTransformer

### LINKED ISSUE #10933: Unhandled promise rejection on HITL interrupts in `createToolCallTransformer`
### Checked other resources

### Example Code

```ts
const agent = createAgent({
  model,
  tools: [/* a tool wired with HumanInTheLoop / interruptOn */],
  checkpointer,
});

const stream = await agent.streamEvents(
  { messages: [{ role: "user", content: "do the thing" }] },
  { version: "v3", configurable: { thread_id: "t-1" } },
);

for await (const message of stream.messages) {
  for await (const _ of message.text) {
    // consume output
  }
}
```

Crash output:

```text
node_modules/langchain/src/agents/stream.ts:291
              pending.rejectOutput(new Error(message));
                                   ^

Error: [
  {
    "id": "...",
    "value": {
      "actionRequests": [
        {
          "name": "...",
          "args": { ... },
          "description": "... requires manual approval"
        }
      ],
      "reviewConfigs": [
        {
          "actionName": "...",
          "allowedDecisions": ["approve", "reject"]
        }
      ]
    }
  }
]
    at Object.process (langchain/src/agents/stream.ts:291)
    at StreamMux.push (langgraph/src/stream/mux.ts:287)
    at pump (langgraph/src/stream/mux.ts:499)
```

The stack trace never enters user code. The process dies because Node reports the promise as unhandled before any consumer can attach a handler.

_No response_

## Summary

When an agent built with `createAgent` (or `createDeepAgent`) triggers a Human-In-The-Loop interrupt while consuming `streamEvents({ version: "v3" })`, the tool-call `output` promise created by `createToolCallTransformer` is rejected before any user-space consumer has had a chance to attach a handler. This produces a fatal `unhandledRejection` that crashes the Node.js process, even when the consumer correctly handles the interrupt via `stream.interrupts`.

## Root cause

In `createToolCallTransformer` (`langchain/dist/agents/stream.js`), each tool call entry creates an `output` promise and stores its `rejectOutput`:

```js
const output = new Promise((res, rej) => {
  resolveOutput = res;
  rejectOutput = rej;
});
// ...
toolCallsLog.push({ name, callId, input, output, status, error });
```

When a `tool-error` event arrives, the transformer rejects that promise:

```js
} else if (data.event === "tool-error") {
  const message = data.message ?? "unknown error";
  if (isHeadlessToolInterruptError(message, toolCallId)) {
    return true;
  }
  pending.rejectOutput(new Error(message));
  // ...
}
```

For HITL interrupts, the `message` is the JSON-serialized interrupt payload (`actionRequests` / `reviewConfigs`). `isHeadlessToolInterruptError` only short-circuits a specific "headless" shape; regular HITL interrupts fall through and rejection happens.

The pump in `@langchain/langgraph/dist/stream/mux.js` invokes `process()` synchronously in a tight loop. The sequence on the microtask queue is:

1. `tool-start` event → `createToolCallEntry` → `toolCallsLog.push(...)` resolves the channel's pending `next()`.
2. Same synchronous tick: `tool-error` event → `pending.rejectOutput(new Error(message))`. The `output` promise is now rejected with no handler attached.
3. Node schedules the `unhandledRejection` check.
4. Only later, on the next microtask, does the consumer's `for await` body resume and have a chance to attach `.catch(...)`.

Because steps 1 and 2 happen synchronously within a single `pump` iteration, no user-space consumer can win the race to attach a handler before Node observes the unhandled rejection.

## Why this cannot be handled reliably in user space

- `createToolCallTransformer` is hardcoded as the first transformer in `ReactAgent`, so consumers cannot replace or wrap it.
- Attaching `toolCall.output.catch(() => {})` inside `for await (const toolCall of stream.toolCalls)` is too late in the failing case, because the rejection may already have happened in the same synchronous pump cycle.
- The only reliable workaround available today is a global `process.on("unhandledRejection", ...)` handler that detects this specific interrupt payload and swallows it.

## Current workaround

```ts
process.on("unhandledRejection", (reason) => {
  if (isLangChainInterruptRejection(reason)) return;
  throw reason;
});
```

This avoids the crash, but it is a process-wide workaround for a library-level race condition.

### System Info

- `langchain` — file `langchain/dist/agents/stream.js` (also `stream.cjs`), function `createToolCallTransformer`
- Reproduced with `@langchain/langgraph` stream mux (`@langchain/langgraph/dist/stream/mux.js`)
- Node.js 20+ (default unhandled-rejection = throw)

chatman-media:
Closes #10933

## Root cause

In `createToolCallTransformer` (`langchain/src/agents/stream.ts`), each tool call entry creates an `output` promise and stores its `rejectOutput` resolver. When a `tool-error` event arrives, the transformer has a fast-path for **headless** tool interrupts via `isHeadlessToolInterruptError` — those return early without rejecting the promise. However, **HITL** interrupts from `humanInTheLoopMiddleware` carry a structurally different payload:

```json
[{ "id": "…", "value": { "actionRequests": […], "reviewConfigs": […] } }]
```

`isHeadlessToolInterruptError` only matches the headless shape (`value.type === "tool"`), so HITL interrupts fell through to `pending.rejectOutput(new Error(message))`. Because the langgraph `pump` fires `tool-started` and `tool-error` synchronously in the same tick, the rejection happened before any user-space consumer could attach a `.catch()` handler, producing a fatal `unhandledRejection` crash.

## Fix

Add `isHITLInterruptError` that detects interrupt arrays whose serialised `value` contains an `actionRequests` array. When matched, return `true` early (keeping the tool call promise pending), exactly as headless interrupts are handled.

## Test

Added a unit test in `libs/langchain/src/agents/tests/stream.test.ts` that:
1. Fires a `tool-started` event for a HITL-gated tool.
2. Immediately fires the `tool-error` event with a HITL interrupt payload (before any consumer attaches a handler — the exact race from the bug).
3. Asserts the `output` promise is still **pending** (not rejected) using `Promise.race` with an immediately-resolved sentinel.
4. Asserts that `finalize()` then cleanly resolves the promise.

The test fails on the unpatched code (the `Promise.race` would throw because the `output` promise is already rejected) and passes with the fix.

### DIFF
--- a/.changeset/fix-hitl-unhandled-rejection.md
+++ b/.changeset/fix-hitl-unhandled-rejection.md
@@ -1,0 +1,7 @@
+---
+"langchain": patch
+---
+
+fix(agents): prevent unhandled promise rejection on HITL interrupts in `createToolCallTransformer`
+
+When a Human-In-The-Loop interrupt arrives via a `tool-error` event, the `output` promise for the affected tool call was rejected before any consumer could attach a `.catch()` handler, causing a fatal `unhandledRejection` crash. The fix adds detection for HITL interrupt payloads (those whose serialised value contains an `actionRequests` array) and skips the rejection, keeping the tool call pending — consistent with the existing handling for headless tool interrupts.
--- a/libs/langchain/src/agents/stream.ts
+++ b/libs/langchain/src/agents/stream.ts
@@ -167,6 +167,27 @@ function isHeadlessToolInterruptError(
 }
 
 /**
+ * Returns true when `message` is a serialised LangGraph interrupt array
+ * containing at least one HITL interrupt (i.e. an entry whose `value` has
+ * an `actionRequests` array).  These interrupts must NOT be surfaced as
+ * rejected output promises — the tool call is merely paused, not errored.
+ */
+function isHITLInterruptError(message: string): boolean {
+  try {
+    const parsed = JSON.parse(message) as unknown;
+    if (!Array.isArray(parsed)) return false;
+    return parsed.some((entry) => {
+      if (entry == null || typeof entry !== "object") return false;
+      const value = (entry as { value?: unknown }).value;
+      if (value == null || typeof value !== "object") return false;
+      return Array.isArray((value as { actionRequests?: unknown }).actionRequests);
+    });
+  } catch {
+    return false;
+  }
+}
+
+/**
  * Detects serialized LangChain `ToolMessage` values that can appear on
  * `tool-finished.output` after crossing a protocol or serialization boundary.
  *
@@ -322,6 +343,9 @@ export function createToolCallTransformer(
                 ((data as Record<string, unknown>).message as string) ??
                 "unknown error";
               if (isHeadlessToolInterruptError(message, toolCallId)) {
+                return true;
+              }
+              if (isHITLInterruptError(message)) {
                 return true;
               }
               pending.rejectOutput(new Error(message));
--- a/libs/langchain/src/agents/tests/stream.test.ts
+++ b/libs/langchain/src/agents/tests/stream.test.ts
@@ -510,6 +510,88 @@ describe("streamEvents", () => {
     expect(await iterator.next()).toEqual({ done: true, value: undefined });
   });
 
+  it("keeps tool call output pending (not rejected) on a HITL interrupt tool-error (issue #10933)", async () => {
+    // Regression test for: unhandled promise rejection when a HITL interrupt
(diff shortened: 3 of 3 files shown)


chatman-media:
Superseded by #11087, which fixes this more generally via `isToolInterrupt()` (covers HITL + raw interrupts). Closing — thanks!

PR #11032, opened 2026-06-07
fix(google): ai parameter validation for "eu" multi-region

neuralmanacle:
Closes langchain-ai/langchainjs#11019

This PR updates the connection logic to correctly resolve the host endpoint for the `eu` region on Google Cloud Platform, ensuring compliance with GCP's regional endpoint requirements.

**Changes:**

* `connection.ts`: Modified the `computedEndpoint` getter to explicitly return `aiplatform.eu.rep.googleapis.com` when the location is set to `eu`.
* `chat_models.test.ts`: Added unit tests to validate correct URL construction for the `eu` location, covering both the default rep endpoint and custom endpoint overrides.

PR #11036, opened 2026-06-08
fix(google): handle EU location for Vertex AI endpoint routing

### LINKED ISSUE #11019: ChatGoogle URL generation 404s on 'eu' location, and custom 'aiplatform.eu.rep.googleapis.com' endpoint yields empty extractions with gemini-3.1-flash-lite
### Checked other resources

### Example Code

```
import {ChatGoogle} from "@langchain/google/node";

// First try
    model: "gemini-3.1-flash-lite",
    location: "eu", // Triggers an invalid base URL: eu-aiplatform.googleapis.com
    maxRetries: 0,
    googleAuthOptions: {},
});
await model.invoke("Hello"); 
// Result: HTTP 404 Not Found

// Second try
    model: "gemini-3.1-flash-lite",
    location: "eu",
    endpoint: "aiplatform.eu.rep.googleapis.com",
   googleAuthOptions: {},
    maxRetries: 0,
});
const response = await model.invoke("Extract the key data points from this document...");
// Result: Downstream parsers evaluate response.content/text as empty
```

**for HTTP 404 Not Found:** "RequestError: Publisher Model `projects/loop-dev-gemini/locations/eu/publishers/google/models/gemini-3.1-flash-lite` was not found or your project does not have access to it. Please ensure you are using a valid model version. For more information, see: https://cloud.google.com/vertex-ai/generative-ai/docs/learn/model-versions\n    at RequestError.fromResponse"

**for Downstream parsers evaluate response.content/text as empty:** "stack":"Error: Empty extraction from Google LLM\n"

### Description

When configuring ChatGoogle for EU data using gemini-3.1-flash-lite, two sequential routing and parsing bugs occur:

Routing Bug (HTTP 404): Setting location: "eu" without a custom endpoint generates an invalid base URL, triggering a 404 error because "eu" isn't handled as a standard region.

Parsing Bug (Empty Extraction): Overriding the endpoint to Google's EU representative gateway (aiplatform.eu.rep.googleapis.com) successfully yields an HTTP 200. However, downstream parsing fails with an empty text extraction error.

Has this issue happened to anyone else?

### System Info

langchain@1.4.4 | MIT | deps: 4 | versions: 445
mac
node version: v22.22.0
pnpm version: v11.5.1

anishesg:
This PR fixes ChatGoogle's handling of the EU location for Vertex AI endpoint routing. When users set `location: "eu"` to ensure EU data residency, the code was incorrectly constructing the endpoint URL as `eu-aiplatform.googleapis.com`, which resulted in 404 errors because "eu" is not a valid Google Cloud region identifier.

The fix introduces two changes in `BaseChatGoogle`:

1. When `location` is set to "eu", the endpoint URL now correctly uses Google's EU representative endpoint `aiplatform.eu.rep.googleapis.com` instead of constructing an invalid region-prefixed URL.

2. A new `apiLocation` getter maps "eu" to the valid EU region "europe-west1" for use in API request paths, ensuring that the location parameter in the URL path is a valid region identifier while preserving the EU representative endpoint.

These changes allow users to simply set `location: "eu"` without needing to manually override the endpoint parameter, and ensure that API requests are properly routed through Google's EU data residency infrastructure.

Fixes #11019

### DIFF
--- a/.changeset/eu-vertex-endpoint-fix.md
+++ b/.changeset/eu-vertex-endpoint-fix.md
@@ -1,0 +1,5 @@
+---
+"@langchain/google": patch
+---
+
+fix(google): handle EU location for Vertex AI endpoint routing
--- a/libs/providers/langchain-google/src/chat_models/base.ts
+++ b/libs/providers/langchain-google/src/chat_models/base.ts
@@ -243,6 +243,21 @@ export abstract class BaseChatGoogle<
     return this._location || "global";
   }
 
+  /**
+   * Returns the effective location for use in API paths.
+   * Maps special location values like "eu" to valid region identifiers.
+   */
+  protected get apiLocation(): string {
+    const loc = this._location || "global";
+    // Map "eu" to a valid EU region for API calls
+    // When users specify "eu", they want EU data residency,
+    // which requires using a valid EU region in the API path
+    if (loc === "eu") {
+      return "europe-west1";
+    }
+    return loc;
+  }
+
   protected get endpoint(): string {
     if (typeof this._endpoint !== "undefined") {
       return this._endpoint;
@@ -253,6 +268,9 @@ export abstract class BaseChatGoogle<
     } else if (this.location === "global") {
       // See https://docs.cloud.google.com/vertex-ai/generative-ai/docs/learn/locations#use_the_global_endpoint
       return "aiplatform.googleapis.com";
+    } else if (this._location === "eu") {
+      // Special case for EU data residency - use the EU representative endpoint
+      return "aiplatform.eu.rep.googleapis.com";
     } else {
       return `${this.location}-aiplatform.googleapis.com`;
     }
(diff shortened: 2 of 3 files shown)


anishesg:
Yeah, this is a fix so I should add a changeset. Let me do that now.

anishesg:
done, added the changeset

afirstenberg:
Good catch for this.
However, this issue is also true for the "us" region.

I'd suggest changing the special case to either test if there isn't a dash (and isn't "global") to use the `aiplatform.{region}` pattern or to special case both regions.

I don't have a preference. Whatever we choose, Google will break it without notice again in the future.

neuralmanacle:
@afirstenberg @anishesg the heuristic logic related to checking for "-" has been added along with unit tests for the multi-region as a new commit on PR #11032. I'm ready to align with whatever merge path makes the most sense for the repository—just let me know your thoughts.

hntrl (commented):
Fixing with #11032

hntrl:
Fixing with #11032

PR #11037, opened 2026-06-09
feat(mcp-adapters): expose MCP tool outputSchema on tool metadata

carefreelife98:
## What

`loadMcpTools` forwards an MCP tool's `inputSchema` (as the LangChain tool `schema`) and `annotations`, but silently drops the spec-defined optional `outputSchema`. The MCP `Tool` shape is `{ name, description?, inputSchema, outputSchema?, annotations? }`.

This exposes `tool.outputSchema` on the resulting `DynamicStructuredTool` metadata when present, mirroring how `annotations` is already surfaced. The key is omitted entirely when the tool declares no output schema, so a simple presence check works.

## Why

Consumers that need an MCP tool's structured-output type currently have to make a second raw `client.listTools()` call to recover `outputSchema`, since the adapter discards it. After this change it's available directly via `tool.metadata.outputSchema`.

## Design notes

**Why expose the raw schema (unlike `inputSchema`, which is dereferenced + simplified)?**
`inputSchema` is transformed via `simplifyJsonSchemaForLLM` because it is *sent to the LLM* and must satisfy provider function-calling constraints (e.g. OpenAI rejecting `anyOf`/`oneOf`). `outputSchema` is never sent to the LLM — it is a declarative type contract for consumers. Lossy simplification would discard schema information consumers rely on, so the server's original schema is preserved as-is.

**Why not just rely on the existing `mcp_structured_content` artifact?**
The adapter already surfaces `result.structuredContent` as an artifact, but that is the *runtime payload* available only *after* a call. `outputSchema` is the *static contract* available *before* invocation — useful for deciding downstream node schemas or describing the expected output shape ahead of time. They serve different needs.

## Notes

- `DynamicStructuredTool.metadata` is t `Record<string, unknown>`, so no cast is needed.
- `Tool.outputSchema` is a first-class optional field in `@modelcontextprotocol/sdk`, so reading it is type-safe.
- Adds a unit test covering both the present and absent cases.
- Verified locally: `oxfmt --check`, `oxlint`, and `vitest run` (16/16) all pass on the changed files.

No existing issue — small additive change (searched issues/PRs for `outputSchema`, none found).


carefreelife98:
Friendly ping 🙂 Opened #11083 describing the gap this addresses.

This is a small additive change to `@langchain/mcp-adapters` — it forwards the MCP spec's optional `outputSchema` onto tool metadata, mirroring how `annotations` is already exposed. CI is green, a changeset is included, and the key is omitted entirely when a tool declares no output schema.

@hntrl would you (or another maintainer) be able to take a look, or let me know if you'd prefer a different approach? Happy to adjust.

PR #11045, opened 2026-06-10
fix(openai): omit empty reasoning item id in Responses API input

jackjin1997:
## Summary

Fixes #11044.

Reasoning blocks reassembled from streaming chunks (e.g. via `agent.streamEvents()`) never carry an `id` — OpenAI's streaming protocol only includes the reasoning item `id` in non-streaming responses. When such a message is replayed as Responses API **input** on the next turn (after a tool result), `convertReasoningBlock` emitted the item with `id: block.id ?? ""`, i.e. `id: ""`. OpenAI rejects that with:

```
400 Invalid 'input[n].id': ''. Expected an ID that begins with 'rs'.
```

This only happens on the `streamEvents` path when the model produces a reasoning summary **and** a tool call in the same turn; `agent.stream()` is unaffected because it reuses the original response output items (which keep their ids).

## Fix

Omit the `id` field entirely when it is absent, instead of defaulting to an empty string. An empty `id` is the only malformed value OpenAI rejects here; a missing `id` is accepted (and is already how the legacy reconstruction path, `convertReasoningSummaryToResponsesReasoningItem`, behaves when a reasoning item has no id). This keeps the two converter paths consistent.

## Tests

- Added a unit test in `convertStandardContentMessageToResponsesInput` covering a reasoning block with no id, asserting the emitted item omits `id` rather than setting `id: ""`.
- Reverse-verified: with the previous `?? ""` behavior the new test fails (output contains `id: ""`).
- Full `converters/tests/responses.test.ts` suite passes (84 tests); `oxfmt --check`, `oxlint`, and the package build are green.

## Areas for careful review

Whether reasoning items should be *dropped* entirely when they have no id (cannot be referenced by OpenAI) vs. sent without an id (current approach). I kept them, matching the existing `convertReasoningSummaryToResponsesReasoningItem` behavior, to preserve the reasoning summary as context.

## AI Disclosure

This bug was investigated and the fix authored with AI assistance (Claude).

PR #11050, opened 2026-06-11
fix(openai): drop reasoning content blocks in legacy Chat Completions serializer

### LINKED ISSUE #11049: Chat Completions serializer forwards `type: "reasoning"` content blocks verbatim → 400 from OpenAI (legacy content path)
### Checked other resources

### Example Code

```typescript
import { AIMessage, HumanMessage } from "@langchain/core/messages";

const model = new ChatOpenAI({ model: "gpt-4o-mini" });

// An assistant turn replayed from history that carries a `reasoning` content
// block — e.g. persisted by a LangGraph checkpoint, or produced when reasoning
// was enabled and then disabled for this follow-up request.
// Note: no `response_metadata.output_version === "v1"` on this message.
const messages = [
  new HumanMessage("What is 2 + 2?"),
  new AIMessage({
    content: [
      { type: "reasoning", reasoning: "The user wants a simple sum." },
      { type: "text", text: "4" },
    ],
  }),
  new HumanMessage("Now what is 3 + 3?"),
];

await model.invoke(messages);
```

```
400 Invalid value: 'reasoning'. Supported values are: 'text', 'image_url',
'input_audio', 'refusal', 'audio', and 'file'.
{
  "message": "Invalid value: 'reasoning'. Supported values are: 'text', 'image_url', 'input_audio', 'refusal', 'audio', and 'file'.",
  "type": "invalid_request_error",
  "param": "messages[1].content[0].type",
  "code": "invalid_value"
}
```

### Description

When a message whose `content` is an **array** containing a `{ type: "reasoning" }` block is sent to a Chat Completions model, `@langchain/openai` forwards the reasoning block **verbatim** to the OpenAI API, which rejects it with a 400. The OpenAI API does not accept `reasoning` as an input content-part type on the Completions endpoint — this is true both for models that cannot reason at all (e.g. `gpt-4o-mini`) and for reasoning-capable models when the request itself is not a reasoning request.

**I expect** the Completions serializer to handle a `reasoning` content block the same way it already does on the standardized-content path — i.e. drop it (keep only the text) — rather than sending it to the API.

**Instead**, whether the reasoning block is dropped or forwarded depends entirely on `message.response_metadata.output_version`:

`@langchain/openai/dist/converters/completions.js`, `convertMessagesToCompletionsMessageParams` (line ~574):

```js
return messages.flatMap((message) => {
  if ("output_version" in message.response_metadata &&
      message.response_metadata?.output_version === "v1")
    return convertStandardContentMessageToCompletionsMessage({ message }); // ← safe path

  // legacy path:
  const content = typeof message.content === "string"
    ? message.content
    : message.content.flatMap((m) => {
        if (isDataContentBlock(m)) return convertToProviderContentBlock(m, completionsApiContentBlockConverter);
        if (typeof m === "object" && m !== null && "type" in m && m.type === "tool_use") return [];
        return m; // ← a `reasoning` block is neither a data block nor tool_use → forwarded AS-IS
      });
  // ...
});
```

- **v1 path** — `convertStandardContentMessageToCompletionsMessage` (same file, line ~457) does `message.contentBlocks.filter((block) => block.type === "text")`, so reasoning blocks are correctly dropped.
- **Legacy path** — a `reasoning` block is not a `DataContentBlock` and not `tool_use`, so it hits `return m;` and is forwarded unchanged → the 400 above.

So the same input (an `AIMessage` carrying a `reasoning` content block) is serialized safely on one path and into an invalid request on the other, gated only on `output_version`.

**Why this is reachable in normal usage:** assistant messages carrying reasoning blocks are routinely replayed as history — most commonly from a LangGraph checkpoint (`PostgresSaver` etc.), or in any flow that switches a reasoning model off (or down to a non-reasoning model such as `gpt-4o-mini` for a cheap auxiliary call like title/summary generation) while feeding prior turns back in. When the persisted message does not carry `output_version: "v1"` (cross-provider history, or messages produced before that flag is set), the legacy path runs and the
request fails.

For reference, the related reasoning issues (#10421 / #10422 / #10456 on the `contentBlocks` getter dropping reasoning, and #10691 on the chunk-merge collision) all concern *reading/merging* reasoning blocks. This is the inverse: *serializing* an already-correct reasoning block on the **send** side. #10691's fix (now shipped) actually makes this more deterministic — you reliably get a clean standalone `{ type: "reasoning" }` block in history, which then reliably 400s on a non-reasoning replay.

**Suggested fix:** in the legacy `message.content.flatMap(...)` branch of `convertMessagesToCompletionsMessageParams`, skip `reasoning` blocks the same way the v1 path filters to text (`return [];` for `m.type === "reasoning"`), so both paths agree.

### System Info

```
@langchain/core: 1.1.48
@langchain/openai: 1.4.7
@langchain/langgraph: 1.3.7
langchain: 1.4.4

Platform: macOS (Darwin 25.5.0)
Runtime: Bun 1.3.14 (also reproduces under Node)
Node: v26.3.0
```

jackjin1997:
## Summary

Fixes #11049.

When an assistant message carrying a `reasoning` content block is sent through the **Chat Completions** API on the legacy content path (i.e. without `response_metadata.output_version === "v1"`), `convertMessagesToCompletionsMessageParams` forwarded the block verbatim. OpenAI rejects it:

```
400 Invalid value: 'reasoning'. Supported values are: 'text', 'image_url', 'input_audio', 'refusal', ...
```

This happens whenever such a message is replayed from history — e.g. persisted by a LangGraph checkpoint, or produced when reasoning was enabled and then disabled for a follow-up request.

## Fix

Drop `reasoning` blocks on the legacy content path, mirroring how `tool_use` blocks are already dropped a few lines above (they're represented elsewhere / would cause an API error). Reasoning is internal model output with no Chat Completions input content part — the v1 content path already filters to `text` only, so this just brings the legacy path in line.

## Tests

- Added a unit test in the "Anthropic cross-provider compatibility" describe block: an `AIMessage` with a `reasoning` + `text` content array (no `output_version`), asserting the reasoning block is dropped and text passes through.
- Reverse-verified: neutralizing the drop makes the new test fail (reasoning leaks through).
- Full `converters/tests/completions.test.ts` passes (20 tests); `oxfmt --check`, `oxlint`, and the package build are green.

## Note

This is a sibling of #11045 (same `reasoning`-block-as-replayed-input problem on the Responses API). The two are independent and touch different files.

## AI Disclosure

This bug was investigated and the fix authored with AI assistance (Claude).

### DIFF
--- a/.changeset/openai-completions-drop-reasoning.md
+++ b/.changeset/openai-completions-drop-reasoning.md
@@ -1,0 +1,7 @@
+---
+"@langchain/openai": patch
+---
+
+fix(openai): drop reasoning content blocks in legacy Chat Completions serializer
+
+When an assistant message carrying a `reasoning` content block (e.g. replayed from a LangGraph checkpoint, or produced when reasoning was enabled then disabled) is sent through the Chat Completions API without `response_metadata.output_version === "v1"`, the legacy content path forwarded the block verbatim. OpenAI rejects it with `400 Invalid value: 'reasoning'. Supported values are: 'text', 'image_url', 'input_audio', 'refusal', ...`. Reasoning blocks are now dropped on this path, matching how `tool_use` blocks are already handled and how the v1 content path already filters to text.
--- a/libs/providers/langchain-openai/src/converters/completions.ts
+++ b/libs/providers/langchain-openai/src/converters/completions.ts
@@ -818,6 +818,18 @@ export const convertMessagesToCompletionsMessageParams: Converter<
             ) {
               return [];
             }
+            // Drop reasoning blocks — the Chat Completions API has no input
+            // content part for reasoning (it only surfaces in the Responses
+            // API or via additional_kwargs.reasoning_content). Replaying one
+            // verbatim from history triggers `400 Invalid value: 'reasoning'`.
+            if (
+              typeof m === "object" &&
+              m !== null &&
+              "type" in m &&
+              m.type === "reasoning"
+            ) {
+              return [];
+            }
             return m;
           });
     // oxlint-disable-next-line @typescript-eslint/no-explicit-any
--- a/libs/providers/langchain-openai/src/converters/tests/completions.test.ts
+++ b/libs/providers/langchain-openai/src/converters/tests/completions.test.ts
@@ -562,5 +562,29 @@ describe("convertCompletionsMessageToBaseMessage", () => {
       // tool_calls should still be present
       expect((result[0] as any).tool_calls).toHaveLength(1);
     });
+
+    it("should drop reasoning blocks from legacy content (not a valid Completions input part)", () => {
+      // An assistant turn replayed from history that carries a `reasoning`
+      // content block (e.g. persisted by a checkpoint, or produced when
+      // reasoning was enabled and later disabled). With no `output_version: v1`
+      // it goes through the legacy content path. Forwarding `type: "reasoning"`
+      // verbatim triggers `400 Invalid value: 'reasoning'` from OpenAI.
+      const message = new AIMessage({
+        content: [
+          { type: "reasoning", reasoning: "The user wants a simple sum." },
+          { type: "text", text: "4" },
+        ],
+      });
+
+      const result = convertMessagesToCompletionsMessageParams({
+        messages: [message],
+      });
+
+      expect(result).toHaveLength(1);
+      const contentArr = result[0].content as any[];
+      // reasoning is dropped; text passes through
+      expect(contentArr.some((c: any) => c.type === "reasoning")).toBe(false);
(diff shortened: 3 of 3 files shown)


jackjin1997:
@christian-bromann now that #11045 landed for the Responses API path, this is the Chat Completions sibling for the same underlying issue (reasoning blocks serialized verbatim into a legacy-path request that OpenAI rejects). Same fix shape, tests green — would be great to get this one in too.

jackjin1997:
This has been superseded by the broader merged fix in #11175, which filters `reasoning`, `reasoning_content`, and `thinking` blocks from the same Chat Completions conversion path. The issue reporter also confirmed the fix in `@langchain/openai` 1.5.5. Closing this duplicate implementation; happy to revisit if the regression reappears.

PR #11065, opened 2026-06-13
fix(openai): route standard url file blocks to native input_file in Responses API

### LINKED ISSUE #9895: File URL content blocks incorrectly routed through completions converter when using Responses API
### Checked other resources

### Example Code

```typescript
import { HumanMessage } from '@langchain/core/messages';

  model: 'gpt-4o',
  useResponsesApi: true,  // Explicitly using Responses API
});

// Using LangChain's standard file content block format
  content: [
    { type: 'text', text: 'What is in this document?' },
    {
      type: 'file',
      source_type: 'url',
      url: 'https://example.com/document.pdf',
      mime_type: 'application/pdf',
      metadata: { filename: 'document.pdf' },
    },
  ],
});

// This throws an error even though Responses API supports file URLs
const response = await model.invoke([message]);
```

```
Error: URL file blocks with source_type url must be formatted as a data URL for ChatOpenAI
```

### Description

When using `ChatOpenAI` with `useResponsesApi: true`, file content blocks with `source_type: 'url'` are incorrectly routed through the Chat Completions API converter instead of being handled by the Responses API converter.

**Root Cause:**

In `@langchain/openai/dist/converters/responses.js`, the routing logic at line ~970 checks `isDataContentBlock(item)` which returns `true` for standard file blocks:

```javascript
// Line ~970 - Standard format gets routed through completions converter
if (isDataContentBlock(item))
    return convertToProviderContentBlock(item, completionsApiContentBlockConverter);

// Line ~992 - Native format passes through directly (this works!)
if (item.type === "input_text" || item.type === "input_image" || item.type === "input_file")
    return item;
```

The `completionsApiContentBlockConverter.fromStandardFileBlock()` then expects the URL to be a base64 data URL, not an HTTP URL, and throws the error.

**The Problem:**
- OpenAI's **Chat Completions API** does NOT support file URLs (only `file_id` or base64 `file_data`)
- OpenAI's **Responses API** DOES support file URLs via the native `input_file` format
- But LangChain routes standard file blocks through the completions converter regardless of which API is being used

**Expected Behavior:**

When `useResponsesApi: true`, file content blocks with `source_type: 'url'` should be converted to OpenAI's native format:

```javascript
// Input (LangChain standard format)
{ type: 'file', source_type: 'url', url: 'https://example.com/doc.pdf', metadata: { filename: 'doc.pdf' } }

// Expected output (OpenAI native format for Responses API)
{ type: 'input_file', file_url: 'https://example.com/doc.pdf' }
```

This native format passes through correctly at line ~992.

**Current Workaround:**

We're extending `ChatOpenAI` and overriding `_generate()` to transform file content blocks to native format before calling `super._generate()`:

### System Info

"@langchain/openai": "^1.2.0",
@langchain/core: 1.1.8
Node.js:  v22.21.1 
Platform: macOS / Linux

rxits:
**Fixes #9895**

### Problem
With `ChatOpenAI({ useResponsesApi: true })`, a standard file content block using `source_type: "url"` throws:

```
Error: URL file blocks with source_type url must be formatted as a data URL for ChatOpenAI
```

even though the Responses API supports file URLs natively.

### Root cause
`convertMessagesToResponsesInput` (used by the Responses chat model) routes every `isDataContentBlock` item through `completionsApiContentBlockConverter`. Its `fromStandardFileBlock` only accepts base64 *data* URLs for `source_type: "url"` and throws on plain HTTP URLs — correct for Chat Completions (which has no file-URL support), but wrong for the Responses API, which accepts `input_file.file_url`.

The newer `convertStandardContentMessageToResponsesInput` path (v1 `contentBlocks`) already handles this correctly via `resolveFileItem`; only the legacy standard-content path was affected.

### Fix
In `convertMessagesToResponsesInput`, convert standard **file** blocks to the native Responses `input_file` shape before falling back to the completions converter:

- `source_type: "url"` → `{ type: "input_file", file_url }`
- `source_type: "id"` → `{ type: "input_file", file_id }`
- `source_type: "base64"` → `{ type: "input_file", file_data }`

Filename is preserved from metadata when present, matching the existing `resolveFileItem` behavior. Non-file data blocks (image/audio) keep their existing routing.

### Tests
Added a regression test in `responses.test.ts` asserting a `source_type: "url"` file block converts to `input_file` with `file_url` (and preserves `filename`). Verified failing before the fix (with the exact reported error) and passing after; full converter suite (103 tests) green with no type errors.

### DIFF
--- a/.changeset/smart-deer-sniff.md
+++ b/.changeset/smart-deer-sniff.md
@@ -1,0 +1,5 @@
+---
+"@langchain/openai": patch
+---
+
+fix(openai): route standard url file blocks to native input_file in Responses API
--- a/libs/providers/langchain-openai/src/converters/responses.ts
+++ b/libs/providers/langchain-openai/src/converters/responses.ts
@@ -1657,6 +1657,35 @@ export const convertMessagesToResponsesInput: Converter<
             });
           }
           if (isDataContentBlock(item)) {
+            // The Responses API supports file URLs natively, but the Chat
+            // Completions converter rejects URL file blocks. Convert standard
+            // file blocks to the native `input_file` shape instead of routing
+            // them through the completions converter.
+            if (item.type === "file") {
+              const filename = getFilenameFromMetadata(item);
+              if (item.source_type === "url") {
+                return {
+                  type: "input_file",
+                  file_url: item.url,
+                  ...(filename ? { filename } : {}),
+                };
+              }
+              if (item.source_type === "id") {
+                return {
+                  type: "input_file",
+                  file_id: item.id,
+                  ...(filename ? { filename } : {}),
+                };
+              }
+              if (item.source_type === "base64") {
+                const mimeType = item.mime_type ?? "";
+                return {
+                  type: "input_file",
+                  file_data: `data:${mimeType};base64,${item.data}`,
+                  filename: getRequiredFilenameFromMetadata(item),
+                };
+              }
+            }
             return convertToProviderContentBlock(
               item,
               completionsApiContentBlockConverter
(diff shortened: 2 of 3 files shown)


christian-bromann (dismissed):
LGTM 👍

PR #11075, opened 2026-06-16
fix(core): keep streamed text and tool calls in separate content blocks

### LINKED ISSUE #11074: Streaming text + tool call can collide content block indexes, causing `AIMessage.tool_calls` to be empty
### Checked other resources

### Example Code

```ts
import { AIMessageChunk } from "@langchain/core/messages";
import { ChatGenerationChunk } from "@langchain/core/outputs";
import { convertChunksToEvents } from "@langchain/core/language_models/compat";
import { ChatModelStream } from "@langchain/core/language_models/stream";

async function* chunks() {
    message: new AIMessageChunk({
      content: "I will call a tool",
      tool_call_chunks: [
        {
          index: 0,
          id: "call_1",
          name: "eval",
          args: "{\"code\"",
        },
      ],
    }),
    text: "I will call a tool",
  });

    message: new AIMessageChunk({
      content: "",
      tool_call_chunks: [
        {
          index: 0,
          args: ":\"new Date()\"}",
        },
      ],
    }),
    text: "",
  });
}

const stream = new ChatModelStream(convertChunksToEvents(chunks()));
const message = await stream.output;

console.log(message.content);
console.log(message.tool_calls);
```

## Actual Behavior

The text block and first tool-call chunk can share content block index `0`. The tool-call fields are attached to the text block rather than emitted as a separate `tool_call_chunk` block.

The final message can have no tool calls:

```ts
message.tool_calls; // []
```

## Expected Behavior

String text and tool calls should become separate content blocks:

```ts
[
  {
    type: "text",
    text: "I will call a tool",
  },
  {
    type: "tool_call",
    id: "call_1",
    name: "eval",
    args: {
      code: "new Date()",
    },
  },
]
```

The final `AIMessage.tool_calls` should contain the tool call so agents can execute it.

_No response_

### Description

# Streaming text + tool call can collide content block indexes, causing `AIMessage.tool_calls` to be empty

## Description

When an OpenAI-compatible streaming model returns both normal text content and a tool call in the same streamed output, LangChain can stop the agent instead of executing the tool call.

The streamed data contains a valid tool call, but after LangChain converts legacy stream chunks into content block events, the final `AIMessage` has `tool_calls: []`. In `ReactAgent`, this causes the agent to exit because it thinks there are no tool calls to run.

## What Happens

In `@langchain/core/dist/language_models/compat.js`, `convertChunksToEvents` appears to use content block index `0` for streamed text content:

```js
const blockIndex = 0;
```

Later, streamed OpenAI tool call chunks also commonly use `toolChunk.index === 0`.

Because both text and the first tool call use block index `0`, the tool call fields are merged into the already-active text block instead of starting a separate `tool_call_chunk` block.

The resulting content can look like this:

```js
[
  {
    type: "text",
    text: "I will call a tool",
    id: "call_123",
    name: "eval",
    args: "{\"code\":\"new Date()\"}"
  }
]
```

Since the block is still `type: "text"`, the final `AIMessage` does not populate `tool_calls`, and the agent stops.

## Expected Behavior

Text content and tool calls should become separate content blocks, for example:

```js
[
  {
    type: "text",
    text: "I will call a tool"
  },
  {
    type: "tool_call",
    id: "call_123",
    name: "eval",
    args: {
      code: "new Date()"
    }
  }
]
```

The final `AIMessage.tool_calls` should contain the tool call so the agent executes it.

## Suggested Fix

`toolChunk.index` should not be used directly as the v3 content block index, because provider tool call indexes are scoped to the tool call list, not to all content blocks.

A local patch that fixed this was to map tool call indexes to separate content block indexes:

```js
const toolCallBlockIndexes = new Map();

// ...

const toolCallKey =
  typeof toolChunk.index === "number"
    ? `index:${toolChunk.index}`
    : typeof toolChunk.id === "string"
      ? `id:${toolChunk.id}`
      : `next:${toolCallBlockIndexes.size}`;

let blockIndex = toolCallBlockIndexes.get(toolCallKey);

if (blockIndex === undefined) {
  blockIndex = nextBlockIndex(activeBlocks);
  toolCallBlockIndexes.set(toolCallKey, blockIndex);
}
```

With this change, streamed text remains block `0`, and the first tool call becomes block `1`, preventing the collision.

### System Info

- `langchain`: `1.4.1`
- `@langchain/core`: `1.1.47`
- `@langchain/openai`: `1.4.6`
- Node: `>=20`

jackjin1997:
## Summary

Fixes #11074.

When a legacy stream chunk carries both string `content` and a tool call (common for OpenAI-compatible models that emit a brief text preamble before a tool call), `convertChunksToEvents` could drop the tool call entirely — the final `AIMessage` ended up with `tool_calls: []` and a `ReactAgent` would stop instead of executing the tool.

### Root cause

Streamed string text is assigned content block index `0`:

```ts
const blockIndex = 0; // text path
```

Tool call chunks then used the provider's `tool_call_chunk.index` directly as the content block index:

```ts
const blockIndex =
  typeof toolChunk.index === "number" ? toolChunk.index : activeBlocks.size;
```

A provider's first tool call commonly has `index === 0`, so it collided with the already-active text block. Because block `0` already existed as `type: "text"`, the tool-call branch skipped creating a new `tool_call_chunk` block and merged the tool fields onto the text block — which stayed `type: "text"`. `finalizeContentBlock` therefore never produced a `tool_call`, so `AIMessage.tool_calls` was empty.

Provider tool-call indexes are scoped to the tool-call list, not to the global content-block list. The fix maps each distinct tool call onto its own freshly allocated content block index (via the existing `nextBlockIndex` helper), keeping streamed text at block `0` and the first tool call at block `1`. Continuation chunks for the same tool call (same provider index / id) still resolve to the same block.

### Behavior unchanged for non-colliding streams

When there is no text block, `nextBlockIndex` allocates `0` for the first tool call and `1` for the second, exactly matching the previous indexes — so existing tool-call streaming is unaffected (verified by the existing `FakeToolCallStreamModel` test).

## Testing

Added a regression test (`FakeTextAndToolCallStreamModel`) mirroring the issue repro: a chunk with both text and a `tool_call_chunk` at `index: 0`, followed by an args continuation chunk. It asserts the text and tool call land on different content block indexes and that the finalized message surfaces the tool call. Reverse-verified: the test fails without the fix.

## AI Disclosure

This fix was prepared with AI assistance (issue analysis, implementation, and tests).

### DIFF
--- a/.changeset/streaming-text-toolcall-index-collision.md
+++ b/.changeset/streaming-text-toolcall-index-collision.md
@@ -1,0 +1,15 @@
+---
+"@langchain/core": patch
+---
+
+fix(core): keep streamed text and tool calls in separate content blocks
+
+When a legacy stream chunk carried both string `content` and a tool call,
+`convertChunksToEvents` placed the text at content block index `0` and used
+the provider's `tool_call_chunk.index` (also `0`) directly as the content
+block index. The two collided: the tool call fields were merged onto the
+existing text block, which stayed `type: "text"`, so the finalized
+`AIMessage` had `tool_calls: []` and agents would stop instead of running
+the tool. Provider tool-call indexes are scoped to the tool-call list, not
+to the global content-block list, so each distinct tool call is now mapped
+onto its own freshly allocated content block index.
--- a/libs/langchain-core/src/language_models/compat.ts
+++ b/libs/langchain-core/src/language_models/compat.ts
@@ -154,6 +154,11 @@ export async function* convertChunksToEvents(
     | undefined;
   let audioStream: AudioStreamState | undefined;
   const emittedImageKeys = new Set<string>();
+  // Provider tool-call indexes are scoped to the tool-call list, not to the
+  // global content-block list, so they cannot be used as block indexes
+  // directly (e.g. text occupies block 0 and a tool call with index 0 would
+  // collide with it). Map each distinct tool call onto its own block index.
+  const toolCallBlockIndexes = new Map<string, number>();
 
   for await (const chunk of chunks) {
     options?.signal?.throwIfAborted();
@@ -239,10 +244,18 @@ export async function* convertChunksToEvents(
       msg.tool_call_chunks.length > 0
     ) {
       for (const toolChunk of msg.tool_call_chunks) {
-        const blockIndex =
+        const toolCallKey =
           typeof toolChunk.index === "number"
-            ? toolChunk.index
-            : activeBlocks.size;
+            ? `index:${toolChunk.index}`
+            : typeof toolChunk.id === "string"
+              ? `id:${toolChunk.id}`
+              : `position:${toolCallBlockIndexes.size}`;
+
+        let blockIndex = toolCallBlockIndexes.get(toolCallKey);
+        if (blockIndex === undefined) {
+          blockIndex = nextBlockIndex(activeBlocks);
+          toolCallBlockIndexes.set(toolCallKey, blockIndex);
+        }
 
         if (!activeBlocks.has(blockIndex)) {
           const initial: ContentBlock = {
--- a/libs/langchain-core/src/language_models/tests/stream_bridge.test.ts
+++ b/libs/langchain-core/src/language_models/tests/stream_bridge.test.ts
@@ -170,6 +170,48 @@ class FakeToolCallStreamModel extends BaseChatModel {
 }
 
 /**
+ * A model that yields string text content and a tool call in the same chunk,
+ * where the provider tool-call index (0) collides with the text block index.
+ */
+class FakeTextAndToolCallStreamModel extends BaseChatModel {
+  _llmType() {
+    return "fake-text-and-tool-stream";
+  }
+
+  async _generate(
+    _messages: BaseMessage[],
+    _options: this["ParsedCallOptions"],
+    _runManager?: CallbackManagerForLLMRun
+  ): Promise<ChatResult> {
+    return { generations: [] };
+  }
+
+  async *_streamResponseChunks(
+    _messages: BaseMessage[],
+    _options: this["ParsedCallOptions"],
+    _runManager?: CallbackManagerForLLMRun
+  ): AsyncGenerator<ChatGenerationChunk> {
+    yield new ChatGenerationChunk({
+      message: new AIMessageChunk({
+        content: "I will call a tool",
+        id: "msg_text_tool",
+        tool_call_chunks: [
+          { id: "call_1", name: "eval", args: '{"code"', index: 0 },
+        ],
+      }),
+      text: "I will call a tool",
+    });
+    yield new ChatGenerationChunk({
+      message: new AIMessageChunk({
+        content: "",
+        tool_call_chunks: [{ args: ':"new Date()"}', index: 0 }],
+      }),
+      text: "",
+    });
+  }
+}
+
+/**
  * A model that yields chunks with usage_metadata.
  */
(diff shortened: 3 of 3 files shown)


jackjin1997:
Bumping this one — it's a core streaming-protocol fix (text + tool-call chunks colliding on content block index 0 can silently drop `tool_calls` from the finalized message), tests are green, no API surface changes. Happy to rebase if main has drifted.

NitsanCohen770:
Independent confirmation from a real production repro (not a synthetic test): we hit this exact bug integrating Moonshot AI's Kimi K3 (OpenAI-compatible Chat Completions API, via `ChatOpenAI` + a custom `baseURL`) as `langchain`'s `createAgent`/`deepagents`' underlying model, dispatched through `CompiledStateGraph.streamEvents(..., {version:"v3"})` — the real production path for our agent runtime (needed for live token/tool-progress streaming to the UI).

Symptom matched precisely: a turn with ~76 bound tools where the model emitted a short text preamble ("Let me pull your current cash balances.") before a tool call. The final `AIMessage` had `tool_calls: []`, `response_metadata.finish_reason: "stop"`, and the tool call's `id`/`name`/`args` merged onto a single `content` block that stayed `type: "text"` — causing the agent to silently stop instead of executing the tool. A plain `.invoke()` (which doesn't attach the `lc_prefer_chat_model_stream_events` handler and takes the legacy `_streamResponseChunks` + `concat()` path) never showed this; only `streamEvents(...,{version:"v3"})` did — consistent with this PR's root-cause analysis of `convertChunksToEvents`.

We independently traced it to the same `blockIndex = 0` (text) vs. `toolChunk.index` (tool call) collision in `convertChunksToEvents`, and applied this PR's exact diff as a local `pnpm patch` against the compiled `@langchain/core@1.2.1` package (both `compat.js` and `compat.cjs`). Verified against:

1. A network-free synthetic repro (two text chunks + two tool-call chunks at `index: 0`) — before: `tool_calls: []`; after: `tool_calls` correctly populated, text and tool call as separate content blocks.
2. The real production agent (deepagents + Kimi K3, 76 real tools bound) driven through `streamEvents(...,{version:"v3"})` — before: broken exactly as described above; after: a full multi-turn tool-calling sequence (4 sequential tool calls across a real user query) all correctly populate `tool_calls`, with clean text/tool_call block separation throughout.

This fix is correct and safe to merge as far as our (unrelated, independent) repro is concerned. One remaining related issue we noticed but which is out of scope for this PR: `message-finish`'s `reason` is hardcoded to `"stop"` in `convertChunksToEvents`/`ChatModelStream._assembleMessage` regardless of the underlying chunk's actual finish reason — worth a follow-up, since some consumers may branch on `response_metadata.finish_reason` rather than `tool_calls.length`.

Thanks for tracking this down — would love to see this merged and released.

jackjin1997:
@christian-bromann The independent production repro above now validates the exact `streamEvents` path and the text/tool-call index collision this patch fixes. The change remains limited to block allocation with a reverse-verified regression test. Are there any blockers I should address to move this forward?

jackjin1997:
@hntrl Following up now that #11074 has three open implementations and the production reproduction remains unresolved. This was the first PR and includes focused, reverse-verified regression coverage; checks remain green. Could you advise which implementation direction should be consolidated for review?

jackjin1997:
Closing this for now after roughly 50 days without maintainer review, despite green checks and an independent production reproduction. There are multiple open implementations for the same issue, so I am reducing duplicate review load. Happy to revisit if maintainers decide to consolidate around this direction.

PR #11077, opened 2026-06-16
feat(core): support exposing bound tools and validate conflicting call options (#7757)

SapphireSmith:
## Summary
This PR addresses two issues (matching #7757) regarding tool binding and structured output:
1. Adds a `.tools` getter property on `RunnableBinding` and `BaseChatModel` so developers can inspect which tools are currently bound.
2. Throws a clear local validation error if a developer tries to pass `tools` in the call options during invocation (`invoke`, `stream`, etc.) of a model configured with structured output.

## Details
- **Tool Inspection**: Added `tools` getter on `RunnableBinding` (recursively checking `this.bound` or config/kwargs) and a default `tools` getter returning `[]` on `BaseChatModel`.
- **Conflict Validation**: In `assembleStructuredOutputPipeline`, we inject the structured output config metadata (`is_structured_output`, `structured_output_tool_name`) on the LLM. If `is_structured_output` is active and custom `tools` are passed in call options during invocation, `_separateRunnableConfigFromCallOptionsCompat` detects the missing structured output tool and throws a descriptive validation error to prevent a 400 API error.

## Verification
- Added automated unit tests to `libs/langchain-core/src/language_models/tests/chat_models.test.ts`.
- Verified using a local reproduction script that `.tools` logs the bound tools successfully and the correct validation error is thrown before calling the API.


libs/langchain-core/src/language_models/chat_models.ts
```diff
@@ -258,6 +258,40 @@ export abstract class BaseChatModel<
...
     (callOptions as this["ParsedCallOptions"]).signal = runnableConfig.signal;
+
+    if (runnableConfig.configurable?.is_structured_output) {
+      const expectedToolName =
+        runnableConfig.configurable.structured_output_tool_name;
+      const parsedTools = (callOptions as Record<string, unknown>).tools;
+      if (parsedTools !== undefined) {
+        const hasExpectedTool =
+          Array.isArray(parsedTools) &&
+          parsedTools.some((tool: unknown) => {
```
sviat-barbutsa, line 275: This is a useful guard, but I think there is one more mixed-tool-list case worth covering because the new test covers `tools: [extraTool]`, but this check still passes if call options include the structured-output tool plus another tool, because `hasExpectedTool` only checks `some(...)` by name.

That seems to leave part of the conflict case open, the model would still be invoked with a caller-provided `tools` list while configured for structured output.

Could you please add some regression test where `withStructuredOutput(...).invoke(...)` receives the expected structured-output tool plus an extra tool, and assert it still fails before the provider call?

SapphireSmith: Good catch! I've updated the validation check to reject mixed lists (expected tool + extra tools) and empty lists too. 

I also added a regression test covering all three cases (only extra, mixed, and only expected tools) to make sure it functions correctly. 

Pushed the commits now!

PR #11088, opened 2026-06-19
feat(weaviate): add X-Weaviate-Client-Integration telemetry header

dudanogueira:
## Summary

Tags Weaviate connections opened by `WeaviateStore` with an
`X-Weaviate-Client-Integration: langchain/<version>` header so Weaviate can
track LangChain.js usage in its telemetry. This mirrors the Python
`langchain-weaviate` integration, which registers the same identifier.

## Motivation

The Python `weaviate-client` exposes a public `integrations.configure(...)` API
that LangChain's Python integration calls to identify itself. The JS client has
no equivalent public API, so until now JS LangChain traffic was indistinguishable
from raw client traffic in Weaviate telemetry. This change closes that gap on the
JS side.

## Design decisions

- **Header registration without private internals.** `getConnectionDetails()`
  returns the client's live `headers` object *by reference*, and the client
  spreads that same object into every HTTP and gRPC request. Mutating it in place
  tags all subsequent requests on both transports without reaching into any
  private client internals.
- **Best-effort and non-blocking.** Registration is fire-and-forget from the
  constructor and the helper never throws — if a future client version changes
  shape (e.g. `getConnectionDetails` missing, or array-form headers), it silently
  skips registration rather than breaking store construction. This matches the
  spirit of Python's registration in `__init__`.
- **Integration name is just `langchain`** (not language-specific), consistent
  with the Python integration; language is already disambiguated by the client's
  own `X-Weaviate-Client` header.
- **Version injection.** `__PKG_VERSION__` is defined at build time via `tsdown`
  and mirrored in `vitest.config.ts` so tests can assert the exact header value.

## Tests

Added unit tests covering the three paths:
- header is registered via `getConnectionDetails` (asserts `langchain/<version>`)
- registration never throws when the client doesn't support it
- array-form headers are left untouched

## Risk

Low. The feature is purely additive telemetry, runs best-effort, and cannot
affect store construction or query behavior even if header registration fails.

### DIFF
--- a/libs/providers/langchain-weaviate/src/tests/vectorstores.test.ts
+++ b/libs/providers/langchain-weaviate/src/tests/vectorstores.test.ts
@@ -54,6 +54,56 @@ function makeSearchMockClient() {
   };
 }
 
+/** Let the fire-and-forget header registration settle before asserting. */
+function flushMicrotasks(): Promise<void> {
+  return new Promise((resolve) => {
+    setTimeout(resolve, 0);
+  });
+}
+
+test("registers the X-Weaviate-Client-Integration header via getConnectionDetails", async () => {
+  const headers: Record<string, string> = {};
+  const client = {
+    collections: {},
+    getConnectionDetails: vi
+      .fn()
+      .mockResolvedValue({ host: "localhost", headers }),
+  } as unknown as WeaviateClient;
+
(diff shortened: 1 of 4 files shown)


christian-bromann (approved):
LGTM 👍

PR #11100, opened 2026-06-22
fix(langchain, openai): decouple strict tools from strict structured output response

colifran:
### Summary
When an agent requests native / provider structured output (ProviderStrategy), the agent core marked every bound tool strict: true. On Anthropic, a large strict-tool surface makes the combined compiled grammar exceed the API's limits:

`400 invalid_request_error: Schema is too complex for compilation. Try reducing the number of tools or simplifying tool schemas.`

Strict function tools are only actually required by OpenAI Chat Completions (its .parse() path). Anthropic, Gemini, and the OpenAI Responses API do not need them. This PR stops the core from forcing tool strictness for all providers and moves the requirement to where it belongs.

### Tests
1. langchain core (AgentNode): stop forcing strict: true on tools for structured output. Tool strictness now comes only from an explicit modelSettings.strict, so each provider decides. (The response schema's own strictness is unchanged.)
2. @langchain/openai (ChatOpenAICompletions.invocationParams): re-apply strict: true to tools only where OpenAI actually needs it, a non-streaming json_schema request (the .parse() path). Not the Responses API, not streaming, not JSON mode, and an explicit strict: false still wins.

### Repro

Before change:
```
=== anthropic (claude-opus-4-7) ===
    scenario      status  tools  strict  native  note
    native-strict 400     19     19      yes     Schema is too complex for compilation. Try reducing the number of tools or simplifying tool schemas.
    native-loose  200     19     0       yes     → {"haiku":"Crimson leaves descend—\n…"}
    tool          200     20     0       no      → {"haiku":"Crimson leaves descend\n…"}
```

After change:
```
=== anthropic (claude-opus-4-7) ===
    scenario      status  tools  strict  native  note
    native-strict 200     19     0       yes     → {"haiku":"Crisp leaves drift slowly\nWhispering on amber winds\nAutumn breathes goodbye"}
    native-loose  200     19     0       yes     → {"haiku":"Crimson leaves descend—\nwhispers of the cooling wind\ncrunch beneath my steps"}
    tool          200     20     0       no      → {"haiku":"Crisp leaves drift downward\nWhispering of fading warmth—\nAutumn breathes goodbye"}
```

### DIFF
--- a/libs/langchain/src/agents/nodes/AgentNode.ts
+++ b/libs/langchain/src/agents/nodes/AgentNode.ts
@@ -1019,7 +1019,13 @@ export class AgentNode<
           kwargs: { method: "json_schema" },
           schema: structuredResponseFormat.strategy.schema,
         },
-        strict: resolvedStrict,
+
+        /**
+         * Don't force strict on tools: it makes Anthropic's combined grammar
+         * "too complex for compilation", and only OpenAI Chat Completions needs
+         * it (re-applied there). Honor an explicit override; else leave unset.
+         */
+        strict: preparedOptions?.modelSettings?.strict,
       });
     }
 
--- a/.changeset/full-dryers-bake.md
+++ b/.changeset/full-dryers-bake.md
@@ -1,0 +1,6 @@
+---
+"@langchain/openai": patch
+"langchain": patch
+---
+
+fix(langchain, openai): decouple strict tools from strict structured output response
--- a/libs/langchain/src/agents/tests/responses.test.ts
+++ b/libs/langchain/src/agents/tests/responses.test.ts
@@ -1,15 +1,21 @@
-import { describe, expect, it } from "vitest";
+import { describe, expect, it, vi } from "vitest";
 import { z } from "zod/v3";
 import { z as z4 } from "zod/v4";
 
 import { ChatOpenAI } from "@langchain/openai";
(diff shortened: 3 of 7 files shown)


libs/langchain/src/agents/nodes/AgentNode.ts
```diff
@@ -1019,7 +1019,13 @@ export class AgentNode<
...
           schema: structuredResponseFormat.strategy.schema,
         },
-        strict: resolvedStrict,
+
+        /**
+         * Don't force strict on tools: it makes Anthropic's combined grammar
+         * "too complex for compilation", and only OpenAI Chat Completions needs
+         * it (re-applied there). Honor an explicit override; else leave unset.
+         */
+        strict: preparedOptions?.modelSettings?.strict,
```
hntrl, line 1028: shouldn't this be
```typescript
preparedOptions?.modelSettings?.strict ??
structuredResponseFormat?.strategy?.strict
```
?

hntrl: also idk how helpful the comment is

colifran: no it shouldn't be - we're decoupling it from the responseFormat strategy they are two separate things

hntrl (approved):

PR #11119, opened 2026-06-29
fix(core): avoid stream tool call block collisions

### LINKED ISSUE #11074: Streaming text + tool call can collide content block indexes, causing `AIMessage.tool_calls` to be empty

mturac:
Fixes #11074.

## Summary
- map legacy `tool_call_chunks` provider indexes to independent content block indexes
- keep streamed text and tool calls in separate content blocks when their indexes overlap
- add regression coverage and a patch changeset for `@langchain/core`

## Testing
- `pnpm --filter @langchain/core test src/language_models/tests/stream_bridge.test.ts`
- `pnpm exec oxfmt --check libs/langchain-core/src/language_models/compat.ts libs/langchain-core/src/language_models/tests/stream_bridge.test.ts .changeset/stream-tool-call-indexes.md`
- `pnpm exec oxlint libs/langchain-core/src/language_models/compat.ts libs/langchain-core/src/language_models/tests/stream_bridge.test.ts`
- `git diff --check`
- `pnpm changeset status --since main`

### DIFF
--- a/.changeset/stream-tool-call-indexes.md
+++ b/.changeset/stream-tool-call-indexes.md
@@ -1,0 +1,5 @@
+---
+"@langchain/core": patch
+---
+
+Fix legacy stream conversion when text and tool call chunks share provider indexes.
--- a/libs/langchain-core/src/language_models/compat.ts
+++ b/libs/langchain-core/src/language_models/compat.ts
@@ -154,6 +154,7 @@ export async function* convertChunksToEvents(
     | undefined;
   let audioStream: AudioStreamState | undefined;
   const emittedImageKeys = new Set<string>();
+  const toolCallBlockIndexes = new Map<string, number>();
 
   for await (const chunk of chunks) {
     options?.signal?.throwIfAborted();
@@ -239,10 +240,22 @@ export async function* convertChunksToEvents(
       msg.tool_call_chunks.length > 0
     ) {
       for (const toolChunk of msg.tool_call_chunks) {
-        const blockIndex =
+        const blockKeys = [
           typeof toolChunk.index === "number"
-            ? toolChunk.index
-            : activeBlocks.size;
+            ? `index:${toolChunk.index}`
+            : undefined,
+          typeof toolChunk.id === "string" ? `id:${toolChunk.id}` : undefined,
+        ].filter((key): key is string => key != null);
+        let blockIndex = blockKeys
+          .map((key) => toolCallBlockIndexes.get(key))
+          .find((index): index is number => index != null);
+
+        if (blockIndex == null) {
+          blockIndex = nextBlockIndex(activeBlocks);
+        }
+        for (const key of blockKeys) {
+          toolCallBlockIndexes.set(key, blockIndex);
+        }
 
         if (!activeBlocks.has(blockIndex)) {
           const initial: ContentBlock = {
@@ -250,7 +263,10 @@ export async function* convertChunksToEvents(
             id: toolChunk.id,
             name: toolChunk.name,
             args: "",
-            index: blockIndex,
+            index:
+              typeof toolChunk.index === "number"
+                ? toolChunk.index
+                : blockIndex,
           };
           activeBlocks.set(blockIndex, {
             type: "tool_call_chunk",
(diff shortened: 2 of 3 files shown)


mturac:
Closing this to avoid splitting review across multiple open fixes for the same issue. #11075, #11091, and #11114 are already open for #11074, so I will leave the active review path there.

PR #11067, opened 2026-06-14
docs(tavily): fix "refering" typo in tool description

atimothee:
## Summary
- Fix "refering" → "referring" in `TavilyCrawl.description` (`libs/providers/langchain-tavily/src/tavily-crawl.ts`).
- This string is the tool description that LLMs see when deciding to invoke the tool, so a cleaner prompt is a small quality win.

## Validation
- `pnpm --filter @langchain/tavily build` → build completes, attw/publint clean
- `git diff` → 1 line changed in a single file

Pre-existing lint and prettier issues on this package were not introduced by this change (they reproduce on `main`).

No linked issue.

Made with [Cursor](https://cursor.com)

PR #11066, opened 2026-06-14
docs(ibm): fix "Recieved" typo in tool_choice error message

atimothee:
## Summary
- Fix a small typo (`Recieved` → `Received`) in the user-facing error message thrown from `_convertToolChoiceToWatsonxToolChoice` in `@langchain/ibm`.
- The error message is what `ChatWatsonx` consumers see when they pass an invalid `tool_choice`, so the typo is visible in the wild.

## Validation
- `pnpm install --frozen-lockfile`
- `pnpm format:check` → all matched files use the correct format
- `pnpm lint` → no issues
- `pnpm --filter @langchain/ibm build` → build successful (incl. `attw` and `publint`)
- `pnpm --filter @langchain/ibm test` → 7 test files / 118 tests passed, no type errors

No behavioral change — single-line spelling fix in an error message.

Made with [Cursor](https://cursor.com)

PR #11126, opened 2026-06-30
fix(groq): wire missing constructor params and add interface fields

jahanzaib-iqbal-dev:
## Problem

Three bugs found in the Groq provider that silently break developer expectations:

### Bug 1 — `topLogprobs`, `reasoningFormat`, `serviceTier` silently ignored

These fields are defined as class properties and used in `invocationParams()`, but were **never assigned in the constructor**. A developer passing any of these at construction time gets no error — the values are silently dropped.

```ts
// Before — these did nothing at construction time
const llm = new ChatGroq({ topLogprobs: 5, reasoningFormat: "parsed", serviceTier: "auto" })
// llm.topLogprobs === undefined ❌
```

```ts
// After
// llm.topLogprobs === 5 ✅
```

### Bug 2 — `reasoningFormat` and `serviceTier` missing from `ChatGroqInput`

Both fields are used internally but were not exposed in the public interface, making it impossible for TypeScript users to set them at construction time with type safety.

### Bug 3 — Deprecated model in JSDoc example

The "Bind tools" example used `"llama3-groq-70b-8192-tool-use-preview"` which is deprecated and not in the model registry. Updated to `"llama-3.3-70b-versatile"`.

## Changes

- `chat_models.ts` — Added 3 missing constructor assignments + 2 new `ChatGroqInput` interface fields with JSDoc + updated deprecated model in example
- `chat_models.test.ts` — Added unit tests covering all three constructor param assignments

## Type
- [x] Bug fix (silent data loss on construction)
- [x] Tests added

### DIFF
--- a/libs/providers/langchain-groq/src/chat_models.ts
+++ b/libs/providers/langchain-groq/src/chat_models.ts
@@ -198,7 +198,7 @@ export interface ChatGroqInput extends BaseChatModelParams {
 
   /**
    * The maximum number of tokens that the model can process in a single response.
-   * This limits ensures computational efficiency and resource management.
+   * This limit ensures computational efficiency and resource management.
    */
   maxTokens?: number;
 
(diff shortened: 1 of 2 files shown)


jahanzaib-iqbal-dev:
@christian-bromann this is a one-word grammar fix in the Groq provider JSDoc — `'This limits ensures'` → `'This limit ensures'`. Same pattern as #11066 and #11067 merged recently. No changeset needed for a comment-only fix. Would appreciate your review! 🙏

PR #11143, opened 2026-07-02
fix(anthropic): do not send unknown-MIME files as application/pdf documents

### LINKED ISSUE #11139: @langchain/anthropic: files with unknown MIME type sent as application/pdf documents, causing 400
## Summary

When an agent reads a binary file whose extension has no MIME type entry (e.g. `.pem`, `.enc`, `.keystore`), `getMimeType()` returns `""`. Three separate code paths in `@langchain/anthropic` treat an empty string as equivalent to `application/pdf` and encode the file as a `document.source.base64` block. The Anthropic API validates the content against the declared type, finds it is not a valid PDF, and returns:

```
messages.N.content.M.tool_result.content.0.document.source.base64.media_type:
  Input should be 'application/pdf'
```

The run aborts immediately with a 400 and there is no recovery path.

## Package

`@langchain/anthropic` v1.5.1

## Affected source lines

| File | Line | Code |
|------|------|------|
| `src/utils/standard.ts` | 308 | `if (mimeType === "" or mimeType === "application/pdf")` |
| `src/utils/content.ts` | 233 | `if (mime_type === "application/pdf" or mime_type === "")` |
| `src/utils/message_inputs.ts` | 280 | `media_type: contentPart.mimeType ?? "application/pdf"` — passes any non-PDF mimeType straight through into a document block |

## Steps to reproduce

1. Use `@langchain/anthropic` with a `deepagents`-based agent that calls `read_file` on a binary file with an unknown extension (e.g. a `.pem` cert file)
2. The tool returns `{ type: "file", mimeType: "", data: <base64> }`
3. One of the three code paths above encodes it as `document.source.base64` with `media_type: "application/pdf"`
4. The Anthropic API returns a 400 and the run aborts

Reproduced using [OpenWiki](https://github.com/langchain-ai/openwiki) (`openwiki --init`) on a Java/Spring Boot monorepo containing `.pem` certificate files.

## Root cause

All three code paths assume that a file with an unknown/empty MIME type is a PDF. This is incorrect — the file could be any binary (cert, keystore, encrypted config, etc.). Sending non-PDF content with `media_type: "application/pdf"` causes the Anthropic API to reject the message.

rodriguescarson:
Fixes #11139

## Problem

When an agent reads a binary file whose extension has no MIME entry (e.g. `.pem`, `.enc`, `.keystore`), `getMimeType()` returns `""`. Three code paths in `@langchain/anthropic` treated an empty MIME type as equivalent to `application/pdf` and encoded the file as a `document.source.base64` block with `media_type: "application/pdf"`. The Anthropic API validates the content against the declared type and returns:

```
messages.N.content.M.tool_result.content.0.document.source.base64.media_type:
  Input should be 'application/pdf'
```

The run aborts with a 400 and there is no recovery path.

## Root cause

Anthropic base64 **document** sources only accept `media_type: "application/pdf"` (text uses a `text` source, images an `image` content source). An arbitrary binary with an unknown MIME type is not a valid document at all, so labeling it `application/pdf` is incorrect.

## Fix

An empty/unknown MIME type is now treated as a PDF **only when the data actually carries the PDF magic bytes** (`%PDF`, i.e. `JVBERi` once base64-encoded), via a small shared `_looksLikePdf()` helper. This keeps genuinely unlabeled PDFs working while other binaries no longer get mislabeled.

- `utils/standard.ts` — base64 file branch: only route to a PDF document for `application/pdf` or an empty MIME that sniffs as a PDF; other unknown binaries fall through to the existing `Unsupported file mime type` error.
- `utils/content.ts` — same change on the base64 branch of `fromStandardFileBlock`.
- `utils/message_inputs.ts` — previously `media_type = contentPart.mimeType ?? "application/pdf"`, which passed an empty string or any non-PDF MIME straight through into the base64 document block (also a 400). Now it emits a PDF document only for an actual PDF and otherwise raises a clear error.

## Tests

Added two regression tests in `standard_content.test.ts`:
- an unlabeled base64 file with PDF magic bytes still becomes an `application/pdf` document;
- an unknown-MIME binary throws `Unsupported file mime type` instead of being mislabeled.

`pnpm exec vitest run src/tests/standard_content.test.ts` passes (8/8, no type errors). I verified the changed code this way; a few unrelated test files in my local filtered install can't resolve `@langchain/core/*` subpaths (core not built locally) — that's an environment artifact, not from this change.

## Note

The URL document branch (`type: "url"`) still maps an empty MIME to a PDF url source; a remote URL can't be sniffed without fetching, and that path is out of scope for the reported base64 400. Happy to address it separately if you'd like.

### DIFF
--- a/.changeset/anthropic-unknown-mime-not-pdf.md
+++ b/.changeset/anthropic-unknown-mime-not-pdf.md
@@ -1,0 +1,7 @@
+---
+"@langchain/anthropic": patch
+---
+
+fix(anthropic): do not send files with unknown MIME type as application/pdf documents
+
+Files whose MIME type is empty/unknown (e.g. `.pem`, `.keystore`) were encoded as `document.source.base64` blocks with `media_type: "application/pdf"`, which the Anthropic API rejects with a 400. Empty MIME types are now only treated as PDF when the data actually has PDF magic bytes; other unknown binaries raise a clear error instead.
--- a/libs/providers/langchain-anthropic/src/tests/standard_content.test.ts
+++ b/libs/providers/langchain-anthropic/src/tests/standard_content.test.ts
@@ -134,6 +134,45 @@ describe("_formatStandardContent", () => {
     });
   });
 
+  it("treats an unlabeled base64 file as a PDF only when it has PDF magic bytes", () => {
+    // "%PDF-1.4" — the PDF header, so an empty MIME type should resolve to a PDF.
+    const pdfData = Uint8Array.from([
+      0x25, 0x50, 0x44, 0x46, 0x2d, 0x31, 0x2e, 0x34,
+    ]);
+    const fileBlock: ContentBlock.Multimodal.File = {
+      type: "file",
+      data: pdfData,
+      mimeType: "",
+    };
+
+    const [content] = _formatStandardContent(
+      createAnthropicMessage([fileBlock])
+    );
+
+    expect(content).toMatchObject({
+      type: "document",
+      source: {
+        type: "base64",
+        data: Buffer.from(pdfData).toString("base64"),
+        media_type: "application/pdf",
+      },
+    });
+  });
+
+  it("throws instead of mislabeling an unknown-MIME binary as a PDF", () => {
+    // Arbitrary binary (e.g. a .pem/.keystore) with no MIME type must not be
+    // sent as an application/pdf document — that yields a 400 from the API.
+    const fileBlock: ContentBlock.Multimodal.File = {
+      type: "file",
+      data: Uint8Array.from([0x00, 0x01, 0x02, 0x03]),
+      mimeType: "",
+    };
+
+    expect(() =>
+      _formatStandardContent(createAnthropicMessage([fileBlock]))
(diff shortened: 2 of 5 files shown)


JacobWeyer:
Also running into this when trying to use Openwiki

colifran:
Thanks @rodriguescarson  - I'll look deeper at this one tomorrow.

@JacobWeyer ya the openwiki one should be resolved - that was an issue in deepagentsjs that was fixed today.

rodriguescarson:
Hey @colifran, friendly ping on this one. Still mergeable with no conflicts, and @JacobWeyer hit the same 400 above from unknown-MIME files being sent as PDF documents. Happy to rebase or adjust anything if that helps it land.

rodriguescarson:
Hi @colifran, checking in on this one. It is still mergeable with green CI, and @JacobWeyer confirmed the same 400 above, so it is hitting more than one person.

The change stops files with an unknown MIME type from being sent as application/pdf, which is what makes the Anthropic API return the 400. Let me know if you want any changes and I will turn them around quickly.

PR #11153, opened 2026-07-04
fix(core): bind splitText when trimMessages receives a TextSplitter instance

### LINKED ISSUE #11152: `trimMessages` throws `TypeError` when `textSplitter` is a `TextSplitter` instance (unbound `splitText`)
### Checked other resources

### Example Code

```typescript
import { trimMessages, HumanMessage } from "@langchain/core/messages";
import { RecursiveCharacterTextSplitter } from "@langchain/textsplitters";

const splitter = new RecursiveCharacterTextSplitter({ chunkSize: 10, chunkOverlap: 0 });

await trimMessages([new HumanMessage("aaaa bbbb cccc dddd eeee ffff gggg hhhh")], {
  maxTokens: 4,
  tokenCounter: (msgs) =>
    msgs.reduce((n, m) => n + String(m.content).split(/\s+/).filter(Boolean).length, 0),
  strategy: "first",
  allowPartial: true,
  textSplitter: splitter, // passing the splitter object, as the JSDoc documents
});
```

```
TypeError: Cannot read properties of undefined (reading '_splitText')
    at _firstMaxTokens (.../langchain-core/src/messages/transformers.ts)
```

(The exact property depends on the splitter - `RecursiveCharacterTextSplitter` hits `this._splitText`
first; a `CharacterTextSplitter` or a custom splitter hits whatever `this.*` its `splitText` uses.)

### Description

`trimMessages` documents `textSplitter` as accepting either a function **or** an object with a
`splitText` method, the JSDoc on `TrimMessagesFields.textSplitter` explicitly names
`BaseDocumentTransformer`, which is exactly what the splitters in `@langchain/textsplitters`
(`RecursiveCharacterTextSplitter`, `CharacterTextSplitter`, …) are.

* I'm trying to use `trimMessages` with `allowPartial: true` and a real `TextSplitter` instance so
  long messages get split instead of dropped whole.
* I expect it to trim/split the message and return the result (the same as when I pass a plain
  splitter function).
* Instead it throws `TypeError: Cannot read properties of undefined` as soon as the partial-splitting
  path runs.

Looking at `_trimMessagesHelper` in `libs/langchain-core/src/messages/transformers.ts`, the object
branch extracts the method **without binding it**:

```ts
if ("splitText" in textSplitter) {
  textSplitterFunc = textSplitter.splitText;   // <-- `this` is lost
}
```

Every real `splitText` implementation relies on `this` (`this.separators`, `this._splitText`,
`this.mergeSplits`, …), so when `_firstMaxTokens` later calls `textSplitterFunc(text)` it runs with
`this === undefined` and throws. Passing a **pre-bound** function
(`splitter.splitText.bind(splitter)`) works, which confirms the lost binding is the whole cause.

For reference, Python's `trim_messages` does the same extraction, but Python bound methods keep the
receiver, so this only bites the JS port.

**Suggested direction:** bind the method to its splitter in that branch, e.g.
`textSplitterFunc = textSplitter.splitText.bind(textSplitter);`. The function-splitter branch is
unaffected. Happy to open a small PR with a unit test covering the object-splitter path (there isn't
one today).

### System Info

@langchain/core: 1.2.1
@langchain/textsplitters: (current main)
platform: linux (reproduced in a node:24 Docker container; also affects the current main tree)
Node: v24.16.0
pnpm: 10.14.0

parveshsaini:
## Description

`trimMessages` accepts a `textSplitter` that is either a function **or** an object with a
`splitText` method - the JSDoc on `TrimMessagesFields.textSplitter` explicitly documents the
`BaseDocumentTransformer` (object) form, which is what every real splitter in
`@langchain/textsplitters` (`RecursiveCharacterTextSplitter`, `CharacterTextSplitter`, …) is.

When such an object is passed, the helper extracts the method **unbound**:

```ts
// libs/langchain-core/src/messages/transformers.ts
if ("splitText" in textSplitter) {
  textSplitterFunc = textSplitter.splitText;   // <- loses `this`
}
```

Every real `splitText` implementation uses `this` (`this.separators`, `this._splitText`,
`this.mergeSplits`, …), so the later call in `_firstMaxTokens` runs with `this === undefined`
and throws. This is hit whenever `allowPartial: true` falls back to string splitting:

```
TypeError: Cannot read properties of undefined (reading '_splitText')
```

The one-line fix binds the method to its splitter:

```ts
textSplitterFunc = textSplitter.splitText.bind(textSplitter);
```

The Python `trim_messages` does the same extraction, but Python bound methods keep the receiver,
so the bug is specific to the JS port. The function-splitter branch is unaffected.

## Reproduction

```ts
import { trimMessages, HumanMessage } from "@langchain/core/messages";
import { RecursiveCharacterTextSplitter } from "@langchain/textsplitters";

const splitter = new RecursiveCharacterTextSplitter({ chunkSize: 10, chunkOverlap: 0 });

await trimMessages([new HumanMessage("aaaa bbbb cccc dddd eeee ffff gggg hhhh")], {
  maxTokens: 4,
  tokenCounter: (msgs) => msgs.reduce((n, m) => n + String(m.content).split(/\s+/).length, 0),
  strategy: "first",
  allowPartial: true,
  textSplitter: splitter, // documented object form -> throws before this change
});
```

Before this change the call above throws `TypeError: Cannot read properties of undefined`.
Passing `splitter.splitText.bind(splitter)` (a pre-bound function) already worked, confirming the
lost binding is the sole cause.

## Test

Adds a unit test to `libs/langchain-core/src/messages/tests/message_utils.test.ts` that passes a
minimal `this`-dependent `TextSplitter`-style object (core cannot depend on
`@langchain/textsplitters`, so a stand-in splitter is used). Verified in a clean `node:24` container:

- **without** the fix: the new test fails with
  `TypeError: Cannot read properties of undefined (reading 'separator')`
- **with** the fix: the full file passes (24/24), `oxlint` reports 0 errors, `oxfmt --check` is clean.

Fixes #11152

Twitter: https://x.com/PrvshHasNoLife


colifran (dismissed):
Thank you @parveshsaini! Looks good.

colifran (approved):

PR #11154, opened 2026-07-05
fix(langchain): recover from invalid tool args when middleware is present

rchasman:
## Problem

`ToolNode`'s documented default for `handleToolErrors` says it "Catches only `ToolInvocationError` (invalid arguments from model) and converts to ToolMessage" so the model can see the validation error and retry.

That holds without middleware. But when any `wrapToolCall` middleware is registered, the `ToolInvocationError` thrown by the base handler propagates through the middleware call, gets wrapped in `MiddlewareError` (once per middleware layer), and hits this branch in `ToolNode.#handleError`:

```ts
if (isMiddlewareError && this.handleToolErrors !== true) {
  throw error;
}
```

so it is re-thrown as fatal before the documented default ever applies. One malformed tool call from the model kills the entire agent run.

This bites any `createAgent`/DeepAgents user in practice: DeepAgents always registers a `wrapToolCall` middleware, so agents built on it (e.g. OpenWiki) hard-abort whenever the model emits a tool call with a missing required field, instead of retrying. Repro observed repeatedly with claude-sonnet-5 emitting a todo item without `status` in OpenWiki 0.0.1: the run dies with `Error invoking tool 'write_todos' ... Received tool input did not match expected schema`.

## Fix

- Add `findToolInvocationError()` that walks the `cause` chain (the `MiddlewareError` wrapping preserves the original error as `cause`, and stacked middleware can nest wrappers).
- Exempt `ToolInvocationError` from the middleware bubble-up, so the existing default/function handlers convert it to a `ToolMessage` exactly as they already do in the no-middleware path.
- Pass the unwrapped error to a function-valued `handleToolErrors` so `instanceof ToolInvocationError` checks in user handlers keep working.

Genuine middleware errors are unaffected: they still bubble up unless `handleToolErrors === true` (the existing test "should still wrap non-GraphBubbleUp errors in MiddlewareError" still passes).

## Test

Added a regression test: a passthrough `wrapToolCall` middleware plus a model that first emits invalid tool args. Before the fix the invoke rejects with `MiddlewareError`; after, the model receives a ToolMessage with the validation error and recovers. Verified red without the fix, green with it; all 331 agents unit tests pass.

### DIFF
--- a/.changeset/spotty-lamps-feel.md
+++ b/.changeset/spotty-lamps-feel.md
@@ -1,0 +1,5 @@
+---
+"langchain": patch
+---
+
+fix(langchain): recover from invalid tool args when middleware is present. ToolInvocationError is now converted to a ToolMessage (per the documented handleToolErrors default) even when a wrapToolCall middleware is registered, instead of fatally aborting the agent run.
--- a/libs/langchain/src/agents/nodes/ToolNode.ts
+++ b/libs/langchain/src/agents/nodes/ToolNode.ts
@@ -104,6 +104,26 @@ const isSendInput = (
   typeof input === "object" && input != null && "lg_tool_call" in input;
 
 /**
+ * Find a ToolInvocationError in an error's cause chain.
+ *
+ * Errors thrown by the base handler inside wrapToolCall middleware are
+ * wrapped in MiddlewareError (possibly nested once per middleware layer),
+ * with the original error preserved as `cause`.
+ */
+function findToolInvocationError(
+  error: unknown
+): ToolInvocationError | undefined {
+  let current: unknown = error;
+  while (current instanceof Error) {
+    if (current instanceof ToolInvocationError) {
+      return current;
+    }
+    current = current.cause;
+  }
+  return undefined;
+}
+
+/**
  * Default error handler for tool errors.
  *
  * This is applied to errors from baseHandler (tool execution).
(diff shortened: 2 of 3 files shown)


michalmodus:
Independent confirmation from production — thanks for the precise root-cause writeup, it matches exactly what we traced.

We hit this with `claude-sonnet-5` served through an OpenAI-compatible gateway backed by Bedrock (langchain sees it as `ChatOpenAI`, tool-call ids are `toolu_bdrk_*`, and args arrive as an OpenAI-style `function.arguments` JSON string). deepagents' always-on `wrapToolCall` (filesystem/token-eviction) routes every `ToolInvocationError` through the `isMiddlewareError && this.handleToolErrors !== true` branch, so one malformed tool call aborts the whole run after many successful steps.

Beyond the `write_todos` (missing `status`) case in your description, we saw the same fatal-abort for other tools where the model returned schema-valid JSON that was just missing/mistyping a required field:

- a report-generation tool called with all its metadata but omitting the required body field (`Either html, react_body, or body_html is required`);
- `read_file` called with a stringified array (`"[209,483,491]"`) for the numeric `offset` param → `Invalid input: expected number, received NaN`.

All are recoverable "fix your args and retry" errors that only become fatal because a `wrapToolCall` middleware is registered. The `findToolInvocationError()` cause-chain unwrap + exempting `ToolInvocationError` from the middleware bubble-up is exactly the right fix. 👍

rchasman:
Thanks for the confirmation and the extra failure shapes - those two examples (missing required field with otherwise-valid JSON, stringified array for a numeric param) make the case well that this isn't specific to one tool or provider: any schema-invalid args from any model become a fatal abort once a wrapToolCall middleware is present. The regression test in this PR covers the missing-required-field shape, and the fix handles all of them the same way since they all surface as ToolInvocationError. Changeset added per the bot.

PR #11157, opened 2026-07-06
fix(openai): infer assistant role for tool call deltas

Kingtous:
## Description

This PR fixes OpenAI Chat Completions streaming delta conversion when an OpenAI-compatible provider emits assistant-only fields before sending `delta.role`.

Currently, `convertCompletionsDeltaToBaseMessageChunk` only returns an `AIMessageChunk` when `delta.role` or `defaultRole` is `"assistant"`. Some OpenAI-compatible streaming responses may emit fields such as `tool_calls`, `function_call`, `reasoning_content`, or `audio` before the initial assistant role delta. In that case, the chunk is converted into a generic `ChatMessageChunk`, causing streamed tool call information to be missed instead of being captured as `tool_call_chunks`.

This change conservatively infers the role as `"assistant"` only when assistant-only delta fields are present and no explicit/default role is available.

## Change

- Detect assistant-only streaming delta fields:
  - `tool_calls`
  - `function_call`
  - `reasoning_content`
  - `audio`
- Infer `role = "assistant"` for those deltas when `delta.role` and `defaultRole` are both unset.
- Preserve existing behavior for explicit roles and previously established default roles.

## Context

This helps OpenAI-compatible providers whose streaming chunks do not send `role: "assistant"` before tool call deltas. Without this, early tool call chunks can be treated as `ChatMessageChunk` instead of `AIMessageChunk`, leading to lost tool call data during streaming.


Sparky579:
I hit the same bug from a different angle while testing streaming adapters against recorded provider traffic, and I can confirm this fix resolves the roleless case. Checked out the branch against `@langchain/openai` 1.5.8 and the tool calls come through as expected.

There is one variant it does not cover: servers that send `role: ""` rather than omitting the field. `??` only falls back on `null`/`undefined`, so an empty string still reaches the generic `ChatMessageChunk` branch and the tool calls are dropped.

With this PR's change applied:

| First delta | tool_call_chunks |
|---|---|
| `{"tool_calls":[…]}` | ✅ (fixed here) |
| `{"role":null,"tool_calls":[…]}` | ✅ (fixed here) |
| `{"role":"","tool_calls":[…]}` | ❌ still empty |
| `{"role":"assistant","tool_calls":[…]}` | ✅ |

Repro:

```ts
const result = convertCompletionsDeltaToBaseMessageChunk({
  delta: {
    role: "",
    tool_calls: [
      { index: 0, id: "call_1", type: "function",
        function: { name: "get_weather", arguments: '{"city":"BJ"}' } },
    ],
  },
  rawResponse: { id: "c", choices: [{ index: 0, delta: {}, finish_reason: null }] } as any,
}) as AIMessageChunk;

// tool_call_chunks is [] — the call is silently lost
```

Plain text is unaffected either way, which is what makes this easy to miss: the stream looks healthy and only tool calls disappear.

Switching the coalescing to `||` covers both shapes and keeps the assistant-only-field inference intact:

```ts
const role =
  delta.role || defaultRole || (isAssistantDelta ? "assistant" : undefined);
```

An explicit `"tool"` / `"system"` / `"user"` role still wins, since those are truthy.

The empty-string shape is not hypothetical. It is what [anomalyco/opencode#28427](https://github.com/anomalyco/opencode/issues/28427) reports from OpenAI-compatible servers; there it surfaced as a validation error in a different SDK, but the same responses take this code path here.

Happy to open a follow-up PR with a test for the empty-string case if you would rather keep this one focused, or you are welcome to fold it in.

Sparky579:
Filed #11468 for the `role: ""` variant so it has somewhere to live independently of this PR, with a runnable `ChatOpenAI` reproduction. It references this PR as the existing fix for the roleless case.

OffsiteGuru:
Real-world case for this PR: **Halogen Flash Server** streams chat completions with no `role` in any chunk, and **LibreChat 0.8.8-rc2** (`@langchain/openai` 1.5.8 via `@librechat/agents`; same code in 1.5.13) silently drops every streamed tool call. The model announces a tool, nothing runs, no error. A proxy that inserts only `{"role":"assistant"}` as the first delta fixes it.

`delta.role` is optional in OpenAI's schema (`ChatCompletionStreamResponseDelta` has no `required`), so servers are entitled to omit it.

Halogen's stream order: `reasoning_content` deltas, then `content`, then `tool_calls`, none with `role`. Under this PR the `content` chunks still become `ChatMessageChunk` while the others become `AIMessageChunk`. Would it be simpler to default to `"assistant"` whenever both `delta.role` and `defaultRole` are absent? A chat completion only has one possible author, and that also covers #11468 if `""` is treated as absent (`||`).

PR #11160, opened 2026-07-07
fix(openai): pass request options for streaming responses

### LINKED ISSUE #11159: OpenAI client does not pass custom request options for streaming responses
### Checked other resources

### Example Code

When calling `client.stream(input, options)` on a `ChatOpenAI` client, the [request options](https://github.com/langchain-ai/langchainjs/blob/1a1fe3ca9cd779abda3baf23529434dca601f00b/libs/providers/langchain-openai/src/chat_models/base.ts#L88) property does not get passed through to the underlying SDK. It works as expected for `.invoke()` (non-streaming).

```typescript

const client = new ChatOpenAI({
  configuration: {
    async fetch(input, init) {
      const headers = new Headers(init?.headers);
      if (!headers.has("x-custom-header")) {
        setTimeout(() => {
          throw new Error("Custom header not set");
        });
      }
      return fetch(input, init);
    },
  },
});

// works as expected
await client.invoke("Hello", {
  options: {
    headers: { "x-custom-header": "1" },
  },
});
console.log("invoke() successful");

// does not pass the header
const stream = await client.stream("Hello", {
  options: {
    headers: { "x-custom-header": "1" },
  },
});
for await (const _ of stream) {
}
console.log("stream() successful");
```

_No response_

### Description

- I am trying to pass a request-specific custom header to `ChatOpenAI.stream()`
- The header does not get passed to the request

### System Info

not applicable

andrew0:
Fixes #11159

### DIFF
--- a/libs/providers/langchain-openai/src/chat_models/completions.ts
+++ b/libs/providers/langchain-openai/src/chat_models/completions.ts
@@ -346,7 +346,10 @@ export class ChatOpenAICompletions<
       stream: true as const,
     };
 
-    const streamIterable = await this.completionWithRetry(params, options);
+    const streamIterable = await this.completionWithRetry(params, {
+      signal: options.signal,
+      ...options.options,
+    });
     const shouldStreamUsage = this.streamUsage ?? options.streamUsage;
 
     const abortableStream = async function* (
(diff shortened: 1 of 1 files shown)


wbrunovieira (commented):
Matches the non-streaming path (`completions.ts:222-232`), which already forwards `{ signal, ...options.options }` to `completionWithRetry` — so streaming is consistent now, and both methods are covered. Also stops the old code from passing the whole `ParsedCallOptions` object where only `RequestOptions` is expected. Looks correct to me.

One suggestion: worth adding a regression test? CONTRIBUTING asks for a unit test on new logic, and it's easy to assert — spy on `completionWithRetry` and check a custom request option reaches it on the streaming path. Would guard against a silent regression.

Thanks!

PR #11161, opened 2026-07-07
fix(langchain): keep structured-output tool name stable across model requests

### LINKED ISSUE #11151: createAgent structured-output tool is renamed `extract-<n>` on every model call, defeating Anthropic prompt caching
## Description

When `createAgent` is given a `responseFormat` schema without an explicit name (the common case: a plain Zod schema, whose JSON-schema conversion has no `title`), the structured-output tool is named from a module-global counter:

```ts
// libs/langchain/src/agents/responses.ts
let bindingIdentifier = 0; // L46
// ...
function getFunctionName(name?: string) {
  // L102
  return name ?? `extract-${++bindingIdentifier}`; // L103
}
```

The tool is not created once per agent: `AgentNode`'s per-call `baseHandler` (`libs/langchain/src/agents/nodes/AgentNode.ts:368`) invokes `#getResponseFormat` (defined at :173, called at :376) on **every model request**, which calls `transformResponseFormat` → `ToolStrategy.fromSchema` → `getFunctionName` again. So in a single agent loop the model sees a different tool list on every call:

```
call 1: tools=[..., extract-89]
call 2: tools=[..., extract-90]
call 3: tools=[..., extract-91]
```

(The counter is also process-global, so concurrent agents interleave increments.)

## Why this matters

Anthropic's prompt cache hashes **tool definitions + system + messages up to the cache breakpoint**. Because the tool block changes on every call, every model call in an agent loop is a guaranteed cache miss: `cache_creation_input_tokens` equals the full prefix on every call and `cache_read_input_tokens` is always 0 — even when the user has correctly marked a large static prefix with `cache_control` (or uses `anthropicPromptCachingMiddleware`).

Observed in production (LangSmith trace, langchain 1.2.27, agent with a ~50k-token prefix, 5-call loop): 5 × 50,372 cache-write tokens, 0 read. With a stable tool name the same run writes once and reads 4 times (~0.1× input price), roughly a 3× cost reduction for the agent. The same prefix-instability principle degrades Gemini/Vertex implicit caching as well.

## Reproduction (minimal)

```ts
import { z } from "zod";
import { tool } from "langchain";

const agent = createAgent({
  model: "anthropic:claude-sonnet-4-5",
  tools: [
    tool(async () => "ok", {
      name: "noop",
      description: "returns ok",
      schema: z.object({}),
    }),
  ],
  responseFormat: z.object({ answer: z.string() }),
});

// Any run where the model calls `noop` before finishing produces >= 2 model calls.
// Inspect the requests (or a LangSmith trace): the structured-output tool gets a
// different `extract-<n>` name on each model call (the counter is process-global,
// so the exact numbers depend on prior usage; the point is that consecutive calls
// differ). With Anthropic prompt caching enabled, usage shows cache_creation on
// every call and cache_read = 0.
```

Workaround we ship: pass a JSON-schema object with a `title` (used as the tool name in `ToolStrategy.fromSchema`), so the name is stable:

```ts
import { toJsonSchema } from "@langchain/core/utils/json_schema";
responseFormat: { ...toJsonSchema(schema), title: "my_output" }
```

## Suggested fixes (either suffices; both are better)

1. **Memoise the response-format strategies per agent** instead of rebuilding them inside the per-call handler — the schema cannot change between loop iterations of one agent.
2. **Derive a stable default name** instead of a global counter, e.g. a fixed `"extract"` (colliding names are impossible: only one unnamed responseFormat exists per agent) or a content hash of the schema.

Note the native/provider strategy already uses a stable `?? "extract"` default (`AgentNode.ts`), so this only affects the **tool** strategy — which is what Anthropic (incl. Claude via Vertex) uses for structured output.

Affected: verified in `langchain` 1.2.27, 1.4.2, and current `main` / 1.5.2 (naming at `libs/langchain/src/agents/responses.ts:103`; per-call regeneration via `baseHandler` at `libs/langchain/src/agents/nodes/AgentNode.ts:368` → `#getResponseFormat`).

bisma-nawaz:
Fixes #11151

## What

Memoizes the structured-output response format inside `AgentNode` so the generated tool keeps a stable name across the model requests of a single agent loop.

## Why

`AgentNode.#getResponseFormat` runs on every model request (via `baseHandler`). When `responseFormat` is a schema with no explicit name, `transformResponseFormat` → `ToolStrategy.fromSchema` derives the tool name from a module-global counter (`extract-${++bindingIdentifier}` in `libs/langchain/src/agents/responses.ts`). Rebuilding on every call therefore produces a new tool name (`extract-89`, `extract-90`, `extract-91`, …) for the same schema, so the model sees a different tool definition on each request.

Because tool definitions are part of the cached prefix for providers that support prompt caching, an unstable tool name turns every request in the agent loop into a cache miss (`cache_read_input_tokens` stays 0). The instability degrades implicit caching on other providers for the same reason.

## How

The strategies for a given `(model, responseFormat)` pair can't change between iterations of one agent loop, so `#getResponseFormat` now caches its result keyed on the `model` and `responseFormat` references it was built from and returns the cached value when both are unchanged. A middleware that swaps the model or response format still produces a fresh result (the references differ), so behavior is unchanged in that case. I traced the rebuild path through `AgentNode.#getResponseFormat` → `transformResponseFormat` → `ToolStrategy.fromSchema` and confirmed the tool name is the only value that varies between otherwise-identical calls. I kept the `extract-` naming scheme intact since it's relied on elsewhere (e.g. routing in `ReactAgent` and `piiRedaction` check `name.startsWith("extract-")`).

## Testing

Added a unit test in `libs/langchain/src/agents/tests/responses.test.ts` that spies on `bindTools` and asserts the structured-output tool is bound with the same generated name on every model request of a multi-step agent loop. It fails before the change (two different `extract-<n>` names) and passes after.

```
$ npx vitest run src/agents/tests/responses.test.ts
 Test Files  1 passed (1)
      Tests  37 passed (37)

$ npx vitest run src/agents/tests/reactAgent.test.ts
 Test Files  1 passed (1)
      Tests  35 passed (35)
```

Also ran the full `src/agents/tests` unit suite (25 files, 390 tests) — all pass. `oxlint` and `oxfmt --check` are clean on the changed files.

### DIFF
--- a/.changeset/stable-extract-tool-name.md
+++ b/.changeset/stable-extract-tool-name.md
@@ -1,0 +1,5 @@
+---
+"langchain": patch
+---
+
+fix(langchain): keep structured-output tool name stable across model requests
--- a/libs/langchain/src/agents/nodes/AgentNode.ts
+++ b/libs/langchain/src/agents/nodes/AgentNode.ts
@@ -146,6 +146,24 @@ export class AgentNode<
   #options: AgentNodeOptions<StructuredResponseFormat, ContextSchema>;
   #systemMessage: SystemMessage;
 
+  /**
+   * Memoized result of {@link AgentNode.#getResponseFormat}, keyed by the
+   * `model` and `responseFormat` references it was derived from.
+   *
+   * `#getResponseFormat` runs on every model request. Recomputing it rebuilds
+   * the structured-output strategies from scratch, and an unnamed schema is
+   * assigned a fresh `extract-<n>` tool name from a module-global counter each
+   * time. That makes the model see a different tool definition on every call,
+   * which defeats provider prompt caching (e.g. Anthropic, Gemini). Reusing the
+   * previous result whenever the inputs are unchanged keeps the tool definition
+   * stable across the agent loop.
+   */
+  #responseFormatCache?: {
+    model: string | LanguageModelLike;
+    responseFormat: ResponseFormatInput | undefined;
+    result: ResponseFormat | undefined;
+  };
+
   constructor(
     options: AgentNodeOptions<StructuredResponseFormat, ContextSchema>
   ) {
@@ -179,6 +197,20 @@ export class AgentNode<
       return undefined;
     }
 
+    /**
+     * Reuse the previously computed response format when neither the model nor
+     * the responseFormat reference has changed, so the structured-output tool
+     * keeps a stable name across model requests within an agent loop.
+     */
+    const cached = this.#responseFormatCache;
+    if (
+      cached &&
+      cached.model === model &&
+      cached.responseFormat === responseFormat
+    ) {
+      return cached.result;
+    }
+
     let resolvedModel: LanguageModelLike | undefined;
     if (isConfigurableModel(model)) {
       resolvedModel = await (
@@ -196,22 +228,19 @@ export class AgentNode<
       resolvedModel
     );
 
+    let result: ResponseFormat | undefined;
     if (strategies.length === 0) {
-      return undefined;
-    }
-
-    /**
-     * we either define a list of provider strategies or a list of tool strategies
-     */
-    const isProviderStrategy = strategies.every(
-      (format) => format instanceof ProviderStrategy
-    );
-
-    /**
-     * Populate a list of structured tool info.
-     */
-    if (!isProviderStrategy) {
-      return {
+      result = undefined;
+    } else if (
+      /**
+       * we either define a list of provider strategies or a list of tool strategies
+       */
+      !strategies.every((format) => format instanceof ProviderStrategy)
+    ) {
+      /**
+       * Populate a list of structured tool info.
+       */
+      result = {
         type: "tool",
         tools: (
           strategies.filter(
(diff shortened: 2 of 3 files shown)


wbrunovieira (commented):
Traced this through — the memoization and the early-return→single-`result` refactor are behavior-preserving, and keeping `extract-` is right (`ReactAgent` routing and `piiRedaction` both key off `startsWith("extract-")`).

One thing I'd check before this closes #11151: the cache is keyed on the `request.model` **reference**, but for a string model (`createAgent({ model: "anthropic:…" })`) `#deriveModel()` calls `initChatModel()` (`AgentNode.ts:327`), which builds a **new `ConfigurableModel` every request** (`universal.ts:1093`) — no memoization. Since `#deriveModel()` runs once per model request, `cached.model === model` never holds across the loop, so a fresh `extract-N` is still generated each request and the prompt-cache miss persists. It only stabilizes when a model *instance* is passed. Does the `bindTools` spy test use a string model or an instance? If it's an instance, it'd go green while the reported (string-model) case is still broken.

If that's real, memoizing here may be the wrong layer — a deterministic `extract-*` name derived from the schema in `responses.ts` (instead of the module-global counter) would be stable regardless of how the model reference churns, while still satisfying the `startsWith("extract-")` consumers. Bigger change / probably its own PR, but might be the more robust fix. Nice catch on the root cause either way.

PR #11167, opened 2026-07-08
fix(langchain): malformed tool input schemas are unrecoverable

colifran:
### Summary
When a `wrapToolCall` middleware is present, a recoverable tool-input error never reaches the model to self-correct and instead the whole run aborts instead.

The root cause is `baseHandler` correctly turns a `ToolInputParsingException` into a `ToolInvocationError`. The `defaultHandleToolErrors` knows how to convert into a retryable `ToolMessage`, but the middleware composer then
rewraps it as a `MiddlewareError`. `#handleError`'s guard re-raises middleware errors unless `handleToolErrors === true` and never looks at .cause, so the recoverable error is treated as fatal. Since middleware always wraps tool calls in deepagents, the self-correction path is unreachable.

This PR unwraps that one case, i.e., if a `MiddlewareError`'s .cause is a `ToolInvocationError` we treat the cause as the error passed to `handleToolErrors` so the existing recovery path applies. Genuine middleware failures stay fatal by default.

This has been heavily reported downstream in openwiki where long --init runs die the first time the model fumbles a tool call.

### Tests

Added two cases to `ToolNode.test.ts`. Both include a `wrapToolCall` middleware that rewraps errors as `MiddlewareError`

### DIFF
--- a/libs/langchain/src/agents/nodes/ToolNode.ts
+++ b/libs/langchain/src/agents/nodes/ToolNode.ts
@@ -247,20 +247,23 @@ export class ToolNode<
 
     /**
      * A recoverable tool error (e.g. tool-input schema validation) can be
-     * rewrapped by the middleware composer as a {@link MiddlewareError} with the
-     * original error on `.cause`. Unwrap it so the intended `handleToolErrors`
-     * self-correction path still applies, while genuine middleware errors stay
-     * fatal by default.
+     * rewrapped as a {@link MiddlewareError} with the original error on `.cause`
+     * — once per `wrapToolCall` middleware, so it may be nested several layers
+     * deep. Walk the cause chain to the root; if it's a {@link ToolInvocationError},
+     * unwrap it so the intended `handleToolErrors` self-correction path still
+     * applies. Genuine middleware errors stay fatal by default.
      */
     let effectiveError = error;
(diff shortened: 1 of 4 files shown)


libs/langchain/src/agents/nodes/ToolNode.ts
```diff
@@ -245,46 +245,67 @@ export class ToolNode<
...
+     * applies. Genuine middleware errors stay fatal by default.
+     */
+    let effectiveError = error;
+    let errorFromMiddleware = isMiddlewareError;
+    if (isMiddlewareError) {
+      let unwrapped: unknown = error;
+      while (unwrapped instanceof MiddlewareError) {
+        unwrapped = unwrapped.cause;
+      }
+      if (unwrapped instanceof ToolInvocationError) {
```
christian-bromann: Could we use `isInstance` instead of `instanceof` (missing in `ToolInvocationError`, maybe worth branding it)?

christian-bromann (dismissed):
Otherwise LGTM 👍

libs/langchain/src/agents/nodes/ToolNode.ts
colifran: Updated.

PR #11169, opened 2026-07-09
fix(openai): emit `output_text` for assistant content in responses input

colifran:
### Summary

The openai responses converter typed all text content as `input_text` regardless of role. For example, an assistant message became `{ role: "assistant", content: [{ type: "input_text", … }] }` and the API rejected it:

```
400 Invalid value: 'input_text'. Supported values are: 'output_text' and 'refusal'.
```

This bug only affects the native openai responses api and only on multi-turn runs since the bad `input_text` is on a prior assistant turn replayed as input meaning a single first turn never triggers it.

This PR makes the text part role-aware, i.e., `output_text` for assistant, `input_text` otherwise, this matches Python's `_construct_responses_api_input`.

### Tests

Added `convertStandardContentMessageToResponsesInput` regression tests verifying that an assistant message converts to `output_text` and a user message stays `input_text`.

### DIFF
--- a/.changeset/mighty-hotels-end.md
+++ b/.changeset/mighty-hotels-end.md
@@ -1,0 +1,5 @@
+---
+"@langchain/openai": patch
+---
+
+fix(openai): emit output_text for assistant content in responses input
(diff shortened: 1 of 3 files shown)


christian-bromann (approved):
LGTM 👍

PR #11171, opened 2026-07-09
fix(core): coerce string v1 AIMessage content to text blocks

christian-bromann:
## Summary
- Normalize string (and other non-array) `content` to content-block arrays when constructing `AIMessage` with `response_metadata.output_version === "v1"`.
- Prevents `TypeError: initParams.contentBlocks.push is not a function`, which shows up when LangGraph stream/history payloads still carry string `content` under v1 metadata (e.g. Fleet agent creation).
- Adds regression coverage for string content with and without `tool_calls`.

No existing open PR covers this. Related merged work: #11047 (preserve existing `contentBlocks`), which does not handle string `content`.

## Test plan
- [x] `pnpm test src/messages/tests/ai.test.ts` in `libs/langchain-core` (17 passed)
- [ ] Confirm agent-creation chat no longer throws after bumping `@langchain/core` in langchainplus

### DIFF
--- a/.changeset/fix-aimessage-v1-string-content.md
+++ b/.changeset/fix-aimessage-v1-string-content.md
@@ -1,0 +1,9 @@
+---
+"@langchain/core": patch
+---
+
+fix(core): coerce string v1 AIMessage content to text blocks
+
+Prevent `contentBlocks.push is not a function` when constructing an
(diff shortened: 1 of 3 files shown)


colifran (approved):
Nice!

PR #11179, opened 2026-07-09
feat(google): surface Vertex trafficType and modelVersion on response_metadata

### LINKED ISSUE #11197: Surface Vertex trafficType and modelVersion on response_metadata
### Feature request

Vertex AI returns two useful fields on every `generateContent`/`streamGenerateContent` response that the LangChain Google SDKs currently drop:

- **`usageMetadata.trafficType`** — `PROVISIONED_THROUGHPUT`, `ON_DEMAND`, or `ON_DEMAND_PRIORITY`. Tells you whether the request actually consumed Provisioned Throughput (PTU) capacity or spilled to on-demand.
- **`modelVersion`** — the exact served model version, which matters when PTU is version-bound.

Today a caller can only see token counts via `usage_metadata`. There's no supported way to observe PTU consumption from the client — you have to intercept the raw HTTP body (e.g. via the `google-response-*` custom event) to get `trafficType` at all, and on **streaming** responses even that is awkward because the field only arrives on the terminal chunk.

### Motivation

We run Gemini on Vertex Provisioned Throughput. After pinning traffic to a PTU-eligible region we needed to confirm requests were actually consuming PTU (not silently billing on-demand). The only signal is `trafficType`, and it isn't reachable from the `AIMessage` — non-streaming or streaming.

### Proposed change

Expose both fields on the returned message's `response_metadata` on both streaming and non-streaming paths, in both `@langchain/google` and `@langchain/google-common`. Omit when absent (e.g. AI Studio responses without `trafficType`), so no shape change for non-Vertex callers.

### Example

```ts
const model = new ChatGoogle({ platformType: "gcp", location: "us", model: "gemini-3.1-flash-lite" });
const res = await model.invoke("hi");
console.log(res.response_metadata.traffic_type);  // "PROVISIONED_THROUGHPUT"
console.log(res.response_metadata.model_version); // "gemini-3.1-flash-lite"
```

### PR

PR #11179 implements this.

saarRozenbaum:
Closes #11197

## Summary

Vertex AI returns two useful fields on every generate response that the SDK currently drops:

- **`usageMetadata.trafficType`** — `PROVISIONED_THROUGHPUT`, `ON_DEMAND`, or `ON_DEMAND_PRIORITY`. Tells you whether the request actually consumed Provisioned Throughput (PTU) capacity or spilled to on-demand.
- **`modelVersion`** — the exact served model version, which matters when PTU is version-bound.

Today a caller can only see token counts via `usage_metadata`. There's no supported way to observe PTU consumption from the client — you have to intercept the raw HTTP body (e.g. via the `google-response-*` custom event) to get `trafficType` at all, and on **streaming** responses even that is awkward because the field only arrives on the terminal chunk.

This surfaces both fields on the returned message's `response_metadata`, on both streaming and non-streaming paths, in both `@langchain/google` and `@langchain/google-common`.

## Motivation

We run Gemini on Vertex Provisioned Throughput. After pinning traffic to a PTU-eligible region we needed to confirm requests were actually consuming PTU (not silently billing on-demand). The only signal is `trafficType`, and it wasn't reachable from the `AIMessage` — non-streaming or streaming. This change makes it a first-class part of the response.

## Changes

- **`@langchain/google`**
  - `_generate` (non-streaming): add `traffic_type` + `model_version` to `message.response_metadata`. The path already read `trafficType` to derive `serviceTier`; now it's also exposed directly.
  - `_streamResponseChunks` (streaming): add them to each chunk's `response_metadata` when present (terminal chunk carries `usageMetadata`), so they land on the final concatenated `AIMessage`.
- **`@langchain/google-common`**
  - `responseToGenerationInfo` + `combineGenerations`: same fields on `generationInfo` and the message's `response_metadata`.

Both fields are omitted when absent (e.g. AI Studio responses without `trafficType`), so no shape change for non-Vertex callers.

## Test plan

- [x] `@langchain/google`: new mock fixtures + tests — `invoke surfaces Vertex traffic_type and model_version on response_metadata`, and `stream surfaces ... on the final concatenated chunk`.
- [x] `@langchain/google-common`: new fixture + `Invoke surfaces Vertex traffic_type and model_version on response_metadata`.
- [x] Existing usage-metadata and streaming tests unchanged.

## Example

```ts
const model = new ChatGoogle({ platformType: "gcp", location: "us", model: "gemini-3.1-flash-lite" });
const res = await model.invoke("hi");
console.log(res.response_metadata.traffic_type);  // "PROVISIONED_THROUGHPUT"
console.log(res.response_metadata.model_version); // "gemini-3.1-flash-lite"

// streaming
let final;
for await (const c of await model.stream("hi")) final = final ? final.concat(c) : c;
console.log(final.response_metadata.traffic_type); // "PROVISIONED_THROUGHPUT"
```

### DIFF
--- a/.changeset/vertex-traffic-type.md
+++ b/.changeset/vertex-traffic-type.md
@@ -1,0 +1,6 @@
+---
+"@langchain/google-common": patch
+"@langchain/google": patch
+---
+
+feat(google): surface Vertex trafficType and modelVersion on response_metadata
--- a/libs/providers/langchain-google-common/src/tests/chat_models.test.ts
+++ b/libs/providers/langchain-google-common/src/tests/chat_models.test.ts
@@ -3456,6 +3456,23 @@ test("Invoke usage_metadata maps token details", async () => {
   expect(result.usage_metadata?.output_token_details?.reasoning).toBe(4);
 });
 
+test("Invoke surfaces Vertex traffic_type and model_version on response_metadata", async () => {
+  const record: Record<string, unknown> = {};
+  const projectId = mockId();
+  const model = new ChatGoogle({
+    authOptions: {
+      record,
+      projectId,
+      resultFile: "chat-traffic-type-mock.json",
+    },
+  });
+
+  const result = await model.invoke("Hello?");
+
+  expect(result.response_metadata?.traffic_type).toBe("PROVISIONED_THROUGHPUT");
+  expect(result.response_metadata?.model_version).toBe("gemini-2.0-flash-001");
+});
+
 test("Stream usage_metadata includes cache_read", async () => {
   const record: Record<string, unknown> = {};
   const projectId = mockId();
--- a/libs/providers/langchain-google-common/src/tests/data/chat-traffic-type-mock.json
+++ b/libs/providers/langchain-google-common/src/tests/data/chat-traffic-type-mock.json
@@ -1,0 +1,25 @@
+{
+  "candidates": [
+    {
+      "content": {
+        "role": "model",
+        "parts": [
+          {
+            "text": "OK"
+          }
+        ]
+      },
+      "finishReason": "STOP"
+    }
+  ],
+  "modelVersion": "gemini-2.0-flash-001",
+  "promptFeedback": {
+    "safetyRatings": []
+  },
+  "usageMetadata": {
+    "promptTokenCount": 10,
+    "candidatesTokenCount": 4,
+    "totalTokenCount": 14,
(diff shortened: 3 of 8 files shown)


saarRozenbaum:
Gentle bump — CI is green and this is still mergeable. Happy to rebase if it's gone stale, or to split it if surfacing both fields in one PR is more than you'd like to take at once.

@afirstenberg no rush if you're still catching up; just flagging that it hasn't had a look yet.

PR #11185, opened 2026-07-11
fix(google-genai): preserve Gemini signatures in stream events

### LINKED ISSUE #11181: google-genai: _streamChatModelEvents drops all Gemini thoughtSignatures (and tool-call ids) — gemini-3 tool replays degrade to empty completions via DUMMY_SIGNATURE
### Environment

- `@langchain/google-genai`: 2.2.0 (latest)
- `@langchain/core`: 1.2.2 (latest)
- `@langchain/langgraph`: 1.4.7 (triggers the path via `StreamProtocolMessagesHandler`, `lc_prefer_chat_model_stream_events = true`)
- Model: `gemini-3.1-pro-preview` (Vertex publisher API behaves the same)

### Summary

`ChatGoogleGenerativeAI`'s native `_streamChatModelEvents` (the path core selects whenever a callback handler prefers `ChatModelStreamEvent`s — e.g. langgraph's v3 message-streaming protocol) **drops every Gemini `thoughtSignature`** (both thought parts and functionCall parts) and also loses the generated tool-call ids. `convertGoogleGenAIStream` never copies `thoughtSignature`, and the message assembled by `ChatModelStream._assembleMessage` is built from events alone, so nothing survives into `additional_kwargs.__gemini_function_call_thought_signatures__`.

On the next turn the serializer can't find a signature for the functionCall part and falls back to the hardcoded `DUMMY_SIGNATURE` bypass token (`utils/common.ts`). **Gemini 3 requires the real signature back**: replaying a functionCall without it is a hard 400 from the API — "Function call is missing a thought_signature in functionCall parts" — and the DUMMY bypass avoids the 400 at the cost of documented degraded behaviour. In our production traffic that degradation is consistent **empty completions (0 output tokens, finishReason STOP)** on every tool-call follow-up turn.

The legacy `_generate` / `_streamResponseChunks` paths handle all of this correctly (signature maps on `additional_kwargs`) — only the new events path loses it.

### Reproduction (no network needed — stubbed SSE)

```js
import { HumanMessage, ToolMessage } from '@langchain/core/messages';
import { ChatModelStream } from '@langchain/core/language_models/stream';
import { ChatGoogleGenerativeAI } from '@langchain/google-genai';

const sse = (objs) => new Response(objs.map((o) => `data: ${JSON.stringify(o)}\r\n\r\n`).join(''),
  { status: 200, headers: { 'content-type': 'text/event-stream' } });

globalThis.fetch = async () => sse([{
  candidates: [{
    content: { parts: [
      { text: 'planning', thought: true, thoughtSignature: 'THOUGHT_SIG' },
      { functionCall: { name: 'get_weather', args: { city: 'Paris' } }, thoughtSignature: 'FC_SIG' },
    ], role: 'model' },
    finishReason: 'STOP', index: 0,
  }],
  usageMetadata: { promptTokenCount: 10, candidatesTokenCount: 5, totalTokenCount: 15 },
}]);

const model = new ChatGoogleGenerativeAI({ model: 'gemini-3.1-pro-preview', apiKey: 'test', baseUrl: 'http://127.0.0.1:9' });

const events = [];
const assembled = await new ChatModelStream((async function* () {
  for await (const e of model._streamChatModelEvents([new HumanMessage('hi')], {})) { events.push(e); yield e; }
})());

console.log(JSON.stringify(events).includes('SIG'));   // false — signatures gone from the event stream
console.log(assembled.additional_kwargs);              // {} — no signature map
console.log(assembled.tool_calls);                     // [{ name, args }] — id lost too
```

Observed output on 2.2.0 / 1.2.2:

```
signatures present in emitted events: false
assembled additional_kwargs: {}
assembled tool_calls: [{"type":"tool_call","name":"get_weather","args":{"city":"Paris"}}]
```

And replaying the assembled message (after working around the `reasoning` serializer crash — filed separately) serializes the functionCall with the DUMMY token instead of `FC_SIG`:

```
thoughtSignature sent on replay: ErYCCrMCAdHtim9kOoOkrPiC …  (expected FC_SIG)
```

### Expected

Signatures captured on the legacy path must survive the events path too — either by copying `thoughtSignature` into the emitted content blocks (`reasoning.signature`, and something equivalent for tool_call blocks) or by attaching the signature map to the `message-finish` event metadata so `_assembleMessage` can persist it. Tool-call ids should also round-trip.

### Impact

Any langgraph v3 / stream-events consumer running Gemini 3 with tools gets silently degraded (empty) completions on every follow-up after a tool call. The raw-API 400 shows the signature is mandatory, so there's no correct request the current path can build.

vismaytiwari:
Fixes #11181

## Summary

The native chat model event path in `@langchain/google-genai` was not carrying Gemini thought signatures or tool call IDs into the assembled message. As a result, replaying a streamed Gemini 3 tool call used the fallback dummy signature instead of the signature returned by Gemini.

This change:

- preserves signatures on streamed reasoning blocks;
- assigns a stable ID to each streamed tool call and keeps its signature mapped to that ID in response metadata;
- reads that metadata when replaying the assembled message; and
- avoids serializing the same standardized tool call from both `content` and `tool_calls`.

## Validation

- Reproduced the missing signatures, missing tool call ID, and dummy-signature replay before the change with a stubbed SSE response.
- Manually verified the same provider entrypoint after the change: both signatures survive, the tool call has an ID, and replay sends one function call with the real signature.
- `pnpm --filter @langchain/google-genai test` (77 passed)
- `pnpm --filter @langchain/google-genai build`
- `pnpm format:check`
- `pnpm lint`

### DIFF
--- a/.changeset/calm-tools-remember.md
+++ b/.changeset/calm-tools-remember.md
@@ -1,0 +1,5 @@
+---
+"@langchain/google-genai": patch
+---
+
+Preserve Gemini thought signatures and tool call IDs when using chat model stream events.
--- a/libs/providers/langchain-google-genai/src/utils/common.ts
+++ b/libs/providers/langchain-google-genai/src/utils/common.ts
@@ -510,14 +510,28 @@ export function convertMessageContentToParts(
   if (Array.isArray(message.content)) {
     messageParts.push(
       ...(message.content
-        .map((c) => _convertLangChainContentToPart(c, isMultimodalModel))
+        .map((c) => {
+          if (
+            isAIMessage(message) &&
+            c.type === "tool_call" &&
+            message.tool_calls?.some(
+              (toolCall) => toolCall.id === c.id && toolCall.name === c.name
+            )
+          ) {
+            return undefined;
+          }
+          return _convertLangChainContentToPart(c, isMultimodalModel);
+        })
         .filter((p) => p !== undefined) as Part[])
     );
   }
 
-  const functionThoughtSignatures = message.additional_kwargs?.[
+  const functionThoughtSignatures = (message.additional_kwargs?.[
     _FUNCTION_CALL_THOUGHT_SIGNATURES_MAP_KEY
-  ] as Record<string, string>;
+  ] ??
+    message.response_metadata?.[_FUNCTION_CALL_THOUGHT_SIGNATURES_MAP_KEY]) as
+    | Record<string, string>
+    | undefined;
 
   if (isAIMessage(message) && message.tool_calls?.length) {
     functionCalls = message.tool_calls.map((tc) => {
--- a/libs/providers/langchain-google-genai/src/utils/stream_events.ts
+++ b/libs/providers/langchain-google-genai/src/utils/stream_events.ts
@@ -11,7 +11,11 @@ import type {
   FinishReason,
 } from "@langchain/core/language_models/event";
 import type { ContentBlock, UsageMetadata } from "@langchain/core/messages";
-import { convertUsageMetadata } from "./common.js";
+import { v4 as uuidv4 } from "@langchain/core/utils/uuid";
+import {
+  _FUNCTION_CALL_THOUGHT_SIGNATURES_MAP_KEY,
+  convertUsageMetadata,
+} from "./common.js";
 
 export interface ConvertGoogleGenAIStreamOptions {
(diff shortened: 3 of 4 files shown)


vismaytiwari:
Bump — #11181 is still open and this is passing; glad to add tests or change anything reviewers want.

PR #11190, opened 2026-07-12
fix(core): coalesce duplicate tracer copies sharing run state

### LINKED ISSUE #11189: LangChainTracer logs "No chain run to end" for every end event when a compiled graph runs inside another StateGraph node
### Checked other resources

### Example Code

Fully self-contained and offline: deterministic fake chat model, local HTTP stub as the
LangSmith endpoint, no real API key needed.

```bash
# setup (Node 20+)
mkdir langsmith-tracer-repro && cd langsmith-tracer-repro
npm init -y
npm install @langchain/core@1.2.2 @langchain/langgraph@1.4.7 langchain@1.5.3 zod@3
# save the script below as repro-langsmith-tracer-bug.mjs, then:

node repro-langsmith-tracer-bug.mjs                # nested agent -> ~12 noise lines (bug)
NESTED=false node repro-langsmith-tracer-bug.mjs   # same agent top-level -> clean
```

```javascript
// repro-langsmith-tracer-bug.mjs
import { createServer } from "node:http";

// Stub LangSmith API so the run is offline and nothing hangs on retries.
// Env must be set before the langchain modules are imported, hence the
// dynamic imports below.
const stub = createServer((_req, res) => {
  res.writeHead(200, { "Content-Type": "application/json" });
  res.end("{}");
});
await new Promise((resolve) => stub.listen(0, "127.0.0.1", resolve));
process.env.LANGSMITH_TRACING = "true";
process.env.LANGSMITH_API_KEY = "lsv2_pt_fake";
process.env.LANGSMITH_ENDPOINT = `http://127.0.0.1:${stub.address().port}`;

const { BaseChatModel } = await import("@langchain/core/language_models/chat_models");
const { AIMessage, HumanMessage } = await import("@langchain/core/messages");
const { tool } = await import("@langchain/core/tools");
const { z } = await import("zod");
const { createAgent } = await import("langchain");
const { StateGraph, START, END, Annotation } = await import("@langchain/langgraph");

// Deterministic offline chat model: one tool call, then a final answer.
class FakeToolCallingModel extends BaseChatModel {
  queue = [
    new AIMessage({
      content: "",
      tool_calls: [{ id: "call_1", name: "echo", args: { text: "hello" } }],
    }),
    new AIMessage("The echo tool returned: hello"),
  ];
  _llmType() {
    return "fake";
  }
  bindTools() {
    return this;
  }
  async _generate() {
    const message = this.queue.shift() ?? new AIMessage("done");
    return { generations: [{ message, text: "" }] };
  }
}

const echo = tool(async ({ text }) => `echoed: ${text}`, {
  name: "echo",
  description: "Echo the given text back.",
  schema: z.object({ text: z.string() }),
});

const ask = { messages: [new HumanMessage("Call the echo tool once, then answer.")] };
const makeAgent = () => createAgent({ model: new FakeToolCallingModel({}), tools: [echo] });

if (process.env.NESTED !== "false") {
  const State = Annotation.Root({ out: Annotation() });
  const graph = new StateGraph(State)
    .addNode("run", async (_state, config) => {
      // Passing (or omitting) `config` makes no difference - the noise appears either way.
      const result = await makeAgent().invoke(ask, config);
      return { out: result.messages.length };
    })
    .addEdge(START, "run")
    .addEdge("run", END)
    .compile();
  const result = await graph.invoke({});
  console.log("nested run finished, messages:", result.out);
} else {
  const result = await makeAgent().invoke(ask);
  console.log("top-level run finished, messages:", result.messages.length);
}

// Let the langsmith background queue flush against the stub, then shut it down.
const { awaitAllCallbacks } = await import("@langchain/core/callbacks/promises");
await awaitAllCallbacks();
stub.close();
stub.unref();
```

```
Error in handler LangChainTracer, handleChainEnd: Error: No chain run to end.
Error in handler LangChainTracer, handleLLMEnd: Error: No LLM run to end.
Error in handler LangChainTracer, handleChainEnd: Error: No chain run to end.
Error in handler LangChainTracer, handleChainEnd: Error: No chain run to end.
Error in handler LangChainTracer, handleChainEnd: Error: No chain run to end.
Error in handler LangChainTracer, handleToolEnd: Error: No tool run to end
Error in handler LangChainTracer, handleChainEnd: Error: No chain run to end.
Error in handler LangChainTracer, handleLLMEnd: Error: No LLM run to end.
Error in handler LangChainTracer, handleChainEnd: Error: No chain run to end.
Error in handler LangChainTracer, handleChainEnd: Error: No chain run to end.
Error in handler LangChainTracer, handleChainEnd: Error: No chain run to end.
Error in handler LangChainTracer, handleChainEnd: Error: No chain run to end.
nested run finished, messages: 4
```

The errors are thrown in `@langchain/core/dist/tracers/base.js` (`handleChainEnd` /
`handleLLMEnd` / `handleToolEnd`: "No ... run to end") and caught and logged by the callback
manager.

### Description

When a compiled LangGraph graph (e.g. an agent built with `createAgent` from langchain v1) is
invoked inside a node of another `StateGraph` with LangSmith tracing enabled, `@langchain/core`
logs a burst of handler errors to stderr - about 12 lines for a single small agent run.

The uploaded trace is complete and correct (verified via the LangSmith API: every run has
`status: success` and an `end_time`), so the lines appear to be purely cosmetic - but they are
indistinguishable from real errors for end users, and any CLI built on nested graphs spams
stderr with them.

The exact same agent invoked top-level (not inside another graph's node) produces no such
lines (`NESTED=false` mode of the repro).

What seems to happen: I instrumented `LangChainTracer.prototype.handle*` to log which tracer
instance receives which event. In the nested case every start/end event is delivered to TWO
distinct `LangChainTracer` instances (fresh copies are created per `CallbackManager.configure()`
via `copyWithTracingConfig`, see `callbacks/manager.js`). The instances share run bookkeeping
(`usesRunTreeMap = true`), so the first delivery ends the run and removes it from the shared
map, and the second delivery then throws "No chain/LLM/tool run to end". Instrumented output
excerpt (Tn = distinct tracer instances):

```
[T4] handleChainStart run=... known=true
[T5] handleChainStart run=... known=true
[T4] handleChainEnd   run=... known=true
[T5] handleChainEnd   run=... known=false
[T5] handleChainEnd   run=... THREW: No chain run to end.
```

Passing or omitting the node's `config` to the inner `agent.invoke()` makes no difference.

Expected behavior: no "Error in handler LangChainTracer" lines in either mode - each end event
should be handled by exactly one tracer instance, as in the top-level case.

### System Info

- @langchain/core 1.2.2
- @langchain/langgraph 1.4.7
- langchain 1.5.3
- langsmith 0.8.1
- Node v22.18.0, npm 10.9.3
- Linux (Ubuntu, kernel 6.17)
- Reproduced both with a real model (ChatOllama) and with the fake model above

pawel-twardziak:
When a compiled LangGraph graph (e.g. an agent built with `createAgent`) is invoked inside a node of another `StateGraph` with LangSmith tracing enabled, every start/end callback event is delivered to two distinct `LangChainTracer` instances that share the same run bookkeeping (`runTreeMap`). The first instance ends the run and deletes it from the shared map, and the second one then throws - so a single small nested agent run spams stderr with ~12 lines of `Error in handler LangChainTracer, handleChainEnd: Error: No chain run to end.` noise. The uploaded trace itself is complete and correct, but each duplicate tracer also issues redundant `postRun()`/`patchRun()` calls.

The duplication has two compounding sources, both introduced with the tracer-defaults work in #10694:

- `CallbackManager._configureSync()` maps `handlers` and `inheritableHandlers` independently and calls `copyWithTracingConfig()` in both passes, producing two distinct copies of the same tracer and breaking the aliasing invariant that `CallbackManager.copy()` relies on (it decides inheritability by object identity).
- `mergeConfigs()` concatenates `handlers`/`inheritableHandlers` from two callback managers without deduplication, so tracer copies sharing one run store accumulate as configs propagate through nested graph execution.

This PR fixes both:

- `_configureSync()` now uses a single replacement map when applying tracer-inheritable metadata/tags, so the same source handler yields the same copy in both arrays (aliasing preserved).
- A new `coalesceTracers()` step collapses `LangChainTracer` copies that share the same run store - keyed via a new internal `LangChainTracer._getRunStoreKey()`, which returns the `runTreeMap` that `copyWithTracingConfig()` deliberately shares between copies - merging their tracing metadata/tags through the existing `copyWithTracingConfig()` precedence rules. Genuinely independent tracers (separate run stores) and non-tracer handlers are left untouched, and the coalescing runs on every configure call, so duplicates produced by later `mergeConfigs()` calls are collapsed on the next nested invocation.

The end handlers were intentionally *not* made silently idempotent - that would hide duplicate callback delivery while keeping duplicate LangSmith writes.

One deliberate behavioral note for review: `coalesceTracers()` also promotes a tracer found only in `inheritableHandlers` into `handlers`, repairing aliases already broken by the independent copy passes.

**Tests** (written first, failing on current `main`, passing with the fix):

- `tracer configuration preserves inheritable handler identity` - asserts the tracer entry in `handlers` is the exact same object as its entry in `inheritableHandlers` after configuring with tracer-inheritable metadata.
- `nested callback configs coalesce tracers sharing run state` - merges configs carrying the same tracer lineage, reconfigures with a `configurable.thread_id`, and asserts a single tracer remains with correctly merged metadata (`{ tenant, thread_id }`) and that invoking a runnable produces no `No chain run to end` warnings.
- `callback configuration keeps independent tracers` - asserts two tracers with separate run stores are not collapsed.

**Verification:**

- Full `@langchain/core` unit suite passes (97 files, 1450 tests) with no type errors; `oxlint` and `oxfmt` clean on the changed files.
- The self-contained reproducer from the issue (nested `StateGraph` + `createAgent` against a stubbed LangSmith endpoint) produces exactly 12 `Error in handler LangChainTracer` lines on published `@langchain/core@1.2.2` and 0 lines with this patch, in both nested and top-level modes, with identical final results.
- Changeset included: patch bump for `@langchain/core`.

Fixes #11189


colifran (approved):
Tested locally after reviewing. Looks great! Thanks @pawel-twardziak!

PR #11198, opened 2026-07-14
fix(google): preserve thought and thoughtSignature on resent text/reasoning blocks

### LINKED ISSUE #10461: bug(@langchain/google): thinking blocks lose thought/thoughtSignature in multi-turn conversations

saarRozenbaum:
## Summary

Fixes #10461.

Both converter paths in `@langchain/google` silently strip `thought` and `thoughtSignature` when resending a prior `AIMessage`'s thinking text as part of a subsequent turn's history:

- **`convertLegacyContentMessageToGeminiContent`**'s `isMessageContentText` branch only copied `.text`, dropping the sibling `thought`/`thoughtSignature` keys that `convertGeminiCandidateToAIMessage` puts directly on a Gemini thinking part (`{ type: "text", thought: true, ... }`) in `AIMessage.content`.
- **`convertStandardContentBlockToGeminiPart`**'s `"text"` case had the same gap, and additionally had **no case at all for `"reasoning"` blocks** — which is the shape `AIMessage.contentBlocks` actually returns for a Gemini thinking part, via `@langchain/core`'s `ChatGoogleTranslator` (every `AIMessage` this package produces sets `response_metadata.model_provider = "google"`, which is exactly the condition that routes `contentBlocks` through that translator). A `"reasoning"` block hit the `default: return null` branch and was dropped entirely.

Per [Gemini's docs](https://ai.google.dev/gemini-api/docs/thinking): *"You should NOT remove or modify thought blocks from the history, as they contain the signatures required for the model to continue its reasoning."* Both gaps mean a thinking model's own prior chain-of-thought either loses its thought marker (legacy path) or disappears outright (standard/`contentBlocks` path) when replayed on the next turn.

## Why this revisits #10462 (closed)

#10462 attempted the same `"text"`-case fix and was closed after @afirstenberg noted thinking parts flow through this package as `"reasoning"` blocks, not `"text"` blocks with `thought=true`. That's correct **for `AIMessage.contentBlocks`** (via `ChatGoogleTranslator`) — but not for `AIMessage.content` itself, which both converters also read directly, and which genuinely does carry `{ type: "text", thought: true }`. I confirmed this against a live production trace (`gemini-3.5-flash` via Vertex):

```json
[
  { "text": "...(real chain-of-thought)...", "thought": true, "type": "text" },
  { "functionCall": { "name": "geo_agent", "args": {...} }, "thoughtSignature": "AY89a19...", "type": "functionCall" }
]
```

This PR fixes **both** shapes rather than picking one — the `"text"` case (legacy path + as a fallback in the standard path when a Gemini part collapses through `originalTextContentBlock`) and a new `"reasoning"` case (standard path, the shape that actually reaches the converter for a real Gemini response with `model_provider: "google"` set and no `output_version` override).

## Changes

Both additions are additive — spread-guarded on key presence — so a plain text/reasoning block with no `thought`/`thoughtSignature` keys serializes identically to before. No behavior change for non-thinking models or non-thought content.

## Test plan

Added 4 unit tests in `converters/tests/messages.test.ts`:
- Legacy path: a `thought:true` text block round-trips with the flag intact
- Legacy path: a plain text block with no `thought` key is unaffected (control case)
- Explicit `output_version: "v1"`: same preservation via the direct `"text"` case
- Real-world `ChatGoogleTranslator` path: asserts (via `.contentBlocks` directly) that the input really does normalize to a `"reasoning"` block, then verifies the converter preserves `thought` through that shape too

All 4 new tests pass. Pre-existing note: 3 unrelated tests in this same file fail in my local environment on unmodified `main` too (confirmed via `git stash`) — a local/CI environment discrepancy unrelated to this change, not something this PR touches.

## AI Disclosure

Investigated and drafted with AI assistance (Claude), including the production trace analysis and the `ChatGoogleTranslator` code trace that explains why #10462's original framing was closed. I reviewed and verified the diagnosis and patch myself before opening this.

### DIFF
--- a/libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
+++ b/libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
@@ -1532,6 +1532,62 @@ describe.each(thinkingModelInfo)(
       );
       expect(hasThoughtSignature).toBe(true);
     });
+
+    test("thought/thoughtSignature survive a multi-turn round trip (regression for #10461)", async () => {
+      // Reproduces the production bug: resending a prior thinking AIMessage
+      // as history silently stripped its thought marker/signature before
+      // this fix, either losing the flag (legacy content array) or dropping
+      // the block outright (contentBlocks -> ChatGoogleTranslator ->
+      // "reasoning" -> default: return null in the standard converter).
+      const llm = newChatGoogle({ reasoningEffort: "low" });
+
+      const firstResult = await llm.invoke("Why is the sky blue?");
+      const firstReasoningBlocks = firstResult.contentBlocks.filter(
+        (b) => b.type === "reasoning"
+      );
+      expect(firstReasoningBlocks.length).toBeGreaterThan(0);
+      const firstSignatures = firstResult.contentBlocks
+        .filter((b) => "thoughtSignature" in b)
+        .map((b) => b.thoughtSignature);
+      expect(firstSignatures.length).toBeGreaterThan(0);
+
+      // Feed the thinking AIMessage back in as history and ask a follow-up.
+      // recorder.request captures the actual outbound HTTP body for the
+      // second call — assert the resent turn's `contents` entry still
+      // carries `thought`/`thoughtSignature` on the wire, not just that the
+      // call succeeds (a successful call alone wouldn't catch a silently
+      // dropped/unflagged thought block, since Gemini doesn't require the
+      // resent history to include one).
+      await llm.invoke([
+        new HumanMessage("Why is the sky blue?"),
+        firstResult,
+        new HumanMessage("Now explain it to a 5 year old."),
+      ]);
+
+      const sentContents = recorder.request?.body?.contents ?? [];
+      const modelTurn = sentContents.find(
+        (c: { role?: string }) => c.role === "model"
+      );
+      expect(modelTurn).toBeDefined();
+
+      const sentThoughtParts = (modelTurn?.parts ?? []).filter(
+        (p: { thought?: boolean }) => p.thought === true
+      );
+      expect(sentThoughtParts.length).toBeGreaterThan(0);
+
+      const sentSignatures = (modelTurn?.parts ?? [])
+        .filter((p: { thoughtSignature?: string }) => "thoughtSignature" in p)
+        .map((p: { thoughtSignature?: string }) => p.thoughtSignature);
+      expect(sentSignatures.length).toBeGreaterThan(0);
+      // At least one signature from the first response must round-trip
+      // verbatim — Gemini validates these, so a regenerated/blank value
+      // would defeat the point of preserving the field at all.
+      expect(
+        sentSignatures.some((sig: string | undefined) =>
+          firstSignatures.includes(sig)
+        )
+      ).toBe(true);
+    });
   }
 );
 
--- a/libs/providers/langchain-google/src/converters/messages.ts
+++ b/libs/providers/langchain-google/src/converters/messages.ts
@@ -363,7 +363,28 @@ function convertStandardContentBlockToGeminiPart(
 ): Gemini.Part | null {
   switch (block.type) {
     case "text":
-      return { text: block.text };
+      return {
+        text: block.text,
+        ...("thought" in block ? { thought: block.thought as boolean } : {}),
+        ...("thoughtSignature" in block
+          ? { thoughtSignature: block.thoughtSignature as string }
+          : {}),
+      };
+    case "reasoning":
+      // ChatGoogleTranslator (block_translators/google.ts) normalizes a
+      // Gemini thinking part into `{ type: "reasoning", reasoning, thought,
+      // thoughtSignature, reasoningContentBlock }` for the standard
+      // `contentBlocks` getter. Round-trip it back into the shape Gemini
+      // expects — without this, a thinking model's own chain-of-thought is
+      // silently dropped (hits the `default` branch below) when history
+      // that went through `.contentBlocks` is replayed on the next turn.
+      return {
+        text: block.reasoning,
+        ...("thought" in block ? { thought: block.thought as boolean } : {}),
+        ...("thoughtSignature" in block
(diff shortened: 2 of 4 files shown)


afirstenberg:
I appreciate the proposed fix. You mentioned in #10461 that you had hit this in production.

Can we get an integration test that demonstrates this problem so we make sure this is a solution and we don't get a regression in the future, please?

saarRozenbaum:
Added an integration test (`thought/thoughtSignature survive a multi-turn round trip`) in `chat_models/tests/index.int.test.ts`, in the existing `Google Thinking` describe block. It invokes a real thinking model twice, resends the first response's thinking `AIMessage` as history, and asserts against `GoogleRequestRecorder`'s captured outbound HTTP body that the resent turn's `contents` entry still carries `thought: true` parts and a matching `thoughtSignature`.

Verified against a live Gemini API key:
- **With the fix**: passes for every currently-available `useApiKey: true` thinking model (`gemini-3-flash-preview`, `gemini-3.1-pro-preview`, `gemini-3.1-flash-lite-preview`). `gemini-3-pro-preview` fails separately with `This model...is no longer available` — a Google-side deprecation, unrelated to this change.
- **With the fix reverted** (I temporarily restored the pre-fix `converters/messages.ts` and reran): all three previously-passing models fail with `AssertionError: expected 0 to be greater than 0` on `sentThoughtParts.length` — confirming the test actually catches the regression rather than passing vacuously.

afirstenberg:
Thanks @saarRozenbaum! I'll try to take a look at everything this week, tho it may be till the weekend. (I'm catching up on a lot of things.)

libs/providers/langchain-google/src/converters/messages.ts
```diff
@@ -363,7 +363,28 @@ function convertStandardContentBlockToGeminiPart(
 ): Gemini.Part | null {
   switch (block.type) {
     case "text":
-      return { text: block.text };
+      return {
+        text: block.text,
+        ...("thought" in block ? { thought: block.thought as boolean } : {}),
```
afirstenberg, line 368: I think part of what I'm struggling with about this change (unless I'm misunderstanding something somewhere, which I very well may be) is that this should never be true.

A Gemini text part (really, any part) that has thought set to true should *always* be converted to a reasoning block. If it isn't - why isn't it?

libs/providers/langchain-google/src/converters/messages.ts
```diff
@@ -363,7 +363,28 @@ function convertStandardContentBlockToGeminiPart(
...
+        ...("thoughtSignature" in block
```
afirstenberg, line 369: Since thought signatures, and even thoughts, can be on any `Gemini.Part`, we may want to apply these rules against everything that could be returned.

libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
```diff
@@ -1532,6 +1532,62 @@ describe.each(thinkingModelInfo)(
       );
       expect(hasThoughtSignature).toBe(true);
     });
+
+    test("thought/thoughtSignature survive a multi-turn round trip (regression for #10461)", async () => {
```
afirstenberg, line 1536: This is mostly testing making sure Gemini accepts what we send. But it was accepting that before.

If I'm following what is happening, I think we should be testing root-cause cases - where blocks with `thinking:true` are not converted to `Reasoning` blocks, and thus the `thinking` attribute is lost.

afirstenberg (commented):
Good analysis, and thank you for doing the work!

But if I am understanding things (and please tell me if I'm not), the underlying issue is that we have LangChain `ContextBlock.Text` blocks that should have `thought: true` but don't. I'd counter that anything that has `thought: true` should be a `ContentBlock.Reasoning` and that the bug is that we're missing this case somewhere.

saarRozenbaum:
You're right, and I don't think we disagree on where the defect belongs — a part with `thought: true` *should* be a `ContentBlock.Reasoning`, and the producer failing to do that is the real bug. Let me answer "why isn't it?" directly.

`convertGeminiPartToContentBlock` returns the part shape straight through:

```ts
if ("text" in part && typeof part.text === "string") return {
  type: "text",
  text: part.text,
};
```

`AIMessage.content` is built from that, so a Gemini thinking part lands in `content` as a `"text"`-typed block. `contentBlocks` normalizes it to `"reasoning"` via `ChatGoogleTranslator`, but both converters also read `content` directly — and `convertStandardContentMessageToGeminiContent` prefers `additional_kwargs.originalTextContentBlock` over the normalized block, which puts the un-normalized shape back in the path even on the standard route.

So: fix the producer to emit `Reasoning` — that's the correct change and I'm happy to move the PR there. But the converter also has to tolerate the legacy shape, because checkpointed conversation history already contains it. In our production traces (`gemini-3.5-flash` via Vertex) that's what's persisted today:

```json
"additional_kwargs": {
  "originalTextContentBlock": {
    "text": "[...]",
    "thoughtSignature": "AY89a191UpdSvFJ96hwFTd+PKI80hA05...",
    "type": "text"
  }
}
```

Those threads get replayed after the producer is fixed. Normalizing at the source doesn't retroactively rewrite them, so dropping the converter tolerance would silently discard signatures from every existing conversation.

On your second point — agreed, and the same traces show it: `thoughtSignature` rides on `functionCall` parts too, as a sibling key. Handling only `"text"` is too narrow; the rule belongs on any `Gemini.Part` that can carry `thought`/`thoughtSignature`.

On the test — fair, mine proves Gemini accepts what we send, which it already did. I'll replace it with the root-cause assertion: that a Gemini thinking part fails to become a `Reasoning` block, which is the actual defect and fails on `main` for the right reason.

Separately, this package has a second production-confirmed issue in the same two functions — Vertex started populating `functionCall.id` on 2026-07-24 20:33 UTC and rejects it on the way back in, 400ing every multi-turn tool-calling conversation. Details and timeline in #11209.

PR #11206, opened 2026-07-15
fix(google-genai): serialize standard v1 reasoning content blocks

### LINKED ISSUE #11180: google-genai: request serializer throws "Unknown content type reasoning" on standard v1 reasoning blocks (standard-block sibling of #10103)
### Environment

- `@langchain/google-genai`: 2.2.0 (latest)
- `@langchain/core`: 1.2.2 (latest)
- Node 22

### Summary

`@langchain/google-genai`'s request serializer (`convertMessageContentToParts` / `_convertLangChainContentToPart` in `utils/common.ts`) throws `Unknown content type reasoning` when the history contains the **standard v1 `reasoning` content block** — which is exactly the shape `@langchain/core`'s own v1 pipeline produces (`ChatModelStream._assembleMessage`, `output_version: "v1"` messages).

This is the standard-block sibling of #10103 (which fixed the provider `thinking` shape). Any multi-turn agent whose state stores v1 messages (e.g. langgraph checkpoints via the v3 stream protocol, `createAgent` with stream-events consumers) crashes on the second model call:

```
Error: Unknown content type reasoning
    at _convertLangChainContentToPart (@langchain/google-genai/dist/utils/common.js:201)
```

### Reproduction (no network needed)

```js
import { AIMessage, HumanMessage } from '@langchain/core/messages';
import { ChatGoogleGenerativeAI } from '@langchain/google-genai';

const model = new ChatGoogleGenerativeAI({ model: 'gemini-3.1-pro-preview', apiKey: 'test' });

// Shape produced by core's ChatModelStream._assembleMessage (output_version v1)
const history = [
  new HumanMessage('hi'),
  new AIMessage({
    content: [
      { type: 'reasoning', reasoning: 'thinking about it' },
      { type: 'tool_call', id: 'tc-1', name: 'get_weather', args: { city: 'Paris' } },
    ],
    tool_calls: [{ name: 'get_weather', args: { city: 'Paris' }, id: 'tc-1' }],
    response_metadata: { output_version: 'v1', model_provider: 'google-genai', finish_reason: 'stop' },
  }),
];

model._buildGenerateContentRequest(history, {});
// => Error: Unknown content type reasoning
```

### Expected

The serializer should translate the standard `reasoning` block back to a Gemini thought part (`{ text, thought: true, thoughtSignature? }`), the same way it already handles the provider `thinking` shape since #10103 — i.e. the package should be able to serialize the message format its own ecosystem produces.

### Impact

In production this surfaces as an unhandled rejection out of `AgentNode #invokeModel` that permanently wedges a thread: every replay of the checkpointed history crashes the run. We currently work around it with a model wrapper that rewrites `reasoning` → `thinking` before serialization.

ondraulehla:
Fixes #11180

## Problem

The request serializer (`_convertLangChainContentToPart` in `libs/providers/langchain-google-genai/src/utils/common.ts`) throws `Unknown content type reasoning` when the history contains the standard v1 `reasoning` content block. That block shape is produced by `@langchain/core`'s own v1 output pipeline (`output_version: "v1"`), so any multi-turn agent that stores v1 messages (for example langgraph checkpoints) crashes on the second model call when the history is replayed.

This is the standard-block sibling of #10103, which fixed the same failure for the provider-specific `thinking` shape.

## Fix

Handle `type: "reasoning"` in `_convertLangChainContentToPart` and convert it back to a Gemini thought part (`{ text, thought: true, thoughtSignature? }`), mirroring the existing `thinking` branch. The `signature` field is read directly from the block, which matches the shape core's own Google block translator emits (`block_translators/google_vertexai.ts`).

## Tests

Three new unit tests in `src/tests/common.test.ts` (no network):
- reasoning block converts to `{ text, thought: true }` and does not gain a `thoughtSignature`
- reasoning block with a `signature` keeps it as `thoughtSignature`
- the exact multi-turn replay shape from the issue (v1 `AIMessage` with `reasoning` + `tool_call` blocks) serializes without throwing

Package checks: `vitest run` for the package passes (75 tests, no type errors), `oxfmt --check` and `oxlint` clean on the touched files, package build passes. Changeset included (patch on `@langchain/google-genai`).

### DIFF
--- a/.changeset/fix-google-genai-reasoning-blocks.md
+++ b/.changeset/fix-google-genai-reasoning-blocks.md
@@ -1,0 +1,5 @@
+---
+"@langchain/google-genai": patch
+---
+
+fix(google-genai): serialize standard v1 `reasoning` content blocks back to Gemini thought parts instead of throwing "Unknown content type reasoning", so multi-turn agents replaying v1 message history (for example langgraph checkpoints) no longer crash on the second model call
--- a/libs/providers/langchain-google-genai/src/tests/common.test.ts
+++ b/libs/providers/langchain-google-genai/src/tests/common.test.ts
@@ -376,3 +376,83 @@ describe("Round-trip thinking content handling", () => {
     });
   });
 });
+
+// https://github.com/langchain-ai/langchainjs/issues/11180
+describe("Round-trip standard v1 reasoning content handling", () => {
+  test("reasoning block converts back to Gemini thought part", () => {
+    const message = new AIMessage({
+      content: [
+        { type: "reasoning", reasoning: "Let me reason about this..." },
+        { type: "text", text: "The answer is 42." },
+      ],
+    });
+
+    const parts = convertMessageContentToParts(message, true, []);
+
+    expect(parts).toHaveLength(2);
+    expect(parts[0]).toEqual({
+      text: "Let me reason about this...",
+      thought: true,
+    });
+    expect(parts[0]).not.toHaveProperty("thoughtSignature");
+    expect(parts[1]).toEqual({ text: "The answer is 42." });
+  });
+
+  test("reasoning block with signature keeps thoughtSignature", () => {
+    const message = new AIMessage({
+      content: [
+        {
+          type: "reasoning",
+          reasoning: "Signed reasoning",
+          signature: "sig456",
+        },
+      ],
+    });
+
+    const parts = convertMessageContentToParts(message, true, []);
+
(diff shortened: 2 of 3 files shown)


thushanth-bengre-langchain:
Hi @ondraulehla. Thank you for this PR.

Our direction is going to be using `@langchain/google` package int he future. Since this is fixing `@langchain/google-genai`, we will be closing this. Please use `@langchain/google` and open new issue if you find similar behavior in that package.

PR #11208, opened 2026-07-15
fix(@langchain/google): omit native functionCall.id and functionResponse.id on Vertex, keep them on gai

### LINKED ISSUE #11209: functionResponse.id still leaks native Gemini/Vertex ids, causing the same 400 fixed in #10292
### Context

#10266 / #10292 fixed Vertex AI's `generateContent` rejecting a `functionResponse.id` field (`Unknown name "id" ... Cannot find field`), by omitting `id` whenever `message.tool_call_id.startsWith("lc-tool-call-")` — i.e. when the id was LangChain-synthesized rather than supplied by the model.

That check only catches the synthetic case. `convertGeminiPartsToToolCalls` (in `libs/providers/langchain-google/src/converters/messages.ts`) falls back to a synthetic `lc-tool-call-*` id **only when the model doesn't supply its own `functionCall.id`**:

```ts
id:
  functionCallPart.functionCall.id ??
  `lc-tool-call-${uuidv4().replace(/-/g, "")}`,
```

Both Gemini (via the Developer API) and Vertex AI frequently *do* supply a native `functionCall.id` (e.g. `"m4v7k35n"`). When that native id round-trips back into a `ToolMessage.tool_call_id` on the next turn, `isGeneratedId` is `false` (it doesn't start with `lc-tool-call-`), so the id is attached to `functionResponse.id` — and on Vertex this hits the exact same 400 error #10266 already reported and supposedly fixed:

```
Invalid JSON payload received. Unknown name "id" at 'contents[2].parts[0].function_response': Cannot find field.
```

This is a **production-observed regression of #10292's own intent** — the fix assumed only synthetic ids needed to be stripped, but Vertex rejects `functionResponse.id` regardless of whether the id is native or synthetic. Any multi-turn tool-calling conversation against Vertex where Gemini supplies its own `functionCall.id` (common — not an edge case) will 400 on the next turn.

### Reproduction

1. Configure `ChatGoogle` with `platformType: "gcp"` (Vertex).
2. Send a message that triggers a tool call. Vertex returns a `functionCall` part with a native `id` (no `lc-tool-call-` prefix).
3. Reply with a `ToolMessage` using that `tool_call_id`.
4. Send the follow-up turn. Vertex 400s with `Unknown name "id" ... Cannot find field`.

### Why the fix isn't "just strip `id` everywhere"

I initially assumed a blanket strip (never send `functionResponse.id`, on any platform) was the right fix, but verified empirically against the Gemini Developer API (`platformType: "gai"`) that `id` **is** load-bearing there: when a turn produces two parallel `functionCall`s to the *same* tool name (so name alone can't disambiguate) and the `functionResponse`s are sent back out of order, the model only pairs them correctly when `id` is present. Without it, responses are matched positionally and produce a wrong/swapped answer. This also matches the existing test `"passes tool_call_id through as functionResponse.id (legacy path)"` in `messages.test.ts`, which explicitly locks in id-passthrough as intended behavior for the no-platform-specified path — just without awareness of the Vertex-specific rejection.

So the correct fix is platform-gated: omit a **native** `functionResponse.id` only when the target platform is Vertex (`gcp`), keep omitting a **generated** id everywhere (unchanged from #10292 — it's never meaningful to the model), and keep passing a native id through everywhere else.

### Fix

Opened as #11208 — threads `platformType` through `convertMessagesToGeminiContents` → `convertStandardContentMessageToGeminiContent` / `convertLegacyContentMessageToGeminiContent`, combining it with the existing `isGeneratedId` check. Includes new tests covering all 4 native/generated × gcp/gai combinations, and a changeset. Verified against a local patch with a live end-to-end run against Vertex (multi-turn tool-calling, matching the production failure) and an empirical test against the Gemini Developer API confirming the parallel-call disambiguation behavior.

saarRozenbaum:
Fixes #11209

## Problem

#10292 fixed Vertex AI's `generateContent` rejecting a `functionResponse.id` field (`Unknown name "id" ... Cannot find field`) by omitting `id` whenever `message.tool_call_id.startsWith("lc-tool-call-")` — i.e. only when the id was LangChain-synthesized.

That check doesn't cover ids Gemini/Vertex supplies natively. `convertGeminiPartsToToolCalls` only falls back to a synthetic `lc-tool-call-*` id when the model doesn't supply its own `functionCall.id`:

```ts
id:
  functionCallPart.functionCall.id ??
  `lc-tool-call-${uuidv4().replace(/-/g, "")}`,
```

When Vertex supplies a native id (common, not an edge case), that id round-trips into `functionResponse.id` on the next turn, `isGeneratedId` is `false`, and Vertex 400s with the exact error #10292 was meant to fix:

```
Invalid JSON payload received. Unknown name "id" at 'contents[2].parts[0].function_response': Cannot find field.
```

This is a real regression of #10292's own intent, hit in production against Vertex with multi-turn tool-calling.

## Why not just strip `id` everywhere

I initially assumed a blanket strip was correct, but verified empirically against the Gemini Developer API (`platformType: "gai"`) that a native `functionResponse.id` **is** load-bearing there: when a turn produces two parallel `functionCall`s to the *same* tool name (so name alone can't disambiguate) and the `functionResponse`s are sent back out of order, the model only pairs them correctly when `id` is present — without it, responses are matched positionally and produce a wrong/swapped answer. The existing test `"passes tool_call_id through as functionResponse.id (legacy path)"` already locks this behavior in for the no-platform-specified path.

So the fix is platform-gated: still omit a **generated** id everywhere (unchanged from #10292 — it's never meaningful to the model), but also omit a **native** id specifically when the platform is Vertex (`gcp`), since Vertex rejects `functionResponse.id` regardless of the id's origin.

## Fix

Threads `platformType` through `convertMessagesToGeminiContents` → `convertStandardContentMessageToGeminiContent` / `convertLegacyContentMessageToGeminiContent`, and combines it with the existing `isGeneratedId` check:

```ts
// before
const isGeneratedId = message.tool_call_id.startsWith("lc-tool-call-");
parts.push({
  functionResponse: {
    ...(isGeneratedId ? {} : { id: message.tool_call_id }),
    ...
  },
});

// after
const isGeneratedId = message.tool_call_id.startsWith("lc-tool-call-");
parts.push({
  functionResponse: {
    ...(isGeneratedId || platformType === "gcp"
      ? {}
      : { id: message.tool_call_id }),
    ...
  },
});
```

`platformType` is available at all 3 call sites of `convertMessagesToGeminiContents` in `chat_models/base.ts` via `this.platform`.

Also switched `convertMessagesToGeminiContents`'s type annotation from `Converter<BaseMessage[], ...>` (a strict 1-arg function alias) to an explicit 2-param signature, since the `Converter` type couldn't express the new optional parameter.

## Tests

All existing tests pass unchanged (none of them pass a `platformType`, so default/no-platform behavior — preserve a native id, omit a generated id — is untouched). Added 4 new tests covering the platform-gated cases:

- native id + `gcp` → omitted (Vertex rejects it)
- native id + `gai` → preserved (needed for parallel-call disambiguation)
- generated id + `gcp` → omitted (unchanged)
- generated id + `gai` → omitted (unchanged — a generated id is never meaningful to the model on either platform)

Verified against a local patch-package equivalent of this fix with a live end-to-end run against Vertex (multi-turn tool-calling, matching the production failure depth) and an empirical test directly against the Gemini Developer API confirming the parallel-call disambiguation behavior described above.

### DIFF
--- a/.changeset/vertex-function-response-id-passthrough.md
+++ b/.changeset/vertex-function-response-id-passthrough.md
@@ -1,0 +1,9 @@
+---
+"@langchain/google": patch
+---
+
+fix: omit native functionResponse.id for Vertex AI, preserve it for the Gemini Developer API
+
+#10292 stripped a LangChain-generated `tool_call_id` from `functionResponse.id` to fix Vertex AI's rejection of that field, but only checked for the `lc-tool-call-` prefix. When Gemini/Vertex supplies its own native `functionCall.id`, that id is not generated, so it still round-trips into `functionResponse.id` and Vertex rejects the request with the same `Unknown name "id" ... Cannot find field` error #10292 was meant to fix.
+
+Vertex rejects `functionResponse.id` unconditionally (native or generated), while the Gemini Developer API accepts it and needs it to disambiguate parallel calls to the same tool name. This now threads the resolved `platformType` through `convertMessagesToGeminiContents` and only omits a native id when the platform is Vertex (`gcp`).
--- a/.changeset/vertex-function-response-id-passthrough.md
+++ b/.changeset/vertex-function-response-id-passthrough.md
@@ -2,8 +2,10 @@
 "@langchain/google": patch
 ---
 
-fix: omit native functionResponse.id for Vertex AI, preserve it for the Gemini Developer API
+fix: omit native functionCall.id / functionResponse.id for Vertex AI, preserve them for the Gemini Developer API
 
 #10292 stripped a LangChain-generated `tool_call_id` from `functionResponse.id` to fix Vertex AI's rejection of that field, but only checked for the `lc-tool-call-` prefix. When Gemini/Vertex supplies its own native `functionCall.id`, that id is not generated, so it still round-trips into `functionResponse.id` and Vertex rejects the request with the same `Unknown name "id" ... Cannot find field` error #10292 was meant to fix.
 
 Vertex rejects `functionResponse.id` unconditionally (native or generated), while the Gemini Developer API accepts it and needs it to disambiguate parallel calls to the same tool name. This now threads the resolved `platformType` through `convertMessagesToGeminiContents` and only omits a native id when the platform is Vertex (`gcp`).
+
+The same leak affects `functionCall.id` in the opposite direction. `convertGeminiPartToContentBlock` stores the model's `functionCall` object verbatim — including a native `id` — and the legacy converter spreads it straight back onto the wire, so replaying a prior tool call 400s with `Unknown name "id" at 'contents[N].parts[0].function_call'`. Vertex began populating `functionCall.id` on 2026-07-24, turning every multi-turn tool-calling conversation into a failure with no client-side change. Both part types are now gated on `platformType` (#11209).
--- a/libs/providers/langchain-google/src/chat_models/base.ts
+++ b/libs/providers/langchain-google/src/chat_models/base.ts
@@ -468,7 +468,7 @@ export abstract class BaseChatGoogle<
     const body = {
       ...this.invocationParams(options),
       systemInstruction: convertMessagesToGeminiSystemInstruction(messages),
-      contents: convertMessagesToGeminiContents(messages),
+      contents: convertMessagesToGeminiContents(messages, this.platform),
     };
 
     const moduleName = this.constructor.name;
@@ -583,7 +583,7 @@ export abstract class BaseChatGoogle<
     const body = {
       ...this.invocationParams(options),
       systemInstruction: convertMessagesToGeminiSystemInstruction(messages),
-      contents: convertMessagesToGeminiContents(messages),
+      contents: convertMessagesToGeminiContents(messages, this.platform),
     };
 
     const url = await this.buildUrl("streamGenerateContent?alt=sse");
@@ -657,7 +657,7 @@ export abstract class BaseChatGoogle<
     const body = {
       ...this.invocationParams(options),
       systemInstruction: convertMessagesToGeminiSystemInstruction(messages),
-      contents: convertMessagesToGeminiContents(messages),
+      contents: convertMessagesToGeminiContents(messages, this.platform),
     };
 
     const url = await this.buildUrl("streamGenerateContent?alt=sse");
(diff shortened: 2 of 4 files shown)


saarRozenbaum:
Widened to cover `functionCall.id` as well — the `functionResponse`-only fix left half the leak open, as @cro13 measured in #11209.

`convertGeminiPartToContentBlock` stores the model's `functionCall` object verbatim (native `id` included), and the legacy converter spreads it straight back onto the wire:

```ts
} else if (item?.type === "functionCall") {
  const { type, functionCall, ...etc } = item;
  parts.push({ ...etc, functionCall });   // id goes back out
```

Neither side is guarded by `isGeneratedId`, so replaying a prior tool call 400s with `Unknown name "id" at 'contents[N].parts[0].function_call'` regardless of #10292.

This became a live outage rather than a latent bug on **2026-07-24 at 20:33 UTC**, when Vertex started populating the field. From production traces (`gemini-3.5-flash`, `platformType: "gcp"`):

| Time (UTC) | Response shape |
|---|---|
| <= 20:32 | `functionCall: {name, args}` — no id, `lc-tool-call-*` synthesized |
| 20:33:28 | `functionCall: {name, args, id: "a323kt5y"}` — native id appears |
| 20:33:33 | first 400 |

Five seconds, no client-side change — same package version deployed a week earlier. 39 failed turns across two users over ~90 minutes, every one on turn >= 2 (turn 1 has no prior `functionCall` to replay).

### Changes

Gated on `platformType` via the same plumbing this PR already threads, so the `gai` passthrough that parallel same-name tool calls rely on for disambiguation is preserved. Extracted as `stripNativeFunctionCallId` since it needs a guarded copy rather than a `delete` — the stored `functionCall` is pushed by reference, so mutating it would strip the id from the caller's message history too (thanks @cro13 for catching that detail).

### Test plan

3 new unit tests: `gcp` strips the id while preserving `name`/`args`/`thoughtSignature`, `gai` preserves it, and the caller's content block is left unmutated. Content blocks use the real production wire shape.

Verified non-vacuous: reverting just the one-line call site flips `omits a native functionCall.id when platformType is gcp (Vertex)` to failing, and restoring it goes green — 24 -> 27 passing.

Pre-existing note, unchanged by this PR: 3 tests in `messages.test.ts` (`merges consecutive ToolMessages...`, `AIMessage with tool_calls produces model turn...`, `strips type field from executableCode...`) fail identically on this branch before my changes — confirmed via `git stash`.

Godrules500:
Hello, I wanted to check in on when this is expected to be resolved?

afirstenberg:
@Godrules500 - See the discussion at #11209 

Do you have a demonstrated test that indicates there is a problem? Every test I've run has suggested this isn't an issue, and I was going to suggest that this and #11209 be closed.

But if you have a test case that clearly demonstrates the problem, I'm very happy to take a look.

PR #11211, opened 2026-07-15
feat(core): add wrapRunExecution hook for run-scoped context propagation

JacksonWeber:
## Summary

Adds an optional `wrapRunExecution(runId, fn)` hook to the callback-handler contract, plus `BaseRunManager.withRunContext` / `withRunContextAsyncIterable` that invoke it around the **execution** of a run body — chat model `_generate`, streaming chunks, and tool `_call`.

Unlike the existing `handle*Start` / `handle*End` callbacks (point-in-time notifications), this lets a handler keep an **ambient context active for the whole duration of the run**.

## Motivation

The motivating use case is distributed tracing. Handlers such as an OpenTelemetry tracer observe runs via callbacks, but callbacks can't wrap execution — so there's no way for a handler to make a run's span the *active* context while the model's network call runs. As a result, lower-level client instrumentations (HTTP/`fetch` via undici, DB drivers) start their spans off `context.active()` at request time, see no active span, and emit **disconnected root traces** instead of spans nested under the LLM/tool run.

Python LangChain avoids this because its callback tracer can `context.attach`/`detach` across `contextvars`; JS callbacks run in a child async scope that reverts before `_generate()` executes, so the equivalent can't be done from a handler alone. This hook closes that gap generically, without coupling core to OpenTelemetry.

## What changed

- **`callbacks/base.ts`** — new optional method on the handler contract:
  ```ts
  wrapRunExecution?<T>(runId: string, fn: () => T): T;
  ```
- **`callbacks/manager.ts`** — `BaseRunManager.withRunContext(fn)` composes every handler's `wrapRunExecution` for the run (skips handlers that don't implement it; zero overhead when none do). `withRunContextAsyncIterable(iterable)` proxies `next`/`return`/`throw` so streaming network calls (initiated lazily on first `next()`) are also covered.
- **`language_models/chat_models.ts`** — wraps the non-streaming `_generate` call and the streaming `_streamResponseChunks` iteration in the run context.
- **`tools/index.ts`** — wraps the tool `_call` in the run context.

## Behavior / compatibility

- **Fully backward compatible.** `wrapRunExecution` is optional; handlers that don't implement it are skipped and `fn` is invoked directly — no added overhead, no behavior change.
- Implementations must invoke `fn` exactly once and return its result (sync return values and rejected promises propagate unchanged).
- No new dependencies; core stays OpenTelemetry-agnostic.

## Example (OpenTelemetry handler)

```ts
wrapRunExecution<T>(runId: string, fn: () => T): T {
  const span = this.spanForRun(runId);
  return span
    ? context.with(trace.setSpan(context.active(), span), fn)
    : fn();
}
```

With this, an outgoing `fetch` inside the model call nests under the run's `chat` span in the same trace, instead of starting a new root trace.

## Notes

- Draft: opening for design feedback. Wrapping is currently applied to chat `_generate` + streaming and tool `_call`; happy to extend to text LLMs (`llms.ts`) and chain/runnable call sites if the approach looks good.
- A changeset is included (`@langchain/core`, minor).
- Tests included:
  - `src/callbacks/tests/wrap_run_execution.test.ts` — unit tests for `withRunContext` (hook invocation, runId passing, multi-handler composition order, plain-handler skip, sync-throw and rejected-promise propagation) and `withRunContextAsyncIterable` (per-`next()` wrapping, passthrough, `return()`/`throw()` forwarding).
  - `src/language_models/tests/chat_models.test.ts` — integration test asserting the hook is active during `_generate`.
  - `src/tools/tests/tools.test.ts` — integration test asserting the hook is active during the tool call body.
  - All affected suites pass (102 tests, no regressions).


mohsinsurani:
Hi @hntrl, this PR is ready for maintainer review. It fixes #11221, includes regression coverage for non-streaming, streaming, and generator-tool execution, and all visible checks are green. @JacksonWeber reviewed the generator follow-up and said LGTM. There is also the older draft #11211 covering the same issue, could you advise which implementation the project would prefer to proceed with? I’m happy to rebase or make any requested changes.

JacksonWeber:
Closing in favor of https://github.com/langchain-ai/langchainjs/pull/11226.

PR #11226, opened 2026-07-20
fix(core): propagate callback context through run execution

### LINKED ISSUE #11221: Provider HTTP spans emit as disconnected root traces
### Checked other resources

### Example Code

A callback handler that follows the documented pattern (start a span in
`handleChatModelStart`, end it in `handleLLMEnd`) has no way to make that span
the active OpenTelemetry context while the model's network call actually runs.
```typescript
import { NodeSDK } from "@opentelemetry/sdk-node";
import {
  InMemorySpanExporter,
  SimpleSpanProcessor,
} from "@opentelemetry/sdk-trace-node";
import { UndiciInstrumentation } from "@opentelemetry/instrumentation-undici";
import { trace, type Span } from "@opentelemetry/api";
import type { LLMResult } from "@langchain/core/outputs";
import { AzureChatOpenAI } from "@langchain/openai";

const exporter = new InMemorySpanExporter();
const sdk = new NodeSDK({
  spanProcessors: [new SimpleSpanProcessor(exporter)],
  instrumentations: [new UndiciInstrumentation()], // instruments global fetch
});
sdk.start();

const tracer = trace.getTracer("langchain");

// Minimal tracing handler following the usual start/end span pattern.
class OtelHandler extends BaseCallbackHandler {
  name = "OtelHandler";
  private spans = new Map<string, Span>();

  async handleChatModelStart(
    _llm: Serialized,
    _messages: BaseMessage[][],
    runId: string
  ): Promise<void> {
    // The span exists, but there is no hook that lets me make it the ACTIVE
    // context for the duration of the model call below.
    this.spans.set(runId, tracer.startSpan("chat"));
  }

  async handleLLMEnd(_output: LLMResult, runId: string): Promise<void> {
    this.spans.get(runId)?.end();
    this.spans.delete(runId);
  }
}

// Any real chat model works. Its outgoing HTTPS request to the provider is what
// the undici/fetch instrumentation traces during the run.
const model = new AzureChatOpenAI({
  azureOpenAIApiKey: process.env.AZURE_OPENAI_API_KEY,
  azureOpenAIBasePath: process.env.AZURE_OPENAI_BASE_PATH,
  azureOpenAIApiDeploymentName: process.env.AZURE_OPENAI_DEPLOYMENT ?? "gpt-4o",
  azureOpenAIApiVersion: process.env.AZURE_OPENAI_API_VERSION ?? "2024-10-21",
});

await model.invoke("Say hello in 3 words.", {
  callbacks: [new OtelHandler()],
});

// Inspect what was exported.
for (const span of exporter.getFinishedSpans()) {
  console.log({
    name: span.name,
    traceId: span.spanContext().traceId,
    parentSpanId: span.parentSpanContext?.spanId,
  });
}

await sdk.shutdown();
```

No error is thrown. The defect is in the emitted trace structure. The provider's
HTTP span lands in a different trace with no parent, instead of nesting under the
`chat` span:
```
{ name: 'chat',                traceId: 'aaaa...', parentSpanId: undefined }
{ name: 'POST api.openai.com', traceId: 'bbbb...', parentSpanId: undefined }
```

Expected (single trace, HTTP span nested under the run):

```
{ name: 'chat',                traceId: 'aaaa...', parentSpanId: undefined }
{ name: 'POST api.openai.com', traceId: 'aaaa...', parentSpanId: '<chat span id>' }
```

### Description

* I'm trying to use a callback handler to trace chat model runs with OpenTelemetry, and have the provider's outgoing HTTP request nest under the run's span in the same trace.
* I expect the HTTP span emitted by the `undici`/`fetch` instrumentation to be a child of the `chat` span, because the network request happens during the model call that the run represents.
* Instead, the HTTP span starts a separate root trace with no parent. The lower-level instrumentation reads `context.active()` at request time, finds no active span (the handler's start/end callbacks have already unwound), and emits a disconnected root span.

The root cause is that there is no seam in core for a handler to keep an ambient context active across the *execution* of a run body. The relevant call sites invoke the body directly:
* `language_models/chat_models.ts` runs the non-streaming `_generate` call and the streaming `_streamResponseChunks` iteration directly.
* `tools/index.ts` runs the tool `_call` directly.

The request is initiated lazily on the first `next()` of the generator and continues on later `next()` calls, none of which run inside the code that creates the generator, so activating context only at creation time would miss the request.

Python LangChain does not hit this because its callback tracer can `context.attach` / `context.detach` across `contextvars`, keeping the run's context active while the body runs. In JS, callbacks run in a child async scope
that reverts before `_generate()` executes, so the same thing can't be done from a handler alone.

An OpenTelemetry-agnostic hook on the handler contract would close the gap, for example an optional `wrapRunExecution(runId, fn)` that core composes around the run body (chat `_generate`, streamed chunks, tool `_call`). Handlers that don't implement it would be unaffected. A draft exploring this is up at langchain-ai/langchainjs#11211.

### System Info

langchain packages:
  @langchain/core: 1.2.2
  @langchain/openai: 1.5.5

platform: windows (the behavior is platform-independent)
Node version: v24.18.0 (needs Node 18+ for global fetch)
package manager: npm 11.16.0

Additional relevant packages:

```
@opentelemetry/sdk-node: 0.220.0
@opentelemetry/sdk-trace-node: 2.9.0
@opentelemetry/instrumentation-undici: 0.30.0
@opentelemetry/api: 1.9.1
```

mohsinsurani:
## Summary

- add an optional callback hook for wrapping the execution of a run body
- compose handler contexts in run managers, including every lazy async-iterator operation
- apply run context propagation to chat model generation, streaming, and tool calls
- add regression coverage for model and tool execution plus wrapper composition and errors

## Testing

- `pnpm --filter @langchain/core test` (98 files, 1455 passed, 1 skipped)
- `oxlint` on changed TypeScript files

Closes #11221

Related: #11211 explores the same callback-hook design.


libs/langchain-core/src/tools/index.ts
```diff
@@ -307,7 +307,8 @@ export abstract class StructuredTool<
 
     let result;
     try {
-      const raw = await this._call(parsed, runManager, config);
+      const call = () => this._call(parsed, runManager, config);
+      const raw = await (runManager ? runManager.withRunContext(call) : call());
```
JacksonWeber, line 311: `withRunContext(call)` here only wraps creating the generator, but an async tool doesn't actually run until `.next()` is called, and that happens later inside `consumeAsyncGenerator(raw, ...)` after the wrapper has already exited. 
So, for generator-based tools the body runs with the context deactivated, and any spans/HTTP calls they make won't nest under the tool run.

mohsinsurani: Good catch- fixed in 5e13569b0. Generator tools are now consumed through the run manager’s context-wrapped async iterator, so each next/return/throw operation executes inside wrapRunExecution. I also added a regression test that observes the context before and after a yield; it failed with [false, false] before the fix and now passes with [true, true].

mohsinsurani: Hi @JacksonWeber, Anything else needed for me to fix this PR?

JacksonWeber: LGTM

mohsinsurani:
Hi @hntrl, this PR is ready for maintainer review. It fixes #11221, includes regression coverage for non-streaming, streaming, and generator-tool execution, and the visible checks are green. @JacksonWeber reviewed the generator follow-up and said LGTM. Could you advise whether this PR or draft #11211 should proceed? I’m happy to address any requested changes.

PR #11231, opened 2026-07-21
fix(aws): handle unstamped tool call content blocks

### LINKED ISSUE #11260: [Bug]: ChatBedrockConverse rejects unstamped standard tool-call blocks during replay
### Checked other resources

- [x] I searched the LangChain.js documentation and GitHub issues.
- [x] I did not find an existing report for this exact failure.
- [x] The failure reproduces on the latest stable `@langchain/aws` and current `main`.

### Example Code

The conversion fails before an AWS request is created, so no credentials are needed. This can be reproduced from `libs/providers/langchain-aws/src/tests`:

```typescript
import { convertToConverseMessages } from "../utils/message_inputs.js";

  content: [
    {
      type: "tool_call",
      id: "tool-call-1",
      name: "write_file",
      args: { file_path: "/tmp/example.txt" },
    },
  ],
});

convertToConverseMessages([message]);
```

### Error Message

```text
Unsupported content block type: tool_call
```

An unstamped `invalid_tool_call` content block reaches the same unsupported-block path.

### Description

`convertAIMessageToConverseMessage()` delegates to the standard v1 converter only when `response_metadata.output_version === "v1"`. Persisted or middleware-reconstructed histories can retain standard `tool_call` / `invalid_tool_call` content blocks after that response metadata is removed. Replaying those messages with `ChatBedrockConverse` then takes the legacy conversion branch and throws before sending a Bedrock request.

Expected behavior:

- replay a valid standard `tool_call` block as a Bedrock `toolUse` block;
- skip diagnostic `invalid_tool_call` blocks, matching the stamped v1 converter;
- preserve inline content order and avoid duplicating calls mirrored in `message.tool_calls`;
- reject a validly typed tool-call block without a usable ID before constructing an invalid Bedrock request.

This was observed downstream during long-running Bedrock agent history replay. Small histories completed, while larger reconstructed histories surfaced the unstamped standard blocks.

A focused fix with regression coverage is available in #11231. The earlier #9766 handles the stamped v1 path; this report covers the residual unstamped replay path.

### System Info

- `@langchain/aws`: 1.4.3
- LangChain.js `main`: `710d74f3fc0ff98d5088c8e0c86752829cfaaa2c`
- Node.js: 24.18.0
- pnpm: 10.14.0
- OS: macOS 26.1 (arm64)

geonwoo-jeong:
Fixes #11260

## Summary

- Replay standard `tool_call` content blocks when `response_metadata.output_version` is missing.
- Skip diagnostic `invalid_tool_call` blocks in the same unstamped path, matching the existing v1 converter.
- Preserve inline content order, avoid duplicate `toolUse` blocks when `tool_calls` mirrors content, and keep inline standard content canonical when mirrored arguments disagree.
- Keep whitespace-only text from corrupting a preceding non-text block and reject missing or blank tool-call IDs before constructing an invalid Bedrock request.

## Context

#9766 routes stamped v1 messages through the standard converter. Persisted or middleware-reconstructed histories can retain v1 `tool_call` or `invalid_tool_call` content blocks after the output-version metadata is lost. Those messages take the legacy branch and currently throw `Unsupported content block type` before a Bedrock request is sent.

Observed downstream in [langchain-ai/openwiki](https://github.com/langchain-ai/openwiki/blob/264ee8465f3c9874b822bcbb7ca68de471143798/src/agent/index.ts#L728-L741) during long Bedrock agent runs: small repositories completed, while large-repository history replay surfaced unstamped v1 content blocks.

This adds a consumer-side fallback only; it does not assume which component removed the metadata. Valid tool calls are replayed as `toolUse`; invalid tool calls remain diagnostic data and are skipped, as they already are in the stamped v1 conversion path.

## Minimal reproduction

```ts
  content: [
    { type: "tool_call", id: "t1", name: "write_file", args: { a: 1 } },
  ],
});

convertToConverseMessages([message]);
```

On current `main`, conversion throws `Unsupported content block type: tool_call`. An unstamped `invalid_tool_call` block reaches the same throw. The expected behavior is to convert valid calls and skip invalid calls while preserving surrounding text. No AWS request or credentials are required to reproduce it.

## Validation

Validated with Node.js 24.18.0 and pnpm 10.14.0:

- `pnpm --filter @langchain/aws test` — 8 files, 133 tests, no type errors
- `pnpm --filter @langchain/aws build` — compile, attw, and publint passed
- `pnpm lint`
- `pnpm exec oxfmt --check .changeset/quick-tools-replay.md libs/providers/langchain-aws/src/tests/chat_models.test.ts libs/providers/langchain-aws/src/utils/message_inputs.ts`

The full monorepo pre-push suite was also attempted. It stops on three existing `@langchain/google` message-converter tests in untouched code; the failures reproduce with the required runtime while both `libs/providers/langchain-google` and `libs/langchain-core` match `upstream/main` exactly.

### DIFF
--- a/.changeset/quick-tools-replay.md
+++ b/.changeset/quick-tools-replay.md
@@ -2,4 +2,4 @@
 "@langchain/aws": patch
 ---
 
-fix(aws): replay unstamped standard tool call content blocks
+fix(aws): replay unstamped standard tool calls and skip invalid tool calls
--- a/.changeset/quick-tools-replay.md
+++ b/.changeset/quick-tools-replay.md
@@ -1,0 +1,5 @@
+---
+"@langchain/aws": patch
+---
+
+fix(aws): replay unstamped standard tool call content blocks
--- a/libs/providers/langchain-aws/src/tests/chat_models.test.ts
+++ b/libs/providers/langchain-aws/src/tests/chat_models.test.ts
@@ -1006,6 +1006,27 @@ describe("unstamped tool call content replay", () => {
       { role: "assistant", content: [{ text: "hello" }] },
     ]);
   });
+
+  it("skips invalid tool calls without output version metadata", () => {
+    const result = convertToConverseMessages([
+      new AIMessage({
+        content: [
+          { type: "text", text: "before" },
+          {
+            type: "invalid_tool_call",
+            id: "tool-call-1",
+            name: "write_file",
+            args: '{"file_path":',
+            error: "Failed to parse tool call arguments as JSON",
+          },
+        ],
+      }),
+    ]);
+
+    expect(result.converseMessages).toEqual([
+      { role: "assistant", content: [{ text: "before" }] },
+    ]);
+  });
 });
 
 test("Streaming supports empty string chunks", async () => {
--- a/libs/providers/langchain-aws/src/tests/chat_models.test.ts
+++ b/libs/providers/langchain-aws/src/tests/chat_models.test.ts
@@ -932,6 +932,82 @@ describe("reasoning content replay", () => {
   });
 });
 
+describe("unstamped tool call content replay", () => {
+  const createToolCallContent = () => [
+    { type: "text", text: "before" },
+    {
+      type: "tool_call",
+      id: "tool-call-1",
+      name: "write_file",
+      args: { file_path: "/tmp/example.txt" },
+    },
+    { type: "text", text: "\n  " },
+    { type: "text", text: "after" },
+  ];
+  const expectedContent = [
+    { text: "before" },
+    {
+      toolUse: {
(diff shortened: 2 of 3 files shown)


geonwoo-jeong:
Hi @hntrl — this is a small `@langchain/aws` replay fix for v1 `tool_call` blocks that lost their `output_version` metadata. The failure was observed downstream in [langchain-ai/openwiki](https://github.com/langchain-ai/openwiki/blob/264ee8465f3c9874b822bcbb7ca68de471143798/src/agent/index.ts#L728-L741) during long Bedrock agent runs. It includes focused regression coverage and a patch changeset; the full Actions suite is awaiting first-time-contributor approval. Would you mind taking a look when you have a chance?

PR #11233, opened 2026-07-21
fix(anthropic): preserve model metadata in streaming chunks

### LINKED ISSUE #11230: ChatAnthropic drops `model` from `response_metadata` when streaming
### Description

`ChatAnthropic` populates `response_metadata.model` on **non-streaming** responses but **omits it on streamed responses** — only `model_provider` and `usage` survive chunk aggregation. The Anthropic streaming API *does* send the model (in the `message_start` event), so the data is available; it just isn't propagated into the aggregated message's `response_metadata`.

This breaks any downstream tooling that attributes a run to its model via `response_metadata.model` (e.g. Braintrust / LangSmith cost estimation): streamed calls log token usage but get no cost because the model is missing.

### Example Code

```ts
import { initChatModel } from "langchain";

for (const streaming of [false, true]) {
  const model = await initChatModel("claude-opus-4-8", {
    modelProvider: "anthropic",
    maxTokens: 1024,
    streaming,
  });
  const res = await model.invoke([{ role: "user", content: "Say hi in one word." }]);
  console.log(streaming, Object.keys(res.response_metadata ?? {}));
}
```

### Actual output

```
false  [ 'model', 'id', 'stop_reason', 'stop_sequence', 'stop_details', 'usage', 'type', 'role', 'model_provider' ]
true   [ 'model_provider', 'usage' ]        // <-- `model` missing when streaming
```

### Expected

`response_metadata.model` should be present in both the streaming and non-streaming paths (populated from the `message_start` event's `model`).

### System Info

- langchain: 1.4.1
- @langchain/anthropic: 1.4.0
- @langchain/core: 1.1.48
- node: v20.20.0
- platform: macOS

### Note

`usage_metadata` (tokens) is populated correctly in both paths — only `response_metadata.model` is affected. (Related but distinct from the token-usage issues #7876 / #8470.)

q1000treadz:
Fixes #11230

This PR fixes Anthropic streaming responses losing the model name after streamed `AIMessageChunk`s are aggregated.

Anthropic sends the model name on the `message_start` streaming event. Previously, LangChain.js preserved that value in `additional_kwargs`, but did not copy it into `response_metadata`, so the final concatenated streamed message lacked `response_metadata.model`. Non-streaming Anthropic responses already include the model in `response_metadata`.

This change copies `message_start.message.model` into the first streamed chunk's `response_metadata.model` when Anthropic provides it, while preserving the existing `additional_kwargs` behavior.

Tests added:
- Verifies `message_start` creates an `AIMessageChunk` with `response_metadata.model`.
- Verifies concatenating multiple chunks preserves `response_metadata.model`.
- Verifies usage metadata and stop metadata are preserved.
- Verifies missing `model` in `message_start` does not throw.
- Verifies `ChatAnthropic` with `streaming = true` preserves model metadata through `invoke()`.

Checks run:
- Targeted regression tests for Anthropic streaming metadata passed.
- Anthropic package unit tests passed.
- Build completed successfully.

### DIFF
--- a/.changeset/fresh-anthropic-model.md
+++ b/.changeset/fresh-anthropic-model.md
@@ -1,0 +1,5 @@
+---
+"@langchain/anthropic": patch
+---
+
+fix(anthropic): preserve model response metadata in streaming chunks
--- a/libs/providers/langchain-anthropic/src/tests/chat_models_stream_events.test.ts
+++ b/libs/providers/langchain-anthropic/src/tests/chat_models_stream_events.test.ts
@@ -299,6 +299,29 @@ describe("ChatAnthropic._streamChatModelEvents (native)", () => {
       ).toBe(25);
     });
 
+    test("streaming invoke preserves message_start model in response metadata", async () => {
+      const model = new MockStreamChatAnthropic(textOnlyEvents());
+      model.streaming = true;
+
+      const message = await model.invoke([new HumanMessage("hello")]);
+
+      expect(message.response_metadata).toMatchObject({
+        model_provider: "anthropic",
+        model: "claude-sonnet-4-20250514",
+        usage: {},
+      });
+      expect(message.additional_kwargs).toMatchObject({
+        model: "claude-sonnet-4-20250514",
+        stop_reason: "end_turn",
+        stop_sequence: null,
+      });
+      expect(message.usage_metadata).toMatchObject({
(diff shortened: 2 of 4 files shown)


Pospelove:
LGTM

jKlag:
This is still an issue - checking back in to see if there are any updates on this merging

PR #11242, opened 2026-07-23
fix(anthropic): preserve gateway (LiteLLM) cost in streamed metadata

talarari:
## Summary

`ChatAnthropic` currently drops a numeric `cost` extension when an Anthropic-compatible gateway adds it to the terminal streaming `message_delta.usage` object.

This change preserves that value at:

```ts
finalChunk.response_metadata.usage.cost;
```

Token accounting remains in `usage_metadata`; no token fields are copied into `response_metadata`.

### LiteLLM

LiteLLM added this response shape specifically for its Anthropic-compatible `/v1/messages` stream in [BerriAI/litellm#15102](https://github.com/BerriAI/litellm/pull/15102).

In the v1.93.0 release:

1. The feature is enabled with `include_cost_in_streaming_usage` ([configuration](https://github.com/BerriAI/litellm/blob/052b5a2169d8d3082e1d66e69f200a72b0c1e274/litellm/proxy/example_config_yaml/pass_through_config.yaml#L41-L42)).
2. The injector selects events whose `type` is `message_delta` and whose `usage` value is an object, calculates the cost, and writes it to `usage.cost` ([source](https://github.com/BerriAI/litellm/blob/052b5a2169d8d3082e1d66e69f200a72b0c1e274/litellm/proxy/common_request_processing.py#L2827-L2889)).
3. LiteLLM's `/v1/messages` integration test asserts that the emitted `usage.cost` exists, is numeric, and is non-negative ([test](https://github.com/BerriAI/litellm/blob/052b5a2169d8d3082e1d66e69f200a72b0c1e274/tests/pass_through_tests/test_anthropic_passthrough.py#L321-L386)).

For example, the resulting terminal event has this shape:

```json
{
  "type": "message_delta",
  "delta": {
    "stop_reason": "end_turn",
    "stop_sequence": null
  },
  "usage": {
    "output_tokens": 42,
    "cost": 0.0123
  }
}
```

LangChain.js handles that event by reconstructing token usage and response metadata, but previously read only `output_tokens` and `delta.context_management`; the gateway extension was therefore discarded ([current handler](https://github.com/langchain-ai/langchainjs/blob/ee76ea0347fb611153e5ec7d0e70fa405f5293a3/libs/providers/langchain-anthropic/src/utils/message_outputs.ts#L53-L72)).

LiteLLM requires the feature flag and resolvable model pricing. If cost calculation fails, the field is omitted. The value comes from LiteLLM's raw `completion_cost()` calculation and can differ from final proxy accounting that applies discounts, margins, or tool costs. This PR forwards the gateway value unchanged; it does not calculate or reinterpret billing data.

### OpenRouter documentation note

OpenRouter's official usage-accounting documentation says that every response includes cost and that streaming usage is sent in the last SSE message ([pinned docs](https://github.com/OpenRouterTeam/docs/blob/e657666ce82f285637d09625659fecd79236b98e/cookbook/administration/usage-accounting.mdx#L21-L32)). Its Anthropic API schema models usage on `message_delta` ([schema](https://github.com/OpenRouterTeam/docs/blob/e657666ce82f285637d09625659fecd79236b98e/openapi/openapi.yaml#L12024-L12110)).

However, that streaming schema does not currently declare `cost`, while OpenRouter's non-streaming Anthropic schema does ([schema](https://github.com/OpenRouterTeam/docs/blob/e657666ce82f285637d09625659fecd79236b98e/openapi/openapi.yaml#L12889-L12949)). OpenRouter's server implementation is not public, so its public sources do not conclusively prove the exact streaming payload. LiteLLM is therefore the verified reproducer for this PR; the fix also handles any Anthropic-compatible gateway that emits the same numeric extension.

## Change

- Read the untyped gateway extension defensively.
- Add `response_metadata.usage.cost` only when the value is a number.
- Leave native Anthropic streams unchanged when `cost` is absent or nonnumeric.
- Keep cost available independently of the `streamUsage` token-accounting option.
- Leave `usage_metadata`, cache-token handling, and `context_management` unchanged.
- Preserve the terminal cost when `AIMessageChunk`s are concatenated.

## Tests

Added unit coverage for:

- a gateway-provided numeric cost;
- an ordinary Anthropic `message_delta` without cost;
- a nonnumeric `cost` extension;
- `streamUsage: false`; and
- the final aggregated chunk after stream concatenation.

Validation:

- [x] `pnpm --filter @langchain/anthropic test` (206 tests, no type errors)
- [x] `pnpm lint`
- [x] `pnpm format:check`
- [x] `pnpm --filter @langchain/anthropic build`


hntrl (approved):
thanks for the pr!

Issue #9798, opened 2026-01-13
@langchain/anthropic streaming leaves `input_json_delta` chunks in AIMessage content, breaking checkpoint resume


simozampa:
### Checked other resources

- [x] This is a bug, not a usage question. For questions, please use the LangChain Forum (https://forum.langchain.com/).
- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain.js documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain.js rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

 ```typescript
  import { ChatAnthropic } from "@langchain/anthropic";
  import { StateGraph, Annotation, Command, interrupt } from "@langchain/langgraph";
  import { PostgresSaver } from "@langchain/langgraph-checkpoint-postgres";
  import { ToolNode } from "@langchain/langgraph/prebuilt";
  import { tool } from "@langchain/core/tools";
  import { z } from "zod";
  import { AIMessage, HumanMessage } from "@langchain/core/messages";

  // Simple tool for reproduction
  const myTool = tool(
    async ({ prompt }) => {
      return JSON.stringify({ result: prompt });
    },
    {
      name: "my_tool",
      description: "A test tool",
      schema: z.object({ prompt: z.string() }),
    }
  );

  const tools = [myTool];

  const chatModel = new ChatAnthropic({
    modelName: "claude-sonnet-4-20250514",
  });

  const modelWithTools = chatModel.bindTools(tools);

  const StateAnnotation = Annotation.Root({
    messages: Annotation<(AIMessage | HumanMessage)[]>({
      reducer: (x, y) => x.concat(y),
    }),
  });

  async function callModel(state: typeof StateAnnotation.State) {
    const response = await modelWithTools.invoke(state.messages);
    return { messages: [response] };
  }

  async function postToolReview(state: typeof StateAnnotation.State) {
    const feedback = interrupt({ message: "Continue?" });
    return new Command({
      goto: "agent",
      update: { messages: [new HumanMessage({ content: feedback })] },
    });
  }

  const graph = new StateGraph(StateAnnotation)
    .addNode("agent", callModel)
    .addNode("tools", new ToolNode(tools))
    .addNode("post_tool_review", postToolReview)
    .addEdge("__start__", "agent")
    .addConditionalEdges("agent", (state) => {
[...]

PR #11245, opened 2026-07-24
fix(openai): drop input_json_delta blocks in chat completions converter

sehee0207:
### What
`convertMessagesToCompletionsMessageParams` now drops `input_json_delta` content
blocks, alongside the `tool_use` / `tool_call` / `reasoning` /
`reasoning_content` / `thinking` blocks it already drops.

### Why
Anthropic streaming can leave `input_json_delta` blocks in `AIMessage.content`,
including after the message is checkpointed and restored. When such a message is
replayed to an OpenAI model, the raw blocks were passed through to the Chat
Completions request and rejected with:

`400 Invalid value: 'input_json_delta'. Supported values are: 'text', 'image_url', 'input_audio', 'refusal', 'audio', 'file'.`

For finalized messages where the consolidated call is already present in
`message.tool_calls`, dropping the residual delta does not lose tool-call
information.

This follows the provider-level fix in #9809 for `@langchain/anthropic`, which
removes the residual `input_json_delta` blocks and rebuilds each `tool_use` input
from `tool_calls` as the primary source (matched by `id`), falling back to
merging the delta chunks by `index` when `tool_calls` is unavailable. OpenAI
carries the finalized call in the separate `tool_calls` field rather than inside
content, so here the redundant delta blocks can simply be dropped. #9809 covered
the Anthropic→Anthropic path; this covers the Anthropic→OpenAI path.

### Repro (no network; converter only)
```ts
import { convertMessagesToCompletionsMessageParams } from "@langchain/openai/converters/completions";

const msg = new AIMessage({
  content: [
    { type: "text", text: "reading" },
    { type: "input_json_delta", index: 0, input: '{"file_path":"/README.md"}' } as any,
  ],
  tool_calls: [{ id: "call_1", name: "read_file", args: { file_path: "/README.md" } }],
});

const [param] = convertMessagesToCompletionsMessageParams({ messages: [msg] });
console.log(JSON.stringify(param.content, null, 2));
// before: content still contains { "type": "input_json_delta", ... } → OpenAI 400
// after:  input_json_delta removed from content; tool_calls preserved
```

### Why the provider converter
The change lives in the OpenAI converter because that is already where blocks the
Chat Completions API cannot accept are stripped (`tool_use`, `tool_call`,
`reasoning`, `reasoning_content`, `thinking`), so adding `input_json_delta` is a
natural extension of the converter's existing responsibility. It also matches the
provider-level precedent set by #9809, and keeps the change narrow and low-risk.

### Test
Added a converter test asserting `input_json_delta` is removed from content while
`tool_calls` is preserved.

Related: #9798, #9809

### DIFF
--- a/.changeset/openai-drop-input-json-delta.md
+++ b/.changeset/openai-drop-input-json-delta.md
@@ -1,0 +1,12 @@
+---
+"@langchain/openai": patch
+---
+
+Drop `input_json_delta` streaming-artifact content blocks when converting
+messages to Chat Completions params. Anthropic streaming can leave
+`input_json_delta` blocks in `AIMessage.content`, including after the message is
+checkpointed and restored; these were passed through raw and rejected by OpenAI
+with `400 Invalid value: 'input_json_delta'` — e.g. when a message authored by a
+Claude model is replayed to an OpenAI model. For finalized messages the
+consolidated tool call is already present in `message.tool_calls`, so dropping
+the residual delta does not lose tool-call information.
--- a/libs/providers/langchain-openai/src/converters/completions.ts
+++ b/libs/providers/langchain-openai/src/converters/completions.ts
@@ -811,17 +811,22 @@ export const convertMessagesToCompletionsMessageParams: Converter<
             //  - Tool-call blocks (`tool_use`, `tool_call`) are already
             //    carried in message.tool_calls, so resending them as content
             //    would be a duplicate/invalid part.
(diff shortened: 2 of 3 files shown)


wookingwoo:
The OpenAI content types actually supported are a fixed, small set per spec, so instead of this legacy path continuing to "blocklist known-bad types," it might be worth switching to "allowlist known-good types" — similar to `convertStandardContentMessageToCompletionsMessage`. That way we wouldn't need to patch this path every time a new provider artifact type shows up.

PR #11254, opened 2026-07-24
fix(langchain): reset run model call count on end

### LINKED ISSUE #11241: modelCallLimitMiddleware: beforeModel end-jump (exitBehavior: "end") skips the afterAgent runModelCallCount reset, locking out checkpointed threads
### Checked other resources

- [x] I searched existing issues and did not find a duplicate.

### Package / version

`langchain@1.5.3` (`langchain/agents` — `modelCallLimitMiddleware`), `@langchain/langgraph@1.4.7`.

### Summary

With a checkpointer and `modelCallLimitMiddleware({ runLimit, exitBehavior: "end" })`, once a thread hits its run limit the `runModelCallCount` state key is **never reset**, so **every subsequent invocation on that thread is immediately terminated after 0 model calls**. The per-run counter is effectively latched at the limit for the life of the thread.

The reset is implemented only in the middleware's `afterAgent` hook, but the `beforeModel` hook ends the run via `jumpTo: "end"` when the limit is reached. Ending the run at the `beforeModel` router routes straight to `END` without executing `afterAgent`, so the reset never runs on exactly the path that needs it. This contradicts the documented per-run semantics of `runModelCallCount` (it is meant to reset each run, unlike the persisted `threadModelCallCount`).

### Where

In the compiled middleware (`dist/agents/middleware/modelCallLimit.js`):

- `beforeModel` — when `runLimit <= runModelCallCount`, returns `{ jumpTo: "end", messages: [new AIMessage(<"Model call limits exceeded…">)] }` under `exitBehavior: "end"`.
- `afterModel` — increments `runModelCallCount`.
- `afterAgent` — returns `{ runModelCallCount: 0 }` (the only place the run counter is reset).

Because the `beforeModel` end-jump goes to `END` (the `afterModel`/`afterAgent` nodes are not on that path), the `afterAgent` reset is skipped whenever the run ends by hitting the limit. On a `MemorySaver` (or any checkpointer) the stale `runModelCallCount` persists into the next `invoke`/`stream` on the same `thread_id`, where `beforeModel` sees `runLimit <= runModelCallCount` on the very first step and ends immediately.

### Reproduction

```ts
import { createAgent, modelCallLimitMiddleware } from "langchain";
import { MemorySaver } from "@langchain/langgraph";
import { HumanMessage, AIMessage } from "@langchain/core/messages";

// A fake model that always emits a tool call, so the agent never stops on its own.
class AlwaysToolModel extends /* your BaseChatModel test double */ {
  calls = 0;
  async _generate() {
    this.calls++;
    return { generations: [{ text: "", message: new AIMessage({ content: "", tool_calls: [{ name: "noop", args: {}, id: `c${this.calls}` }] }) }] };
  }
}

const model = new AlwaysToolModel();
const agent = createAgent({
  model,
  tools: [/* a noop tool */],
  middleware: [modelCallLimitMiddleware({ runLimit: 2, exitBehavior: "end" })],
  checkpointer: new MemorySaver(),
});

const cfg = { configurable: { thread_id: "t1" }, recursionLimit: 100 };

await agent.invoke({ messages: [new HumanMessage("go")] }, cfg);
console.log(model.calls); // 2  ✅ (limit hit as expected)

await agent.invoke({ messages: [new HumanMessage("again")] }, cfg);
console.log(model.calls); // 2  ❌ expected 4 — the second run made ZERO model calls
```

### Expected behavior

`runModelCallCount` should be reset at the start of each run (or on the end-jump path), so a thread that exhausts its per-run budget once still gets a full budget on the next run. `exitBehavior: "end"` is the mode most likely to be used with a checkpointer (it avoids leaving an `AIMessage` with unfulfilled `tool_calls` in history), so it is precisely the affected combination.

### Actual behavior

`runModelCallCount` is latched at `runLimit` after the first exhausting run; every later run on the same checkpointed thread ends after 0 model calls.

### Workaround

Write `runModelCallCount: 0` into the per-turn stream/invoke **input** (it wins the `{ ...checkpointValues, ...input }` merge), restoring the budget each run. This couples app code to an internal state-key name, which we'd like to drop once the reset is fixed upstream.

### Suggested fix

Reset `runModelCallCount` on the `beforeModel` end-jump path as well (e.g. include `runModelCallCount: 0` in the state update returned alongside `jumpTo: "end"`), or move the per-run reset to a `beforeAgent` hook so it runs before the limit check regardless of how the previous run ended.

ousamabenyounes:
Fixes #11241.

## Summary

The model call limit middleware now clears the per-run model call counter when `exitBehavior: "end"` stops execution from `beforeModel`. This prevents a persisted checkpointer state from carrying a stale run-level count into the next invocation.

The regression test covers two invocations on the same checkpoint thread. It also asserts that `threadModelCallCount` keeps accumulating across checkpointed invocations, so the fix is scoped to the run-level counter only.

Added a patch changeset for the langchain package.

## Exit path audit

`modelCallLimit.ts` only accepts `exitBehavior: "error" | "end"`; the deprecated `"throw"` alias is normalized to `"error"`. The only two `jumpTo: "end"` branches in this middleware are the thread-limit and run-limit paths, and both now reset `runModelCallCount`.

## Test verification (RED -> GREEN)

RED before fix:

`pnpm --filter ./libs/langchain test src/agents/middleware/tests/modelCallLimit.test.ts -t "reset run count when exitBehavior end"`

Failed with:

`AssertionError: expected 1 to be +0`

GREEN after fix:

`pnpm --filter ./libs/langchain test src/agents/middleware/tests/modelCallLimit.test.ts -t "reset run count when exitBehavior end"`

Passed: 1 test file, 1 test.

## Full local suite

Full local suite after rebasing onto latest `origin/main`:

`BASE_REF=origin/main ./run-ci.sh`

Passed: format check, oxlint on changed files, langchain build, and `pnpm --filter ./libs/langchain test src/agents/middleware/tests/modelCallLimit.test.ts` with 7 tests passing.


ousamabenyounes:
Updated this PR in `001bb5767ebf53fbe512c9171f377f8b71f59d21` after rebasing onto the latest `main`.

I also audited `modelCallLimit.ts`: `exitBehavior` is only `"error" | "end"` (`"throw"` maps to `"error"`), and the only two `jumpTo: "end"` branches are the thread/run limit paths. The regression test now asserts that `runModelCallCount` resets while `threadModelCallCount` continues accumulating across checkpointed invocations.

Local validation: `BASE_REF=origin/main ./run-ci.sh` passed: format, oxlint on changed files, langchain build, and `modelCallLimit.test.ts` with 7 tests passing.

PR #11262, opened 2026-07-27
fix(google): preserve tool call ids in native stream events

### LINKED ISSUE #11261: @langchain/google native stream events drop tool call IDs
### Checked other resources

### Example Code

A focused source-level repro on current `main`:

```typescript
const events = await collectEvents([{
  candidates: [{
    content: {
      parts: [{
        functionCall: {
          id: "server-assigned-id-123",
          name: "web_search",
          args: { query: "weather" },
        },
      }],
    },
  }],
}]);

const start = events.find((event) => event.event === "content-block-start");
const finish = events.find((event) => event.event === "content-block-finish");

console.log(start?.content);
// { type: "tool_call_chunk", name: "web_search", args: "", index: 0 }

console.log(finish?.content);
// { type: "tool_call", name: "web_search", args: { query: "weather" }, id: undefined }
```

The same result occurs when the raw `functionCall` omits `id`: unlike the non-streaming converter, this path does not generate an `lc-tool-call-*` fallback.

No exception is thrown. The finalized tool_call content block silently has id: undefined.

### Description

`libs/providers/langchain-google/src/utils/stream_events.ts` does not put a tool-call ID on the block accumulator, `content-block-start`, or `content-block-delta` in the `functionCall` branch. `finalizeContentBlock` therefore produces a standard `tool_call` block without an ID.

The non-streaming converter in the same package already preserves `functionCall.id` and generates an `lc-tool-call-*` UUID fallback when Gemini omits it. Native stream events should have the same contract: one stable ID must survive from block start through block delta and finalization so a later `ToolMessage.tool_call_id` can match the assistant tool call.

This breaks agent loops using the native ChatModelStreamEvent path because the finalized tool call cannot be correlated with its tool result. The sibling `@langchain/google-genai` bug is being addressed by #11258, but that PR explicitly does not cover `@langchain/google`.

Expected: preserve a model-provided `functionCall.id`; otherwise generate the same `lc-tool-call-*` fallback used by `convertGeminiPartsToToolCalls`, and propagate it through start, delta, and finish events.

### System Info

@langchain/google: 0.2.1 / current main @ 710d74f3f
Platform: macOS Darwin 25.5.0 arm64
Node: v25.5.0
pnpm: 10.14.0

jackjin1997:
## Summary

Fixes #11261

- preserve a model-provided Gemini `functionCall.id` in native stream events
- generate the same `lc-tool-call-*` fallback used by the non-streaming converter when Gemini omits the ID
- propagate one stable ID through content-block start, delta, and finish events
- cover the converter, public `ChatGoogle.streamEvents` path, and a live API integration test

This is the `@langchain/google` sibling of #11258; that PR only changes `@langchain/google-genai`.

## Test plan

- `pnpm --filter @langchain/google exec vitest run src/utils/tests/stream_events.test.ts src/chat_models/tests/chat_models_stream_events.test.ts` — 10/10 passed, typecheck clean
- `pnpm --filter @langchain/google build` — passed, including publint and attw
- `pnpm exec oxfmt --check <changed files>` — passed
- `pnpm exec oxlint <changed TypeScript files>` — passed
- `pnpm --filter @langchain/google test` — 133 passed; 3 pre-existing failures remain in `converters/tests/messages.test.ts`, which this PR does not modify or import
- live API integration test added; not executed locally because `TEST_API_KEY` / `TEST_CREDENTIALS` are unavailable

Reverse verification: both focused ID tests fail on current `main` because `content-block-start` has no ID, then pass with this patch.

## AI Disclosure

AI assisted with code review, issue triage, test drafting, and implementation. The diff, reverse verification, and local checks were reviewed before submission.

### DIFF
--- a/.changeset/calm-tools-stream.md
+++ b/.changeset/calm-tools-stream.md
@@ -1,0 +1,5 @@
+---
+"@langchain/google": patch
+---
+
+Preserve tool call IDs in Google native stream events.
--- a/libs/providers/langchain-google/src/chat_models/tests/chat_models_stream_events.test.ts
+++ b/libs/providers/langchain-google/src/chat_models/tests/chat_models_stream_events.test.ts
@@ -91,6 +91,7 @@ const toolChunks: Gemini.GenerateContentResponse[] = [
           parts: [
             {
               functionCall: {
+                id: "call-weather-1",
                 name: "web_search",
                 args: { query: "weather" },
               },
@@ -138,7 +139,11 @@ describe("ChatGoogle.streamEvents", () => {
     await expect(
       mockChatGoogle(toolChunks).streamEvents("Hello")
     ).toHaveStreamToolCalls([
-      { name: "web_search", args: { query: "weather" } },
+      {
+        id: "call-weather-1",
+        name: "web_search",
+        args: { query: "weather" },
+      },
     ]);
   });
 
--- a/libs/providers/langchain-google/src/chat_models/tests/chat_models_stream_events.test.ts
+++ b/libs/providers/langchain-google/src/chat_models/tests/chat_models_stream_events.test.ts
@@ -140,11 +140,41 @@ describe("ChatGoogle.streamEvents", () => {
       mockChatGoogle(toolChunks).streamEvents("Hello")
     ).toHaveStreamToolCalls([
       {
-        id: "call-weather-1",
         name: "web_search",
         args: { query: "weather" },
       },
     ]);
+  });
+
+  test("preserves tool call ids in native events", async () => {
+    const events = [];
+    for await (const event of mockChatGoogle(toolChunks).streamEvents(
+      "Hello"
+    )) {
+      events.push(event);
+    }
+
+    expect(
+      events.find(
+        (event) =>
+          event.event === "content-block-start" &&
+          event.content.type === "tool_call_chunk"
+      )
+    ).toMatchObject({ content: { id: "call-weather-1" } });
+    expect(
+      events.find(
+        (event) =>
+          event.event === "content-block-delta" &&
+          event.delta.type === "block-delta"
+      )
+    ).toMatchObject({ delta: { fields: { id: "call-weather-1" } } });
+    expect(
+      events.find(
+        (event) =>
+          event.event === "content-block-finish" &&
+          event.content.type === "tool_call"
+      )
+    ).toMatchObject({ content: { id: "call-weather-1" } });
   });
 
   test("streams usage", async () => {
--- a/libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
+++ b/libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
@@ -426,50 +426,6 @@ const readPages = tool(async ({ fileId }) => `page text for ${fileId}`, {
   schema: z.object({ fileId: z.string() }),
 });
 
-describe("Google stream event tool calls", () => {
-  test("include a stable id from the live API path", async () => {
-    const model = new ChatGoogle({
-      model: "gemini-2.0-flash",
-      apiKey: getEnvironmentVariable("TEST_API_KEY"),
-    }).bindTools([weatherTool], {
-      tool_choice: "get_weather",
-    });
-
-    const events = [];
-    for await (const event of model.streamEvents(
-      "What is the weather in New York?"
-    )) {
-      events.push(event);
-    }
-
-    const toolCallStart = events.find(
-      (event) =>
-        event.event === "content-block-start" &&
-        event.content.type === "tool_call_chunk"
-    );
-    const toolCallFinish = events.find(
-      (event) =>
-        event.event === "content-block-finish" &&
-        event.content.type === "tool_call"
-    );
-
-    expect(toolCallStart).toMatchObject({
-      content: {
-        type: "tool_call_chunk",
-        id: expect.any(String),
-        name: "get_weather",
-      },
-    });
-    expect(toolCallFinish).toMatchObject({
-      content: {
-        type: "tool_call",
-        id: toolCallStart?.content.id,
-        name: "get_weather",
-      },
-    });
-  });
-});
-
 const coreModelInfo: ModelInfo[] = filterTestableModels([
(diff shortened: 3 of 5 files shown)


jackjin1997:
Quick follow-up after two weeks: the focused Google stream-event tests and visible PR checks are green. Is there a preferred reviewer or any additional integration evidence needed before this can move forward?

jackjin1997:
@afirstenberg, following up after another week: this includes unit coverage plus the real `ChatGoogle.streamEvents` integration path, while #11258 only changes `@langchain/google-genai`. All visible checks remain green. Is keeping the provider-specific fixes separate the right boundary, or should this be consolidated differently?

jackjin1997:
@afirstenberg, one week later, this and the newer #11453 still cover the same `@langchain/google` stream-event ID loss, while #11258 remains `@langchain/google-genai`-specific. This earlier PR includes both focused unit coverage and the real `ChatGoogle.streamEvents` integration path, with all visible checks green. Could you confirm whether this should be the consolidation target, or point me to the current Google reviewer?

jackjin1997:
Weekly status check: this remains mergeable with green visible checks, and the overlapping #11453 is also still open. Would you like this older PR kept as the consolidation target because it includes the `ChatGoogle.streamEvents` integration regression, or should I close it to reduce duplicate review surface?

jackjin1997:
@afirstenberg, following up at the weekly boundary: this earlier PR and #11453 are still both open, while this one carries the public `ChatGoogle.streamEvents` integration regression and the concurrent-call coverage. All visible checks remain green. Could you confirm whether to consolidate here or close it in favor of the newer path?

afirstenberg:
First of all, @jackjin1997, thank you for the repeated pings on these issues and keeping me on my toes.

I'm starting here since this one has an integration test while #11453 doesn't.

Unfortunately - at the moment, I can't get the integration test to compile. I'm still working on it, but wanted to make sure you're aware that I'm at least looking at the issue and trying to understand it.

jackjin1997:
Thanks, Allen — I reproduced the integration-test compile failure on `29e8cfa20`. I used `bindTools()`, which returns a `Runnable`: its `streamEvents` overload requires a version and returns tracing events, not the native content-block events this test needs. Adding `version: "v2"` would exercise the wrong protocol.

The correction is to call the chat model directly:

```ts
const llm = newChatGoogle();
const stream = llm.streamEvents("What is the weather in New York?", {
  tools: [weatherTool],
  tool_choice: "get_weather",
});
```

I have corrected that locally, narrowed the start/finish event types before reading IDs, and moved the credentialed case into the existing model matrix and `newChatGoogle()` setup. I also found that `toHaveStreamToolCalls` does not accept an `id` expectation; the public-path unit test now explicitly checks the ID on start, delta, and finish events instead. The five diagnostics in the two changed test files are gone, and the focused suite is 12/12 passing.

The update is not pushed yet because the full local validation gate is still red. After rebasing locally onto `b0a0d3f7`, the package build (including publint/attw), lint, and formatting pass, but the full suite is 179 passed / 6 failed. Running unmodified `b0a0d3f7` gives 175 passed / the same 6 failures, all in `src/converters/tests/messages.test.ts`; raw package-wide `tsc` also reports errors outside these two test files. I have kept those separate from this fix rather than changing unrelated converter behavior or using CI to debug it. No live API execution is claimed: I do not have the integration credentials locally.

jackjin1997:
Pushed `74a5eddb7` — the integration test compiles now.

The failure came from `bindTools()`: it returns a `RunnableBinding`, whose `streamEvents` overloads all require `version: "v1" | "v2"` and emit callback tracing events rather than the native `content-block-*` events this test asserts on. Adding `version: "v2"` would have compiled but exercised the wrong protocol, so the test now calls the chat model directly and passes `tools` / `tool_choice` as call options — they reach the request through `combineGoogleChatModelFields` → `invocationParams`, the same path `bindTools` uses.

Two related notes from going through the compile errors:

- `toHaveStreamToolCalls` only compares `name` and `args` (`StreamToolCallExpectation` has no `id` field), so the `id` in that unit-test expectation was a type error that never asserted anything. Tool call ids are now asserted on the native `content-block-start` / `content-block-delta` / `content-block-finish` events, plus a case for two concurrently streamed calls.
- The branch is rebased onto current main (`b0a0d3f73`), hence the force-push.

Locally: the focused stream-event suites pass 12/12, `oxfmt --check` and `oxlint` are clean, and reverting only `src/utils/stream_events.ts` to main makes 4 of those tests fail, so the coverage really does exercise the fix. I have no `TEST_API_KEY` here, so the integration test itself is still yours to confirm — `@langchain/google` is not in the unit-test CI matrix either, so nothing upstream will run it for us.

jackjin1997:
One more thing before you sink more time into this branch: #11602 landed a few hours ago on the same `convertGoogleGeminiStream` id path, additionally carries `thoughtSignature` through to `tool_calls`, and lists #11261 under Fixes — so it is a superset of what this PR fixes. I'd rather not have two PRs competing for the same review slot; if you prefer to consolidate there, say the word and I'll close this one.

The only coverage here that #11602 doesn't have is two concurrently streamed function calls keeping distinct ids (separate `tool:${toolIdx}` accumulator keys) and the explicit id assertions on the native start/delta/finish events. That's portable as a unit test if you want it. FWIW the compile fix I pushed (`74a5eddb7`) uses the same direct `streamEvents(input, { tools, tool_choice })` shape that #11602's integration test uses, so the two branches agree on that part.

afirstenberg:
#11602 is by @thushanth-bengre-langchain, who works at Langchain, so we will likely go with that one. Plus it adds the `thoughtSignature` fixes, which have been our big focus.

Thanks for pointing it out, @jackjin1997, and for all the work you've put into this!

thushanth-bengre-langchain:
Thank you for this PR @jackjin1997.

We will go with #11602 for this one as it addresses multiple issues.

I have ported your tests to that PR and added you as a co-author there.

jackjin1997:
Closing per Allen's call above — #11602 carries the superset fix (tool call id + thoughtSignature), and thushanth-bengre-langchain already ported this branch's two integration tests over in cd4e997a9, so the concurrent-call coverage lives on there. Happy to reopen if the consolidation direction changes.

PR #11264, opened 2026-07-28
fix(langchain): propagate `control` through createAgent options

sysradium:
`RunControl` ([`@langchain/langgraph` ≥ 1.4.0](https://docs.langchain.com/oss/javascript/langgraph/fault-tolerance#graceful-shutdown)) lets a caller request a cooperative drain, so a run stops at the next superstep boundary with its checkpoint committed and can be resumed later. The documented use is hooking SIGTERM for a graceful shutdown, which is exactly what an agent serving long-running work during a k8s rolling deploy needs in our case.

It can't be passed to a `createAgent` agent without a cast:

```ts
const control = new RunControl();

await agent.stream(input, { ...config, streamMode: "updates", control });
//                                                             ^^^^^^^
// TS2353: Object literal may only specify known properties, and 'control'
// does not exist in type 'Partial<Pick<PregelOptions<...>,
// CreateAgentPregelOptions | CreateAgentPregelStreamOptions>> & {...}'.
```

`control` is a `PregelOptions` member, and `ReactAgent` already forwards it  `invoke`/`stream` merge via `mergeConfigs`, which preserves unknown keys, then pass the merged config straight to the compiled graph:

```ts
async stream(state, config) {
  const mergedConfig = mergeConfigs(this.#defaultConfig, config);
  const initializedState = await this.#initializeMiddlewareStates(state, mergedConfig);
  return this.#graph.stream(initializedState, mergedConfig);
}
```

So adding `control` to `CreateAgentPregelOptions` addresses the problem.


sysradium:
Anyone?

PR #11266, opened 2026-07-29
fix(google-common): serialize tool choice modes for Vertex AI

### LINKED ISSUE #11265: bug(@langchain/google-common): `functionCallingConfig.mode` is serialized lowercase, which Vertex AI silently ignores — `tool_choice` ("any"/"none"/forced function name) is a no-op on ChatVertexAI
### Checked other resources

### Example Code

```typescript

  model: "gemini-2.5-flash",
  temperature: 0,
}).withConfig({
});

```

_No response_

### Description

forced-function-name case:

```ts
  config = {
    functionCallingConfig: {
      ...
```

mis-set):

`tool_choice` does anything.

### Suggested fix

```ts
```

### System Info

- Node.js 22.x

jackjin1997:
## Summary
Fixes #11265
- serialize user-facing `auto`, `any`, and `none` tool choices as the uppercase Vertex AI wire enums
- use `ANY` for forced function names while preserving `allowedFunctionNames`
- cover all modes with request-formatting tests and assert the live `ChatVertexAI` request payload in the existing integration test

## Test plan
- `pnpm --filter @langchain/google-common test` — 8 files, 298 tests passed; typecheck clean
- `pnpm --filter @langchain/google-common --filter @langchain/google-vertexai build` — passed
- `pnpm exec oxfmt --check <changed files>` — passed
- `pnpm exec oxlint <changed TypeScript files>` — passed
- live Vertex AI assertion added to the existing forced-tool integration test; not run locally because GCP credentials are unavailable

Reverse verification: the four focused request-formatting cases fail on current `main` with lowercase modes, then pass with this patch.

## AI Disclosure
AI assisted with issue triage, test drafting, and implementation. The diff, reverse verification, and local checks were reviewed before submission.

### DIFF
--- a/.changeset/calm-tools-call.md
+++ b/.changeset/calm-tools-call.md
@@ -1,0 +1,5 @@
+---
+"@langchain/google-common": patch
+---
+
+fix Google tool choice modes sent to Vertex AI
--- a/libs/providers/langchain-google-common/src/tests/utils.test.ts
+++ b/libs/providers/langchain-google-common/src/tests/utils.test.ts
@@ -848,6 +848,33 @@ describe("streaming", () => {
   });
 });
 
+describe("gemini tool config formatting", () => {
+  test.each([
+    ["auto", "AUTO"],
+    ["any", "ANY"],
+    ["none", "NONE"],
+  ] as const)("serializes %s mode as %s", async (toolChoice, expectedMode) => {
+    const api = getGeminiAPI();
+    const request = (await api.formatData([new HumanMessage("Use tools.")], {
+      tool_choice: toolChoice,
+    })) as GeminiRequest;
+
+    expect(request.toolConfig?.functionCallingConfig?.mode).toBe(expectedMode);
+  });
+
+  test("serializes a forced function name as ANY", async () => {
+    const api = getGeminiAPI();
+    const request = (await api.formatData([new HumanMessage("Use tools.")], {
+      tool_choice: "lookup_weather",
+    })) as GeminiRequest;
+
+    expect(request.toolConfig?.functionCallingConfig).toEqual({
+      mode: "ANY",
+      allowedFunctionNames: ["lookup_weather"],
+    });
+  });
+});
+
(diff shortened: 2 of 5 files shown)


jackjin1997:
This has been open for a little over two weeks, with the focused request-formatting coverage, a live Vertex assertion, and all visible checks green. There are now two later implementations of the same issue (#11288 and #11323); could a maintainer confirm which PR should be kept so the duplicate review paths can be consolidated?

jackjin1997:
@afirstenberg, following up after another week: this is still the earliest of the three implementations (#11288 and #11323 are the later duplicates), includes the live Vertex assertion, and all visible checks remain green. Could you confirm whether this should be the consolidation target, or whether you would prefer a different scope before review?

jackjin1997:
@afirstenberg, closing the loop after another week: neither later implementation has received a review decision, while this remains the earliest one with the live Vertex assertion and green visible checks. Could you choose the consolidation target or point me to the current owner? I can rebase or adjust scope immediately once the preferred path is clear.

jackjin1997:
@afirstenberg, another week has passed with all three implementations still open and no consolidation decision. This remains the earliest path and includes the live Vertex assertion with green visible checks. Could you choose the review target or point me to the current Google owner so the duplicate surface can be reduced?

afirstenberg:
Sorry, @jackjin1997 - We've mostly been focusing on the `@langchain/google` package and getting it fully working to eventually replace the `@langchain/google-common` packages as well as the `@langchaiin/google-genai` package.

Is there a reason that moving to `@langchain/google` is infeasible at this time? (A quick glance there does suggest we're handling tool_choice / mode correctly there.)

That said - #LGTM @hntrl @thushanth-bengre-langchain

jackjin1997:
Thanks, @afirstenberg — this was mainly for existing `@langchain/google-common` users who cannot migrate atomically. I agree that `@langchain/google` is the destination and was not suggesting the migration is infeasible long-term. I appreciate you merging the compatibility fix.

Issue #11110, opened 2026-06-25
`contentBlocks` sets the Gemini `tool_call` id from `message.id` instead of `functionCall.id`


elierotenberg:
### Checked other resources

- [x] This is a bug, not a usage question. For questions, please use the LangChain Forum (https://forum.langchain.com/).
- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain.js documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain.js rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```typescript
import { AIMessage } from "@langchain/core/messages";

const message = new AIMessage({
  content: [
    { type: "functionCall", functionCall: { id: "abc123", name: "calculator", args: { expression: "1 + 1" } } },
  ],
  tool_calls: [
    { type: "tool_call", id: "abc123", name: "calculator", args: { expression: "1 + 1" } },
  ],
  response_metadata: { model_provider: "google" },
});

console.log("tool_calls[0].id:", message.tool_calls[0].id);
console.log("contentBlocks[0].id:", message.contentBlocks[0].id);
```

Output:

```
tool_calls[0].id: abc123
contentBlocks[0].id: undefined
```

### Error Message and Stack Trace (if applicable)

Not directly applicable although downstream code expecting `contentBlock.id` won't work.

### Description

`AIMessage.contentBlocks` returns Gemini `tool_call` blocks without id, while `AIMessage.tool_calls` has it.

The problems seems to come from [`google_genai.ts`](https://github.com/langchain-ai/langchainjs/blob/f1d64ff8d6076fb6656c1f327b7b58eac520314b/libs/langchain-core/src/messages/block_translators/google_genai.ts#L50) (and equivalent in google.ts):

```typescript
return {
  type: "tool_call",
  id: message.id, // should be block.functionCall.id
  name: block.functionCall.name,
  args: block.functionCall.args,
};
```

For the record, the following pnpm patch fixed the issue:

```patch
diff --git a/dist/messages/block_translators/google.cjs b/dist/messages/block_translators/google.cjs
index b2ab67138f019884d80bf6b116f25c74b78a77ba..aeff2f40c3a364f2d8e9eb27fd03b50ed20352f7 100644
--- a/dist/messages/block_translators/google.cjs
+++ b/dist/messages/block_translators/google.cjs
[...]

PR #11271, opened 2026-07-30
fix(google): preserve streamed tool call signatures

syncsiva:
## Summary

- preserve Gemini thought signatures when Google provider messages are translated to v1 standard content blocks
- replay signed v1 tool calls as a single Gemini function call
- retain the provider tool call ID so standard content and `tool_calls` stay aligned

## Root cause

The `@langchain/google` v1 conversion path rebuilt Gemini function calls from `message.tool_calls`, which omitted the `thoughtSignature` stored on the standardized content block. The Google block translator also used the message ID instead of the corresponding tool call ID, allowing the same call to be represented twice.

Gemini 3 requires the original thought signature when a streamed tool call is returned in conversation history. Losing it caused the subsequent tool-result request to fail with `Function call is missing a thought_signature in functionCall parts`.

## Validation

- `pnpm --filter @langchain/google exec vitest run src/converters/tests/messages.test.ts --reporter=verbose`
- `pnpm --filter @langchain/core exec vitest run src/messages/block_translators/tests/google.test.ts --reporter=verbose`
- `pnpm --filter @langchain/google build`
- targeted `oxfmt --check`
- `git diff --check`


syncsiva:
Fixes https://github.com/langchain-ai/langchainjs/issues/11110

PR #11275, opened 2026-07-31
fix(langchain-mongodb): route Atlas-provisioned Voyage keys to MongoDB AI endpoint

okxint:
MongoDB Atlas–provisioned Voyage AI API keys start with `al-` and must be
routed to `https://ai.mongodb.com/v1` instead of the default
`https://api.voyageai.com/v1`. The Python `VoyageEmbeddings` class already
implements this routing via `get_default_base_url(api_key)`, but the JS
class always uses the Voyage AI endpoint regardless of key prefix.

**Fix:** detect the `al-` prefix in the constructor and set the base URL
accordingly, unless the caller explicitly passed a `basePath` override.

The class-field default `basePath = "https://api.voyageai.com/v1"` is
removed since the constructor now always assigns deterministically.

### DIFF
--- a/.changeset/fix-voyage-mongodb-atlas-base-url.md
+++ b/.changeset/fix-voyage-mongodb-atlas-base-url.md
@@ -1,0 +1,5 @@
+---
+"@langchain/mongodb": patch
+---
+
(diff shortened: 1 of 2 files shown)


okxint:
Added a changeset in the latest commit (`.changeset/fix-voyage-mongodb-atlas-base-url.md`) to resolve the changeset-bot warning.

PR #11278, opened 2026-07-31
fix(core): redact MCP tracing credentials

jkennedyvz:
Remote MCP credentials were included in chat-model callback metadata and could be persisted in LangSmith traces. OpenAI MCP tools exposed `headers` and `authorization`; Anthropic MCP server definitions exposed `authorization_token`.

This adds a tracing-only invocation-parameter hook in `BaseChatModel`, uses it for streaming, uncached, and cached callback starts, and sanitizes corresponding callback options. OpenAI and Anthropic override the hook to redact their MCP credential fields without mutating caller input or provider request parameters.

Provider unit tests verify trace payloads are redacted while request parameters retain the original credentials.

## Release note

OpenAI and Anthropic chat-model traces now redact credentials from remote MCP tool and server definitions.

This contribution was prepared with AI-agent assistance.

### DIFF
--- a/libs/providers/langchain-openai/src/chat_models/base.ts
+++ b/libs/providers/langchain-openai/src/chat_models/base.ts
@@ -70,6 +70,8 @@ import {
   createFunctionCallingParser,
 } from "@langchain/core/language_models/structured_output";
 
+const MCP_CREDENTIALS_REDACTED = "**REDACTED**";
+
 interface OpenAILLMOutput {
   tokenUsage: {
     completionTokens?: number;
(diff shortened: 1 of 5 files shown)


libs/providers/langchain-openai/src/chat_models/base.ts
```diff
@@ -438,6 +438,61 @@ export abstract class BaseChatOpenAI<
...
+      tools: params.tools.map((tool) => {
+        if (tool?.type !== "mcp") {
+          return tool;
+        }
+        const redactedTool: Record<string, unknown> = { ...tool };
+        if ("headers" in redactedTool) {
+          redactedTool.headers = "**REDACTED**";
+        }
+        if ("authorization" in redactedTool) {
+          redactedTool.authorization = "**REDACTED**";
```
christian-bromann: Can we move `"**REDACTED**"` into a constant and reuse it?

christian-bromann (dismissed):
One comment, otherwise LGTM

libs/providers/langchain-openai/src/chat_models/base.ts
jkennedyvz: Addressed in 3bb1658. The MCP credential redaction placeholder now uses a reusable constant in both affected provider implementations.

PR #11288, opened 2026-08-02
fix(google-common): send functionCallingConfig.mode as the uppercase Gemini enum

### LINKED ISSUE #11265: bug(@langchain/google-common): `functionCallingConfig.mode` is serialized lowercase, which Vertex AI silently ignores — `tool_choice` ("any"/"none"/forced function name) is a no-op on ChatVertexAI

onatozmenn:
Fixes #11265

## Problem

`formatToolConfig` in `@langchain/google-common` serializes `toolConfig.functionCallingConfig.mode` with the lowercase user-facing values (`"auto"` / `"any"` / `"none"`), while the Gemini/Vertex API defines `FunctionCallingConfig.mode` as an enum whose values are `AUTO` / `ANY` / `NONE`. The lowercase value is not honored, so requests fall back to default `AUTO` semantics:

- `tool_choice: "any"` (and `tool_choice: "<function name>"`, which routes through the same code path) does not force a tool call — the model may answer with plain text instead, and the call loses ANY-mode's schema anchoring.
- `tool_choice: "none"` does not suppress tool calls.
- `allowed_function_names` is not enforced, since it only applies together with a valid mode.

Nothing errors, which is what makes this hard to spot: the model usually calls the tool anyway when the prompt asks for it, so the behavior just degrades intermittently.

The newer `@langchain/google` package already sends the enum values correctly (`libs/providers/langchain-google/src/converters/tools.ts`, typed from `gemini-openapi.json`), as does `@langchain/google-genai` via `FunctionCallingMode`, so this was the odd one out among the Google wrappers.

## Change

Map the lowercase values to the wire enum in `formatToolConfig`, mirroring how `formatGenerationConfig` in the same file already maps `reasoningLevel` to the uppercase `thinkingLevel` values, and update the `GeminiRequest["toolConfig"]` type accordingly. The forced-function-name branch now sends `mode: "ANY"` instead of `"any"`.

The public API is unchanged: callers keep passing lowercase `tool_choice` values, which are normalized in `processToolChoice` as before — only the serialized request body changes.

## Tests

Added mock-transport tests in `chat_models.test.ts` asserting the request body:

- `tool_choice` `auto` / `any` / `none` produce `mode: "AUTO"` / `"ANY"` / `"NONE"`
- `tool_choice: "my_tool"` produces `{ mode: "ANY", allowedFunctionNames: ["my_tool"] }`

`vitest run src/tests/chat_models.test.ts` → 105 passed (101 existing + 4 new). No existing test asserted the lowercase spelling.

### DIFF
--- a/.changeset/tender-moons-repeat.md
+++ b/.changeset/tender-moons-repeat.md
@@ -1,0 +1,5 @@
+---
+"@langchain/google-common": patch
+---
+
+Send `toolConfig.functionCallingConfig.mode` using the uppercase Gemini enum values (`AUTO` / `ANY` / `NONE`). The lowercase values sent previously were ignored by the API, which made `tool_choice` (including forcing a specific function) a silent no-op.
--- a/libs/providers/langchain-google-common/src/tests/chat_models.test.ts
+++ b/libs/providers/langchain-google-common/src/tests/chat_models.test.ts
@@ -2865,6 +2865,76 @@ describe("Mock ChatGoogle - Gemini", () => {
     ).toBeUndefined();
   });
 
+  test.each([
+    ["auto", "AUTO"],
+    ["any", "ANY"],
+    ["none", "NONE"],
+  ])(
+    "6. tool_choice %s is sent as the %s enum value",
+    async (toolChoice, expectedMode) => {
+      const record: Record<string, any> = {};
+      const projectId = mockId();
+      const authOptions: MockClientAuthInfo = {
+        record,
+        projectId,
+        resultFile: "chat-6-mock.json",
+      };
+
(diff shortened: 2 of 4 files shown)


thushanth-bengre-langchain:
Thank you for the PR. Going with #11266 for this bugfix.

PR #11292, opened 2026-08-03
fix(core): format every data content block for tracing

### LINKED ISSUE #11291: _formatForTracing only converts the first base64 content block per message
### Checked other resources

### Example Code

```typescript

  name = "CaptureInputHandler";

  async handleChatModelStart(
    _llm: Serialized,
    messages: BaseMessage[][],
  ): Promise<void> {
    this.messages = messages[0];
  }
}

  content: [
    {
      type: "file",
      source_type: "base64",
      mime_type: "application/pdf",
      data: "JVBERi0xLjQK",
    },
    {
      type: "file",
      source_type: "base64",
      mime_type: "application/pdf",
      data: "JVBERi0xLjUK",
    },
  ],
});

  callbacks: [handler],
});

```

_No response_

### Description

### Actual output

```json
[
  {
    "type": "text",
    "text": "two PDFs"
  },
  {
    "type": "image_url",
    "image_url": {
    }
  },
  {
    "type": "file",
    "source_type": "base64",
    "data": "JVBERi0xLjUK"
  }
]
```

### Expected output

```json
[
  {
    "type": "text",
    "text": "two PDFs"
  },
  {
    "type": "image_url",
    "image_url": {
    }
  },
  {
    "type": "image_url",
    "image_url": {
    }
  }
]
```

### Root cause

### Docs / shape note

|---|---|---|

### Proposed fix

### System Info

```
langchain: 1.5.3
Platform: macOS
Node: 24.18.0
pnpm: 11.15.1
```

jackjin1997:
## Summary

Fixes #11291
- convert every URL or base64 data content block in the tracing copy
- preserve the original message while avoiding repeated message reconstruction
- add a callback-path regression test with multiple attachments

## Testing

- pnpm --filter @langchain/core test
- pnpm --filter @langchain/core build
- oxfmt --check on changed files
- oxlint on changed TypeScript files

## AI Disclosure

This bug was fixed with AI assistance. The regression test was written first and verified to fail on the previous implementation before the production change.

### DIFF
--- a/.changeset/bright-geese-trace.md
+++ b/.changeset/bright-geese-trace.md
@@ -1,0 +1,5 @@
+---
+"@langchain/core": patch
+---
+
+fix(core): format every data content block for tracing
--- a/libs/langchain-core/src/language_models/chat_models.ts
+++ b/libs/langchain-core/src/language_models/chat_models.ts
@@ -172,24 +172,22 @@ function _formatForTracing(messages: BaseMessage[]): BaseMessage[] {
   const messagesToTrace: BaseMessage[] = [];
   for (const message of messages) {
     let messageToTrace = message;
-    if (Array.isArray(message.content)) {
-      for (let idx = 0; idx < message.content.length; idx++) {
-        const block = message.content[idx];
-        if (isURLContentBlock(block) || isBase64ContentBlock(block)) {
-          if (messageToTrace === message) {
-            // Also shallow-copy content
-            // oxlint-disable-next-line @typescript-eslint/no-explicit-any
-            messageToTrace = new (message.constructor as any)({
-              ...messageToTrace,
-              content: [
-                ...message.content.slice(0, idx),
-                convertToOpenAIImageBlock(block),
-                ...message.content.slice(idx + 1),
-              ],
-            });
-          }
-        }
-      }
+    if (
+      Array.isArray(message.content) &&
+      message.content.some(
+        (block) => isURLContentBlock(block) || isBase64ContentBlock(block)
+      )
+    ) {
+      // Also shallow-copy content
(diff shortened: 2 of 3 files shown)


jackjin1997:
Now that this has crossed two weeks: the focused tracing regression, the full local core suite, and all visible checks are green. The later #11338 has not surfaced a distinct behavior beyond this patch. Is there a preferred reviewer for the tracing conversion path, or any additional coverage you would like before review?

jackjin1997:
@christian-bromann, could you help route this to the current core/tracing reviewer? It has now been another week since the last review request, and the author of the later duplicate #11450 has agreed to consolidate into this earlier PR. The focused regression, full local core suite, and visible checks remain green; I can address any scope or coverage concern promptly.

jackjin1997:
@christian-bromann, this core tracing fix has now gone another week without a reviewer response. The later duplicate author agreed to consolidate here, and the focused regression plus visible checks remain green. Could you route it to the current tracing owner, or flag any scope concern that should be resolved first?

jackjin1997:
@christian-bromann, following up at the weekly boundary on the tracing conversion fix. The later duplicate author already agreed to consolidate here, and the focused regression plus visible checks remain green. Could you route this to the tracing owner or call out any scope issue that blocks review?

jackjin1997:
A fifth implementation for #11291 has now appeared (#11613), on top of #11297, #11338, and #11450. To reduce an already crowded review queue, I am closing this older PR despite it having been the prior consolidation target. It remains the version with the required changeset, callback-path regression, and full local core validation; happy to reopen it if maintainers prefer the earliest complete implementation or want any of its coverage carried forward.

PR #11294, opened 2026-08-03
fix(core): merge Anthropic input JSON deltas

### LINKED ISSUE #11293: [@langchain/core 1.2.0] hasMismatchedMergeableType breaks tool_call streaming for Anthropic models via ChatOpenAI
## Description

`@langchain/core 1.2.0` introduced `hasMismatchedMergeableType` in `_findMergeTarget` ([`libs/langchain-core/src/messages/base.ts`](https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-core/src/messages/base.ts)). This guard prevents merging streaming chunks that have different `type` fields at the same index — but it breaks tool call streaming for **Anthropic models routed through `ChatOpenAI`** (the OpenAI-compatible endpoint).

## Root cause

When Anthropic streams a tool call via the OpenAI-compat path, the chunks look like this:

| chunk | `index` | `type` |
|---|---|---|
| content block start | `0` | `"tool_use"` |
| argument delta | `0` | `"input_json_delta"` |

Both chunks share `index: 0`, but their `type` values differ. `getMergeableTypeBase` strips the `_delta` suffix before comparing:

```
getMergeableTypeBase("tool_use")         → "tool_use"
getMergeableTypeBase("input_json_delta") → "input_json"
```

`"tool_use" !== "input_json"` → `hasMismatchedMergeableType` returns `true` → merge blocked.

The delta chunks are pushed as separate entries instead of accumulating into the base chunk. `AIMessage.tool_calls` ends up empty after streaming aggregation, so any code that routes on `tool_calls` (e.g. a LangGraph `conversationAgentRouter`) always sees zero tool calls and skips tool execution entirely.

## What changed

This function **did not exist in 1.1.36**. The regression is a direct result of the new check added in 1.2.0.

## Proposed fix

A `_delta` item is always intended to update an existing entry — its type name intentionally describes the *kind of delta*, not the *type of the target*. The guard should skip the check when the incoming item is a delta:

```typescript
function hasMismatchedMergeableType(left: unknown, right: unknown): boolean {
  if (typeof left !== "object" || left === null) return false;
  if (typeof right !== "object" || right === null) return false;
  if (!("type" in left) || !("type" in right)) return false;

  // A "_delta" item is always meant to update an existing entry, regardless of
  // whether the type-name prefix matches (e.g. Anthropic streams `input_json_delta`
  // chunks into a `tool_use` block — different prefixes, but a legitimate merge).
  if (typeof right.type === "string" && right.type.endsWith("_delta")) return false;

  return (
    typeof left.type === "string" &&
    typeof right.type === "string" &&
    getMergeableTypeBase(left.type) !== getMergeableTypeBase(right.type)
  );
}
```

This preserves the original intent of the guard (preventing e.g. `text` blocks from merging with `tool_use` blocks at the same index) while allowing Anthropic's streaming format to work correctly.

## Reproduction

Use `ChatOpenAI` pointing at an Anthropic-compatible endpoint (e.g. AWS Bedrock, or any proxy that forwards to Anthropic but exposes an OpenAI-compatible API). Call `.invoke()` with tools bound. In 1.1.36 `response.tool_calls` is populated; in 1.2.0 it is empty.

## Version

- `@langchain/core`: `1.2.0` (regression vs `1.1.36`)

jackjin1997:
## Summary

Fixes #11293

- allow Anthropic input_json_delta chunks to merge into their indexed tool_use content blocks
- preserve the semantic type guard for unrelated content block types
- add a focused core regression test and patch changeset

The same protocol compatibility was proposed in #10923. This follow-up isolates the core change and regression test against current main; #10923 currently also contains unrelated provider changes.

## Testing

- pnpm exec vitest run --typecheck.enabled=false (1367 passed, 1 skipped)
- pnpm --filter @langchain/core build
- pnpm exec oxfmt --check on changed TypeScript files
- pnpm exec oxlint on changed TypeScript files

## AI Disclosure

This contribution was prepared with AI-agent assistance.

### DIFF
--- a/.changeset/tidy-tools-stream.md
+++ b/.changeset/tidy-tools-stream.md
@@ -1,0 +1,5 @@
+---
+"@langchain/core": patch
+---
+
+Merge Anthropic input JSON deltas into their indexed tool-use content blocks.
--- a/libs/langchain-core/src/messages/base.ts
+++ b/libs/langchain-core/src/messages/base.ts
@@ -573,12 +573,14 @@ function hasMismatchedMergeableType(left: unknown, right: unknown): boolean {
   if (typeof left !== "object" || left === null) return false;
   if (typeof right !== "object" || right === null) return false;
   if (!("type" in left) || !("type" in right)) return false;
+  if (typeof left.type !== "string" || typeof right.type !== "string") {
+    return false;
+  }
+  if (left.type === "tool_use" && right.type === "input_json_delta") {
+    return false;
+  }
 
-  return (
-    typeof left.type === "string" &&
-    typeof right.type === "string" &&
-    getMergeableTypeBase(left.type) !== getMergeableTypeBase(right.type)
-  );
+  return getMergeableTypeBase(left.type) !== getMergeableTypeBase(right.type);
 }
 
 /**
--- a/libs/langchain-core/src/messages/tests/base_message.test.ts
+++ b/libs/langchain-core/src/messages/tests/base_message.test.ts
@@ -248,6 +248,42 @@ test("_mergeLists keeps different content block types separate when indices coll
   ]);
 });
 
+test("_mergeLists merges Anthropic input JSON deltas into tool use blocks", () => {
+  const chunk1 = new AIMessageChunk({
+    content: [
+      {
+        type: "tool_use",
+        index: 0,
+        id: "toolu_abc123",
+        name: "get_weather",
+        input: {},
+      },
+    ],
+  });
+
+  const chunk2 = new AIMessageChunk({
+    content: [
+      {
+        type: "input_json_delta",
+        index: 0,
+        partial_json: '{"location":"San Francisco"}',
+      },
+    ],
+  });
+
+  const merged = chunk1.concat(chunk2);
+  expect(merged.content).toEqual([
+    {
+      type: "tool_use",
+      index: 0,
+      id: "toolu_abc123",
+      name: "get_weather",
+      input: {},
+      partial_json: '{"location":"San Francisco"}',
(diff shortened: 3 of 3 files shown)


jackjin1997:
Now that this has been open for two weeks: the focused core regression, full local core suite, and all visible checks are green. This also keeps the fix isolated from the broader provider changes in #10923. Is there a preferred reviewer for the content-block aggregation path, or any additional provider-level coverage you would like before review?

jackjin1997:
@christian-bromann, could you help route this to the current core messages reviewer? It has been another week since the last review request. The patch remains isolated from the broader provider work in #10923, with the focused aggregation regression, full local core suite, and visible checks green; I can add any provider-level evidence the reviewer considers necessary.

jackjin1997:
@christian-bromann, following up after another week on the core content-block aggregation path. The change remains isolated from the provider work, with the focused regression and visible checks green. Could you route it to the current messages reviewer, or identify any provider-level evidence still needed?

jackjin1997:
@christian-bromann, another week has passed on the core input-JSON aggregation fix. It remains isolated from the broader provider PR, with focused regression coverage and green visible checks. Could you point it to the current messages reviewer or identify any remaining provider-level evidence?

jackjin1997:
@hntrl since you have been reviewing the core streaming paths recently, would you mind looking at this one? It fixes how Anthropic input JSON deltas are merged during content-block aggregation, with a focused core regression and the full local core suite green. It has been open since early August without a reviewer, and it stays independent of the broader provider work in #10923.

PR #11297, opened 2026-08-04
fix(core): convert all image content blocks when formatting for tracing

### LINKED ISSUE #11291: _formatForTracing only converts the first base64 content block per message

anshusaurav:
## Description

`_formatForTracing` (in `libs/langchain-core/src/language_models/chat_models.ts`) only converted the **first** base64/URL content block in a multimodal message. After the first matching block triggered a shallow copy of the message, the `messageToTrace === message` guard was `false` for every subsequent block, so any additional image blocks reached tracing (e.g. LangSmith) unconverted.

This rewrites the loop to map over the content once: every URL/base64 block is converted via `convertToOpenAIImageBlock`, and the message is shallow-copied a single time only when a block actually changed (preserving the original copy-once optimization).

## Changes

- Fix `_formatForTracing` to convert **all** image content blocks, not just the first.
- Add a regression test in `chat_models.test.ts` that traces a message containing two base64 image blocks (via a `handleChatModelStart` capture handler) and asserts both are converted, while the original message object is left untouched.

## Testing

- New test fails on `main` (second block remains a raw `source_type: "base64"` block) and passes with this change.
- Full `libs/langchain-core/.../chat_models.test.ts` suite passes (23/23), type-check clean, `oxlint` clean.

Fixes #11291

### DIFF
--- a/.changeset/tidy-trace-image-blocks.md
+++ b/.changeset/tidy-trace-image-blocks.md
@@ -1,0 +1,5 @@
+---
+"@langchain/core": patch
+---
+
+Fix tracing of multimodal messages so that every base64/URL image content block is converted, not just the first one in a message.
--- a/libs/langchain-core/src/language_models/chat_models.ts
+++ b/libs/langchain-core/src/language_models/chat_models.ts
@@ -173,22 +173,21 @@ function _formatForTracing(messages: BaseMessage[]): BaseMessage[] {
   for (const message of messages) {
     let messageToTrace = message;
     if (Array.isArray(message.content)) {
-      for (let idx = 0; idx < message.content.length; idx++) {
-        const block = message.content[idx];
+      let converted = false;
+      const content = message.content.map((block) => {
         if (isURLContentBlock(block) || isBase64ContentBlock(block)) {
-          if (messageToTrace === message) {
-            // Also shallow-copy content
-            // oxlint-disable-next-line @typescript-eslint/no-explicit-any
-            messageToTrace = new (message.constructor as any)({
-              ...messageToTrace,
-              content: [
-                ...message.content.slice(0, idx),
-                convertToOpenAIImageBlock(block),
-                ...message.content.slice(idx + 1),
-              ],
-            });
-          }
+          converted = true;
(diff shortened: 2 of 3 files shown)


jackjin1997:
Thanks for working on this. This duplicates #11292, which was opened yesterday and already links #11291. Both patches use the same clone-once, map-all-matching-blocks fix; #11292 follows the exact two-PDF callback reproduction from the issue and records full core test, build, lint, and format validation. To avoid parallel review of functionally identical changes, could we consolidate on the earlier PR?

anshusaurav:
You're right, @jackjin1997 — thanks for the heads-up. This duplicates #11292, which was opened first and takes the same clone-once / map-all-blocks approach. Closing in favor of it. Nice work on the earlier fix! 👍

PR #11305, opened 2026-08-05
feat: support LangSmith Gateway environment routing

jacoblee93:
## Description
Adds JavaScript parity for configuring OpenAI, Anthropic, and Fireworks chat models with `LANGSMITH_GATEWAY` and `LANGSMITH_GATEWAY_API_KEY`. The shared resolver now handles only LangSmith-owned configuration; each provider resolves its own JavaScript/Python-compatible base URL and API key environment variables.

## Test Plan
- [x] Focused core, OpenAI, Anthropic, and Fireworks unit tests
- [x] Core and provider package builds
- [x] `pnpm lint` and focused formatting checks

Made by [Open SWE](https://openswe.vercel.app/agents/e9fc0108-1034-cda2-f558-aafd85fca2fe)

### DIFF
--- a/libs/langchain-core/src/utils/gateway.ts
+++ b/libs/langchain-core/src/utils/gateway.ts
@@ -1,5 +1,4 @@
 import { getEnvironmentVariable } from "./env.js";
-import { validateSafeUrl } from "./ssrf.js";
 
 const LANGSMITH_GATEWAY = "LANGSMITH_GATEWAY";
 const LANGSMITH_GATEWAY_API_KEY = "LANGSMITH_GATEWAY_API_KEY";
@@ -8,19 +7,16 @@ const DEFAULT_LANGSMITH_GATEWAY = "https://gateway.smith.langchain.com";
 const TRUE_VALUES = ["true", "1", "yes"];
 const FALSE_VALUES = ["false", "0", "no"];
 
-type EnvironmentVariableNames = string | readonly string[];
+export interface LangSmithGatewayConfigOptions<TApiKey> {
+  baseURL?: string;
+  apiKey?: TApiKey;
+  providerApiKey?: string;
+  providerPath: string;
+}
 
-function firstEnvironmentVariable(
-  names: EnvironmentVariableNames
-): string | undefined {
-  const environmentVariableNames = typeof names === "string" ? [names] : names;
-  for (const name of environmentVariableNames) {
-    const value = getEnvironmentVariable(name);
-    if (value) {
-      return value;
-    }
-  }
-  return undefined;
+export interface LangSmithGatewayConfig<TApiKey> {
+  baseURL?: string;
+  apiKey?: TApiKey | string;
 }
 
 function resolveLangSmithGatewayBaseURL(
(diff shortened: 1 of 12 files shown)


libs/providers/langchain-openai/src/chat_models/base.ts
```diff
@@ -524,10 +525,14 @@ export abstract class BaseChatOpenAI<
...
         : undefined;
-    this.apiKey =
-      fields?.apiKey ??
-      configApiKey ??
-      getEnvironmentVariable("OPENAI_API_KEY");
+    const gatewayConfig = resolveLangSmithGatewayConfig<OpenAIApiKey>({
+      baseURL: fields?.configuration?.baseURL,
+      apiKey: fields?.apiKey ?? configApiKey,
+      providerPath: "openai/v1",
+      baseURLEnv: "OPENAI_API_BASE",
```
hntrl: `|| "OPENAI_BASE_URL"`?

libs/providers/langchain-fireworks/src/chat_models.ts
```diff
@@ -91,10 +91,15 @@ export class ChatFireworks extends ChatOpenAICompletions<ChatFireworksCallOption
...
 
-    const fireworksApiKey =
-      fields.apiKey ||
-      fields.fireworksApiKey ||
-      getEnvironmentVariable("FIREWORKS_API_KEY");
+    const gatewayConfig = resolveLangSmithGatewayConfig<string>({
+      baseURL: fields.configuration?.baseURL,
+      apiKey: fields.apiKey || fields.fireworksApiKey,
+      providerPath: "fireworks",
+      baseURLEnv: "FIREWORKS_API_BASE",
```
hntrl: same thing - i don't know what standard convention for base urls are here. I don't think its this

libs/langchain-core/src/utils/gateway.ts
```diff
@@ -0,0 +1,97 @@
+import { getEnvironmentVariable } from "./env.js";
+import { validateSafeUrl } from "./ssrf.js";
+
+const LANGSMITH_GATEWAY = "LANGSMITH_GATEWAY";
+const LANGSMITH_GATEWAY_API_KEY = "LANGSMITH_GATEWAY_API_KEY";
+const LANGSMITH_API_KEY = "LANGSMITH_API_KEY";
+const DEFAULT_LANGSMITH_GATEWAY = "https://gateway.smith.langchain.com";
+const TRUE_VALUES = ["true", "1", "yes"];
+const FALSE_VALUES = ["false", "0", "no"];
```
hntrl, line 8: this doesn't seem exhaustive

libs/langchain-core/src/utils/gateway.ts
```diff
@@ -0,0 +1,97 @@
...
+      });
+  return `${baseURL}/${providerPath}`;
+}
+
+export function resolveLangSmithGatewayConfig<TApiKey = string>({
+  baseURL,
+  apiKey,
+  providerPath,
+  baseURLEnv = [],
+  apiKeyEnv = [],
```
hntrl: why do this instead of having provider pass in `getEnvironmentVariable`

libs/langchain-core/src/utils/gateway.ts
```diff
@@ -0,0 +1,97 @@
...
+  apiKey?: TApiKey;
+  providerPath: string;
+  baseURLEnv?: EnvironmentVariableNames;
+  apiKeyEnv?: EnvironmentVariableNames;
+  defaultBaseURL?: string;
+}): {
+  baseURL?: string;
+  apiKey?: TApiKey | string;
+  baseURLFromGateway: boolean;
+} {
```
hntrl: nit: can these be named types instead of opaque type assigns

libs/langchain-core/src/utils/gateway.ts
jacoblee93: People should only ever be setting `true` or `false` IMO but Python has an obsession with "user friendliness"

Let's just keep it so people don't wonder why something works in one language but not the other

libs/langchain-core/src/utils/gateway.ts
jacoblee93: Distinguish between an explicitly provided key from one passed into the env since gateway env should override provider env:

```
- Explicit `apiKey` always wins, hence the early return on line 47.
- A provider env key should lose to `LANGSMITH_GATEWAY_API_KEY` when the gateway supplies the base URL.
- Otherwise, the provider env key should win.
```

libs/langchain-core/src/utils/gateway.ts
```diff
@@ -0,0 +1,63 @@
...
+    : undefined;
+  if (!gatewayApiKey && baseURLFromGateway) {
+    gatewayApiKey = getEnvironmentVariable(LANGSMITH_API_KEY);
+  }
+
+  return {
+    baseURL: resolvedBaseURL,
+    apiKey: baseURLFromGateway
+      ? gatewayApiKey || providerApiKey
+      : providerApiKey || gatewayApiKey,
```
hntrl: why would we fail over to gateway api key if we're not pointing at a gateway base url

libs/langchain-core/src/utils/gateway.ts
```diff
@@ -0,0 +1,63 @@
...
+  const baseURL = TRUE_VALUES.includes(value.toLowerCase())
+    ? DEFAULT_LANGSMITH_GATEWAY
+    : value.replace(/\/+$/, "");
+  return `${baseURL}/${providerPath}`;
+}
+
+export function resolveLangSmithGatewayConfig<TApiKey = string>({
+  baseURL,
+  apiKey,
+  providerApiKey,
```
hntrl: i still don't get why apiKey + providerApiKey is distinct

hntrl (approved):

PR #11312, opened 2026-08-06
ci: harden the npm release workflows and add a Plumber score

thomasboni:
Two hardening changes on the npm release path. A third commit adds the check that found them. It's independent and can be dropped.

**1. No dependency cache on the publish jobs** (`publish.yml`, `dev-release.yml`)

Six test and compatibility workflows run on pushes to `main` with the same pnpm store cache key family, and `pnpm install` executes lifecycle scripts from the dependency tree during those runs. The release jobs restore that main-scoped cache before installing and publishing to npm, so anything those runs write into the cache can end up in published packages, and the `restore-keys` prefix fallback keeps a stale store eligible even after the lockfile changes. Dropping `cache: "pnpm"` on the two publish jobs keeps the release install derived from the lockfile and registry alone. Test workflows keep their cache, so the cost is a longer install on release runs only.

**2. Only forward `NPM_TOKEN` to the called release workflows** (`release.yml`)

`secrets: inherit` hands every repository secret to the called workflows, while `publish.yml` and `dev-release.yml` only read `NPM_TOKEN` (`GITHUB_TOKEN` is provided by the runner either way). The callees now declare the one secret they need and the entry point passes it explicitly, so future secrets added to the repo stay out of the release jobs' reach by default.

**3. Plumber workflow and badge**

The two issues above were found with the Plumber CLI, verified on `main`. The last commit adds it as a workflow so the same checks run on future changes:

- It scans the workflows for CI/CD security issues (unpinned actions, over-broad token permissions, caches on release paths) and uploads results as SARIF to code scanning.
- It runs on built-in defaults, no config needed. The small `.plumber.yaml` here only adds `amannn/action-semantic-pull-request` (already pinned by SHA) to the trusted actions list.
- The badge works like OpenSSF Scorecard's published results: it shows the default branch score, a failed publish never fails CI, and it reads UNKNOWN until the first run on `main`. The gate is 85/100 so a single small finding does not block PRs.

I work on Plumber. If you do not want the tool, no hard feelings, say so and I remove that commit, the hardening stands on its own.

### DIFF
--- a/.github/workflows/dev-release.yml
+++ b/.github/workflows/dev-release.yml
@@ -9,6 +9,10 @@ on:
         required: false
         default: "dev"
         type: string
+    secrets:
+      NPM_TOKEN:
+        description: "npm token used to authenticate the publish"
+        required: true
 
 permissions:
   contents: read
--- a/.github/workflows/dev-release.yml
+++ b/.github/workflows/dev-release.yml
@@ -38,7 +38,6 @@ jobs:
         uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
         with:
           node-version-file: ".nvmrc"
-          cache: "pnpm"
 
       - name: 📥 Install deps
         run: pnpm install --frozen-lockfile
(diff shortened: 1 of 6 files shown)


hntrl (commented):
Hi Thomas,

We currently do not have a need for additional CI scanning

thomasboni:
@hntrl ok, that's clear. 

Do you still want me to open a PR to add the CI hardenings in the existing workflows? Points 1 and 2 above, without adding anything in the workflow?

PR #11318, opened 2026-08-07
feat: add isRetryable to LangChainError, respect it in retry mw defaults

thushanth-bengre-langchain:
Adds `isRetryable` to LangChainError (defaults to true) so `modelRetryMiddleware` and `toolRetryMiddleware` can skip retrying errors that can never succeed, instead of burning `maxRetries` attempts on them by default. Only changes the default `retryOn` — an explicit `retryOn` is unaffected.

Classified as non-retryable: `ModelAbortError`, `ContextOverflowError`, and `@langchain/google`'s `ConfigurationError`, `AuthError`, `PromptBlockedError`, `InvalidToolError`, `ToolCallNotFoundError`, `InvalidInputError`. `NoCandidatesError` and `MalformedOutputError` are left on the base default (`true`) with a `TODO` — genuinely ambiguous, deferred rather than guessed.

**Breaking** (`@langchain/google`, **minor bump**): `RequestError.isRetryable` is now a property, not a method — `error.isRetryable()` will throw. No in-repo callers existed outside this class's own doc comments.

Also fixes `modelRetryMiddleware`'s array-form `retryOn` matching via `error.constructor === ErrorConstructor` instead of `instanceof` — a subclass of a listed error type was never matched; now consistent with `toolRetryMiddleware`


libs/langchain-core/src/errors/index.ts
```diff
@@ -48,6 +48,16 @@ export const ns = baseNs.sub("error");
...
 
+  /**
+   * Whether this error is safe to retry (e.g. a transient network/rate-limit
+   * failure) as opposed to deterministic (e.g. bad credentials, malformed
+   * input) — retrying a deterministic failure unchanged will fail identically.
+   *
+   * Defaults to `true` to preserve existing retry-everything behavior for any
+   * subclass that doesn't explicitly override it.
+   */
+  readonly isRetryable: boolean = true;
```
hntrl: this should just be a condition on APIError

libs/langchain/src/agents/middleware/modelRetry.ts
```diff
@@ -147,9 +147,10 @@ export function modelRetryMiddleware(config: ModelRetryMiddlewareConfig = {}) {
       return retryOn(error);
     }
     // retryOn is an array of error constructors
-    return retryOn.some(
-      (ErrorConstructor) => error.constructor === ErrorConstructor
-    );
+    return retryOn.some((ErrorConstructor) => {
+      // oxlint-disable-next-line no-instanceof/no-instanceof
+      return error instanceof ErrorConstructor;
```
hntrl: I think we may want to condition on `.isInstance`

libs/langchain/src/agents/middleware/modelRetry.ts
```diff
@@ -147,9 +147,10 @@ export function modelRetryMiddleware(config: ModelRetryMiddlewareConfig = {}) {
       return retryOn(error);
     }
     // retryOn is an array of error constructors
-    return retryOn.some(
-      (ErrorConstructor) => error.constructor === ErrorConstructor
```
hntrl, line 151: change toolRetry to use this so we don't hatch lint

libs/langchain-core/src/errors/index.ts
hntrl: nvm

PR #11322, opened 2026-08-08
fix(core): preserve web search results in content blocks

### LINKED ISSUE #11320: web_search_call.results is missing from contentBlocks when streaming
### Checked other resources

### Example Code

The following code: 
```typescript
  { messages },
  { streamMode: "messages" }
);

for await (const [message] of stream) {
  console.log(message.contentBlocks);
}
```

When the Web Search tool returns Image Search Results, the raw `⁠web_search_call`⁠ contains `⁠results⁠`, but the corresponding ⁠`server_tool_call_result`⁠ in ⁠`message.contentBlocks⁠` does not.

**Expected:** ⁠ `web_search_call.results`⁠ is available in the corresponding `⁠server_tool_call_result`⁠, including Image Search Results.

**Actual:** The ⁠`server_tool_call_result⁠` contains `⁠action`⁠, but `⁠results`⁠ is missing.

N/A - no error is thrown. The `results` field is missing from the converted `contentBlocks`

### Description

When using OpenAI Web Search with `⁠agent.stream(..., { streamMode: "messages" })⁠`, the `⁠web_search_call`⁠ is available in ⁠message.contentBlocks⁠, but its `⁠results`⁠ are missing.

This prevents us from consuming OpenAI Image Search Results while streaming, since the image URLs are returned in ⁠`web_search_call.results⁠`.

The raw tool output contains ⁠`results`⁠, but they appear to be dropped when converting the OpenAI message to LangChain ⁠`contentBlocks⁠`.

I traced this to:

`⁠libs/langchain-core/src/messages/block_translators/openai.ts⁠`

The current conversion preserves ⁠`action`⁠:

```typescript
if (_isObject(toolOutput.action)) {
  output.action = toolOutput.action;
}
```

but not `⁠results⁠`.
Would it make sense to preserve ⁠`results⁠` as well?

```typescript
if (_isArray(toolOutput.results)) {
  output.results = toolOutput.results;
}
```

We need to consume Image Search Results through the streaming message interface in a production use case.

OpenAI Image Search Results:
https://developers.openai.com/api/docs/guides/tools-web-search#image-search-results

I'm happy to submit a PR with this change and a unit test if this approach looks good

### System Info

@langchain/core: 1.2.4
@langchain/openai: 1.5.5
platform: macOS
Node: 24.12.0
pnpm 11.15.1

fuji-151a:
Fixes #11320 

Preserves `web_search_call.resutls` when converting OpenAI Web Search output to LangChain `contentBlocks`.

Previously, the conversion only preserved `action`, causing `results` to be missing from `message.contentBlocks` when using `streamMode: "message"`. This prevents consumers from accessing OpenAI Image Search Results, including the returned image URLs, through the streaming message interface.

OpenAI Image Search Results: https://developers.openai.com/api/docs/guides/tools-web-search#image-search-results

This change preserves `results` as well and adds a unit test covering the behavior.


fuji-151a:
Hi! Just checking in on this PR.

Is there anything else needed from my side to move this forward? Thanks!

PR #11330, opened 2026-08-10
fix(langchain): preserve model errors propagated through middleware

### LINKED ISSUE #11324: MiddlewareError.wrap discards `code`, custom properties, and the prototype, silently breaking both documented `retryOn` forms of modelRetryMiddleware
### Checked other resources

- [x] The bug is not resolved by updating to the latest stable version (reproduced on `langchain@1.5.5` / `@langchain/core@1.2.5`).

### Example Code

Self-contained. `npm i langchain@1.5.5 @langchain/core@1.2.5 && node repro.mjs`

```typescript
import { createAgent, createMiddleware, modelRetryMiddleware, MiddlewareError } from "langchain";
import { SimpleChatModel } from "@langchain/core/language_models/chat_models";

class MyRetryableError extends Error {
  $retryable = true;
  constructor() { super("boom"); this.name = "MyRetryableError"; this.code = "ECONNRESET"; }
}

// ── 1. MiddlewareError.wrap drops `code`, custom props, and the prototype ──
const raw = new MyRetryableError();
const wrapped = MiddlewareError.wrap(raw, "SomeMiddleware");
console.log("raw     : code=%s $retryable=%s instanceof=%s", raw.code, raw.$retryable, raw instanceof MyRetryableError);
console.log("wrapped : code=%s $retryable=%s instanceof=%s", wrapped.code, wrapped.$retryable, wrapped instanceof MyRetryableError);

// ── 2. Both documented retryOn forms go dead with a middleware inside the retry ──
let calls = 0;
class FailingModel extends SimpleChatModel {
  _llmType() { return "failing"; }
  bindTools() { return this; }
  async _call() { calls += 1; throw new MyRetryableError(); }
}

const passthrough = createMiddleware({
  name: "Passthrough",
  wrapModelCall: async (request, handler) => handler(request),
});

async function modelCalls(retryOn, extraMiddleware) {
  calls = 0;
  const agent = createAgent({
    model: new FailingModel({}),
    tools: [],
    middleware: [
      modelRetryMiddleware({ maxRetries: 2, retryOn, initialDelayMs: 0, onFailure: "error" }),
      ...extraMiddleware,
    ],
  });
  try { await agent.invoke({ messages: [new HumanMessage("hi")] }); } catch { /* expected */ }
  return calls;
}

// Form A — the JSDoc's "Retry specific exceptions only" example
console.log("array form, no middleware  : %d calls (expect 3)", await modelCalls([MyRetryableError], []));
console.log("array form, one middleware : %d calls (expect 3)", await modelCalls([MyRetryableError], [passthrough]));

// What the predicate actually receives
await modelCalls((e) => {
  console.log("predicate sees: constructor=%s name=%s code=%s $retryable=%s instanceof=%s",
    e.constructor.name, e.name, e.code, e.$retryable, e instanceof MyRetryableError);
  return false;
}, [passthrough]);
```

Output:

```
raw     : code=ECONNRESET $retryable=true instanceof=true
wrapped : code=undefined  $retryable=undefined instanceof=false

array form, no middleware  : 3 calls (expect 3)
array form, one middleware : 1 calls (expect 3)

predicate sees: constructor=MiddlewareError name=MyRetryableError code=undefined $retryable=undefined instanceof=false
```

The JSDoc's other example — `"Custom exception filtering"`, which tests
`error.name === "HTTPError" && "statusCode" in error` — fails the same way
(3 calls → 1 call) because `statusCode` is dropped.

Wrapping also nests once per middleware, so the depth is not fixed:

```
middlewares=0  chainDepth=1  Tagged
middlewares=1  chainDepth=2  MiddlewareError -> Tagged
middlewares=3  chainDepth=4  MiddlewareError -> MiddlewareError -> MiddlewareError -> Tagged
```

No error is raised. That is the problem — retries silently do not happen.

### Description

`AgentNode` wraps every middleware's throw in `MiddlewareError`
(`agents/nodes/AgentNode.js`, `throw MiddlewareError.wrap(error, currentMiddleware.name)`).
The constructor copies only `message` and `name` and moves the original to
`cause`:

```js
constructor(error, middlewareName) {
  const errorMessage = error instanceof Error ? error.message : String(error);
  super(errorMessage);
  this.name = error instanceof Error ? error.name : `...Error`;
  if (error instanceof Error) this.cause = error;
}
```

So custom properties and the prototype are gone by the time an outer middleware
inspects the error.

I expect `modelRetryMiddleware`'s documented `retryOn` forms to work in a normal
agent. Instead **both examples in its own JSDoc stop working as soon as any
middleware sits inside the retry**, which is the common case:

```ts
// @example Retry specific exceptions only
retryOn: [TimeoutError, NetworkError],          // matched via error.constructor === Ctor

// @example Custom exception filtering
if (error.name === "HTTPError" && "statusCode" in error) { ... }
```

The array form is matched with `error.constructor === ErrorConstructor`
(`modelRetry.js`), which cannot hold after re-wrapping. The `statusCode` example
fails because the property is stripped.

Because nesting depth varies with middleware count, a predicate cannot know how
far to walk `cause`; `name` is the only property that survives arbitrarily many
wraps. But `name` alone cannot express "HTTP 503" or "ECONNRESET".

**Why this matters more than the workaround suggests.** The failure is silent and
indistinguishable from success — a retry that never fires looks exactly like a
fault that never occurs. We ran `$retryable` and `code` checks in production for
months believing they worked; they had never matched once. We only found out by
asserting attempt counts through the real middleware stack.

**Suggested directions**

1. Preserve the original error's own properties and prototype when wrapping
   (e.g. `Object.setPrototypeOf` + copy own keys), or
2. Unwrap `MiddlewareError` before evaluating `retryOn`, so both the predicate and
   the class-array form see the error as thrown, or
3. If lossy wrapping must stay, make the array form unwrap-aware and fix the
   JSDoc examples, since as written they do not work.

(1) or (2) look preferable — they fix every user-written predicate, not just the
built-in one. Happy to open a PR if you'd like.

### System Info

```
langchain@1.5.4 and @langchain/core@1.2.3 (also reproduced on langchain@1.5.5 / @langchain/core@1.2.5)
@langchain/aws@1.4.3
@langchain/langgraph@1.4.8

platform: mac (Darwin arm64)
Node: v24.14.0
pnpm: 10.14.0
```

HDPark95:
## Description

A `wrapModelCall` middleware that only forwards to `handler(request)` currently turns downstream model errors into `MiddlewareError`. That changes the error identity, prototype, and custom fields before an outer middleware sees it, silently breaking both documented `modelRetryMiddleware.retryOn` forms:

- the class-array form compares `error.constructor` with the configured constructor;
- predicates commonly read custom fields such as `code` or `statusCode`.

The initial revision unwrapped errors inside `modelRetryMiddleware`. After review, I confirmed that the tool-call path already avoids wrapping errors merely propagated by middleware. This revision moves the fix to the model-call wrap site instead, so the two paths follow the same rule.

## Fix

`AgentNode` now tracks the exact values thrown by the inner model handler. If the current middleware propagates the same value, it is re-thrown unchanged. Errors originating in the middleware are still wrapped in `MiddlewareError`, preserving middleware attribution.

A patch changeset for `langchain` is included.

## Tests

Added three regression tests with a pass-through middleware inside the retry:

- the class-array form performs the initial call plus two retries;
- the predicate receives the original error with `statusCode` intact;
- exhausted retries format the original error type, and `onFailure: "error"` re-throws the original instance.

All three fail on `main` without the `AgentNode` change. Validation on this branch:

- `pnpm build --filter langchain`
- `pnpm run test:unit:ci --filter langchain` — 59 files, 851 tests, no type errors
- `pnpm run lint`
- `pnpm run format:check`

Fixes #11324

### DIFF
--- a/.changeset/soft-dogs-retry.md
+++ b/.changeset/soft-dogs-retry.md
@@ -1,0 +1,5 @@
+---
+"langchain": patch
+---
+
+fix(langchain): preserve model errors propagated through middleware
--- a/libs/langchain/src/agents/middleware/modelRetry.ts
+++ b/libs/langchain/src/agents/middleware/modelRetry.ts
@@ -5,7 +5,6 @@ import { z } from "zod/v3";
 import { AIMessage } from "@langchain/core/messages";
 
 import { createMiddleware } from "../middleware.js";
-import { MiddlewareError } from "../errors.js";
 import { sleep, calculateRetryDelay } from "./utils.js";
 import { RetrySchema } from "./constants.js";
 import { InvalidRetryConfigError } from "./error.js";
@@ -141,29 +140,6 @@ export function modelRetryMiddleware(config: ModelRetryMiddlewareConfig = {}) {
   } = data;
 
   /**
-   * Recover the error as it was originally thrown.
-   *
-   * `AgentNode` re-wraps every middleware throw in a `MiddlewareError`, which
-   * keeps only `message`/`name` and moves the original error to `cause`. So when
-   * any middleware sits inside the retry, `retryOn` would otherwise see the
-   * wrapper - losing `constructor`, `code`, custom properties, and the prototype -
-   * and never match, silently disabling retries. Walk the `cause` chain so both
-   * the class-array (`error.constructor === Ctor`) and predicate forms see the
-   * error as thrown, regardless of how many middleware wrapped it.
-   */
-  const unwrapMiddlewareError = (error: Error): Error => {
-    let current = error;
-    while (
-      MiddlewareError.isInstance(current) &&
-      // oxlint-disable-next-line no-instanceof/no-instanceof
-      current.cause instanceof Error
-    ) {
-      current = current.cause;
-    }
-    return current;
-  };
-
-  /**
    * Check if the exception should trigger a retry.
    */
   const shouldRetryException = (error: Error): boolean => {
@@ -220,14 +196,10 @@ export function modelRetryMiddleware(config: ModelRetryMiddlewareConfig = {}) {
           const attemptsMade = attempt + 1; // attempt is 0-indexed
 
           // Ensure error is an Error instance
-          const rawError =
+          const err =
             error && typeof error === "object" && "message" in error
               ? (error as Error)
               : new Error(String(error));
-
-          // Unwrap any MiddlewareError layers added by AgentNode so retryOn and
-          // the failure handler see the error as it was originally thrown.
-          const err = unwrapMiddlewareError(rawError);
 
           // Check if we should retry this exception
           if (!shouldRetryException(err)) {
--- a/libs/langchain/src/agents/middleware/modelRetry.ts
+++ b/libs/langchain/src/agents/middleware/modelRetry.ts
@@ -5,6 +5,7 @@ import { z } from "zod/v3";
 import { AIMessage } from "@langchain/core/messages";
 
 import { createMiddleware } from "../middleware.js";
+import { MiddlewareError } from "../errors.js";
 import { sleep, calculateRetryDelay } from "./utils.js";
 import { RetrySchema } from "./constants.js";
 import { InvalidRetryConfigError } from "./error.js";
@@ -140,6 +141,29 @@ export function modelRetryMiddleware(config: ModelRetryMiddlewareConfig = {}) {
   } = data;
 
   /**
+   * Recover the error as it was originally thrown.
+   *
+   * `AgentNode` re-wraps every middleware throw in a `MiddlewareError`, which
+   * keeps only `message`/`name` and moves the original error to `cause`. So when
+   * any middleware sits inside the retry, `retryOn` would otherwise see the
+   * wrapper - losing `constructor`, `code`, custom properties, and the prototype -
+   * and never match, silently disabling retries. Walk the `cause` chain so both
+   * the class-array (`error.constructor === Ctor`) and predicate forms see the
+   * error as thrown, regardless of how many middleware wrapped it.
+   */
+  const unwrapMiddlewareError = (error: Error): Error => {
+    let current = error;
+    while (
+      MiddlewareError.isInstance(current) &&
+      // oxlint-disable-next-line no-instanceof/no-instanceof
+      current.cause instanceof Error
+    ) {
+      current = current.cause;
+    }
+    return current;
+  };
+
+  /**
    * Check if the exception should trigger a retry.
    */
   const shouldRetryException = (error: Error): boolean => {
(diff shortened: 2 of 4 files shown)


kdelay:
Read the diff and traced the wrap path on `main`. The unwrap does fix the reported case, but the asymmetry behind it seems worth surfacing before this lands, because it decides where the fix belongs.

**The tool-call path already refuses to wrap errors it merely propagated.**

`libs/langchain/src/agents/utils.ts` composes `wrapToolCall` with the same re-wrap, but it first records the exact values thrown by the downstream handler:

```ts
// Track exact values thrown by the downstream handler so unchanged
// propagation is not misclassified as a failure in this middleware.
const downstreamErrors = new Set<unknown>();
// ...
} catch (error: unknown) {
  if (downstreamErrors.has(error)) {
    throw error;
  }
  throw MiddlewareError.wrap(error, m.name);
}
```

`AgentNode.ts` has no equivalent: line 701 is its only wrap site and it wraps unconditionally, so a middleware whose `wrapModelCall` does nothing but `handler(request)` still produces a `MiddlewareError` for an error it never raised. That is exactly the `passthrough` case in your new tests.

That leaves two places to fix this:

- at the consumer (this PR): `modelRetryMiddleware` recovers the error as thrown. Every other consumer still sees wrappers, and a genuine middleware failure gets unwrapped too, which drops the "this middleware failed" attribution `MiddlewareError` exists to carry.
- at the wrap site: give `AgentNode` the same `downstreamErrors` guard as `utils.ts`. Errors passing through a middleware unchanged then reach `retryOn` as thrown, errors originating in a middleware keep their wrapper, and the class-array form starts working for user-written middleware too, not only inside `modelRetryMiddleware`.

The second is a narrower behavioural change than it first looks, since it only stops attributing pass-through errors, and it makes the two middleware paths agree. If the consumer side was a deliberate choice over that, a line in the PR body saying why would help review.

**If the unwrap stays, the failure path changes behaviour and nothing pins it.**

`handleFailure` now receives the unwrapped error, so with a middleware inside the retry `formatFailureMessage`'s `error.constructor.name` goes from `MiddlewareError` to `TimeoutError`, and `onFailure: "error"` re-throws the original error rather than the inner wrapper (`AgentNode` re-wraps one level up, but the `cause` chain now points straight at the original). Both read as improvements, and neither is asserted: both new tests use `onFailure: "continue"` and check call counts and the predicate argument. The class-array test already exhausts its retries, so asserting the returned `AIMessage` content there would lock the message, and one `onFailure: "error"` case would lock the re-raise.

**Changeset.** changeset-bot flagged this one. The closest merged analogue, #11277 (`fix(langchain)`, same package), carried a `.changeset/` entry, so without one this fix would land without shipping in a release.

HDPark95:
Done in 53e6175. I confirmed the model-call/tool-call asymmetry and moved the fix to `AgentNode`: errors thrown by the inner handler now keep their identity when a middleware only propagates them, while errors originating in middleware still go through `MiddlewareError.wrap`. I removed the consumer-side unwrap, added the `langchain` patch changeset, and pinned both the exhausted-message and `onFailure: "error"` paths. The three new tests fail on `main`; the revised branch passes all 851 `langchain` unit tests.

PR #11267, opened 2026-07-29
fix(langchain): prevent summary output stream leaks

jackjin1997:
## Summary

Fixes #11238

- prevent the internal summarization model invocation from inheriting user-facing callbacks
- keep parent tags and metadata while isolating the housekeeping model call from message streams
- reuse the summarized state message ID so the synthetic summary remains in agent state without being emitted as a new message

## Test plan

- `pnpm --filter langchain test` — 59 files, 845 tests passed; typecheck clean
- `pnpm --filter langchain build` — passed, including publint and attw
- `pnpm exec oxfmt --check <changed files>` — passed
- `pnpm exec oxlint <changed TypeScript files>` — passed

Reverse verification covers both leak paths: removing callback isolation emits the internal summarizer model output, while assigning a fresh ID to the synthetic summary emits the `lc_source: "summarization"` `HumanMessage`.

## AI Disclosure

AI assisted with issue triage, test repair, implementation, and validation. The diff and reverse verification were reviewed before submission.

PR #11331, opened 2026-08-10
fix(langchain): exclude middleware-internal model calls from the message projection

### LINKED ISSUE #11238: JS Summarization Middleware: internal model.invoke() is streamed to UI - regression
### Checked other resources

### Example Code

```typescript

import {
  summarizationMiddleware,
  createAgent
} from 'langchain'
import { Overwrite } from '@langchain/langgraph'

const coreAgent = createAgent({
  model: 'gpt-5.4-mini',
  systemPrompt,
  middleware: [
    summarizationMiddleware({
      model: 'gpt-5.4-mini',
      trigger: { messages: 3 },
      keep: { messages: 2 }
    })
  ],
  name: 'CoreAgent',
  version: 'v1'
}).withConfig({
  runName: 'CoreAgent',
  tags: ['user-visible']
})

const result = await coreAgent.invoke(
  { messages: state.messages },
  { ...config, recursionLimit: 100 }
)

return {
  messages: new Overwrite(result.messages ?? [])
}
```

(part of a LangGraph state graph)

_No response_

### Description

I am using the Summarization middleware in a JS LangChain/LangGraph app. 

## Expected behavior
Summarization runs silently in the background, compacting the context. It is not output from the `invoke()` call and is not visible in the resulting streamed messages.

## Actual behavior
Summarization messages are output from the `invoke()` and streamed to the chat.
Example:

> USER: tell me a story
> SYSTEM: Here is a summary of the conversation to date:
Error generating summary: Error: Missing credentials. Please pass an apiKey, workloadIdentity, adminAPIKey, or set the OPENAI_API_KEY or OPENAI_ADMIN_KEY environment variable.I'm only able to assist with topics related to Bubble.io and your app. If you have any questions about building your app or need help with Bubble features, I'm here to help!

(note: this error message is because I have not configured the ChatGPT model used for summarization)

## Details
I think this is a regression of #9455
The fix for that issue (in #9640) was to add `callbacks: []` to the `invoke()` options for the summarization model. That fix was removed in #9812. 
A similar issue (#10042) with LLM Tool Selector middleware was fixed in the same way in #10160

## Workarounds
I've found two workarounds:
1. ~override the middleware's `invoke()` in a hacky way to set the config callbacks to `[]`~
    - I was incorrect: this does not appear to affect the output
2. filter out messages from the stream if their metadata contains `lc_source==='summarization'` or `langgraph_node==='SummarizationMiddleware.before_model'`Neither of these is ideal
    - Checking `additional_kwargs?.lc_source === 'summarization'` also works to detect the HumanMessage

After further investigation it seems workaround 1 only works if we also check `additional_kwargs.lc_source` to filter the summarization messages. It seems that we are getting messages in the stream both from the Summarization LLM call, and also from the HumanMessage returned from the middleware.

## pnpm info langchain
```
langchain@1.5.3 | MIT | deps: 4 | versions: 456

dist
.tarball: https://registry.npmjs.org/langchain/-/langchain-1.5.3.tgz
.shasum: 1e53dd26388696ccdcf7e7f3fe588ac010c47839
.integrity: sha512-YRYPy1xPq4CgKe0NdPoscYHnKBW5cViEyWiZsUKQfN3xL1X5DpGlIpaBQg1A/t6o9ePKE7J5b7nqrfLz8t7/Rw==
.unpackedSize: 3.2 MB

dependencies:
@langchain/langgraph-checkpoint: ^1.1.3
@langchain/langgraph: ^1.4.7
langsmith: >=0.5.0 <1.0.0
zod: ^3.25.76 || ^4

maintainers:
- hntrl <hunter@hntrl.io>
- lc-oss-admin <oss-admin@langchain.dev>

dist-tags:
latest: 1.5.3
dev: 1.5.1-dev-1781824996270
next: 1.0.0-alpha.9
alpha: 1.0.0-alpha.9
tag-for-publishing-older-releases: 0.2.20

published a week ago by GitHub Actions <npm-oidc-no-reply@github.com>
```

## platform
mac

## Node version
v24.13.0

## pnpm --version
10.33.4

thushanth-bengre-langchain:
`summarizationMiddleware`, `toolEmulatorMiddleware`, and `llmToolSelectorMiddleware` make bookkeeping model calls that surfaced as extra assistant messages in `run.messages` and `stream({ streamMode: "messages" })`. Tag them with LangGraph's `nostream` tag so its messages handler skips them. `toolEmulator` passed no config at all, so it also gains the ambient-config merge.

Ports langchain-ai/langchain#39252 with a different mechanism: that PR filters via a `StreamTransformer`, which @langchain/langgraph 1.4.8 cannot support — it drops the event metadata a transformer would read, and has no `before_builtins` ordering. Suppressing at emission matches #10160, which fixed this for `llmToolSelector`.

Behavior change: internal calls no longer appear in `run.messages` or
classic `streamMode: "messages"`. They remain observable via
`streamEvents({ version: "v2" })`, identified by `lc_source`.

Co-authored-by: jackjin1997 <28055621+jackjin1997@users.noreply.github.com>


jackjin1997:
One boundary looks intentionally different from the linked user report in #11238: the reporter confirmed that both the internal summarizer invocation and the synthetic `HumanMessage` should be absent from the message stream. This PR suppresses the invocation, but `internalCall.stream.test.ts` currently asserts that `"Here is a summary ..."` remains in `run.messages`.

`INTERNAL_CALL_TAG` addresses the model-call half only; the middleware state write can still project the synthetic summary as a newly added message. The focused regression in #11267 covers the second path by reusing the ID of the first replaced conversation message, so the summary stays in state without being emitted as a new message. Could this broader fix include equivalent handling (and invert that assertion), or should #11267 remain as the complementary fix for that path?

libs/langchain/src/agents/middleware/internalCall.ts
hntrl, line 1: nit: i'd rather not have this in a separate file, I think fine to inline inside of the middleware configs directly

libs/langchain/src/agents/tests/internalCall.observability.test.ts
hntrl, line 1: `blastRadius`?

libs/langchain/src/agents/tests/internalCall.blastRadius.test.ts
```diff
@@ -0,0 +1,96 @@
...
+
+import { createAgent, createMiddleware } from "../index.js";
+import { INTERNAL_CALL_TAG } from "../middleware/internalCall.js";
+
+// Pins which stream surfaces `INTERNAL_CALL_TAG` affects.
+
+function buildAgent(tagged: boolean) {
+  const internalModel = fakeModel().respond(
+    new AIMessage("INTERNAL_CALL")
+  ) as unknown as BaseChatModel;
```
hntrl: hmmmmm whats this type assign for?

libs/langchain/src/agents/tests/internalCall.stream.test.ts
hntrl, line 1: could we validate .stream also?

libs/langchain/src/agents/tests/internalCall.observability.test.ts
thushanth-bengre-langchain: Used that name because this test was checking which stream surface the tag reaches. 
Renamed this to `observability` which makes more sense.

libs/langchain/src/agents/tests/internalCall.blastRadius.test.ts
thushanth-bengre-langchain: Good catch — unnecessary. fakeModel() returns FakeBuiltModel, which extends BaseChatModel, so no assertion is needed.

Removed it

libs/langchain/src/agents/tests/internalCall.stream.test.ts
thushanth-bengre-langchain: Added `stream({ streamMode: "messages" })` coverage for both middleware.

thushanth-bengre-langchain:
> One boundary looks intentionally different from the linked user report in #11238: the reporter confirmed that both the internal summarizer invocation and the synthetic `HumanMessage` should be absent from the message stream. This PR suppresses the invocation, but `internalCall.stream.test.ts` currently asserts that `"Here is a summary ..."` remains in `run.messages`.
> 
> `INTERNAL_CALL_TAG` addresses the model-call half only; the middleware state write can still project the synthetic summary as a newly added message. The focused regression in #11267 covers the second path by reusing the ID of the first replaced conversation message, so the summary stays in state without being emitted as a new message. Could this broader fix include equivalent handling (and invert that assertion), or should #11267 remain as the complementary fix for that path?

Either order works; the only conflict is one hunk in `createSummary` (this PR adds `tags`, [#11267](https://github.com/langchain-ai/langchainjs/pull/11267) adds `callbacks: []` and rewrites the comment above). The ID-reuse change and the test file don't overlap with anything here.

If this lands first: [#11267](https://github.com/langchain-ai/langchainjs/pull/11267) rebases and I'd suggest dropping `callbacks: []` — with the tag in place it's redundant for suppression, and it additionally removes the summarization call from `streamEvents({ version: "v2" })` (measured: `[{ content: "THE_SUMMARY", lc_source: "summarization" }, { content: "MAIN" }] → [{ content: "MAIN" }]`). That's the escape hatch this PR's changeset points people to. [#11267](https://github.com/langchain-ai/langchainjs/pull/11267) keeps the ID-reuse fix, which the tag structurally can't do.

If [#11267](https://github.com/langchain-ai/langchainjs/pull/11267) lands first: I rebase, drop `callbacks: []` in favour of the tag for the same reason, and amend the changeset if it stays.

Either way the end state is the same: tag for the model-call half across all three middleware, ID reuse for the state-write half.

hntrl:
re: above

Remediation mentioned here: https://github.com/langchain-ai/langchainjs/pull/11267#pullrequestreview-4902771879

We're just going to move ahead with the tags assignment approach

.changeset/plenty-rules-smile.md
hntrl, line 1: this changelog is very verbose, also lets make it patch since this is technically a fix

libs/langchain/src/agents/tests/internalCall.observability.test.ts
hntrl, line 1: can we roll this into one test suite in the existing middleware files? `internalCall` 4x feels a little opaque

hntrl (commented):
also is there any way we can have accompanying tests per middleware? just dont want to regress later down the line

hntrl (approved):

PR #11338, opened 2026-08-11
fix(core): _formatForTracing converts all base64/URL content blocks, not just the first

### LINKED ISSUE #11291: _formatForTracing only converts the first base64 content block per message

okxint:
Fixes #11291.

## Problem

`_formatForTracing` in `chat_models.ts` used a `if (messageToTrace === message)` one-shot guard inside its inner loop. Once the **first** base64 (or URL) content block was found and the message was shallow-copied into `messageToTrace`, the guard evaluated to `false` for every subsequent block in the same message. Those blocks were left as raw base64 data in the traced payload and never converted.

```ts
// before — only the first base64 block is converted
for (let idx = 0; idx < message.content.length; idx++) {
  const block = message.content[idx];
  if (isURLContentBlock(block) || isBase64ContentBlock(block)) {
    if (messageToTrace === message) {   // ← false after the first hit
      messageToTrace = new (...) { ... }
    }
    // second, third, … blocks fall through unconverted
  }
}
```

## Fix

Replace the loop-with-guard with a single `.map()` pass that converts **every** base64/URL block in one sweep, and only allocates a new message instance when at least one block changed:

```ts
// after — all base64/URL blocks are converted
let changed = false;
const convertedContent = message.content.map((block) => {
  if (isURLContentBlock(block) || isBase64ContentBlock(block)) {
    changed = true;
    return convertToOpenAIImageBlock(block);
  }
  return block;
});
if (changed) {
  messagesToTrace.push(new (message.constructor as any)({ ...message, content: convertedContent }));
} else {
  messagesToTrace.push(message);
}
```

## Test

Added a regression test in `libs/langchain-core/src/language_models/tests/chat_models.test.ts` that sends a `HumanMessage` with **two** base64 PDF blocks through `model.invoke()`, captures the message seen by `handleChatModelStart`, and asserts that **both** blocks have been converted (no `source_type` field, `type === "image_url"`).

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/libs/langchain-core/src/language_models/chat_models.ts
+++ b/libs/langchain-core/src/language_models/chat_models.ts
@@ -171,27 +171,29 @@ export type BaseChatModelCallOptions = BaseLanguageModelCallOptions & {
 function _formatForTracing(messages: BaseMessage[]): BaseMessage[] {
   const messagesToTrace: BaseMessage[] = [];
   for (const message of messages) {
-    let messageToTrace = message;
     if (Array.isArray(message.content)) {
-      for (let idx = 0; idx < message.content.length; idx++) {
-        const block = message.content[idx];
+      let changed = false;
+      const convertedContent = message.content.map((block) => {
         if (isURLContentBlock(block) || isBase64ContentBlock(block)) {
-          if (messageToTrace === message) {
-            // Also shallow-copy content
-            // oxlint-disable-next-line @typescript-eslint/no-explicit-any
-            messageToTrace = new (message.constructor as any)({
-              ...messageToTrace,
-              content: [
-                ...message.content.slice(0, idx),
-                convertToOpenAIImageBlock(block),
-                ...message.content.slice(idx + 1),
-              ],
-            });
-          }
+          changed = true;
+          return convertToOpenAIImageBlock(block);
         }
+        return block;
+      });
+      if (changed) {
+        // oxlint-disable-next-line @typescript-eslint/no-explicit-any
+        messagesToTrace.push(
+          new (message.constructor as any)({
+            ...message,
+            content: convertedContent,
+          })
+        );
+      } else {
+        messagesToTrace.push(message);
       }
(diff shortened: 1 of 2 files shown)


jackjin1997:
Thanks for digging into this. This covers the same issue, function, and callback-path regression as #11292, which has been open since Aug 3 and is already green. #11292 also includes the required @langchain/core changeset and asserts the complete converted tracing payload. To keep review consolidated, could we continue on #11292? I am happy to incorporate any distinct improvement, but I do not see a semantic gap between the current fixes.

PR #11342, opened 2026-08-11
feat(core): mark errors as retryable or not, and stop retrying the ones that aren't

thushanth-bengre-langchain:
## Summary

Retry middleware retried every failure up to `maxRetries`, including deterministic ones like a bad API key or an unknown model. This adds a way to mark an error's retryability and has the retry middleware respect it.

`stampRetryable` sets a non-enumerable `Symbol.for("langchain.errors.retryable")` directly on the error, leaving its class and shape intact, so a provider SDK's own error can be classified in place without breaking `instanceof` for consumers. Errors marked non-retryable now fail on the first attempt; unclassified errors retry exactly as before, and an explicit `retryOn` is unaffected.

## Changes

- Add `stampRetryable(error, retryable)` and `getRetryable(error)` to `@langchain/core/errors`. `getRetryable` returns `undefined` for unclassified errors and follows the `.cause` chain, depth-capped against cycles.
- Mark `ModelAbortError` and `ContextOverflowError` non-retryable at construction, and mark what `AsyncCaller` already classifies — `STATUS_NO_RETRY` codes, aborts, and exhausted quota as non-retryable, rate limits as retryable. It previously acted on these internally with no way to communicate the verdict to an outer retry loop.
- Change the default `retryOn` for `modelRetryMiddleware` and `toolRetryMiddleware` to `getRetryable(error) ?? true`.
- Mark provider errors in Google, OpenAI, Anthropic, and Fireworks — one commit each, covering the statuses each dispatcher already recognizes. Google's `RequestError` publishes the verdict from its existing `isRetryable()`.
- Fix `FireworksEmbeddings` throwing a plain `Error` with the HTTP status only inside the message string, hiding it from both the retry middleware and `AsyncCaller`. The status is now a field; message text is unchanged.
- Move the `@langchain/core` peer range to `workspace:^` for `@langchain/google` and `@langchain/fireworks`.
- Document that a custom `onFailedAttempt` replaces the built-in handler and opts out of its marking.
- Add coverage for symbol semantics, cause-chain traversal, per-provider status mapping, and the middleware behavior change end-to-end.
- Add a changeset: minor for `@langchain/core`, `langchain`, `@langchain/google`, and `@langchain/fireworks`; patch for `@langchain/openai` and `@langchain/anthropic`.

### DIFF
--- a/.changeset/brave-moons-invent.md
+++ b/.changeset/brave-moons-invent.md
@@ -1,18 +1,9 @@
 ---
 "@langchain/google": minor
-"@langchain/openai": patch
-"@langchain/anthropic": patch
-"@langchain/fireworks": minor
 ---
 
-feat(providers): mark provider errors as retryable or not
+feat(google): mark Google provider errors as retryable or not
 
-Builds on `stampRetryable` in `@langchain/core`, so the retry middleware can tell a transient provider failure from a deterministic one. Each dispatcher marks the errors it already recognizes; the SDK's own error object is marked in place, so `instanceof` against the provider's error classes is unaffected. Statuses a dispatcher does not recognize stay unmarked and keep retrying as before.
+Builds on `stampRetryable` in `@langchain/core` so the retry middleware can tell a transient failure from a deterministic one. Timeouts, rate limits, and server errors are marked retryable; bad credentials, blocked prompts, invalid tools, and invalid input non-retryable. Errors that may succeed on a regeneration stay unmarked and retry as before.
 
-**`@langchain/google`**: `RequestError` publishes the verdict from its existing `isRetryable()`, and `AuthError` is marked from its status code, so a bad credential is non-retryable while a transient failure of the auth endpoint itself is not. `ConfigurationError`, `PromptBlockedError`, `InvalidToolError`, `ToolCallNotFoundError`, and `InvalidInputError` are marked non-retryable. `NoCandidatesError` and `MalformedOutputError` are deliberately left unmarked, since both can resolve on a regeneration.
-
-**`@langchain/openai`** and **`@langchain/anthropic`**: connection timeouts and rate limits retryable; user aborts, context overflow, invalid tool results, authentication failures, and unknown models non-retryable.
-
-**`@langchain/fireworks`**: fixes `FireworksEmbeddings` throwing a plain `Error` with the HTTP status only inside the message string. With no `status` field, neither the retry middleware nor `AsyncCaller` could act on it, so a bad API key was retried to exhaustion on every embed call. The status is now a field and the error is marked; the message text is unchanged.
-
-`@langchain/google` and `@langchain/fireworks` move their `@langchain/core` peer range from `^1.0.0` to `workspace:^`, matching `@langchain/openai` and `@langchain/anthropic`. It publishes as the core version shipped in the same release, which is the one that introduces `stampRetryable`. Left at `^1.0.0`, npm would resolve these packages against a core lacking the export, and the failure would only surface on the error path.
+The `@langchain/core` peer range moves from `^1.0.0` to `workspace:^`, since this release depends on `stampRetryable`.
--- a/.changeset/brave-moons-invent.md
+++ b/.changeset/brave-moons-invent.md
@@ -1,0 +1,18 @@
+---
+"@langchain/google": minor
+"@langchain/openai": patch
+"@langchain/anthropic": patch
+"@langchain/fireworks": minor
+---
+
+feat(providers): mark provider errors as retryable or not
+
+Builds on `stampRetryable` in `@langchain/core`, so the retry middleware can tell a transient provider failure from a deterministic one. Each dispatcher marks the errors it already recognizes; the SDK's own error object is marked in place, so `instanceof` against the provider's error classes is unaffected. Statuses a dispatcher does not recognize stay unmarked and keep retrying as before.
+
+**`@langchain/google`**: `RequestError` publishes the verdict from its existing `isRetryable()`, and `AuthError` is marked from its status code, so a bad credential is non-retryable while a transient failure of the auth endpoint itself is not. `ConfigurationError`, `PromptBlockedError`, `InvalidToolError`, `ToolCallNotFoundError`, and `InvalidInputError` are marked non-retryable. `NoCandidatesError` and `MalformedOutputError` are deliberately left unmarked, since both can resolve on a regeneration.
+
+**`@langchain/openai`** and **`@langchain/anthropic`**: connection timeouts and rate limits retryable; user aborts, context overflow, invalid tool results, authentication failures, and unknown models non-retryable.
+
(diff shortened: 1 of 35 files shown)


.changeset/olive-hounds-repeat.md
hntrl, line 1: can you justify the minor changes? I think worth separating into distinct core, provider package, and langchain changesets

libs/langchain-core/src/errors/index.ts
```diff
@@ -27,6 +27,66 @@ export function addLangChainErrorFields(
...
+  return error;
+}
+
+/**
+ * Read an error's retryability mark, following the `.cause` chain.
+ *
+ * @returns `true`/`false` when marked, `undefined` when unclassified —
+ *   supply your own default, e.g. `getRetryable(error) ?? true`.
+ */
+export function getRetryable(error: unknown): boolean | undefined {
```
hntrl, line 64: ooc in what scenario are we wrapping a would-be retryable error in `.cause`? if all errors are directly emitted from the model

libs/langchain-core/src/errors/index.ts
```diff
@@ -27,6 +27,66 @@ export function addLangChainErrorFields(
...
+        retryableSymbol
+      );
+      if (descriptor) {
+        return descriptor.value as boolean;
+      }
+      current = (current as { cause?: unknown }).cause;
+    } catch {
+      // A throwing `cause` getter shouldn't take down the error path.
+      return undefined;
+    }
```
hntrl: unsafe access, would just do something like `if ("cause" in current)`

libs/langchain-core/src/errors/index.ts
```diff
@@ -103,6 +163,8 @@ export class ModelAbortError extends ns.brand(LangChainError, "model-abort") {
   constructor(message: string, partialOutput?: AIMessageChunk) {
     super(message);
     this.partialOutput = partialOutput;
+    // Retrying would mean ignoring the caller's explicit abort signal.
+    stampRetryable(this, false);
```
hntrl, line 150: do we need to explicitly stamp here? if we don't stamp here isn't it going to be false anyways?

libs/langchain-core/src/utils/async_caller.ts
hntrl, line 1: is async caller going to have any unexpected bearing on how model/tool retry works?

hntrl: like for instance if model retry knows to retry 3 times, and if async caller retries 3 times, we get 9 attempts in practice

libs/langchain-core/src/errors/index.ts
hntrl: disregard, I saw the condition in model retry

libs/providers/langchain-fireworks/package.json
```diff
@@ -28,7 +28,7 @@
     "@langchain/openai": "workspace:*"
   },
   "peerDependencies": {
-    "@langchain/core": "^1.0.0"
+    "@langchain/core": "workspace:^"
```
hntrl, line 31: how come?

.changeset/olive-hounds-repeat.md
thushanth-bengre-langchain: Separated this into 2 changesets. 
One for core+langchain and providers as a different changeset.

libs/langchain-core/src/errors/index.ts
thushanth-bengre-langchain: None in practice — you're right, dropped it

libs/langchain-core/src/utils/async_caller.ts
thushanth-bengre-langchain: It's 21 actually because 7 from async caller and 3 from middleware if retries are not set. This is pre existing. 
This PR will stop deterministic failures from doing that 21 loop. Non-retryable is 1 call now. 

Retryable and Unclassified errors still does the 21 default loops.

libs/providers/langchain-fireworks/package.json
thushanth-bengre-langchain: Since `stampRetryable` ships in core 1.3.0, but thats not yet available, `open-swe` will catch that. 
So followed what we do for other providers.

.changeset/olive-hounds-repeat.md
hntrl, line 1: step away from internal detail language in changeset

.changeset/brave-moons-invent.md
hntrl, line 1: separate per package changeset

libs/langchain-core/src/utils/async_caller.ts
hntrl: we should probably fix this:
* we might need some way to indicate that async caller shouldn't use its retry logic inside of createAgent calls

libs/langchain-core/src/utils/async_caller.ts
```diff
@@ -335,6 +347,9 @@ export interface AsyncCallerParams {
    * Custom handler to handle failed attempts. Takes the originally thrown
    * error object as input, and should itself throw an error if the input
    * error is not retryable.
+   *
+   * Replaces the built-in handler entirely, so supplying one opts out of its
+   * retryability marking (see `stampRetryable` in `@langchain/core/errors`).
```
hntrl: remove

libs/langchain-core/src/utils/async_caller.ts
thushanth-bengre-langchain: This is now addressed. The retry will be handled by `middleware`

.changeset/olive-hounds-repeat.md
```diff
@@ -0,0 +1,15 @@
+---
+"@langchain/core": minor
+"langchain": minor
```
hntrl: patch

.changeset/brave-moons-invent.md
```diff
@@ -0,0 +1,11 @@
+---
+"@langchain/google": minor
```
hntrl: patch

.changeset/warm-berries-shake.md
```diff
@@ -0,0 +1,13 @@
+---
+"@langchain/fireworks": minor
```
hntrl: patch

libs/providers/langchain-fireworks/src/chat_models.ts
```diff
@@ -162,19 +162,22 @@ export class ChatFireworks extends ChatOpenAICompletions<ChatFireworksCallOption
 
   async completionWithRetry(
     request: OpenAIClient.Chat.ChatCompletionCreateParamsStreaming,
-    options?: OpenAICoreRequestOptions
+    options?: OpenAICoreRequestOptions,
+    callerMaxRetries?: number
```
hntrl: couple of places where we're adding positional args for max retries, can we just have it on equivalent params/options?

hntrl (approved):

PR #11345, opened 2026-08-11
fix(aws): classify Bedrock stream-idle timeouts, cover pre-response hang

### LINKED ISSUE #11325: [@langchain/aws] ChatBedrockConverse stream-idle timeout is unclassifiable and puts `lc_error_code` in `cause` instead of on the error
### Checked other resources

- [x] The bug is not resolved by updating to the latest stable version (`@langchain/aws@1.4.3` is current).

### Example Code

The error is constructed in `@langchain/aws/dist/utils/stream_timeout.js`:

```typescript
/** Creates the catchable error thrown when Bedrock stops yielding stream chunks. */
function createStreamIdleTimeoutError(timeout) {
  return new Error(
    `Bedrock Converse stream timed out after ${timeout} ms without receiving a chunk.`,
    { cause: { lc_error_code: "MODEL_STREAM_TIMEOUT" } },
  );
}
```

Inspecting what a caller can actually see:

```typescript
const error = createStreamIdleTimeoutError(120_000); // as thrown by withStreamIdleTimeout

error.name;                    // "Error"          — not distinctive
error.code;                    // undefined
error.$retryable;              // undefined
error.lc_error_code;           // undefined        — it is in `cause`, not on the error
error.cause instanceof Error;  // false            — plain object, so a cause-walk stops here
error.cause.lc_error_code;     // "MODEL_STREAM_TIMEOUT"
```

So a retry predicate has nothing stable to match on:

```typescript
import { modelRetryMiddleware } from "langchain";

modelRetryMiddleware({
  // Cannot use the class-array form: the class is not exported.
  // Cannot match on `name`: it is the generic "Error".
  // Cannot match on `code`/`$retryable`: absent.
  // Only remaining options: the message string, or reaching into `cause`.
  retryOn: (e) => e.cause?.lc_error_code === "MODEL_STREAM_TIMEOUT",
});
```

```
Error: Bedrock Converse stream timed out after 120000 ms without receiving a chunk.
```

`name` is `"Error"`; there is no `code`, `$retryable`, or `lc_error_code` on the
error itself.

### Description

Two small things that combine into "the stall watchdog fires and the run just
dies with no retry".

**1. The error is not classifiable.** `ChatBedrockConverse`'s stream watchdog
throws a bare `Error`. Its class isn't exported, `name` is generic, and there is
no retryable marker — so neither `modelRetryMiddleware` nor any caller-written
predicate can recognise it except by matching the message string.

This is made worse by #11324: `MiddlewareError` wrapping means `name` is the only
property that reliably survives a middleware boundary, and here `name` carries no
information.

**2. `lc_error_code` is in the wrong place.** `@langchain/core`'s own convention
puts it *on* the error:

```js
// @langchain/core/dist/errors/index.js
function addLangChainErrorFields(error, lc_error_code) {
  error.lc_error_code = lc_error_code;   // on the error
  ...
}
```

and core now ships a `LangChainError` base class. Here it is nested inside
`cause` as a plain object instead. Because a plain object isn't an `Error`, it
also terminates any `cause`-chain walk, so even code that walks causes won't find
it.

I expect a stalled stream to be retryable, or at least identifiable. Instead it
is neither. In our production traces this accounted for 21 failures over four
days, none of them retried.

**Suggested fix.** Throw a `LangChainError` subclass — or at minimum set `name`
and `lc_error_code` on the error itself — so it is identifiable by `name` and
consistent with core's convention:

```typescript
class ModelStreamTimeoutError extends LangChainError {
  // name = "ModelStreamTimeoutError"
  // lc_error_code = "MODEL_STREAM_TIMEOUT"
}
```

A distinctive `name` is what makes it usable across a middleware boundary.

---

**Related, possibly worth splitting: `streamIdleTimeout` cannot see an
unanswered request.**

`withStreamIdleTimeout` wraps `response.stream`, which only exists after
`await this.client.send(...)` resolves — i.e. after response headers arrive:

```js
const response = await this.client.send(command, { abortSignal: abortController.signal });
if (response.stream) {
  const streamIdleTimeout = resolveStreamIdleTimeout(options.streamIdleTimeout ?? this.streamIdleTimeout);
  for await (const chunk of withStreamIdleTimeout(response.stream, streamIdleTimeout, abortController)) {
```

A request that is accepted but never answered is entirely outside the watchdog's
window, even though the option reads like a general stall guard. We hit this in
production: a run waited ~180s with no response and no stream, and the only thing
that eventually failed it was an unrelated transport-level `sessionTimeout` we had
configured for memory reasons.

Would you accept a `firstResponseTimeout`, or extending `streamIdleTimeout` to
cover the `client.send` phase? Happy to open a PR if the direction is agreeable.

### System Info

```
@langchain/aws@1.4.3
langchain@1.5.4 (also reproduced on 1.5.5)
@langchain/core@1.2.3 (also reproduced on 1.2.5)
@aws-sdk/client-bedrock-runtime@3.1097.0

platform: mac (Darwin arm64)
Node: v24.14.0
pnpm: 10.14.0
```

thushanth-bengre-langchain:
Fixes #11325 

`ChatBedrockConverse`'s stream watchdog threw a bare `Error` with `lc_error_code` nested under `cause`, so it couldn't be classified via `LangChainError.isInstance()`. It also only started timing once the response stream existed — a hang before Bedrock ever responds wasn't covered by `streamIdleTimeout` at all.

- Added `ModelStreamTimeoutError` (extends `LangChainError`) in `@langchain/core/errors`, alongside `ModelAbortError`/`ContextOverflowError`.
- `createStreamIdleTimeoutError` now throws it instead of the bare `Error`/`cause` hack.
- Added `withRequestIdleTimeout`, which races the initial `client.send()` call against the timeout too, closing the pre-response gap.
- New/updated tests cover both the pre-response hang and the existing mid-stream stall case.

### DIFF
--- a/libs/langchain-core/src/errors/index.ts
+++ b/libs/langchain-core/src/errors/index.ts
@@ -107,6 +107,42 @@ export class ModelAbortError extends ns.brand(LangChainError, "model-abort") {
 }
 
 /**
+ * Error class representing a stalled model stream in LangChain.
+ *
+ * This error is thrown when a streaming model operation stops producing output
+ * (e.g. no response is received, or no new chunk arrives) within a configured
+ * idle timeout window.
+ *
+ * @remarks
+ * - This error extends the {@link LangChainError} base class with the marker `"model-stream-timeout"`.
+ *
+ * @example
+ * ```typescript
+ * try {
+ *   for await (const chunk of await model.stream(input)) {
+ *     // ...
+ *   }
+ * } catch (err) {
+ *   if (ModelStreamTimeoutError.isInstance(err)) {
+ *     // Handle the stall, e.g. retry or surface a timeout to the user
+ *   } else {
+ *     throw err;
+ *   }
+ * }
+ * ```
+ */
+export class ModelStreamTimeoutError extends ns.brand(
+  LangChainError,
+  "model-stream-timeout"
+) {
+  readonly name = "ModelStreamTimeoutError";
+
+  constructor(message?: string) {
+    super(message ?? "The model stream timed out.");
+  }
+}
+
+/**
  * Error class representing a context window overflow in a language model operation.
  *
  * This error is thrown when the combined input to a language model (such as prompt tokens,
--- a/libs/langchain-core/src/errors/index.ts
+++ b/libs/langchain-core/src/errors/index.ts
@@ -107,42 +107,6 @@ export class ModelAbortError extends ns.brand(LangChainError, "model-abort") {
 }
 
 /**
- * Error class representing a stalled model stream in LangChain.
- *
- * This error is thrown when a streaming model operation stops producing output
- * (e.g. no response is received, or no new chunk arrives) within a configured
- * idle timeout window.
- *
- * @remarks
- * - This error extends the {@link LangChainError} base class with the marker `"model-stream-timeout"`.
- *
- * @example
- * ```typescript
- * try {
- *   for await (const chunk of await model.stream(input)) {
- *     // ...
- *   }
- * } catch (err) {
- *   if (ModelStreamTimeoutError.isInstance(err)) {
- *     // Handle the stall, e.g. retry or surface a timeout to the user
- *   } else {
- *     throw err;
- *   }
- * }
- * ```
- */
-export class ModelStreamTimeoutError extends ns.brand(
-  LangChainError,
-  "model-stream-timeout"
-) {
-  readonly name = "ModelStreamTimeoutError";
-
-  constructor(message?: string) {
-    super(message ?? "The model stream timed out.");
-  }
-}
-
-/**
  * Error class representing a context window overflow in a language model operation.
  *
  * This error is thrown when the combined input to a language model (such as prompt tokens,
--- a/libs/providers/langchain-aws/src/utils/stream_timeout.ts
+++ b/libs/providers/langchain-aws/src/utils/stream_timeout.ts
@@ -1,5 +1,3 @@
-import { addLangChainErrorFields } from "./errors.js";
-
 /** Default milliseconds to wait for the first or next Bedrock stream chunk. */
 export const DEFAULT_STREAM_IDLE_TIMEOUT = 60_000;
 
@@ -24,7 +22,8 @@ export function createStreamIdleTimeoutError(
   const error = new Error(
     `Bedrock Converse timed out after ${timeout} ms while waiting for the ${phase}.`
   );
-  return addLangChainErrorFields(error, "MODEL_STREAM_TIMEOUT");
+  (error as { lc_error_code?: string }).lc_error_code = "MODEL_STREAM_TIMEOUT";
+  return error;
 }
 
 /** Creates an AbortController that follows an optional caller-provided signal. */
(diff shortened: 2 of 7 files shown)


libs/langchain-core/src/errors/index.ts
```diff
@@ -106,6 +106,42 @@ export class ModelAbortError extends ns.brand(LangChainError, "model-abort") {
   }
 }
 
+/**
```
hntrl: lets not add this yet

libs/providers/langchain-aws/src/utils/stream_timeout.ts
```diff
@@ -14,11 +16,13 @@ export function resolveStreamIdleTimeout(timeout: number | undefined) {
   return timeout;
 }
 
-/** Creates the catchable error thrown when Bedrock stops yielding stream chunks. */
-export function createStreamIdleTimeoutError(timeout: number) {
-  return new Error(
-    `Bedrock Converse stream timed out after ${timeout} ms without receiving a chunk.`,
-    { cause: { lc_error_code: "MODEL_STREAM_TIMEOUT" } }
```
hntrl, line 21: does `{ lc_error_code: "MODEL_STREAM_TIMEOUT" }` work instead of `{ cause: { lc_error_code: "MODEL_STREAM_TIMEOUT" } }`

thushanth-bengre-langchain: This worked.

libs/providers/langchain-aws/src/utils/stream_timeout.ts
```diff
@@ -14,12 +14,16 @@ export function resolveStreamIdleTimeout(timeout: number | undefined) {
...
+/** Creates the catchable error thrown when Bedrock stops responding within the idle timeout. */
+export function createStreamIdleTimeoutError(
+  timeout: number,
+  phase: "initial response" | "next chunk" = "next chunk"
+) {
+  const error = new Error(
+    `Bedrock Converse timed out after ${timeout} ms while waiting for the ${phase}.`
   );
+  (error as { lc_error_code?: string }).lc_error_code = "MODEL_STREAM_TIMEOUT";
```
hntrl, line 25:
```suggestion
  const error = addLangChainErrorFields(
    new Error(
      `Bedrock Converse timed out after ${timeout} ms while waiting for the ${phase}.`
    ),
    "MODEL_STREAM_TIMEOUT" as LangChainErrorCodes
  );
```

libs/providers/langchain-aws/src/utils/stream_timeout.ts
```diff
@@ -40,6 +44,40 @@ export function createLinkedAbortController(signal?: AbortSignal) {
...
+export async function withRequestIdleTimeout<T>(
+  request: Promise<T>,
+  timeout: number | undefined,
+  abortController: AbortController
+): Promise<T> {
+  if (timeout === undefined) {
+    return request;
+  }
+
+  request.catch(() => undefined);
```
hntrl:
```suggestion
```

libs/providers/langchain-aws/src/utils/stream_timeout.ts
thushanth-bengre-langchain: Can't use this function here since the troubleshooting docs does not have this error code yet. 
So the troubleshooting link that shows up points to a non-existing page. 
Either we have to update the docs to add this error code or keep it as it is.

hntrl (approved):

PR #11378, opened 2026-08-15
fix(langchain): preserve error identity in MiddlewareError.wrap

### LINKED ISSUE #11324: MiddlewareError.wrap discards `code`, custom properties, and the prototype, silently breaking both documented `retryOn` forms of modelRetryMiddleware

haoku123:
Fixes #11324.

## Problem

`AgentNode` wraps every middleware's throw in `MiddlewareError` (`agents/nodes/AgentNode.ts`), once per middleware layer. `MiddlewareError.wrap` only copies `message`, `name`, and `cause`, so:

- `wrapped instanceof MyError` is `false`
- `wrapped.constructor === MyError` is `false` — breaking the `retryOn: [MyError]` array form of `modelRetryMiddleware` (which matches via `error.constructor === ErrorConstructor`)
- own properties like `code`, `statusCode`, or custom flags like `\` are dropped — breaking the documented `retryOn` predicate patterns (`error.name === "HTTPError" && "statusCode" in error`)

Result: with any middleware inside the retry, both documented `retryOn` forms silently stop retrying.

## Solution

`wrap` now rebuilds the original error's identity on the wrapped instance:

```ts
if (error instanceof Error) {
  Object.setPrototypeOf(wrapped, Object.getPrototypeOf(error));
  Object.assign(wrapped, error);
  if (error.stack) wrapped.stack = error.stack;
}
```

This is safe with `MiddlewareError`'s existing recognition logic: `Symbol.hasInstance` checks the `~brand` own property (set in the constructor), not the prototype chain, so the wrapped error is still detected as a `MiddlewareError` and the `cause`-chain unwrap loop in `ToolNode` keeps working.

## Tests

New `middlewareErrorWrapping.test.ts` covers:

- unit: prototype chain / `constructor` / `instanceof` / own props / `code` / `\` / message / name / cause / stack preserved, including multi-layer wrapping
- integration: `modelRetryMiddleware` array form retries 3x with and without an extra middleware; predicate form receives the original identity; the documented `name + statusCode` predicate retries 3x

Verified locally (Node 24.6): new suite 6/6, plus `middlewareErrorHandling` + `modelRetry` + `toolRetry` suites 60/60, oxlint clean.

### DIFF
--- a/.changeset/nice-tigers-sell.md
+++ b/.changeset/nice-tigers-sell.md
@@ -1,0 +1,5 @@
+---
+"langchain": patch
+---
+
+Fix `MiddlewareError.wrap` to preserve the wrapped error's prototype chain and own properties, so `modelRetryMiddleware`'s documented `retryOn` forms keep working when the error is wrapped by other middleware.
--- a/libs/langchain/src/agents/errors.ts
+++ b/libs/langchain/src/agents/errors.ts
@@ -124,7 +124,27 @@ export class MiddlewareError extends Error {
     if (isGraphBubbleUp(error)) {
       return error;
     }
-    return new MiddlewareError(error, middlewareName);
+
+    const wrapped = new MiddlewareError(error, middlewareName);
+
+    if (error instanceof Error) {
+      // Preserve the original error's identity:
+      // - its prototype chain, so `instanceof` checks and the `retryOn`
+      //   constructor-matching form keep working,
+      // - its own enumerable properties (`code`, custom flags, ...),
+      // - its stack trace.
+      //
+      // The wrapped error is still recognized as a `MiddlewareError` because
+      // `Symbol.hasInstance` checks the `~brand` property, which is set in the
+      // constructor and remains an own property of the instance.
+      Object.setPrototypeOf(wrapped, Object.getPrototypeOf(error));
+      Object.assign(wrapped, error);
+      if (error.stack) {
+        wrapped.stack = error.stack;
+      }
+    }
+
+    return wrapped;
   }
 
   /**
--- a/libs/langchain/src/agents/tests/middlewareErrorWrapping.test.ts
+++ b/libs/langchain/src/agents/tests/middlewareErrorWrapping.test.ts
@@ -1,0 +1,201 @@
+/* oxlint-disable @typescript-eslint/no-explicit-any */
+import { describe, it, expect, vi } from "vitest";
+import { HumanMessage } from "@langchain/core/messages";
+import { StructuredTool } from "@langchain/core/tools";
+import { RunnableBinding } from "@langchain/core/runnables";
+
+import { createAgent, createMiddleware } from "../index.js";
+import { modelRetryMiddleware } from "../middleware/modelRetry.js";
+import { MiddlewareError } from "../errors.js";
+import { FakeToolCallingModel } from "./utils.js";
+
+class RetryableError extends Error {
+  code = "ECONNRESET";
+
+  $retryable = true;
+
+  constructor() {
+    super("boom");
+    this.name = "RetryableError";
+  }
+}
+
+class HttpError extends Error {
+  statusCode = 503;
+
+  constructor() {
+    super("service unavailable");
+    this.name = "HTTPError";
+  }
+}
+
+describe("MiddlewareError.wrap identity preservation (#11324)", () => {
+  it("preserves own properties, prototype chain, and stack", () => {
+    const raw = new RetryableError();
+    const wrapped = MiddlewareError.wrap(raw, "testMiddleware") as RetryableError;
+
+    // prototype chain is preserved, so instanceof and constructor checks work
+    expect(wrapped).toBeInstanceOf(RetryableError);
+    expect(wrapped).toBeInstanceOf(Error);
+    expect(wrapped.constructor).toBe(RetryableError);
+
+    // own enumerable properties survive the wrap
+    expect(wrapped.code).toBe("ECONNRESET");
+    expect(wrapped.$retryable).toBe(true);
+
+    // message, name, cause, and stack are intact
+    expect(wrapped.message).toBe("boom");
+    expect(wrapped.name).toBe("RetryableError");
+    expect(wrapped.cause).toBe(raw);
+    expect(wrapped.stack).toBe(raw.stack);
+
+    // still recognized as a MiddlewareError through the ~brand check
+    expect(MiddlewareError.isInstance(wrapped)).toBe(true);
+    expect(wrapped).toBeInstanceOf(MiddlewareError);
+  });
+
+  it("is chainable across multiple wrap layers", () => {
+    const raw = new RetryableError();
+    const once = MiddlewareError.wrap(raw, "inner");
+    const twice = MiddlewareError.wrap(once, "outer");
+
+    expect(twice).toBeInstanceOf(RetryableError);
+    expect(twice.constructor).toBe(RetryableError);
+    expect((twice as RetryableError).code).toBe("ECONNRESET");
+    expect(MiddlewareError.isInstance(twice)).toBe(true);
+  });
+});
+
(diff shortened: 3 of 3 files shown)


HDPark95:
Disclosure: I have an open PR for the same issue (#11330), so this is a note, not a verdict.

Ran this branch: unit suite green (60 files, 874 tests). Measured: after `wrap` the value satisfies both `instanceof MiddlewareError` and `instanceof TimeoutError`, and is a copy, not the thrown instance.

The other call site: `agents/utils.ts` filters propagated errors through a `downstreamErrors` set before calling `wrap`, so there `wrap` only sees errors raised inside the middleware, where adopting the original prototype and `name` drops that attribution. `AgentNode.ts` has no such filter.

haoku123:
Thanks for running the full suite and for the fair disclosure — that verification is genuinely useful (874 tests, `instanceof` both ways, copy semantics).

On the `agents/utils.ts` attribution concern, two points:

1. **`name` was already copied before this PR.** The pre-existing constructor does `this.name = error instanceof Error ? error.name : ...`, so for `Error` subclasses the middleware attribution in `name` was already gone — this PR changes nothing about `name` behavior. The only new thing is the prototype chain and own properties.

2. **The official recognition channel is unaffected.** `MiddlewareError.isInstance` / `Symbol.hasInstance` check the `~brand` own property, which survives `setPrototypeOf` (own properties are untouched). So `wrapped instanceof MiddlewareError` stays true — as your measurement confirms — and downstream code that classifies via `MiddlewareError.isInstance` keeps working.

On the structural point: in `agents/utils.ts` the `downstreamErrors` set filters by reference *before* `wrap` is called, so propagated errors never reach `wrap` — they are rethrown unchanged. `wrap` only ever sees errors raised inside the middleware itself, where preserving the original identity is exactly the behavior #11324 needs (that is the case `modelRetryMiddleware`'s `retryOn` forms rely on). So there is no scenario where this PR steals attribution from a propagated error.

If you think middleware provenance should be *more* explicit (e.g. a dedicated `middleware` field on `MiddlewareError`), that could be a worthwhile follow-up, but it is a separate enhancement from the identity-preservation bug this PR fixes. Happy to compare notes with #11330 if you see overlap in our approaches.

PR #11387, opened 2026-08-16
fix(core): make EventStreamCallbackHandler end handlers idempotent

### LINKED ISSUE #11360: EventStreamCallbackHandler warns "Run ID … not found in run map" for every run when tracing is enabled and a tool invokes a nested graph (duplicate handler delivery not covered by #11190)
### Example code

```js
// repro.mjs — run: LANGCHAIN_TRACING_V2=true LANGCHAIN_API_KEY=lsv2_dummy LANGCHAIN_ENDPOINT=http://127.0.0.1:9 node repro.mjs
import { createAgent, tool, FakeToolCallingModel } from "langchain";
import { z } from "zod";

const innerAgent = createAgent({
  model: new FakeToolCallingModel({ toolCalls: [[]] }),
  tools: [],
});

const nested = tool(
  async (_input, config) => {
    await innerAgent.invoke({ messages: [{ role: "user", content: "hi" }] }, config);
    return "done";
  },
  { name: "nested_graph_tool", schema: z.object({}) },
);

const agent = createAgent({
  model: new FakeToolCallingModel({
    toolCalls: [[{ name: "nested_graph_tool", args: {}, id: "1", type: "tool_call" }], []],
  }),
  tools: [nested],
});

for await (const _ev of agent.streamEvents(
  { messages: [{ role: "user", content: "go" }] },
  { version: "v2" },
)) { /* drain */ }
```

### Error message and stack trace

```
Error in handler EventStreamCallbackHandler, handleChainEnd: Error: onChainEnd: Run ID 019ff6… not found in run map.
Error in handler EventStreamCallbackHandler, handleLLMEnd: Error: onLLMEnd: Run ID 019ff6… not found in run map.
Error in handler EventStreamCallbackHandler, handleToolEnd: Error: onToolEnd: Run ID 019ff6… not found in run map.
```

(4 warnings for this minimal repro; a real nested-agent app served by `langgraph dev` produces a dozen+ per turn.)

### Description

Running the script above:

- `LANGCHAIN_TRACING_V2` unset → **0 warnings**
- `LANGCHAIN_TRACING_V2=true` → **4 warnings**
- flat agent (no nested-graph tool) → 0 warnings either way

The warnings are non-fatal (`raiseError` is false, so the callback dispatcher logs and swallows them), but `langgraph dev` consumes every run via `streamEvents({version: "v2"})`, so any real nested-agent app with tracing enabled emits a burst of them on every turn, drowning the dev logs.

**What's happening:** we instrumented `EventStreamCallbackHandler`'s end handlers to record already-ended run ids. Every warning fires with `alreadyEnded=true` — the handler already processed that exact run's end once. The same handler instance ends up registered twice in the callback manager used inside the nested graph, so `handleChainEnd`/`handleLLMEnd`/`handleToolEnd` fan out to it twice concurrently (`Promise.all`); the first delivery deletes the `runInfoMap` entry (the end handlers delete before validating), and the second throws.

This looks like the same mechanism as #11189 ("No chain run to end" from `LangChainTracer` in nested graphs). PR #11190 fixed that by coalescing duplicate handler copies sharing run bookkeeping — but the fix is hard-scoped to `handler instanceof LangChainTracer` (`coalesceTracers` in `callbacks/manager.js`), so `EventStreamCallbackHandler` (and presumably any other non-tracer handler) still gets duplicated when the tracing-enabled configure path runs for a tool-invoked nested graph.

A related asymmetry noticed while debugging: `EventStreamCallbackHandler` implements `onToolError` but not `onChainError`, so error-path chain runs (e.g. a `GraphInterrupt` bubbling out of a tool) never remove their `runInfoMap` entry.

**Expected behavior:** each end event is delivered once per handler instance — no warnings, exactly like the non-tracing case.

### System Info

- langchain 1.5.4
- @langchain/core 1.2.4 (reproduced identically with 1.2.5 via npm override)
- @langchain/langgraph 1.4.8
- @langchain/langgraph-api 1.4.4 (`langgraph dev`) — where the noise surfaces in practice
- Node.js 26.0.0, macOS 15 (darwin 25.5.0)

haoku123:
Fixes #11360.

## Problem

When a compiled graph (e.g. an agent built with `createAgent`) runs inside a nested graph with LangSmith tracing enabled, the same `EventStreamCallbackHandler` instance ends up registered twice in the callback manager. `handleChainEnd` / `handleLLMEnd` / `handleToolEnd` / `handleRetrieverEnd` fan out to it twice concurrently; the first delivery deletes the `runInfoMap` entry, and the second throws `Run ID … not found in run map`. The warnings are non-fatal, but `langgraph dev` consumes runs via `streamEvents`, so real nested-agent apps emit a burst of them every turn.

`#11190` fixed the same class of issue for `LangChainTracer` at the callback-manager level (tracer coalescing). This PR addresses the remaining handler family.

## Solution

`EventStreamCallbackHandler` now overrides the four `handle*End` entry points with an idempotent guard: when the run id is not in `runInfoMap`, the run was already ended by a previous delivery (or was never started), so the handler returns without throwing. The `on*End` hooks receive the same guard so directly-invoked hooks are idempotent too.

The end event is still emitted exactly once — the first delivery sends it, duplicate deliveries are skipped rather than re-sent.

## Tests

Added `event_stream.test.ts` with four tests: a duplicated chain/llm/tool/retriever end emits its end event exactly once and does not throw. Verified non-vacuous: all four fail against the previous implementation. Full tracers suite: 40 tests passing, oxlint clean.

### DIFF
--- a/.changeset/fix-context-overflow-status.md
+++ b/.changeset/fix-context-overflow-status.md
@@ -1,0 +1,5 @@
+---
+"@langchain/core": patch
+---
+
+Preserve the underlying HTTP status in `ContextOverflowError.fromError` so retry logic and callers can still inspect it
--- a/.changeset/fix-event-stream-duplicate-end.md
+++ b/.changeset/fix-event-stream-duplicate-end.md
@@ -1,0 +1,5 @@
+---
+"@langchain/core": patch
+---
+
+Make `EventStreamCallbackHandler` end handlers idempotent so duplicate handler delivery in nested graph runs no longer throws "Run ID not found in run map" warnings
--- a/.changeset/fix-responses-service-tier.md
+++ b/.changeset/fix-responses-service-tier.md
@@ -1,0 +1,5 @@
+---
+"@langchain/openai": patch
+---
+
+Honor the per-call `service_tier` option on the Responses API path instead of only reading the instance value
--- a/.changeset/fix-zod-v4-ref-siblings.md
+++ b/.changeset/fix-zod-v4-ref-siblings.md
@@ -1,0 +1,5 @@
+---
+"@langchain/openai": patch
+---
+
+Inline `$ref` sibling keywords in zod v4 interop response formats so OpenAI strict mode no longer rejects schemas where a field is chained `.default().describe()`
--- a/libs/langchain-core/src/errors/index.ts
+++ b/libs/langchain-core/src/errors/index.ts
@@ -197,6 +197,13 @@ export class ContextOverflowError extends ns.brand(
    */
   cause?: Error;
 
+  /**
+   * HTTP status of the underlying error, preserved by
+   * {@link ContextOverflowError.fromError} so retry logic and callers can
+   * still inspect it after wrapping.
+   */
+  status?: number;
+
   constructor(message?: string) {
     super(message ?? "Input exceeded the model's context window.");
     // The same oversized input fails identically; it needs trimming, not another attempt.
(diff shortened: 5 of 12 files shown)


davidgovea:
I opened https://github.com/langchain-ai/langchainjs/pull/11601 to address the duplicate-instance dispatch behind #11360 at core's callback configuration boundary. It normalizes active and inheritable handlers independently after tracer coalescing, so configured runs deliver each lifecycle event once per handler instance.

The useful distinction is that this also prevents duplicate starts, token callbacks, and custom-handler side effects. Your end guards can still help with directly repeated lifecycle calls that never pass through configuration; our change doesn't cover that case.

The implementation is +5/-2 lines with 18 regression/compatibility tests, including real EventStream delivery. All 1,531 core tests pass, with one skipped. Sharing here so we can coordinate on the dispatch fix and any end-handler protection that's still useful.

Issue #7848, opened 2025-03-14
@langchain/google-genai does not properly handle MALFORMED_FUNCTION_CALL


ymfpfp:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain.js documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain.js rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```js
try {
  response = await modelWithTools.invoke([
    new SystemMessage(actionPrompt(mappedChatHistory, toolsToUse)),
    new HumanMessage(mappedChatHistory)
  ]);
} catch (err) {
  sparrowMonitor.error(`Ran into error while sampling, try ${tries}`, err);
}
```

### Error Message and Stack Trace (if applicable)

I have an error logger wrapper - this is the stack it returns:
```
[12:11:32.944] ERROR: Ran into error while sampling, try 0 {"reqId":"ITwa6etzSrhGZdlyILjNL","reqPath":"POST /api/v1/_internal/benchmark/actions","reqUrl":"POST /api/v1/_internal/benchmark/actions","levelLabel":"ERROR"}
    err: {
      "type": "TypeError",
      "message": "Cannot read properties of undefined (reading 'length')",
      "stack":
          Ran into error while sampling, try 0: Cannot read properties of undefined (reading 'length')
              at mapGenerateContentResultToChatResult (/Users/jc/Documents/node_modules/.pnpm/@langchain+google-genai@0.1.11_@langchain+core@0.2.26_openai@4.73.0_zod@3.23.8___zod@3.23.8/node_modules/@langchain/google-genai/dist/utils/common.cjs:229:32)
              at ChatGoogleGenerativeAI._generate (/Users/jc/Documents/node_modules/.pnpm/@langchain+google-genai@0.1.11_@langchain+core@0.2.26_openai@4.73.0_zod@3.23.8___zod@3.23.8/node_modules/@langchain/google-genai/dist/chat_models.cjs:661:87)
              at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
              at async Promise.allSettled (index 0)
[...]

PR #11400, opened 2026-08-18
fix(google-genai): throw ContentBlockedError when Gemini candidate has no content

### LINKED ISSUE #10741: @langchain/google-genai: TypeError in convertResponseContentToChatGenerationChunk when streaming candidate has no content (partial fix in #10319)
## Summary

`convertResponseContentToChatGenerationChunk` in `libs/providers/langchain-google-genai/src/utils/common.ts` throws `TypeError: Cannot read properties of undefined (reading 'parts')` when Gemini returns a streaming chunk whose candidate has `content: undefined` (e.g., when a safety/recitation filter fires mid-stream, or `finishReason` is `SAFETY` / `RECITATION` / `OTHER`).

This is the streaming counterpart to #10318. That issue was closed by #10319, which added a `?.` guard to `mapGenerateContentResultToChatResult` (the non-streaming path) but did **not** apply the same guard to `convertResponseContentToChatGenerationChunk`. The streaming path remains broken in the latest release (\`2.1.28\`) and on \`main\`.

## Reproduction

Any streaming call to a Gemini model that triggers a safety/recitation block mid-stream will surface this. Minimal repro:

\`\`\`ts
import { ChatGoogleGenerativeAI } from \"@langchain/google-genai\";

const llm = new ChatGoogleGenerativeAI({ model: \"gemini-2.5-flash\" });

// Any prompt that the model partially answers and then aborts via a safety
// filter mid-stream will produce a streaming chunk shaped like:
//   { candidates: [{ content: undefined, finishReason: \"SAFETY\", ... }], ... }
for await (const chunk of await llm.stream(\"<prompt that trips a filter>\")) {
  console.log(chunk);
}
// → TypeError: Cannot read properties of undefined (reading 'parts')
\`\`\`

## Root cause

[\`libs/providers/langchain-google-genai/src/utils/common.ts\` line 785](https://github.com/langchain-ai/langchainjs/blob/main/libs/providers/langchain-google-genai/src/utils/common.ts#L785) destructures \`content\` off the candidate and then accesses \`.parts\` without a guard:

\`\`\`ts
const functionCalls = candidateContent.parts?.reduce(  // ← throws if candidateContent is undefined
  ...
);
\`\`\`

The same function is internally inconsistent about whether \`candidateContent\` may be undefined:

- L785 — unsafe direct access (\`candidateContent.parts?\`)
- L801 — already uses \`candidateContent?.parts\`
- L898 — explicitly handles \`!candidateContent ? undefined : candidateContent.role\`

So the rest of the function clearly anticipates the undefined case; only L785 was missed.

## Suggested fix

Two small options, both safe:

1. Match the upstream pattern from #10319 (early-return \`null\` when no content):
   \`\`\`ts
   if (!candidateContent) return null;
   \`\`\`

2. Or simply add the missing optional chain to mirror L644 in \`mapGenerateContentResultToChatResult\`:
   \`\`\`ts
   const functionCalls = candidateContent?.parts?.reduce( ... );
   \`\`\`

I'm happy to open a PR with whichever approach the maintainers prefer (option 1 is what we currently apply via a local \`pnpm patch\` on \`2.1.28\`, and matches the surrounding fix from #10319).

## Affected versions

\`@langchain/google-genai\` 2.1.21 → 2.1.28 (current) and \`main\`.

## Related

- Issue #10318 (closed) — same bug, non-streaming path
- PR #10319 (merged) — fixed only \`mapGenerateContentResultToChatResult\`
- PR #9659 (closed) and PR #9510 (closed) — earlier attempts that also covered this
- deepagentsjs#70 — open, downstream report of the same crash

thushanth-bengre-langchain:
Fixes #10741 & #7848 

## Summary

`ChatGoogleGenerativeAI` silently returned empty content (non-streaming) or crashed with a raw `TypeError` (streaming) whenever Gemini returned a candidate with no `content` — a safety/recitation block, a malformed function call, or a thinking model burning its whole token budget on reasoning. Both paths now throw a typed, catchable `ContentBlockedError` carrying `finishReason`/`blockReason`, instead of crashing or returning something misleading.

## Breaking change

Non-streaming calls (`.invoke()`, `.generate()`, `.batch()`) that previously resolved with empty `content`/`generations` on a blocked/malformed response now **throw** instead. See the changeset for the migration note. Marked `minor` because of the breaking changes.


.changeset/throw-content-blocked-error.md
```diff
@@ -0,0 +1,7 @@
+---
+"@langchain/google-genai": minor
+---
+
+fix(google-genai): throw a typed, catchable `ContentBlockedError` instead of crashing (streaming) or silently returning empty content / an empty `generations` array (non-streaming) when Gemini returns a candidate with no content (safety/recitation block, malformed function call) or no candidates at all (blocked prompt).
+
+**Migration:** if your code inspects a response for signs of a block — an empty `content`/`tool_calls`, `additional_kwargs.finishReason`, or an empty `generations` array from `.generate()`/`.batch()` — that code path is now unreachable; the call throws instead. Wrap it in `try/catch` and check `ContentBlockedError.isInstance(error)` (exposed from `@langchain/google-genai`), inspecting `error.finishReason` / `error.blockReason`.
```
hntrl, line 7: eh I don't think we need to call out

libs/providers/langchain-google-genai/src/utils/common.ts
```diff
@@ -783,10 +786,9 @@ export function convertResponseContentToChatGenerationChunk(
   const [candidate] = response.candidates;
   const { content: candidateContent, ...generationInfo } = candidate;
   if (!candidateContent) {
-    // No content on the candidate (e.g. safety/recitation filter triggered
-    // mid-stream, or finishReason is SAFETY/RECITATION/OTHER). Skip this
-    // chunk rather than throwing on candidateContent.parts.
-    return null;
+    throw new ContentBlockedError({
+      finishReason: generationInfo.finishReason,
+    });
```
hntrl, line 791: we probably shouldn't be throwing this in the streaming path

libs/providers/langchain-google-genai/src/utils/errors.ts
```diff
@@ -0,0 +1,71 @@
...
+ * try {
+ *   await model.invoke("...");
+ * } catch (error) {
+ *   if (ContentBlockedError.isInstance(error)) {
+ *     console.log(`Blocked: ${error.finishReason}`);
+ *   }
+ * }
+ * ```
+ */
+export class ContentBlockedError extends ns.brand(
```
hntrl, line 49: `EmptyContentError`

hntrl (approved):

PR #11402, opened 2026-08-18
fix(core): include tool_call blocks and skip empty text blocks in ChatVertexAI contentBlocks

### LINKED ISSUE #9950: ChatVertexAI tool call not added in message parts
### Checked other resources

### Example Code

```typescript
import dotenv from 'dotenv';
import { createAgent } from 'langchain';
import { z } from 'zod';
import { ChatVertexAI } from '@langchain/google-vertexai';

dotenv.config();

const agent = createAgent({
    model: new ChatVertexAI({
        model: 'gemini-3-pro-preview',
        location: 'global',
    }),
    tools: [
        tool(({ a, b }) => String(a + b), {
            name: 'add',
            description: 'Adds two numbers together',
            schema: z.object({
                a: z.number(),
                b: z.number(),
            }),
        }),
        // get value
        tool(
            ({ key }) => {
                return key === 'a' ? 1 : 2;
            },
            {
                name: 'get_value',
                description: 'Gets a value for a given key, either "a" or "b"',
                schema: z.object({
                    key: z.string(),
                }),
            },
        ),
        //print number
        tool(
            ({ value }) => {
                console.log('Value:', value);
                return 'Printed value to console';
            },
            {
                name: 'print_value',
                description: 'Prints a value to the console',
                schema: z.object({
                    value: z.number(),
                }),
            },
        ),
    ],
    systemPrompt: `You are a calculator, first, in parallel ask for the values of "a" and "b" using the "get_value" tool. Once you have both values, use the "add" tool to add them together. Finally, use the "print_value" tool to print the result. the explain a summary of what you did and why.`,
});

async function runAgent() {
    const response = await agent.invoke({
        messages: [
            {
                role: 'user',
                content: 'Calculate the sum of a and b',
            },
        ],
    });
    const parts = response.messages.map((v) => [v.type, v.contentBlocks]);
    console.log('Agent response:', JSON.stringify(parts, null, 2));
}

runAgent();
```

<details>

```json
[
  [
    "human",
    [
      {
        "type": "text",
        "text": "Calculate the sum of a and b"
      }
    ]
  ],
  [
    "ai",
    [
      {
        "type": "text",
        "text": ""
      }
    ]
  ],
  [
    "tool",
    [
      {
        "type": "text",
        "text": "1"
      }
    ]
  ],
  [
    "tool",
    [
      {
        "type": "text",
        "text": "2"
      }
    ]
  ],
  [
    "ai",
    [
      {
        "type": "text",
        "text": ""
      }
    ]
  ],
  [
    "tool",
    [
      {
        "type": "text",
        "text": "3"
      }
    ]
  ],
  [
    "ai",
    [
      {
        "type": "text",
        "text": ""
      }
    ]
  ],
  [
    "tool",
    [
      {
        "type": "text",
        "text": "Printed value to console"
      }
    ]
  ],
  [
    "ai",
    [
      {
        "type": "text",
        "text": "I have successfully completed the task.\n\n**Summary of actions:**\n1.  **Retrieval**: I requested the values for variables \"a\" and \"b\" simultaneously to maximize efficiency. The system returned `1` for \"a\" and `2` for \"b\".\n2.  **Calculation**: Using the retrieved numbers, I performed an addition operation ($1 + 2$).\n3.  **Output**: Finally, I printed the resulting sum, `3`, to the console.\n\n**Why:**\nThis sequence ensured that the correct values were obtained before processing, the mathematical operation was handled by the dedicated calculation tool, and the final result was displayed as requested."
      }
    ]
  ]
]
```

</details>

### Description

When using tool calls with ChatVertexAI, tool call contents have type `text` instead of `tool_call` and there is no metadata inside of it.

If i use openai instead, each tool call makes a `tool_call` content block inside of the message.

There is also no thought signature inside of them. I'd need to rebuild the content blocks out of the messages myself.

### System Info

langchain@1.2.18 | MIT | deps: 5 | versions: 381
Windows
node 24

thushanth-bengre-langchain:
Fixes #9950

- `ChatVertexTranslator.translateContent` only read `message.content`, never `message.tool_calls`, so Vertex/Gemini tool calls surfaced as an empty `{type: "text", text: ""}` block instead of a `tool_call block`, unlike OpenAI/GoogleGenAI.
- Adds a `tool_call` block per entry in `message.tool_calls` (matches OpenAI's field shape: `id`, `name`, `args`).

- Skips empty `text` blocks in the same translator, removing the spurious block that was masking the missing tool call.

### DIFF
--- a/libs/langchain-core/src/messages/block_translators/google_vertexai.ts
+++ b/libs/langchain-core/src/messages/block_translators/google_vertexai.ts
@@ -79,16 +79,16 @@ function convertToV1FromChatVertexMessage(
       }
       yield { type: "non_standard", value: block };
     }
+    for (const toolCall of message.tool_calls ?? []) {
+      yield {
+        type: "tool_call",
+        id: toolCall.id,
+        name: toolCall.name,
+        args: toolCall.args,
+      };
+    }
   }
-  const toolCallBlocks: ContentBlock.Standard[] = (
-    message.tool_calls ?? []
-  ).map((toolCall) => ({
-    type: "tool_call",
-    id: toolCall.id,
-    name: toolCall.name,
-    args: toolCall.args,
-  }));
-  return [...iterateContent(), ...toolCallBlocks];
+  return Array.from(iterateContent());
 }
 
 export const ChatVertexTranslator: StandardContentBlockTranslator = {
--- a/libs/langchain-core/src/messages/block_translators/google_vertexai.ts
+++ b/libs/langchain-core/src/messages/block_translators/google_vertexai.ts
@@ -49,6 +49,9 @@ function convertToV1FromChatVertexMessage(
         };
         continue;
       } else if (_isContentBlock(block, "text") && _isString(block.text)) {
+        if (block.text === "") {
+          continue;
+        }
         yield { type: "text", text: block.text };
         continue;
       } else if (_isContentBlock(block, "image_url")) {
@@ -77,7 +80,15 @@ function convertToV1FromChatVertexMessage(
       yield { type: "non_standard", value: block };
     }
   }
-  return Array.from(iterateContent());
+  const toolCallBlocks: ContentBlock.Standard[] = (
+    message.tool_calls ?? []
+  ).map((toolCall) => ({
+    type: "tool_call",
+    id: toolCall.id,
+    name: toolCall.name,
+    args: toolCall.args,
+  }));
+  return [...iterateContent(), ...toolCallBlocks];
 }
 
 export const ChatVertexTranslator: StandardContentBlockTranslator = {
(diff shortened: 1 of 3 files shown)


libs/langchain-core/src/messages/block_translators/google_vertexai.ts
```diff
@@ -77,7 +80,15 @@ function convertToV1FromChatVertexMessage(
...
-  return Array.from(iterateContent());
+  const toolCallBlocks: ContentBlock.Standard[] = (
+    message.tool_calls ?? []
+  ).map((toolCall) => ({
+    type: "tool_call",
+    id: toolCall.id,
+    name: toolCall.name,
+    args: toolCall.args,
+  }));
+  return [...iterateContent(), ...toolCallBlocks];
```
hntrl: nit: could we have this live inside of the iterator? e.g. just at the very end we iterate and loop over message.tool_calls

hntrl (approved):

PR #11407, opened 2026-08-19
fix(anthropic): round-trip tool search server-tool result blocks

chenzimin:
The tool search tool (`tool_search_tool_bm25_20251119` / `tool_search_tool_regex_20251119`) is a server-side tool. Anthropic runs the search and returns a `server_tool_use` block **and** its paired `tool_search_tool_*_tool_result` block in the same assistant message.

`@langchain/anthropic` already supports the *send* side (tool definitions via `tools.toolSearchBM25_20251119()` / `toolSearchRegex_20251119()`, `defer_loading`, and auto-injecting the beta header), but it does not carry the **result** block through a conversation:
- `_makeMessageChunkFromAnthropicEvent` (streaming) whitelists `content_block_start` types and omits the tool-search result block, so it is dropped from the accumulated `AIMessage`.
- `_convertMessagesToAnthropicPayload` (`_formatContentBlocks`) whitelists `toolTypes` and omits it too, so even a manually-constructed message loses it on re-send.

As a result, any multi-turn (or multi-step tool-loop) conversation that uses tool search fails on the next request with:
```
  400 invalid_request_error: messages.N: tool_search_tool_bm25 tool use with id srvtoolu_...
  was found without a corresponding tool_search_tool_bm25_tool_result block
```

This mirrors the existing handling for web_search_tool_result

### DIFF
--- a/.changeset/tool-search-result-round-trip.md
+++ b/.changeset/tool-search-result-round-trip.md
@@ -1,0 +1,5 @@
+---
+"@langchain/anthropic": patch
+---
+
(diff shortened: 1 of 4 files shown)


christian-bromann (approved):
LGTM 👍

PR #11419, opened 2026-08-20
fix(openai): send content null (not []) for tool-call-only v1 assistant messages

### LINKED ISSUE #11416: @langchain/openai: Chat Completions converter emits empty content array for v1 tool-call-only AIMessage
### Checked other resources

### Example Code

import {
  convertMessagesToCompletionsMessageParams,
} from "@langchain/openai";

  content: [
    {
      type: "tool_call",
      id: "call_123",
      name: "get_current_date_time",
      args: {},
    },
  ],
  tool_calls: [
    {
      type: "tool_call",
      id: "call_123",
      name: "get_current_date_time",
      args: {},
    },
  ],
  response_metadata: {
    output_version: "v1",
  },
});

const result = convertMessagesToCompletionsMessageParams({
  messages: [message],
});

console.dir(result, { depth: null });

400 Invalid 'messages[4].content': empty array. Expected an array with minimum length 1, but got an empty array instead.

Error: 400 Invalid 'messages[4].content': empty array. Expected an array with minimum length 1, but got an empty array instead.
    at APIError.generate (.../node_modules/openai/src/core/error.ts:73:14)
    at OpenAI.makeStatusError (.../node_modules/openai/src/client.ts:636:28)
    at OpenAI.makeRequest (.../node_modules/openai/src/client.ts:932:24)
    at async <anonymous> (.../@langchain/openai/src/chat_models/completions.ts:540:18)

### Description

When using OpenAI Chat Completions with a v1 `AIMessage` that contains only a
`tool_call` content block, `convertMessagesToCompletionsMessageParams()` emits
an assistant message with an empty `content` array.

The original LangChain `AIMessage` contains a valid tool call, but after
conversion it becomes:

```typescript
{
  role: "assistant",
  content: [],
  tool_calls: [
    {
      id: "call_123",
      type: "function",
      function: {
        name: "get_current_date_time",
        arguments: "{}"
      }
    }
  ]
}
```

OpenAI Chat Completions rejects this payload with:

```text
400 Invalid 'messages[4].content': empty array.
Expected an array with minimum length 1, but got an empty array instead.
```

This breaks normal multi-step tool-calling flows:

```text
AIMessage(tool call)
-> ToolMessage
-> next model invocation
-> HTTP 400
```

I initially encountered this in a LangGraph application, but confirmed that
the issue is not caused by LangGraph or application-specific code.

I reproduced the issue in a standalone project by directly calling
`convertMessagesToCompletionsMessageParams()`.

The issue is reproducible with:

- `@langchain/core`: 1.2.9
- `@langchain/openai`: 1.5.9

The standalone reproduction returns:

```typescript
[
  {
    role: "assistant",
    content: [],
    tool_calls: [
      {
        id: "call_123",
        type: "function",
        function: {
          name: "get_current_date_time",
          arguments: "{}"
        }
      }
    ]
  }
]
```

### Expected behavior

A tool-call-only assistant message should not be serialized with
`content: []`.

For example:

```typescript
{
  role: "assistant",
  content: "",
  tool_calls: [...]
}
```

As a temporary workaround, I normalize tool-call-only `AIMessage`s before the
next model invocation so that they are serialized with `content: ""` while
preserving the `tool_calls` / `tool_call_id` relationship.

### System Info

Platform: Windows
Node.js: 24.12.0
pnpm: 10.28.0

@langchain/core: 1.2.9
@langchain/openai: 1.5.9

OpenAI SDK: 6.48.0
OpenAI model: gpt-5-nano
API: OpenAI Chat Completions

chiliec:
### Problem

`convertMessagesToCompletionsMessageParams` (`@langchain/openai`) builds an assistant message's `content` by filtering the message's content blocks down to text blocks:

```ts
content: message.contentBlocks.filter((block) => block.type === "text"),
```

For a v1-output `AIMessage` whose content is **only tool-call blocks** (a tool-call-only turn), that filter yields an empty array, so the outgoing request carries `content: []`. The OpenAI Chat Completions API rejects it:

```
400 Invalid 'messages[..].content': empty array.
Expected an array with minimum length 1, but got an empty array instead.
```

Reproduction (from the issue): an `AIMessage` with `output_version: "v1"` and content consisting solely of a `tool_call` block round-trips through the converter to `content: []` and then fails the API call.

### Fix

After assembling the assistant param, if `content` resolved to an empty array **and** `tool_calls` are present, send `content: null` instead. `null` is a valid value for `ChatCompletionAssistantMessageParam.content`, and the `tool_calls` field carries the actual payload — which is exactly the shape the API expects for a tool-call-only assistant turn.

The change is confined to the `assistant` branch; text-bearing assistant messages, string-content messages, and non-assistant roles are untouched.

### Tests

- Updated the existing test `should preserve tool_calls for output_version v1 assistant messages` — it previously asserted the **buggy** `content: []` output. It now asserts `content: null` (and is renamed to describe the corrected behavior).
- Verified RED/GREEN locally: with the fix removed the test fails with `content: []`; with the fix it passes.
- Full converter test suite green: **112 passed**, no type errors (`vitest run src/converters/`).

### Verification commands

```
# built @langchain/core, then:
cd libs/providers/langchain-openai
npx vitest run src/converters/    # 112 passed, Type Errors: no errors
```

### DIFF
--- a/.changeset/four-actors-move.md
+++ b/.changeset/four-actors-move.md
@@ -1,0 +1,5 @@
+---
+"@langchain/openai": patch
+---
+
+fix(openai): send content null (not []) for tool-call-only v1 assistant messages
--- a/libs/providers/langchain-openai/src/converters/completions.ts
+++ b/libs/providers/langchain-openai/src/converters/completions.ts
@@ -658,10 +658,13 @@ export const convertStandardContentMessageToCompletionsMessage: Converter<
       content: message.contentBlocks.filter((block) => block.type === "text"),
     };
   } else if (role === "assistant") {
+    const textContent = message.contentBlocks.filter(
+      (block) => block.type === "text"
+    );
     const completionParam: OpenAIClient.Chat.Completions.ChatCompletionAssistantMessageParam =
       {
         role: "assistant",
-        content: message.contentBlocks.filter((block) => block.type === "text"),
+        content: textContent,
       };
     if (AIMessage.isInstance(message) && !!message.tool_calls?.length) {
       completionParam.tool_calls = message.tool_calls.map(
@@ -670,6 +673,18 @@ export const convertStandardContentMessageToCompletionsMessage: Converter<
     } else if (message.additional_kwargs.tool_calls != null) {
       completionParam.tool_calls = message.additional_kwargs
         .tool_calls as OpenAIClient.Chat.Completions.ChatCompletionMessageToolCall[];
+    }
+    // The OpenAI Chat Completions API rejects an assistant message whose
+    // `content` is an empty array ("empty array. Expected an array with minimum
+    // length 1"). A tool-call-only AIMessage (all content blocks are tool_call
+    // blocks, so the text filter yields []) must therefore send `content: null`
+    // instead of `[]`; tool_calls carry the payload.
+    if (
+      Array.isArray(completionParam.content) &&
+      completionParam.content.length === 0 &&
+      completionParam.tool_calls != null
+    ) {
+      completionParam.content = null;
     }
     return completionParam;
   } else if (role === "tool" && ToolMessage.isInstance(message)) {
--- a/libs/providers/langchain-openai/src/converters/tests/completions.test.ts
+++ b/libs/providers/langchain-openai/src/converters/tests/completions.test.ts
@@ -349,7 +349,7 @@ describe("convertCompletionsMessageToBaseMessage", () => {
       });
     });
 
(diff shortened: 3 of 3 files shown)


christian-bromann (dismissed):
LGTM 👍

chiliec:
Thanks for the quick merge, @christian-bromann! Happy to keep contributing on the JS SDK / provider integrations — I work across TS/Go with an LLM-tooling and crypto/mobile background, and I'm currently open to remote contract work if an extra pair of hands would be useful. Either way I'll keep sending fixes as I hit them.

PR #11420, opened 2026-08-20
fix(openai): serialize empty v1 assistant content as string

### LINKED ISSUE #11416: @langchain/openai: Chat Completions converter emits empty content array for v1 tool-call-only AIMessage

jackjin1997:
Fixes #11416

V1 assistant messages with no text content now use an empty string instead of an invalid empty array when sent to Chat Completions. This covers both tool-call-only and reasoning-only history.

---

The standard-content converter filters assistant content down to text blocks. When none remain, an empty string keeps the payload valid while preserving any tool calls.

## Release note

`@langchain/openai` now serializes v1 assistant messages without text content as an empty string for Chat Completions.

## AI disclosure

This contribution was implemented and validated with AI assistance.

### DIFF
--- a/.changeset/cuddly-tools-smile.md
+++ b/.changeset/cuddly-tools-smile.md
@@ -2,4 +2,4 @@
 "@langchain/openai": patch
 ---
 
-Fix Chat Completions serialization of v1 tool-call-only assistant messages.
+Fix Chat Completions serialization of v1 assistant messages without text content.
--- a/.changeset/cuddly-tools-smile.md
+++ b/.changeset/cuddly-tools-smile.md
@@ -1,0 +1,5 @@
+---
+"@langchain/openai": patch
+---
+
+Fix Chat Completions serialization of v1 tool-call-only assistant messages.
--- a/libs/providers/langchain-openai/src/converters/completions.ts
+++ b/libs/providers/langchain-openai/src/converters/completions.ts
@@ -671,6 +671,13 @@ export const convertStandardContentMessageToCompletionsMessage: Converter<
       completionParam.tool_calls = message.additional_kwargs
         .tool_calls as OpenAIClient.Chat.Completions.ChatCompletionMessageToolCall[];
     }
+    if (
+      Array.isArray(completionParam.content) &&
+      completionParam.content.length === 0 &&
+      completionParam.tool_calls != null
+    ) {
+      completionParam.content = "";
+    }
     return completionParam;
   } else if (role === "tool" && ToolMessage.isInstance(message)) {
     return {
--- a/libs/providers/langchain-openai/src/converters/completions.ts
+++ b/libs/providers/langchain-openai/src/converters/completions.ts
@@ -673,8 +673,7 @@ export const convertStandardContentMessageToCompletionsMessage: Converter<
     }
     if (
       Array.isArray(completionParam.content) &&
-      completionParam.content.length === 0 &&
-      completionParam.tool_calls != null
+      completionParam.content.length === 0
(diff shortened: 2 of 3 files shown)


jackjin1997:
Closing in favor of #11419, which has now been merged and resolves the reported tool-call-only case. This branch also covered reasoning-only replay; happy to revisit that separately if it remains reproducible after the merged fix.

PR #11421, opened 2026-08-20
fix(anthropic): preserve generic tool_search_tool_result blocks

priprocess:
The API emits the un-suffixed tool_search_tool_result type for both tool search variants (observed in live traffic; matches the response format in Anthropic's tool-search docs). 

#11407 allowlisted only the variant-suffixed names, so streaming still drops the block and the next request 400s with INVALID_TOOL_RESULTS. Adds the generic name to both allowlists, mirrors the existing tests for it.


christian-bromann (approved):
LGTM 👍

Issue #9731, opened 2026-01-03
google-genai: cached token count incorrect when streaming


bradvogel:
### Checked other resources

- [x] This is a bug, not a usage question. For questions, please use the LangChain Forum (https://forum.langchain.com/).
- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain.js documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain.js rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```typescript
import { ChatGoogleGenerativeAI } from '@langchain/google-genai';
import { HumanMessage, SystemMessage } from '@langchain/core/messages';

// Large system prompt to trigger implicit caching (>1024 tokens for Flash)
const SYSTEM_PROMPT = 'You are a helpful assistant. '.repeat(200);

const model = new ChatGoogleGenerativeAI({
  model: 'gemini-2.5-flash',
  apiKey: process.env.GEMINI_API_KEY!,
  streaming: true
});

const messages = [
  new SystemMessage(SYSTEM_PROMPT),
  new HumanMessage('Explain how to solve 2 + 2 step by step.')
];

// Run twice - first warms cache, second shows the bug
for (let run = 1; run <= 2; run++) {
  console.log(`\n--- Run ${run} ---`);
  const stream = await model.stream(messages);
  let chunkNum = 0;

  for await (const chunk of stream) {
    const usage = chunk.usage_metadata as any;
    if (usage?.input_tokens !== undefined) {
      chunkNum++;
      const inputTokens = usage.input_tokens;
      const cacheRead = usage.input_token_details?.cache_read ?? 0;

      console.log(`Chunk ${chunkNum}: input_tokens=${inputTokens}, cache_read=${cacheRead}`);

      // BUG: cache_read is cumulative, input_tokens is delta
      // On run 2, chunk 2+: cache_read > 0 but input_tokens = 0
      if (cacheRead > 0 && inputTokens === 0) {
        console.log('  ^^^ BUG: cache_read > 0 but input_tokens = 0');
      }
    }
  }
  await new Promise(r => setTimeout(r, 2000)); // Wait for cache
}
```

### Error Message and Stack Trace (if applicable)

[...]

Issue #8266, opened 2025-05-28
@langchain/google-genai usage_metadata bug


keisokoo:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain.js documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain.js rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

same input, same output
1.
genai sdk response
  "modelVersion": "models/gemini-2.5-pro-preview-05-06",
  "usageMetadata": {
    "promptTokenCount": 17,
    "candidatesTokenCount": 1698,
    "totalTokenCount": 3100,
    "promptTokensDetails": [
      {
        "modality": "TEXT",
        "tokenCount": 17
      }
    ],
    "thoughtsTokenCount": 1385
  }
2.
langchain/genai response
      "additional_kwargs": {},
      "response_metadata": {
        "index": 0,
        "finishReason": "STOP"
      },
      "tool_calls": [],
      "tool_call_chunks": [],
      "invalid_tool_calls": [],
      "usage_metadata": {
        "input_tokens": 17,
        "output_tokens": 24803,
        "total_tokens": 26172
      }

### Description

Issue: usage_metadata is being double-counted when using ChatGoogleGenerativeAI with streaming

When using ChatGoogleGenerativeAI from LangChain with either streaming: true or the .streamEvents() method, the usage_metadata values (e.g., token usage) are being counted multiple times, leading to inaccurate usage reporting.

This happens because the same metadata appears to be merged or emitted more than once during the streaming lifecycle.

Expected behavior:
Usage metadata should only be recorded once per response, regardless of whether streaming is enabled.

Steps to reproduce:
	1.	Initialize ChatGoogleGenerativeAI with streaming: true or call .streamEvents().
	2.	Inspect the returned usage_metadata.
	3.	Observe that the same token usage appears to be counted more than once.

Let me know if you’d like to add logs or example code to reproduce.

### System Info

macOS
pnpm 10.11.0
node v23.11.0

Godrules500:
I don't have a solution, but are you using streamEvents there when you get a response_metadata in your response? I'm using streamEvents and it is always {}. If you are, do you mind pasting the code you're using that's using streamEvents and getting hte response_metadata back?

bradvogel:
[...]

PR #11423, opened 2026-08-20
fix(google-genai): report Gemini reasoning tokens in usage metadata

### LINKED ISSUE #11422: `@langchain/google-genai`: `thoughtsTokenCount` is dropped, so reasoning tokens are missing and `input_tokens + output_tokens` no longer equals `total_tokens`
## Checked other resources

- [x] The bug is not resolved by updating to the latest stable version (reproduced on `@langchain/google-genai` 2.3.0).

## Example Code

```typescript

  model: "gemini-2.5-flash",
  thinkingConfig: { includeThoughts: true, thinkingBudget: 1024 },
});

const res = await model.invoke(
  "A bat and a ball cost $1.10 in total. The bat costs $1.00 more than " +
    "the ball. How much does the ball cost? Answer with just the amount."
);

console.log(res.usage_metadata);
```

Actual:

```json
{ "input_tokens": 41, "output_tokens": 5, "total_tokens": 389 }
```

Expected (what the API actually reported for the same call):

```json
{
  "input_tokens": 41,
  "output_tokens": 348,
  "total_tokens": 389,
  "output_token_details": { "reasoning": 343 }
}
```

## Description

For Gemini thinking models, `usage_metadata` omits reasoning tokens entirely and
the object is not internally consistent: `input_tokens + output_tokens` is 46
while `total_tokens` is 389. The missing 343 tokens are the thinking tokens,
which are billed but appear in no field of the result.

Any cost tracker reading `output_tokens` sees **5** for a call that produced
**389** tokens.

### Root cause

`convertUsageMetadata` in
[`libs/providers/langchain-google-genai/src/utils/common.ts`](https://github.com/langchain-ai/langchainjs/blob/main/libs/providers/langchain-google-genai/src/utils/common.ts#L977-L1013)
builds the result from four fields and never reads `thoughtsTokenCount`:

```ts
const output: UsageMetadata = {
  input_tokens: usageMetadata?.promptTokenCount ?? 0,
  output_tokens: usageMetadata?.candidatesTokenCount ?? 0,
  total_tokens: usageMetadata?.totalTokenCount ?? 0,
};
```

The field is absent because the legacy `@google/generative-ai` `UsageMetadata`
interface does not declare it — it declares only `promptTokenCount`,
`candidatesTokenCount`, `totalTokenCount` and `cachedContentTokenCount`. The
value does arrive at runtime:

```json
{
  "promptTokenCount": 41,
  "candidatesTokenCount": 5,
  "totalTokenCount": 405,
  "thoughtsTokenCount": 359,
  "serviceTier": "standard"
}
```

Gemini reports thinking tokens separately from `candidatesTokenCount`
(`prompt + candidates + thoughts == total`), while `UsageMetadata.output_tokens`
is documented as "Sum of all output token types" — so reasoning tokens should be
included in `output_tokens` and broken out in `output_token_details.reasoning`.

`@langchain/google-common` already does exactly this in
[`src/utils/gemini.ts`](https://github.com/langchain-ai/langchainjs/blob/main/libs/providers/langchain-google-common/src/utils/gemini.ts#L1243-L1263);
`@langchain/google-genai` has no equivalent (`thoughtsTokenCount` appears in zero
files in the package).

### Note for whoever fixes this

Fixing `convertUsageMetadata` alone is not sufficient and makes streaming worse.
`_streamResponseChunks` converts `input_tokens` / `output_tokens` /
`total_tokens` into per-chunk deltas *after* `convertUsageMetadata` returns, but
leaves nested detail fields cumulative. Since `mergeOutputTokenDetails` in
`@langchain/core` sums `reasoning` across chunks, a cumulative value gets added
up once per chunk.

With a three-chunk stream carrying cumulative thoughts of 100 / 250 / 381, the
merged result reports **731** instead of 381 — the same double-counting class as
 #8266. The reasoning count needs the same delta treatment the other three
counts already get.

(This is also why `input_token_details.cache_read` is wrong while streaming,
reported separately in #9731.)

### Unrelated bug in the same function

`convertUsageMetadata` has an operator-precedence bug in the
`gemini-3-pro-preview` bracket tracking:

```ts
const over200k = Math.max(0, usageMetadata?.promptTokenCount ?? 0 - 200000);
```

`??` binds looser than `-`, so `0 - 200000` is evaluated first and the line reads
`promptTokenCount ?? -200000`. For any real prompt this is
`Math.max(0, promptTokenCount)`, so a 500-token prompt reports
`over_200k: 500`. The next line has the same problem for `cache_read_over_200k`.

I have a patch with unit tests for all of the above and will open a PR shortly.

## System Info

```
@langchain/google-genai 2.3.0
@langchain/core 1.2.9
@google/generative-ai 0.24.1
platform: windows
Node: v24.12.0
```

ainazulfiqar99acc:
Fixes #11422

## The problem

For Gemini thinking models, `usage_metadata` omits reasoning tokens and the
object is not internally consistent. A real `gemini-2.5-flash` call:

```json
{ "input_tokens": 41, "output_tokens": 5, "total_tokens": 389 }
```

`41 + 5` is not `389`. The missing 343 tokens are thinking tokens — billed, but
present in no field of the result. Any cost tracker reading `output_tokens` sees
**5** for a call that produced **389** tokens.

After this change, same call:

```json
{
  "input_tokens": 41,
  "output_tokens": 348,
  "total_tokens": 389,
  "output_token_details": { "reasoning": 343 }
}
```

## Root cause

`convertUsageMetadata` never read `thoughtsTokenCount`. The field is not declared
on the legacy `@google/generative-ai` `UsageMetadata` interface — which declares
only `promptTokenCount`, `candidatesTokenCount`, `totalTokenCount` and
`cachedContentTokenCount` — but it does arrive at runtime, so it is now read
defensively via a small helper.

Gemini reports thinking tokens separately from `candidatesTokenCount`
(`prompt + candidates + thoughts == total`), while `UsageMetadata.output_tokens`
is documented as "Sum of all output token types". So reasoning tokens are counted
toward `output_tokens` and broken out in `output_token_details.reasoning`.

This matches what `@langchain/google-common` already does in
[`src/utils/gemini.ts#L1243-L1263`](https://github.com/langchain-ai/langchainjs/blob/main/libs/providers/langchain-google-common/src/utils/gemini.ts#L1243-L1263),
so the two packages now agree.

## Why the streaming change is also required

Fixing `convertUsageMetadata` alone is not sufficient, and on its own it makes
streaming worse.

`_streamResponseChunks` converts `input_tokens` / `output_tokens` /
`total_tokens` into per-chunk deltas *after* `convertUsageMetadata` returns, but
leaves nested detail fields cumulative. `mergeOutputTokenDetails` in
`@langchain/core` sums `reasoning` across chunks, so a cumulative value is added
once per chunk.

With a three-chunk stream carrying cumulative thoughts of 100 / 250 / 381, the
converter-only change reports **731** instead of 381 — the same double-counting
class as #8266. `chat_models_stream_usage.test.ts` covers this specifically and
fails if only the converter is patched.

(The same structural gap is why `input_token_details.cache_read` is wrong while
streaming, reported in #9731. This PR does not change that path.)

## Unrelated fix in the same function

```ts
const over200k = Math.max(0, usageMetadata?.promptTokenCount ?? 0 - 200000);
```

`??` binds looser than `-`, so `0 - 200000` evaluates first and the expression
reads `promptTokenCount ?? -200000`. For any real prompt this is
`Math.max(0, promptTokenCount)`, so a 500-token prompt reported
`over_200k: 500`. Same problem on the next line for `cache_read_over_200k`.
Both are now parenthesized.

Happy to split this into its own PR if you'd prefer it separate.

## Tests

10 new unit tests, no network required:

- `common.test.ts` — reasoning extraction, `output_tokens` inclusion, the
  `input + output == total` invariant, a no-thinking regression guard, and both
  200k bracket cases.
- `chat_models_stream_usage.test.ts` — a stubbed stream carrying cumulative usage,
  asserting the merged result reports the final totals rather than their sum.

Full package suite: **103 passing**, no type errors. `oxlint` and
`oxfmt --check` clean.

Verified end to end against the live Gemini API on `gemini-2.5-flash`, with the
raw SDK's `thoughtsTokenCount` as the reference value.

### DIFF
--- a/.changeset/olive-moons-reason.md
+++ b/.changeset/olive-moons-reason.md
@@ -1,0 +1,16 @@
+---
+"@langchain/google-genai": patch
+---
+
+Report Gemini reasoning (thinking) tokens in `usage_metadata`
+
+`convertUsageMetadata` never read `thoughtsTokenCount`, so reasoning tokens were
+missing from `output_token_details.reasoning` and excluded from `output_tokens`.
+Because Gemini reports thinking tokens separately from `candidatesTokenCount`,
+this left `input_tokens + output_tokens` short of `total_tokens` on any thinking
+model. Streaming now converts the reasoning count to a per-chunk delta so
+concatenated chunks do not sum the cumulative totals.
+
+Also fixes an operator precedence bug in the `gemini-3-pro-preview` bracket
+tracking, where `?? 0 - 200000` parsed as `?? -200000` and reported the entire
+prompt token count as the amount over 200k.
--- a/libs/providers/langchain-google-genai/src/chat_models.ts
+++ b/libs/providers/langchain-google-genai/src/chat_models.ts
@@ -1009,7 +1009,8 @@ export class ChatGoogleGenerativeAI
     let usageMetadata: UsageMetadata | undefined;
     // Keep prior cumulative counts for calculating token deltas while streaming
     let prevPromptTokenCount = 0;
-    let prevCandidatesTokenCount = 0;
+    let prevOutputTokenCount = 0;
+    let prevThoughtsTokenCount = 0;
     let prevTotalTokenCount = 0;
     let index = 0;
     for await (const response of stream) {
@@ -1037,13 +1038,28 @@ export class ChatGoogleGenerativeAI
         );
         prevPromptTokenCount = newPromptTokenCount;
 
-        const newCandidatesTokenCount =
-          response.usageMetadata.candidatesTokenCount ?? 0;
+        // `output_tokens` covers every output token type, so reasoning tokens
+        // count toward it alongside `candidatesTokenCount`. Both are cumulative
+        // on the wire and have to be turned into per-chunk deltas, otherwise
+        // concatenating the chunks sums the running totals.
+        const newThoughtsTokenCount =
+          usageMetadata.output_token_details?.reasoning ?? 0;
+        const newOutputTokenCount =
+          (response.usageMetadata.candidatesTokenCount ?? 0) +
+          newThoughtsTokenCount;
         usageMetadata.output_tokens = Math.max(
           0,
-          newCandidatesTokenCount - prevCandidatesTokenCount
+          newOutputTokenCount - prevOutputTokenCount
         );
-        prevCandidatesTokenCount = newCandidatesTokenCount;
+        prevOutputTokenCount = newOutputTokenCount;
+
+        if (usageMetadata.output_token_details?.reasoning !== undefined) {
+          usageMetadata.output_token_details.reasoning = Math.max(
+            0,
+            newThoughtsTokenCount - prevThoughtsTokenCount
+          );
+        }
+        prevThoughtsTokenCount = newThoughtsTokenCount;
 
         const newTotalTokenCount = response.usageMetadata.totalTokenCount ?? 0;
         usageMetadata.total_tokens = Math.max(
--- a/libs/providers/langchain-google-genai/src/tests/chat_models_stream_usage.test.ts
+++ b/libs/providers/langchain-google-genai/src/tests/chat_models_stream_usage.test.ts
@@ -1,0 +1,106 @@
+import { describe, expect, test, vi, afterEach } from "vitest";
+import type { GenerateContentRequest } from "@google/generative-ai";
+import { AIMessageChunk } from "@langchain/core/messages";
+import { ChatGoogleGenerativeAI } from "../chat_models.js";
+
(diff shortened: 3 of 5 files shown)


thushanth-bengre-langchain:
Thank you for the PR @ainazulfiqar99acc .
However, we are moving in a direction of using `@langchain/google` and after verifying, we are not able to reproduce this issue in that package.

Hence closing this.

PR #11425, opened 2026-08-20
fix(core,google-genai): report correct token usage in streaming llmOutput

### LINKED ISSUE #11424: `llmOutput.tokenUsage` is wrong on both streaming paths, so callback-based token tracking under-reports Gemini usage by orders of magnitude
## Checked other resources

- [x] The bug is not resolved by updating to the latest stable version.

## Example Code

```typescript
import type { LLMResult } from "@langchain/core/outputs";

const handler = {
  handleLLMEnd(output: LLMResult) {
    console.log("llmOutput          :", JSON.stringify(output.llmOutput));
    const msg = (output.generations[0][0] as any).message;
    console.log("message.usage_metadata:", JSON.stringify(msg.usage_metadata));
  },
};

const opts = {
  model: "gemini-2.5-flash",
  thinkingConfig: { includeThoughts: true, thinkingBudget: 1024 },
  callbacks: [handler],
};
const PROMPT = "A bat and a ball cost $1.10 in total. The bat costs $1.00 more than the ball. How much does the ball cost?";

await new ChatGoogleGenerativeAI(opts).invoke(PROMPT);
await new ChatGoogleGenerativeAI({ ...opts, streaming: true }).invoke(PROMPT);
for await (const _ of await new ChatGoogleGenerativeAI(opts).stream(PROMPT));
```

Output:

```
[invoke, streaming:false]
llmOutput             : {"tokenUsage":{"promptTokens":41,"completionTokens":366,"totalTokens":407}}
message.usage_metadata: {"input_tokens":41,"output_tokens":366,"total_tokens":407,...}

[invoke, streaming:true]
llmOutput             : {"estimatedTokenUsage":{}}
message.usage_metadata: {"input_tokens":41,"output_tokens":360,"total_tokens":401,...}

[stream()]
llmOutput             : {"tokenUsage":{"promptTokens":0,"completionTokens":3,"totalTokens":3}}
message.usage_metadata: {"input_tokens":41,"output_tokens":417,"total_tokens":458,...}
```

## Description

`message.usage_metadata` is correct in all three cases. `llmOutput` is correct in
only one of them.

| call path | `llmOutput` | true usage |
| --- | --- | --- |
| `invoke()`, streaming off | `tokenUsage: { 41, 366, 407 }` | 407 |
| `invoke()`, `streaming: true` | `estimatedTokenUsage: {}` | 401 |
| `.stream()` | `tokenUsage: { 0, 3, 3 }` | 458 |

This matters because `llmOutput` is what callback-based accounting reads.
Sentry's JavaScript LangChain integration, for example, reads token counts
**only** from `llmOutput.tokenUsage` (or `llmOutput.usage` for Anthropic) with no
`usage_metadata` fallback — so a streamed Gemini call is recorded as either no
tokens at all or 3 tokens instead of 458.

There are two independent causes.

### 1. `.stream()` — core keeps the last chunk instead of the accumulation

In `libs/langchain-core/src/language_models/chat_models.ts`, `_streamIterator`
**assigns** `llmOutput` on every chunk:

```ts
if (isAIMessageChunk(chunk.message) && chunk.message.usage_metadata !== undefined) {
  llmOutput = {
    tokenUsage: {
      promptTokens: chunk.message.usage_metadata.input_tokens,
      completionTokens: chunk.message.usage_metadata.output_tokens,
      totalTokens: chunk.message.usage_metadata.total_tokens,
    },
  };
}
```

So the value that survives is the **final chunk's** usage. That is correct for
providers that report cumulative totals on a final usage chunk (the OpenAI
convention), but wrong for any provider that emits **per-chunk deltas** — the
surviving value is the last delta. `@langchain/google-genai` emits deltas (see
`_streamResponseChunks`), which is why the totals come out as `0 / 3 / 3`.

The loop already maintains `generationChunk`, the concatenated result whose
`usage_metadata` is correctly summed. Building `llmOutput` from
`generationChunk.message.usage_metadata` after the loop instead of from each
chunk inside it would be correct for both conventions.

### 2. `invoke()` with `streaming: true` — the object is never populated

In `libs/providers/langchain-google-genai/src/chat_models.ts`, the streaming
branch of `_generate` declares an empty object, never writes to it, and returns
it under a key nothing reads:

```ts
const tokenUsage: TokenUsage = {};
const stream = this._streamResponseChunks(messages, options, runManager);
// ... loop concatenates chunks; tokenUsage is never touched ...
return { generations, llmOutput: { estimatedTokenUsage: tokenUsage } };
```

The non-streaming path of the same class returns
`llmOutput: { tokenUsage: {...} }` via `mapGenerateContentResultToChatResult`, so
the two paths disagree on both the key and the content. The concatenated
`generations` already carry correct `usage_metadata` and could populate it.

## System Info

```
@langchain/google-genai 2.3.0
@langchain/core 1.2.9
platform: windows
Node: v24.12.0
```

Related: #11422 (reasoning tokens dropped from `usage_metadata`) — separate
defect, same area. The numbers above were measured with #11423 applied, so the
`usage_metadata` column is already correct; these `llmOutput` problems are
independent of it.

ainazulfiqar99acc:
Fixes #11424

## The problem

`message.usage_metadata` is correct on every call path. `llmOutput` is not — and
`llmOutput` is what callback-based token accounting reads.

Measured on `gemini-2.5-flash`:

| call path | `llmOutput` | true usage |
| --- | --- | --- |
| `invoke()`, streaming off | `tokenUsage: { 41, 366, 407 }` | 407 |
| `invoke()`, `streaming: true` | `estimatedTokenUsage: {}` | 401 |
| `.stream()` | `tokenUsage: { 0, 3, 3 }` | 458 |

Three tokens reported for a call that used 458. A missing field reads as broken
and gets noticed; `3` renders fine on a dashboard and does not.

Sentry's JavaScript LangChain integration, for example, reads token counts only
from `llmOutput.tokenUsage` with no `usage_metadata` fallback, so streamed
Gemini calls are recorded as either nothing or a small fraction of actual usage.

### `langchain-core` — the last chunk wins instead of the accumulation

`_streamIterator` **assigned** `llmOutput` inside the chunk loop, so the value
that survived was whatever the final chunk carried. That is correct for
providers reporting cumulative totals on a final usage chunk (the OpenAI
convention) and wrong for providers emitting per-chunk deltas, where the
survivor is the last delta.

The loop already maintains `generationChunk`, whose `usage_metadata` is summed
correctly for either convention, so the counts are now derived from it after the
loop. Providers on the cumulative convention are unaffected — there is a test
covering that case explicitly.

### `@langchain/google-genai` — an object that was never filled in

```ts
const tokenUsage: TokenUsage = {};
const stream = this._streamResponseChunks(messages, options, runManager);
// ... loop concatenates chunks; tokenUsage is never touched ...
return { generations, llmOutput: { estimatedTokenUsage: tokenUsage } };
```

Declared empty, never written to, and returned under a key the non-streaming
path of the same class does not use — `mapGenerateContentResultToChatResult`
returns `tokenUsage`. The concatenated `generations` already carry correct
usage metadata, so they now populate it under the matching key.

`estimatedTokenUsage` is a real convention elsewhere (`@langchain/cohere` uses it
for client-side estimates and reads it back in `_combineLLMOutput`), but these
are exact counts from the API, and nothing in this package ever read the value —
`_combineLLMOutput` here returns `[]`.

## Tests

6 new unit tests, no network required.

`langchain-core`:
- delta-emitting provider — merged counts, not the final delta
- cumulative-final provider — unchanged behaviour, guarding the OpenAI convention
- `llmOutput` agrees with the concatenated `usage_metadata`

`@langchain/google-genai`:
- streaming `_generate` reports usage under `tokenUsage`
- the values agree with the message usage metadata
- the usage object is not empty

Suites: `@langchain/core` **1515 passing**, `@langchain/google-genai`
**96 passing**, no type errors. `oxlint` and `oxfmt --check` clean.

## Note

Separate from #11422 / #11423, which fix reasoning tokens going missing from
`usage_metadata`. The measurements above were taken with that patch applied, so
the `usage_metadata` values are already correct there; these `llmOutput`
problems are independent and this PR branches from `main`.

Happy to split the core and provider changes into separate PRs if you would
rather review them apart.

### DIFF
--- a/.changeset/brave-otters-listen.md
+++ b/.changeset/brave-otters-listen.md
@@ -1,0 +1,22 @@
+---
+"@langchain/core": patch
+"@langchain/google-genai": patch
+---
+
+Fix `llmOutput` token usage on the streaming paths
+
+`_streamIterator` reassigned `llmOutput` on every chunk, so the value that
+survived was the final chunk's usage. That is correct for providers that report
+cumulative totals on a last usage chunk, but for providers emitting per-chunk
+deltas it truncated the result to the final delta — a streamed Gemini call
+reported 3 total tokens instead of 458. It now derives the counts from the
+concatenated chunk, which is already summed correctly for both conventions.
+
+`ChatGoogleGenerativeAI._generate` also returned
+`llmOutput: { estimatedTokenUsage: {} }` on its streaming branch: an object that
+was declared, never populated, and keyed differently from the non-streaming path
+of the same class. It now reports the real counts under `tokenUsage`, matching
+`mapGenerateContentResultToChatResult`.
+
+Callback-based token tracking reads `llmOutput`, so both paths previously
+under-reported usage while `message.usage_metadata` was correct.
--- a/libs/langchain-core/src/language_models/chat_models.ts
+++ b/libs/langchain-core/src/language_models/chat_models.ts
@@ -818,18 +818,6 @@ export abstract class BaseChatModel<
           } else {
             aggregated = concat(aggregated, chunk);
           }
-          if (
-            isAIMessageChunk(chunk.message) &&
-            chunk.message.usage_metadata !== undefined
-          ) {
-            llmOutput = {
-              tokenUsage: {
-                promptTokens: chunk.message.usage_metadata.input_tokens,
-                completionTokens: chunk.message.usage_metadata.output_tokens,
-                totalTokens: chunk.message.usage_metadata.total_tokens,
-              },
-            };
-          }
         }
         // Check if stream ended due to abort (provider returned early)
         if (parsedOptions.signal?.aborted) {
@@ -843,6 +831,21 @@ export abstract class BaseChatModel<
         }
         if (aggregated === undefined) {
           throw new Error("Received empty response from chat model call.");
+        }
+        // As in `_streamIterator`, derive usage from the aggregated chunk so
+        // providers that emit per-chunk deltas are summed rather than reduced
+        // to their final delta.
+        if (
+          isAIMessageChunk(aggregated.message) &&
+          aggregated.message.usage_metadata !== undefined
+        ) {
+          llmOutput = {
+            tokenUsage: {
+              promptTokens: aggregated.message.usage_metadata.input_tokens,
+              completionTokens: aggregated.message.usage_metadata.output_tokens,
+              totalTokens: aggregated.message.usage_metadata.total_tokens,
+            },
+          };
         }
         if (outputVersion === "v1") {
           aggregated.message = castStandardMessageContent(
--- a/libs/langchain-core/src/language_models/chat_models.ts
+++ b/libs/langchain-core/src/language_models/chat_models.ts
@@ -583,18 +583,25 @@ export abstract class BaseChatModel<
           } else {
             generationChunk = generationChunk.concat(chunk);
           }
-          if (
-            isAIMessageChunk(chunk.message) &&
-            chunk.message.usage_metadata !== undefined
-          ) {
-            llmOutput = {
-              tokenUsage: {
-                promptTokens: chunk.message.usage_metadata.input_tokens,
-                completionTokens: chunk.message.usage_metadata.output_tokens,
-                totalTokens: chunk.message.usage_metadata.total_tokens,
-              },
-            };
-          }
+        }
+        // Derive token usage from the concatenated chunk rather than from the
+        // last chunk seen. Providers differ: some report cumulative totals on a
+        // final usage chunk, others emit per-chunk deltas. Reading the last
+        // chunk is only correct for the former, and truncates the latter to its
+        // final delta. `generationChunk` has already summed either shape.
+        if (
+          generationChunk !== undefined &&
+          isAIMessageChunk(generationChunk.message) &&
+          generationChunk.message.usage_metadata !== undefined
+        ) {
+          llmOutput = {
+            tokenUsage: {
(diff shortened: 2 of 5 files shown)


jackjin1997 (commented):
Thanks for tackling this. There is a second copy of the last-chunk `llmOutput` logic in `BaseChatModel._generateUncached`'s implicit-streaming branch (`hasStreamingHandler && ...`, around lines 790–845). That branch still assigns `llmOutput` inside the chunk loop, so `invoke()` with a handler whose `lc_prefer_streaming` is true remains under-reported for delta-emitting providers.

I reproduced it locally with two chunks carrying `{4, 6, 10}` and `{0, 3, 3}`: `handleLLMEnd` receives `{0, 3, 3}` instead of the aggregated `{4, 9, 13}`. Could this PR derive `llmOutput` from `aggregated.message.usage_metadata` after the loop there as well? The added regression case and the full core suite pass with that change.

ainazulfiqar99acc:
Yep, missed that one. Fixed in 998d862.

Same in-loop assignment in the implicit-streaming branch. Got `{0, 3, 3}` instead of `{4, 9, 13}` reproducing your case.

Now pulls from `aggregated.message.usage_metadata` after the loop like `_streamIterator` does, plus a regression test using an `lc_prefer_streaming` handler. Core suite green.

thushanth-bengre-langchain:
Thank you for the PR @ainazulfiqar99acc 

We have verified that this issue is not reproducible in @langchain/google.
Our direction going forward will be to use `@langchain/google`.

PR #11426, opened 2026-08-20
fix(google): stop losing/mangling ToolMessage content in Gemini function responses

### LINKED ISSUE #10297: Images in ToolMessage content lost when converted to Gemini functionResponse

### LINKED ISSUE #10439: bug(@langchain/google) ToolMessage serializes content into JSON string, even though Gemini API accepts structured objects as function results

thushanth-bengre-langchain:
Fixes #10297
Fixes #10439

ToolMessage content sent to Gemini as a `functionResponse` had two related bugs: media content (images/audio/video/files) was getting JSON-stringified into `response.result`, where Gemini can't interpret it; and non-string JSON content (objects/arrays) was double-stringified instead of being passed through as the JSON object Gemini's API actually accepts.

* Keep media content blocks out of `functionResponse.response.result`
* Preserve them as sibling `inlineData`/`fileData` Gemini parts instead of discarding them
* Pass non-string ToolMessage content straight through as the JSON object/array Gemini expects, instead of double-stringifying it
* Applies to both the legacy (v0) and v1 content conversion paths

### DIFF
--- a/.changeset/fix-tool-message-media-lost-in-function-response.md
+++ b/.changeset/fix-tool-message-media-lost-in-function-response.md
@@ -1,0 +1,5 @@
+---
+"@langchain/google": patch
+---
+
+fix: keep image/audio/video content in ToolMessages as sibling Gemini parts instead of losing it inside functionResponse.response.result JSON (#10297)
--- a/.changeset/fix-tool-message-media-lost-in-function-response.md
+++ b/.changeset/fix-tool-message-media-lost-in-function-response.md
@@ -2,4 +2,4 @@
 "@langchain/google": patch
 ---
 
-fix: keep image/audio/video content in ToolMessages as sibling Gemini parts instead of losing it inside functionResponse.response.result JSON (#10297)
+fix: keep image/audio/video content in ToolMessages as sibling Gemini parts instead of losing it inside functionResponse.response.result JSON, and stop double-stringifying non-string tool results (#10297, #10439)
--- a/libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
+++ b/libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
@@ -753,6 +753,138 @@ describe.each(coreModelInfo)(
       }
     });
 
+    test("function reply with image content is sent as a sibling part, not lost in JSON (#10297)", async () => {
+      const dataPath = "src/chat_models/tests/data/blue-square.png";
+      const dataType = "image/png";
+      const data = await fs.readFile(dataPath);
+      const data64 = data.toString("base64");
+      const dataUri = `data:${dataType};base64,${data64}`;
+
+      const tools: Gemini.Tool[] = [
+        {
+          functionDeclarations: [
+            {
+              name: "get_screenshot",
+              description: "Get a screenshot of the current screen.",
+              parameters: { type: "object", properties: {} },
+            },
+          ],
+        },
+      ];
+      const llm = newChatGoogle().bindTools(tools);
+      const messages: BaseMessage[] = [
+        new HumanMessage(
+          "Take a screenshot of the current screen and tell me what color it is."
+        ),
+        new AIMessage({
+          tool_calls: [
+            {
+              type: "tool_call",
+              id: "lc-tool-call-screenshot-id",
+              name: "get_screenshot",
+              args: {},
+            },
+          ],
+        }),
+        new ToolMessage({
+          content: [
+            { type: "text", text: "Screenshot captured successfully." },
+            { type: "image_url", image_url: dataUri },
+          ],
+          tool_call_id: "lc-tool-call-screenshot-id",
+        }),
+      ];
+
+      const res = await llm.invoke(messages);
+      expect(res).toBeDefined();
+
+      const sentContents = recorder.request?.body?.contents as
+        | Gemini.Content[]
+        | undefined;
+      expect(sentContents).toBeDefined();
+      const toolResponseContent = sentContents!.find((c) =>
+        c.parts?.some((p) => "functionResponse" in p)
+      );
+      expect(toolResponseContent).toBeDefined();
+
+      const functionResponsePart = toolResponseContent!.parts!.find(
+        (p) => "functionResponse" in p && p.functionResponse
+      ) as Gemini.Part.FunctionResponse;
+      const resultStr = JSON.stringify(
+        functionResponsePart.functionResponse!.response
+      );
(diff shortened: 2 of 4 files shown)


afirstenberg:
Do we have integration tests illustrating the problem and the fix?

thushanth-bengre-langchain:
> Do we have integration tests illustrating the problem and the fix?

@afirstenberg Added int test for both issues

hntrl (approved):

PR #11430, opened 2026-08-20
fix(mcp-adapters): drop regex patterns that fail to compile under the u flag

### LINKED ISSUE #11351: mcp-adapters: MCP tools with unicode-invalid JSON-schema regex patterns fail every invocation (SyntaxError from @cfworker/json-schema's /u compile)
### Checked other resources

- This is a bug in `@langchain/mcp-adapters` / the `@langchain/core` JSON-schema validation path, not a usage question.
- I searched existing issues for "Invalid regular expression" / pattern-unicode reports and found none.

### Package versions

`@langchain/mcp-adapters` 1.1.3, `@langchain/core` 1.1.47, `@cfworker/json-schema` 4.1.1 (core's validator).

### Problem

Real-world MCP servers ship tool input schemas whose `pattern` is a legal plain ECMAScript regex but **invalid under the `u` flag** — PayPal's production MCP server (`mcp.paypal.com`) is a live example, with patterns shaped like `"^https\\:\\/\\/"` (`\:` is an Annex-B identity escape, rejected under `/u`).

`StructuredTool` validates input with `@cfworker/json-schema`, which compiles `pattern` (and each `patternProperties` key) via `new RegExp(pattern, 'u')`, uncaught. The tool loads fine — the compile is lazy — and then **every invocation** fails:

```
SyntaxError: Invalid regular expression: /^https\:\/\//u: Invalid escape
```

The tool is permanently uncallable, and the error reads like a bug in the caller's arguments rather than the vendor's schema.

### Reproduction

```js
import { loadMcpTools } from "@langchain/mcp-adapters";

const toxicTool = {
  name: "toxic",
  description: "d",
  inputSchema: {
    type: "object",
    properties: { url: { type: "string", pattern: "^https\\:\\/\\/" } },
  },
};
const fakeClient = {
  listTools: async () => ({ tools: [toxicTool] }),
  callTool: async () => ({ content: [{ type: "text", text: "ok" }] }),
};

const tools = await loadMcpTools("srv", fakeClient, { throwOnLoadError: true });
// loads fine — the failure is per-invocation:
await tools[0].invoke({ url: "https://x.com" });
// SyntaxError: Invalid regular expression: /^https\:\/\//u: Invalid escape
```

### Suggested fix

`loadMcpTools` already normalizes schemas for `@cfworker/json-schema` compatibility (`dereferenceJsonSchema` exists specifically because that validator can't resolve `$ref`; `simplifyJsonSchemaForLLM` loosens composition keywords for LLM tool-calling). A natural companion step would drop (or rewrite) `pattern` / `patternProperties` entries that fail `new RegExp(p, 'u')`, so a vendor's non-unicode-safe pattern degrades to a skipped best-effort check instead of a permanently broken tool.

One subtlety if dropping `patternProperties` entries: `@cfworker` marks pattern-matched keys as evaluated, so under `additionalProperties: false` a naive drop turns the crash into a false rejection of previously-valid keys — the relaxation needs to extend to `additionalProperties` at that level.

Alternatives: `@langchain/core` catching the SyntaxError around `validate()` and skipping the pattern check, or an upstream `@cfworker/json-schema` change to fall back to a non-unicode compile when `/u` fails.

We ship a downstream sanitizer with tests covering the above (including the `patternProperties`/`additionalProperties` interaction) in [stigmer/stigmer#525](https://github.com/stigmer/stigmer/pull/525), happy for any of it to be reused. Context: [stigmer/stigmer#420](https://github.com/stigmer/stigmer/issues/420).

bisma-nawaz:
Fixes #11351

## What

MCP tools whose JSON Schema carries a regex `pattern` (or a `patternProperties` key) that is valid in a plain `RegExp` but invalid under the `u` flag — e.g. an Annex B identity escape like `\:` — fail on **every** invocation with a `SyntaxError`.

## Why

Tool-input validation runs through `@cfworker/json-schema`, which compiles each `pattern` with `new RegExp(pattern, "u")`. `simplifyJsonSchemaForLLM` in `libs/langchain-mcp-adapters/src/tools.ts` strips several unsupported keywords but passes `pattern`/`patternProperties` through untouched, so an unenforceable-under-`u` pattern reaches the validator and throws, leaving the tool permanently unusable.

## How

At the end of `simplifyJsonSchemaForLLM`, drop any `pattern` (and any `patternProperties` key) that fails to compile under the `u` flag, via a small `isUnicodeCompilableRegex` helper. When a `patternProperties` key is removed, `additionalProperties: false` at that level is relaxed so the dropped constraint doesn't turn into a rejection of keys that previously validated. The simplifier already recurses into nested schemas, so nested patterns are covered too.

## Testing

Added `drops regex patterns that fail to compile under the Unicode flag` to `libs/langchain-mcp-adapters/src/tests/tools.test.ts`: it loads a tool whose schema carries a `\:` pattern and invokes it. Before the change the invocation throws `SyntaxError: Invalid regular expression: /^\d+\:\d+$/u: Invalid escape`; after, it succeeds. The full test file passes (16) with no type errors.

before: SyntaxError on invoke / after: 16 passed

### DIFF
--- a/.changeset/mcp-adapters-unicode-pattern.md
+++ b/.changeset/mcp-adapters-unicode-pattern.md
@@ -1,0 +1,5 @@
+---
+"@langchain/mcp-adapters": patch
+---
+
+Drop MCP tool JSON-schema regex patterns that fail to compile under the `u` flag. Such patterns (e.g. Annex B identity escapes like `\:`) are valid in a plain `RegExp` but throw a `SyntaxError` under the `u` flag that `@cfworker/json-schema` uses, which previously made every invocation of the tool fail.
--- a/libs/langchain-mcp-adapters/src/tests/tools.test.ts
+++ b/libs/langchain-mcp-adapters/src/tests/tools.test.ts
@@ -710,6 +710,58 @@ describe("Simplified Tool Adapter Tests", () => {
       expect(result).toBe("Event created");
     });
 
+    test("drops regex patterns that fail to compile under the Unicode flag", async () => {
+      // `\:` is a legal identity escape in a plain RegExp but throws under the
+      // "u" flag that @cfworker/json-schema compiles patterns with, which
+      // previously made every invocation of a tool carrying such a pattern fail
+      // with a SyntaxError. See issue #11351.
+      const schemaWithInvalidUnicodePattern = {
+        type: "object" as const,
+        additionalProperties: false,
+        properties: {
+          ratio: {
+            type: "string",
+            pattern: "^\\d+\\:\\d+$",
+          },
+        },
+        patternProperties: {
+          "^x\\:": { type: "string" },
+        },
+        required: ["ratio"],
+      };
+
+      mockClient.listTools.mockReturnValueOnce(
+        Promise.resolve({
+          tools: [
+            {
+              name: "set_ratio",
+              description: "Set an aspect ratio",
+              inputSchema: schemaWithInvalidUnicodePattern,
+            },
+          ],
+        })
+      );
+
+      mockClient.callTool.mockImplementation(() => {
+        return Promise.resolve({
+          content: [{ type: "text", text: "Ratio set" }],
+        });
+      });
+
+      const tools = await loadMcpTools(
+        "mockServer(invalid unicode pattern)",
+        mockClient as Client
+      );
+
+      expect(tools.length).toBe(1);
+
+      // Before the fix this threw a SyntaxError while compiling the `\:` pattern
+      // under the "u" flag; the unenforceable pattern is now dropped so the call
+      // succeeds.
+      const result = await tools[0].invoke({ ratio: "16:9" });
+      expect(result).toBe("Ratio set");
+    });
+
     test("should simplify schemas with anyOf at top level", async () => {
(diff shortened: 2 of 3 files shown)


Bubblegunn:
Thanks for taking this on, @bisma-nawaz — dropping the unenforceable constraint rather than
letting it brick the tool reads right to me, and relaxing `additionalProperties` so the drop
can't tighten validation is a good catch.

One gap I ran into while testing this at `0e54630`: the scan checks `patternProperties` *keys*,
but the sub-schemas behind the surviving keys are never simplified, so a `/u`-invalid `pattern`
one level in still throws.

```ts
patternProperties: { "^opt_": { type: "string", pattern: "^\\d+\\:\\d+$" } }
```

`invoke({ opt_a: "16:9" })` → `SyntaxError: Invalid regular expression: /^\d+\:\d+$/u: Invalid escape`

Recursing into the surviving values covers it, and matches the docstring's "applied recursively
to nested schemas":

```ts
keptPatternProps[patternKey] = simplifyJsonSchemaForLLM(subSchema as JsonSchemaObject);
```

Full package suite stayed green for me with that (169 passed). Happy to share the test.

PR #11433, opened 2026-08-21
fix(google): route Vertex multi-region endpoints

### LINKED ISSUE #11432: `@langchain/google` uses an invalid endpoint for Vertex multi-regions
### Checked other resources

- [x] I searched the LangChain.js documentation and API reference.
- [x] I searched existing issues and pull requests.

### Example code

```ts
  model: "gemini-2.5-flash",
  platformType: "gcp",
  location: "eu",
});
await model.invoke("What is 1 + 1?");
```

### Error message and stack trace

The request is sent to:

```text
https://eu-aiplatform.googleapis.com/v1/projects/.../locations/eu/publishers/google/models/gemini-2.5-flash:generateContent
```

Vertex AI multi-region locations use:

```text
https://aiplatform.eu.rep.googleapis.com/v1/projects/.../locations/eu/publishers/google/models/gemini-2.5-flash:generateContent
```

### Description

`@langchain/google` handles `global` and standard regional locations, but treats the `eu` and `us` multi-regions as standard regions. The sibling `@langchain/google-common` package already applies the required `aiplatform.<location>.rep.googleapis.com` routing after #11032.

A fix should align `@langchain/google` with that endpoint selection and cover both the generated request URL and the credentialed Vertex integration path.

jackjin1997:
Fixes #11432

---

`@langchain/google` generated standard regional Vertex hostnames for the `eu` and `us` multi-regions. Vertex instead exposes those locations through `aiplatform.<location>.rep.googleapis.com`, as documented by Google Cloud and already handled by the sibling `@langchain/google-common` package in #11032.

This change:

- routes the explicit `eu` and `us` locations through their multi-region endpoints;
- verifies both full request URLs through the public `ChatGoogle.invoke()` path;
- adds a credentialed Node/Vertex integration assertion for the `eu` endpoint;
- includes a patch changeset.

Verification:

- `pnpm --filter @langchain/google exec vitest run src/chat_models/tests/index.test.ts` — 91 passed, type errors 0
- `pnpm exec oxlint libs/providers/langchain-google/src` — passed
- `pnpm --filter @langchain/google build` — passed, including attw and publint
- full package test — 162 passed; the 3 converter failures were reproduced unchanged on a clean `origin/main` worktree (161 passed, same 3 failures)
- credentialed integration test was not run locally because Google credentials are unavailable

## Release note

`@langchain/google` now routes Vertex AI multi-region locations `eu` and `us` through their supported multi-region endpoints.

Prepared with AI assistance and reviewed and tested by the contributor.

### DIFF
--- a/libs/providers/langchain-google/src/chat_models/base.ts
+++ b/libs/providers/langchain-google/src/chat_models/base.ts
@@ -360,7 +360,7 @@ export abstract class BaseChatGoogle<
     } else if (this.location === "global") {
       // See https://docs.cloud.google.com/vertex-ai/generative-ai/docs/learn/locations#use_the_global_endpoint
       return "aiplatform.googleapis.com";
-    } else if (["eu", "us"].includes(this.location)) {
+    } else if (!this.location.includes("-")) {
       return `aiplatform.${this.location}.rep.googleapis.com`;
     } else {
       return `${this.location}-aiplatform.googleapis.com`;
--- a/libs/providers/langchain-google/src/chat_models/base.ts
+++ b/libs/providers/langchain-google/src/chat_models/base.ts
@@ -360,6 +360,8 @@ export abstract class BaseChatGoogle<
     } else if (this.location === "global") {
       // See https://docs.cloud.google.com/vertex-ai/generative-ai/docs/learn/locations#use_the_global_endpoint
       return "aiplatform.googleapis.com";
+    } else if (["eu", "us"].includes(this.location)) {
+      return `aiplatform.${this.location}.rep.googleapis.com`;
     } else {
       return `${this.location}-aiplatform.googleapis.com`;
     }
--- a/.changeset/bright-zebras-listen.md
+++ b/.changeset/bright-zebras-listen.md
@@ -1,0 +1,5 @@
+---
+"@langchain/google": patch
+---
+
+Fix Vertex AI endpoint routing for multi-region locations such as `eu` and `us`.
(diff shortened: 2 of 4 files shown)


libs/providers/langchain-google/src/chat_models/base.ts
```diff
@@ -360,6 +360,8 @@ export abstract class BaseChatGoogle<
     } else if (this.location === "global") {
       // See https://docs.cloud.google.com/vertex-ai/generative-ai/docs/learn/locations#use_the_global_endpoint
       return "aiplatform.googleapis.com";
+    } else if (["eu", "us"].includes(this.location)) {
```
afirstenberg: I'd rather we go with the solution we applied for #11032 which does not lock us to specific names.
(Which I thought we applied already.)

hntrl:
```suggestion
    } else if (!this.location.includes("-")) {
```

jackjin1997: Agreed. hntrl applied the same shape-based condition directly, so the implementation no longer hard-codes `eu` / `us`. I re-ran the focused unit test (91/91), build, lint, and format checks against the updated head; the tests stay on the two documented multi-region endpoints.

hntrl (approved):

jackjin1997:
@hntrl thanks for the approval last week. The branch is still mergeable against main with checks green, and Allen's earlier concern about hard-coded region names was resolved by the shape-based condition you suggested. Is there anything else needed from my side before this can land, or is it waiting on a second sign-off for the Google package?

PR #11445, opened 2026-08-22
fix(google): keep ToolMessage and following HumanMessage in separate Gemini user turns

### LINKED ISSUE #11444: @langchain/google: ToolMessage followed by HumanMessage merges into one Gemini `user` content, and Vertex rejects the request
## Checked other resources

- [x] The title is short and descriptive.
- [x] I searched the LangChain.js documentation.
- [x] I searched existing GitHub issues. The closest are #9624 and langchain-ai/langchain-google#1364, both about `thought_signature`. This is a different problem (no thought signatures are involved in the repro below).
- [x] The example code is mine.
- [x] I reproduced on the latest release, `@langchain/google` 0.2.3, as well as 0.2.1.

---

## Example Code

Running this prints all three lines in one go. The first case fails, the two controls succeed, so the trigger is isolated without editing anything.

```bash
npm i @langchain/google @langchain/core zod tsx
export GOOGLE_CLOUD_PROJECT=your-project
export GOOGLE_APPLICATION_CREDENTIALS=/path/to/sa.json
npx tsx repro.ts
```

```ts
import { ChatGoogle } from '@langchain/google/node'
import { AIMessage, HumanMessage, ToolMessage } from '@langchain/core/messages'
import { tool } from '@langchain/core/tools'
import { z } from 'zod'

const readPages = tool(async () => 'page text', {
  name: 'read_document_pages',
  description: 'read pages of a document',
  schema: z.object({ fileId: z.string() }),
})

  model: 'gemini-3.7-flash',
  vertexai: true,
  apiVersion: 'v1',
  location: 'global',
  temperature: 0,
  maxOutputTokens: 256,
  googleAuthOptions: {
    scopes: ['https://www.googleapis.com/auth/cloud-platform'],
    projectId: process.env.GOOGLE_CLOUD_PROJECT,
  },
}).bindTools([readPages])

const toolCall = { name: 'read_document_pages', args: { fileId: 'f1' }, id: 'call_1' }
const aiTurn = (content: string) => new AIMessage({ content, tool_calls: [toolCall] })
const toolTurn = () =>
  new ToolMessage({ content: 'page text', tool_call_id: 'call_1', name: 'read_document_pages' })

const cases: Record<string, any[]> = {
  'FAILS   ai(text + tool_call), tool, human': [
    new HumanMessage('read it'),
    aiTurn('Reading the agreement now.'),
    toolTurn(),
    new HumanMessage('continue'),
  ],
  'control ai(text + tool_call), tool       ': [
    new HumanMessage('read it'),
    aiTurn('Reading the agreement now.'),
    toolTurn(),
  ],
  'control ai(EMPTY + tool_call), tool, human': [
    new HumanMessage('read it'),
    aiTurn(''),
    toolTurn(),
    new HumanMessage('continue'),
  ],
}

async function main() {
  for (const [label, messages] of Object.entries(cases)) {
    try {
      await model.invoke(messages)
      console.log(`${label} -> 200 OK`)
    } catch (e: any) {
      console.log(`${label} -> ${e?.message}`)
    }
  }
}

void main()
```

Output:

```
FAILS   ai(text + tool_call), tool, human -> Requests ending with a model turn are not supported.
control ai(text + tool_call), tool        -> 200 OK
control ai(EMPTY + tool_call), tool, human -> 200 OK
```

---

## Error Message and Stack Trace

```
RequestError: Requests ending with a model turn are not supported.
    at RequestError.fromResponse (node_modules/@langchain/google/src/utils/errors.ts:543:12)
    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
    at async <anonymous> (node_modules/@langchain/google/src/chat_models/base.ts:606:19)
    at async Object.pRetry (node_modules/@langchain/core/src/utils/p-retry/index.js:246:22)
    at async run (node_modules/p-queue/dist/index.js:163:29)
```

---

## Description

I am calling `ChatGoogle` with a history where an assistant turn carries both text and a tool call, followed by its `ToolMessage`, followed by a `HumanMessage`.

I expect the call to succeed. That sequence is ordinary: it is what a conversation looks like when a user says something after a tool has run.

Instead Vertex returns a 400 saying the request ends with a model turn. It does not. The converted payload ends with a `user` content, which is why the error sends you looking in the wrong place.

The cause appears to be in `convertMessagesToGeminiContents`, which merges adjacent messages of the same role. `ToolMessage` and `HumanMessage` both map to `role: "user"`, so the user's text is folded into the same content as the `functionResponse`:

```json
{
  "role": "user",
  "parts": [
    { "functionResponse": { "name": "read_document_pages", "response": { "result": "page text" } } },
    { "text": "continue" }
  ]
}
```

To check that against the API directly, I posted hand-built payloads to `:generateContent`, bypassing the library:

| `contents` | Result |
| --- | --- |
| `model:[text, functionCall]` then `user:[functionResponse, text]` | 400 |
| `model:[text, functionCall]` then `user:[functionResponse]` then `user:[text]` | 200 OK |

Splitting that single merged content is the only difference between the two.

I saw the same on a real 57 message conversation. It converts to 55 contents, of which exactly one mixes a `functionResponse` with text. As it stands the request returns 400. With that one content split into two `user` contents, the identical history returns 200.

The suggested fix is this: when merging adjacent `user` contents, do not merge one carrying a `functionResponse` with one carrying text. Emit them as separate `user` contents.

On how we hit it in practice, an agent run was killed mid tool by an infrastructure timeout, and the user sent a follow up message. From that point every turn on the thread failed, because the offending pair stays in the history permanently. Our workaround is a `wrapModelCall` middleware that inserts a short assistant turn between a `ToolMessage` and a following `HumanMessage`, which prevents the merge. I am happy to open a PR against the converter if the suggested fix looks right.

One observation I cannot explain, in case it is diagnostic: the third case above passes. Setting the assistant turn's content to an empty string leaves the same merged `user` content in place, and the request succeeds.

---

## System Info

```
langchain@1.5.4
@langchain/core@1.2.9
@langchain/google@0.2.1 (also reproduced on 0.2.3)
@langchain/langgraph@1.4.8

platform: darwin (macOS 26.5.2, arm64)
node: v24.18.0
pnpm: 11.5.2
```

chiliec:
## What

Closes #11444.

`convertMessagesToGeminiContents` merged every pair of adjacent contents that
shared a role. A `ToolMessage` converts to `role: "user"` with a
`functionResponse` part, and a following `HumanMessage` converts to
`role: "user"` with a text part, so the two were folded into a single `user`
content:

```json
{ "role": "user", "parts": [ { "functionResponse": ... }, { "text": "continue" } ] }
```

Vertex/Gemini rejects a `user` content that mixes a `functionResponse` with
text and returns a 400 ("request ends with a model turn" — misleading, the
converted payload actually ends with `user`). This happens for the ordinary
history shape `ai(text + tool_call), tool, human`, e.g. when a user sends a
follow-up message after a tool has run. Once that pair is in the history every
subsequent turn on the thread fails.

## Fix

Only merge adjacent same-role contents when both carry a `functionResponse` or
neither does. A tool result therefore stays in its own `user` turn and is not
combined with a following human text message. Adjacent tool results (parallel
tool calls) still merge into one turn, as before.

The fix matches the maintainer-facing diagnosis in the issue: split the merged
content into two separate `user` contents.

## Tests

Added `keeps a ToolMessage and a following HumanMessage in separate user turns
(legacy path)` in
`libs/providers/langchain-google/src/converters/tests/messages.test.ts`,
reproducing the issue's exact failing sequence and asserting four separate
contents (`user`, `model`, `user`(functionResponse), `user`(text)).

## Validation (real results on this machine)

Node 22.14.0, `pnpm@10.14.0`, built `@langchain/core` locally, then:

```
pnpm exec vitest run src/converters/tests/messages.test.ts
```

- With the fix: **21 passed**, my new test green.
- RED→GREEN: reverting only `converters/messages.ts` makes the new test fail
  ("expected length 3 to be 4" — the functionResponse and human text are merged);
  restoring it makes it pass.
- `oxlint` on `src/converters/` — clean; `oxfmt` — no changes.

I could not run the live Vertex integration test (needs Google Cloud
credentials), so the behavior is verified at the converter level via the unit
test above, which is where the merge happens.

Note: three tests in this file (`merges consecutive ToolMessages…`, `AIMessage
with tool_calls produces model turn…`, `strips type field from executableCode…`)
fail identically on a clean `main` checkout in my local setup (a
`response_metadata is undefined` error unrelated to this change) and are outside
this diff; this PR changes only the adjacent-merge condition and adds one test.

First-time contributor here — happy to adjust naming or the merge condition if
you'd prefer a different shape.

### DIFF
--- a/.changeset/gemini-tool-human-merge.md
+++ b/.changeset/gemini-tool-human-merge.md
@@ -1,0 +1,9 @@
+---
+"@langchain/google": patch
+---
+
+Fix `convertMessagesToGeminiContents` merging a `ToolMessage` (functionResponse)
+with a following `HumanMessage` (text) into one `user` content. Both map to the
+`user` role, and Vertex/Gemini rejects a single `user` content that mixes a
+`functionResponse` with text, so a tool result followed by a user message would
+fail with a 400. Adjacent tool results (parallel tool calls) still merge.
--- a/libs/providers/langchain-google/src/converters/messages.ts
+++ b/libs/providers/langchain-google/src/converters/messages.ts
@@ -849,7 +849,18 @@ export const convertMessagesToGeminiContents: Converter<
     });
     if (content) {
       const prev = contents[contents.length - 1];
-      if (prev && prev.role === content.role) {
+      // Merge adjacent contents of the same role, but do NOT merge a content
+      // carrying a functionResponse (a converted ToolMessage) with one that
+      // does not (e.g. a following HumanMessage's text). Both map to the
+      // `user` role, and Gemini/Vertex rejects a single `user` content that
+      // mixes a functionResponse with text. Two functionResponse contents
+      // (parallel tool results) still merge, as do two plain contents.
+      // See issue #11444.
+      if (
+        prev &&
+        prev.role === content.role &&
+        hasFunctionResponse(prev) === hasFunctionResponse(content)
+      ) {
         prev.parts.push(...content.parts);
       } else {
         contents.push(content);
@@ -859,6 +870,14 @@ export const convertMessagesToGeminiContents: Converter<
 
   return contents;
 };
+
+/**
+ * Returns true if any part of the given content is a functionResponse part
+ * (i.e. it was produced from a ToolMessage).
+ */
+function hasFunctionResponse(content: Gemini.Content): boolean {
+  return content.parts.some((part) => "functionResponse" in part);
+}
 
 /**
  * Converts LangChain system messages to Gemini API system instruction format.
--- a/libs/providers/langchain-google/src/converters/tests/messages.test.ts
+++ b/libs/providers/langchain-google/src/converters/tests/messages.test.ts
@@ -283,6 +283,56 @@ describe("convertMessagesToGeminiContents", () => {
     expect(responses[1].functionResponse!.id).toBe("call-london");
   });
 
+  test("keeps a ToolMessage and a following HumanMessage in separate user turns (legacy path)", () => {
+    // A ToolMessage (functionResponse) and a following HumanMessage (text) both
+    // map to the `user` role. They must NOT be merged into one content: Gemini /
+    // Vertex rejects a single `user` content that mixes a functionResponse with
+    // text (see issue #11444). Expected: user, model, user(functionResponse),
+    // user(text) — four separate contents.
+    const messages = [
+      new HumanMessage("read it"),
+      new AIMessage({
+        content: "Reading the agreement now.",
+        tool_calls: [
+          {
+            name: "read_document_pages",
+            args: { fileId: "f1" },
+            id: "call_1",
+            type: "tool_call",
+          },
+        ],
+      }),
+      new ToolMessage({
+        content: "page text",
+        tool_call_id: "call_1",
+        name: "read_document_pages",
+      }),
+      new HumanMessage("continue"),
+    ];
+
+    const contents = convertMessagesToGeminiContents(messages);
+
+    expect(contents).toHaveLength(4);
+
+    expect(contents[0].role).toBe("user");
+    expect(contents[0].parts.some((p) => "text" in p)).toBe(true);
+    expect(contents[0].parts.some((p) => "functionResponse" in p)).toBe(false);
+
+    expect(contents[1].role).toBe("model");
+
+    // The tool response turn carries ONLY the functionResponse, no text.
+    expect(contents[2].role).toBe("user");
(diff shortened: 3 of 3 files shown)


glorat:
That was quick! I tested the change and it worked

hntrl (commented):
Hi Vlad! Thanks for contributing!

I'm going to continue this on in #11477. There's some additional tests semantic things I wanted to address before merging. Will make sure you get proper accreditation!

PR #11450, opened 2026-08-23
fix(core): convert every content block when formatting for tracing

### LINKED ISSUE #11291: _formatForTracing only converts the first base64 content block per message

likalight:
Fixes #11291.

## The bug

`_formatForTracing` (`libs/langchain-core/src/language_models/chat_models.ts`)
guards its shallow copy with `messageToTrace === message`:

```ts
if (isURLContentBlock(block) || isBase64ContentBlock(block)) {
  if (messageToTrace === message) {
    messageToTrace = new (message.constructor as any)({ ... });
  }
}
```

After the first convertible block the guard is false, so every later block is
skipped. A message carrying two images or two files traces the first converted
and the rest in their original shape.

There is a second latent problem in the same expression: each copy slices from
the original `message.content`, so even with the guard removed a second
conversion would have discarded the first.

## The fix

Collect the conversions into one copy of the content array, then construct the
traced message once, and only when something actually changed. Messages with no
convertible blocks are still passed through by reference, exactly as before.

## Verification

- Added a test using the reporter's repro shape — a `HumanMessage` with a text
  block and two base64 PDF blocks — asserting both file blocks convert and the
  text block is carried through untouched.
- Verified it fails without the source change:
  `AssertionError: expected { type: 'file', …(3) } to not have property "source_type"`
- `libs/langchain-core` `src/language_models/`: **102/102 tests pass**, no type errors
- `pnpm lint` and `pnpm format:check`: clean
- Changeset included.

### DIFF
--- a/.changeset/format-for-tracing-all-blocks.md
+++ b/.changeset/format-for-tracing-all-blocks.md
@@ -1,0 +1,5 @@
+---
+"@langchain/core": patch
+---
+
+Convert every URL and base64 content block when formatting messages for tracing, not just the first. `_formatForTracing` only built its shallow copy for the first convertible block, and each copy sliced from the original `message.content`, so a message carrying two images or two files traced the first converted and the rest in their original shape.
--- a/libs/langchain-core/src/language_models/chat_models.ts
+++ b/libs/langchain-core/src/language_models/chat_models.ts
@@ -173,22 +173,26 @@ function _formatForTracing(messages: BaseMessage[]): BaseMessage[] {
   for (const message of messages) {
     let messageToTrace = message;
     if (Array.isArray(message.content)) {
-      for (let idx = 0; idx < message.content.length; idx++) {
-        const block = message.content[idx];
+      const content = message.content;
+      // Collect every conversion into one copy. Converting in place against a
+      // freshly sliced `message.content` per block would discard the previous
+      // block's conversion, so a message carrying two images or two files
+      // traced only the first.
+      let convertedContent: typeof content | undefined;
+      for (let idx = 0; idx < content.length; idx++) {
+        const block = content[idx];
         if (isURLContentBlock(block) || isBase64ContentBlock(block)) {
-          if (messageToTrace === message) {
-            // Also shallow-copy content
-            // oxlint-disable-next-line @typescript-eslint/no-explicit-any
-            messageToTrace = new (message.constructor as any)({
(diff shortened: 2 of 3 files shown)


jackjin1997:
This is the same #11291 tracing conversion and test path already covered by #11292, open since August 3; #11338 is another later implementation. #11292 converts every eligible block in one copied message, preserves the original, includes the changeset and callback-path regression, and remains green. Could we keep the review on the earliest PR and fold in any genuinely distinct case from this version?

likalight:
@jackjin1997 agreed, and thanks for the pointer — #11292 predates this by nearly three weeks and I should have found it before opening. Reviewing it now, it covers the same ground: one copied message, every eligible block converted, original preserved, changeset and callback-path regression included.

I don't see a distinct case here worth carrying over. Consolidate on #11292; happy to close this whenever maintainers prefer, or leave it for reference if that's more useful.

jackjin1997:
Thanks for comparing them directly and confirming there is no distinct case to carry over. I will keep #11292 as the consolidation target.

PR #11453, opened 2026-08-23
fix(google): carry tool call ids through native stream events

### LINKED ISSUE #11261: @langchain/google native stream events drop tool call IDs

likalight:
Fixes #11261.

## The bug

The `functionCall` branch of `convertGoogleGeminiStream`
(`libs/providers/langchain-google/src/utils/stream_events.ts`) never set an `id`
— not on the block accumulator, not on `content-block-start`, not on
`content-block-delta`. `finalizeContentBlock` therefore emitted:

```js
{ type: "tool_call", name: "web_search", args: { query: "weather" }, id: undefined }
```

Nothing throws. A later `ToolMessage.tool_call_id` simply has no id to match,
and a server-assigned `functionCall.id` is discarded on the way through.

## The fix

Issue the id once, at block start, and read it from the accumulator afterwards
so one stable value survives start → delta → finalization:

- A server-assigned `functionCall.id` is preserved.
- When Gemini omits one, generate the same `lc-tool-call-*` shape the
  non-streaming converter in this package already uses
  (`convertGeminiPartsToToolCalls`), so both paths now have the same contract.

## Verification

Three tests in `stream_events.test.ts`:

1. A server-assigned id appears on `content-block-start`, on
   `content-block-delta`, and on the finalized `tool_call`.
2. With no id from Gemini, the fallback matches `/^lc-tool-call-[0-9a-f]{32}$/`
   and the finish event carries the **same** id as block start — one id, not a
   fresh one per event.
3. Two concurrent `functionCall` parts get distinct ids.

All three fail without the source change
(`expected undefined to be 'server-assigned-id-123'`, `expected undefined to be defined`).

- `libs/providers/langchain-google` unit tests: **164 pass**, no type errors
- `pnpm lint` and `pnpm format`: clean
- Changeset included.

## Unrelated pre-existing failures

`src/converters/tests/messages.test.ts` has 3 failures on this branch, but they
reproduce identically on a clean `main` with no changes applied:

- `merges consecutive ToolMessages into a single function turn for parallel tool calls (legacy path)`
- `AIMessage with tool_calls produces model turn with functionCall parts (legacy path)`
- `strips type field from executableCode content blocks`

Untouched by this PR — flagging in case they are not already known.

### DIFF
--- a/.changeset/google-stream-tool-call-ids.md
+++ b/.changeset/google-stream-tool-call-ids.md
@@ -1,0 +1,5 @@
+---
+"@langchain/google": patch
+---
+
+Carry tool-call IDs through native Gemini stream events. The `functionCall` branch of `convertGoogleGeminiStream` never set an `id` on the block accumulator, `content-block-start` or `content-block-delta`, so `finalizeContentBlock` produced a `tool_call` block with `id: undefined` and a later `ToolMessage.tool_call_id` had nothing to match. A server-assigned `functionCall.id` is now preserved, and when Gemini omits one the same `lc-tool-call-*` fallback the non-streaming converter uses is generated — one stable id from block start through finalization.
--- a/libs/providers/langchain-google/src/utils/stream_events.ts
+++ b/libs/providers/langchain-google/src/utils/stream_events.ts
@@ -10,6 +10,7 @@ import type {
   FinishReason,
 } from "@langchain/core/language_models/event";
 import type { ContentBlock, UsageMetadata } from "@langchain/core/messages";
+import { v4 as uuidv4 } from "@langchain/core/utils/uuid";
 import type { Gemini } from "../chat_models/api-types.js";
 
 export type GeminiStreamResponse = Gemini.GenerateContentResponse;
@@ -123,10 +124,17 @@ export async function* convertGoogleGeminiStream(
       } else if (part.functionCall) {
         const key: BlockKey = `tool:${toolIdx}`;
         const args = JSON.stringify(part.functionCall.args ?? {});
+        // One id must survive block start -> delta -> finalization so a later
+        // `ToolMessage.tool_call_id` can match it. Gemini may omit `id`, so
+        // fall back to the same generated shape the non-streaming converter
+        // uses rather than leaving the finalized block without one.
+        const toolCallId =
+          part.functionCall.id ?? `lc-tool-call-${uuidv4().replace(/-/g, "")}`;
         const { index, isNew } = getOrCreateBlockIndex(key, {
           type: "tool_call_chunk",
           name: part.functionCall.name,
           args: "",
+          id: toolCallId,
           index: toolIdx,
         });
         if (isNew) {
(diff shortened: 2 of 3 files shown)


jackjin1997:
This appears to cover the same #11261 start/delta/final ID propagation as #11262, which has been open since July 27. The earlier PR preserves server IDs, generates one stable fallback, and covers the converter, public `ChatGoogle.streamEvents` path, and a credentialed integration test source. To avoid parallel review, could we consolidate on #11262 unless there is a distinct semantic case here? I am happy to bring over any missing edge.

likalight:
@jackjin1997 agreed — #11262 has been open since 27 July and covers the same start/delta/finalization ID propagation, plus a credentialed integration test this one doesn't have. My apologies for the duplicate; I checked assignees and comments before starting, which misses an existing PR entirely.

The only edge I'd flag for #11262 is a test asserting that two concurrent `functionCall` parts get *distinct* ids — worth a line if it isn't already covered, since one shared fallback across parallel calls would still satisfy a single-call test. Otherwise nothing distinct here. Consolidate on #11262; happy to close this on your or a maintainer's word.

jackjin1997:
Thanks for checking the overlap and for calling out the parallel-call case. The implementation already allocates each part from its own accumulator key, and I have now pushed 55fae9222 to #11262 with an explicit two-concurrent-function-calls regression asserting distinct fallback ids. The focused stream-event suite, formatting, lint, and package build all pass.

PR #11454, opened 2026-08-23
fix(google-common): send functionCallingConfig.mode in uppercase

### LINKED ISSUE #11265: bug(@langchain/google-common): `functionCallingConfig.mode` is serialized lowercase, which Vertex AI silently ignores — `tool_choice` ("any"/"none"/forced function name) is a no-op on ChatVertexAI

likalight:
Fixes #11265.

## The bug

The Gemini API's `FunctionCallingConfig.Mode` is a case-sensitive enum — `AUTO`,
`ANY`, `NONE`. `formatToolConfig` in `@langchain/google-common` emitted
lowercase in both branches:

```ts
// ...and for a forced function name:
mode: "any",
```

A lowercase value matches no enum member, so Vertex AI silently ignores the
field and the request stays in the default AUTO mode. The practical effect is
that `tool_choice` of `"any"`, `"none"`, or a forced function name is a **no-op**
on `ChatVertexAI`: the model may answer in plain text instead of calling the
function, and forced calls lose ANY-mode schema-constrained decoding. Nothing
errors, so it fails quietly.

## The fix

Uppercase the wire value in both branches, and narrow the type in `types.ts`
from `"auto" | "any" | "none"` to the enum's actual members.

`tool_choice` is still accepted lowercase from users — only what goes on the
wire changes.

This also brings the package in line with its sibling: the newer
`@langchain/google` already emits uppercase and even normalizes a lowercase
input (`convertToolChoiceToGeminiConfig`, see
`libs/providers/langchain-google/src/converters/tests/tools.test.ts`).
`google-common` was the outlier.

## One thing to confirm

The `types.ts` change narrows a publicly exported type. Anyone hand-constructing
a `GeminiRequest` with `mode: "any"` will now see a type error — which is the
point, since that value never worked, but it is visible rather than silent. Say
the word if you would rather accept both cases in the type and normalize
internally.

## Verification

Two tests in `chat_models.test.ts`, using the existing mock-request `record`
pattern:

1. `tool_choice` of `auto` / `any` / `none` each serialize to `AUTO` / `ANY` /
   `NONE`.
2. A forced function name serializes to
   `{ mode: "ANY", allowedFunctionNames: ["my_tool"] }`.

Both fail without the source change
(`expected 'auto' to be 'AUTO'`, `expected { mode: 'any', …(1) } to deeply equal { mode: 'ANY', …(1) }`).

- `@langchain/google-common` unit tests: **302/302 pass**, no type errors
- Downstream consumers (`google-gauth`, `google-vertexai`, `google-vertexai-web`,
  `google-webauth`) typecheck clean against the narrowed type
- `pnpm lint` and `pnpm format`: clean
- Changeset included.

### DIFF
--- a/.changeset/google-common-uppercase-function-calling-mode.md
+++ b/.changeset/google-common-uppercase-function-calling-mode.md
@@ -1,0 +1,5 @@
+---
+"@langchain/google-common": patch
+---
+
+Serialize `toolConfig.functionCallingConfig.mode` in the Gemini enum's uppercase form. The Gemini API's `FunctionCallingConfig.Mode` is case-sensitive (`AUTO`, `ANY`, `NONE`), so the lowercase values sent previously matched no member and were silently ignored — leaving requests in the default AUTO mode. `tool_choice` of `"any"`, `"none"` or a forced function name was therefore a no-op on ChatVertexAI, and forced calls lost ANY-mode schema anchoring. `tool_choice` is still accepted lowercase; only the wire value changes. This matches the newer `@langchain/google` package, which already emits uppercase.
--- a/libs/providers/langchain-google-common/src/tests/chat_models.test.ts
+++ b/libs/providers/langchain-google-common/src/tests/chat_models.test.ts
@@ -2890,6 +2890,75 @@ describe("Mock ChatGoogle - Gemini", () => {
     );
   });
 
+  test("6. tool_choice serializes functionCallingConfig.mode uppercase", async () => {
+    const myTool = tool(async ({ query }) => `Result for ${query}`, {
+      name: "my_tool",
+      description: "A custom tool",
+      schema: z.object({ query: z.string() }),
+    });
+
+    // The Gemini enum is case-sensitive; a lowercase mode is silently ignored
+    // and the request falls back to AUTO.
+    const cases: [string, string][] = [
+      ["auto", "AUTO"],
+      ["any", "ANY"],
+      ["none", "NONE"],
+    ];
+
+    for (const [toolChoice, expected] of cases) {
+      const record: Record<string, any> = {};
+      const projectId = mockId();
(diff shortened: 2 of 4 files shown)


jackjin1997:
This overlaps the same issue and wire-format change as #11266, which has been open since July 29. #11266 also covers all three modes plus forced function names, includes a `ChatVertexAI` integration-path assertion, and all visible checks are green. To keep review consolidated, could we compare any distinct value in the public type narrowing here and carry it into the earlier PR if needed? I am happy to incorporate that edge if maintainers prefer it.

likalight:
@jackjin1997 agreed, #11266 came first (29 July) and has the `ChatVertexAI` integration-path assertion this one lacks. Consolidate on it.

One piece here may be genuinely distinct, which is the bit you flagged: this also narrows the exported type in `types.ts` from `"auto" | "any" | "none"` to `"AUTO" | "ANY" | "NONE"`. That's what stops the lowercase value being reintroduced by a future caller — a fix confined to `formatToolConfig` leaves the type still advertising the values that never worked. It is a visible narrowing, so anyone hand-constructing a `GeminiRequest` with `mode: "any"` would get a type error, which is the point but worth a maintainer's call.

Happy for you to carry that into #11266, or to close this either way.

jackjin1997:
Thanks for confirming. #11266 already includes the same exported `GeminiRequest` narrowing to `"AUTO" | "ANY" | "NONE"`, together with the wire normalization and downstream typechecks, so there is no missing edge to carry over.

PR #11457, opened 2026-08-24
fix(openai): avoid duplicate usage metadata in streams

### LINKED ISSUE #11054: ChatOpenAI streaming duplicates response_metadata.model_name / finish_reason (and doubles response_metadata.usage) when the provider emits multiple finish_reason chunks (e.g. OpenRouter)
### Checked other resources

- [x] I used the GitHub search to find a similar issue; the closest is #10159 (ChatGoogle, `usage_metadata`) — this is a distinct case (ChatOpenAI, different fields).
- [x] The bug is not resolved by updating to the latest stable version (confirmed against `@langchain/openai` 1.4.7).

### Example Code

```ts

  model: "anthropic/claude-haiku-4.5",
  apiKey: process.env.OPENROUTER_KEY,
  streaming: true,
  streamUsage: true,
  configuration: { baseURL: "https://openrouter.ai/api/v1" },
});

const msg = await model.invoke("Say hello in 3 words.");
console.log(msg.response_metadata.model_name);     // "anthropic/...-haiku...anthropic/...-haiku..."  (doubled)
console.log(msg.response_metadata.finish_reason);  // "stopstop"
console.log(msg.response_metadata.usage.prompt_tokens, msg.usage_metadata.input_tokens); // 30  15
//                                                  ^ response_metadata doubled; usage_metadata correct
```

_No response — silent data corruption, not a thrown error._

### Description

When a provider's stream emits **more than one chunk with `finish_reason != null`** (OpenRouter sends two `stop` chunks), `ChatOpenAI`'s completions streaming produces corrupted `response_metadata`:

- `model_name`, `finish_reason`, `system_fingerprint`, `service_tier` are set on **every** finish-reason chunk:

  ```ts
  // src/chat_models/completions.ts (~L173-177)
  if (choice.finish_reason != null) {
    generationInfo.finish_reason = choice.finish_reason;
    generationInfo.system_fingerprint = data.system_fingerprint;
    generationInfo.model_name = data.model;
    generationInfo.service_tier = data.service_tier;
  }
  ```

  `AIMessageChunk.concat` then **string-concatenates** these scalar fields across chunks → `"stopstop"`, a doubled `model_name`, etc.

- Numeric `response_metadata.usage` is **summed** across chunks (e.g. `prompt_tokens: 30` for a 15-token prompt). The normalized `usage_metadata` is unaffected (stays `15`), so token tracking via `usage_metadata` is fine, but anything reading `response_metadata` (downstream tracers such as Langfuse, and the raw provider `usage.cost`) gets 2×.

The OpenAI spec sends `finish_reason` on a single chunk, so this only surfaces with OpenAI-compatible providers that emit it more than once — but the result (a garbage `model_name`) is clearly unintended regardless of the upstream quirk.

Related: #10159 (same class of bug for ChatGoogle / `usage_metadata`).

### Proposed fix

Treat these as **scalar, last-wins** metadata rather than letting chunk-concat accumulate them — e.g. only attach `model_name` / `finish_reason` / `system_fingerprint` / `service_tier` once, or de-duplicate them when merging finish-reason chunks.

### System Info

- `@langchain/openai` 1.4.5 (also reproduced reading 1.4.7 source)
- Node.js 24
- Provider: OpenRouter via OpenAI-compatible `baseURL`

brianpham93:
Fixes #11054

## Summary

Prevent duplicate token usage in both streamed `AIMessageChunk.response_metadata` and native content-block usage events when an OpenAI-compatible provider repeats cumulative usage snapshots.

This provider-agnostic fix applies to raw OpenAI, Azure OpenAI, LiteLLM, OpenRouter, GPTSL, and other adapters using the shared OpenAI Completions stream converter.

Related: #11084 addresses duplicated scalar metadata such as `finish_reason` and `model_name`; this PR addresses the remaining numeric usage aggregation issue from the same class of multi-terminal streams.

## Root cause

`ChatOpenAICompletions` forwarded usage metadata from choice-bearing chunks and then emitted the same usage again on the dedicated final usage chunk. LangChain chunk aggregation adds response metadata, so repeated cumulative usage was counted more than once.

The native content-block converter also emitted every repeated cumulative snapshot as a separate `usage` event. Consumers of `streamEvents()` or the usage sub-stream could therefore observe duplicate telemetry events.

The provider raw usage values are not necessarily wrong. The client must not add or re-emit the same cumulative snapshot repeatedly.

## Changes

- Remove `response_metadata.usage` from choice-bearing chunks and preserve usage on the dedicated final usage chunk.
- Deduplicate identical normalized cumulative usage snapshots in the shared OpenAI Completions stream-event converter while preserving genuinely changed snapshots.
- Keep the latest usage snapshot on the final `message-finish` event.
- Preserve the raw provider response in `additional_kwargs.__raw_response`.
- Avoid per-token metadata cleanup when the raw chunk does not contain usage.
- Add regression coverage for repeated and changing usage snapshots plus OpenAI-compatible, Azure OpenAI, and LiteLLM-compatible configurations.
- Add patch releases for `@langchain/core` and `@langchain/openai`.

### Azure OpenAI through LiteLLM

Tested LiteLLM 1.98.0 routing to the configured Azure OpenAI deployment.

The raw LiteLLM stream contained one usage event: 13 input tokens, 10 output tokens, 23 total tokens.

With the unpatched `@langchain/openai` commit `fca7d2f8b`:

- The stream exposed the same non-empty usage snapshot twice.
- `invoke()` returned response metadata of 26 input, 20 output, and 46 total tokens.
- `usage_metadata` remained 13 input, 10 output, and 23 total tokens.

With this patch, the same path returned one usage snapshot and the correct totals.

### OpenRouter

A live OpenRouter stream reproduced the same behavior with the unpatched client:

- Raw provider usage: 15 input, 9 output, 24 total tokens.
- Unpatched response metadata: 30 input, 18 output, 48 total tokens.
- Patched response metadata: 15 input, 9 output, 24 total tokens.

## Tests

- [x] Focused core converter tests: 3 passed with no type errors.
- [x] Focused OpenAI streaming tests: 14 passed with no type errors.
- [x] Full `@langchain/core` suite: 1,513 passed, 1 skipped, no type errors.
- [x] Full `@langchain/openai` suite executed 329 passing tests; the command still reports 104 pre-existing dependency/API type errors outside this change.
- [x] Package builds passed from the package directories, including `attw` and `publint`.
- [x] Lint, formatting, and `git diff --check` passed.

### DIFF
--- a/.changeset/fix-openai-stream-usage-metadata.md
+++ b/.changeset/fix-openai-stream-usage-metadata.md
@@ -1,0 +1,5 @@
+---
+"@langchain/openai": patch
+---
+
+fix duplicate usage metadata when OpenAI-compatible providers repeat cumulative usage snapshots during streaming
--- a/.changeset/fix-openai-stream-usage-metadata.md
+++ b/.changeset/fix-openai-stream-usage-metadata.md
@@ -1,4 +1,5 @@
 ---
+"@langchain/core": patch
 "@langchain/openai": patch
 ---
 
--- a/libs/langchain-core/src/language_models/openai_completions_stream.ts
+++ b/libs/langchain-core/src/language_models/openai_completions_stream.ts
@@ -96,6 +96,17 @@ export interface ConvertOpenAICompletionsStreamOptions {
 
 type BlockKey = "text" | "reasoning" | "audio" | `tool:${number}`;
 
+const TOKEN_DETAIL_KEYS = [
+  "audio",
+  "cache_creation",
+  "cache_read",
+  "document",
+  "image",
+  "reasoning",
+  "text",
+  "video",
+] as const;
+
 /**
  * Convert an async iterable of OpenAI Chat Completions-shaped stream chunks into
  * LangChain `ChatModelStreamEvent`s with typed deltas.
@@ -116,7 +127,7 @@ export async function* convertOpenAICompletionsStream(
   const blockKeyToIndex = new Map<BlockKey, number>();
   let nextBlockIndex = 0;
   let messageStarted = false;
-  let usageSnapshot: UsageMetadata | undefined;
+  let lastEmittedUsageSnapshot: UsageMetadata | undefined;
   let finishReason: FinishReason | undefined;
   let responseMetadata: Record<string, unknown> | undefined;
   let emittedProviderMetadata = false;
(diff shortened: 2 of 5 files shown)


brianpham93:
Follow-up review findings addressed in the latest branch update (head `640482ac45e7b38c8b195c189b18b63b00b7b399`). The shared OpenAI Completions stream converter now suppresses repeated identical cumulative usage events while preserving changed snapshots and the final latest snapshot. The provider-side cleanup is also guarded so it does not allocate metadata work for ordinary token chunks without raw usage. Added core regression coverage for repeated/changed snapshots; focused core/OpenAI tests and the full core suite pass.

PR #11490, opened 2026-08-26
fix(openai): skip kwargs.tool_calls when invalid_tool_calls is set

guda-art:
## Description

When streaming tool-call arguments fail to parse, `@langchain/core` records `invalid_tool_calls` with `Malformed args.` and leaves `tool_calls` empty. LangGraph `toolsCondition` therefore routes to `END` without a `ToolMessage`.

On the next model call, `@langchain/openai` still serializes `additional_kwargs.tool_calls` (the raw OpenAI payload kept for compatibility) onto the Completions request. The provider then sees an assistant turn with `tool_calls` and no matching `role: tool` results, and returns `400 INVALID_TOOL_RESULTS`.

This change skips the `additional_kwargs.tool_calls` fallback when `invalid_tool_calls` is non-empty. Messages that only have legacy kwargs (no `invalid_tool_calls`) are unchanged.

## Motivation

Same `AIMessage` currently has two semantics:
- Graph routing looks at empty `tool_calls` → no tool execution
- Outbound conversion looks at `additional_kwargs.tool_calls` → request still has `tool_calls`

That mismatch is not implied by the OpenAI tool-calling protocol; it is leftover compatibility behavior.

## Changes

- `convertStandardContentMessageToCompletionsMessage`
- `convertMessagesToCompletionsMessageParams`

## Test plan

- [x] Unit tests for the converter (legacy + v1 / standard-content paths)
- [x] Persist an AIMessage with empty `tool_calls`, non-empty `invalid_tool_calls`, and `additional_kwargs.tool_calls`; next `invoke` must not send `tool_calls`
- [x] Legacy AIMessage with only `additional_kwargs.tool_calls` still serializes them

### DIFF
--- a/.changeset/skip-kwargs-tool-calls-when-invalid.md
+++ b/.changeset/skip-kwargs-tool-calls-when-invalid.md
@@ -1,0 +1,5 @@
+---
+"@langchain/openai": patch
+---
+
+fix(openai): skip additional_kwargs.tool_calls when invalid_tool_calls is set
--- a/libs/providers/langchain-openai/src/converters/completions.ts
+++ b/libs/providers/langchain-openai/src/converters/completions.ts
@@ -670,7 +670,13 @@ export const convertStandardContentMessageToCompletionsMessage: Converter<
       completionParam.tool_calls = message.tool_calls.map(
         convertLangChainToolCallToOpenAI
       ) as OpenAIClient.Chat.Completions.ChatCompletionMessageToolCall[];
-    } else if (message.additional_kwargs.tool_calls != null) {
+    } else if (
+      message.additional_kwargs.tool_calls != null &&
+      !(
+        AIMessage.isInstance(message) &&
+        (message.invalid_tool_calls?.length ?? 0) > 0
+      )
+    ) {
       completionParam.tool_calls = message.additional_kwargs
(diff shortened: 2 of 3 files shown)


yesprasad:
Nice fix — this closes a real mismatch between graph routing (`tool_calls` empty) and outbound serialization (legacy `additional_kwargs.tool_calls` still present).
One review suggestion: because the guard is applied in both the standard-content and legacy converter paths, could we add regression coverage for:
1. `invalid_tool_calls` + raw legacy `additional_kwargs.tool_calls` → no outbound tool calls
2. legacy kwargs without `invalid_tool_calls` → remains unchanged
3. both converter paths
That would lock in the compatibility boundary this PR is restoring.

guda-art:
Thanks for the review — I added regression coverage for both converter paths:

1. `invalid_tool_calls` + legacy `additional_kwargs.tool_calls` → no outbound `tool_calls`
2. legacy kwargs without `invalid_tool_calls` → still serialized unchanged
3. both the legacy `convertMessagesToCompletionsMessageParams` path and the v1 / `convertStandardContentMessageToCompletionsMessage` path

Also added a patch changeset for `@langchain/openai` so the fix is included in the next version bump.

PR #11498, opened 2026-08-28
feat(langchain): support HITL respond decisions

### LINKED ISSUE #11356: HITL docs specify a `respond` decision that no released version implements
### Summary

The JS human-in-the-loop docs document a `respond` decision and recommend it for `ask_user`-style tools, but no published `langchain` release implements it. Sending the documented decision throws at runtime.

### What the docs say

https://docs.langchain.com/oss/javascript/langchain/human-in-the-loop lists four decision types — `approve`, `edit`, `reject`, `respond` — and says:

> Use `respond` for "ask user" style tools where the tool's real implementation is the human's reply.

with the example:

```ts
{ type: "respond", message: "Blue." }
```

### What the package does

`langchain/dist/agents/middleware/hitl.d.ts`:

```ts
declare const DecisionType: z.ZodEnum<["approve", "edit", "reject"]>;
```

So resuming with the documented decision fails:

```
Unexpected human decision: {"type":"respond","message":"ap-south-1"}.
Decision type 'respond' is not allowed for tool 'ask_user'.
Expected one of ["approve","edit","reject"] based on the tool's configuration.
```

### Versions

`langchain` 1.5.2 and 1.5.5 (latest at time of writing) — `["approve","edit","reject"]` in both. The newest prerelease tag (`next` = 1.0.0-alpha.9) is older.

### Impact

There is no documented alternative for an ask-the-human tool. `reject`'s `message` does reach the model as the tool result, but it is recorded with `status: "error"`, and any `reject` in a batch drops the approved calls in that same batch — so it is not equivalent.

### Ask

Either ship `respond`, or update the JS docs to match the released decision set and describe the intended pattern for ask-the-human tools.

xianjianlf2:
Adds support for a `respond` human-in-the-loop decision so reviewers can answer on behalf of a tool without executing it.

This creates a successful `ToolMessage` from the reviewer-provided response, keeps responded tool calls in the revised assistant message, and routes back to the model when there are no remaining approved/edited tool calls to execute. The default `interruptOn: true` behavior remains unchanged and still expands to approve/edit/reject only; `respond` must be explicitly allowed.

Fixes #11356

Validation:
- `corepack pnpm --filter langchain build`
- `corepack pnpm --dir libs/langchain exec vitest run src/agents/middleware/tests/hitl.test.ts`
- `corepack pnpm exec oxfmt --check libs/langchain/src/agents/middleware/hitl.ts libs/langchain/src/agents/middleware/tests/hitl.test.ts`
- `git diff --check`

### DIFF
--- a/.changeset/hitl-respond-decision.md
+++ b/.changeset/hitl-respond-decision.md
@@ -1,0 +1,5 @@
+---
+"langchain": patch
+---
+
+feat(langchain): support HITL respond decisions
--- a/libs/langchain/src/agents/middleware/hitl.ts
+++ b/libs/langchain/src/agents/middleware/hitl.ts
@@ -73,9 +73,10 @@ export type DescriptionFactory = z.infer<typeof DescriptionFunctionSchema>;
 /**
  * The type of decision a human can make.
  */
-const ALLOWED_DECISIONS = ["approve", "edit", "reject"] as const;
+const ALLOWED_DECISIONS = ["approve", "edit", "reject", "respond"] as const;
 const DecisionType = z.enum(ALLOWED_DECISIONS);
 export type DecisionType = z.infer<typeof DecisionType>;
+const DEFAULT_ALLOWED_DECISIONS: DecisionType[] = ["approve", "edit", "reject"];
 
 const InterruptOnConfigSchema = z.object({
   /**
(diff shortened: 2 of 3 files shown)


Parker-Fawcett:
HITL respond decisions are a great fit for mechanical contract enforcement - the decision interface (what fields, what format, what happens on timeout) should be locked before agents start producing responses. I built rebuild-dossier around this exact pattern for rebuild specs. Related work: https://arxiv.org/abs/2608.23616

PR #11505, opened 2026-08-30
fix(mistralai): preserve streamed tool call IDs

ahjulstad:
(This PR is AI-authored and human-supervised. )

## Summary
- preserve the prior tool-call ID when Mistral streams an unnamed continuation with `id: "null"`
- add a regression test for the observed named-call / null-ID continuation sequence

## Problem
Mistral-hosted `glm-5-2` can stream a tool call in two chunks: the first has the function name and a normal ID, while the continuation has arguments but `name: ""` and `id: "null"`. The adapter assigned a new UUID to that continuation, preventing LangChain from merging its arguments with the original tool call.

## Validation
- `pnpm --filter @langchain/mistralai exec vitest run src/tests/chat_models.test.ts -t null`

The package-wide TypeScript check remains blocked by two unrelated existing errors in `internal/standard-tests/src/integration_tests/chat_models.ts` and `src/utils/stream_events.ts`.

This fixes the sequential continuation shape. Parallel responses that omit distinct IDs/indexes for multiple continuations remain ambiguous at the provider boundary.

### DIFF
--- a/libs/providers/langchain-mistralai/src/chat_models.ts
+++ b/libs/providers/langchain-mistralai/src/chat_models.ts
@@ -449,13 +449,14 @@ function mistralAIResponseToChatMessage(
   }
 }
 
-function _convertDeltaToMessageChunk(
+export function _convertDeltaToMessageChunk(
   delta: {
     role?: string | null | undefined;
     content?: string | MistralAIContentChunk[] | null | undefined;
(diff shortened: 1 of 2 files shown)


ahjulstad:
A repo that reproduces the issue is [here](https://github.com/ahjulstad/mistral-parallel-tool-call-repro)

PR #11518, opened 2026-09-01
feat(deepseek): surface DeepSeek-R1 reasoning_content in non-streaming responses

### LINKED ISSUE #8201: please surport for deepseek-reasoner(deepseek-r1)
please surport for deepseek-reasoner(deepseek-r1):
Please support the output of deepseek-reasoner's thinking process in the form of <think>thinking process</think>, achieving the same effect as the commonly displayed thinking process in the market.

LeonxLJX:
## Summary
Closes #8201

Non-streaming `invoke()` now surfaces DeepSeek-R1 reasoning the same way the streaming path already does:
- exposes the dedicated `reasoning_content` field (deepseek-reasoner) into `additional_kwargs.reasoning_content`
- extracts inline `<think>...</think>` reasoning from `content` and moves it into `additional_kwargs.reasoning_content`

This makes R1's thinking process consistently available whether or not streaming is used.

## Test plan
- [ ] maintainer review / CI

### DIFF
--- a/.changeset/tidy-waves-reason.md
+++ b/.changeset/tidy-waves-reason.md
@@ -1,0 +1,13 @@
+---
+"@langchain/deepseek": patch
+---
+
+Expose DeepSeek-R1 `reasoning_content` in non-streaming responses.
+
+Non-streaming `invoke()` previously dropped the model's reasoning, so callers
+saw an inconsistent shape depending on whether they used `stream()` (which
+already surfaced `reasoning_content`) or `invoke()`. The change copies the
+provider's `reasoning_content`/`reasoning` field into
+`response_metadata.additional_kwargs.reasoning_content`, and also extracts an
+inline `<think>...</think>` block into the same key so both DeepSeek-R1
+response styles are represented consistently.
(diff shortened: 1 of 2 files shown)


LeonxLJX:
Design notes for the non-streaming reasoning surfacing.

**Root cause.** The streaming path already exposed DeepSeek-R1 `reasoning_content`, but non-streaming `invoke()` didn't — so the same model gave inconsistent shapes depending on how you called it (issue #8201). That breaks anyone building on `additional_kwargs` reasoning uniformly.

**Judgment.** I prefer the dedicated `reasoning_content` (deepseek-reasoner) / `reasoning` field first, then fall back to extracting inline `<think>...</think>` from `content` and moving it into `additional_kwargs.reasoning_content`. Cleaning the `<think>` wrapper out of `content` keeps the visible message clean and mirrors what the streaming path already does. The guard `!langChainMessage.additional_kwargs.reasoning_content` avoids clobbering a value the field path already set.

**Suggestion.** Some DeepSeek-compatible endpoints name it `reasoning` rather than `reasoning_content`; I already fall back to `reasoning`, so we're covered. A unit test asserting the non-streaming round-trip (field path + inline `<think>` path) would lock this in — I can add one if useful.

LeonxLJX:
Bumping this for visibility. PR extends the existing DeepSeek-R1 `reasoning_content` surfacing to the non-streaming `invoke()` path (the streaming path already had it via the `additional_kwargs.reasoning_content` extraction). On the non-streaming path the reasoning now lands in `result.additional_kwargs.reasoning_content` instead of being concatenated into the regular content. Verified against the DeepSeek API docs and the test fixture already in `libs/providers/langchain-deep-seek/src/tests`. Changeset bot already detected the bump. Happy to iterate if the @langchain/core team wants the field under a different name.

PR #11539, opened 2026-09-02
fix(openai): pass full message to parser in jsonSchema fallback (fixe…

### LINKED ISSUE #11535: ChatOpenAI jsonSchema fallback passes ContentBlock[] to parser as string
### Checked other resources

- [x] This is a bug, not a usage question. For questions, please use the LangChain Forum.
- [x] I added a descriptive title.
- [x] I searched the LangChain.js docs and GitHub issues/PRs.
- [x] I reproduced this on the latest stable packages.

### Example Code

The failure can be reproduced without API credentials by injecting a mock chat-completions implementation that returns the same valid v1 content-block response:

```javascript
import { StructuredOutputParser } from "@langchain/core/output_parsers";
import { z } from "zod";

const schema = z.object({ answer: z.string() });
  content: [{ type: "text", text: '{"answer":"ok"}' }],
  response_metadata: { output_version: "v1" },
});

const parser = StructuredOutputParser.fromZodSchema(schema);

// This succeeds because the parser receives a BaseMessage.
console.log(await parser.invoke(message));
// => { answer: "ok" }

const mockCompletions = {
  getLsParams() {
    return {};
  },
  invocationParams() {
    return {};
  },
  async _generate() {
    return {
      generations: [{ text: message.text, message }],
    };
  },
};

  apiKey: "not-used",
  model: "gpt-5",
  completions: mockCompletions,
});

await model
  .withStructuredOutput(schema, { method: "jsonSchema" })
  .invoke("Return JSON");
```

```text
OutputParserException: Failed to parse. Text: "undefined". Error: TypeError: Cannot read properties of undefined (reading 'trim')

Troubleshooting URL: https://docs.langchain.com/oss/javascript/langchain/errors/OUTPUT_PARSING_FAILURE/
```

### Description

`ChatOpenAI.withStructuredOutput(..., { method: "jsonSchema" })` has a fallback for responses where `additional_kwargs.parsed` is absent. The fallback currently passes `aiMessage.content` to the content parser after a TypeScript-only cast:

```ts
if ("parsed" in aiMessage.additional_kwargs) {
  return aiMessage.additional_kwargs.parsed as RunOutput;
}
return altParser.invoke(aiMessage.content as string);
```

Source: [`libs/providers/langchain-openai/src/chat_models/base.ts`](https://github.com/langchain-ai/langchainjs/blob/921437ef1f345459eba1dee5d64679b6435fb393/libs/providers/langchain-openai/src/chat_models/base.ts#L1155-L1163)

When `aiMessage.content` is a `ContentBlock[]`, the cast does not convert it at runtime. `BaseLLMOutputParser.invoke()` sees a non-string value and treats the array as a `BaseMessage`. `StructuredOutputParser._baseMessageToString()` then reads `message.text` from the array, producing `undefined`, and `parse()` fails at `text.trim()`.

The response itself is valid:

```ts
new AIMessage({
  content: [{ type: "text", text: '{"answer":"ok"}' }],
  additional_kwargs: {},
  response_metadata: { output_version: "v1" },
});
```

Expected: `{ answer: "ok" }`.

Actual: `OUTPUT_PARSING_FAILURE` with `Text: "undefined"`.

This was originally observed with an Azure OpenAI GPT-5-compatible deployment, where structured JSON was returned as a v1 text content block and `additional_kwargs.parsed` was absent. The mock above isolates the provider-independent fallback bug.

Passing the full message appears to be the minimal fix:

```diff
- return altParser.invoke(aiMessage.content as string);
+ return altParser.invoke(aiMessage);
```

This lets `StructuredOutputParser` / `StandardSchemaOutputParser` use the `BaseMessage.text` accessor, consistent with the fix merged in #10814. Passing `aiMessage.text` would also avoid the invalid runtime type, but passing the message preserves the parser contract directly.

### Related work / duplicate check

- #10437 and #10814 fixed content-block parsing when a parser receives a full `BaseMessage`.
- This is a different call path: the OpenAI `jsonSchema` fallback strips the message and passes `ContentBlock[]` as though it were a string.
- I found no issue or PR covering this exact fallback path.

### System Info

Reproduced with the originally reported versions:

```text
@langchain/core 1.1.48
@langchain/openai 1.4.7
langchain 1.4.4
zod 4.4.3
```

Also reproduced with the latest stable versions as of 2026-09-02:

```text
@langchain/core 1.2.9
@langchain/openai 1.5.11
langchain 1.5.10
zod 4.4.3
Node.js v25.5.0
macOS 26.5.2 (arm64)
```

Margalit111:
Fixes #11535

### Description
In `@langchain/openai`, `ChatOpenAI.withStructuredOutput` (using the `jsonSchema` method) has a fallback path when `additional_kwargs.parsed` is missing. Previously, it passed `aiMessage.content as string` to the alternative parser (`altParser.invoke(...)`). 

When `aiMessage.content` is returned as an array of content blocks (`ContentBlock[]`), the cast to `string` is invalid at runtime. The parser then treated it incorrectly, attempting to access `.trim()` on `undefined` and throwing an `OutputParserException`.

This PR updates the call to pass the full `aiMessage` object directly (`altParser.invoke(aiMessage)`). This allows the parser to properly extract the text via the standard `BaseMessage.text` accessor, resolving the runtime error.

### Tests
- Added a regression unit test in `completions.test.ts` verifying that `withStructuredOutput` successfully parses output when the model returns an array of content blocks.
- Verified all tests pass via `pnpm vitest run src/chat_models/tests/completions.test.ts`.

### DIFF
--- a/.changeset/tall-places-pump.md
+++ b/.changeset/tall-places-pump.md
@@ -1,0 +1,6 @@
+---
+"@langchain/openai": patch
+---
+
+fix(openai): pass full message to parser in jsonSchema fallback (fixe…
+)
--- a/libs/providers/langchain-openai/src/chat_models/base.ts
+++ b/libs/providers/langchain-openai/src/chat_models/base.ts
@@ -1159,7 +1159,7 @@ export abstract class BaseChatOpenAI<
             if ("parsed" in aiMessage.additional_kwargs) {
               return aiMessage.additional_kwargs.parsed as RunOutput;
             }
-            return altParser.invoke(aiMessage.content as string);
+            return altParser.invoke(aiMessage);
           }
         );
       } else {
--- a/libs/providers/langchain-openai/src/chat_models/tests/completions.test.ts
+++ b/libs/providers/langchain-openai/src/chat_models/tests/completions.test.ts
@@ -4,6 +4,8 @@ import { toJsonSchema } from "@langchain/core/utils/json_schema";
 import { HumanMessage, AIMessageChunk } from "@langchain/core/messages";
 import { CallbackManagerForLLMRun } from "@langchain/core/callbacks/manager";
 import { ChatOpenAICompletions } from "../completions.js";
+import { AIMessage } from "@langchain/core/messages";
+import { ChatOpenAI } from "../index.js";
 
 describe("ChatOpenAICompletions constructor", () => {
   it("supports string model shorthand", () => {
@@ -373,5 +375,36 @@ describe("ChatOpenAICompletions strict tools for structured output", () => {
     expect(
       toolStrict({ response_format: { type: "json_object" } })
     ).toBeUndefined();
+  });
+
+  it("should parse structured output when content is an array of content blocks", async () => {
+    const schema = z.object({ answer: z.string() });
+
+    const message = new AIMessage({
+      content: [{ type: "text", text: '{"answer":"ok"}' }],
+      response_metadata: { output_version: "v1" },
+    });
+
+    const mockCompletions = {
+      getLsParams: () => ({}),
+      invocationParams: () => ({}),
+      async _generate() {
+        return {
+          generations: [{ text: message.text, message }],
+        };
+      },
+    };
+
+    const model = new ChatOpenAI({
+      apiKey: "test",
+      model: "gpt-5",
+      completions: mockCompletions as any,
+    });
+
+    const res = await model
+      .withStructuredOutput(schema, { method: "jsonSchema" })
+      .invoke("Return JSON");
+
+    expect(res).toEqual({ answer: "ok" });
   });
 });


davidgovea:
Independently confirmed this fixes #11535. We encountered the same fallback failure when a structured-output call nested in a streamed agent run returned valid JSON content blocks without `additional_kwargs.parsed`.

I reduced it to this [20-line, credential-free test](https://github.com/davidgovea/langchainjs/blob/724bfc6f7e19981e8626cae12c25cd0f066b4ff5/libs/providers/langchain-openai/src/chat_models/tests/structured-output-content-blocks.test.ts), mocking only the model invocation so the real `withStructuredOutput` parser runs:

```ts
import { afterEach, expect, it, vi } from "vitest";
import { z } from "zod/v3";
import { ChatOpenAI } from "../index.js";

afterEach(() => vi.restoreAllMocks());

it("parses JSON from content blocks without provider-parsed output", async () => {
  const model = new ChatOpenAI({ model: "gpt-4o-mini", apiKey: "unused" });
  vi.spyOn(ChatOpenAI.prototype, "invoke").mockResolvedValue(
    new AIMessage({ content: [{ type: "text", text: '{"answer":"ok"}' }] })
  );
  await expect(
    model
      .withStructuredOutput(z.object({ answer: z.string() }), {
        method: "jsonSchema",
      })
      .invoke("Return JSON")
  ).resolves.toEqual({ answer: "ok" });
});
```

On current main (`56130def875051cc39c5fd8af8e6bfa6119fd9c5`, `@langchain/openai` 1.5.12-rc.0 / core 1.2.9), it fails with:

```text
Failed to parse. Text: "undefined". Error: TypeError: Cannot read properties of undefined (reading 'trim')
```

Applying this PR's diff (`dd4ff70722748c2a7a80639371b67d17d159078c`) to that main revision makes it pass. Running this test alongside `completions.test.ts` gives **13 passing tests**, with no type errors:

```sh
pnpm --filter @langchain/openai test \
  src/chat_models/tests/structured-output-content-blocks.test.ts \
  src/chat_models/tests/completions.test.ts
```

Thanks for the fix. Passing the message preserves the parser's text extraction and resolves the independently reproduced failure.

PR #11553, opened 2026-09-03
fix(google): preserve thought and thoughtSignature on resent text/reasoning/functionCall parts

### LINKED ISSUE #10461: bug(@langchain/google): thinking blocks lose thought/thoughtSignature in multi-turn conversations

thushanth-bengre-langchain:
## Summary

Fixes #10461 — resending a prior thinking turn's history to Gemini stripped
`thought`/`thoughtSignature` markers, contradicting Gemini's documented
requirement to echo thought blocks back verbatim

This picks up and extends #11198.

## What's included

- The original fix from #11198: `convertStandardContentBlockToGeminiPart`
  now preserves `thought`/`thoughtSignature` on both the `"text"` and
  `"reasoning"` content-block cases, and the legacy-path text branch does
  the same.
- **New**: `convertStandardContentMessageToGeminiContent` rebuilds
  `functionCall` parts directly from `AIMessage.tool_calls` (not from a
  content-block item), which was a separate path that silently dropped
  `thoughtSignature` on tool calls made by a thinking model. Addresses
  @afirstenberg's review comment that thought signatures "can be on any
  `Gemini.Part`, not just text"

## Deliberately out of scope

@afirstenberg's review also raised a deeper architectural point: a Gemini
part with `thought: true` arguably should already be normalized to a
`ContentBlock.Reasoning` at the producer (`convertGeminiPartToContentBlock`),
rather than the converters having to tolerate the raw `{ type: "text",
thought: true }` shape on the way back out.

That is a  **breaking change**, not a bugfix — the
raw shape is what `AIMessage.content` has always contained for thinking
responses, and `BaseMessage.text` in core currently concatenates it in. 
Changing the producer's output shape would change `.text`'s output for every existing
thinking-model caller.

Hence this will be done in a separate PR., explicitly-scoped
follow-up rather than bundling a breaking change into this fix.

Credit to @saarRozenbaum for the original diagnosis and fix in #11198.

### DIFF
--- a/libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
+++ b/libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
@@ -1816,6 +1816,12 @@ describe.each(thinkingModelInfo)(
         .thoughtSignature;
       expect(toolCallSignature).toBeDefined();
 
+      // Force v1 - the default (legacy) route already preserves this signature
+      firstResult.response_metadata = {
+        ...firstResult.response_metadata,
+        output_version: "v1",
+      };
+
       await llm.invoke([
         new HumanMessage("What is the weather in New York?"),
         firstResult,
--- a/libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
+++ b/libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
@@ -1743,6 +1743,62 @@ describe.each(thinkingModelInfo)(
       );
       expect(hasThoughtSignature).toBe(true);
     });
+
+    test("thought/thoughtSignature survive a multi-turn round trip (regression for #10461)", async () => {
+      // Reproduces the production bug: resending a prior thinking AIMessage
+      // as history silently stripped its thought marker/signature before
+      // this fix, either losing the flag (legacy content array) or dropping
+      // the block outright (contentBlocks -> ChatGoogleTranslator ->
+      // "reasoning" -> default: return null in the standard converter).
+      const llm = newChatGoogle({ reasoningEffort: "low" });
+
+      const firstResult = await llm.invoke("Why is the sky blue?");
+      const firstReasoningBlocks = firstResult.contentBlocks.filter(
+        (b) => b.type === "reasoning"
+      );
+      expect(firstReasoningBlocks.length).toBeGreaterThan(0);
+      const firstSignatures = firstResult.contentBlocks
+        .filter((b) => "thoughtSignature" in b)
+        .map((b) => b.thoughtSignature);
+      expect(firstSignatures.length).toBeGreaterThan(0);
+
+      // Feed the thinking AIMessage back in as history and ask a follow-up.
+      // recorder.request captures the actual outbound HTTP body for the
+      // second call — assert the resent turn's `contents` entry still
+      // carries `thought`/`thoughtSignature` on the wire, not just that the
+      // call succeeds (a successful call alone wouldn't catch a silently
+      // dropped/unflagged thought block, since Gemini doesn't require the
+      // resent history to include one).
+      await llm.invoke([
+        new HumanMessage("Why is the sky blue?"),
+        firstResult,
+        new HumanMessage("Now explain it to a 5 year old."),
+      ]);
+
+      const sentContents = recorder.request?.body?.contents ?? [];
+      const modelTurn = sentContents.find(
+        (c: { role?: string }) => c.role === "model"
+      );
+      expect(modelTurn).toBeDefined();
+
+      const sentThoughtParts = (modelTurn?.parts ?? []).filter(
+        (p: { thought?: boolean }) => p.thought === true
+      );
+      expect(sentThoughtParts.length).toBeGreaterThan(0);
+
+      const sentSignatures = (modelTurn?.parts ?? [])
+        .filter((p: { thoughtSignature?: string }) => "thoughtSignature" in p)
+        .map((p: { thoughtSignature?: string }) => p.thoughtSignature);
+      expect(sentSignatures.length).toBeGreaterThan(0);
+      // At least one signature from the first response must round-trip
+      // verbatim — Gemini validates these, so a regenerated/blank value
+      // would defeat the point of preserving the field at all.
+      expect(
+        sentSignatures.some((sig: string | undefined) =>
+          firstSignatures.includes(sig)
+        )
+      ).toBe(true);
+    });
   }
 );
 
--- a/libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
+++ b/libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
@@ -1801,11 +1801,7 @@ describe.each(thinkingModelInfo)(
     });
 
     test("functionCall thoughtSignature survives a multi-turn round trip after a real tool call (root cause: tool_calls-rebuild path)", async () => {
-      // convertStandardContentMessageToGeminiContent rebuilds functionCall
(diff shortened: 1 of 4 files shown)


libs/providers/langchain-google/src/converters/messages.ts
```diff
@@ -392,7 +392,28 @@ function convertStandardContentBlockToGeminiPart(
 ): Gemini.Part | null {
   switch (block.type) {
     case "text":
-      return { text: block.text };
+      return {
+        text: block.text,
+        ...("thought" in block ? { thought: block.thought as boolean } : {}),
```
afirstenberg: Once again - why should we have "thought" be in a text block?

libs/providers/langchain-google/src/converters/messages.ts
```diff
@@ -392,7 +392,28 @@ function convertStandardContentBlockToGeminiPart(
...
+      // ChatGoogleTranslator (block_translators/google.ts) normalizes a
+      // Gemini thinking part into `{ type: "reasoning", reasoning, thought,
+      // thoughtSignature, reasoningContentBlock }` for the standard
+      // `contentBlocks` getter. Round-trip it back into the shape Gemini
+      // expects — without this, a thinking model's own chain-of-thought is
+      // silently dropped (hits the `default` branch below) when history
+      // that went through `.contentBlocks` is replayed on the next turn.
+      return {
+        text: block.reasoning,
+        ...("thought" in block ? { thought: block.thought as boolean } : {}),
```
afirstenberg: If this is a reasoning block, then "thought" should always be true.

libs/providers/langchain-google/src/converters/messages.ts
```diff
@@ -392,7 +392,28 @@ function convertStandardContentBlockToGeminiPart(
...
+    case "reasoning":
+      // ChatGoogleTranslator (block_translators/google.ts) normalizes a
+      // Gemini thinking part into `{ type: "reasoning", reasoning, thought,
+      // thoughtSignature, reasoningContentBlock }` for the standard
+      // `contentBlocks` getter. Round-trip it back into the shape Gemini
+      // expects — without this, a thinking model's own chain-of-thought is
+      // silently dropped (hits the `default` branch below) when history
+      // that went through `.contentBlocks` is replayed on the next turn.
+      return {
+        text: block.reasoning,
```
afirstenberg, line 406: Not all reasoning blocks are text blocks with Gemini (although they seem to have removed it from thoughts in recent models).

afirstenberg (commented):
If I follow, this fix is supposed to handle *just* the issue with `thoughtSignature`, correct?

That changing how `AIMessage.content` formats messages would be a breaking change, so will be handled elsewhere. (Using `AIMessage.content` has many other problems. See #10047. But that is outside this scope.)

To me it looks like this tries to go outside the scope of handling just `thoughtSignature`, but I'm getting honestly confused about what this is really supposed to fix. And the integration tests don't make it any clearer.

libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
```diff
@@ -1743,6 +1743,101 @@ describe.each(thinkingModelInfo)(
       );
       expect(hasThoughtSignature).toBe(true);
     });
+
+    test("thought/thoughtSignature survive a multi-turn round trip (regression for #10461)", async () => {
```
afirstenberg, line 1747: I'm confused. I don't see a regression.

When I run this test against the current main branch, it passes.
What is it supposed to be testing?

thushanth-bengre-langchain: I ran this on unmodified main and I see the failure.(`gemini-3.6-flash`)
`sentThoughtParts.length` is 0 on `main`.
So the flag is getting stripped. 

Happy to share the full run output.
Also curious, what model/config showed it passing for you. I couldn't reproduce that.

libs/providers/langchain-google/src/converters/messages.ts
thushanth-bengre-langchain: Check 5 different models from this year and all has 
`{ text, thought: true }` + `{ text, thoughtSignature }`.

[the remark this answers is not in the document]
thushanth-bengre-langchain: Added a gate to do this only if `thought: true`

libs/providers/langchain-google/src/converters/messages.ts
thushanth-bengre-langchain: Yes. You are right.

Added `thought: true` to this case

libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
afirstenberg: My mistake, I tagged the wrong test as passing.

libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
```diff
@@ -1743,6 +1743,97 @@ describe.each(thinkingModelInfo)(
...
+      // verbatim — Gemini validates these, so a regenerated/blank value
+      // would defeat the point of preserving the field at all.
+      expect(
+        sentSignatures.some((sig: string | undefined) =>
+          firstSignatures.includes(sig)
+        )
+      ).toBe(true);
+    });
+
+    test("functionCall thoughtSignature survives a multi-turn round trip after a real tool call (root cause: tool_calls-rebuild path)", async () => {
```
afirstenberg, line 1803: My mistake above. This test appears to be passing on everything at the moment. (All current models gemini-3.1 through gemini-3.8, both gai and gcp platforms.)

libs/providers/langchain-google/src/converters/messages.ts
hntrl: this is how its getting passed through on the translation end atm: https://github.com/langchain-ai/langchainjs/pull/11553/changes#diff-5418c413c4cad4ec819452d2eebf798a513d305f8788e0261cb970757d0b1220R1140-R1142

I think ideally this would be standard content shaped but I think I have to concede because it is already there

hntrl (dismissed):

libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
thushanth-bengre-langchain: No issues with default v0.
But when we use `output_version` v1, it fails on main.

I changed output version in the latest commit. This should fail on `main`.

thushanth-bengre-langchain:
> If I follow, this fix is supposed to handle _just_ the issue with `thoughtSignature`, correct?
> 
> That changing how `AIMessage.content` formats messages would be a breaking change, so will be handled elsewhere. (Using `AIMessage.content` has many other problems. See #10047. But that is outside this scope.)
> 
> To me it looks like this tries to go outside the scope of handling just `thoughtSignature`, but I'm getting honestly confused about what this is really supposed to fix. And the integration tests don't make it any clearer.

Correct.
It handles `thought` too.
Covers 3 gaps:
`text`, `reasoning-block` and `tool-call`(v1 only).

Int-tests fail on `main` now.

afirstenberg:
> It handles thought too.

I guess I'm still confused about why it has to handle `thought` being set in text.  Under what conditions is that happening?

thushanth-bengre-langchain:
Closing this in favor of #11574

PR #11563, opened 2026-09-05
fix(google-genai): forward abort signal to non-streaming requests

### LINKED ISSUE #9667: Abort Controller for GoogleGenAI
### Checked other resources

### Example Code

const model = new ChatGoogleGenerativeAI(parameters)
await model.invoke(messages, { signal: step.abortSignal() })

_No response_

### Description

Error on @langchain/google-genai/chat_models.js

The parameters are rewritten on the _generate method by using the invocationParams method.
It'll then disregard additional parameters, specially the abort attribute which prevents from using an abort controller to suspend the current run.

### System Info

node@22
@langchain/google-genai@2.1.0

adimalkar:
Fixes #9667

`ChatGoogleGenerativeAI._generate` called `completionWithRetry` with only the request, dropping the second `options` argument. `completionWithRetry` uses that argument in two places, so both received `undefined`:

```ts
return this.caller.callWithOptions(
  { signal: options?.signal },
  async () => await this.client.generateContent(request, { signal: options?.signal }),
);
```

The result is that `model.invoke(messages, { signal })` cannot abort an in-flight non-streaming request. The `options.signal?.throwIfAborted()` at the top of `_generate` only covers the moment before dispatch.

Both streaming paths already forward the signal — `_streamChatModelEvents` and `_streamResponseChunks` each pass it to `callWithOptions` and to `generateContentStream` — so this was a single missed call site on the non-streaming path. `langchain-anthropic` uses the same idiom at three call sites.

## How I verified it

- Added a regression test asserting the signal reaches `completionWithRetry`. It **fails** before the change and passes after.
- Package suite goes from 93 to 94 passing, no type errors, no existing test affected.
- `oxfmt --check` and `oxlint` clean on both changed files.

The package had no existing abort/signal coverage, so the new test is the first.

**Scope of verification:** the integration tests require live API keys, so I confirmed the signal is plumbed through at the unit level rather than that a real Gemini request cancels end to end.

### DIFF
--- a/.changeset/calm-impalas-learn.md
+++ b/.changeset/calm-impalas-learn.md
@@ -1,0 +1,5 @@
+---
+"@langchain/google-genai": patch
+---
+
+Forward abort signals to non-streaming Google Generative AI requests.
--- a/libs/providers/langchain-google-genai/src/chat_models.ts
+++ b/libs/providers/langchain-google-genai/src/chat_models.ts
@@ -933,7 +933,8 @@ export class ChatGoogleGenerativeAI
     }
 
     const res = await this.completionWithRetry(
-      this._buildGenerateContentRequest(messages, options)
+      this._buildGenerateContentRequest(messages, options),
+      options
     );
 
     let usageMetadata: UsageMetadata | undefined;
--- a/libs/providers/langchain-google-genai/src/tests/chat_models.test.ts
+++ b/libs/providers/langchain-google-genai/src/tests/chat_models.test.ts
@@ -1234,3 +1234,29 @@ describe("withStructuredOutput - StandardSchema", () => {
     expect((result as any).parsed).toEqual({ name: "cobalt" });
   });
 });
+
+test("Google AI - abort signal reaches the non-streaming request", async () => {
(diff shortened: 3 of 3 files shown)


adimalkar:
Closing this — #9667 was closed as `NOT_PLANNED` on the grounds that `@langchain/google-genai` is being superseded by `@langchain/google`, and I checked that claim rather than just taking it.

`libs/providers/langchain-google/src/chat_models/base.ts` already forwards the signal on both request paths, in both of the places this PR adds it to the older package:

```ts
response = await this.caller.callWithOptions(
  { signal: options.signal, maxRetries: options.maxRetries },
  async () => {
    const nextResponse = await this.apiClient.fetch(
      new Request(url, { method: "POST", headers, body: JSON.stringify(body), signal: options.signal })
    );
```

That holds for `_generate` and for `_streamResponseChunks`, so there is no gap to re-target and no second issue worth opening. The defect was real in `langchain-google-genai` — `completionWithRetry` received `undefined` where `options` was expected, so the non-streaming path never saw the signal while both streaming paths did — but it is not worth landing in a package being retired.

Thanks for the pointer to the unified package.

PR #11565, opened 2026-09-05
fix(core): handle zero and negative maxima during normalization

### LINKED ISSUE #11564: normalize returns NaN for all-zero distance matrices
### Checked other resources

- [x] I searched the existing issues and pull requests for this normalization problem.
- [x] The bug is present on the current main branch.

### Example Code

```ts
import { euclideanDistance, normalize } from '@langchain/core/utils/math';

const distances = euclideanDistance([[1, 0]], [[1, 0]]);
console.log(distances); // [[0]]
console.log(normalize(distances)); // [[NaN]]
console.log(normalize(distances, true)); // [[NaN]]
```

### Description

`normalize` divides each entry by the matrix maximum. An all-zero distance matrix (including distances between identical embeddings) has a maximum of zero, so both distance and similarity normalization compute `0 / 0` and return `NaN`.

Expected: zero distances remain zero, and converting those zero distances to similarities returns 1. Nonzero matrices should retain their existing results.

A focused regression test against the repository source fails with `expected [[NaN, NaN], [NaN, NaN]] to deeply equal [[0, 0], [0, 0]]`. Using 1 as the divisor when the maximum is zero keeps the output finite. Tests also cover empty matrix shapes and nonzero similarity normalization.

I have prepared this focused fix and its tests and would like to contribute it; please assign this issue to me if assignment is needed.

### System Info

macOS arm64, Node.js v24.19.0, pnpm 10.14.0, current `@langchain/core` source (package version 1.2.9). No external services or API keys are needed. Prepared with AI assistance.

mehuljariwala:
Fixes #11564.

`normalize` divides all-zero matrices by zero, while its maximum helper treats all-negative matrices as having a maximum of zero and lets empty rows contaminate the result with NaN. Find the maximum from nonempty rows and use a divisor of 1 only when that maximum is exactly zero. A NaN maximum is no longer silently replaced with 1.

Keep the existing maximum-based scaling and `1 - value / divisor` similarity calculation. Document that finite non-negative inputs produce values in `[0, 1]`, while signed inputs can fall outside that range. Positive-input behavior is preserved.

Regression coverage includes zero matrices in both modes, all-negative matrices, mixed signs, empty rows among nonempty rows, entirely empty shapes, signed inputs with a zero maximum, NaN maxima, and unchanged input. Include a patch changeset for `@langchain/core`.

Validation: three added regression tests failed against the preceding PR revision. The updated utility suite passes 286 tests across 15 files with no reported type errors. Repository Oxlint and formatting checks pass. No provider requests or API keys were used. Prepared with OpenAI Codex assistance.

### DIFF
--- a/.changeset/finite-zero-normalization.md
+++ b/.changeset/finite-zero-normalization.md
@@ -1,0 +1,5 @@
+---
+"@langchain/core": patch
+---
+
+Avoid NaN scores when normalizing an all-zero distance matrix.
--- a/.changeset/finite-zero-normalization.md
+++ b/.changeset/finite-zero-normalization.md
@@ -2,4 +2,4 @@
 "@langchain/core": patch
 ---
 
-Avoid NaN scores when normalizing an all-zero distance matrix.
+Handle zero maxima, all-negative matrices, and empty rows when normalizing matrices.
--- a/libs/langchain-core/src/utils/math.ts
+++ b/libs/langchain-core/src/utils/math.ts
@@ -47,7 +47,7 @@ export function matrixFunc(
 }
 
 export function normalize(M: number[][], similarity = false): number[][] {
-  const max = matrixMaxVal(M);
+  const max = matrixMaxVal(M) || 1;
   return M.map((row) =>
     row.map((val) => (similarity ? 1 - val / max : val / max))
   );
--- a/libs/langchain-core/src/utils/math.ts
+++ b/libs/langchain-core/src/utils/math.ts
@@ -46,10 +46,17 @@ export function matrixFunc(
   );
 }
 
+/**
+ * Divide each value by the matrix maximum, using a divisor of 1 when it is zero.
+ * When similarity is true, return 1 minus each scaled value.
+ * Finite non-negative inputs produce values in [0, 1]; signed inputs can produce
+ * values outside that range. This does not perform min-max scaling or clamping.
+ */
 export function normalize(M: number[][], similarity = false): number[][] {
-  const max = matrixMaxVal(M) || 1;
+  const max = matrixMaxVal(M);
+  const divisor = max === 0 ? 1 : max;
   return M.map((row) =>
-    row.map((val) => (similarity ? 1 - val / max : val / max))
+    row.map((val) => (similarity ? 1 - val / divisor : val / divisor))
   );
 }
 
@@ -172,8 +179,13 @@ function argMax(array: number[]): MaxInfo {
 }
 
 function matrixMaxVal(arrays: number[][]): number {
-  return arrays.reduce(
-    (acc, array) => Math.max(acc, argMax(array).maxValue),
-    0
-  );
+  let max: number | undefined;
+  for (const array of arrays) {
+    if (array.length === 0) {
+      continue;
+    }
+    const { maxValue } = argMax(array);
+    max = max === undefined ? maxValue : Math.max(max, maxValue);
+  }
+  return max ?? 0;
 }
--- a/libs/langchain-core/src/utils/tests/math_utils.test.ts
+++ b/libs/langchain-core/src/utils/tests/math_utils.test.ts
@@ -165,6 +165,45 @@ test("Test normalize", async () => {
   expect(actual).toEqual(expected);
 });
 
+test("normalize returns finite distances for an all-zero matrix", () => {
+  const input = [
+    [0, 0],
+    [0, 0],
+  ];
+  expect(normalize(input)).toEqual([
+    [0, 0],
+    [0, 0],
+  ]);
+  expect(normalize(input, true)).toEqual([
+    [1, 1],
+    [1, 1],
+  ]);
+  expect(input).toEqual([
+    [0, 0],
+    [0, 0],
+  ]);
+});
+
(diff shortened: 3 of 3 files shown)


yeshwarprasad:
@mehuljariwala  My core Evidence Engine Reviewed this PR - Observations requiring validation


ADVISORY
Area to check 1
libs/langchain-core/src/utils/math.ts:177
Changed code
```
function matrixMaxVal(arrays: number[][]): number {
  return arrays.reduce(
    (acc, array) => Math.max(acc, argMax(array).maxValue),
    0
  );
}
```
Concern
matrixMaxVal initializes the reducer accumulator to 0, which returns 0 for matrices whose per-row maxima are all negative; normalize() then falls back to dividing by 1, producing incorrect scaling (and for similarity=true, values can exceed 1).

Impacted code path
`libs/langchain-core/src/utils/math.ts:matrixMaxVal `→ `libs/langchain-core/src/utils/math.ts:normalize`
Failure / impact
`matrixMaxVal` initializes the reducer accumulator to 0, which returns 0 for matrices whose per-row maxima are all negative; normalize() then falls back to dividing by 1, producing incorrect scaling (and for similarity=true, values can exceed 1).

Failure condition
`normalize()` is called with a matrix M where every row’s maximum value is < 0 (e.g., cosine similarity matrices that are all negative) or rows may be empty. matrixMaxVal returns 0, causing `normalize()` to use `max=1` and produce unscaled negatives or 1 - (negative)/1 > 1 for similarity=true.

Suggested repair
Initialize the reducer with -Infinity and handle empty rows explicitly, e.g., return -Infinity for empty rows and then coerce overall max to 0 only when all values are -Infinity. Alternatively, compute the max across all numbers while skipping NaN. Consider full min-max normalization when inputs may be negative: const min = matrixMinVal(M); const max = matrixMaxVal(M); then scale by (max - min), guarding zero range.

Required regression coverage
Add tests for normalize() with: (1) matrices where all values are negative, (2) mixed sign matrices, and (3) matrices containing empty rows. Verify outputs stay in expected bounds (e.g., for similarity=true, do not exceed 1) and match intended normalization semantics.

@mehuljariwala  let me know if this was helpful in making this PR more stronger with evidence. Happy to Help.

yesprasad:
@mehuljariwala  `normalize()` can behave incorrectly for matrices with **all-negative** values due to `matrixMaxVal’s` accumulator initialization.

mehuljariwala:
@yeshwarprasad @yesprasad Thanks for pointing out the negative-maximum and empty-row cases. I reproduced both and addressed them in 5e7ddad.

The maximum now starts from the first nonempty row, so an all-negative matrix uses its actual maximum and empty rows are ignored. The unit-divisor fallback is now explicitly `max === 0`, so it does not hide a NaN maximum. Added regression coverage for all-negative inputs, mixed signs, empty rows, zero maxima, and NaN maxima.

There is one distinction about the expected bounds: this function divides by the maximum; it is not a min-max normalizer. For example, `normalize([[-4, -2]])` now returns `[[2, 1]]`, and the existing formula for `normalize([[-2, 0, 4]], true)` returns `[[1.5, 1, 0]]`. Bounds of `[0, 1]` apply to finite non-negative inputs. I documented that explicitly and retained the existing formula, since changing to min-max scaling would also change established positive-input results.

Three new regressions failed on the prior PR revision. The updated utility suite passes 286 tests across 15 files with no type errors; repository lint and formatting checks pass as well.

yeshwarprasad:
@mehuljariwala  `TracePull` reviewed this PR and here are the observations - 

Assessment: Reviewed 2 connected review units: 2 model-assessed and 0 risk-skipped. This group only adds a changeset note with no code changes; no issues or cross-unit impacts detected. Changes in math.ts improve edge-case handling for normalization; no cross-boundary contract or graph issues detected and no static findings to confirm.

Go/No-Go: GO

Recommended action: Safe to merge based on the reviewed evidence.

Severity: advisory

Contract status: ambiguous

🧠 Why this matters
Reviewed 2 file groups with 0 deterministic findings. 0 issues identified.

🧭 TracePull impact
Changed files: 3
Scope: contained
Affected artifacts: 0
Direct consumers: 0
Transitive consumers: 0
Maximum depth: 0
Affected projects: 0
Resolution: complete
🧩 Review-unit coverage
| Unit | Status | Files | Findings |

| --- | --- | --- | --- |

| 1 | completed | libs/langchain-core/src/utils/math.ts | 0 |

| 2 | completed | .changeset/finite-zero-normalization.md | 0 |

🧭 Verified behavior flow
```mermaid
flowchart TD
  code-40caa805b6["libs/langchain-core/src/utils/math.ts"]
  style code-40caa805b6 fill:#e8f5e9
  doc-633d109361[".changeset/finite-zero-normalization.md"]
  style doc-633d109361 fill:#e8f5e9
  test-8e665281e6["libs/langchain-core/src/utils/tests/math_utils.test.ts"]
  style test-8e665281e6 fill:#e8f5e9
  doc-633d109361 -.->|"documents contract for"| code-40caa805b6
  test-8e665281e6 -.->|"validates"| code-40caa805b6
```

hntrl (commented):
See https://github.com/langchain-ai/langchainjs/issues/11564#issuecomment-5554065710

PR #11574, opened 2026-09-07
fix(google): thought and thoughtSignature on round trip

### LINKED ISSUE #10461: bug(@langchain/google): thinking blocks lose thought/thoughtSignature in multi-turn conversations

afirstenberg:
Fixes #10461 
Supersedes: #11198 #11553

### Background

This addresses four specific issues:
* "thought" and "thoughtSignature" not being present in the conversation history sent to Gemini when using v0 messages
* "thought" and "thoughtSignature" not being present in the conversation history sent to Gemini when using v1 messages
* "thoughtSignature" not being present in v1 tool calls specifically
* `ContentBlock.Reasoning` not being handled in v1 at all

It does so by, largely, keeping the existing logic converting from LangChain representation to a `Google.Part` in both the v0 (legacy) and v1 (standard) paths intact. (This is done in both in the messages.ts file.)
* In `convertLegacyContentMessageToGeminiContent()`, the conversion is moved to a new function: `convertLegacyPartToGeminiPart()` for consistency with the standard path and to standardize where and how the thought/thoughtSignature fields are added
* Both `convertLegacyPartToGeminiPart()` and `convertStandardContentBlockToGeminiPart()` get a `baseGeminiPart()` function that returns the `Gemini.Part` _without_ the thought or thoughtSignature fields
* Both then add the additional fields, if necessary, independent of what the underlying `Gemini.Part` is.

The last point is done because Gemini can attach "thought" or "thoughtSignature" to _any_ block type. Not just text.

### What about the question @afirstenberg kept raising about why this wasn't in a "reasoning" block?

Well, there were two reasons:
* In v0, there is no reasoning block. And there shouldn't be.
* We were missing the "reasoning" handling in the v1 path.

So it was correct that we make sure that text blocks get this additional metadata added back to it.

The v0 content blocks are mostly, but not completely, like the format that is sent by Gemini. It is almost arguable we should just send it back after removing the "type" field. But I wasn't going to dig into that.

I am satisfied, at this point, that we store the v0 in the message `content`, that we should keep doing so, and that we are handling things correctly.

### Why this and not #11553?

This adds the fields back in a slightly different way than #11553 that I think is more correct.

* Issue 11553 adds it back to specific block types. Gemini can take this field on any block type.
* In the v1 path, issue 11553 handles a "reasoning" block by assuming text reasoning, which is an invalid assumption.

### Why wasn't this caught by previous tests?

In some cases, Gemini is tolerant. In other cases... I'm not actually sure why it wasn't.

### Tests

I have added all the tests in chat_models/tests/index.int.test.ts and converters/tests/messages.test.ts from #11553. (But see note about a change to the integration tests.) I have done integration testing against all current models and platforms that we are testing against.

Changes to integration tests:
* Both added tests originally set the reasoningLevel to "low". This didn't guarantee that Gemini would do any thinking at all.
* Changed the reasoning level for both to "high". Which still doesn't guarantee it, but there is a much much higher chance.

Results:
* All "Google Core" and "Google Thinking" tests pass in the integration tests, including the new tests.
* Except as noted below, all the messages tests pass.

Failing tests:
* Four previous tests in messages.test.ts fail. However, these four _also_ failed in issue 11553 and in the current main branch. Several of them _do_ involve tool calls, so they should be investigated more closely.
  * merges consecutive ToolMessages into a single function turn for parallel tool calls (legacy path)
  * AIMessage with tool_calls produces model turn with functionCall parts (legacy path)
  * ToolMessage v1 standard image content block is excluded from functionResponse.result (v1 path)
  * strips type field from executableCode content blocks
* v1 contentBlocks (explicit output_version): preserves thought and thoughtSignature on a resent text block
  * I'm not sure this is a valid test. 
  * It is, essentially, saying that if you add the "thought" field to a text block, which is not a valid v1 attribute, it should "do the right thing".
* a non-Google reasoning block with no thought flag is dropped, not sent as visible text
  * I think this is an invalid test, or at least doesn't match the description.
  * This doesn't test anything about "visible text", and it is explicitly a v1 block, so it is expected to be converted to a Gemini.Part with text and thinking set to true.

### Thanks

My thanks to @jackjin1997 @saarRozenbaum @thushanth-bengre-langchain and @hntrl for helping dig into this.

### DIFF
--- a/libs/providers/langchain-google/src/converters/messages.ts
+++ b/libs/providers/langchain-google/src/converters/messages.ts
@@ -381,6 +381,25 @@ function convertStandardVideoContentBlockToGeminiPart(
   return ret;
 }
 
+function convertStandardReasoningBlockToGeminiPart(
+  block: ContentBlock.Reasoning
+): Gemini.Part | null {
+  // Reasoning blocks created from Gemini should contain the base
+  // standard block they would be if "thought" wasn't set to true.
+  // If it exists, we'll use that. Otherwise, we'll assume it was
+  // originally a text block.
+  const ret = "reasoningContentBlock" in block && block.reasoningContentBlock
+    ? convertStandardContentBlockToGeminiPart(block.reasoningContentBlock as ContentBlock.Standard)
+    : {text: block.reasoning}
+
+  // Then add that this came from the thinking process
+  if (ret) {
+    ret.thought = true;
+  }
+
+  return ret;
+}
+
 /**
  * Converts a single LangChain standard content block (v1 format)
  * into a Gemini.Part.
@@ -390,19 +409,32 @@ function convertStandardVideoContentBlockToGeminiPart(
 function convertStandardContentBlockToGeminiPart(
   block: ContentBlock.Standard
 ): Gemini.Part | null {
-  switch (block.type) {
-    case "text":
-      return { text: block.text };
-    case "image":
-    case "audio":
-    case "text-plain":
-    case "file":
-      return convertStandardDataContentBlockToGeminiPart(block);
-    case "video":
-      return convertStandardVideoContentBlockToGeminiPart(block);
-    default:
-      return null;
+
+  function baseGeminiPart(): Gemini.Part | null {
+    switch (block.type) {
+      case "text":
+        return { text: block.text };
+      case "reasoning":
+        return convertStandardReasoningBlockToGeminiPart(block);
+      case "image":
+      case "audio":
+      case "text-plain":
+      case "file":
+        return convertStandardDataContentBlockToGeminiPart(block);
+      case "video":
+        return convertStandardVideoContentBlockToGeminiPart(block);
+      default:
+        return null;
+    }
   }
+
+  const ret: Gemini.Part | null = baseGeminiPart();
+  if (ret) {
+    if ("thoughtSignature" in ret) {
+      ret.thoughtSignature = ret.thoughtSignature!;
+    }
+  }
+  return ret;
 }
 
 /**
--- a/libs/providers/langchain-google/src/converters/messages.ts
+++ b/libs/providers/langchain-google/src/converters/messages.ts
@@ -727,6 +727,13 @@ function convertLegacyPartToGeminiPart(
   }
 
   const ret = baseGeminiPart();
+  const itemRecord = item as Record<string,unknown>;
+  if ("thought" in itemRecord) {
+    ret.thought = itemRecord.thought as boolean;
+  }
+  if ("thoughtSignature" in itemRecord) {
+    ret.thoughtSignature = itemRecord.thoughtSignature as string;
+  }
   return ret;
 }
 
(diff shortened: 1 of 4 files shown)


libs/providers/langchain-google/src/converters/messages.ts
```diff
@@ -381,6 +381,28 @@ function convertStandardVideoContentBlockToGeminiPart(
...
+  // Reasoning blocks created from Gemini should contain the base
+  // standard block they would be if "thought" wasn't set to true.
+  // If it exists, we'll use that. Otherwise, we'll assume it was
+  // originally a text block.
+  const ret =
+    "reasoningContentBlock" in block && block.reasoningContentBlock
+      ? convertStandardContentBlockToGeminiPart(
+          block.reasoningContentBlock as ContentBlock.Standard
+        )
+      : { text: block.reasoning };
```
hntrl, line 396: I don't think this condition is correct? when does reasoningContentBlock track onto the actual standard block object?

hntrl (approved):

libs/providers/langchain-google/src/converters/messages.ts
afirstenberg: A reasonable question. In libs/langchain-core/src/messages/block_translators/google.ts we determine what the block would be if it wasn't a "thought" block first.
If it isn't a thought block - we return it.
If it _is_, we create a "reasoning" block and set `reasoningContentBlock` to what the original block looked like.

libs/providers/langchain-google/src/converters/messages.ts
```diff
@@ -381,6 +381,28 @@ function convertStandardVideoContentBlockToGeminiPart(
...
+
+  // Then add that this came from the thinking process
+  if (ret) {
+    ret.thought = true;
```
thushanth-bengre-langchain, line 400: Doesn't this set `thought: true` unconditionally, for every reasoning block? Is there a case where a `ContentBlock.Reasoning` shouldn't be flagged as Gemini's own thought — e.g. one from a provider that never sets `thought` at all?

afirstenberg: Ah, that would explain the second test that I flagged.

Is this considered an issue? 
I can't think why `Reasoning` blocks shouldn't be treated as if they were `thought: true` Parts from Gemini.

I'm not sure Gemini does anything with the thinking blocks anyway that are sent back because they're really "thought summaries".

afirstenberg: I guess I would ask how we have a `Reasoning` block if the hypothetical provider didn't in some way mark this as a reasoning / thinking block.

thushanth-bengre-langchain (approved):

PR #11588, opened 2026-09-08
fix(openai): Fix base url deserialization

jacoblee93:
OpenAI manifests pulled from the prompt hub did not respect base url. This fixes this.

### DIFF
--- a/libs/providers/langchain-openai/src/chat_models/base.ts
+++ b/libs/providers/langchain-openai/src/chat_models/base.ts
@@ -228,12 +228,6 @@ export interface BaseChatOpenAIFields
    * Optional configuration options for the OpenAI client.
    */
   configuration?: ClientOptions;
-
-  /**
-   * Base URL for the OpenAI API, also used when loading Python `base_url` fields.
-   * Prefer `configuration.baseURL` in JavaScript; it takes precedence.
-   */
-  baseUrl?: string;
 }
 
 export function getChatOpenAIModelParams<TParams extends BaseChatOpenAIFields>(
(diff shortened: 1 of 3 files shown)


libs/providers/langchain-openai/src/chat_models/base.ts
```diff
@@ -228,6 +228,12 @@ export interface BaseChatOpenAIFields
    * Optional configuration options for the OpenAI client.
    */
   configuration?: ClientOptions;
+
+  /**
+   * Base URL for the OpenAI API, also used when loading Python `base_url` fields.
+   * Prefer `configuration.baseURL` in JavaScript; it takes precedence.
+   */
+  baseUrl?: string;
```
hntrl: does this need to be in types? We have a couple of places where we respect the ser/de logic but don't embed it in types

libs/providers/langchain-openai/src/chat_models/base.ts
```diff
@@ -228,6 +228,12 @@ export interface BaseChatOpenAIFields
    * Optional configuration options for the OpenAI client.
    */
   configuration?: ClientOptions;
+
+  /**
+   * Base URL for the OpenAI API, also used when loading Python `base_url` fields.
```
hntrl: also nit: fine to not mention base_url here IMO

libs/providers/langchain-openai/src/chat_models/base.ts
jacoblee93: Ah yeah not sure. Will try

hntrl (approved):

PR #11589, opened 2026-09-08
fix(langchain): keep structured output tool names stable for prompt caching

### LINKED ISSUE #11151: createAgent structured-output tool is renamed `extract-<n>` on every model call, defeating Anthropic prompt caching

richardscarrott:
> **Alternative to #11600 for #11151. Only one of the two should merge.**
> Same bug, same fix, different generated name. This PR derives the name from
> a hash of the schema; #11600 numbers by position. See the comparison table on
> #11600.

Fixes #11151

### Problem

When a `responseFormat` schema has no title and resolves to the tool strategy rather than the provider's native structured output, the structured-output tool is named from a module-global counter — `extract-1` on the first model request, `extract-2` on the second. Providers render tool definitions at the very front of the cached prompt prefix, so a name that changes between requests invalidates the entire cache, system prompt and message history included.

It is also nondeterministic: the counter is global to the process, so concurrent agents interleave their increments and a restart resets the sequence.

### Measurement

Two runs of the same agent against the live Anthropic API — same system prompt, same tool, same middleware, same number of model requests:

```
before   req#1  tools=[get_weather, extract-1]                 cache_write=10790  cache_read=0
         req#2  tools=[get_weather, extract-2]                 cache_write=10863  cache_read=0

after    req#1  tools=[get_weather, extract-b8db57530ae59966]  cache_write=10797  cache_read=0
         req#2  tools=[get_weather, extract-b8db57530ae59966]  cache_write=73     cache_read=10797
```

### Change

The generated name is now `extract-` followed by a truncated hash of the schema the tool definition is built from, so the same schema always produces the same name: within one agent loop, across turns of a conversation, and across separate processes. The counter has been removed.

- Schemas carrying a title are unaffected, so anyone who adopted the title workaround keeps their current names.
- The `extract-` prefix is retained, since `ReactAgent` and the PII redaction middleware both classify structured-output tool calls by it. A custom title produces a name with no such prefix, so I plan to investigate those call sites separately to see whether a titled schema is problematic there.
- The hash reuses `sha256` from `@langchain/core/utils/hash`, already used by the PII middleware. No new dependency, no Node-only APIs, still synchronous.

One behavioural change: two identical schemas in one `responseFormat` list now collapse to a single tool, the last entry winning. The agent already keys its strategy map by name, so this needs no new code. Previously the two were offered to the model as separate tools differing only in name, and whichever the model called decided which entry's options applied.

No user-facing API changes and no migration. The generated name remains an implementation detail; the escape hatch for a specific name is still the schema title.

### Credit

The root cause was correctly diagnosed by @bisma-nawaz in #11161. This PR takes a different approach — deriving the name from schema content rather than memoising the derived strategies — because the agent node rebuilds the model on every request, and for a model passed as an identifier string that produces a fresh instance each time, so a memo keyed on the model never hits. Content-derived naming is also the only variant stable across processes.

### DIFF
--- a/libs/langchain/src/agents/responses.ts
+++ b/libs/langchain/src/agents/responses.ts
@@ -8,6 +8,7 @@ import {
 } from "@langchain/core/utils/types";
 import { type AIMessage } from "@langchain/core/messages";
 import { toJsonSchema, Validator } from "@langchain/core/utils/json_schema";
+import { sha256 } from "@langchain/core/utils/hash";
 import { type FunctionDefinition } from "@langchain/core/language_models/base";
 import {
   type SerializableSchema,
@@ -41,9 +42,20 @@ export type ResponseFormatUndefined = {
 const PROVIDER_STRATEGY_DEFAULT_STRICT = true;
 
 /**
- * This is a global counter for generating unique names for tools.
+ * Derives the name of the tool used to extract structured output.
+ *
+ * Tools must have a name so a tool call can be mapped back to the strategy that
+ * produced it. When the schema carries no title the name is hashed from the
+ * schema the tool definition is built from, so the same schema always yields
+ * the same name - within an agent loop, across conversation turns, and across
+ * processes.
+ *
+ * @param schema - The schema the tool definition is built from
+ * @param title - The schema's title, used verbatim when present
  */
-let bindingIdentifier = 0;
+function getFunctionName(schema: Record<string, unknown>, title?: string) {
+  return title ?? `extract-${sha256(JSON.stringify(schema)).slice(0, 16)}`;
+}
 
 /**
  * Information for tracking structured output tool metadata.
@@ -95,20 +107,12 @@ export class ToolStrategy<_T = unknown> {
     schema: InteropZodObject | SerializableSchema | Record<string, unknown>,
     outputOptions?: ToolStrategyOptions
   ): ToolStrategy<any> {
-    /**
-     * It is required for tools to have a name so we can map the tool call to the correct tool
-     * when parsing the response.
-     */
-    function getFunctionName(name?: string) {
-      return name ?? `extract-${++bindingIdentifier}`;
-    }
-
     if (isSerializableSchema(schema) || isInteropZodSchema(schema)) {
       const asJsonSchema = toJsonSchema(schema);
       const tool = {
         type: "function" as const,
         function: {
-          name: getFunctionName(asJsonSchema.title),
+          name: getFunctionName(asJsonSchema, asJsonSchema.title),
           strict: false,
           description:
             asJsonSchema.description ??
@@ -128,7 +132,7 @@ export class ToolStrategy<_T = unknown> {
       functionDefinition = schema as unknown as FunctionDefinition;
     } else {
       functionDefinition = {
-        name: getFunctionName(schema.title as string),
+        name: getFunctionName(schema, schema.title as string),
         description: (schema.description as string) ?? "",
         parameters: schema.schema || (schema as Record<string, unknown>),
       };
--- a/.changeset/deterministic-structured-output-tool-naming.md
+++ b/.changeset/deterministic-structured-output-tool-naming.md
@@ -1,0 +1,18 @@
+---
+"langchain": patch
+---
+
+fix(langchain): derive structured output tool names from the schema
+
+An untitled `responseFormat` schema produced a tool named from a module-global
+counter, so the structured output tool was called `extract-1` on one model
+request and `extract-2` on the next. Providers render tool definitions at the
+front of the cached prompt prefix, so the changing name invalidated the prompt
+cache on every request in an agent loop — the system prompt and message history
+included.
+
+The generated name is now a hash of the emitted JSON Schema, so the same schema
+yields the same tool name across every request in a loop, across turns of a
+conversation, and across separate processes. Schemas carrying a title are
+unaffected. Two identical schemas in one `responseFormat` list now collapse to a
+single tool rather than two indistinguishable ones.
--- a/libs/langchain/src/agents/tests/reactAgent.test.ts
+++ b/libs/langchain/src/agents/tests/reactAgent.test.ts
@@ -199,12 +199,14 @@ describe("createAgent", () => {
 
     type WeatherResponse = z.infer<typeof WeatherResponseSchema>;
 
+    const [structuredTool] = toolStrategy(WeatherResponseSchema);
+
     const toolCalls = [
       [
         {
           args: { temperature: 75 },
           id: "2",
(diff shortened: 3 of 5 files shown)


libs/langchain/src/agents/responses.ts
```diff
@@ -41,9 +42,20 @@ export type ResponseFormatUndefined = {
...
+ * the same name - within an agent loop, across conversation turns, and across
+ * processes.
+ *
+ * @param schema - The schema the tool definition is built from
+ * @param title - The schema's title, used verbatim when present
  */
-let bindingIdentifier = 0;
+function getFunctionName(schema: Record<string, unknown>, title?: string) {
+  return title ?? `extract-${sha256(JSON.stringify(schema)).slice(0, 16)}`;
+}
```
hntrl, line 58: Am not totally sold this is the way to fix. Attaching shas to tools are fine but my intuition is that simpler, non-obfuscated tool names fare better. I think a determining order for 
* if a zod object/ json schema has a title already (e.g. `.meta({ title })`), we attach that as the suffix of the tool name
* we can expect that `responseFormat` wont change between conversation turns, so for each response format object we get we just keep a monotonic counter and thats what we add as suffix to the tool name

richardscarrott: Yeah, I wasn't entirely sure about it initially either, but I settled on hashing for a few reasons:

1. The name is deterministic across instances and processes, which keeps it stable for prompt caching across follow-up turns.
2. It's an uncommon case, but the Python SDK can use a uuid suffix for the name, `response_format_{uuid4[:4]}`, so there is some precedent for more obscure names.

That said, I think the array-index approach you mentioned offline preserves the same stability property. For example:

```ts
const agent = createAgent({
  // ...
  responseFormat: [InvoiceSchema, SupportTicketSchema], // [extract-1, extract-2], always
});
```

The only issue is that the JS SDK allows callers to compose strategies separately:

```ts
const [invoiceStrategy] = toolStrategy(InvoiceSchema); // extract-1
const [supportTicketStrategy] = toolStrategy(SupportTicketSchema); // extract-1

const agent = createAgent({
  // ...
  responseFormat: [invoiceStrategy, supportTicketStrategy], // would need to be rewritten to extract-1, extract-2
});
```

I think that name rewrite is maybe acceptable but tbh I still lean _slightly_ toward hashing because each strategy receives its final name when callers create it.

Having said that, I'm unsure how much `extract-1` vs `extract-b8db57530ae59966` affects model behaviour — If readable names materially improve tool selection, that matters more than avoiding the name rewrite, since the model reliably terminating the loop with the extract tool is of course critical.

My hunch is the description and input schema does the heavy lifting more than the tool name as both extract-n and extract-{hash} are pretty meaningless tbh...

We could consider truncating the hash from 64-bit to 32-bit, giving us `extract-b8db5753`. The Python UUID suffix is 16-bit, I think. 🤔

richardscarrott: @hntrl I've pushed up an alternative impl. keeping the `extract-n` naming convention as discussed yesterday — https://github.com/langchain-ai/langchainjs/pull/11600

There's also a comparison table in that PR between the hash vs index-based impl so we can pick which is best.

PR #11602, opened 2026-09-10
fix(google): preserve tool call id and thoughtSignature in native streaming

thushanth-bengre-langchain:
`convertGoogleGeminiStream` (the native `streamEvents()` converter for `@langchain/google`) dropped both the tool call `id` and Gemini's `thoughtSignature` on every streamed function call.

Fixes:

* [#11181](https://github.com/langchain-ai/langchainjs/issues/11181)
* [#9624](https://github.com/langchain-ai/langchainjs/issues/9624)
* [#11261](https://github.com/langchain-ai/langchainjs/issues/11261)

`thoughtSignature` is preserved on the tool call's content block (its correct home) and read back from there on replay, by tool call id — no `@langchain/core` changes needed.

### DIFF
--- a/libs/langchain-core/src/messages/ai.ts
+++ b/libs/langchain-core/src/messages/ai.ts
@@ -148,6 +148,9 @@ export class AIMessage<TStructure extends MessageStructure = MessageStructure>
               id: toolCall.id,
               name: toolCall.name,
               args: toolCall.args,
+              ...("thoughtSignature" in toolCall
+                ? { thoughtSignature: toolCall.thoughtSignature }
+                : {}),
             }))
           );
         }
@@ -172,6 +175,9 @@ export class AIMessage<TStructure extends MessageStructure = MessageStructure>
               id: block.id!,
               name: block.name,
               args: block.args as Record<string, unknown>,
+              ...("thoughtSignature" in block
+                ? { thoughtSignature: block.thoughtSignature }
+                : {}),
             })),
           ] as $InferToolCalls<TStructure>[];
         }
--- a/libs/langchain-core/src/messages/ai.ts
+++ b/libs/langchain-core/src/messages/ai.ts
@@ -148,9 +148,6 @@ export class AIMessage<TStructure extends MessageStructure = MessageStructure>
               id: toolCall.id,
               name: toolCall.name,
               args: toolCall.args,
-              ...("thoughtSignature" in toolCall
-                ? { thoughtSignature: toolCall.thoughtSignature }
-                : {}),
             }))
           );
         }
@@ -175,9 +172,6 @@ export class AIMessage<TStructure extends MessageStructure = MessageStructure>
               id: block.id!,
               name: block.name,
               args: block.args as Record<string, unknown>,
-              ...("thoughtSignature" in block
-                ? { thoughtSignature: block.thoughtSignature }
-                : {}),
             })),
           ] as $InferToolCalls<TStructure>[];
         }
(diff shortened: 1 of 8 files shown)


libs/providers/langchain-google/src/chat_models/tests/index.int.test.ts
```diff
@@ -1709,6 +1709,21 @@ describe.each(thinkingModelInfo)(
       expect(hasThoughtSignature).toBe(true);
     });
 
+    test("thought signature - function via streamEvents", async () => {
```
afirstenberg: See also the integration test in #11262 which should be included (although I'm still  trying to validate it).

thushanth-bengre-langchain: Ported 2 tests from #11262 here.

Thank you.

libs/langchain-core/src/messages/ai.ts
hntrl: I am hard pressed to say the fix is in changing core message logic

libs/langchain-core/src/messages/ai.ts
```diff
@@ -148,6 +148,9 @@ export class AIMessage<TStructure extends MessageStructure = MessageStructure>
               id: toolCall.id,
               name: toolCall.name,
               args: toolCall.args,
+              ...("thoughtSignature" in toolCall
```
afirstenberg: We're making a core change for a platform-specific attribute? Or do other platforms also use "thoughtSignature"?

libs/langchain-core/src/messages/ai.ts
```diff
@@ -172,6 +175,9 @@ export class AIMessage<TStructure extends MessageStructure = MessageStructure>
               id: block.id!,
               name: block.name,
               args: block.args as Record<string, unknown>,
+              ...("thoughtSignature" in block
```
afirstenberg: See above about core change for platform-specific attribute

afirstenberg (commented):
I don't like the idea of platform attributes in code.
Is there a reason we're not using a spread here?

libs/langchain-core/src/messages/ai.ts
thushanth-bengre-langchain: Was able to do this without touching core.

libs/langchain-core/src/messages/ai.ts
thushanth-bengre-langchain: Was able to do this without touching core.

libs/providers/langchain-google/src/utils/stream_events.ts
```diff
@@ -123,33 +126,42 @@ export async function* convertGoogleGeminiStream(
       } else if (part.functionCall) {
         const key: BlockKey = `tool:${toolIdx}`;
         const args = JSON.stringify(part.functionCall.args ?? {});
+        // Only used to seed a *new* block; an id already assigned to this
+        // block must not change on later chunks.
+        const candidateId =
+          part.functionCall.id ?? `lc-tool-call-${uuidv4().replace(/-/g, "")}`;
```
hntrl, line 132: @afirstenberg do you remember why we needed this? for the life of me I can't remember

hntrl (approved):

libs/providers/langchain-google/src/utils/stream_events.ts
afirstenberg: It depends which "this" we're talking about.

If I remember:
* Something in langchain indexed off the function call ID. (I don't remember what.)
* Gemini wasn't always providing one.
* We needed to make sure that if we generated one it was **not** sent to Gemini, because then Gemini would reject it (which is why there is an "lc-tool-call-" prefix), but that we still needed to keep it around for future turns, perhaps?

As for why the UUID without the dashes... no clue.

Issue #10470, opened 2026-03-19
bug(@langchain/google): `timeout` isn't respected on model `invoke`, `AbortSignal` isn't passed to `fetch`


vjancik:
### Checked other resources

- [x] This is a bug, not a usage question. For questions, please use the LangChain Forum (https://forum.langchain.com/).
- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain.js documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain.js rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

# Bug: `ChatGoogle.invoke()` ignores `RunnableConfig.timeout` — non-streaming path drops the AbortSignal

**Package:** `@langchain/google`
**Version confirmed:** `0.1.6`, `0.1.7` (with `@langchain/core` up to `1.1.34`)
**Affects:** non-streaming `.invoke()` calls (i.e. `streaming: false`, which is the default)
**Does NOT affect:** streaming path (`.stream()` / `streaming: true`)

---

## Summary

Passing `{ timeout: N }` to `ChatGoogle.invoke()` has no effect. If the Gemini API stalls, the call hangs indefinitely regardless of the configured timeout. The abort signal that `@langchain/core` creates from the timeout value is fully formed and reachable inside `_generate`, but it is silently dropped when the library constructs the `Request` object for the `generateContent` fetch call.

---

## Root cause — exact location

`@langchain/core` converts `RunnableConfig.timeout` into an `AbortSignal` inside `ensureConfig` (`runnables/config.js`):

```js
// @langchain/core — runnables/config.js
const timeoutSignal = AbortSignal.timeout(originalTimeoutMs);
// merged with any existing signal, then stored as config.signal
empty.signal = timeoutSignal; // (or AbortSignal.any([...]) if there was already a signal)
delete empty.timeout;         // timeout is consumed here; signal is the sole carrier from now on
```

This signal is then passed as `options.signal` all the way down the runnable chain. `@langchain/core` also wraps the `_generate` return value with `raceWithSignal(promise, options.signal)`, which will reject the outer promise as soon as the signal fires — **but only if the inner fetch is already settled or the signal actually fires**.

[...]

PR #11604, opened 2026-09-10
fix(google-genai): forward abort signal to non-streaming generateContent

### LINKED ISSUE #11388: ChatGoogleGenerativeAI: AbortSignal never reaches fetch in non-streaming _generate (options not forwarded to completionWithRetry)
### Checked other resources

- I searched existing issues for `google-genai` + `AbortSignal`/`completionWithRetry` and found no report of this.

### Package (exact versions verified)

`@langchain/google-genai` **2.1.14** and **2.2.0** (latest at time of writing — bug present in both, verified against the published tarballs).

### The bug

In the **non-streaming** path, `_generate` calls `completionWithRetry` with a **single argument**, dropping the `options` object that carries the caller's `AbortSignal`:

```js
// dist/chat_models.cjs:554 (2.2.0); :542 in 2.1.14
const res = await this.completionWithRetry(this._buildGenerateContentRequest(messages, options));
```

The receiving side is already wired correctly — it forwards the signal to both the retry caller and the SDK fetch **when it receives options**:

```js
// dist/chat_models.cjs:614 (2.2.0); same in 2.1.14
async completionWithRetry(request, options) {
    return this.caller.callWithOptions({ signal: options?.signal }, async () => {
        return await this.client.generateContent(request, { signal: options?.signal });
        ...
```

Since `options` is never passed, only the pre-flight `options.signal?.throwIfAborted()` in `_generate` has any effect. Once the request is in flight, **the signal can never cancel it**: `AbortSignal.timeout(...)` / manual aborts on `invoke()` are silently ignored for this provider in non-streaming mode, and the underlying fetch runs (and bills) to completion. We observed a production call hanging for ~995s with a live abort signal attached.

The **streaming** path is unaffected — `_streamResponseChunks(messages, options, ...)` forwards `options` properly.

### Reproduction sketch

```ts
const model = new ChatGoogleGenerativeAI({ model: "gemini-2.5-flash", apiKey });
// expected: rejects after ~1ms; actual: resolves normally after full generation
await model.invoke([new HumanMessage("write a long story")], { signal: AbortSignal.timeout(1) });
```

(The pre-flight throwIfAborted only fires if the signal is *already* aborted before the call starts; a timeout that expires mid-request never cancels anything.)

### Suggested fix (one line)

Forward the options at the `_generate` call site:

```js
const res = await this.completionWithRetry(this._buildGenerateContentRequest(messages, options), options);
```

The receiver already handles everything else correctly.

cpruijsen:
Non-streaming `ChatGoogleGenerativeAI._generate` now passes `options` as the second argument to `completionWithRetry`, so the caller's `AbortSignal` reaches `generateContent`, matching the streaming path and what `completionWithRetry` already implements (`callWithOptions` plus SDK `{ signal }`). `_generate` previously called `completionWithRetry` with only the request. Pre-flight `options.signal?.throwIfAborted()` still ran, but a timeout or abort after the request started was ignored and the fetch ran (and billed) to completion.

Related: #10501 (open, conflicts with main, same one-line fix) and closed unmerged #11401. Same class of bug as #10470 / #10471 in `@langchain/google`.

Streaming `_streamChatModelEvents` / `_streamResponseChunks` are unchanged.

Fixes #11388

## Decision

The unit test spies `completionWithRetry` and asserts it receives `{ signal }` (same as #11401 / #10501). The alternative is to mock `client.generateContent`, abort after the call starts, and assert the same `AbortSignal` object reached the SDK (what #10471 does for `ChatGoogle`, and what was offered on this issue). `completionWithRetry` already forwards `options?.signal` to both the retry caller and `generateContent`, so the spy is the smallest reversible check. Can switch to the generateContent test.

## Test plan

- [x] Unit test: `invoke` with `{ signal }` forwards that signal to `completionWithRetry` (fake API key, no Gemini network)
- [x] Test fails without the one-line pass-through and passes with it
- [x] Existing `chat_models.test.ts` suite (45 tests) passes
- [ ] Optional live check: `invoke(..., { signal: AbortSignal.timeout(1) })` against Gemini

### DIFF
--- a/.changeset/google-genai-abort-signal.md
+++ b/.changeset/google-genai-abort-signal.md
@@ -1,0 +1,5 @@
+---
+"@langchain/google-genai": patch
+---
+
+fix(google-genai): forward abort signals to non-streaming generateContent calls
--- a/libs/providers/langchain-google-genai/src/chat_models.ts
+++ b/libs/providers/langchain-google-genai/src/chat_models.ts
@@ -933,7 +933,8 @@ export class ChatGoogleGenerativeAI
     }
 
     const res = await this.completionWithRetry(
-      this._buildGenerateContentRequest(messages, options)
+      this._buildGenerateContentRequest(messages, options),
+      options
     );
 
     let usageMetadata: UsageMetadata | undefined;
--- a/libs/providers/langchain-google-genai/src/tests/chat_models.test.ts
+++ b/libs/providers/langchain-google-genai/src/tests/chat_models.test.ts
@@ -139,10 +139,12 @@ test("Google AI - passes system instruction and response schema per request", as
     },
   } as unknown as Schema;
   const requests: GenerateContentRequest[] = [];
-  vi.spyOn(model, "completionWithRetry").mockImplementation(async (request) => {
-    requests.push(request as GenerateContentRequest);
-    return { response: mockGenerateContentResponse("ok") } as never;
-  });
+  vi.spyOn(model, "completionWithRetry").mockImplementation(
+    async (request, _options) => {
+      requests.push(request as GenerateContentRequest);
+      return { response: mockGenerateContentResponse("ok") } as never;
+    }
+  );
 
   await model.invoke(
     [new SystemMessage("First system instruction"), new HumanMessage("Hello")],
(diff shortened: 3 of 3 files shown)


cpruijsen:
Closing this one. I saw #11608 closed in favour of the unified `@langchain/google` package, so I checked whether the same gap exists there before assuming this was worth porting.

It does not. `ChatGoogle._generate` builds a `Request` and passes the abort signal into `fetch`, so the unified package already cancels a non-streaming call. The defect here is real but only in `langchain-google-genai`, which is the package being consolidated away, so fixing it is not worth your review time.

Happy to look at something in the active packages instead.

PR #11608, opened 2026-09-11
fix(google-genai): rewrite list-valued JSON Schema type for Gemini

### LINKED ISSUE #11328: ChatGoogleGenerativeAI: withStructuredOutput sends a list-valued `type` to responseSchema, which the proto rejects (400)
### Checked other resources

- [x] I searched GitHub issues for this and found no existing report (0 hits for `Proto field is not repeating`, 0 for `nullable responseSchema`). #7758 is adjacent — same neighbourhood (nullability through the Gemini structured-output path) but a different symptom and mechanism; it is about an `anyOf`-with-null branch binding incorrectly, this is a hard 400 before the model runs.
- [x] Reproduced against `@langchain/google-genai@2.2.0`.

### Summary

`withStructuredOutput()` sends a list-valued `type` (`{"type": ["string","null"]}`) straight into Gemini's `responseSchema` field, which is a proto message with no repeated `type` field. Every request 400s. It is triggered by an ordinary `z.string().nullable()` on zod 3 + `zod-to-json-schema`.

### Repro

```ts
import { z } from "zod";
import { schemaToGenerativeAIParameters } from "@langchain/google-genai/dist/utils/zod_to_genai_parameters.js";

const S = z.object({ nickname: z.string().nullable(), age: z.number() });
console.log(JSON.stringify(schemaToGenerativeAIParameters(S)));
```

Output (verbatim):

```json
{"type":"object","properties":{"nickname":{"type":["string","null"]},"age":{"type":"number"}},"required":["nickname","age"]}
```

`chat_models.js:676` then does `this.withConfig({ responseSchema: jsonSchema })`, so that document is assigned to `generationConfig.responseSchema`.

### The API's response

Google validates the payload **before** auth, so this needs no valid key — a dummy one returns the real schema verdict:

```
POST https://generativelanguage.googleapis.com/v1beta/models/gemini-2.0-flash:generateContent?key=DUMMY
{"contents":[{"parts":[{"text":"hi"}]}],
 "generationConfig":{"responseMimeType":"application/json",
   "responseSchema":{"type":"object","properties":{"nickname":{"type":["string","null"]},"age":{"type":"number"}},"required":["nickname","age"]}}}
```

```
Invalid JSON payload received. Unknown name "type" at
'generation_config.response_schema.properties[0].value':
Proto field is not repeating, cannot start list.
```

**Control**, same schema with the proto's own nullability spelling, in the same request shape — reaches `API key not valid`, i.e. it passes payload validation:

```json
{"nickname":{"type":"string","nullable":true},"age":{"type":"number"}}
```

So this is not a general "Gemini dislikes the schema" — it is precisely the list-valued `type`.

### Cause

`schemaToGenerativeAIParameters` (`utils/zod_to_genai_parameters.ts`) is `removeAdditionalProperties(...)` plus a `$schema` destructure. It removes exactly two keys and forwards everything else. `responseSchema` is the narrow `Schema` proto, whose `type` is a single-valued enum, so a JSON Schema union type cannot survive it — but nothing in the path rewrites it.

### Suggested fix

`nullable` is a real field of the `Schema` proto (verified accepted above), and it is the standard spelling for this:

- `{"type": ["X","null"]}` → `{"type": "X", "nullable": true}`
- `{"type": ["X","Y"]}` (no null) → `{"anyOf":[{"type":"X"},{"type":"Y"}]}` — `anyOf` is a proto field and is accepted (verified).

Worth noting that **every other Gemini client I measured already does this**, so the behaviour here is the outlier rather than the norm — `google-adk` 2.6.3, `agno` 2.8.7, `@ai-sdk/google` 4.0.39 and `litellm` 1.96.0 all rewrite a list-valued `type` before sending. Measured 2026-08-10, each on the real converter output (JS via an intercepting `fetch` on the wire payload).

### Two things that would make a regression test actually fail

1. **A zod 4 fixture passes either way.** zod 4's native `z.toJSONSchema()` emits `{"anyOf":[{"type":"string"},{"type":"null"}]}` for `.nullable()`, and `anyOf` *is* a proto field, so it never reaches this bug. The failing shape comes from zod 3 + `zod-to-json-schema`, and from any hand-written / OpenAPI-3.1 JSON Schema passed to `withStructuredOutput` directly. A test written on zod 4 will be green while the bug ships.
2. **Assert on the emitted document, not on a mocked call succeeding.** The rejection happens server-side in payload validation, so a test with a stubbed transport cannot see it. Asserting that no `type` value is an array is enough, and the dummy-key request above makes it checkable in CI with no credentials.

### Scope

I have not used a real API key, so nothing here is a claim about how the model *behaves* with the corrected schema — only about which documents the endpoint will accept.

cpruijsen:
Fixes #11328

`schemaToGenerativeAIParameters` now rewrites list-valued JSON Schema `type` before `ChatGoogleGenerativeAI.withStructuredOutput()` (default `jsonSchema` path) sends it as Gemini `responseSchema`. The same sanitizer also feeds tool `parameters`.

Gemini's proto `type` is a single enum, not a repeating field. A JSON Schema array such as `["string","null"]` 400s before auth:

`Proto field is not repeating, cannot start list.`

The usual trigger is Zod 3 `.nullable()` on an unchecked primitive (`zod-to-json-schema` emits a type array). Zod 4 `.nullable()` emits `anyOf` instead, so a Zod 4 fixture would not catch this.

The reporter's control, `{"type":"string","nullable":true}`, is accepted (dummy-key request reaches `API key not valid`). Reproduced here the same way.

Before this, `schemaToGenerativeAIParameters` only stripped `$schema` / `additionalProperties` / `strict` and forwarded every other keyword. `@langchain/google-common` and `@langchain/google` already rewrite list-valued `type` in `adjustObjectType`; `@langchain/google-genai` did not.

This ports that helper onto google-genai's existing sanitizer (one call inside `removeAdditionalProperties`):

- `["X","null"]` → `{type:"X", nullable:true}`
- `["X"]` → `"X"`
- `["null","null"]` (Zod 3 `z.null().nullable()`) → `{type:"null", nullable:true}` rather than a typeless `{nullable:true}`
- `["null"]` and non-null unions such as `["string","number"]` throw, matching the sibling packages

Chose the google-common throws for `["null"]` and `["X","Y"]`. Alternative, from the reporter's follow-up: unwrap `["null"]` to `"null"` and rewrite `["X","Y"]` to `anyOf` (both accepted by the proto). Reasoning: match the repo's existing helper rather than make google-genai more permissive than google-common / `@langchain/google`. Happy to switch.

Did not copy google-common's `anyOf` rejection. google-genai already forwards `anyOf` (Zod 4 `.nullable()`), and `anyOf` is a real proto field.

Unit tests assert the emitted document (Zod 3 fixtures) and that no remaining `type` value is an array. They fail on the unfixed converter and pass with this change. Dummy-key requests confirm the rewritten documents pass proto validation.

This is not a claim about how the model behaves with the corrected schema.

### DIFF
--- a/.changeset/fix-google-genai-list-valued-type.md
+++ b/.changeset/fix-google-genai-list-valued-type.md
@@ -1,0 +1,5 @@
+---
+"@langchain/google-genai": patch
+---
+
+fix(google-genai): rewrite JSON Schema list-valued `type` (e.g. `["string","null"]` from Zod 3 `.nullable()`) into Gemini's single-valued `type` plus `nullable` before sending `responseSchema`
--- a/libs/providers/langchain-google-genai/src/utils/tests/zod_to_genai_parameters.test.ts
+++ b/libs/providers/langchain-google-genai/src/utils/tests/zod_to_genai_parameters.test.ts
@@ -1,0 +1,133 @@
+import { describe, expect, test } from "vitest";
+import { z } from "zod/v3";
+import {
+  jsonSchemaToGeminiParameters,
+  schemaToGenerativeAIParameters,
+} from "../zod_to_genai_parameters.js";
+
+function pathsWithArrayValuedType(node: unknown, path = "$"): string[] {
+  if (Array.isArray(node)) {
+    return node.flatMap((item, i) =>
+      pathsWithArrayValuedType(item, `${path}[${i}]`)
+    );
+  }
+  if (node === null || typeof node !== "object") {
+    return [];
+  }
+  const obj = node as Record<string, unknown>;
+  const hits = Array.isArray(obj.type) ? [path] : [];
+  return hits.concat(
+    Object.entries(obj).flatMap(([key, value]) =>
+      pathsWithArrayValuedType(value, `${path}.${key}`)
+    )
+  );
+}
+
+describe("schemaToGenerativeAIParameters", () => {
+  test("rewrites Zod 3 z.string().nullable() list-valued type to nullable string", () => {
+    const schema = z.object({
+      nickname: z.string().nullable(),
+      age: z.number(),
+    });
+    const result = schemaToGenerativeAIParameters(schema);
+
+    expect(result.properties?.nickname).toMatchObject({
+      type: "string",
+      nullable: true,
+    });
+    expect(result.properties?.age).toMatchObject({ type: "number" });
+    expect(result.required).toEqual(["nickname", "age"]);
+    expect(pathsWithArrayValuedType(result)).toEqual([]);
+  });
+
+  test("rewrites Zod 3 z.null().nullable() ['null','null'] instead of dropping type", () => {
+    const schema = z.object({
+      n: z.null().nullable(),
+    });
+    const result = schemaToGenerativeAIParameters(schema);
+
+    expect(result.properties?.n?.type).toBe("null");
+    expect(result.properties?.n?.nullable).toBe(true);
+    expect(pathsWithArrayValuedType(result)).toEqual([]);
+  });
+
+  test("rewrites nullable items nested under arrays", () => {
+    const schema = z.object({
+      tags: z.array(z.string().nullable()),
+    });
+    const result = schemaToGenerativeAIParameters(schema);
+
+    expect(result.properties?.tags).toMatchObject({
+      type: "array",
+      items: { type: "string", nullable: true },
+    });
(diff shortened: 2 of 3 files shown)


thushanth-bengre-langchain:
Thank you for opening the PR @cpruijsen.

We are focusing on having a stable unified `@langchain/google` package.
We fixed this issue in that package in PR #11606.

Closing this PR.

PR #11605, opened 2026-09-10
feat(mcp-adapters): add MCPAdapter API and consolidate Zod4 boundaries

byhow:
## What changes?

Add `MCPAdapter({ servers })` and `listTools()` as the canonical API. Replace `getTools()` with `listTools()` for the major release; keep `MultiServerMCPClient` and `mcpServers` as deprecated aliases.

Composed Zod 4 schemas validate server configuration before connecting. This PR defaults to modern MCP; legacy HTTP, stdio, and SSE servers use `mode: "legacy"`. Notification and progress callbacks belong to individual servers, and initialization and SSE fallback are legacy-only. Configuration snapshots preserve callback and OAuth provider identity while copying mutable options.

Use SDK schemas for protocol payloads and SDK validation for tool arguments after hooks run. Preserve server tool schemas, keep model-facing overrides separate from invocation validation, and expose branded errors with original causes and structured Zod issues. Tool results use standard LangChain content blocks; remove `useStandardContentBlocks`.

Keep a concise package README and runnable examples. Remove package-local reference and migration guides for relocation to `langchain-ai/docs`.

## How was it tested?

- 286 tests passed, with four existing skips, including real modern/legacy HTTP behavior, tool schema constraints, and hook artifact preservation.
- Source, examples, and modified test files typechecked.
- ESM/CommonJS builds, attw, publint, lint, formatting, and whitespace checks passed.

## Scope

Builds on #11597. Discovery and content fidelity follow in #11612; protocol negotiation and server interactions in #11614; invocation and checkpointed elicitation in #11642–#11645. OAuth workflow enhancements are outside this stack. No CI changes.

PR #11612, opened 2026-09-11
feat(mcp-adapters): preserve tool fidelity and isolate connection catalogs

byhow:
## What changes?

Preserve structured output, resource provenance, and protocol metadata in tool artifacts. Native `ToolMessage`, `Command`, and graph-interrupt behavior survive hooks and conversion. Resource links are not fetched implicitly.

Add `listToolsets()` for executable tools grouped by server; retain `initializeConnections()` as a deprecated wrapper. `listTools([], { cacheMode: "use" | "refresh" | "bypass" })` delegates freshness and pagination to the SDK. Unchanged descriptors reuse adapted tools, and previously returned tools remain unchanged.

Isolate connections and catalogs by effective headers and OAuth provider identity. Deduplicate concurrent connections, release connections after failed setup, and attempt every close during shutdown. Remove tool-level debug logging while preserving errors and MCP server logging APIs.

Keep runnable examples with the package. Detailed discovery, reference, and migration material belongs in `langchain-ai/docs`; remove the package guides and leave the README unchanged from #11605.

## How was it tested?

- 314 tests passed, with two existing skips, including real SDK 1.30 stdio alongside SDK 2 HTTP/SSE servers, artifacts, graph controls, connection isolation, cache behavior, and lifecycle races.
- Source, examples, and modified test files typechecked.
- ESM/CommonJS builds, attw, publint, lint, formatting, and whitespace checks passed.

`@modelcontextprotocol/sdk ^1.30.0` is a development dependency used only by the legacy server fixture. This verifies compatibility with an actual SDK 1 server implementation; the shipped adapter uses the SDK 2 client/core packages.

## Scope

Stacked on #11605, which owns configuration schemas, tool argument validation, standard content defaults, and branded errors. Protocol negotiation and server interactions follow in #11614; invocation and checkpointed elicitation follow in #11642–#11645. OAuth workflow enhancements are outside this stack. No CI changes.


libs/langchain-mcp-adapters/src/tools.ts
```diff
@@ -39,9 +38,9 @@ import {
 } from "./types.js";
 import type { ToolHooks } from "./hooks.js";
 import type { Client } from "./connection.js";
-import { getDebugLog } from "./logging.js";
+import debug from "debug";
```
hntrl, line 40: do we need debug logging?

libs/langchain-mcp-adapters/package.json
```diff
@@ -57,6 +57,7 @@
     "@langchain/openai": "workspace:*",
     "@langchain/tsconfig": "workspace:*",
     "@modelcontextprotocol/node": "^2.0.0",
+    "@modelcontextprotocol/sdk": "^1.30.0",
```
hntrl, line 60: ?

libs/langchain-mcp-adapters/README.md
```diff
@@ -73,6 +73,9 @@ Construction validates options with Zod 4 and opens no connections. Discovery
 and invocation open connections as needed. Use `listTools("serverName")` to
 select tools and always await `close()` when finished.
 
+Use `listToolsets()` for a map of server names to tools. Both discovery methods
```
hntrl, line 76: docs should live in langchain-ai/docs I imagine

libs/langchain-mcp-adapters/src/tools.ts
byhow: this is for parity but can remove / removed

libs/langchain-mcp-adapters/package.json
byhow: integration test on regression and can remove if a bit too much

byhow: this is only for dev dep

libs/langchain-mcp-adapters/src/tools.ts
hntrl: yeah lets

libs/langchain-mcp-adapters/package.json
hntrl: ah ok nvm

libs/langchain-mcp-adapters/src/client.ts
```diff
@@ -57,7 +60,15 @@ export class MCPAdapter {
   /**
    * Cached map of server names to tools
    */
-  #serverNameToTools: Record<string, DynamicStructuredTool[]> = {};
+  #generation = 0;
+  #closing = false;
```
hntrl: can you explain why we need this lifecycle tracking?

libs/langchain-mcp-adapters/src/client.ts
```diff
@@ -523,11 +549,16 @@ export class MCPAdapter {
    * Close all connections.
    */
   async close(): Promise<void> {
-    debugLog(`INFO: Closing all MCP connections...`);
-    this.#serverNameToTools = {};
-    this.#failedServers.clear();
-    await this.#clientConnections.delete();
-    debugLog(`INFO: All MCP connections closed`);
+    this.#generation += 1;
```
hntrl: is the implication from incrementing generation that I can later restore an MCPAdapter using new clients?

libs/langchain-mcp-adapters/src/tools.ts
```diff
@@ -237,24 +227,18 @@ async function _convertCallToolResult({
     );
   }
 
-  const convertedContent: (ContentBlock | ContentBlock.Multimodal.Standard)[] =
-    (
-      await Promise.all(
```
hntrl, line 242: lol why was this async in the first place

byhow: lol, this is from origin/main

byhow: updated to just flatmap

libs/langchain-mcp-adapters/src/client.ts
byhow: this is not right, we don't need to track life cycle here actually

libs/langchain-mcp-adapters/src/connection.ts
```diff
@@ -66,6 +63,37 @@ const transportTypes = ["http", "sse", "stdio"] as const;
  */
 export class ConnectionManager {
   #connections: Map<ClientKeyObject, Connection> = new Map();
+  #identities: ClientKeyObject[] = [];
+  #pending = new Map<ClientKeyObject, Promise<Client>>();
+  #closing?: Promise<void>;
+
+  constructor(
+    private readonly onToolsChanged?: (options: TransportOptions) => void
+  ) {}
```
hntrl, line 72: positional kwargs like this feels bad, can we make this a params object?

hntrl (dismissed):

hntrl (approved):

PR #11613, opened 2026-09-11
fix(core): convert every media content block in _formatForTracing

### LINKED ISSUE #11291: _formatForTracing only converts the first base64 content block per message

Lubaoshuai:
## Description

`_formatForTracing` in `@langchain/core` only converted the **first** URL/base64 media block of each message: the clone guard (`if (messageToTrace === message)`) also gated the per-block conversion, so after the first block was rewritten the guard became false and every later media block stayed inline in the traced payload — only the first attachment got offloaded to a data URI, and remaining base64 stayed in traces (e.g. when a message carries two PDF attachments, as reported in #11291).

This change clones the message once when any media block is present and converts **every** matching block. Messages without media blocks keep their original reference, so behavior is unchanged for the common case. As before, this affects the tracing copy only — providers still receive the original message.

Fixes #11291

## Test

- Added a regression test in `libs/langchain-core/src/language_models/tests/chat_models.test.ts` that traces a `HumanMessage` with two base64 file blocks and asserts both are rewritten to `image_url` data URIs and that the original message is not mutated. It fails on `main` (second block left unconverted) and passes with this change.
- `vitest run src/language_models/tests/chat_models.test.ts` in `libs/langchain-core`: 25 passed, no type errors.

### DIFF
--- a/.changeset/quiet-donkeys-trace-media.md
+++ b/.changeset/quiet-donkeys-trace-media.md
@@ -1,0 +1,5 @@
+---
+"@langchain/core": patch
+---
+
+fix(core): convert every media content block in `_formatForTracing`, not only the first, so multimodal messages trace completely
--- a/libs/langchain-core/src/language_models/chat_models.ts
+++ b/libs/langchain-core/src/language_models/chat_models.ts
@@ -172,22 +172,20 @@ function _formatForTracing(messages: BaseMessage[]): BaseMessage[] {
   for (const message of messages) {
     let messageToTrace = message;
     if (Array.isArray(message.content)) {
-      for (let idx = 0; idx < message.content.length; idx++) {
-        const block = message.content[idx];
-        if (isURLContentBlock(block) || isBase64ContentBlock(block)) {
-          if (messageToTrace === message) {
-            // Also shallow-copy content
-            // oxlint-disable-next-line @typescript-eslint/no-explicit-any
-            messageToTrace = new (message.constructor as any)({
-              ...messageToTrace,
-              content: [
-                ...message.content.slice(0, idx),
-                convertToOpenAIImageBlock(block),
-                ...message.content.slice(idx + 1),
-              ],
-            });
-          }
-        }
+      const hasMediaBlock = message.content.some(
+        (block) => isURLContentBlock(block) || isBase64ContentBlock(block),
+      );
+      if (hasMediaBlock) {
+        // Clone once and convert every media block, not just the first
+        // oxlint-disable-next-line @typescript-eslint/no-explicit-any
+        messageToTrace = new (message.constructor as any)({
+          ...message,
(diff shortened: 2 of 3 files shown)


jackjin1997:
Thanks for the clear reproduction. This is the same #11291 fix already covered by #11292, which has been open since August 3 and changes the same function/test path. #11292 also includes the required changeset, avoids an `any` in the callback test, passed the full local core suite, and the author of the other duplicate #11450 has already agreed to consolidate there. Could we keep review on #11292 and close this duplicate unless there is a genuinely distinct case to carry over?

jackjin1997:
Update: with this becoming the fifth related implementation for #11291, I have closed #11292 to reduce the review queue rather than keep asking for consolidation. Its changeset, typed callback regression, and full local core validation remain available if maintainers want that coverage carried into the surviving PR.

Lubaoshuai:
Thanks @jackjin1997 — makes sense. Since #11292 is closed, I've added the missing changeset here so this one is complete against the bot requirements. If maintainers want the typed callback regression from your branch carried in, happy to review a PR to this branch or cherry-pick it myself — whatever keeps the review queue shortest.

Lubaoshuai:
Done — the capture handler now types `_llm` as `Serialized` (same shape as the typed regression in #11292), so the test carries no `any` anymore. Kept the extra assertion that tracing doesn't mutate the original message, and the changeset is already in place. Happy for this branch to be the surviving PR for #11291 — it's ready for review.

PR #11614, opened 2026-09-11
feat(mcp-adapters): scope interactions to modern and legacy servers

byhow:
## What changes?

Omitting `mode` uses the SDK's automatic protocol negotiation, so one adapter can connect to modern and legacy servers without callers identifying their protocol ahead of time. Explicit `mode: "modern"` pins the modern revision with no fallback. Explicit `mode: "legacy"` uses the legacy handshake and enables legacy callback options in the configuration types.

SSE uses the legacy handshake directly. Automatic HTTP negotiation can fall back to SSE on 404/405 responses; authentication and server failures do not trigger transport fallback.

Legacy servers accept per-server `onElicitation` callbacks for form and URL input. Answers are parsed with SDK validators and preserve Zod issue paths. Modern configurations reject this callback; modern `logLevel` and `maxElicitationRounds` are rejected on legacy servers and at the adapter root.

Modern catalog subscriptions invalidate cached tools even without an application observer. Subscription setup failures reject the connection. SDK validators define logging levels; `setLoggingLevel()` remains legacy-only.

Per-server `resourceSubscriptions` selects resource URIs. Modern servers use `subscriptions/listen`; legacy servers use `resources/subscribe`. Unsupported subscriptions fail explicitly. Modern `reconnect` options are rejected, and losing a tool response does not automatically replay the operation. The README explains protocol negotiation; detailed interaction and migration guidance is preserved in the separate `langchain-ai/docs` draft.

## How was it tested?

- 348 tests passed after merging #11612's review fixes, with two existing skips; type checks and compiled examples pass
- Automatic negotiation with real modern HTTP, legacy HTTP, SDK 1.30 stdio, direct SSE, and HTTP-to-SSE fallback servers; explicit-modern mismatch rejection and transport fallback error boundaries
- Real modern resource delivery/filtering, unsupported capability rejection, legacy resource subscription routing, modern lost-response execution count, and configuration/type boundaries
- Legacy callback accept/decline/cancel, invalid answers, callback failures, missing handlers, and modern configuration rejection
- Real modern catalog changes with and without observers, cache freshness, subscriptions, request logging, and mixed-generation discovery/invocation
- Build, attw, publint, lint, formatting, and scoped anti-slop review pass

## Scope

Stacked on #11612. Invocation and default modern LangGraph interruption follow in #11642–#11645. OAuth workflow enhancements are outside this stack. Legacy callbacks wait within the active request and cannot survive a connection closing. No CI changes or new dependencies.


hntrl (commented):
should be using mcp sdks auto negotiation

hntrl (approved):
should be using mcp sdks auto negotiation

hntrl (approved):

hntrl (dismissed):
lets collapse changeset/docs changes, we don't need to be this verbose

hntrl (approved):

PR #11615, opened 2026-09-11
feat(mcp-adapters): checkpoint modern elicitation with LangGraph interrupts

byhow:
## Replaced by smaller draft PRs

The elicitation work is split into this review order, starting above #11614:

1. #11642 — tool invocation boundaries (649 changed lines).
2. #11643 — direct continuation rounds (721 changed lines).
3. #11644 — LangGraph elicitation interrupts (716 changed lines).
4. #11645 — recovery tests, runnable examples and documentation (541 changed lines).

OAuth additions are excluded from these replacements and deferred for separate design work. Earlier PRs in the stack are unchanged. The SDK revision-specific validator prerequisite for the major release remains open.

The replacement stack passes 398 tests with two existing skips, type checks, lint, formatting, ESM/CommonJS builds, attw and publint. The local HTTP example completes form/URL input and verifies decline/cancel. Each replacement was also checked independently at its own commit.

<details>
<summary>Original combined scope and verification (superseded)</summary>

> **Release prerequisite:** the SDK must publish revision-specific elicitation validators, and the adapter must replace its local modern URL-schema exception with that public API before the major release. A local upstream proposal is prepared and boundary-tested; no unpublished dependency or private SDK import is introduced here.

## What changes?

Modern MCP tools now pause a checkpointed LangGraph run for elicitation by default. No callback or `elicitationMode` setting is required. Direct modern calls continue bounded state-only rounds outside a graph, preserving effective arguments and cancellation; input requests outside a graph explain the checkpoint requirement. Explicit legacy servers continue using per-server callbacks.

Automatic protocol negotiation from #11614 selects this behavior from the negotiated protocol. The interrupt client advertises elicitation support on modern requests; a connection that negotiates legacy does not advertise callback support unless configured for it.

Each wire round checkpoints effective arguments and opaque server continuation with automatic retries disabled. Resume answers must match the pending keys and server schemas. Public interrupt values expose questions, while continuation data stays in the checkpoint. Tool invocation selects graph/direct execution and header policy at invocation time, so tools discovered outside a graph can later run inside one.

SDK validators define requests and answers through a Zod bridge that preserves defaults and issue paths. One documented modern URL-request schema compensates for the public SDK schema still requiring the removed legacy `elicitationId`; it does not invent an ID or import private SDK schemas. Internal interruption uses a capability guard and branded pending errors instead of class-identity checks.

`finishAuth(serverName, callbackParams, expectedState)` checks callback state, delegates issuer/token handling to the SDK, and clears the old connection/catalog after success. Applications own redirects, one-time state consumption, user binding, and credential storage. Direct modern calls retain header overrides; graph calls require connection-level headers/authentication.

The stack also adds explicit resource URI subscriptions and modern/legacy routing, rejects modern stream-replay configuration, and clarifies deprecated logging/SSE/DCR guidance. The SDK schema prerequisite is documented in the migration guide; this PR does not claim it is resolved.

## How was it tested?

- 403 tests passed, two existing skips; type checks, compiled examples, and the complete README checkpoint example pass
- Automatic mixed-generation negotiation and server-visible elicitation capabilities: modern requests advertise form and URL support; automatic legacy connections do not advertise reverse callbacks
- Modern and legacy HTTP/stdio servers together, graph pause/resume, accept/decline/cancel, URL requests without legacy IDs, invalid answers, state-only rounds, round limits, cancellation, and reconstructed adapters
- Direct continuation regression tests cover completion, round limits, cancellation, concurrent per-call header isolation, and a state-only round followed by a real question
- Local OAuth refresh, DCR/CIMD selection, authorization callbacks, state/issuer rejection, and scope step-up, plus native/web registration metadata and re-registration after refreshed discovery selects a different issuer
- Live HTTP reproductions: 67 checks each against source, ESM, and CommonJS; 20 stdio checks
- Wire-level protocol metadata, header declaration filtering, legacy results without `resultType`, malformed modern results, exact protocol error codes, arbitrary JSON artifacts, fractional schema constraints, and private resource-cache behavior
- Builds, attw, publint, lint, formatting, and scoped anti-slop review pass

## Scope and limits

S4 of four follow-up PRs, stacked on #11614. No CI changes or new dependencies. The SDK protected continuation and request-metadata overrides remain maintenance dependencies. Modern default interruption is an intentional JavaScript API choice, not a claim that Python already defaults to it.

MemorySaver reconstruction does not verify a separate-process persistent database. Applications must enforce checkpoint/thread/account ownership, and servers must make side effects safe to resume: a crash between a server action and checkpoint persistence cannot guarantee exactly-once execution. OAuth acceptance uses local fixtures, not production identity providers. Sampling and roots are outside this elicitation bridge; the modern revision deprecates those features and removes roots list-change notifications.

</details>


libs/langchain-mcp-adapters/src/client.ts
hntrl: why do we have oauth hatches in a PR for elicitations?

PR #11625, opened 2026-09-13
fix(openai): recognize gpt-6 models as reasoning models

### LINKED ISSUE #11624: ChatOpenAI drops explicit reasoning for gpt-6-astra despite reasoning-enabled model profile
### Checked other resources

### Example Code

This is network-free; it inspects the request produced by `ChatOpenAI`.

```ts

let body: Record<string, unknown> | undefined;

  model: "gpt-6-astra",
  apiKey: "test",
  useResponsesApi: true,
  reasoning: {
    effort: "max",
    summary: "auto",
  },
  configuration: {
    fetch: (_input, init) => {
      body = JSON.parse(String(init?.body));
      return Promise.resolve(
        Response.json(
          { error: { message: "stop after request capture" } },
          { status: 400 },
        ),
      );
    },
  },
});

await model.invoke("hello").catch(() => undefined);
console.log(body?.reasoning);
```

Actual output:
`undefined`

Expected output:
`{ "effort": "max", "summary": "auto" }`

The same configuration is forwarded for a model name already accepted by the
reasoning predicate, such as o3.

_No response_

### Checked version

`@langchain/openai@1.5.13`

The current `main` branch still appears affected. Commit `ac2e45c` added
`gpt-6-astra` to the model profile/catalog as reasoning-enabled, but Responses
request construction still filters explicit reasoning through a separate model-name
predicate that does not recognize this model.

### Cause

The model profile/catalog identifies `gpt-6-astra` as reasoning-capable, but
`BaseChatOpenAI._getReasoningParams()` first applies the separate reasoning-model
name predicate. That predicate recognizes o* models and non-chat gpt-5*
models, but excludes `gpt-6-astra`. As a result, explicit reasoning supplied by
the caller is discarded before `ChatOpenAIResponses.invocationParams()` builds
the request.
This leaves the profile and request-construction paths inconsistent.

### Suggested fix
Use the resolved model profile as the source of truth when deciding whether to
forward explicitly configured reasoning. A narrow `gpt-6-astra` predicate update
would also fix the immediate regression, but profile-driven behavior would avoid
the same mismatch for future reasoning models.

Please add request-construction tests covering:
1. `gpt-6-astra` with explicit reasoning forwards the object unchanged.
2. A model without configured reasoning does not gain a reasoning field.
3. Existing reasoning-model behavior remains unchanged.

*NOTE*: This bug has beed found by an AI agent while using your deepagents library and trying to call a GPT 6 Astra model. This whole issue has been written by that AI agent.

### System Info

I'm running it inside Deno runtime 2.9.6 on macOS 26.6

chiliec:
Fixes #11624

## What

`ChatOpenAI` drops explicitly configured `reasoning` for `gpt-6-astra`. The
profile added in #11555 marks it `reasoningOutput: true`, but
`_getReasoningParams()` first gates on `isReasoningModel()`, which only
recognises `o*` and non-chat `gpt-5*` names — so the caller's reasoning
object never reaches the Responses request body.

## Fix

Extend `isReasoningModel()` in `libs/providers/langchain-openai/src/utils/misc.ts`
with a `gpt-6` branch mirroring the existing `gpt-5` one (same `-chat`
exclusion). One line.

## Tests

`index.test.ts`:
- `isReasoningModel`: `gpt-6-astra` / `gpt-6` → true, `gpt-6-chat-latest` → false
- request construction: the issue's network-free repro — `gpt-6-astra` +
  `useResponsesApi` + `reasoning: { effort, summary }` → the request body
  carries the reasoning object unchanged (mirrors the sibling
  `reasoningEffort` tests using a mocked `fetch`)

Existing non-reasoning / gpt-5 / o-series cases unchanged.

## Validation

```
$ cd libs/providers/langchain-openai
$ npx vitest run
 Test Files  29 passed (29)
      Tests  345 passed (345)
 Type Errors  no errors
$ oxfmt --check / oxlint on the two touched files: clean
```

RED→GREEN: with the source change reverted (tests kept) both new cases fail
(`expected false to be true`, `expected undefined to deeply equal { effort:
'high', summary: 'auto' }`); restoring the one-line fix turns them green.

Happy to adjust.

### DIFF
--- a/.changeset/brave-owls-reason.md
+++ b/.changeset/brave-owls-reason.md
@@ -1,0 +1,5 @@
+---
+"@langchain/openai": patch
+---
+
+Recognize gpt-6 models as reasoning models so explicit `reasoning` config is forwarded.
--- a/libs/providers/langchain-openai/src/chat_models/tests/index.test.ts
+++ b/libs/providers/langchain-openai/src/chat_models/tests/index.test.ts
@@ -899,8 +899,14 @@ describe("ChatOpenAI", () => {
       expect(isReasoningModel("gpt-5.3-codex")).toBe(true);
     });
 
+    it("should return true for gpt-6 family models", () => {
+      expect(isReasoningModel("gpt-6-astra")).toBe(true);
+      expect(isReasoningModel("gpt-6")).toBe(true);
+    });
+
     it("should return false for gpt-5-chat models", () => {
       expect(isReasoningModel("gpt-5-chat-latest")).toBe(false);
+      expect(isReasoningModel("gpt-6-chat-latest")).toBe(false);
     });
 
     it("should return false for non-reasoning models", () => {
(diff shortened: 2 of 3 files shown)


mdnanocom:
up on this one

byhow (approved):
Thank you for looking into this issue and coming up with the targeted fix! Changes LGTM and will be merging this in.

PR #11627, opened 2026-09-13
fix: per-call service_tier parameter ignored when using Responses API

### LINKED ISSUE #11353: ChatOpenAI: per-call service_tier parameter is ignored when useResponsesApi is true
### Checked other resources

### Example Code

Runs without an API key: the injected `fetch` captures the request body before it leaves the process.

```ts

let sentBody: Record<string, unknown> = {};

const captureFetch: typeof fetch = async (_url, init) => {
  sentBody = JSON.parse(String(init?.body));
  throw new Error("request captured");
};

async function sentServiceTier(useResponsesApi: boolean) {
    model: "gpt-4.1-mini",
    apiKey: "sk-not-used",
    useResponsesApi,
    configuration: { fetch: captureFetch },
  });
  try {
    // `service_tier` is declared on BaseChatOpenAICallOptions and listed in `callKeys`
    await model.invoke("hi", { service_tier: "flex" });
  } catch {
    /* expected: the captured request never goes out */
  }
  return sentBody.service_tier;
}

console.log("completions:", await sentServiceTier(false)); // "flex"
console.log("responses:  ", await sentServiceTier(true));  // undefined  <-- bug
```

Output:

```
completions: flex
responses:   undefined
```

No exception, the parameter is silently dropped.

### Description

I am trying to select OpenAI's processing tier per call, so that a background turn can run on the
cheaper `flex` tier while an interactive one uses the default. `service_tier` is declared as a call
option (`BaseChatOpenAICallOptions.service_tier`, `chat_models/base.ts`) and is listed in
`BaseChatOpenAI.callKeys`, so I expect it to be honoured whichever API the model talks.

- **Expected:** `service_tier` passed via call options (`invoke(..., { service_tier })`,
  `withConfig`, or `modelSettings` in agent middleware) ends up in the request body for both the
  Completions API and the Responses API.
- **Actual:** it is honoured only on the Completions path. On the Responses path the value is read
  from the instance alone and the call option is discarded.

Related history:
#9897 reported that `service_tier` did not reach the API at all under `useResponsesApi`. #9898 fixed
that by adding `service_tier: this.service_tier` and the issue was closed as completed; #9953 was
closed unmerged. 

### System Info

@langchain/openai@npm:1.5.6
@langchain/core: 1.2.5
langchain: 1.5.5

platform: linux (6.8.0 x86_64)
node: v24.13.0
yarn: 4.18.0

KooshaPari:
Fixes #11353

## Problem
When using the ChatOpenAI Responses API, the per-call `service_tier` parameter was being ignored. The `invocationParams` method was using only the instance-level `this.service_tier`, so passing `service_tier` as a runtime argument (e.g., via `.invoke()` or `.bind()`) had no effect.

## Changes
- Modified `ChatOpenAIResponses.invocationParams()` to prefer the per-call `options.service_tier` over the instance-level `this.service_tier` using nullish coalescing
- This is consistent with how other per-call options (e.g., `previous_response_id`, `truncation`) are handled in the same method
- Also consistent with the Completions API which already supports per-call `service_tier` overrides

## Testing
- Added 3 new unit tests verifying: per-call override, instance fallback, and undefined when neither is set
- All existing tests continue to pass


KooshaPari:

PR #11631, opened 2026-09-13
fix: RunnableBranch._invoke no longer falls through to default on falsy results

### LINKED ISSUE #11583: bug(core): RunnableBranch._invoke falls through to default on falsy branch output (0/""/false)
### Checked other resources

### Example Code

import { RunnableBranch } from "@langchain/core/runnables";

const branch = RunnableBranch.from([
  [(x: number) => x > 0, (x: number) => 0],
  (x: number) => -1,
]);
console.log(await branch.invoke(5)); // expected: 0, actual: -1

N/A - no throw, silent wrong routing

### Description

* I'm trying to use RunnableBranch where a branch returns 0 / "" / false.
* I expect the matched branch output to be returned.
* Instead libs/langchain-core/src/runnables/branch.ts:170 does `if (!result)` so falsy output is discarded and default runs. Same file :236 in _streamIterator correctly uses `if (stream === undefined)`.

### System Info

platform: win32 x64 (Windows)
Node: v24.6.0
npm: 11.5.1
pnpm: 10.14.0
@langchain/core: 1.2.9
HEAD: 56130def8 chore: version packages (rc) (#11532)
libs/langchain-core/src/runnables/branch.ts:170

KooshaPari:
Fixes #11583

## Problem
The `RunnableBranch._invoke` method used `if (!result)` to determine whether a branch matched. This incorrectly executed the default branch when a matched branch returned a falsy value (`0`, `false`, `null`, empty string `""`).

## Changes
- Replaced the `if (!result)` check with a boolean `matched` flag
- The flag is set to `true` when any condition evaluates to `true`
- The default branch now only runs when `matched === false` (no condition matched)
- Added a type assertion for the return value since `result` may be falsy
- The `_streamIterator` path was already correct (checks `stream === undefined`)

## Testing
- Added 7 unit tests covering: default execution, first match, falsy results (0, false, empty string), negative numbers, and RunnableLambda conditions


KooshaPari:

PR #11636, opened 2026-09-14
feat(google): Agentic Video Understanding support

afirstenberg:
Adds support for Agentic Video Understanding (see https://blog.google/innovation-and-ai/models-and-research/gemini-models/introducing-agentic-video-in-gemini/)

* Adds support for `mediaProcessing` parameter in a video content block
* Changes API version for GCP / GEAP to "v1beta1", which is necessary for `mediaProcessing`
* Adds tests for same, specific to models that can accept it
* Adds gemini-3.8-flash to test suite
* Updates to latest API
* Changes how we save the OpenAPI spec for the API (sort parameters alphabetically so future diffs will be useful).

TODO:
* Check merge conflict
* Make sure thought signatures that are necessary are sent back
* Double-check the tests to make sure they are valid


libs/providers/langchain-google/src/converters/messages.ts
```diff
@@ -307,6 +307,28 @@ function stripMediaBlocksForFunctionResponse(content: unknown): unknown {
...
+    block.mediaProcessing ??
+    block.media_processing ??
+    (block.metadata as Record<string, unknown> | undefined)?.mediaProcessing ??
+    (block.metadata as Record<string, unknown> | undefined)?.media_processing;
+  if (typeof raw === "string") {
+    const upper = raw.toUpperCase();
+    if (
+      upper === "AGENTIC" ||
+      upper === "STATIC" ||
+      upper === "MEDIA_PROCESSING_UNSPECIFIED"
```
afirstenberg, line 323: This should be removed. Despite being valid in the API, it shouldn't be valid for a user to set it.

PR #11643, opened 2026-09-15
feat(mcp-adapters): answer MCP elicitation with graph interrupts

byhow:
Answer modern MCP elicitation with LangGraph interrupts, matching the Python adapter's replay contract.

A modern server needing form or URL input answers `tools/call` with an `input_required` result and expects the call retried with the answers. This raises each round as an `interrupt()` and resumes with `createMCPElicitationResume(interrupt, responses)`, which targets the answer at the interrupt it was written for rather than an unrelated one paused in the same run.

## Why the adapter drives the rounds

The SDK does this itself, and for a callback-shaped consumer that is the right answer — Mastra registers an `elicitation/create` handler and lets the driver retry, with no loop of its own. That path is unchanged here.

It cannot answer from a graph. The driver fulfils every pending request concurrently while LangGraph matches resume values to `interrupt()` calls **by order**, and `interrupt()` suspends by throwing, which the driver reads as a handler failure. So rounds are driven from the frame that issued the call, one question at a time. The Python adapter bypasses its own driver for the same two reasons.

Interception happens at the SDK's own `_resolveNonCompleteResult` seam — the counterpart to its `allowInputRequired` manual mode — so everything `callTool` wraps around a response still applies: SEP-2243 `Mcp-Param-*` header mirroring and output-schema validation, including the `toolDefinition` forwarding from #11612 that keeps the discovered descriptor and the validation view from disagreeing. Driving rounds through the public `request()` instead would forfeit both, since the SDK implements them with internal helpers.

## The replay contract

Resuming replays the tool call. The initial `tools/call` is issued again, the server repeats its question, and `interrupt()` returns the supplied answer instead of pausing. Only then is the answered follow-up sent, carrying the `requestState` from the response that execution just replayed.

So work a server performs before asking runs again on resume, `beforeToolCall` runs once per execution, and the adapter promises no exactly-once effects — **servers and hooks must be replay-safe**. Mastra's server-side implementation of this protocol leg documents the same constraint.

Answers are validated against the requested schemas. An invalid one re-asks on the same thread without another round trip, and one budget spans server rounds and re-asks together, so neither a server nor a caller can hold a run open.

An `input_required` response carrying no input requests is rejected as unsupported rather than polled, an aborted call reports the abort rather than the response it happened to receive, and calling such a tool outside a graph reports how to answer it instead of hanging.

## Scope

This supersedes #11644, #11655 and #11645, which are closed. Those four PRs were 3073 changed lines; this is 2581 — implementation **567**, tests 1977, changeset 37. Consolidating removed a `continuation.ts` module, a callback inversion five layers deep between the round loop and `client.callTool`, an `MCPContinuation` struct that carried state a local variable now holds, and a mutable closure memoising the client fork. No new implementation module is added — everything lands in existing files.

A follow-up commit restructures the round loop that consolidation produced. That loop drove server rounds and answer corrections together from one `for (;;)`, selecting between "fetch a question" and "ask it" through a sentinel and falling through from the first into the second, so its termination lived mid-body and the header could carry no condition. It is now two loops that each have one: `sendElicitationRound` converts the SDK's pending-input throw back into a value — the throw is required at the SDK seam, since it is what lets an `input_required` response escape `callTool`'s output-schema validation, but it need not shape the loop — and `askUntilAnswered` owns the re-ask, bounded by the shared round budget because a valid answer returns from inside it. Behaviour is unchanged: the refusals keep their order, messages and `pending` cause, and `maxRounds` still bounds questions asked rather than server calls, so an accepted answer is always delivered before the budget can end the run. Counter-bounding the outer loop would have broken that.

Everything here is elicitation. The lower-stack defect fixes and doc corrections that were previously in this branch (background reconnection reporting, the `onProgress` contract, the `mode: modern` error wording, the SSE changeset claim) moved to #11663, which stacks on this.

The test share is high because the behaviour only exists over the wire: a mock cannot produce an `input_required` result with a `requestState` the SDK will decode. Roughly 470 of those lines are server fixtures and graph harness before any assertion, and ~650 are suites recovered from the closed PRs — which earned their place by catching two real bugs in this refactor.

Base is #11642, which is approved and untouched.


libs/langchain-mcp-adapters/src/connection.ts
```diff
@@ -192,10 +193,13 @@ export class ConnectionManager {
...
-      { name: packageJson.name, version: packageJson.version },
-      protocolClientOptions(options)
-    );
+    const identity = { name: packageJson.name, version: packageJson.version };
+    const clientOptions = protocolClientOptions(options);
+
+    const mcpClient =
+      options.mode === "legacy"
+        ? new MCPClient(identity, clientOptions)
+        : new InterruptMCPClient(identity, clientOptions);
```
hntrl: this is kind of awkward aint it

libs/langchain-mcp-adapters/src/connection.ts
```diff
@@ -694,10 +698,9 @@ export class ConnectionManager {
     return new StdioClientTransport({
       command,
       args,
+      env,
       stderr,
       cwd,
-      // oxlint-disable-next-line no-process-env
-      ...(env ? { env: { PATH: process.env.PATH!, ...env } } : {}),
```
hntrl, line 652: i wont even ask

libs/langchain-mcp-adapters/src/continuation.ts
hntrl, line 1: whats a continuation

hntrl (commented):
it's not immediately clear to me why this is needed

libs/langchain-mcp-adapters/src/tools.ts
```diff
@@ -370,8 +396,21 @@ function createToolInvocationFactory(
     };
   }
 
-  const direct = executor(client, modern);
+  const unbound = executor(client, modern);
 
+  // Structural, not `instanceof`: duplicate module copies would break identity.
+  const rounds = (client as Partial<InterruptMCPClient>).maxElicitationRounds;
```
byhow: why cast instead of parse

libs/langchain-mcp-adapters/src/continuation.ts
byhow: its gone now, over engineered retry mechanism

libs/langchain-mcp-adapters/src/tools.ts
```diff
@@ -326,6 +327,8 @@ type CallToolArgs = {
    */
   afterToolCall?: ToolHooks["afterToolCall"];
   inputSchema: ToolInputSchema;
+  /** Graph task state for this execution, or absent outside a graph. */
+  graph?: { state: unknown };
```
hntrl: feels very awkward to have as input here?

libs/langchain-mcp-adapters/src/elicitation.ts
```diff
@@ -141,3 +180,387 @@ export function configureElicitation(
...
+
+  while (budget.asks < budget.max) {
+    source.signal?.throwIfAborted();
+    budget.asks += 1;
+
+    const answer = await answerSchema.safeParseAsync(interrupt(asked));
+    if (answer.success) return answer.data.responses;
+
+    asked = { ...asked, validationError: z.prettifyError(answer.error) };
+  }
```
hntrl: interrupts assume that the process will be dumped, so this budget tracking state is going to get lost

libs/langchain-mcp-adapters/src/elicitation.ts
```diff
@@ -141,3 +180,387 @@ export function configureElicitation(
...
+
+/**
+ * Raise one question until it comes back with usable answers.
+ *
+ * A rejected answer is re-asked on the same thread with the reason attached
+ * and costs no further round trip to the server: that question and its
+ * `requestState` are still current. Every attempt spends the shared budget,
+ * so the loop is bounded by it rather than running until something throws.
+ */
+async function askUntilAnswered(
```
hntrl: This behavior is novel but idt is strictly necessary

hntrl: I think a lot of this logic can be simplified if we get rid of the need to "ask until we get a valid answer"

hntrl (commented):
talked live, the elicitation mechanism is pretty complicated

libs/langchain-mcp-adapters/src/elicitation.ts
```diff
@@ -141,3 +138,227 @@ export function configureElicitation(
...
+    const responses = await buildResponses(
+      requests,
+      ask({
+        type: "mcp_elicitation",
+        server: source.server,
+        tool: source.tool,
+        requests,
+      }),
+      source
+    );
```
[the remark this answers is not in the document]
byhow: Fixed in 1f1ce5845 — the review is correct on both paths, and the `$10` → `$1,000` framing is exactly right.

Each interrupt now carries a `questionId` derived from the question's content plus the call's effective arguments, and `createMCPElicitationResume` binds the answer to it. It has to be a content hash rather than a generated id, because replay must reproduce it. The resume is then parsed as a whole against the question it answers — `questionId` as a `z.literal`, exactly the server's keys via `strictObject`, each answer against that question's requested schema — so a changed question or changed effective arguments fails the parse instead of sending the old approval with the new `requestState`.

This covers the `beforeToolCall` case you raised too, since the effective arguments are part of the identity.

The refused run is rolled back, so the original question stays pending and no consent is recorded against the operation nobody agreed to. Regression tests in `interrupts.test.ts` cover a changed argument set, a changed question, and that the pending question is untouched by the refusal.

libs/langchain-mcp-adapters/src/elicitation.ts
```diff
@@ -1,59 +1,59 @@
...
-  constructor(
-    info: ConstructorParameters<typeof Client>[0],
-    options: ConstructorParameters<typeof Client>[1],
-    private readonly onCancelled?: (
-      notification: CancelledNotificationParams
-    ) => void | Promise<void>
-  ) {
-    super(info, options);
-  }
+export const elicitationAnswerSchema = ElicitResultSchema;
```
byhow: can we not alias this unless this is used elsewhere

libs/langchain-mcp-adapters/src/elicitation.ts
```diff
@@ -141,3 +144,170 @@ export function configureElicitation(
...
+  ) => Promise<CallToolResult | InputRequiredResult>,
+  params: CallToolRequest["params"],
+  server: string,
+  tool: string,
+  signal?: AbortSignal
+): Promise<CallToolResult> {
+  signal?.throwIfAborted();
+  let result = await round(params);
+
+  while (isInputRequiredResult(result)) {
```
byhow, line 282: do we really need this loop or it is ok to drop this as well, as in this mcp adapter does not provide agent loop but only provide the primitives to build that

libs/langchain-mcp-adapters/src/tools.ts
```diff
@@ -393,36 +447,61 @@ function createToolInvocationFactory(
...
+
+  if (!parsed.success)
+    throw new ToolException(
+      `MCP tool "${toolName}" on server "${serverName}" returned output its schema rejects: ${z.prettifyError(parsed.error)}`
+    );
 }
 
 /** Keep the SDK's JSON Schema semantics while exposing a Zod parsing boundary. */
-function createToolInputSchema(
+function jsonSchemaParser<T>(
```
byhow, line 474: do we need this function at all? or we can do all these inline whenever we need to?

libs/langchain-mcp-adapters/src/tools.ts
```diff
@@ -549,13 +623,30 @@ async function _callTool(
...
-    );
+    const result = invocation.elicitation
+      ? await callToolWithElicitation(
+          round,
+          prepared.request,
+          serverName,
+          toolName,
+          config?.signal
+        )
+      : ((await round(prepared.request)) as CallToolResult);
```
byhow: why cast not parse

hntrl (approved):

PR #11644, opened 2026-09-15
feat(mcp-adapters): replay modern elicitation through LangGraph interrupts

byhow:
## What changes?

Modern tool calls pause with a LangGraph interrupt for form or URL input, using the same replay contract as the Python adapter.

### Rescoped: replay instead of checkpointed rounds

Both adapter-owned `task()` wrappers are removed — `mcp.tool.invocation` and `mcp.tool.round`. They existed to checkpoint each completed RPC round so a resume could reuse it rather than re-issue the request. That was a stronger guarantee than the ecosystem offers, it was not asked for, and it depended on released LangGraph nested-replay behavior (langchain-ai/langgraphjs#2679, still an upstream defect this PR does not fix).

Resuming now replays the call. The tool node re-runs from the top, the initial `tools/call` is issued again, `interrupt()` returns the supplied answer instead of pausing, and only then is the answered follow-up sent — carrying the `requestState` returned by the response **that execution just replayed**, never one saved from an earlier execution.

Measured against a real `createAgent` over HTTP:

| | request sequence |
|---|---|
| one round | `round 0` · `round 0` · `round 1 accept` |
| two rounds | `round 0` · `round 0` · `round 1 accept` · `round 0` · `round 1 accept` · `round 2 accept` |
| correction | `round 0` · `round 0` *(bad answer rejected before the wire)* · `round 0` · `round 1 accept` |

Rounds already answered replay too. So work a server performed before asking runs again on resume, and `beforeToolCall` runs once per execution. The changeset states plainly that there are **no exactly-once effects**; servers and hooks must be replay-safe.

### Targeted resume

`createMCPElicitationResume(interrupt, responses)` builds `{ [interruptId]: { responses } }`, and the inner envelope is strict, so an answer reaches the interrupt it was written for rather than an unrelated one paused in the same run. Requested-key and requested-schema validation and correction re-interruption are unchanged.

Targeting does not make replayed server work idempotent, and it does not isolate arbitrary `Promise.all` calls inside a single user node. Parallel MCP tool calls issued by `createAgent` are answered independently, and that is the path covered by tests.

### Header policy

The blanket rejection of `beforeToolCall` headers inside a graph is removed. It presented an adapter restriction as a LangGraph requirement. Under replay the hook runs again on each execution, so headers are recomputed for that execution through the client's own fork rather than restored from a checkpoint, and no credential is persisted between executions. Legacy elicitation keeps its callbacks.

A response that requests input while carrying no input requests is rejected rather than polled, inherited from #11643.

## How was it tested?

- 415 adapter unit tests pass, two existing skips; package and example type checks, lint and formatting pass.
- **New `agent.elicitation.test.ts`** — five regressions driving a real `createAgent` against a modern HTTP MCP server: the exact request sequence and state propagation for one and two rounds, correction after an invalid answer, rejection of an answer under the wrong key, and answering one parallel question while its sibling stays pending. This is the owning-head end-to-end coverage for this PR's behavior.
- `interrupts.test.ts` is updated to the same contract: state-only responses reject on every path, failures surface their own message now that no task repackages them, and the replayed request and the extra hook execution are asserted rather than hidden.
- **Removed** `routes parallel same-tool answers by invocation when rounds finish out of order`. It asserted isolation for `Promise.all` inside one user node, which the outer task provided and this contract deliberately does not promise; the supported `createAgent` parallel path is covered in the new file instead.
- `interrupts.test.ts` is reorganized into named describe groups, with the scenario-string setup replaced by a `harness()` helper taking named options.

## Scope

Third replacement slice of #11615, stacked on #11643. No OAuth additions, dependency or CI changes. The `maxElicitationRounds` server option still bounds the answered loop inside a single execution and still reaches the SDK as `inputRequired.maxRounds`.


libs/langchain-mcp-adapters/src/tests/interrupts.test.ts
```diff
@@ -0,0 +1,588 @@
...
+  "direct-state-then-question",
+  "bad-content",
+  "wrong-key",
+  "extra-key",
+  "concurrent",
+  "headers",
+  "outside-graph",
+  "outside-graph-complete",
+  "outside-graph-headers",
+  "without-checkpointer",
```
hntrl: this feels like a really awkward way to articulate test, maybe better suited for a describe block?

libs/langchain-mcp-adapters/src/continuation.ts
```diff
@@ -25,6 +35,27 @@ export class PendingMCPInput extends ns
...
+      return envelope;
+    }
+    const capabilities = ClientCapabilitiesSchema.parse(
+      envelope?.[CLIENT_CAPABILITIES_META_KEY] ?? {}
+    );
+    return {
+      ...envelope,
+      [CLIENT_CAPABILITIES_META_KEY]: {
+        ...capabilities,
+        elicitation: {
```
hntrl, line 80: isn't this meta object already attached as standard on the client

libs/langchain-mcp-adapters/src/continuation.ts
```diff
@@ -99,15 +132,32 @@ type RoundResult<T> =
       request: CallToolRequest["params"];
     };
 
-/** Continue state-only responses without replaying the initial hook or losing call headers. */
+type MCPInvocationContext = {
+  server: string;
+  tool: string;
+  maxRounds: number;
+  execution?: "direct" | "graph";
```
hntrl: why do we need to clarify this

libs/langchain-mcp-adapters/src/continuation.ts
```diff
@@ -149,19 +207,51 @@ export async function withMCPInterrupts<T>(
...
+                elicitationAnswerFor(request, modernElicitationAnswerSchema),
+              ] satisfies [string, ReturnType<typeof elicitationAnswerFor>]
+          )
+        )
       );
+
+      let question = elicitation;
+
+      while (true) {
+        const answer = await resumeSchema.safeParseAsync(interrupt(question));
```
hntrl, line 301: would love to see something e2e -- on the face this feels like it works but looping like this may potentially be unsafe? https://docs.langchain.com/oss/python/langgraph/interrupts

libs/langchain-mcp-adapters/src/tools.ts
```diff
@@ -347,6 +347,18 @@ type InvokeToolRound = (
...
+function toolExecutionContext(config?: RunnableConfig) {
+  try {
+    return { kind: "graph", state: getCurrentTaskInput(config) } satisfies {
+      kind: "graph";
+      state: unknown;
+    };
+  } catch {
+    return { kind: "direct" } satisfies { kind: "direct" };
+  }
+}
```
hntrl: why do we need to clarify between graph and direct tool calling

libs/langchain-mcp-adapters/src/tools.ts
```diff
@@ -377,7 +389,15 @@ function createToolInvocationFactory(
       ? client.withInterrupts.bind(client)
       : undefined;
 
-  function selectHeaderPolicy() {
+  function selectHeaderPolicy(config?: RunnableConfig) {
+    if (runInterrupts && toolExecutionContext(config).kind === "graph") {
+      return async (_headers: NonNullable<ToolCallModification["headers"]>) => {
+        throw new ToolException(
+          "Durable MCP calls require authentication and headers in the server connection configuration, not beforeToolCall header overrides"
+        );
```
hntrl: does it?

libs/langchain-mcp-adapters/src/continuation.ts
```diff
@@ -99,15 +132,32 @@ type RoundResult<T> =
...
-    signal?: AbortSignal;
+  source: MCPInvocationContext
+): Promise<T> {
+  if (source.execution === "direct") {
+    return runMCPContinuation(invoke, source);
   }
+
+  return task({ name: "mcp.tool.invocation", retry: { maxAttempts: 1 } }, () =>
+    runMCPContinuation(invoke, source)
+  )();
```
hntrl: this almost definitely isn't what we want

libs/langchain-mcp-adapters/src/tests/interrupts.test.ts
```diff
@@ -0,0 +1,588 @@
...
+      },
+    }),
+    config
+  );
+  expect(result.results).toEqual(["slow", "fast"]);
+  expect(answers).toEqual(["fast", "slow"]);
+  expect(calls).toEqual(["slow", "fast"]);
+});
+
+it.each(["state-only", "limit", "abort", "transport"])(
```
hntrl: same again

libs/langchain-mcp-adapters/src/tests/interrupts.test.ts
```diff
@@ -0,0 +1,588 @@
...
+  "bounds graph continuation without user questions: %s",
+  async (scenario) => {
+    const controller = new AbortController();
+    let calls = 0;
+    const State = Annotation.Root({ result: Annotation<string>() });
```
hntrl: use StateSchema instead of Annotations -- these are in model weights and we should probably deprecate

libs/langchain-mcp-adapters/src/tests/interrupts.test.ts
byhow: good point, now a describe block

byhow:

PR #11645, opened 2026-09-15
feat(mcp-adapters): run modern elicitation recovery through createAgent

byhow:
## What changes?

A runnable `createAgent` elicitation example, and the recovery coverage behind it, aligned with the replay contract from #11644.

### The example now uses createAgent

`modern_elicitation.ts` drove a hand-built `StateGraph`, which did not show the path applications actually take, and it answered questions with the flat resume map the replay redesign replaced. It now builds a `createAgent` run over the MCP tools and answers each question with `createMCPElicitationResume`, so the targeted `{ [interruptId]: { responses } }` envelope is what the example demonstrates.

It still rebuilds the adapter between rounds while keeping the checkpointed run. That is the recovery claim worth making: **losing the client does not lose the pending question.** The example's header comment and the examples README now state plainly that a resume replays the tool call, that pre-elicitation work and `beforeToolCall` therefore run again, and that there are no exactly-once effects.

The example's server URL accepts an `MCP_EXAMPLE_URL` override so a test can drive it on an ephemeral port.

### The example server authenticates its own retry state

`modern_server.ts` previously parsed `requestState` straight off the wire and trusted the phase it found. Under replay that value crosses the wire on every round, so it now seals and verifies it with `createRequestStateCodec`, bound to the MCP method, and validates each answer's content. The README says the demo key is ephemeral and single-process, and that a production application needs a stable shared key bound to its authenticated principal.

It also exports `createModernServer`/`listenModernServer`, matching the pattern applied to the calculator example in #11614.

## How was it tested?

- 429 adapter unit tests pass, two existing skips; `vitest run` reports no type errors. Lint and formatting pass.
- **The runnable example is verified as runnable.** Three cases spawn `modern_elicitation.ts` against the example server for `accept`, `decline` and `cancel` and assert its printed result. Because the example reconstructs the adapter between rounds, reaching that result also proves a resume survives losing the client.
- A protocol-level flow drives the example server directly: a forged phase is rejected, and decline and cancel outcomes are preserved across the three-question sequence.
- This head's own reconstruction coverage — a modern stdio interrupt resumed after rebuilding both the adapter and the server process, and accept/decline/cancel against real modern and mixed legacy/modern stdio servers — is moved onto the `{ responses }` envelope, with hook counts corrected for replay.

## Scope

Final replacement slice of #11615, stacked on #11644. No OAuth additions, dependency or CI changes.


libs/langchain-mcp-adapters/examples/modern_elicitation.ts
hntrl, line 1: createAgent elicitation?

byhow:

PR #11652, opened 2026-09-16
fix(langchain): resolve jumpTo "model" to the loop entry node

richardscarrott:
## Problem

The [middleware guide](https://docs.langchain.com/oss/javascript/langchain/middleware/custom#agent-jumps) defines the `'model'` jump target as "jump to the model node (or the first `before_model` hook)". JavaScript only does the first half. A `jumpTo: "model"` from an `afterModel` hook goes straight to the model node, so no `beforeModel` hook runs before that model call.

This applies to every hook, not just `afterModel`. A jump from a later `beforeModel` hook does not re-run the earlier ones, and jumps from `beforeAgent` and `afterAgent` land in the same wrong place.

A second bug in the same router hides the first one. When the last AI message has no tool calls, the `afterModel` router ends the run before it reads `jumpTo`, so the jump is discarded. Only the last `afterModel` hook to execute runs that check, so declaring another `afterModel` middleware changes whether a jump works at all.

## Fix

`"model"` now resolves to the entry of the agent loop, which is the first `beforeModel` hook when one exists and the model node otherwise. Tool results already return to that node.

The `afterModel` router also reads `jumpTo` before its exit check, so an explicit jump is no longer discarded.

## Python

This brings JavaScript in line with the docs and with Python, which already resolves `"model"` to the loop entry.

One divergence is deliberate. When a `beforeModel` hook is itself the loop entry and jumps to `"model"`, Python raises `KeyError: '<name>.before_model'`. Its guard drops the model destination from a node's declared edges when that node is the destination, but `_resolve_jump` still returns it, so LangGraph looks up a branch target that was never registered. This PR declares the self-edge instead, and the jump works.

Allowing it seems right. The hook can set state alongside the jump and read it back to decide whether to re-enter, which is how every other jump terminates. A middleware also has no way to know whether it happens to be declared first, so a jump that works in one position and crashes in another is hard to reason about.

## Follow-up

`jumpTo: "end"` has the same defect. The [guide](https://docs.langchain.com/oss/javascript/langchain/middleware/custom#agent-jumps) defines `'end'` as "jump to the end of the agent execution (or the first `after_agent` hook)", but an `"end"` jump from a `beforeModel` hook routes to `END` and skips `afterAgent`. Not fixed here, to keep this PR to one jump target.


richardscarrott:
For my own reference, here's some fallout issue worth looking into:

1. jumpTo: "end" skips afterAgent hooks — the serious one

Jumping to "end" from beforeModel (or a non-last afterModel) goes straight to __end__ instead of via the afterAgent nodes. Python runs them.

Real damage: modelCallLimitMiddleware uses its own afterAgent hook to reset the per-run counter. It terminates via beforeModel + jumpTo: "end", so the reset never runs. With a checkpointer, the first run that hits runLimit permanently bricks that thread — every later run dies with zero model calls.

Also affects moderation.ts (same shape) and toolCallLimit.ts (position-dependent).

Fix: pass exitNode instead of END in two routers. Careful: the sequence router is shared with afterAgent nodes, which must keep END.
Repro: scratch/jumpto-issue/src/end-jump-callcount.ts

---

2. Jumping to "tools" when the agent has no tools crashes

Both SDKs error, differently. JS throws at invoke time (Branch condition returned unknown or null destination); Python throws at build time with a better message naming the node. JS's router intends to exit, but the edge was never declared.

Open question: should it error or exit? Needs deciding, not just patching.

---

3. jumpTo isn't always cleared after use

A "end" jump returns a plain node name without clearing the sentinel, so a downstream afterAgent router can still see a stale jumpTo and act on it. There's already an unmerged branch for this: "clear consumed jumpTo before afterAgent routing". Closely related to No.1 — fix together.

---

4. Python bug: loop-entry self-jump crashes

In Python, if the first before_model middleware jumps to "model", you get KeyError: '<name>.before_model'. Their guard omits the destination from the declared list but still returns it. JS handles this correctly after our fix. Upstream report, not urgent.

libs/langchain/src/agents/ReactAgent.ts
```diff
@@ -1053,67 +1114,80 @@ export class ReactAgent<
...
+      const { jumpTo } = builtInState;
+      if (!jumpTo) {
         return nextDefault;
       }
-      const destination = parseJumpToTarget(builtInState.jumpTo);
-      if (destination === END) {
-        return END;
+      const destination = resolveJump(jumpTo, {
+        modelDestination: loopEntryNode,
+        endDestination: END,
```
hntrl, line 1174: am guessing this is what you mean by we don't resolve to afterAgent?

hntrl (approved):

libs/langchain/src/agents/ReactAgent.ts
richardscarrott: yeah exactly. I'll rework in the morning to sort that out.

richardscarrott:
I've moved to draft as I plan to circle back around to this with an additional fix for { jumpTo: "end" } routing before merging

PR #11655, opened 2026-09-17
test(mcp-adapters): cover graph elicitation end to end

byhow:
Split out of #11644 so the feature and its end-to-end coverage are each reviewable on their own. #11644 is now 949 lines; this is 560.

No implementation changes — tests and fixtures only.

**`agent.elicitation.test.ts`** drives graph-answered elicitation through `createAgent` against real MCP servers, asserting the exact replay transcript:

- resuming replays the initial `tools/call`, then answers it with that response's `requestState` (1 round = 3 server calls)
- a multi-round elicitation replays each earlier answered round (2 rounds = 6 calls)
- an invalid answer re-interrupts with a validation error and then accepts a correction; the malformed answer never reaches the server
- an answer under the wrong request key is rejected
- one of two parallel questions can be answered without answering its sibling

**`interrupts.test.ts`** gains the cases the adapter bounds itself: a graph continuation it cannot answer (state-only, abort and transport failure), and round bounding that replays each earlier request.

**`fixtures/agent-mcp-servers.ts`** holds the servers those tests need.


byhow:

PR #11663, opened 2026-09-17
fix(mcp-adapters): report silent failures and correct stale contracts

byhow:
Two silent failures and three stale contracts. Everything here was out of scope for the feature PRs in the stack. Stacks on #11643.

**167 changed lines** — implementation 45, tests 97, changesets 22, README 3.

## Background reconnection failed silently

`_attemptReconnect` swallowed every attempt in a bare `catch {}` and returned with no signal when `maxAttempts` ran out, so a server that never came back produced nothing: no log, no handler call, no notification. The `debug` dependency that used to carry this was removed package-wide, and nothing replaced it.

An exhausted budget now reports through `onConnectionError` with the last error. It deliberately does **not** throw — reconnection runs detached, so there is no caller to throw at, and throwing would surface as an unhandled rejection.

## Three contracts that no longer matched the code

- `onProgress` claimed callback failures were "reported through the adapter's debug logger" — a logger the package no longer has. They are ignored, so an observer cannot fail a tool call that already completed.
- The `maxElicitationRounds` and `logLevel` errors said `requires mode: modern`. The schema also accepts `auto`, which is the default, so the message sent users to a setting they did not need.
- `.changeset/quiet-lemons-answer.md` said SSE needs an explicit `mode: "legacy"`. SSE always speaks legacy whatever the mode and rejects `mode: "modern"` outright; the explicit setting is only needed to reach the legacy callbacks or `automaticSSEFallback`.

## README points at the docs site

An earlier revision of this PR documented elicitation and the error surface in the README directly. Those two sections are gone.

The reference content for this package belongs in [langchain-ai/docs](https://github.com/langchain-ai/docs): #11605 removed `libs/langchain-mcp-adapters/docs/client-reference.md` on exactly those grounds, and the README has linked to the docs site nowhere since. What remains here is the three-line `**Documentation**:` pointer that `langchain` and `langchain-classic` both carry. The prose goes up separately against the docs repo, where `src/oss/javascript/langchain/mcp.mdx` currently documents only the legacy callback path.

## Not included: re-listening a dropped subscription

An earlier revision of this PR added a backoff loop that re-established a dropped `subscriptions/listen` stream. It has been removed.

The underlying gap is real — the SDK does not re-listen, modern HTTP sets `maxRetries: 0`, and a dropped stream silently stops list-changed notifications, resource updates and the tool-catalog refresh behind them. But the fix was bespoke background concurrency in a package that otherwise delegates lifecycle to the SDK, and it introduced a shutdown stall that @open-swe caught: aborting only cancelled the backoff sleep, not an in-flight `listen()`, so `close()` could block on an unresponsive server during recovery.

Neither reference implementation does this. The Python adapter has no subscription code at all, and Mastra's client never calls `listen()`. Closing this properly is an API decision — most likely a callback so the application owns the retry policy, or `listen()` gaining a reconnect option upstream — not something to invent here.


libs/langchain-mcp-adapters/src/connection.ts
```diff
@@ -678,6 +649,35 @@ function serializeHeaders(
...
+function matchSdkHeaderCase(
+  headers: Record<string, string>
+): Record<string, string> {
+  return Object.fromEntries(
+    Object.entries(headers).map(([name, value]) => [
+      SDK_HEADER_SPELLING.get(name.toLowerCase()) ?? name,
+      value,
+    ])
+  );
+}
```
hntrl: why do we need this

hntrl (dismissed):

hntrl (approved):

PR #11664, opened 2026-09-17
fix(google): configure scopes for implicit ADC

### LINKED ISSUE #11654: Cannot create `ChatGoogle` LLM with `GOOGLE_APPLICATION_CREDENTIALS`
### Checked other resources

### Example Code

Before running, export `GOOGLE_APPLICATION_CREDENTIALS` in your shell environment, pointing to a file containing the application credentials. This example is [roughly the same as the docs](https://docs.langchain.com/oss/javascript/integrations/chat/google#credentials-through-gemini-enterprise-agent-platform-oauth-application-default-credentials-/-adc).

```typescript
import { ChatGoogle } from '@langchain/google'

const model = new ChatGoogle('gemini-3.7-flash', {
  platformType: 'gcp',
  maxOutputTokens: 2048,
  location: 'global',
  vertexai: true,
})

const result = await model.invoke([
  new SystemMessage(
    'You are a helpful assistant that translates English to French. Translate the user sentence.',
  ),
  new HumanMessage('I love programming.'),
])
```

```
err: {
  "type": "AuthError",
  "message": "No authentication method available. Please provide either an apiKey or credentials.",
  "stack":
      AuthError: No authentication method available. Please provide either an apiKey or credentials.
          at WebApiClient.fetch (file:///usr/src/app/node_modules/.pnpm/@langchain+google@0.2.6_@langchain+core@1.2.11_@opentelemetry+api@1.9.1_ws@8.21.0_/node_modules/@langchain/google/dist/clients/index.js:207:48)
          at file:///usr/src/app/node_modules/.pnpm/@langchain+google@0.2.6_@langchain+core@1.2.11_@opentelemetry+api@1.9.1_ws@8.21.0_/node_modules/@langchain/google/dist/chat_models/base.js:298:47
          at pRetry.onFailedAttempt.error.error (file:///usr/src/app/node_modules/.pnpm/@langchain+core@1.2.11_@opentelemetry+api@1.9.1_ws@8.21.0/node_modules/@langchain/core/dist/utils/async_caller.js:191:44)
          at pRetry (file:///usr/src/app/node_modules/.pnpm/@langchain+core@1.2.11_@opentelemetry+api@1.9.1_ws@8.21.0/node_modules/@langchain/core/dist/utils/p-retry/index.js:124:25)
          at queue.add.throwOnTimeout (file:///usr/src/app/node_modules/.pnpm/@langchain+core@1.2.11_@opentelemetry+api@1.9.1_ws@8.21.0/node_modules/@langchain/core/dist/utils/async_caller.js:191:31)
          at run (/usr/src/app/node_modules/.pnpm/p-queue@6.6.2/node_modules/p-queue/dist/index.js:157:104)
          at PQueue._tryToStartAnother (/usr/src/app/node_modules/.pnpm/p-queue@6.6.2/node_modules/p-queue/dist/index.js:105:17)
          at /usr/src/app/node_modules/.pnpm/p-queue@6.6.2/node_modules/p-queue/dist/index.js:171:18
          at new Promise (<anonymous>)
          at PQueue.add (/usr/src/app/node_modules/.pnpm/p-queue@6.6.2/node_modules/p-queue/dist/index.js:152:16)
  "name": "AuthError"
}
```

### Description

I'm trying to use this for a new app and not start with the deprecated `ChatVertexAI` from `@langchain/google-vertexai`, but I can't seem to get it to work with `GOOGLE_APPLICATION_CREDENTIALS`.

### System Info

```
/usr/src/app # cat package.json | grep langchain
    "@langchain/core": "^1.2.11",
    "@langchain/google": "^0.2.6",
    "@langchain/google-vertexai": "^2.3.2",
    "@langchain/pgvector": "^0.1.0",
    "@langchain/textsplitters": "^1.0.1",
/usr/src/app # cat package.json | grep langchain
    "@langchain/core": "^1.2.11",
    "@langchain/google": "^0.2.6",
    "@langchain/google-vertexai": "^2.3.2",
    "@langchain/pgvector": "^0.1.0",
    "@langchain/textsplitters": "^1.0.1",
/usr/src/app # cat /etc/alpine-release 
3.23.3
/usr/src/app # node --version
v24.13.0
/usr/src/app # pnpm --version
10.30.0
/usr/src/app # 
```

jackjin1997:
Fixes #11654

## Summary

- configure the provider-required OAuth scopes when the Node client falls back to Application Default Credentials
- preserve the existing API-key, inline-credential, explicit Google Auth, and custom-client paths
- add focused regression coverage plus a credential-gated Vertex AI integration test

## Testing

- `pnpm exec vitest run src/chat_models/tests/node.test.ts --reporter=verbose` (2 passed, type errors: none)
- `pnpm build` in `libs/providers/langchain-google` (build, attw, and publint passed)
- `pnpm lint` (passed)
- `pnpm exec oxfmt --check` on changed files (passed)
- credential-gated Vertex test collected and skipped locally because ADC credentials are unavailable
- package-wide `pnpm test` has 6 pre-existing failures in `src/converters/tests/messages.test.ts`; the same 6 failures reproduce with this patch reverted

## AI Disclosure

I used AI tooling to assist with investigation and implementation. I reviewed and validated all changes and am responsible for the result.


wbyoung (commented):
Looks like it'll fix my issue well. Thanks!

libs/providers/langchain-google/src/chat_models/tests/node.int.test.ts
```diff
@@ -0,0 +1,25 @@
...
+
+test.skipIf(!hasApplicationDefaultCredentials)(
+  "authenticates to Vertex AI with implicit application default credentials",
+  async () => {
+    const model = new ChatGoogle({
+      model: "gemini-2.5-flash",
+      platformType: "gcp",
+      vertexai: true,
+      location: "global",
+      maxOutputTokens: 8,
```
afirstenberg: I'd suggest omitting this. We're going to need to change the model soon, and `maxOutputTokens` can be limiting. (I spent 15 minutes trying to figure out why I was getting a false negative testing this, and this is why.)

libs/providers/langchain-google/src/chat_models/tests/node.int.test.ts
```diff
@@ -0,0 +1,25 @@
+import { expect, test } from "vitest";
+import { getEnvironmentVariable } from "@langchain/core/utils/env";
+import { ChatGoogle } from "../node.js";
+
+const hasApplicationDefaultCredentials = Boolean(
+  getEnvironmentVariable("GOOGLE_APPLICATION_CREDENTIALS") &&
+  getEnvironmentVariable("GOOGLE_CLOUD_PROJECT")
+);
+
+test.skipIf(!hasApplicationDefaultCredentials)(
```
afirstenberg: I'd suggest we remove the `skipIf`. I tested this for ADC as well as GOOGLE_APPLICATION_CREDENTIALS and it works. I'm testing it with GOOGLE_CLOUD_CREDENTIALS and there's something not quite working here, but I think that might not be a new problem.

My suggestion would be to do something like what we do in index.int.test.ts where we iterate over the possibilities to make sure everything gets tested so we don't introduce accidental regressions.

libs/providers/langchain-google/src/chat_models/tests/node.int.test.ts
```diff
@@ -0,0 +1,25 @@
...
+  "authenticates to Vertex AI with implicit application default credentials",
+  async () => {
+    const model = new ChatGoogle({
```
afirstenberg: I'd love it if we used the same method to construct the `ChatGoogle` as we do in the integration test. Or at least add the `GoogleRequestLogger()` since I use that to help debug issues.

afirstenberg (commented):
Thank you for diagnosing and doing this!

These are minor nits. Given there is an identified problem, the developer impacted confirms this patch works, and my tests concur:
#LGTM. @thushanth-bengre-langchain @hntrl

jackjin1997:
Thanks for testing it against the real credential paths — and sorry about the fifteen minutes; that false negative was mine to avoid.

All three are addressed in f077b604:

- **`maxOutputTokens` is gone.** It bought nothing here and, as you found, silently turns a model change into a failure that looks like an auth failure.
- **`skipIf` is gone**, replaced by an iteration over the credential sources the way `index.int.test.ts` iterates over models. The old gate required `GOOGLE_APPLICATION_CREDENTIALS` *and* `GOOGLE_CLOUD_PROJECT`, which meant a plain `gcloud auth application-default login` — the most common setup, and the one you were on — skipped the test entirely. Now `GOOGLE_APPLICATION_CREDENTIALS`, `GOOGLE_CLOUD_CREDENTIALS` and ambient ADC are separate cases, each running when the environment provides it, so a regression in one cannot hide behind another being unset. With no Google variables set, `vitest list --mode int` resolves to the ambient case alone, which is the one that used to be skipped.
- **Construction now matches the integration test**, including a `GoogleRequestRecorder` and a `GoogleRequestLogger` gated on `GOOGLE_LOG_REQUESTS`, copied from `buildTestCallbacks`.

On `GOOGLE_CLOUD_CREDENTIALS`: I think you are right that it is not new. That variable is read in `NodeApiClient`'s own credential resolution rather than going through `google-auth-library`'s implicit lookup, so it never reaches `ensureAuthScopes` — which is why this patch deliberately excludes it from `usesImplicitGoogleAuth`. It now has its own case in the test, so whatever is wrong there will surface as a named failure rather than as a silent skip. Happy to look at it separately if you want it filed.

`oxfmt --check` and `oxlint` are clean, and the package's `tsc --noEmit` error count is unchanged at 232, all pre-existing and outside these files.

thushanth-bengre-langchain:
Thank you for the fix @jackjin1997.

PR #11669, opened 2026-09-18
feat(anthropic): accept a SystemMessage anywhere in the message list

richardscarrott:
## Problem

`ChatAnthropic` throws "System messages are only permitted as the first passed message" for any `SystemMessage` that isn't at index 0.

[Anthropic now accepts a system instruction part-way through a conversation](https://platform.claude.com/docs/en/build-with-claude/mid-conversation-system-messages) as a `role: "system"` entry in the messages array. It applies from that point on, and adding one leaves the prompt cache for the earlier turns intact.

## Fix

Hoist the contiguous run of system messages starting at index 0 into the top-level `system` field. Send every other system message at its own position as a `role: "system"` entry.

No model conditioning and no placement pre-validation. If a position is illegal or the model doesn't support the feature, the provider returns a 400 and that error propagates.


libs/providers/langchain-anthropic/src/utils/message_inputs.ts
```diff
@@ -454,6 +448,50 @@ function _formatContent(message: BaseMessage, toolCalls?: ToolCall[]) {
...
 
+/**
+ * Normalizes the content of a system message for the wire.
+ *
+ * Shared by the hoisted top-level `system` field and by in-place
+ * `role: "system"` entries, so both strip the same framework-internal fields.
+ *
+ * String content stays a string: the `system` field sits near the front of the
+ * hashed cache prefix, so promoting it to a block array would invalidate every
+ * existing cached prompt. Text blocks are reduced to the keys Anthropic
```
richardscarrott, line 459: I don't know if this is true tbh. Does this result in different caches
```
system: "Be nice",

// vs

system: [{ type: "text", text: "Be nice" }]
```

hntrl (approved):

PR #11671, opened 2026-09-18
fix(typesafe): make tsc pass, and make it actually run

byhow:
Stacked on #11670.

`tsc --noEmit` reports **33 errors** in this package — 31 bare `process.env` reads and 2 `node:util` imports across the tests, none resolving because no Node ambient types are configured. **No CI job typechecks this package**, so they were invisible. That blindness had already hidden 6 genuine new `TS2349` errors, which reached a final review undetected.

## Two parts, neither of which works alone

**`@types/node` at `^26.4.0`** — the version seven other packages in this repo already declare. On its own it changes nothing: measured 33 errors before and after, because TypeScript's automatic `@types/*` inclusion does not fire under `typescript ~7.0.2` with this package's small dependency graph. `langchain-openai` works with just the bare dependency only because its much larger graph pulls the globals in transitively.

**`"types": ["node"]` in the tsconfig.** No sibling provider sets `types`, and none uses a `/// <reference types="node" />` directive, so there was no precedent either way. One tsconfig line beat eight unexplained directives scattered through test files. The comment explains why this package alone needs it.

Together: **33 → 0**.

## Keeping it at zero

`src/tests/types.test-d.ts` exists so vitest's typecheck actually runs. `typecheck: { enabled: true }` was already set in `vitest.config.ts`, but vitest only invokes tsc when the package contains at least one `*.test-d.ts` file — without one it prints `Type Errors  no errors` while checking nothing.

Verified non-vacuous: injecting a type error into `modelRouter.ts` makes `npx vitest run` exit **1**, where before it exited **0**.

Worth knowing for anyone reading that output: the `Type Errors` line only ever counts errors inside `*.test-d.ts` assertions. Source errors appear on a separate `Errors` line and drive the exit code. Reading `Type Errors  no errors` as "the package typechecks" is what let the six errors through.

## Verification

`tsc --noEmit` exit 0, `oxfmt --check` clean, 140 unit tests, 24 integration against the live API.


byhow:
Closing as out of scope.

The premise was wrong. I framed the 33 `TS2591` errors as an oversight specific to this package, but every provider has them — xai 98, openai 96, groq 39, this one 33, anthropic 21. No CI job runs `tsc --noEmit` and no provider package is clean, so fixing one in isolation is unilateral divergence rather than a repair.

The half worth keeping is the observation, not this patch: vitest prints `Type Errors  no errors` when its typecheck never runs, and it only runs when a package contains a `*.test-d.ts` file. Six genuine `TS2349` errors survived nine reviews in #11670 because of that. Fixing it belongs at the provider level for all packages together, not here.

PR #11675, opened 2026-09-19
fix(core): preserve structured tool inputs in traces

hntrl:
Preserve object-valued tool arguments as structured tracer inputs so LangSmith can render individual fields instead of treating a JSON string as a user message.

- thread structured inputs through tool-start callbacks
- retain the existing string callback argument and fallback for compatibility
- cover structured multi-field tool inputs with a regression test
- add a patch changeset for `@langchain/core`

Made by [Open SWE](https://github.com/langchain-ai/open-swe) · [view thread](https://openswe.vercel.app/agents/e306d924-0c42-5183-8992-2c4455f9c285) · openai:gpt-5.6-sol (medium)


libs/langchain-core/src/callbacks/base.ts
```diff
@@ -195,7 +195,8 @@ abstract class BaseCallbackHandlerMethodsClass {
     tags?: string[],
     metadata?: Record<string, unknown>,
     runName?: string,
-    toolCallId?: string
+    toolCallId?: string,
+    inputs?: Record<string, unknown>
```
hntrl: yuck, how hard would it be to update the `input` signature to widen and accept record objects?

libs/langchain-core/src/tools/index.ts
```diff
@@ -294,8 +294,13 @@ export abstract class StructuredTool<
...
       this.toJSON(),
-      // Log the original raw input arg
-      typeof arg === "string" ? arg : JSON.stringify(arg),
+      typeof inputForValidation === "object" &&
+        inputForValidation !== null &&
+        !Array.isArray(inputForValidation)
+        ? (inputForValidation as Record<string, unknown>)
+        : typeof arg === "string"
+          ? arg
+          : JSON.stringify(arg),
```
hntrl: this ternary is a little crazy, any way we can make it better?

libs/langchain-core/src/tools/index.ts
```diff
@@ -292,10 +292,18 @@ export abstract class StructuredTool<
...
 
+    let callbackInput: string | Record<string, unknown> =
+      typeof arg === "string" ? arg : JSON.stringify(arg);
+    if (
+      typeof inputForValidation === "object" &&
+      inputForValidation !== null &&
+      !Array.isArray(inputForValidation)
+    ) {
+      callbackInput = inputForValidation as Record<string, unknown>;
+    }
```
hntrl:
```suggestion
    let callbackInput: string | Record<string, unknown> | undefined;
    if (
      typeof inputForValidation === "object" &&
      inputForValidation !== null &&
      !Array.isArray(inputForValidation)
    ) {
      callbackInput = inputForValidation;
    }
    if (!callbackInput) {
      if (typeof arg === "string") {
        callbackInput = arg;
      } else {
        callbackInput = JSON.stringify(arg);
      }
    }
```

PR #11680, opened 2026-09-21
fix(google): authenticate service account credentials correctly

### LINKED ISSUE #11677: Node ChatGoogle token issues with GOOGLE_CLOUD_CREDENTIALS
### Privileged issue

### Issue Content

I was testing a modified version of the node integration test for #11664 that also allows for the other forms of authentication.

While testing with GOOGLE_CLOUD_CREDENTIALS set, it was giving an error:

```
RequestError: Request had invalid authentication credentials.
```

Investigating, the issue is in `NodeApiClient.fetch()` [where it calls](https://github.com/langchain-ai/langchainjs/blob/bd44054c8b269e17dc4296218f9eb9897811ce08/libs/providers/langchain-google/src/clients/node.ts#L310) `getGCPCredentialsAccessToken()` without an `await`. 

Adding the `await` seems to resolve the issue, but then causes the error

```
AuthError: Invalid OAuth scope or ID token audience provided.
```

Which is caused [inside gcp-auth.ts](https://github.com/langchain-ai/langchainjs/blob/bd44054c8b269e17dc4296218f9eb9897811ce08/libs/providers/langchain-google/src/utils/gcp-auth.ts#L256), `getGCPCredentialsAccessToken().promise()`

The JWT created by `getCGPCustomToken()` seems to be setting "aud" to "https://oauth2.googleapis.com/token" and doesn't seem to contain a scope.

I'm continuing to look, but need to step away shortly. Documenting this here in case @jackjin1997 or @thushanth-bengre-langchain have a chance to look before I do.

jackjin1997:
Fixes #11677.

Two defects sit on top of each other on the `GOOGLE_CLOUD_CREDENTIALS` path, exactly as @afirstenberg traced.

## The promise in the header

`getGCPCredentialsAccessToken` is `async`, and `NodeApiClient.fetch` interpolated it straight into a template literal. Every request went out as:

```
Authorization: Bearer [object Promise]
```

which Google answers with `RequestError: Request had invalid authentication credentials`. The new client test pins this: without the `await` it asserts against the literal string `Bearer [object Promise]`.

## The missing scope claim

Awaiting the token exposes the second failure, `AuthError: Invalid OAuth scope or ID token audience provided`. `getGCPCustomToken` signs an assertion with `iss`, `aud`, `sub`, `iat` and `exp` — but no `scope`.

For the `urn:ietf:params:oauth:grant-type:jwt-bearer` exchange this path performs, [Google's service account documentation](https://developers.google.com/identity/protocols/oauth2/service-account) lists `scope` as a required member of the claim set, "a space-delimited list of the permissions that the application requests", alongside `aud`, which "is always `https://oauth2.googleapis.com/token`" for an access token request. So the audience was already correct and the absent scope is what the error is reporting. Nothing downstream can fill it in, because this path mints its own JWT rather than delegating to `google-auth-library`.

## Changes

- `await` the access token before building the header.
- Thread scopes into `getGCPCredentialsAccessToken` and `getGCPCustomToken`, and set the `scope` claim on the assertion.
- Add the scopes to the token cache key. The cache keyed on `token_uri` plus `private_key_id`, so without this a token minted for one scope set would be served to a caller that asked for another.
- `NodeApiClient` reuses `googleAuthOptions.scopes` when they are configured and otherwise falls back to `VERTEX_AI_AUTH_SCOPES`, since service account credentials target Vertex.

I kept this entirely inside the auth layer so it does not touch `chat_models/node.ts` and can merge independently of #11664, in either order. If you would rather the scopes that #11664 computes be passed down explicitly once that lands, say so and I will follow up.

## Testing

Seven new tests, all of which fail on `main` for the right reason:

- the header carries the resolved token rather than `Bearer [object Promise]`
- the default scope is `cloud-platform`, and configured `googleAuthOptions.scopes` win
- the assertion carries a space-delimited `scope` claim
- two different scope sets do not share a cached token

Package suite: 199 passing. Six failures remain in `src/converters/tests/messages.test.ts`; they reproduce identically on `bd44054c8` and are untouched here. `oxfmt --check` and `oxlint` are clean, and `tsc --noEmit` is unchanged at 232 pre-existing errors elsewhere in the package.

I could not exercise the real `GOOGLE_CLOUD_CREDENTIALS` flow — no service account to hand — so the end-to-end confirmation is yours to make. The unit coverage pins both mechanisms and the cache behaviour.

## AI Disclosure

Diagnosed and written with AI assistance, starting from @afirstenberg's trace in #11677.


afirstenberg:
I've tested this with the new node.int.test.ts tests against all three OAuth methods. All pass.

#LGTM
@thushanth-bengre-langchain 

Thanks for your work on this @jackjin1997 !

thushanth-bengre-langchain:
Looks good. Thanks for the fix @jackjin1997

